PR #673, opened 2024-07-11
Core services tracing spans

danielSanchezQ:
First step on structure logs in services.
This PR adds:

* A general span with a debug id for the node (just debug mode)
* A base span for consensus
* A base span for da verifier

### DIFF
--- a/nodes/nomos-node/Cargo.toml
+++ b/nodes/nomos-node/Cargo.toml
@@ -26,8 +26,8 @@ nomos-network = { path = "../../nomos-services/network", features = ["libp2p"] }
 nomos-api = { path = "../../nomos-services/api" }
 nomos-log = { path = "../../nomos-services/log" }
 nomos-mempool = { path = "../../nomos-services/mempool", features = [
-  "mock",
-  "libp2p",
+    "mock",
+    "libp2p",
 ] }
 nomos-metrics = { path = "../../nomos-services/metrics" }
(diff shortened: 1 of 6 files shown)


bacv:
Looks like `rustls@0.23.9` has issues with `rust-libp2p`, we can either wait until https://github.com/libp2p/rust-libp2p/issues/5487 is resolved or upload Cargo.lock file with pinned versions.

It seems that the da services integration tests are failing too (never quits), this might not be related to this issue, I'll investigate it.

bacv:
`cargo test -p nomos-da-tests -F libp2p` succeeds with pinned rustls version. These tests use network service for communication between nodes.

danielSanchezQ:
> Looks like `rustls@0.23.9` has issues with `rust-libp2p`, we can either wait until [libp2p/rust-libp2p#5487](https://github.com/libp2p/rust-libp2p/issues/5487) is resolved or upload Cargo.lock file with pinned versions.
> 
> It seems that the da services integration tests are failing too (never quits), this might not be related to libp2p issue, I'll investigate it.

Can we pin the `rustls` version instead? I would like to avoid versioning the `Cargo.lock` file.

bacv:
> Can we pin the rustls version instead? I would like to avoid versioning the Cargo.lock file.

Since libp2p recently started using deps from the dependent project workspace, we should be able to pin it in the nomos-libp2p Cargo.toml

bacv (approved):

PR #676, opened 2024-07-19
chore: Da kzgrs unit tests update

romanzac:
Batch of tests for data availability - kzgrs to improve coverage


bacv (approved):

nomos-da/kzgrs/src/common.rs
```diff
@@ -179,4 +182,55 @@ mod test {
...
+        assert_eq!(
+            format!("{}", BlstError(BLST_ERROR::BLST_BAD_SCALAR)),
+            "Bad scalar value"
+        );
+
+        assert_eq!(
+            format!("{}", KzgRsError::from(BLST_ERROR::BLST_SUCCESS)),
+            "BLST error: Operation successful"
+        );
+    }
```
danielSanchezQ, line 225: I'm wondering why do we test this. As we are using a wrapping library, shouldn't it be tested in the library? :thinking:

nomos-da/kzgrs/src/common.rs
```diff
@@ -179,4 +182,55 @@ mod test {
...
+
+    #[test]
+    fn test_compute_roots_of_unity() {
+        let roots = compute_roots_of_unity(10);
+
+        let domain: GeneralEvaluationDomain<Fr> = GeneralEvaluationDomain::new(10).unwrap();
+        let frs: Vec<Fr> = domain.elements().take(10).collect();
+
+        assert_eq!(roots, frs);
+    }
```
danielSanchezQ, line 235: Also I'm curious about this test. I guess this will make sense if at some point the implementation change (?). Otherwise is just using whats inside the function that is testing right? 
Also to increase the coverage probably?

danielSanchezQ (approved):

nomos-da/kzgrs/src/common.rs
romanzac: Yes, test is improving coverage. To have it more contextual, and valuable, I need to reach higher knowledge about project. My learning curve is a bit in collision with dev changes. I aim so far to help with unit tests, later take over integration tests. I'd like to get more "involved", but that requires more of your time, mostly to answer my questions, describe concepts. 

I infer meaning from code, if it is stable enough, and later create integration tests mostly independent. Alternatively, I can get more involved, and would need some knowledge transfer and could help more, faster. Which mode would you prefer ?

nomos-da/kzgrs/src/common.rs
romanzac: Which library do you refer to  ? This test is for coverage increase of fmt::Display implementation. Test with interesting context would be more valuable, I agree.

nomos-da/kzgrs/src/common.rs
danielSanchezQ: > Yes, test is improving coverage. To have it more contextual, and valuable, I need to reach higher knowledge about project. My learning curve is a bit in collision with dev changes. I aim so far to help with unit tests, later take over integration tests. I'd like to get more "involved", but that requires more of your time, mostly to answer my questions, describe concepts.

No criticism, I was just trying to understand the approach.

> I infer meaning from code, if it is stable enough, and later create integration tests mostly independent. Alternatively, I can get more involved, and would need some knowledge transfer and could help more, faster. Which collab. mode would you prefer ?
We can work on acquiring the knowledge if you need it. That is not a problem. We can create issues and discuss about things until we have a batch of tests planned or something like that. Wdyt?

PR #678, opened 2024-07-23
DA Prost integration

bacv:
Protobuf integration and new `nomos-da-subnets` crate for containing building blocks required for DA Subnets services.


.github/workflows/build-test.yml
```diff
@@ -21,6 +21,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: true
+      - uses: arduino/setup-protoc@v3
```
danielSanchezQ, line 24: What is this? what is it used for?

Cargo.toml
```diff
@@ -4,6 +4,7 @@ members = [
     "nomos-da/kzgrs",
     "nomos-da/kzgrs-backend",
     "nomos-da/storage",
+    "nomos-da/subnets",
```
danielSanchezQ, line 7: I would prefer to call it `network`

nomos-da/subnets/proto/dispersal.proto
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    CloseReason reason = 1;
+}
+
+message SessionReq {
+    oneof message_type {
+        CloseMsg close_msg = 1;
+    }
+}
```
danielSanchezQ, line 75: Just realize, is this really needed? 
* Why would a node alert when is closing connections?
* Why would a node tell another node of changing subnet?
* Why the subnet sample fails message? When is that triggered?

Probably we can go without this. I probably should have commented this in the original poc PR. But we can change that later.

nomos-da/subnets/proto/dispersal.proto
```diff
@@ -0,0 +1,87 @@
...
+
+// WRAPPER MESSAGE
+
+message DispersalMessage {
+    oneof message_type {
+        DispersalReq dispersal_req = 1;
+        DispersalRes dispersal_res = 2;
+        SampleReq sample_req = 3;
+        SampleRes sample_res = 4;
+        SessionReq session_req = 5;
```
danielSanchezQ, line 85: Yeah, not sure we should have a session message for now. Until we define the internal subnet protocol itself.

nomos-da/subnets/src/proto/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+    R: AsyncReadExt + Unpin,
+{
+    let mut length_prefix = [0u8; MAX_MSG_LEN_BYTES];
+    reader.read_exact(&mut length_prefix).await?;
+    let data_length = u16::from_be_bytes(length_prefix) as usize;
+
+    let mut data = vec![0u8; data_length];
+    reader.read_exact(&mut data).await?;
+    M::decode(Bytes::from(data)).map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))
+}
```
danielSanchezQ, line 40: Nitpick: I have the feeling that this can be hook to serde somehow. Or is this standard for prost?

bacv: Protobufs are just a way to serialize messages, nothing else. It's up to the user to build a protocol out of it.
Prost already provides ser/de for message types via `encode`/`decode` methods. The packing that is done here is just to know how many bytes to pull from the reader when data is coming from a wire. As prost is only responsible for ser/de, we need to take care of the "what to parse" and "when" ourselves.

nomos-da/subnets/proto/dispersal.proto
bacv: This protocol is for communication between Executor and DA Node (https://github.com/logos-co/nomos-specs/pull/100), not for Node to Node.

nomos-da/subnets/proto/dispersal.proto
bacv: This protocol is for communication between **Executor and DA Node** (https://github.com/logos-co/nomos-specs/pull/100), not for Node to Node. 

>    Why would a node alert when is closing connections?

Executor, or DA Node, will want to inform each other about the closing connection.
Executor would want to know this in advance because even though quic has ways to determine the liveliness of connection (keepalive ping/timeout) they are time based. Same message type could be used in both directions.

>    Why would a node tell another node of changing subnet?

It would tell the Executor that.

>    Why the subnet sample fails message? When is that triggered?

The executor, validator or light node would tell this reason to the DA Node in case of failed sampling. If we are planning to implement any kind of ranking, such message would allow DA Node to self regulate itself.

Cargo.toml
bacv: Maybe `dispersal`? As it would encapsulate the executors dispersing and da nodes replicating inside the subnets?

.github/workflows/build-test.yml
bacv: To add `protoc` binary into the path on `linux/mac/windows` environments (used to compile the `.proto` files).

nomos-da/subnets/proto/dispersal.proto
danielSanchezQ: > > Why would a node alert when is closing connections?
> 
> Executor, or DA Node, will want to inform each other about the closing connection. Executor would want to know this in advance because even though quic has ways to determine the liveliness of connection (keepalive ping/timeout) they are time based. Same message type could be used in both directions.

Cool, thanks for this explanation.

> 
> > Why would a node tell another node of changing subnet?
> 
> It would tell the Executor that.
I do not think so, at most a validator can refuse the connection because maybe is too busy, and then the executor should select a different one in the subnetwork. A changing subnetwork message is irrelevant IMO. We need to revisit this.

> 
> > Why the subnet sample fails message? When is that triggered?
> 
> The executor, validator or light node would tell this reason to the DA Node in case of failed sampling. If we are planning to implement any kind of ranking, such message would allow DA Node to self regulate itself.

The node do not need to a message for this, if a node receives sampling requests and he do not have the chunk then he knows already. This is something internal. As a node you cannot trust the exterior neither.

nomos-da/subnets/src/proto/mod.rs
danielSanchezQ: > Protobufs are just a way to serialize messages, nothing else. It's up to the user to build a protocol out of it. Prost already provides ser/de for message types via `encode`/`decode` methods. The packing that is done here is just to know how many bytes to pull from the reader when data is coming from a wire. As prost is only responsible for ser/de, we need to take care of the "what to parse" and "when" ourselves.

Uhm, I would like to understand this properly as I may be missing something.But, I do not really think this is the best approach, but we can work on it together later on.

nomos-da/subnets/proto/dispersal.proto
bacv: This will be improved in the technical spec, the main idea here is that both parties in the ideal conditions would inform about the closing connection. I would consider as a place holder for that, with potentially more (or less) reasons that we might come up. With protobufs it will be easy to change it.

Cargo.toml
danielSanchezQ: I think this will be a core of the da-network. We will have stuff for dispersal and sampling inside. Also it was  a good choice to have this separated crate as it can be part of the sdk later on. So, i think we will have:
```
nomos-da /network
    /core
        /backends
            /libp2p
    /assignation
    /sampling
    /dispersal
```
More or less. Does ti makes sense?

bacv: This makes sense, I'll move this code in this pr to `nomos-da/network/dispersal` crate `nomos-da-dispersal`.

danielSanchezQ (approved):

PR #680, opened 2024-07-24
Da: network core

danielSanchezQ:
* Sketch up da network libp2p stack
* Implemented first rough version of broadcast protocol


nomos-da/network/subnetworks-assignations/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+use std::collections::HashSet;
+use std::hash::Hash;
+
+pub trait MembershipHandler {
```
bacv, line 4: As of now I feel that this trait would better fit inside the `network/core/membership/mod.rs`. The actual functions to calculate the subnetwork idx from the pov key could also live there (assuming `nomos-da` crates are considered as a libraries).

The composition of these parts will happen in `nomos-services`, where we'll be able to provide different adapters (eg. `MockMemebershipAdapter`?

nomos-da/network/core/src/broadcast/behaviour.rs
```diff
@@ -0,0 +1,116 @@
...
+
+use subnetworks_assignations::MembershipHandler;
+
+use super::handler::{BroadcastHandler, DaMessage, HandlerEventToBehaviour};
+
+pub type SubnetworkId = u16;
+
+/// Nomos DA BroadcastEvents to be bubble up to logic layers
+pub enum BroadcastEvent {
+    IncomingMessage { peer_id: PeerId, message: DaMessage },
```
bacv, line 19: Looks like we won't need packing after all. Seems that libp2p will take care of it. Will update messages crate if true.

bacv (approved):
Great progress, it would be great to have a simple service as a next step to integrate the messages and quic transport.

danielSanchezQ:
> Great progress, it would be great to have a simple service as a next step to integrate the messages and quic transport.

I think a test instead will be better, shouldn't be too complicated, just spawning 2 "nodes" as tasks and interchange a few random blobs.

nomos-da/network/core/src/broadcast/behaviour.rs
danielSanchezQ: > Looks like we won't need packing after all. Seems that libp2p will take care of it. Will update messages crate if true.

It makes sense is like that, Thanks for taking a look at it.

nomos-da/network/subnetworks-assignations/src/lib.rs
danielSanchezQ: Yeah, the trait would fit there as well. But I think as those algorithms are "simple" but will change in the future they deserved their own crate. Then in the nomos-services we just compose them indeed.

bacv: 🤝

bacv (approved):

danielSanchezQ:
Unrelated lint. Merging

PR #684, opened 2024-07-26
CI: Rust 1.80

bacv:
Added a `allow[dead_code]` for `Nonce` in consensus for Clippy.
Unused methods in da libp2p will be handled in a separate PRs.


danielSanchezQ (approved):
Unrelated errors that are gonna be fixed in further prs

PR #685, opened 2024-07-31
Da: network replication fix connection

danielSanchezQ:
Restructure handler
Wake up polls so they do not hang
Added basic test

### DIFF
--- a/nomos-da/network/core/src/replication/handler.rs
+++ b/nomos-da/network/core/src/replication/handler.rs
@@ -2,8 +2,9 @@ use std::io::Error;
 use std::task::{Context, Poll};
 
 use futures::Future;
+use futures::future::BoxFuture;
 use futures::prelude::*;
-use libp2p::{Stream, StreamProtocol};
+use libp2p::{PeerId, Stream, StreamProtocol};
 use libp2p::core::upgrade::ReadyUpgrade;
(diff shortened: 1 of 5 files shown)


danielSanchezQ:
Unrelated errors

nomos-da/network/core/src/replication/handler.rs
```diff
@@ -156,28 +194,39 @@ impl ConnectionHandler for ReplicationHandler {
...
-        if let Err(error) = self.poll_pending_outgoing_messages(cx) {
-            return Poll::Ready(ConnectionHandlerEvent::NotifyBehaviour(
-                HandlerEventToBehaviour::OutgoingMessageError { error },
-            ));
-        }
+        match self.poll_pending_outgoing_messages(cx) {
+            Ok(Some(event)) => {
+                // bubble up event
+                return Poll::Ready(event);
+            }
```
youngjoon-lee, line 201: So, it seems that this was the key reason why it didn't work? and the [waker](https://github.com/logos-co/nomos-node/pull/685/files#diff-dbcbf9be7adee5a4301bc6baf61185f77850f8a513e696115cea43959a68f86aR222) stuff.

danielSanchezQ: yeah, some issues with polling and waker.

bacv (approved):

PR #686, opened 2024-07-31
DA: DispersedBlobInfo trait in nomos-core and kzgrs-backend

bacv:
Modify DaEncoder and DaDispersal traits to work with DispersedBlobInfo. Before moving to the cli app, changes needs to be applied to the core and services.


nomos-core/src/da/mod.rs
```diff
@@ -1,29 +1,27 @@
...
+    fn verify(&self, blob: &Self::DaBlob) -> Result<(), Self::Error>;
 }
 
 pub trait DaDispersal {
     type EncodedData;
-    type Certificate;
+    type Error;
 
-    fn disperse(&self, encoded_data: Self::EncodedData) -> Result<Self::Certificate, impl Error>;
+    fn disperse(&self, encoded_data: Self::EncodedData) -> Result<(), Self::Error>;
```
danielSanchezQ, line 24: We could have left the certificate that way it would be up to implementation to use `()`. But his way is closer to the actual spec.

nomos-core/src/da/blob/info.rs
```diff
@@ -0,0 +1,8 @@
+use super::metadata::Metadata;
+
+pub trait DispersedBlobInfo: Metadata {
+    type BlobId;
+
+    fn blob_id(&self) -> Self::BlobId;
+    fn size(&self) -> usize;
```
danielSanchezQ, line 7: I'm wondering now if we could have just go for the blob struct directly. But we will see it when composing the final implementation.

nomos-da/kzgrs-backend/benches/verifier.rs
```diff
@@ -26,7 +26,7 @@ fn verify<const SIZE: usize>(bencher: Bencher, column_size: usize) {
 
             let encoder = DaEncoder::new(params);
             let data = rand_data(SIZE * MB / DaEncoderParams::MAX_BLS12_381_ENCODING_CHUNK_SIZE);
-            let encoded_data = encoder.encode(&data).unwrap();
+            let encoded_data = nomos_core::da::DaEncoder::encode(&encoder, &data).unwrap();
```
danielSanchezQ, line 29: Why this change?

danielSanchezQ (approved):
Necessary refactor, thanks for the effort.

nomos-core/src/da/blob/info.rs
bacv: If we use blob directly, the metadata will need to be optional, I think similar case was with the Certificate.

nomos-da/kzgrs-backend/benches/verifier.rs
bacv: I think I had a rust-analyzer bit laggy, because it didn't let recognize the `encode` method from the DaEncoder even after the import. Tried now and it works, fixed.

PR #688, opened 2024-08-06
Da: Network dispersal

danielSanchezQ:
* Validator behaviour
* Executor behaviour


danielSanchezQ:
@romanzac Stuff going on here you may be interested about :wink:

bacv (approved):
Impressive speed and quality. Will be easy to integrate in to network service. 🚀

nomos-da/network/core/src/dispersal/executor/behaviour.rs
```diff
@@ -0,0 +1,439 @@
...
+        tasks: &mut FuturesUnordered<StreamHandlerFuture>,
+        idle_streams: &mut HashMap<Membership::Id, DispersalStream>,
+        membership: &mut Membership,
+        connected_subnetworks: &mut HashMap<PeerId, ConnectionId>,
+        to_disperse: &mut HashMap<PeerId, VecDeque<(Membership::NetworkId, DaBlob)>>,
+        subnetwork_id: SubnetworkId,
+        blob: DaBlob,
+    ) {
+        let members = membership.members_of(&subnetwork_id);
+        let peers = members
```
holisticode, line 284: I think to understand we are filtering here for already connected peers. In the hypothetical scenario (which seems more and more unlikely) that we have more members in a subnet than already connected peers - we would not disperse to them?

holisticode (approved):
Great job! It's taking shape.

I was not seeing protobuf stuff first but I guess it had been added in earlier iterations :+1:

nomos-da/network/core/src/dispersal/executor/behaviour.rs
danielSanchezQ: The take here is that we are connected to at least one node in the subnetwork. If we are connected to more than one we try to send it as well yes.

PR #689, opened 2024-08-09
Da: Network sampling

danielSanchezQ:


danielSanchezQ:
Need of #688 to be merged first.

danielSanchezQ:
@romanzac More stuff on DA networking (libp2p) here.

romanzac:
> @romanzac More stuff on DA networking (libp2p) here.

I'll have more time to go through the code next week. libp2p is used for all types of messaging - consensus, da,  ?

danielSanchezQ:
> > @romanzac More stuff on DA networking (libp2p) here.
> 
> I'll have more time to go through the code next week. libp2p is used for all types of messaging - consensus, da, ?

Yes, everything uses libp2p, although DA is probably the one that goes deeper in the stack.

romanzac:
Hi, @danielSanchezQ I have a question about maximum subnet size. We were talking about 4096. I haven't seen any check on number of members in the implementations of trait MembershipHandler. It is kind of a restriction, so perhaps it will be added later ?

danielSanchezQ:
> Hi, @danielSanchezQ I have a question about maximum subnet size. We were talking about 4096. I haven't seen any check on number of members in the implementations of trait MembershipHandler. It is kind of a restriction, so perhaps it will be added later ?

4096 is a theoretical limit (because of current hardware encoding times). We probably gonna go with just 2048 subnets for now. 
This is gonna be a configuration that should be in one of the upper layers (tbd). Or even in the final implementation of the MembershipHandler. I can ping you to the the code once its done :smiley:

nomos-da/network/core/src/sampling/behaviour.rs
```diff
@@ -0,0 +1,592 @@
...
+        event: THandlerOutEvent<Self>,
+    ) {
+        let Either::Left(event) = event else {
+            unreachable!()
+        };
+        self.stream_behaviour
+            .on_connection_handler_event(peer_id, connection_id, event)
+    }
+
+    fn poll(
```
bacv, line 491: Would it be possible to reuse common parts from nomos behaviors to simplify `poll` and other network behavior methods?

bacv (approved):
Looks good!

nomos-da/network/core/src/sampling/behaviour.rs
danielSanchezQ: I thought about it. Some patterns are really similar, and probably we could make some common functions. But the small differences (types, reactions, etc) may be a burden later on to maintain IMO.

PR #690, opened 2024-08-12
DA: Dispersal in Cli App with mock backend

bacv:
Dispersal command with simplified da network backend mock for executor.

This version has simplified configuration for the cli, with a plan to improve it in subsequent pull requests with more options for different da network backends (libp2p or mock).

Full replication option was removed as an option in the dispersal cli app.

### DIFF
--- a/nomos-cli/src/api/mempool.rs
+++ b/nomos-cli/src/api/mempool.rs
@@ -2,14 +2,14 @@ use super::CLIENT;
 use reqwest::{Error, Response, Url};
 use serde::Serialize;
 
-pub async fn send_certificate<C>(node: &Url, cert: &C) -> Result<Response, Error>
+pub async fn send_blob_info<I>(node: &Url, info: &I) -> Result<Response, Error>
 where
-    C: Serialize,
+    I: Serialize,
 {
     const NODE_CERT_PATH: &str = "mempool/add/cert";
     CLIENT
         .post(node.join(NODE_CERT_PATH).unwrap())
-        .json(cert)
+        .json(info)
         .send()
         .await
 }
--- a/nomos-cli/src/cmds/disseminate/mod.rs
+++ b/nomos-cli/src/cmds/disseminate/mod.rs
@@ -18,9 +18,6 @@ pub struct Disseminate {
     /// Path to the network config file
     #[clap(short, long)]
     pub network_config: PathBuf,
-    /// The data availability protocol to use. Defaults to full replication.
-    // #[clap(flatten)]
-    // pub da_protocol: DaProtocolChoice,
     /// Timeout in seconds. Defaults to 120 seconds.
     #[clap(short, long, default_value = "120")]
     pub timeout: u64,
--- a/nomos-cli/src/cmds/disseminate/mod.rs
+++ b/nomos-cli/src/cmds/disseminate/mod.rs
@@ -1,4 +1,6 @@
-use crate::da::disseminate::{DisseminateApp, DisseminateAppServiceSettings, Settings, Status};
+use crate::da::disseminate::{
+    DisseminateApp, DisseminateAppServiceSettings, KzgrsSettings, Settings, Status,
+};
 use clap::Args;
 use nomos_da_network_service::backends::mock::executor::MockExecutorBackend as NetworkBackend;
 use nomos_da_network_service::NetworkService;
@@ -64,6 +66,7 @@ impl Disseminate {
                     send_blob: Settings {
                         payload: Arc::new(Mutex::new(payload_rx)),
                         timeout,
+                        kzgrs_settings: KzgrsSettings::default(),
                         status_updates,
                         node_addr,
                         output,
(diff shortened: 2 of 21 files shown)


nomos-cli/src/api/mempool.rs
```diff
@@ -2,14 +2,14 @@ use super::CLIENT;
...
+    I: Serialize,
 {
     const NODE_CERT_PATH: &str = "mempool/add/cert";
     CLIENT
         .post(node.join(NODE_CERT_PATH).unwrap())
-        .json(cert)
+        .json(info)
         .send()
         .await
 }
```
danielSanchezQ, line 15: We still have this. But we should remove it after we have the executor swarm. In where we should plug the mempool gossip network :thinking:

nomos-cli/src/cmds/disseminate/mod.rs
```diff
@@ -19,8 +19,8 @@ pub struct Disseminate {
     #[clap(short, long)]
     pub network_config: PathBuf,
     /// The data availability protocol to use. Defaults to full replication.
-    #[clap(flatten)]
-    pub da_protocol: DaProtocolChoice,
+    // #[clap(flatten)]
+    // pub da_protocol: DaProtocolChoice,
```
danielSanchezQ, line 23: Leftover? Probably we just want the real stuff. Full-replication should be used for mocking testing purposes only (?).

nomos-cli/src/cmds/disseminate/mod.rs
```diff
@@ -67,7 +66,7 @@ impl Disseminate {
                     send_blob: Settings {
                         payload: Arc::new(Mutex::new(payload_rx)),
                         timeout,
-                        da_protocol,
+                        kzgrs_settings: KzgrsSettings::default(),
```
danielSanchezQ, line 69: were this setting using the `toeplitz_cache`? We probably want to use that feature (as it is much faster).

nomos-cli/src/da/disseminate.rs
```diff
@@ -1,76 +1,79 @@
...
     time::Duration,
 };
 use tokio::sync::{mpsc::UnboundedReceiver, Mutex};
 
-pub async fn disseminate_and_wait<D, B, N, A, C>(
-    mut da: D,
+use super::network::adapters::mock::MockExecutorDispersalAdapter;
+use crate::api::mempool::send_blob_info;
+
+const APP_ID: &[u8; 32] = b"1234567890abcdef1234567890abcdef";
```
danielSanchezQ, line 36: lol
We could add something meaningful though, like: `32bytes_hash("nomos-cli")`.

nomos-cli/src/da/disseminate.rs
```diff
@@ -1,76 +1,79 @@
...
+    <D as nomos_core::da::DaDispersal>::Error: std::error::Error + Send + Sync + 'static,
 {
     // 1) Building blob
     status_updates.send(Status::Encoding)?;
-    let blobs = da.encode(data);
+    let encoded_data = encoder.encode(&data).map_err(Box::new)?;
+    let blob_hash = build_attestation_message(
+        &encoded_data.aggregated_column_commitment,
+        &encoded_data.row_commitments,
+    );
```
danielSanchezQ, line 58: We do not have attestation per se now. We should probably rename this and adapt to latest version of the protocol. Do you mind opening an issue for this?

nomos-cli/src/da/disseminate.rs
```diff
@@ -199,64 +208,17 @@ impl ServiceCore for DisseminateService {
...
 }
 
-impl Default for FullReplicationSettings {
+impl Default for KzgrsSettings {
     fn default() -> Self {
         Self {
-            voter: [0; 32],
-            num_attestations: 1,
+            num_columns: 4096,
+            with_cache: true,
```
danielSanchezQ, line 221: Ok, this answers my question above. Do not mind. Thanks!

danielSanchezQ (approved):
Looking good. We will need to remove stuff once we have things plugged in but can be done in further PRs when things are more clear.

nomos-cli/src/da/disseminate.rs
bacv: I'll rename it here to `build_blob_id` (if this the name sounds good).

danielSanchezQ: `build_blob_info`?`

nomos-cli/src/da/disseminate.rs
bacv: Moving this to app settings, will be passed via vars.

nomos-cli/src/api/mempool.rs
bacv: I'm not sure I get it. At the moment the mempool gossip network is between nodes as I understand. Once the item is added (in this case via http), the node will gossip it. I don't think we have a way to add items via libp2p currently.

danielSanchezQ: Yeah, you are partially right. In this case we are emulating an executor, an executor should be participating as a validator itself so we could just go ahead and participate in the network here. For now this is fine though.

nomos-cli/src/da/disseminate.rs
bacv: `blob_info` is `{blob_id + metadata}`. As the result of this function is hash, maybe `build_blob_hash`?

danielSanchezQ: `build_blob_id` works then. Sorry, I mixed the call with other stuff.

PR #691, opened 2024-08-19
Da: validator network behaviour & swarm

danielSanchezQ:
Sketch up validator composed behavior and swarm wrapped. This is the first step for integrating into the DA network backend. Some methods of swarm are not implemented and need to be filler later on when needed during the integration with the service itself.


bacv (approved):
Great work.

PR #693, opened 2024-08-20
ColumnIndex in DaBlob

bacv:
Column index as DABlob field for sampling and other usecases.


nomos-da/kzgrs-backend/src/common/mod.rs
```diff
@@ -19,6 +19,7 @@ pub struct Row(pub Vec<Chunk>);
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Column(pub Vec<Chunk>);
 pub struct ChunksMatrix(pub Vec<Row>);
+pub type ColumnIndex = u32;
```
danielSanchezQ, line 22: Actually now that I think of this. We should probably use the smallest fitting size which would be `u16`?

danielSanchezQ (approved):

PR #694, opened 2024-08-20
DA: Store and load multiple columns for the same blob in a single node

bacv:
To allow verification of multiple columns for the same encoded data in a single node Blob holds column_idx, which is later used in da storage service and indexer.

### DIFF
--- a/nomos-services/data-availability/verifier/src/storage/adapters/rocksdb.rs
+++ b/nomos-services/data-availability/verifier/src/storage/adapters/rocksdb.rs
@@ -5,7 +5,7 @@ use std::{marker::PhantomData, path::PathBuf};
 use nomos_core::da::blob::Blob;
 use nomos_da_storage::{
     fs::write_blob,
-    rocksdb::{key_bytes, DA_ATTESTED_KEY_PREFIX},
+    rocksdb::{key_bytes, DA_VERIFIED_KEY_PREFIX},
 };
 use nomos_storage::{
     backends::{rocksdb::RocksBackend, StorageSerde},
@@ -34,6 +34,7 @@ where
     A: Serialize + DeserializeOwned + Clone + Send + Sync,
     B: Blob + Serialize + Clone + Send + Sync + 'static,
     B::BlobId: AsRef<[u8]> + Send + Sync + 'static,
+    B::ColumnIndex: AsRef<[u8]> + Send + Sync + 'static,
     S: StorageSerde + Send + Sync + 'static,
 {
     type Backend = RocksBackend<S>;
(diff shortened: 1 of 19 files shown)


nomos-services/data-availability/verifier/src/storage/adapters/rocksdb.rs
```diff
@@ -107,6 +107,14 @@ where
     }
 }
 
+fn create_blob_idx(blob_id: &[u8], column_idx: &[u8]) -> [u8; 34] {
```
danielSanchezQ, line 110: Isn't the column_idx 16 bytes? 
`32 + 16 != 34`

nomos-services/storage/src/backends/mod.rs
```diff
@@ -45,6 +45,7 @@ pub trait StorageBackend: Sized {
     fn new(config: Self::Settings) -> Result<Self, Self::Error>;
     async fn store(&mut self, key: Bytes, value: Bytes) -> Result<(), Self::Error>;
     async fn load(&mut self, key: &[u8]) -> Result<Option<Bytes>, Self::Error>;
+    async fn load_prefix(&mut self, prefix: &[u8]) -> Result<Vec<Bytes>, Self::Error>;
```
danielSanchezQ, line 48: What does this do? Maybe is up to the user to actually compute this? it feels bad to add something that is not common to other storages. Just trying to understand it.

nomos-services/storage/src/lib.rs
```diff
@@ -25,6 +25,10 @@ pub enum StorageMsg<Backend: StorageBackend> {
         key: Bytes,
         reply_channel: tokio::sync::oneshot::Sender<Option<Bytes>>,
     },
+    LoadPrefix {
+        prefix: Bytes,
+        reply_channel: tokio::sync::oneshot::Sender<Vec<Bytes>>,
+    },
```
danielSanchezQ, line 31: I do not get what does this do.

nomos-services/storage/src/backends/mod.rs
bacv: In key value store (sled, rocksdb etc) there's an ability to load values by incomplete key. For example `/da/verified/blob_id` in this case would return all blobs from different subnetworks that the node has (keys in that case would be `/da/verified/blob_id1` .. `/da/verifier/blob_idN`. I think this is useful feature of such databases, and could also be implemented in relational dbs, although the name `LoadPrefix` could be improved. In case of sled and rocksdb this name reflects what it does directly.

nomos-services/data-availability/verifier/src/storage/adapters/rocksdb.rs
danielSanchezQ: True!! I had a brain fart! Thanks for the explanation!

danielSanchezQ (approved):

PR #695, opened 2024-08-21
Da: network service backend

danielSanchezQ:
Implement the libp2p backend for the DA network service using the specified swarm and behaviors for a validator.

### DIFF
--- a/nomos-da/network/core/src/protocols/sampling/behaviour.rs
+++ b/nomos-da/network/core/src/protocols/sampling/behaviour.rs
@@ -1,5 +1,7 @@
 // std
+use bincode::ErrorKind;
 use std::collections::{HashMap, HashSet, VecDeque};
+use std::error::Error;
 use std::task::{Context, Poll};
 // crates
 use either::Either;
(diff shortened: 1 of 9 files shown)


nomos-da/network/core/src/protocols/sampling/behaviour.rs
```diff
@@ -72,6 +73,74 @@ impl SamplingError {
...
+        ErrorKind::InvalidBoolEncoding(bool) => ErrorKind::InvalidBoolEncoding(*bool),
+        ErrorKind::InvalidCharEncoding => ErrorKind::InvalidCharEncoding,
+        ErrorKind::InvalidTagEncoding(tag) => ErrorKind::InvalidTagEncoding(*tag),
+        ErrorKind::DeserializeAnyNotSupported => ErrorKind::DeserializeAnyNotSupported,
+        ErrorKind::SizeLimit => ErrorKind::SizeLimit,
+        ErrorKind::SequenceMustHaveLength => ErrorKind::SequenceMustHaveLength,
+        ErrorKind::Custom(custom) => ErrorKind::Custom(custom.clone()),
+    })
+}
+
```
danielSanchezQ, line 143: For broadcasting `SamplingErrors` the braodcast message needs to be clone....

nomos-da/network/core/src/swarm/validator.rs
```diff
@@ -53,21 +62,48 @@ where
...
         }
     }
 
-    async fn handle_dispersal_event(&mut self, _event: DispersalEvent) {
-        // TODO: hook incoming dispersal events => to replication
-        unimplemented!()
+    async fn handle_dispersal_event(&mut self, event: DispersalEvent) {
+        match event {
+            // Send message for replication
+            DispersalEvent::IncomingMessage { message } => {
```
holisticode, line 82: Only one match case?

nomos-da/network/core/src/swarm/validator.rs
```diff
@@ -83,33 +119,32 @@ where
...
-                    SwarmEvent::ExpiredListenAddr{ .. } => {}
-                    SwarmEvent::ListenerClosed{ .. } => {}
-                    SwarmEvent::ListenerError{ .. } => {}
-                    SwarmEvent::Dialing{ .. } => {}
-                    SwarmEvent::NewExternalAddrCandidate{ .. } => {}
-                    SwarmEvent::ExternalAddrConfirmed{ .. } => {}
-                    SwarmEvent::ExternalAddrExpired{ .. } => {}
-                    SwarmEvent::NewExternalAddrOfPeer{ .. } => {}
+                    }
+                    SwarmEvent::ConnectionEstablished { .. } => {}
```
holisticode, line 129: What do you think of adding already (at least debug) logs for each of these events? It might be helpful for debugging.

nomos-services/data-availability/network/src/backends/libp2p/validator.rs
```diff
@@ -0,0 +1,217 @@
...
+
+type BlobId = [u8; 32];
+type ColumnIdx = u32;
+
+const BROADCAST_CHANNEL_SIZE: usize = 128;
+
+/// Message that the backend replies to
+#[derive(Debug)]
+pub enum DaNetworkMessage {
+    /// Quickstart a network sapling
```
holisticode, line 31: typo

nomos-da/network/core/src/swarm/validator.rs
danielSanchezQ: Yeah, I could add one debug at the top of this method, or instrument it

nomos-services/data-availability/network/src/backends/libp2p/validator.rs
danielSanchezQ: Typo not, brainfart :sweat_smile:

nomos-services/data-availability/network/src/backends/libp2p/validator.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// Identification key
+    key: Keypair,
+    /// Membership of DA network PoV set
+    membership: Membership,
+}
+
+impl<Membership> DaNetworkValidatorBackend<Membership> {
+    /// Send the sampling request to the underlying sampling behaviour
+    async fn handle_sample_request(&self, subnetwork_id: SubnetworkId, blob_id: BlobId) {
+        if let Err(SendError((subnetwork_id, blob_id))) =
```
holisticode, line 91: This is just handling the error case, is it not desirable to also handle the success case - maybe just with a debug log for now, so we can better track things until we are stable?

danielSanchezQ: The success means it arrived to the other side. It is an error because it should not happen and we cannot recover from it. It means that the service in the other side consuming this is down or erred. I could add a `trace!` log for success but I think it would clog the log system with useless information in this case.

holisticode: I am just imagining the situation where we are debugging sampling, and want to be sure that a message was sent (and then on the other side received, but debugging across process boundaries is often messy).

If we have any means already to make sure that we can reconstruct this information, then I agree we don't need any more logs.

danielSanchezQ: If we do not see this error is guaranteed it was sent. Then we would have to check on the other side if it arrived, and log if we have a receive error though.

holisticode (approved):
great job

PR #698, opened 2024-08-22
DA: Implement subnetwork handler v1

danielSanchezQ:
Implemented subnetwork handler as per the [initial specification](https://www.notion.so/NomosDA-Network-Specification-c6664294d630470ba20aefb21a218f8c?pvs=4#ae967b1d50fe4a278e4750bf6a812772), **skipping PoV** for now:

```python
from itertools import cycle
from typing import List, Sequence, Set, TypeAlias
import ipaddress

# A NodeDef consists of a bytes array for the PoV and an IP address
# on which it is publicly reachable
PoV: TypeAlias = bytes
type NodeDef = tuple[PoV, ipaddress.ip_address]

def calculate_subnets(
    nodes_list: Sequence[NodeDef], num_subnets: int, dispersal_factor: int
) -> List[Set[NodeDef]]:
    """
    Calculate in which subnet(s) to place each node.
    Fills up each subnet up to a count in an iterative mode
    """

    # key of dict is the subnet number
    nodes_list = sorted(nodes_list)
    # create a cycling iterator which will cycle through 
    # all nodes continuosly
    nodes = cycle(nodes_list)
    # for each subnet, cycle through the node list and create a set 
    # (collection) up to the number of 
    subnets = [
        set(next(nodes) for _ in range(dispersal_factor)) for _ in range(num_subnets)
    ]
    return subnets
```

### DIFF
--- a/nomos-da/network/subnetworks-assignations/src/versions/v1.rs
+++ b/nomos-da/network/subnetworks-assignations/src/versions/v1.rs
@@ -1,0 +1,88 @@
+use crate::MembershipHandler;
+use libp2p_identity::PeerId;
+use std::collections::HashSet;
+
+/// Fill a `N` sized set of "subnetworks" from a list of peer ids members
+pub struct FillFromNodeList {
+    pub assignations: Vec<HashSet<PeerId>>,
+    pub subnetwork_size: usize,
+    pub dispersal_factor: usize,
+}
+
+impl FillFromNodeList {
+    pub fn new(peers: &[PeerId], subnetwork_size: usize, dispersal_factor: usize) -> Self {
+        Self {
+            assignations: Self::fill(peers, subnetwork_size, dispersal_factor),
+            subnetwork_size,
+            dispersal_factor,
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


nomos-da/network/subnetworks-assignations/src/versions/v1.rs
```diff
@@ -0,0 +1,88 @@
...
+            replication_factor: dispersal_factor,
+        }
+    }
+
+    fn fill(
+        peers: &[PeerId],
+        subnetwork_size: usize,
+        replication_factor: usize,
+    ) -> Vec<HashSet<PeerId>> {
+        assert!(!peers.is_empty());
```
bacv, line 26: Nit: Because of HashSet (only one unique entry) the test bellow fails, because each subnet has only one entry when replication factor is 2.

bacv: Edited comment, this happens when there's only one peer in the list.

bacv (approved):

PR #701, opened 2024-08-23
DA: Network service executor

bacv:
Libp2p executor backend and swarm. 

The functionality for dialing all validator nodes in required nodes will be improved in upcoming PRs.


nomos-cli/src/da/network/adapters/libp2p.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+impl std::error::Error for DispersalError {}
+
+impl From<String> for DispersalError {
+    fn from(s: String) -> Self {
+        DispersalError(s)
+    }
+}
```
danielSanchezQ, line 28: Better to use `thiserror`

nomos-cli/src/da/network/backend.rs
```diff
@@ -0,0 +1,186 @@
...
+#[derive(Debug)]
+pub enum Command {
+    /// Disperse a blob to a subnetwork.
+    Disperse {
+        subnetwork_id: ColumnIdx,
+        blob: DaBlob,
+    },
+}
+
+/// Sampling events coming from da network
```
danielSanchezQ, line 45:
```suggestion
/// Dispersal events coming from da network
```

nomos-cli/src/da/network/swarm.rs
```diff
@@ -0,0 +1,133 @@
...
+#[derive(Debug)]
+pub enum DispersalEvent {
+    /// A blob successfully arrived its destination
+    DispersalSuccess {
+        blob_id: BlobId,
+        subnetwork_id: SubnetworkId,
+    },
+    /// Something went wrong delivering the blob
+    DispersalError { error: DispersalError },
+}
```
danielSanchezQ, line 33: You could probably reuse this type in the Backend directly as they are exactly the same?

nomos-da/network/core/src/protocols/dispersal/executor/behaviour.rs
```diff
@@ -75,7 +76,75 @@ impl DispersalError {
...
+        bincode::ErrorKind::InvalidCharEncoding => bincode::ErrorKind::InvalidCharEncoding,
+        bincode::ErrorKind::InvalidTagEncoding(tag) => bincode::ErrorKind::InvalidTagEncoding(*tag),
+        bincode::ErrorKind::DeserializeAnyNotSupported => {
+            bincode::ErrorKind::DeserializeAnyNotSupported
+        }
+        bincode::ErrorKind::SizeLimit => bincode::ErrorKind::SizeLimit,
+        bincode::ErrorKind::SequenceMustHaveLength => bincode::ErrorKind::SequenceMustHaveLength,
+        bincode::ErrorKind::Custom(custom) => bincode::ErrorKind::Custom(custom.clone()),
+    })
+}
```
danielSanchezQ, line 147: Isn't this the same method I did? If we can reuse it we can move the method to a pub(crate) shared file so we can reuse it.

nomos-cli/src/da/network/adapters/libp2p.rs
```diff
@@ -0,0 +1,66 @@
...
+            };
+
+            self.network_relay
+                .send(DaNetworkMsg::Process(Command::Disperse {
+                    blob,
+                    subnetwork_id: i as u32,
+                }))
+                .await
+                .map_err(|(e, _)| e.to_string())?
+        }
```
danielSanchezQ, line 63: Nitpick: this could probably be done asyncrhonously (but it will do for now.)

nomos-cli/src/da/network/swarm.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    pub fn dial(&mut self, addr: Multiaddr) -> Result<ConnectionId, DialError> {
+        let opt = DialOpts::from(addr.clone());
+        let connection_id = opt.connection_id();
+
+        tracing::debug!("attempting to dial {addr}. connection_id:{connection_id:?}",);
+        self.swarm.dial(opt)?;
+        Ok(connection_id)
+    }
```
danielSanchezQ, line 79: Nitpick: I think you can use the `self.swarm.dial` directly (iirc Multiaddress implements Into<DialOpts>). But im not sure.

nomos-libp2p/src/config.rs
```diff
@@ -142,7 +142,7 @@ impl From<GossipsubConfigDef> for gossipsub::Config {
     }
 }
 
-mod secret_key_serde {
+pub mod secret_key_serde {
```
danielSanchezQ, line 145: This are the kind of things I warn about. Everytime something we want to change we will have to bridge it first. There is no need.

nomos-libp2p/src/lib.rs
```diff
@@ -6,7 +6,7 @@ use std::task::{Context, Poll};
 
 use std::time::Duration;
 
-pub use config::SwarmConfig;
+pub use config::{secret_key_serde, SwarmConfig};
```
danielSanchezQ, line 9: Yep, sum up example

nomos-da/network/subnetworks-assignations/src/versions/v1.rs
```diff
@@ -40,7 +42,7 @@ impl FillFromNodeList {
 }
 
 impl MembershipHandler for FillFromNodeList {
-    type NetworkId = u16;
+    type NetworkId = u32;
```
danielSanchezQ, line 45: I thought we said `u16` maybe we need to change the SubnetworkId (can be done later on).
There is some job to stabilize what types do we use. I see you have already pierce through the `BlobId` nightmare we had :wink:

nomos-da/network/subnetworks-assignations/Cargo.toml
```diff
@@ -4,8 +4,9 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-libp2p-identity = { version = "0.2", features = ["peerid"] }
+libp2p-identity = { version = "0.2", features = ["peerid", "serde"] }
+serde = "1"
 
 
 [dev-dependencies]
-libp2p-identity = { version = "0.2", features = ["rand"] }
```
danielSanchezQ, line 11: git is drunk :champagne:

nomos-da/network/subnetworks-assignations/src/versions/v1.rs
bacv: So we have `ColumnIdx` which goes into the blob and is used by the storage. Then we have `SubnetworkId`, which is used by the protobufs. We can either have one type or two. Atm SubnetworkId is u32 because that maps conveniently into the protobuf type (where it will be uaed the most). Either way we'll need to make a conversion somewhere.

nomos-cli/src/da/network/adapters/libp2p.rs
bacv: To have async here we'll need a bit more work, because dialing might take time (depending on the host config, there's a cap on how many pending conns you can have), and we might start pushing data before we have a conn for the subnetwork. Sync will be easier to debug in the beginning.

bacv: Ah, thought the comment is for the swarm init, will make this async.

nomos-cli/src/da/network/swarm.rs
bacv: Used your suggestion. Will reintroduce connection id when needed.

PR #704, opened 2024-08-26
DA Network in node

bacv:
Use da network service with verifier backend in Nomos Node.

### DIFF
--- a/nomos-services/data-availability/network/src/backends/libp2p/validator.rs
+++ b/nomos-services/data-availability/network/src/backends/libp2p/validator.rs
@@ -1,7 +1,7 @@
 use crate::backends::NetworkBackend;
 use futures::{Stream, StreamExt};
 use kzgrs_backend::common::blob::DaBlob;
-use libp2p::identity::secp256k1;
+use libp2p::identity::ed25519;
 use libp2p::{Multiaddr, PeerId};
 use log::error;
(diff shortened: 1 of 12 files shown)


nomos-services/data-availability/network/src/backends/libp2p/validator.rs
```diff
@@ -75,13 +77,14 @@ pub struct DaNetworkValidatorBackend<Membership> {
...
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct DaNetworkValidatorBackendSettings<Membership> {
-    /// Identification key
-    key: Keypair,
+    // Identification Secp256k1 private key in Hex format (`0x123...abc`). Default random.
+    #[serde(with = "secret_key_serde", default = "secp256k1::SecretKey::generate")]
+    pub node_key: secp256k1::SecretKey,
```
danielSanchezQ, line 84: I thought we would use `ed2xxxx`. There is no specification for this though. But it is important as it is attached to the `PeerId` itself.

danielSanchezQ (approved):

nomos-services/data-availability/network/src/backends/libp2p/validator.rs
bacv: Updated

PR #705, opened 2024-08-27
DA: Sampling service

holisticode:
First iteration of sampling service.

Correct message handling and service structure. Missing backend implementation.


nomos-services/data-availability/sampling/src/network/adapters/libp2p.rs
```diff
@@ -0,0 +1,127 @@
...
+            rng: ChaCha20Rng::from_entropy(),
+        }
+    }
+
+    async fn start_sampling(&mut self, blob_id: BlobId) -> Result<(), DynError> {
+        let mut ctx: SamplingContext = SamplingContext {
+            blob_id: (blob_id),
+            subnets: vec![],
+        };
+        for _i in 1..self.settings.num_samples {
```
danielSanchezQ, line 86: why from 1?, this would result in off by 1 error I think.

nomos-services/data-availability/sampling/src/network/adapters/libp2p.rs
```diff
@@ -0,0 +1,127 @@
...
+            self.network_relay
+                .send(DaNetworkMsg::Process(DaNetworkMessage::RequestSample {
+                    blob_id,
+                    subnetwork_id,
+                }))
+                .await
+                .expect("RequestSample message should have been sent")
+        }
+        // TODO: needs to be updated after handling the event in the backend
+        self.pending_sampling.insert(blob_id, ctx);
```
danielSanchezQ, line 98: What does the pending_sampling do? Once the message is sent is sent.

nomos-services/data-availability/sampling/src/network/adapters/libp2p.rs
```diff
@@ -0,0 +1,127 @@
...
+        + 'static,
+{
+    settings: DaNetworkSamplingSettings,
+    network_relay: OutboundRelay<
+        <NetworkService<DaNetworkValidatorBackend<Membership>> as ServiceData>::Message,
+    >,
+    pending_sampling: HashMap<BlobId, SamplingContext>,
+    // TODO: is there a better place for this? Do we need to have this even globally?
+    // Do we already have some source of randomness already?
+    rng: ChaCha20Rng,
```
danielSanchezQ, line 53: This is actually really important. And we should link it to the corresponding part in the spec, and if it is missing we should add it. (We can open an issue to track it).

danielSanchezQ (approved):

danielSanchezQ:
> First iteration of sampling service.
> 
> Has still lots of issues due to incompatible rust types and lifetimes. Does not compile

You should change the description before merging as well.

nomos-services/data-availability/sampling/src/network/adapters/libp2p.rs
holisticode: Yes, but we need to accumulate, until we reach the desired number of successful samples for a blob? Only then we can consider sampling successful.

nomos-services/data-availability/sampling/src/network/adapters/libp2p.rs
holisticode: You are referring to the specification for randomness as such right? Not actually the content of the comment (where in the code it has to live).

In that case, yes, we should open an issue.

nomos-services/data-availability/sampling/src/network/adapters/libp2p.rs
danielSanchezQ: Actually, I just realize, that this is wrong. Iterating like this may produce the same number twice (or more). Which is undesirable. 
I think instead of itereting, you need to generate the subnetwork ids using somehting like [`random::sample`](https://docs.rs/rand/latest/rand/distributions/trait.Distribution.html#method.sample_iter). So `N` different subnetwork are selected, and with that you can build the message.

danielSanchezQ (changes requested):
The random sampling need to be fixed as per the comments

nomos-services/data-availability/sampling/src/network/adapters/libp2p.rs
danielSanchezQ: Where it lives it doesn't matter much. Actually I think it could be spawn per request unless we need to use an specific seed from a source (I think it should not be the case). But questions like:
* What random algorithm to use?
* Do we need a common seed?
Need to be answered in the specification and matched in code (unless we consider is an implementation detail, which I'm not sure).

nomos-services/data-availability/sampling/src/network/adapters/libp2p.rs
danielSanchezQ: > Yes, but we need to accumulate, until we reach the desired number of successful samples for a blob? Only then we can consider sampling successful.

But that is not done here. That accumulation is done in the backend, not in the network adapter. You never receive any notification for the incoming message here.

nomos-services/data-availability/sampling/src/network/adapters/libp2p.rs
holisticode: You're right. Moved to use the suggestion you gave.

nomos-services/data-availability/sampling/src/network/adapters/libp2p.rs
holisticode: Decided to move everything into backend, and have the backend generate the list and pass it as a parameter to the network backend via the service.

nomos-services/data-availability/sampling/src/network/adapters/libp2p.rs
```diff
@@ -73,19 +60,16 @@ where
...
-            blob_id: (blob_id),
-            subnets: vec![],
-        };
-        for _i in 1..self.settings.num_samples {
-            let subnetwork_id = self.rng.gen_range(0..self.settings.subnet_size);
-            ctx.subnets.push(subnetwork_id);
+    async fn start_sampling(
+        &mut self,
+        blob_id: BlobId,
+        subnets: &Vec<SubnetworkId>,
```
danielSanchezQ, line 69: Nipick: is more rusty that if you are not gonna consume the `Vec` you can use a slice (Vec dereferences into slice). 
```rust
subnets: &[SubnetworkId]
```

nomos-services/data-availability/sampling/src/network/adapters/libp2p.rs
```diff
@@ -73,19 +60,16 @@ where
...
+    ) -> Result<(), DynError> {
+        for id in subnets.iter() {
```
danielSanchezQ, line 71: Nitpick: You probably do not need the `.iter`, just `for id in &subnets` should be enough (it later desugars into `subnets.iter()` though).

nomos-services/data-availability/sampling/src/network/adapters/libp2p.rs
```diff
@@ -73,19 +60,16 @@ where
...
+            let subnetwork_id = id.clone();
```
danielSanchezQ, line 72: Nipick: SubnetworkId is `Copy` so you do not need to `Clone` implecetely.

danielSanchezQ (approved):
Good job. Added a few nipicks relate to having better rusty code. But it is fine overall. Thanks!

PR #708, opened 2024-08-28
DA: Consensus sampling

danielSanchezQ:
Add sampling to consensus.
Added a verification of block so it is not included if as a non-proposer you do not see the included blobs in your sample view.
As a proposer include only the blobs that you see available in your sample view.

### DIFF
--- a/nomos-da/network/subnetworks-assignations/src/versions/v2.rs
+++ b/nomos-da/network/subnetworks-assignations/src/versions/v2.rs
@@ -1,7 +1,9 @@
 use crate::MembershipHandler;
 use libp2p_identity::PeerId;
+use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
 
+#[derive(Default, Debug, Clone, Serialize, Deserialize)]
 pub struct FillWithOriginalReplication {
     pub assignations: Vec<HashSet<PeerId>>,
     pub subnetwork_size: usize,
@@ -63,7 +65,7 @@ impl FillWithOriginalReplication {
 }
 
 impl MembershipHandler for FillWithOriginalReplication {
-    type NetworkId = u16;
+    type NetworkId = u32;
     type Id = PeerId;
 
     fn membership(&self, id: &Self::Id) -> HashSet<Self::NetworkId> {
--- a/nomos-services/cryptarchia-consensus/src/lib.rs
+++ b/nomos-services/cryptarchia-consensus/src/lib.rs
@@ -16,6 +16,8 @@ use nomos_core::{
     block::{builder::BlockBuilder, Block},
     header::cryptarchia::Builder,
 };
+use nomos_da_sampling::backend::DaSamplingServiceBackend;
+use nomos_da_sampling::DaSamplingService;
 use nomos_mempool::{
     backend::MemPool, network::NetworkAdapter as MempoolAdapter, DaMempoolService, MempoolMsg,
     TxMempoolService,
@@ -29,6 +31,7 @@ use overwatch_rs::services::{
     state::{NoOperator, NoState},
     ServiceCore, ServiceData, ServiceId,
 };
+use rand::{RngCore, SeedableRng};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use serde_with::serde_as;
 use std::hash::Hash;
@@ -134,8 +137,19 @@ impl<Ts, Bs> CryptarchiaSettings<Ts, Bs> {
     }
 }
 
-pub struct CryptarchiaConsensus<A, ClPool, ClPoolAdapter, DaPool, DaPoolAdapter, TxS, BS, Storage>
-where
+pub struct CryptarchiaConsensus<
+    A,
+    ClPool,
+    ClPoolAdapter,
+    DaPool,
+    DaPoolAdapter,
+    TxS,
+    BS,
+    Storage,
+    SamplingBackend,
+    SamplingNetworkAdapter,
+    SamplingRng,
+> where
     A: NetworkAdapter,
     ClPoolAdapter: MempoolAdapter<Payload = ClPool::Item, Key = ClPool::Key>,
     ClPool: MemPool<BlockId = HeaderId>,
@@ -150,6 +164,12 @@ where
     TxS: TxSelect<Tx = ClPool::Item>,
     BS: BlobSelect<BlobId = DaPool::Item>,
     Storage: StorageBackend + Send + Sync + 'static,
+    SamplingRng: SeedableRng + RngCore,
+    SamplingBackend: DaSamplingServiceBackend<SamplingRng> + Send,
+    SamplingBackend::Settings: Clone,
+    SamplingBackend::Blob: Debug + 'static,
+    SamplingBackend::BlobId: Debug + 'static,
+    SamplingNetworkAdapter: nomos_da_sampling::network::NetworkAdapter,
 {
     service_state: ServiceStateHandle<Self>,
     // underlying networking backend. We need this so we can relay and check the types properly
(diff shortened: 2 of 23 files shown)


nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -553,29 +695,51 @@ where
...
+    fn validate_block(
+        block: &Block<ClPool::Item, DaPool::Item>,
+        sampled_blobs_ids: &BTreeSet<DaPool::Key>,
+    ) -> bool {
+        let validated_blobs = block
+            .blobs()
+            .all(|blob| sampled_blobs_ids.contains(&blob.blob_id()));
+        validated_blobs
+    }
 }
```
danielSanchezQ, line 743: This is very temporary and will have to include many other things later one. But I think for now will do.

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -553,29 +695,51 @@ where
...
                 let Ok(block) = BlockBuilder::new(tx_selector, blob_selector)
                     .with_cryptarchia_builder(Builder::new(parent, proof))
                     .with_transactions(cl_txs)
-                    .with_blobs_certificates(da_certs)
+                    .with_blobs_info(
+                        da_blobs_info.filter(move |info| blobs_ids.contains(&info.blob_id())),
+                    )
                     .build()
                 else {
                     panic!("Proposal block should always succeed to be built")
```
holisticode, line 714: Is this temporary too?

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -553,29 +695,51 @@ where
...
         da_mempool_relay: MempoolRelay<DaPoolAdapter::Payload, DaPool::Item, DaPool::Key>,
+        sampling_relay: SamplingRelay<SamplingBackend::BlobId>,
     ) -> Option<Block<ClPool::Item, DaPool::Item>> {
         let mut output = None;
         let cl_txs = get_mempool_contents(cl_mempool_relay);
         let da_certs = get_mempool_contents(da_mempool_relay);
-
-        match futures::join!(cl_txs, da_certs) {
-            (Ok(cl_txs), Ok(da_certs)) => {
+        let blobs_ids = get_sampled_blobs(sampling_relay);
```
holisticode, line 703: So this is the proposer who is querying for sampled blobs, not the mempool, right? Does it actually make any difference here?

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -633,3 +797,28 @@ async fn mark_in_block<Payload, Item, Key>(
...
+
+async fn mark_blob_in_block<BlobId: Debug>(
+    sampling_relay: SamplingRelay<BlobId>,
+    blobs_id: Vec<BlobId>,
+) {
+    if let Err((_e, DaSamplingServiceMsg::MarkInBlock { blobs_id })) = sampling_relay
+        .send(DaSamplingServiceMsg::MarkInBlock { blobs_id })
+        .await
+    {
+        error!("Error marking in block for blobs ids: {blobs_id:?}");
```
holisticode, line 809: I am actually seeing a couple of places where errors are just logged. 
What would actually the consequence be of an error here, for example? Is logging enough?

holisticode (approved):
Basically looks good to me. Left a couple of questions, mainly.

nomos-services/cryptarchia-consensus/src/lib.rs
danielSanchezQ: The mempool will trigger the sampling, but it has no usage for the already sampled blobs.
Btw, mempools will change once we actually have TXs, as blobs are part of the TXs themselves. Probably DA mempool gonna be remove entirely.

nomos-services/cryptarchia-consensus/src/lib.rs
danielSanchezQ: This should never happen as it is internally to the node (msg from service to service). We could actually panic here. Or try to recover. There is no clear policy on this cases. Ideas and suggestions are welcome (open an issue for them).

nomos-da/network/subnetworks-assignations/src/versions/v2.rs
```diff
@@ -63,7 +65,7 @@ impl FillWithOriginalReplication {
 }
 
 impl MembershipHandler for FillWithOriginalReplication {
-    type NetworkId = u16;
+    type NetworkId = u32;
```
bacv, line 68: There's `SubnetworkId` type in `nomos-da-network-core`, maybe we could use it in all network related places?

nomos-services/data-availability/sampling/src/backend/kzgrs.rs
```diff
@@ -1,2 +1,41 @@
...
+
+    fn new(_settings: Self::Settings) -> Self {
+        todo!()
+    }
+
+    async fn get_validated_blobs(&self) -> BTreeSet<Self::BlobId> {
+        todo!()
+    }
+
+    async fn mark_in_block(&mut self, _blobs_id: &[Self::BlobId]) {
```
bacv, line 26: In the context sampling, it seems that this could better be a `mark_complete` or similar.
 `mark_in_block` is actually happening in the consensus, this will just receive a notification about it and make actions according to it.

nomos-da/network/subnetworks-assignations/src/versions/v2.rs
danielSanchezQ: Yes. I changed it as it was a miss aligned type. It could probably be changed in a separated PR though.

nomos-services/data-availability/sampling/src/backend/kzgrs.rs
danielSanchezQ: Sure, this makes sense.

nomos-services/cryptarchia-consensus/src/lib.rs
holisticode: I have learnt in other contexts that when a node encounters an issue it can't actually recover from, it should shutdown, but not panic.

The question here is, what does such an error as this one mean here. Is it fatal, an unrecoverable situation? Actually not really, but something is very fishy still.

Still, a node shutdown should be preferred to a panic IMO.

danielSanchezQ: At the moment there is no really a difference between shutdown and panic. 
* A panic should be used when a path is a **bug**, meaning that the programmed logic is wrong, it should be fixed. 
* An error that does not matter (the node may continue) should be logged and execution should continue.
* When something neither the node nor we expect  then as you said it should be shutdown/restart.

In this specific case it was logged so we can see something went wrong then we should try to recover the relay later on (as it was broken), and if its not possible then the node could shut down. But still do not have any recoverable measures in place (anywhere). Not something that we are prioritising now. 
But if you are interested it would be really beneficial to prepare a design for such policies!

danielSanchezQ:
@bacv, is it possible that this now fails because of the tests we just added? As I'm changing consensus :thinking: 
How are we checking the dissemination is ok? Consensus wise should be fine as we do not have any blob in the blocks.

bacv:
> @bacv, is it possible that this now fails because of the tests we just added? As I'm changing consensus 🤔 How are we checking the dissemination is ok? Consensus wise should be fine as we do not have any blob in the blocks.

Da integration tests are using consensus, this pr changed the generic params that needs to be passed to it, and the tests still uses old ones. I'll make a pr that fixes them (to run locally use `cargo test -p nomos-da-tests -F libp2p`)

danielSanchezQ:
> > @bacv, is it possible that this now fails because of the tests we just added? As I'm changing consensus 🤔 How are we checking the dissemination is ok? Consensus wise should be fine as we do not have any blob in the blocks.
> 
> Da integration tests are using consensus, this pr changed the generic params that needs to be passed to it, and the tests still uses old ones. I'll make a pr that fixes them (to run locally use `cargo test -p nomos-da-tests -F libp2p`)

Ok, I see them now. I couldn't reproduce them before. I'll fix it in this PR don't worry!

bacv:
> @bacv, is it possible that this now fails because of the tests we just added? As I'm changing consensus 🤔 How are we checking the dissemination is ok? Consensus wise should be fine as we do not have any blob in the blocks.

I also see another issue in the nodes test, the `DA-Sampling` service is required by consensus, but it's not started in the node.

danielSanchezQ:
> > @bacv, is it possible that this now fails because of the tests we just added? As I'm changing consensus 🤔 How are we checking the dissemination is ok? Consensus wise should be fine as we do not have any blob in the blocks.
> 
> I also see another issue in the nodes test, the `DA-Sampling` service is required by consensus, but it's not started in the node.

Ah damn, its true. You added recently in a different PR right? i'll wait for that then.

bacv:
> > > @bacv, is it possible that this now fails because of the tests we just added? As I'm changing consensus 🤔 How are we checking the dissemination is ok? Consensus wise should be fine as we do not have any blob in the blocks.
> > 
> > 
> > I also see another issue in the nodes test, the `DA-Sampling` service is required by consensus, but it's not started in the node.
> 
> Ah damn, its true. You added recently in a different PR right? i'll wait for that then.

In that one the verifier and indexer were added, so sampling should be added in a similar fashion.

danielSanchezQ:
> > > > @bacv, is it possible that this now fails because of the tests we just added? As I'm changing consensus 🤔 How are we checking the dissemination is ok? Consensus wise should be fine as we do not have any blob in the blocks.
> > > 
> > > 
> > > I also see another issue in the nodes test, the `DA-Sampling` service is required by consensus, but it's not started in the node.
> > 
> > 
> > Ah damn, its true. You added recently in a different PR right? i'll wait for that then.
> 
> In that one the verifier and indexer were added, so sampling should be added in a similar fashion.

Cool, so I'll just add it here as well then. Thanks!

bacv (approved):

danielSanchezQ:
Thanks for the effort @bacv ! :guitar:

PR #709, opened 2024-08-28
Sampling backend implementation

holisticode:
This PR adds the sampling backend implementation to the sampling backend service.

TODO:
The monitoring thread checking for old blobs needs some work:
1. Proper synchronization between threads
2. Correct implementation of lifetimes
3. Abstraction of the RNG implementation

### DIFF
--- a/nomos-services/data-availability/sampling/Cargo.toml
+++ b/nomos-services/data-availability/sampling/Cargo.toml
@@ -23,6 +23,7 @@ tracing = "0.1"
 thiserror = "1.0.63"
 rand = "0.8.5"
 rand_chacha = "0.3.1"
+chrono = "0.4.38"
 
 [features]
 default = ["libp2p"]
--- a/nomos-services/data-availability/sampling/src/backend/kzgrs.rs
+++ b/nomos-services/data-availability/sampling/src/backend/kzgrs.rs
@@ -1,2 +1,131 @@
+use std::borrow::BorrowMut;
+// std
+use std::collections::{BTreeSet, HashMap};
+use std::fmt::Debug;
+use std::thread;
+use std::time::Duration;
 
+// crates
+use chrono::{naive::NaiveDateTime, Utc};
+use rand::distributions::Standard;
+use rand::prelude::*;
+use rand_chacha::ChaCha20Rng;
 
+use kzgrs_backend::common::blob::DaBlob;
+
+//
+// internal
+use super::DaSamplingServiceBackend;
+use nomos_core::da::BlobId;
+use nomos_da_network_core::SubnetworkId;
+
+pub struct SamplingContext {
+    blob_id: BlobId,
+    subnets: Vec<SubnetworkId>,
+    started: NaiveDateTime,
+}
+
+#[derive(Debug, Clone)]
+pub struct KzgrsDaSamplerSettings {
+    pub num_samples: u16,
+    pub old_blobs_check_duration: Duration,
+    pub blobs_validity_duration: Duration,
+}
+
+pub struct KzgrsDaSampler {
+    settings: KzgrsDaSamplerSettings,
+    validated_blobs: BTreeSet<BlobId>,
+    // TODO: This needs to be properly synchronized, if this is going to be accessed
+    // by independent threads (monitoring thread)
+    pending_sampling_blobs: HashMap<BlobId, SamplingContext>,
+    // TODO: is there a better place for this? Do we need to have this even globally?
+    // Do we already have some source of randomness already?
+    rng: ChaCha20Rng,
+}
+
+impl KzgrsDaSampler {
+    // TODO: this might not be the right signature, as the lifetime of self needs to be evaluated
+    async fn start_pending_blob_monitor(&'static mut self) {
+        //let mut sself = self;
+        let monitor = thread::spawn(move || {
+            loop {
+                thread::sleep(self.settings.old_blobs_check_duration);
+                // everything older than cut_timestamp should be removed;
+                let cut_timestamp = Utc::now().naive_utc() - self.settings.blobs_validity_duration;
+                // retain all elements which come after the cut_timestamp
+                self.pending_sampling_blobs
+                    .retain(|_, ctx| ctx.started.gt(&cut_timestamp));
+            }
+        });
+        monitor.join().unwrap();
+    }
+}
+
+#[async_trait::async_trait]
+impl<'a> DaSamplingServiceBackend for KzgrsDaSampler {
+    type Settings = KzgrsDaSamplerSettings;
+    type BlobId = BlobId;
+    type Blob = DaBlob;
+
+    fn new(settings: Self::Settings) -> Self {
+        let bt: BTreeSet<BlobId> = BTreeSet::new();
+        Self {
+            settings: settings,
+            validated_blobs: bt,
+            pending_sampling_blobs: HashMap::new(),
+            rng: ChaCha20Rng::from_entropy(),
+        }
+        // TODO: how to start the actual monitoring thread with the correct ownership/lifetime?
+    }
+
+    async fn get_validated_blobs(&self) -> BTreeSet<Self::BlobId> {
+        self.validated_blobs.clone()
+    }
+
+    async fn mark_in_block(&mut self, blobs_ids: &[Self::BlobId]) {
+        for id in blobs_ids {
+            if self.pending_sampling_blobs.contains_key(id) {
+                self.pending_sampling_blobs.remove(id);
+            }
+
+            if self.validated_blobs.contains(id) {
+                self.validated_blobs.remove(id);
+            }
+        }
+    }
+
+    async fn handle_sampling_success(&mut self, blob_id: Self::BlobId, blob: Self::Blob) {
+        // this should not even happen
+        if !self.pending_sampling_blobs.contains_key(&blob_id) {}
+
+        let ctx = self.pending_sampling_blobs.get_mut(&blob_id).unwrap();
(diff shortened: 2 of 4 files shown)


bacv:
Are you planning to add network functionality to the sampler backend in this PR?

nomos-services/data-availability/sampling/src/backend/kzgrs.rs
```diff
@@ -1,2 +1,113 @@
...
+pub struct SamplingContext {
+    subnets: Vec<SubnetworkId>,
+    started: Instant,
+}
+
+#[derive(Debug, Clone)]
+pub struct KzgrsDaSamplerSettings {
+    pub num_samples: u16,
+    pub old_blobs_check_duration: Duration,
+    pub blobs_validity_duration: Duration,
```
bacv, line 27: It seems this is not being used anywhere.

nomos-services/data-availability/sampling/src/lib.rs
```diff
@@ -182,10 +193,15 @@ where
                 tokio::select! {
                     Some(service_message) = service_state.inbound_relay.recv() => {
                         Self::handle_service_message(service_message, &mut network_adapter, &mut sampler).await;
+                        // cleanup not on time samples
+                        sampler.prune();
                     }
                     Some(sampling_message) = sampling_message_stream.next() => {
                         Self::handle_sampling_message(sampling_message, &mut sampler).await;
+                        // cleanup not on time samples
+                        sampler.prune();
```
bacv, line 202: As I understand, pruning now is done only after some interaction with the service. Could there be a situation in which the sampling was triggered way before the `old_blobs_check_duration`, for some reason no other interactions happen with the service, and block producer requests for outdated validated blob (because pruning is done after the message is handled). Maybe we should have a ticker for pruning?

danielSanchezQ: A ticker is a good idea probably. Let's do this, lets delegate ticker creation to the backend (we need to add the method to the trait), because is the one that should know about timings, then in the service on tick we prune. Nice catch @bacv !

holisticode: Yes that was the initial idea with the thread in the backend, which we then moved out to the service. A combination of the two sounds like the best approach. If I am understanding correctly, a ticker in the backend, but executing pruning from the service.

danielSanchezQ: The backend is just responsible of building the ticker (as it holds the proper configuration), and pruning. The service will call the prune when the ticker ticks, as it owns the main loop.

nomos-services/data-availability/sampling/Cargo.toml
```diff
@@ -23,6 +23,7 @@ tracing = "0.1"
 thiserror = "1.0.63"
 rand = "0.8.5"
 rand_chacha = "0.3.1"
+chrono = "0.4.38"
```
bacv, line 26: Nit: this is not needed anymore.

bacv (approved):
Looks good! The tests helps to follow the flow.

nomos-services/data-availability/sampling/src/backend/kzgrs.rs
```diff
@@ -1,2 +1,397 @@
...
+            self.validated_blobs.remove(id);
+        }
+    }
+
+    async fn handle_sampling_success(&mut self, blob_id: Self::BlobId, blob: Self::Blob) {
+        if let Some(ctx) = self.pending_sampling_blobs.get_mut(&blob_id) {
+            ctx.subnets.insert(blob.column_idx as SubnetworkId);
+
+            println!("{}", self.settings.num_samples);
+            println!("{}", ctx.subnets.len());
```
danielSanchezQ, line 89: Leftover?
You can [instrument](https://docs.rs/tracing/latest/tracing/attr.instrument.html) the function if you want to debug it.

nomos-services/data-availability/sampling/src/backend/kzgrs.rs
```diff
@@ -1,2 +1,397 @@
...
+impl<R: Rng> KzgrsDaSampler<R> {
+    fn prune_by_time(&mut self) {
+        self.pending_sampling_blobs.retain(|_blob_id, context| {
+            context.started.elapsed() < self.settings.blobs_validity_duration
+        });
+    }
+}
+
+// TODO there is no logging at all here. Should we not do some logging?
+// Or can we assume that it is done on an upeer level by the clients?
```
danielSanchezQ, line 52: We can add whatever logging we need here. Then it is gonna be wrapped in the service context.

nomos-services/data-availability/sampling/src/backend/kzgrs.rs
```diff
@@ -1,2 +1,397 @@
...
+        let bt: BTreeSet<BlobId> = BTreeSet::new();
+        Self {
+            settings,
+            validated_blobs: bt,
+            pending_sampling_blobs: HashMap::new(),
+            rng,
+        }
+    }
+
+    async fn next_prune_interval(&self) -> Interval {
```
danielSanchezQ, line 69: Nitpick: `prune_interval`

nomos-services/data-availability/sampling/src/backend/kzgrs.rs
```diff
@@ -1,2 +1,397 @@
...
+                    self.validated_blobs.insert(blob_id);
+                    // cleanup from pending samplings
+                    self.pending_sampling_blobs.remove(&blob_id);
+                }
+                len if len > self.settings.num_samples as usize => {
+                    unreachable!("{}", "more subnets than expected after sampling success!");
+                }
+                // do nothing if smaller
+                _ => {}
+            }
```
danielSanchezQ, line 102:
```suggestion
        // sampling of this blob_id terminated successfully
        if ctx.subnets.len() == self.settings.num_samples as usize {
            self.validated_blobs.insert(blob_id);
            // cleanup from pending samplings
            self.pending_sampling_blobs.remove(&blob_id);
        }
```

nomos-services/data-availability/sampling/src/backend/kzgrs.rs
```diff
@@ -1,2 +1,397 @@
...
+        // If it fails a single time we consider it failed.
+        // We may want to abstract the sampling policies somewhere else at some point if we
+        // need to get fancier than this
+        self.pending_sampling_blobs.remove(&blob_id);
+        self.validated_blobs.remove(&blob_id);
+    }
+
+    // TODO: This also would be an error which should never happen, but what if a client starts
+    // init_sampling of a blob which is already pending? Or worse, which already is validated?
+    // Should we not therefore return an error here?
```
danielSanchezQ, line 118: Imo this object should be **idempotent**. Why?, Because we cannot ensure that a single service will try to start sampling or check how things are going (even if for now it is the case). So, maybe here we could add a check and just start sampling if it wasn't pending already. wdyt?

nomos-services/data-availability/sampling/src/lib.rs
```diff
@@ -115,15 +118,21 @@ where
                 sampler.handle_sampling_success(blob_id, *blob).await;
             }
             SamplingEvent::SamplingError { error } => {
+                // TODO: in most of these error cases we can't get the blob_id from the error
+                // Shouldn't the error contain that?
+                // We can of course stop tracking that blob_id in the backend via timeout,
+                // which we want to have anyways, but could it be nicer to remove it here too,
+                // by calling the handler_sampling_error method?
```
danielSanchezQ, line 125: This is a specific error. We can try to get the blob_id from the error. It is has we sort circuit the sampling (calling the error handling method with the blobid). Otherwise we wait till time expires. It can be done here or in a different PR.
Steps would be:
* Create a `blob_id` method for the sampling error which returns an `Option<BlobId>`
* Call the method here and check if it is `Some(blob_id)` to call the error handling method.
* Log it otherwise

danielSanchezQ (commented):
Looking good overall.

nomos-services/data-availability/sampling/src/backend/kzgrs.rs
holisticode: I am not much of a fan of hiding away situations which _should_ not happen but if they do, they are critical if not fatal...why would we do that? Yes, the code is nicer. Is there any place in the rest of the codebase which makes this error impossible?

nomos-services/data-availability/sampling/src/lib.rs
holisticode: Added in this PR

nomos-services/data-availability/sampling/src/backend/kzgrs.rs
holisticode: Currently, if already sampling, this would result in the network adapter resending SampleRequest messages. Out of band you suggested an `enum` return signaling the state, so that's what I did here.

nomos-services/data-availability/sampling/src/backend/kzgrs.rs
danielSanchezQ: As we discussed privately. It is not possible it is more than the expected threshold. We increment 1 by 1 (ensured by `&muy self` when we have enough the queue is empty and the blob is considered validated. Something else may happen. But not that.

nomos-services/data-availability/sampling/src/backend/kzgrs.rs
holisticode: Added just 2 tracing::info

danielSanchezQ (approved):
Made a tiny refactor, I'm merging this after CI is :green_circle: 
Good job! Thanks!

PR #711, opened 2024-08-29
DA: Nomos Cli dispersal tests

bacv:
Integration tests for nomos cli dispersing Kzgrs encoded data. For faster test times, the number of columns is set to `32`, cache is disabled.

To run these tests locally:
```bash
cargo test -p tests  -- --nocapture
```

### DIFF
--- a/nodes/nomos-node/src/api.rs
+++ b/nodes/nomos-node/src/api.rs
@@ -1,7 +1,8 @@
+// std
 use std::error::Error;
 use std::ops::Range;
 use std::{fmt::Debug, hash::Hash};
-
+// crates
 use axum::{
     extract::{Query, State},
     http::HeaderValue,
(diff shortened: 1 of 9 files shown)


danielSanchezQ (commented):
To be clear. This does disseminate a blob to a single node right? I cant see if more nodes are spawn.

bacv:
> To be clear. This does disseminate a blob to a single node right? I cant see if more nodes are spawn.

In the test two nodes are used: https://github.com/logos-co/nomos-node/pull/711/files#diff-389e28b061a855a0fb1d6b1265375992e4e02d659110eeb724f64e2af734ab0dR46

bacv:
> To be clear. This does disseminate a blob to a single node right? I cant see if more nodes are spawn.

Oh, you are right, only one, that method creates configs only, for some reason I assumed it spawns, fixing now.

holisticode:
Are we going to integrate sampling into this disseminate test or should we create an independent test? Feels like adding here might more most useful, unless test run times be too long or other impedance.

bacv:
> Are we going to integrate sampling into this disseminate test or should we create an independent test? Feels like adding here might more most useful, unless test run times be too long or other impedance.

If tests will be using nomos cli, then yes. For the parts that doesn't require cli, but could be tested using http api or other ways, a new directory could be created in this module and NomosNode could be used directly.

danielSanchezQ (approved):

PR #713, opened 2024-08-30
DA: Verifier and Indexer in Node

bacv:
Verifier and Indexer services were created and tested in a simplified node, this PR integrates them into the Nomos Node.

### DIFF
--- a/nomos-services/data-availability/verifier/src/backend/kzgrs.rs
+++ b/nomos-services/data-availability/verifier/src/backend/kzgrs.rs
@@ -4,6 +4,7 @@ use core::fmt;
 use blst::{min_sig::PublicKey, min_sig::SecretKey};
 use kzgrs_backend::{common::blob::DaBlob, verifier::DaVerifier as NomosKzgrsVerifier};
 use nomos_core::da::DaVerifier;
+use serde::{Deserialize, Serialize};
 // internal
 use super::VerifierBackend;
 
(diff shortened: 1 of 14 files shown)


nomos-services/data-availability/verifier/src/backend/kzgrs.rs
```diff
@@ -30,7 +31,22 @@ impl VerifierBackend for KzgrsDaVerifier {
...
+            .iter()
+            .map(|pk_hex| {
+                let pk_bytes =
+                    hex::decode(pk_hex).expect("Public key string should decode to bytes");
+                PublicKey::from_bytes(&pk_bytes)
+                    .expect("Public key should be reconstructed from bytes")
+            })
+            .collect::<Vec<PublicKey>>();
+
+        let verifier = NomosKzgrsVerifier::new(secret_key, &nodes_public_keys);
```
danielSanchezQ, line 49: I think this is not needed anymore. But i have to check on it.

danielSanchezQ (approved):
Looks good. Except for the verification with public keys which I think should be removed from the verifier is ok. It can be done in a different pr though.

nomos-services/data-availability/verifier/src/backend/kzgrs.rs
bacv: I've seen that sk is not used anymore, but the index is still required. Lets do it in a separate PR.

PR #715, opened 2024-09-02
DA network in verifier libp2p adapter

bacv:
Verifier was using adapter for gossipsub protocol, updated it to use subscribe to validator network service instead.


danielSanchezQ (approved):
:rocket:

PR #718, opened 2024-09-02
Add PoL

zeegomo:
Promote PoL and CL crates from nomos-pocs.
CL is mostly unchanged, while PoL has gone through some refactors. I didn't include the crate in the workspace as it has special build instruction and  e.g. does not support clippy


danielSanchezQ (approved):
Looking good!
CI is failing in the risc0 tests though

cl/cl/src/merkle.rs
```diff
@@ -0,0 +1,239 @@
...
+        assert!(i < N);
+        leaves[i] = leaf(element);
+    }
+
+    leaves
+}
+
+pub fn leaf(data: &[u8]) -> [u8; 32] {
+    let mut hasher = Sha256::new();
+    hasher.update(b"NOMOS_MERKLE_LEAF");
```
thomaslavaur, line 17: We should consider moving to tag codes if necessary. In Merkle tree constraints, that could save a lot of cycles.

thomaslavaur (approved):
LGTM

cl/cl/src/merkle.rs
zeegomo: Could you elaborate?

bacv (approved):

PR #721, opened 2024-09-03
Integrate PoL into consensus

zeegomo:
Unfortunately, testing now takes a lot longer since proving PoL on M2 takes ~10s, not sure about CI


bacv:
Did you try `macos-latest` runner, they are supposed to be m1. m2 should be coming if not available yet.

.github/workflows/build-test.yml
```diff
@@ -29,6 +29,8 @@ jobs:
           profile: minimal
```
danielSanchezQ, line 29: Not sure if its worth it, but we may have a separated workflow for risc0 if needed. Wdyt @bacv ?

nomos-services/cryptarchia-consensus/src/leadership.rs
```diff
@@ -1,59 +1,112 @@
...
+                let res = tokio::task::spawn_blocking(move || {
+                    Risc0LeaderProof::build(
+                        public_inputs,
+                        LeaderPrivate {
+                            input,
+                            input_cm_path,
+                        },
+                    )
+                })
+                .await;
```
danielSanchezQ, line 87: Is this a heavy operation? I don't get the spawn_blocking to immediately await.

zeegomo: yes, this is heavy and cpu intenstive (i.e. does not yield).`spawn_blocking` uses a dedicated thread for this to avoid blocking other tasks

zeegomo:
> Verifying is cheap right?

yes, verifying takes ~13ms on my laptop

zeegomo:
Apparently risc0 does not work on Windows. How would you be with dropping that for now?

danielSanchezQ:
> Apparently risc0 does not work on Windows. How would you be with dropping that for now?

It doesnt work here or risc0 itself does not support windows? The later would be a big problem imo

zeegomo:
I'm afraid it's the latter but:
* We plan to switch away from Risc0 as the base proofs (even beacause they are big), so you won't need it to validate incoming blocks
* We can use docker in the meantime?

danielSanchezQ:
> I'm afraid it's the latter but:
> 
>     * We plan to switch away from Risc0 as the base proofs (even beacause they are big), so you won't need it to validate incoming blocks
> 
>     * We can use docker in the meantime?

There should be no problem in moving away from windows temporary. But whatever we choose should be available widely across OSs. I'll bubble up this in our channels.

danielSanchezQ (approved):
A++ grade stuff. Thanks a lot!

PR #722, opened 2024-09-04
Trigger sampling from DA mempool

holisticode:
Trigger sampling when a new blob is seen


nomos-services/mempool/src/da/service.rs
```diff
@@ -55,6 +67,12 @@ where
     P::Item: Debug + 'static,
     P::Key: Debug + 'static,
     P::BlockId: Debug + 'static,
+    DB: DaSamplingServiceBackend<R, BlobId = P::Key> + Send,
+    DB::Blob: Debug + 'static,
+    DB::BlobId: Debug + 'static,
+    DB::Settings: Clone,
+    DN: DaSamplingNetworkAdapter,
+    R: SeedableRng + RngCore,
```
danielSanchezQ, line 75: I'm no fan of single or double digits generics, you never know what they are. Although is the rusty style. We have them mixed in the codebase.... :cry:

danielSanchezQ (approved):

nomos-services/mempool/src/da/service.rs
bacv: I think in such services it's convenient to have them short, where I found longer names helpful are in places where there are half a million of params.

danielSanchezQ: They keep growing in certain main places (like consensus where everything converges).  But for example,  `DB` tells me data base, where is not, `DN` absolutely nothing. You are forced to read the bounds what they are. But this is a personal preference.

bacv: I would actually prefer consistency, and wouldn't mind using long names. We could make this a review requirement, if others don't mind. Unfortunately it doesn't seem that rust lints support such rule.

danielSanchezQ: We can open an issue to discuss it. And then little by little refactor everything. Maybe we can come up with an intermediate solution like capitalizing the main bound. For example for `DaSamplingServiceBackend` => `DSSB`. Anyway is not a blocker, it is more about the codebase beauty :tulip:

holisticode: I myself just saw a lot of existing single and double character generics, so I assumed this to be the "standard". 

I don't like them either too much. On the other hand, using full names leads to the very long "headers", for example how will this look like if extended...

`impl<N, P, DB, DN, R> ServiceData for DaMempoolService<N, P, DB, DN, R>`

I don't think that is very readable and nice either to be honest.

I think it's a good idea to set some time apart to discuss.

PR #723, opened 2024-09-04
Trigger sampling suggestion

bacv:
The goal of this PR is to attach sampling service to the consensus and other places where it's used. 
Node tests will require correct configuration for the da network peers, for now I suggest to use `trigger_sampling` branch as base for further sampling related changes.


danielSanchezQ (approved):
You are hereby named *Generics hell master wrangler*

PR #730, opened 2024-09-06
DA: Incoming sampling request handling

bacv:
Incoming sample request handling in Sampling service.

### DIFF
--- a/nomos-da/network/core/src/protocols/sampling/behaviour.rs
+++ b/nomos-da/network/core/src/protocols/sampling/behaviour.rs
@@ -17,8 +17,10 @@ use libp2p::swarm::{
 };
 use libp2p::{Multiaddr, PeerId, Stream};
 use libp2p_stream::{Control, IncomingStreams, OpenStreamError};
+use nomos_core::da::BlobId;
+use nomos_da_messages::sampling::sample_err::SampleErrType;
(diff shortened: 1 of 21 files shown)


nomos-da/network/core/src/protocols/sampling/behaviour.rs
```diff
@@ -132,13 +143,64 @@ impl Clone for SamplingError {
                 peer_id: *peer_id,
                 error: *error,
             },
+            SamplingError::InvalidBlobId { blob_id, peer_id } => SamplingError::InvalidBlobId {
+                peer_id: *peer_id,
+                blob_id: blob_id.clone(),
```
danielSanchezQ, line 148: Nitpick: maybe is bacause of generics, but real blob_id is copy.

danielSanchezQ (approved):
Awesome work. Thanks a lot!

nomos-da/network/core/src/protocols/sampling/behaviour.rs
bacv: The types differ for these invalid blob errors, they are Vec<u8> because this is what we get from the net. In the other errors, that come from the inside the node, the type is [u8;32]. If we'll move away from protobufs, maybe this will be solved.

PR #732, opened 2024-09-06
Implemented get_blob on sampling rocksdb adapter

holisticode:
This PR adds a get_blob implementation on the rocksdb Adapter for the sampling service.

### DIFF
--- a/nomos-services/data-availability/sampling/src/storage/adapters/rocksdb.rs
+++ b/nomos-services/data-availability/sampling/src/storage/adapters/rocksdb.rs
@@ -70,8 +70,7 @@ where
                 blob_id.as_ref(),
                 &column_idx.to_be_bytes(),
             )
-            .await
-            .unwrap();
+            .await?;
             Ok(S::deserialize(blob_bytes)
                 .map(|blob| Some(blob))
                 .unwrap_or_default())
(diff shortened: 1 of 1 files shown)


nomos-services/data-availability/sampling/src/storage/adapters/rocksdb.rs
```diff
@@ -36,22 +40,44 @@ where
...
+            let blob_bytes = load_blob(
+                self.settings.blob_storage_directory.clone(),
+                blob_id.as_ref(),
+                &column_idx.to_be_bytes(),
+            )
+            .await
+            .unwrap();
+            Ok(S::deserialize(blob_bytes)
+                .map(|blob| Some(blob))
+                .unwrap_or_default())
```
bacv, line 77: Just a suggestion for grouping deserialization with the loading. The `load_blob().unwrap()` can be changed to `?`.
```suggestion
            let blob = S::deserialize(
                load_blob(
                    self.settings.blob_storage_directory.clone(),
                    blob_id.as_ref(),
                    &column_idx.to_be_bytes(),
                )
                .await?,
            )
            .map(|b| Some(b))
            .unwrap_or_default();
            Ok(blob)
```

bacv (approved):
Looks good! Thanks

nomos-services/data-availability/sampling/src/storage/adapters/rocksdb.rs
bacv: Actually your variant is better, only lets change the unwrap and it's good to go!

bacv (approved):

PR #734, opened 2024-09-09
Specification related PRs template

danielSanchezQ:


.github/PULL_REQUEST_TEMPLATE/specification.md
```diff
@@ -0,0 +1,33 @@
+## What does this implement?
+
+// Please check on this template details
+here: https://www.notion.so/Specification-hardening-0cd6547233fa43a1a03d6e360de9eb2b
```
zeegomo, line 4: are we ok with this being public?

zeegomo (approved):
Good stuff

.github/PULL_REQUEST_TEMPLATE/specification.md
danielSanchezQ: Well, repo is public but we have private things. I just hope notion is secure enough so they do not get hacked because of this link 😅

holisticode: Is this a PR template only for internal PRs or also for public PRs (PRs coming from the community)?
Because this repo is public, the PR template I guess is public too.

But then it becomes confusing in my opinion.

If it's a PR template for a public repo, this means it applies for any PR, also from the community. Therefore, all instructions for a successful PR should be public.

danielSanchezQ: Until we have the specifications public we cannot have it public. Meaning as well that no one else is gonna implement nothing specification related, hence no need to use the template. This is meant to be used for us atm.

bacv (approved):

PR #737, opened 2024-09-17
DA: Fix compliation errors in KZG backend bench

youngjoon-lee:
Updated `kzgrs-backend/benches/verifier.rs` according to the updated `DaBlob` and `DaVerifier.verify` definitions.


danielSanchezQ (approved):
Good catch. Thanks!

bacv (approved):

PR #738, opened 2024-09-17
chore: Improve coverage workflow

romanzac:
Few changes to tweak coverage workflow to occupy less disk space and run tests.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -8,9 +8,10 @@ name: Codecov
 
 jobs:
   test:
-    name: Test
     env:
-      RUSTFLAGS: -C instrument-coverage
+      CARGO_INCREMENTAL: 0
+      RUSTFLAGS: "-C instrument-coverage"
+      LLVM_PROFILE_FILE: "nomos-node-%p-%m.profraw"
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


romanzac:
Based on tests in my fork original wf would finish successfully, but no coverage would render on codecov.io
https://github.com/romanzac/nomos-node/actions/runs/10845309954
https://app.codecov.io/gh/romanzac/nomos-node/commit/11f0a3b4b8aa7f651915e371e9b2c3e70f6549a9 

Is the infrequent update on https://app.codecov.io/gh/logos-co/nomos-node/commits due to our license ?

romanzac:
Right now, coverage workflow takes around 26 minutes. If that is too long to have it after each push, I can change it as scheduled, once a day for example? 

https://github.com/romanzac/nomos-node/actions/runs/10896693219

bacv:
> Right now, coverage workflow takes around 26 minutes. If that is too long to have it after each push, I can change it as scheduled, once a day for example?

Your suggestion to have coverage workflow less frequently is great, at least for the short term. I think once a day would be a good option. Regarding the changes in PR, I would keep grcov installed with the `binstall`, that would use even less space.

bacv (approved):
Looks good, thanks!

PR #764, opened 2024-09-23
DA: Replication integration

bacv:
Replication protocol fixes and integration tests updates.


nomos-da/kzgrs-backend/src/global.rs
```diff
@@ -1,7 +1,30 @@
...
+use kzgrs::{global_parameters::global_parameters_from_file, GlobalParameters};
 use once_cell::sync::Lazy;
 
 pub static GLOBAL_PARAMETERS: Lazy<GlobalParameters> = Lazy::new(|| {
-    let mut rng = rand::thread_rng();
-    global_parameters_from_randomness(&mut rng)
+    // TODO: This is for development only, the seed needs to come from the trusted setup
+    // configuration.
+
+    global_parameters_from_file("/Users/bacv/code/logos/nomos-node/tests/kzgrs/kzgrs_test_params")
```
danielSanchezQ, line 8: Probably we can define a relative path or something for now?
I think probably this needs a bit of a refactor. Probably we should ditch `once_cell` as it is gonna be cumbersome. And will push global params as an actual parameter that we can feed. All to be done in a later PR. For now lets do the path relative otherwise it would just work on your machine :D

nomos-da/network/core/src/protocols/replication/behaviour.rs
```diff
@@ -43,7 +43,7 @@ pub struct ReplicationBehaviour<Membership> {
     /// nomos DA subnetworks
     membership: Membership,
     /// Relation of connected peers of replication subnetworks
-    connected: HashMap<PeerId, ConnectionId>,
+    connected: HashMap<PeerId, HashSet<ConnectionId>>,
```
danielSanchezQ, line 46: uhm, so why would a peer has more than a a connection id? It should be a single connnection per behaviour right?
I understand that you change it because it wasnt working, but I would like to understand what was happening.

bacv: Yes, you are right there should only be one connection per peer. But until we have an actual executor node, the nomos-cli is acting on behalf of a chosen node to make a dispersal.

When dispersing, the node A and nomos-cli are creating two connections to node B with the same peer id. As this is still draft, I didn't add a TODO comment, but will do that, unless you have a better idea how we could continue with happy path tests. 

If we include nomos-cli to the membership list with a distinct peer_id, then tests might sometimes fail, because sampling depends on it.

danielSanchezQ: > Yes, you are right there should only be one connection per peer. But until we have an actual executor node, the nomos-cli is acting on behalf of a chosen node to make a dispersal.
> 
> When dispersing, the node A and nomos-cli are creating two connections to node B with the same peer id. As this is still draft, I didn't add a TODO comment, but will do that, unless you have a better idea how we could continue with happy path tests.
> 
> If we include nomos-cli to the membership list with a distinct peer_id, then tests might sometimes fail, because sampling depends on it.

Ok, I understand now. Lets add a `todo` there and open an issue. We will catch up later. So for now we can go ahead and change it later. Thanks!

tests/src/tests/cli.rs
```diff
@@ -112,64 +118,48 @@ async fn disseminate_blob() {
...
-    let mut config = Disseminate {
-        file: Some(file.path().to_path_buf()),
-        timeout: 180,
-        app_id: APP_ID.into(),
-        index: 0,
-        columns: 2,
-        ..Default::default()
-    };
+    let node1_blobs = nodes[0]
+        .get_indexer_range(app_id.clone().try_into().unwrap(), from..to)
```
bacv, line 139: @romanzac, this is an example of using http api in the node and cli tests.

danielSanchezQ (approved):

bacv:
Closes https://github.com/logos-co/nomos-node/issues/773

PR #765, opened 2024-09-23
Mix: NetworkBehaviour

youngjoon-lee:
Implemented `libp2p::NetworkBehaviour` for the Nomos Mix protocol, which will be used for the mix nodes.

The full details of Nomos Mix protocol can be found in the [Nomos Mix: Overview](https://www.notion.so/WIP-Nomos-Mix-Overview-1058f96fb65c808abb1dcb33cc2ed417?pvs=25) (though it's not an official spec doc), but I'm summarizing the key point here to make your code reviews easier.
- Like the previous mixnet protocol, the protocol consists of two big parts:
  - Mix phase: the messages are forwarded and unwrapped through mix node
  - Broadcasting phase: The fully unwrapped message is broadcasted by the last mix node via gossipsub.
- The biggest difference is the mix phase. In the new protocol, mix nodes forward messages to all of their connected peers. In other words, the message creator doesn't need to know the IP:Port of mix nodes. Messages are forwarded in the similar way as gossiping. The mix node, who has a private key to unwrap the message, wraps it and forwards the unwrapped message to all its peers again. All other peers, who don't have a proper private key, just forward the message as it is.

This PR contains only the "mix phase":
- `network` crate: Implements `NetworkBehaviour` and `ConnectionHandler`
- `message` crate: Wrapping/unwrapping messages. Currently, this is just a mock implementation. Sphinx encoding will be implemented in next PRs.
- `queue` crate: Implements a mix queue that mixes the order of messages to be emitted to a single connection. Currently, this is just a mock.

So, you can focus on the `network` crate in this PR.

TIP: See the test first in the `nomos-mix/network/src/lib.rs`.

If you need more information, the background of design choices that I made to implement `NetworkBehaviour` can be found in [here](https://www.notion.so/Mixnet-on-Libp2p-Implementation-d8b2827c2ba4429a81ce875747934319).

What's next:
- Integrating the `network` crate to the nomos-node implementation (Substituting the old implementation).
- Implement the real algorithm for the `queue` and `message` crates.


nomos-mix/network/src/behaviour.rs
```diff
@@ -0,0 +1,231 @@
...
+            tracing::debug!("Registering event for peer {:?} to send msg", peer_id);
+            self.events.push_back(ToSwarm::NotifyHandler {
+                peer_id,
+                handler: NotifyHandler::Any,
+                event: FromBehaviour::Message(message.clone()),
+            });
+        }
+
+        Ok(())
+    }
```
danielSanchezQ, line 97: If we do this manually. Why do we need `gossipsub` for anyway?

nomos-mix/network/src/behaviour.rs
```diff
@@ -0,0 +1,231 @@
...
+            }
+        }
+    }
+
+    /// Polls for things that swarm should do.
+    fn poll(&mut self, _: &mut Context<'_>) -> Poll<ToSwarm<Self::ToSwarm, THandlerInEvent<Self>>> {
+        if let Some(event) = self.events.pop_front() {
+            Poll::Ready(event)
+        } else {
+            Poll::Pending
```
danielSanchezQ, line 228: Where do we schedule the waker for waking up?
In my previous experience this may be problematic. If you return pending without telling the waker when to reschedule for polling it may never be called again. Probably we need to store the waker. And when we push something to events we would want to call for the waker to schedule the task again.

nomos-mix/network/src/behaviour.rs
youngjoon-lee: We don't need `gossipsub` in this crate. The `gossipsub` is necessary only for the "broadcasting phase", which broadcasts fully-unwrapped messages (block proposals) to the network. Unlike the mix phase, the block proposals should (must) be delivered to all nodes in the network, as further as possible. For that, I think we need some practices that the gossipsub has, such as IHAVE/IWANT.
But, for the mix phase that this crate implements, we don't need `gossipsub`. A mix node forwards messages to all nodes that are currently connected to it. If the message was not delivered to a certain node because the node was offline, this crate do nothing to eventually deliver the message. Instead, the initial message sender (origin) should re-send the message.

danielSanchezQ: Ok. So my questions is. Does the broadcasting phase messages need to be scheduled as part of the mix protocol or not? I think I am missing something here.

nomos-mix/network/src/behaviour.rs
youngjoon-lee: Oh, let me check it again. I saw that the waker is used in the DA implementation, but also saw that the `ping` and `gossipsub` in libp2p don't call the waker. Let me recheck it, and figure out the proper way.

nomos-mix/network/src/behaviour.rs
youngjoon-lee: The broadcasting phase is totally independent from the mix phase, but is the part of the Nomos Mix protocol. In other words, the fully-unwrapped messages must be broadcasted (from their last mix node)

youngjoon-lee: Actually, that was why I implemented the composite network behaviour that has both `MixBehaviour` and `gossipsub::Behaviour`. But, I removed it because it can be done outside of this crate, even using a different swarm, as we discussed in the call.

nomos-mix/network/src/behaviour.rs
youngjoon-lee: I just found that people have been complaining about the way how libp2p guys implement `NetworkBehaviour` and `ConnectionHandler` (e.g. [here](https://discuss.libp2p.io/t/implementing-custom-protocols-with-rust-libp2p/649)). I also found an [answer](https://github.com/libp2p/rust-libp2p/issues/5147#issuecomment-1937958768) from the rust-libp2p maintainer. He says that we don't need to call `Waker.wake()` if any change on the `self.events` is made by the API defined in the `NetworkBehaviour` trait because the `Swarm` polls the behaviour again if those APIs are called. That's why any protocol implementations in rust-libp2p (including `ping` and `gossipsub`) don't call `Waker.wake()`. 
In this PR, the data structures (like `self.events`) are updated only by trait APIs. However, I am not sure if the way that the `Swarm` uses is good. I think it can be problematic if the implementer updates the data structure via additional APIs and forgets to call `Waker.wake()`. 
What do you think about this issue? I'm thinking about calling `Waker.wake()` as you did for DA, which is the standard polling convention. But, also thinking about the following sentence that the rust-libp2p maintainer wrote:
> This has been done to avoid excessive use of Wakers.

youngjoon-lee: [a3b2a5b](https://github.com/logos-co/nomos-node/pull/765/commits/a3b2a5bc97f144188563c4ca31b5ccafcd1f7e96) I think it's much better to follow the standard convention, as you did, regardless of what rust-libp2p does behind the scene. Actually, the `wake` calls that I added is not so different from what the rust-libp2p guy said, because I call the `wake` at the end of `on_connection_handler_event` each time.

danielSanchezQ: > [a3b2a5b](https://github.com/logos-co/nomos-node/pull/765/commits/a3b2a5bc97f144188563c4ca31b5ccafcd1f7e96) I think it's much better to follow the standard convention, as you did, regardless of what rust-libp2p does behind the scene. Actually, the `wake` calls that I added is not so different from what the rust-libp2p guy said, because I call the `wake` at the end of `on_connection_handler_event` each time.

I think is better. Mostly because what they claim to poll is not true. I encountered issues when not calling `wake` in where the future is never polled again and it just stalls. In this case is really simple because is just storing the waker and calling wake whenever pushing an event.

nomos-mix/network/src/behaviour.rs
danielSanchezQ: > The broadcasting phase is totally independent from the mix phase, but is the part of the Nomos Mix protocol. I’ll double check with Marcin whether the same network should be used for both phases.

I think we need to see this through. If we are forwarding messages manually here I do not think we need to have any gossipsub at all (because you are really doing the job here). If the gossiping is just the last phase in where the message is completely unwrapped we do not need gossiping from the mixnet behaviour or swarm. Let's have a chat about this in the proper channels. It feels important.

nomos-mix/network/src/behaviour.rs
youngjoon-lee: I agree. Good point. Thank you. Let’s go with the waker that I added

nomos-mix/network/src/behaviour.rs
youngjoon-lee: True. That’s why the crate in this PR does NOT have the gossipsub behaviour. This crate implements only the MIX phase, not the broadcasting phase. I believe you noticed that this PR doesn’t use gossipsub at all. I’m just checking if we don’t have any misunderstanding. 

Whether to use gossipsub can be determined outside of this crate, if the broadcasting phase should be considered as a separate protocol (potentially using a different set of nodes). Of course, we have to clarify it before merging this PR. I started a thread about this topic in Discord.

youngjoon-lee: According to our discussion made in Discord, we can go with this crate which implements only the mix phase. In the upper layer, we can implement the broadcast phase using gossipsub, which broadcasts messages returned from this crate.

danielSanchezQ (approved):

PR #766, opened 2024-09-24
Add executor binary crate

danielSanchezQ:


danielSanchezQ:
@bacv, we can merge this before or after the DA changes what you think is gonna be easier.

bacv:
> @bacv, we can merge this before or after the DA changes what you think is gonna be easier.

Executor won't be usable in tests until it has the ability to encode and disperse blobs. I'm fine with holding #764 unmerged until we have this functionality.

danielSanchezQ:
> > @bacv, we can merge this before or after the DA changes what you think is gonna be easier.
> 
> Executor won't be usable in tests until it has the ability to encode and disperse blobs. I'm fine with holding #764 unmerged until we have this functionality.

No, #764 should go ahead, its working so its fine. Same tests will have to pass when we have the executor. It will imply changing the tests a bit, but not so much. 

Lets merge #764 and then ill rebase this if needed.

danielSanchezQ:
Unrelated error.

nodes/nomos-executor/src/main.rs
```diff
@@ -0,0 +1,102 @@
...
+                registry: registry.clone(),
+            },
+            da_network: config.da_network,
+            da_indexer: config.da_indexer,
+            da_sampling: config.da_sampling,
+            da_verifier: config.da_verifier,
+            cryptarchia: config.cryptarchia,
+            #[cfg(feature = "metrics")]
+            metrics: MetricsSettings { registry },
+            storage: RocksBackendSettings {
```
bacv, line 90: Nit: recently this was made configurable via the config, we can remove the const and just use config.storage

bacv (approved):

nodes/nomos-executor/src/main.rs
danielSanchezQ: Nice!, will add, thanks!

PR #768, opened 2024-09-25
DA: Executor behaviour + swarm

danielSanchezQ:
Added executor composited behaviour and executor swarm.

### DIFF
--- a/nomos-da/network/core/src/swarm/executor.rs
+++ b/nomos-da/network/core/src/swarm/executor.rs
@@ -11,9 +11,12 @@ use tokio::sync::mpsc::{unbounded_channel, UnboundedSender};
 use tokio_stream::wrappers::UnboundedReceiverStream;
 // internal
 use crate::address_book::AddressBook;
-use crate::behaviour::validator::{ValidatorBehaviour, ValidatorBehaviourEvent};
+use crate::behaviour::executor::{ExecutorBehaviour, ExecutorBehaviourEvent};
 use crate::protocols::{
-    dispersal::validator::behaviour::DispersalEvent, replication::behaviour::ReplicationEvent,
+    dispersal::{
+        executor::behaviour::DispersalExecutorEvent, validator::behaviour::DispersalEvent,
+    },
+    replication::behaviour::ReplicationEvent,
     sampling::behaviour::SamplingEvent,
 };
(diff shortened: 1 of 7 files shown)


nomos-da/network/core/src/swarm/executor.rs
```diff
@@ -0,0 +1,174 @@
...
+            })
+            .build()
+    }
+
+    pub fn dial(&mut self, addr: Multiaddr) -> Result<(), DialError> {
+        self.swarm.dial(addr)?;
+        Ok(())
+    }
+
+    pub fn local_peer_id(&self) -> &PeerId {
```
bacv, line 93: `dial` and `local_peer_id` are a wrapper functions for interacting with the libp2p swarm. I think it's a good decision because we might tweak this logic without changing the backend. In the same sense we could expose `blob_sender` and `open_stream_sender` channels, they'll be needed by the executor also.

nomos-da/network/core/src/swarm/executor.rs
```diff
@@ -0,0 +1,174 @@
...
+        self.swarm.local_peer_id()
+    }
+
+    pub fn protocol_swarm(&self) -> &Swarm<ExecutorBehaviour<Membership>> {
```
bacv, line 97: For executor swarm to hide libp2p completely we'd need to expose `listen_on` and `sampling_sender` (same applies to the validator swarm). From the looks of it, after we gather all these swam communication channels and related data, we move the whole swarm into a thread and run it, so ownership is lost, but we don't require it anymore either.

danielSanchezQ: > For executor swarm to hide libp2p completely we'd need to expose `listen_on` and `sampling_sender` (same applies to the validator swarm). From the looks of it, after we gather all these swam communication channels and related data, we move the whole swarm into a thread and run it, so ownership is lost, but we don't require it anymore either.

Yeah, makes sense. They can be exposed later when needed maybe.
Related to the communication channels. As we have to take ownership, it is difficult to wrap them in methods, I think they are not clone. But I can try.

bacv (approved):
Looks good, thanks! (errors not related and will be solved after https://github.com/status-im/infra-misc/issues/364)

PR #796, opened 2024-09-27
chore: Da integration tests update

romanzac:
## What does this PR implement?

Batch of tests to extend existing integration tests for data availability.


tests/src/tests/cli.rs
```diff
@@ -153,13 +159,50 @@ async fn disseminate_and_retrieve() {
...
+    };
+
+    let s_b = Wire::serialize(blob);
+
     // Index zero shouldn't be empty, node 2 replicated both blobs to node 1 because they both
     // are in the same subnetwork.
     for b in node1_idx_0_blobs.iter() {
-        assert!(!b.is_empty())
+        assert!(!b.is_empty());
+        assert_eq!(**b, s_b);
```
bacv, line 201: In case of encoding "hello world" we'll get an encoded data with two columns. Each node will have blobs for both columns, and this assert might be true for only one blob with `column_idx = 0`.

romanzac: Thanks for spotting this. Fixed at https://github.com/logos-co/nomos-node/pull/796/commits/99f342ad00e2e3a8fb2df9855eacbedcf5b71fa3

bacv:
Great idea here, but let's postpone the changes to cli tests until the executor node is added and nomos-cli is updated.

PR #798, opened 2024-09-29
fix: Coverage and PR workflows - no space left on device

romanzac:
## What does this PR implement?

- Remove Android, Dotnet and Haskell support from Ubuntu Github runner to free up disk space. About 15 GB freed. 
- Add local `actions/prune-vm`
- Update coverage and PR workflows

### DIFF
--- a/.github/actions/prune-vm/action.yml
+++ b/.github/actions/prune-vm/action.yml
@@ -1,0 +1,27 @@
+# Inspired by https://github.com/AdityaGarg8/remove-unwanted-software
+# to free up disk space. Currently removes Dotnet, Android and Haskell.
+name: Remove unwanted software
+description: Default GitHub runners come with a lot of unnecessary software
+runs:
+  using: "composite"
+  steps:
+    - name: Disk space report before modification
+      shell: bash
+      run: |
+        echo "==> Available space before cleanup"
+        echo
+        df -h
+    - name: Maximize build disk space
+      shell: bash
+      run: |
+          set -euo pipefail
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /usr/local/lib/android
+          sudo rm -rf /opt/ghc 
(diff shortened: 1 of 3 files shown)


romanzac:
Hi @bacv, we can use `AdityaGarg8/remove-unwanted-software@v4.1` for PR workflow as well. Let me know, I can prepare the fix. It is a bit of a hack, but works instantly.

bacv:
I would prefer not to use this, unless https://github.com/status-im/infra-misc/issues/364 takes more time to be closed.

In that case lets fork the action and pull changes from upstream only after reviewing them. Would DST team be willing to take maintenance of this action?

@apentori do  you think we should wait for self hosted runners or add temporal solution?

romanzac:
> I would prefer not to use this, unless [status-im/infra-misc#364](https://github.com/status-im/infra-misc/issues/364) takes more time to be closed.
> 
> In that case lets fork the action and pull changes from upstream only after reviewing them. Would DST team be willing to take maintenance of this action?
> 
> @apentori do you think we should wait for self hosted runners or add temporal solution?

Another alternative is to add rm command to our wfs? It is all what it does:
https://github.com/AdityaGarg8/remove-unwanted-software/blob/8831c82abf29b34eb2caac48d5f999ecfc0d8eef/action.yml#L95

romanzac:
> > Another alternative is to add rm command to our wfs? It is all what it does:
> 
> This could be added as self contained action in `.github/actions/prune-vm` or similar with `dotnet` and `android` deps removed. We would still rely on undocumented github stuff, but it would be slightly easier to track changes done to it if something stops working.

Fixed both workflows since https://github.com/logos-co/nomos-node/pull/798/commits/4c4498236dc070ea2fa295208591345d5bd50ac8

bacv (approved):
Thank you for looking into this and solving it!

PR #801, opened 2024-10-01
DA: Executor http api

bacv:

This pr implements a separate executor http api, which reuses common endpoint handlers from the nomos node.

## Checklist

* [x] 1. Description added.

### DIFF
--- a/nodes/nomos-executor/Cargo.toml
+++ b/nodes/nomos-executor/Cargo.toml
@@ -41,5 +41,5 @@ uuid = { version = "1.10.0", features = ["v4"] }
 [features]
 default = ["tracing"]
 mixnet = ["nomos-node/mixnet"]
-metrics = []
+metrics = ["nomos-node/metrics"]
 tracing = ["nomos-node/tracing"]
--- a/nodes/nomos-executor/Cargo.toml
+++ b/nodes/nomos-executor/Cargo.toml
@@ -4,14 +4,36 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+async-trait = "0.1"
+axum = { version = "0.6" }
+clap = { version = "4.5.13", features = ["derive"] }
 color-eyre = "0.6.0"
+hyper = { version = "0.14", features = ["full"] }
+kzgrs-backend = { path = "../../nomos-da/kzgrs-backend" }
+nomos-api = { path = "../../nomos-services/api" }
+nomos-core = { path = "../../nomos-core" }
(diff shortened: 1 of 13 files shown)


nodes/nomos-executor/Cargo.toml
```diff
@@ -4,17 +4,42 @@ version = "0.1.0"
...
 uuid = { version = "1.10.0", features = ["v4"] }
-serde_yaml = "0.9.34+deprecated"
 
 [features]
 default = ["tracing"]
 mixnet = ["nomos-node/mixnet"]
-metrics = ["nomos-node/metrics"]
-tracing = ["nomos-node/tracing"]
\ No newline at end of file
+metrics = []
```
danielSanchezQ, line 44: shouldn't we activate the nomos-node lib metrics feature as well? :thinking:

nodes/nomos-executor/src/api/backend.rs
```diff
@@ -0,0 +1,295 @@
...
+
+pub struct AxumBackend<
+    A,
+    B,
+    C,
+    M,
+    V,
+    VB,
+    T,
+    S,
```
danielSanchezQ, line 51: Now I know my ABC :ab:

nodes/nomos-node/src/api/backend.rs
```diff
@@ -0,0 +1,295 @@
...
+                        SamplingNetworkAdapter,
+                        SamplingRng,
+                        SamplingStorage,
+                        SIZE,
+                    >,
+                ),
+            )
+            .route("/da/add_blob", routing::post(add_blob::<A, B, M, VB, S>))
+            .route(
+                "/da/get_range",
```
danielSanchezQ, line 258: Yeah, nice changes. Better to compose here than the over-abstraction of converting the endopoints in types. It was my fault though, have been thinking about this for time. So, nice catch.

danielSanchezQ (approved):
Nice :rocket: !

nodes/nomos-executor/Cargo.toml
bacv: For completeness I've added the feature back, but basically we are not using `nomos-node/metrics` related code in executor. Each node constructs their own registry according to the feature flag at the moment. I'll revisit the feature flags when working on the metrics expansion.

danielSanchezQ: Cool. I just wanted to understand as I was missing something indeed. Thanks!

PR #802, opened 2024-10-01
kzg main testing feature update and variable renaming

TakeASwing-420:

> [!NOTE]  
> Review  
>
the [Specification to Implementation Process document](https://www.notion.so/Specification-to-Implementation-Process-0cd6547233fa43a1a03d6e360de9eb2b).

> [!IMPORTANT]  
> explained.  

> [!IMPORTANT]  
> advancing  
> the review.

> [!IMPORTANT]

> [!IMPORTANT]
    modification/solutions.

## Checklist

> [!WARNING]  

* [ ] 1. Description added.

### DIFF
--- a/nomos-da/kzgrs/Cargo.toml
+++ b/nomos-da/kzgrs/Cargo.toml
@@ -22,7 +22,6 @@ num-traits = "0.2.18"
 rand = "0.8.5"
 rayon = { version = "1.10", optional = true }
 
-
 [dev-dependencies]
 divan = "0.1"
 
@@ -39,7 +38,7 @@ name = "fk20"
 harness = false
 
 [features]
-default = ["single"]
+default = ["single", "rayon"]
 single = []
 parallel = [
     "rayon",
--- a/nomos-da/kzgrs/src/kzg.rs
+++ b/nomos-da/kzgrs/src/kzg.rs
@@ -24,7 +24,7 @@ pub fn commit_polynomial(
         .map(|(commitment, _)| commitment)
 }
 
-/// Compute a witness polynomial in that satisfies `witness(x) = (f(x)-v)/(x-u)`
+/// Compute a quotient polynomial 'q' that satisfies `q(x) = (f(x)-v)/(x-u)`
 pub fn generate_element_proof(
     element_index: usize,
     polynomial: &DensePolynomial<Fr>,
@@ -33,13 +33,11 @@ pub fn generate_element_proof(
     domain: GeneralEvaluationDomain<Fr>,
 ) -> Result<Proof<Bls12_381>, KzgRsError> {
     let u = domain.element(element_index);
-    // Instead of evaluating over the polynomial, we can reuse the evaluation points from the rs encoding
-    // let v = polynomial.evaluate(&u);
     let v = evaluations.evals[element_index];
     let f_x_v = polynomial + &DensePolynomial::<Fr>::from_coefficients_vec(vec![-v]);
     let x_u = DensePolynomial::<Fr>::from_coefficients_vec(vec![-u, Fr::one()]);
-    let witness_polynomial: DensePolynomial<_> = &f_x_v / &x_u;
-    let proof = commit_polynomial(&witness_polynomial, global_parameters)?;
+    let quotient_polynomial: DensePolynomial<_> = &f_x_v / &x_u;
+    let proof = commit_polynomial(&quotient_polynomial, global_parameters)?;
     let proof = Proof {
         w: proof.0,
         random_v: None,
@@ -66,6 +64,7 @@ pub fn verify_element_proof(
 }
 
 #[cfg(test)]
+#[cfg(feature = "rayon")]
 mod test {
     use crate::common::{bytes_to_evaluations, bytes_to_polynomial};
     use crate::kzg::{commit_polynomial, generate_element_proof, verify_element_proof};
@@ -75,6 +74,7 @@ mod test {
     use ark_poly_commit::kzg10::{UniversalParams, KZG10};
     use once_cell::sync::Lazy;
     use rand::{thread_rng, Fill};
+    use rayon::prelude::*;
 
     const COEFFICIENTS_SIZE: usize = 16;
     static GLOBAL_PARAMETERS: Lazy<UniversalParams<Bls12_381>> = Lazy::new(|| {
@@ -89,13 +89,11 @@ mod test {
 
     static DOMAIN: Lazy<GeneralEvaluationDomain<Fr>> =
         Lazy::new(|| GeneralEvaluationDomain::new(COEFFICIENTS_SIZE).unwrap());
+
     #[test]
     fn test_poly_commit() {
         let poly = DensePolynomial::from_coefficients_vec((0..10).map(|i| Fr::from(i)).collect());
-        assert!(matches!(
-            commit_polynomial(&poly, &GLOBAL_PARAMETERS),
-            Ok(_)
-        ));
+        assert!(matches!(commit_polynomial(&poly, &GLOBAL_PARAMETERS), Ok(_)));
     }
 
     #[test]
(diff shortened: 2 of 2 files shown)


nomos-da/kzgrs/Cargo.toml
```diff
@@ -22,7 +22,6 @@ num-traits = "0.2.18"
 rand = "0.8.5"
 rayon = { version = "1.10", optional = true }
 
-
```
bacv, line 25: This is unnecessary change

nomos-da/kzgrs/Cargo.toml
```diff
@@ -39,7 +38,7 @@ name = "fk20"
 harness = false
 
 [features]
-default = ["single"]
+default = ["single", "rayon"]
```
bacv, line 41: To enable `rayon` support please use `parallel` feature, default for now should stay as is.

nomos-da/kzgrs/src/kzg.rs
```diff
@@ -24,7 +24,7 @@ pub fn commit_polynomial(
         .map(|(commitment, _)| commitment)
 }
 
-/// Compute a witness polynomial in that satisfies `witness(x) = (f(x)-v)/(x-u)`
+/// Compute a quotient polynomial 'q' that satisfies `q(x) = (f(x)-v)/(x-u)`
```
bacv, line 27: In the context of KZG (Kate-Zaverucha-Goldberg) polynomial commitments, the concept of a witness polynomial is used to verify the evaluation of a polynomial at a specific point. Lets keep the original comment.

nomos-da/kzgrs/src/kzg.rs
```diff
@@ -33,13 +33,11 @@ pub fn generate_element_proof(
...
 ) -> Result<Proof<Bls12_381>, KzgRsError> {
     let u = domain.element(element_index);
-    // Instead of evaluating over the polynomial, we can reuse the evaluation points from the rs encoding
-    // let v = polynomial.evaluate(&u);
     let v = evaluations.evals[element_index];
     let f_x_v = polynomial + &DensePolynomial::<Fr>::from_coefficients_vec(vec![-v]);
     let x_u = DensePolynomial::<Fr>::from_coefficients_vec(vec![-u, Fr::one()]);
-    let witness_polynomial: DensePolynomial<_> = &f_x_v / &x_u;
-    let proof = commit_polynomial(&witness_polynomial, global_parameters)?;
+    let quotient_polynomial: DensePolynomial<_> = &f_x_v / &x_u;
```
bacv, line 39: For context lets keep `witness_polynomial` variable name

nomos-da/kzgrs/src/kzg.rs
```diff
@@ -66,6 +64,7 @@ pub fn verify_element_proof(
 }
 
 #[cfg(test)]
+#[cfg(feature = "rayon")]
```
bacv, line 67: To test with this feature it's better to create a separate test.

nomos-da/kzgrs/src/kzg.rs
```diff
@@ -89,13 +89,11 @@ mod test {
...
         Lazy::new(|| GeneralEvaluationDomain::new(COEFFICIENTS_SIZE).unwrap());
+
     #[test]
     fn test_poly_commit() {
         let poly = DensePolynomial::from_coefficients_vec((0..10).map(|i| Fr::from(i)).collect());
-        assert!(matches!(
-            commit_polynomial(&poly, &GLOBAL_PARAMETERS),
-            Ok(_)
-        ));
+        assert!(matches!(commit_polynomial(&poly, &GLOBAL_PARAMETERS), Ok(_)));
```
bacv, line 96: Before submitting please run:
``` bash
cargo fmt --all -- --check
cargo clippy args: --all --no-default-features --features libp2p -- --deny warnings
cargo clippy args: --all --no-default-features --features libp2p,mixnet -- --deny warnings
```

nomos-da/kzgrs/src/kzg.rs
```diff
@@ -89,13 +89,11 @@ mod test {
 
     static DOMAIN: Lazy<GeneralEvaluationDomain<Fr>> =
         Lazy::new(|| GeneralEvaluationDomain::new(COEFFICIENTS_SIZE).unwrap());
+
```
bacv, line 92: Lets leave previously approved formatting as is.

bacv (changes requested):

TakeASwing-420:
@bacv, I have done the necessary changes as per requested. Please check if there still any issue with the code.

nomos-da/kzgrs/src/kzg.rs
```diff
@@ -76,6 +76,10 @@ mod test {
     use once_cell::sync::Lazy;
     use rand::{thread_rng, Fill};
 
+    // Import rayon only if parallel feature is enabled
+    #[cfg(feature = "parallel")]
+    use rayon::prelude::*;
+
```
danielSanchezQ, line 82: As I already said this is not the way to use this. What is the meaning of using rayon here?

danielSanchezQ (changes requested):

nomos-da/kzgrs/src/kzg.rs
```diff
@@ -132,4 +136,47 @@ mod test {
...
+        let mut bytes: [u8; 310] = [0; 310];
+        let mut rng = thread_rng();
+        bytes.try_fill(&mut rng).unwrap();
+        let (eval, poly) = bytes_to_polynomial::<31>(&bytes, *DOMAIN).unwrap();
+        let commitment = commit_polynomial(&poly, &GLOBAL_PARAMETERS).unwrap();
+        let proofs: Vec<_> = (0..10)
+            .map(|i| generate_element_proof(i, &poly, &eval, &GLOBAL_PARAMETERS, *DOMAIN).unwrap())
+            .collect();
+
+        eval.evals
```
TakeASwing-420, line 152: Here I reused the same eval struct instance for iterating instead of creating the vector from `bytes_to_evaluations`...is it okay?

nomos-da/kzgrs/src/kzg.rs
```diff
@@ -109,18 +107,23 @@ mod test {
...
-                &GLOBAL_PARAMETERS
-            ));
-            // verification fails for other items
-            for ii in i + 1..10 {
+
+        evaluations
+        .par_iter()
+        .zip(proofs.par_iter())
+        .enumerate() 
+        .for_each(|(i, (element, proof))| {
```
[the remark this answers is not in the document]
TakeASwing-420: Yes actually every verify_element_proof process are independent from each other..so I thought of using data parallelization.

nomos-da/kzgrs/src/kzg.rs
```diff
@@ -132,4 +136,47 @@ mod test {
...
         }
     }
+
+    #[test]
+    #[cfg(feature = "parallel")]
+    fn parallelized_generate_proof_and_validate() {
+        let mut bytes: [u8; 310] = [0; 310];
+        let mut rng = thread_rng();
+        bytes.try_fill(&mut rng).unwrap();
+        let (eval, poly) = bytes_to_polynomial::<31>(&bytes, *DOMAIN).unwrap();
```
danielSanchezQ, line 146: I think we have a constant for those `31` it would be better to use it actually.

nomos-da/kzgrs/src/kzg.rs
danielSanchezQ: Yes, I just do not see the value of this tests vs the ones there are in place already. Could you elaborate?

nomos-da/kzgrs/src/kzg.rs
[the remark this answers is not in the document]
danielSanchezQ: But then why a new test? I believe we have this kind of test in place already. Adding a new test that does the same but faster do not really adds too much value :sweat_smile: 
It is a good idea to parallelize though.

TakeASwing-420:
@danielSanchezQ, one thing I forgot to mention that in the original testing function you are iterating the inner loop from j = i + 1 to 10(exclusive)
``
for ii in i + 1..10 {
                assert!(!verify_element_proof
``
so for index 3 you are checking for i = 4 to i = 9... but I am iterating from 0 to 10 except the `i`th index....
that's why my test and yours are taking same time 😅(almost 24 sec in my system)

TakeASwing-420:
I have updated my code.. now you can check the runtime diffrence between the two tests
in my system mine is takin 15 sec but the original is taking 26 sec

danielSanchezQ:
> I have updated my code.. now you can check the runtime diffrence between the two tests in my system mine is takin 15 sec but the original is taking 26 sec also added rayon as dev dependency

Sorry, I may not be explaining myself properly. 

Yes, I understand this is faster and that is a good idea. The thing is that what I do not see necessary is to have 2 tests that test the same one that is faster than the other. So we have two options:
* Remove the original test
* Adapt the original tests to be parallelized.

I would vote for the second. Also it would be beneficial it all possible tests are updated not just this single one. At least to keep consistency. Otherwise I think this PR should be discarded because of those motives. 
Does it makes sense to you @TakeASwing-420 ? Effort is greatly appreciated btw!

TakeASwing-420:
@danielSanchezQ, do you want just a single test function named `generate_proof_and_validate()` with parallelization feature?
Actually, I pushed the previous commit to actually check the difference😅... I would refactor the original testing function and remove mine with a new commit in that case.

> Also it would be beneficial it all possible tests are updated not just this single one. At least to keep consistency. Otherwise I think this PR should be discarded because of those motives.

@danielSanchezQ, of course I would love to contribute to this project... I have just started exploring the zk and protocol enngineering domain for the past 2 months and working my way through it.. As I gradually explore other sections of this project I will surely contribute wherever possible..

danielSanchezQ:
> @danielSanchezQ, do you want just a single test function named `generate_proof_and_validate()` with parallelization feature? Actually, I pushed the previous commit to actually check the difference😅... I would refactor the original testing function and remove mine with a new commit in that case.

Yes, just modify the original test and add the parallelization, lets do this:

* Close this PR
* Open a new one with:
  * Add `rayon` as a `dev-dependency`
  * Modify the original test to add parallelization
  
Sounds good?

PR #811, opened 2024-10-08
Add Executor client

danielSanchezQ:

This PR introduces a new executor http client. It just have a single method to publish blob for now.
There is a tiny refactor of the http endpoints in nomos-node and nomos-executor so paths can be reusable.

Yes

No specification related

--
--
## Checklist
* [ ] 1. Description added.

### DIFF
--- a/nodes/nomos-node/src/api/handlers.rs
+++ b/nodes/nomos-node/src/api/handlers.rs
@@ -201,7 +201,7 @@ where
     post,
     path = "/da/add_blob",
     responses(
-        (status = 200, description = "Attestation for DA blob", body = Option<Attestation>),
+        (status = 200, description = "Blob to be published received"),
         (status = 500, description = "Internal server error", body = String),
     )
 )]
--- a/nodes/nomos-node/src/api/paths.rs
+++ b/nodes/nomos-node/src/api/paths.rs
@@ -1,0 +1,11 @@
+pub const CL_METRICS: &str = "/cl/metrics";
+pub const CL_STATUS: &str = "/cl/status";
+pub const CRYPTARCHIA_INFO: &str = "/cryptarchia/info";
+pub const CRYOTARCHIA_HEADERS: &str = "/cryptarchia/headers";
+pub const DA_ADD_BLOB: &str = "/da/add-blob";
+pub const DA_GET_RANGE: &str = "/da/get-range";
+pub const NETWORK_INFO: &str = "/network/info";
+pub const STORAGE_BLOCK: &str = "/storage/block";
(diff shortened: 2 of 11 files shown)


nodes/nomos-node/src/api/handlers.rs
```diff
@@ -201,7 +201,7 @@ where
     post,
     path = "/da/add_blob",
     responses(
-        (status = 200, description = "Attestation for DA blob", body = Option<Attestation>),
+        (status = 200, description = "Blob to be published received"),
```
bacv, line 204: Nit: We could readd the `body`, as we are actually responding with it, even though attestation in this case is `()` (unit type).

bacv (approved):
Looks good. In the future, we might want to have executor client trait, so that we could plugin open api generated client. But this is to be discussed and might not actually be part of nomos node codebase.

nodes/nomos-node/src/api/handlers.rs
danielSanchezQ: > Nit: We could readd the `body`, as we are actually responding with it, even though attestation in this case is `()` (unit type).

I think we should remove the attestation though. Also we are just gonna remove this or change the endpoint so we actually do not need to return anything.

danielSanchezQ:
> Looks good. In the future, we might want to have executor client trait, so that we could plugin open api generated client. But this is to be discussed and might not actually be part of nomos node codebase.

Yeah, I thought about this. But I'm not sure how worth it is. I haven't used that kind of generated stuff from rust yet, so I do not have a clear opinion.

nodes/nomos-node/src/api/handlers.rs
bacv: Since we are still getting `DispersalResponse` from individual nodes, `Attestation` could be renamed to something similar. This probably could be done in a seperate PR.

nodes/nomos-node/src/api/paths.rs
```diff
@@ -0,0 +1,11 @@
+pub const CL_METRICS: &str = "/cl/metrics";
+pub const CL_STATUS: &str = "/cl/status";
+pub const CRYPTARCHIA_INFO: &str = "/cryptarchia/info";
+pub const CRYOTARCHIA_HEADERS: &str = "/cryptarchia/headers";
+pub const DA_ADD_BLOB: &str = "/da/add-blob";
+pub const DA_GET_RANGE: &str = "/da/get-range";
```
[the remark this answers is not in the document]
danielSanchezQ: Ah, let me update that!

PR #821, opened 2024-10-14
Mix service integration

youngjoon-lee:

This PR:
- removes all old mixnet-related code (No `mixnet` feature flag now)
- implements the Mix service (in the `nomos-service/mix`)
- integrates the Mix service to the node

> [!NOTE]  
> I recommend to review each commit separately.

> [!IMPORTANT]  
> explained.  

- The Mix service implements the `Mix Network` stack that is described in the `Architecture` diagram in the [Nomos Mix: Introduction](https://www.notion.so/WIP-Nomos-Network-Introduction-1058f96fb65c808abb1dcb33cc2ed417?pvs=4#e50b23ed98a74ae281187820d13268c6). This document is not linked in our official spec page yet, but is the latest one currently. In engineer-speak, block proposals are first propagated through the `Mix Network` in the form of Sphinx packet. The main purpose of the `Mix Network` is hiding block proposers from network observers. And then, block proposals returned from the `Mix Network` are broadcasted through the `Broadcast Network`. Therefore, the consensus service uses both the Mix service (for `Mix Network`) and the Network service (for `Broadcast Network`) SEPARATELY. The network service is the one that we've already had, but without old mixnet stuff.
- The node uses different IP/port for each Mix service and Network service. In other words, it uses separate network connections for each. This is mentioned in the [Nomos Mix: Tier 1 — Persistent Transmission Module](https://www.notion.so/WIP-Nomos-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#11c8f96fb65c8090ac04c8a7cee3d7b3)

> [!NOTE]  
> I recommend to not focus on the mix protocol details when reviewing this PR. The protocol details have been implemented in the `nomos-mix` crate by the previous PR (though some details must be updated). This PR just implements the Mix service that uses the `nomos-mix` crate, and uses it in the consensus service.

> [!IMPORTANT]  
> advancing  
> the review.

> [!IMPORTANT]

The spec that this PR implements is accurate and complete. In other words, the fact that the `Mix Network` is used by consensus service before messages are broadcasted through the `Broadcast Network` is true. 

> [!IMPORTANT]
    modification/solutions.

No

## Checklist

> [!WARNING]  

* [ ] 1. Description added.


danielSanchezQ:
This should have probably be better in 3 PRs though. One removes old. One adds. One integrates :sweat_smile:

danielSanchezQ:
> > This should have probably be better in 3 PRs though. One removes old. One adds. One integrates 😅
> 
> Yeah, I thought about doing so (as I've done so far). But, I recalled our conversation that it might be better to put both removing old and adding new into one PR. Also, I was not sure how to write PR descriptions according to our new template, if I split this into multiple PRs. But now I think it's doable and better for easy code review.

As now it is safe to remove the old mixnet (feature gated) and keeping the network service, everything should work anyway. There is no problem but there is a lot of changes to track here and I think it would be more understandable separated. At least for review. 
You can copy/past the PR description for the 2 last of them which are the implementation and the integration.

youngjoon-lee:
Closing. I'll split this into three PRs.

PR #823, opened 2024-10-14
Mix: Add mix service

youngjoon-lee:

This PR implements the mix service that uses the `NetworkBehaviour` defined in the `nomos-mix` crate. While the `nomos-mix` crate implements the mix protocol details, the mix service just creates a libp2p swarm by just using the `nomos-mix` crate. This service can be used by the consensus service along with the network service. The internal code structure of this service is similar as the network service.

> [!IMPORTANT]  
> explained.  

- The mix service implements the `Mix Network` stack that is described in the `Architecture` diagram in the [Nomos Mix: Introduction](https://www.notion.so/WIP-Nomos-Network-Introduction-1058f96fb65c808abb1dcb33cc2ed417?pvs=4#e50b23ed98a74ae281187820d13268c6). This document is not linked in our official spec page yet, but is the latest one currently. In engineer-speak, block proposals should be first propagated through the `Mix Network` in the form of Sphinx packet. The main purpose of the `Mix Network` is hiding block proposers from network observers. And then, block proposals returned from the `Mix Network` are broadcasted through the `Broadcast Network`. Therefore, the consensus service should use both the mix service (for `Mix Network`) and the network service (for `Broadcast Network`). The network service is the one that we've already had. 
- The node should use different IP/port for each Mix service and Network service. In other words, it uses separate network connections for each. This is mentioned in the [Nomos Mix: Tier 1 — Persistent Transmission Module](https://www.notion.so/WIP-Nomos-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#11c8f96fb65c8090ac04c8a7cee3d7b3). That's why we need the mix service separated from the network service.

> [!NOTE]  
> I recommend to not focus on the mix protocol details when reviewing this PR. The protocol details have been implemented in the `nomos-mix` crate by the previous PR (though some details must be updated). This PR just implements the Mix service that uses the `nomos-mix` crate.

> [!IMPORTANT]  
> advancing  
> the review.

> [!IMPORTANT]

Yes.

> [!IMPORTANT]
    modification/solutions.

No

## Checklist

> [!WARNING]  

* [x] 1. Description added.


nomos-services/mix/src/backends/libp2p.rs
```diff
@@ -0,0 +1,224 @@
...
+    task: JoinHandle<()>,
+    msgs_tx: mpsc::Sender<Libp2pNetworkBackendMessage>,
+    events_tx: broadcast::Sender<Libp2pNetworkBackendEvent>,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Libp2pNetworkBackendSettings {
+    pub listening_address: Multiaddr,
+    #[serde(with = "secret_key_serde", default = "ed25519::SecretKey::generate")]
+    pub node_key: ed25519::SecretKey,
```
danielSanchezQ, line 33: Is this key used for both encrypting messages as for the peer id generation?

nomos-services/mix/src/backends/libp2p.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+fn extract_peer_id(multiaddr: &Multiaddr) -> Option<PeerId> {
+    for protocol in multiaddr.iter() {
+        if let Protocol::P2p(peer_id) = protocol {
+            return Some(peer_id);
+        }
+    }
+    None
+}
```
danielSanchezQ, line 224: Nitpick, its totally fine like this, but you can use [`find](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find)` or [`find_map`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find_map) with iterators:

```suggestion
fn extract_peer_id(multiaddr: &Multiaddr) -> Option<PeerId> {
    multiaddr
        .iter()
        .find_map(|protocol| if let Some(peer_id) = protocol {Some(peer_id)} else { None })
}
```

nomos-services/mix/src/backends/libp2p.rs
```diff
@@ -0,0 +1,224 @@
...
+                None => false,
+            })
+            .choose_multiple(&mut rand::thread_rng(), config.peering_degree)
+            .iter()
+            .cloned()
+            .for_each(|addr| {
+                if let Err(e) = swarm.dial(addr.clone()) {
+                    tracing::error!("failed to dial to {:?}: {:?}", addr, e);
+                }
+            });
```
danielSanchezQ, line 93: Eventually this should probably be part of the mixnet behavior as well. 
We will probably handle it internally in the da behaviors too. But for now this should do.

danielSanchezQ (approved):
Super clean service, thanks!

nomos-services/mix/src/backends/libp2p.rs
youngjoon-lee: Sorry, we haven't clearly defined what keys are needed in the specification. We have to do it ASAP. @madxor Let's define the [node credentials](https://www.notion.so/Nomos-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#11c8f96fb65c80f8b754fd18c331514b) clearly together.

In practice, we have two parts that a certain key should be used (except the incentivization part for now):
1. Establishing "secure" connections
2. Sphinx encryption

The `node_key` in this code is for the 1. Since we're using QUIC and libp2p, the `node_key` is used for [TLS 1.3 handshake](https://docs.libp2p.io/concepts/transports/quic/#quic-in-libp2p) which is automatically done by libp2p. During the handshake, the `node_key` is used to generate a self-signed certificate for authentication. During the handshake, an ephemeral symmetric key is shared between peers, and that is used for encrypting traffic. 

For the 2, I'm thinking of using X25519, which is what we had used in the previous implementation. I didn't add it in our codebase yet because we just use a mock instead of Sphinx. But I'll add it soon.

nomos-services/mix/src/backends/libp2p.rs
danielSanchezQ: Maybe we can add a `todo` here.

nomos-services/mix/src/backends/libp2p.rs
youngjoon-lee: Added a comment to explain the role of `node_key`: 3a6de0887c6f116db0d08377cf0af83f7d673562

nomos-services/mix/src/backends/libp2p.rs
youngjoon-lee: Thank you 3a6de0887c6f116db0d08377cf0af83f7d673562

nomos-services/mix/src/backends/libp2p.rs
youngjoon-lee: Good point 3a6de0887c6f116db0d08377cf0af83f7d673562

bacv (approved):
Looks great 🚀

PR #824, opened 2024-10-14
Mix: Integrate mix service to the system

youngjoon-lee:

This PR integrates the mix service (defined in #823) to the system. Specifically, the consensus service uses the mix service to propagate block proposals through the Mix Network. After that, the consensus service (of the exit mix node) broadcasts the block proposals through the Broadcast Network using the network service.

> [!IMPORTANT]  
> explained.  

- The relationship between the Mix Network and the Broadcast Network is defined in the `Architecture` diagram in the [Nomos Mix: Introduction](https://www.notion.so/WIP-Nomos-Network-Introduction-1058f96fb65c808abb1dcb33cc2ed417?pvs=4#e50b23ed98a74ae281187820d13268c6). This document is not linked in our official spec page yet, but is the latest one currently. In engineer-speak, block proposals should be first propagated through the `Mix Network` in the form of Sphinx packet. The main purpose of the `Mix Network` is hiding block proposers from network observers. And then, block proposals returned from the `Mix Network` are broadcasted through the `Broadcast Network`. That's why the consensus service uses both the mix service (for `Mix Network`) and the network service (for `Broadcast Network`).
- The node should use different IP/port for each Mix service and Network service. In other words, it uses separate network connections for each. This is mentioned in the [Nomos Mix: Tier 1 — Persistent Transmission Module](https://www.notion.so/WIP-Nomos-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#11c8f96fb65c8090ac04c8a7cee3d7b3). That's why we need the mix service separated from the network service.

> [!IMPORTANT]  
> advancing  
> the review.

> [!IMPORTANT]

Yes. The fact that the consensus service should use both the mix service and the network service separately is true.

> [!IMPORTANT]
    modification/solutions.

No.

## Checklist

> [!WARNING]  

* [x] 1. Description added.


nodes/nomos-node/src/config.rs
```diff
@@ -72,6 +74,24 @@ pub struct NetworkArgs {
     pub initial_peers: Option<Vec<Multiaddr>>,
 }
 
+#[derive(Parser, Debug, Clone)]
+pub struct MixArgs {
+    #[clap(long = "mix-addr", env = "MIX_ADDR")]
+    mix_addr: Option<Multiaddr>,
+
+    #[clap(long = "mix-node-key", env = "MIX_NODE_KEY")]
+    mix_node_key: Option<String>,
```
danielSanchezQ, line 83: nitpick: I would rather try to have something that is not a string for this. Either the raw bytes or the key type directly. Then clap probably have something to transform to that type directly in the cli. Otherwise we have to deal with errors ourselves instead of delegating to clap itself.

nomos-mix/message/src/lib.rs
```diff
@@ -4,7 +4,7 @@ pub use error::Error;
 
 use sha2::{Digest, Sha256};
 
-pub const MSG_SIZE: usize = 1024;
+pub const MSG_SIZE: usize = 2048;
```
danielSanchezQ, line 7: why are we doubling this?

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -446,6 +469,17 @@ where
...
+                        match wire::deserialize::<Block<ClPool::Item, DaPool::Item>>(&mixed_msg) {
+                            Ok(block) => {
+                                adapter.broadcast(NetworkMessage::Block(block)).await;
+                            },
+                            Err(e) => {
+                                tracing::error!("Error deserializing mixed message: {e}");
+                            },
+                        }
+                    }
+
```
danielSanchezQ, line 482: Why is this in consensus?
I think this is wrong. We are not validating anything in the block yet. We are just piping to somewhere else. This definitely should be done by the mixnet service itself. Consensus should just read from the public network and write to the mixntet (?)

nomos-services/cryptarchia-consensus/src/mix/adapters/libp2p.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+#[async_trait::async_trait]
+impl NetworkAdapter for LibP2pAdapter {
+    type Backend = Libp2pNetworkBackend;
+
+    async fn new(
+        network_relay: OutboundRelay<<NetworkService<Self::Backend> as ServiceData>::Message>,
+    ) -> Self {
+        let relay = network_relay.clone();
```
danielSanchezQ, line 29: No need for clone here I think, `network_relay` is already owned.

nomos-services/cryptarchia-consensus/src/mix/adapters/libp2p.rs
```diff
@@ -0,0 +1,86 @@
...
+        let mixed_msgs = broadcast::Sender::new(BUFFER_SIZE);
+        let mixed_msgs_sender = mixed_msgs.clone();
+        // this wait seems to be helpful in some cases since we give the time
+        // to the network to establish connections before we start sending messages
+        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
danielSanchezQ, line 34: Please add a todo here. We are planning to add a status system to await for services readyness, hopefully we can remove this kind of stuff.

nomos-services/cryptarchia-consensus/src/mix/adapters/libp2p.rs
```diff
@@ -0,0 +1,86 @@
...
+                            tracing::error!("error sending mixed message to consensus: {e}");
+                        }
+                    }
+                    Err(RecvError::Lagged(n)) => {
+                        tracing::error!("lagged messages: {n}")
+                    }
+                    Err(RecvError::Closed) => unreachable!(),
+                }
+            }
+        });
```
danielSanchezQ, line 63: I think this as a task is unnecessary?
When requesting for a subscriptrion you have a channel that can be transformed into a stream and map the necessary operations. This should be done on demand in the `mixed_msgs_stream` method directly.

nomos-services/cryptarchia-consensus/src/mix/adapters/libp2p.rs
```diff
@@ -0,0 +1,86 @@
...
+            .send(NetworkMsg::Process(Libp2pNetworkBackendMessage::Mix(
+                message,
+            )))
+            .await
+        {
+            tracing::error!("error sending message to mix network: {e}: {msg:?}",);
+        }
+    }
+
+    async fn mixed_msgs_stream(&self) -> BoxedStream<Vec<u8>> {
```
danielSanchezQ, line 83: nipick: `mixed_messages_stream`, even if its chattier lets be explicit.

nomos-services/cryptarchia-consensus/src/mix/adapters/mock.rs
```diff
@@ -0,0 +1,78 @@
...
+    type Backend = Mock;
+
+    async fn new(
+        network_relay: OutboundRelay<<NetworkService<Self::Backend> as ServiceData>::Message>,
+    ) -> Self {
+        let relay = network_relay.clone();
+        let mixed_msgs = broadcast::Sender::new(BUFFER_SIZE);
+        let mixed_msgs_sender = mixed_msgs.clone();
+
+        tokio::spawn(async move {
```
danielSanchezQ, line 30: Same stuff for this task.

Also, are we using the mock version?

danielSanchezQ (changes requested):
A few things to polish.
Btw, now we have the executor node binary as well, mix service will need to be added to it as well here. It shouldn't be too much more.

nomos-services/cryptarchia-consensus/src/lib.rs
youngjoon-lee: I was also thinking about this when implementing this code. To be clear, like the network adapter that returns a stream of `Block`, do you mean that it would be better to make the mix adapter return a stream of `Block` (not a stream of `Vec<u8>`)?

nomos-mix/message/src/lib.rs
youngjoon-lee: It's because the size of block proposal is between 1024 and 2048 currently in our tests. But, this file is just a mock, and we haven't determined the proper block size yet. So, I'm using just 2028 here for now.

youngjoon-lee: I think it's better to put this in the configuration.

nodes/nomos-node/src/config.rs
youngjoon-lee: I agree. I'll try it in other PRs because there are many `String`s for node keys in this file.

nomos-services/cryptarchia-consensus/src/lib.rs
danielSanchezQ: > I was also thinking about this when implementing this code. To be clear, like the network adapter that returns a stream of `Block`, do you mean that it would be better to make the mix adapter return a stream of `Block` (not a stream of `Vec<u8>`)?

Uhm, from the mixnet perspective its fine its just bytes. But what I mean is that this relaying and deserializing should not be done in consensus at all.

youngjoon-lee: I think I don't catch your point correctly. Do you think that reading messages from the mix network and sending them to the broadcast network should be done inside the mix service? The mix service implements only the mix network. And, the user of the mix service (i.e. consensus service) can choose what to do with messages returned from the mix network.

youngjoon-lee: Ah, okay. We don't need to stick to the following sentence that I wrote:
> The mix service implements only the mix network.

We can make the mix service use the broadcast service. But, I'd like to check if you're talking about this point.

nodes/nomos-node/src/config.rs
youngjoon-lee: Added TODO 851c7af290fc44b51179552e5064357339b15b91

nomos-services/cryptarchia-consensus/src/mix/adapters/libp2p.rs
youngjoon-lee: Added TODO 851c7af290fc44b51179552e5064357339b15b91

nomos-services/cryptarchia-consensus/src/mix/adapters/libp2p.rs
youngjoon-lee: Fixed 851c7af290fc44b51179552e5064357339b15b91

nomos-services/cryptarchia-consensus/src/mix/adapters/libp2p.rs
youngjoon-lee: True. Fixed in 14b8de118e5eda5e8d1c3a20631e3ea6eff4fd84.
I adopted this style from the network adapter in the consensus service. If my fix looks good, I will update the network adapter in consensus as well in the next PR, to have the same style over our codebase.

nomos-services/cryptarchia-consensus/src/mix/adapters/libp2p.rs
youngjoon-lee: Fixed in 14b8de118e5eda5e8d1c3a20631e3ea6eff4fd84

nomos-services/cryptarchia-consensus/src/mix/adapters/mock.rs
youngjoon-lee: Fixed in 14b8de118e5eda5e8d1c3a20631e3ea6eff4fd84

In consensus service, no. We're using the libp2p adapter.

This mock adapter is for [DA integration test](https://github.com/logos-co/nomos-node/pull/824/files#diff-9d5cf4b9aae836e92a04ce9c9f4f1aec0b317470c26fa25907ac680a95a17bfa). I decided to use this mock because of [the comment](https://github.com/logos-co/nomos-node/pull/822/files#diff-80df468bf4cbe900aecbf5af48340caef3d57e1de0387d3edef75dad46fd4743L1) (which I removed in #822), but now I think it may be okay to use the real libp2p adapter for DA integration test, because it is already using the real libp2p adapter for the network service. Let me try it.

youngjoon-lee:
> A few things to polish. Btw, now we have the executor node binary as well, mix service will need to be added to it as well here. It shouldn't be too much more.

Oh, right. Thank you. I added it to the nomos-executor as well: 8150151fbbb45db2b470e9c15689790df6468be0

nomos-services/cryptarchia-consensus/src/lib.rs
youngjoon-lee: Okay, I did a sync with @madxor . I'll restructure the services like below:
- Consensus service listens messages from the network service (the broadcast network, in a term familiar to Marcin).
- Consensus service sends messages to the mix service.
- Mix service wraps the messages and propagate them through the mix network.
**- Mix service sends fully-unwrapped messages to the network service.**

The bold one is what @danielSanchezQ is pointing out in this thread, if I understand him correctly. That's right. I'll update the implementation accordingly.

danielSanchezQ: Yeah, this is exactly the flow I meant. I would go for it!

bacv: If I get it correctly, in case of other services using mixnet for other types of messages would use different kind of mixnet adapter? This would allow to have one mixnet service and multiple adapters that would take care of filtering and deserialization.

In this case MixNetworkAdapter is already defined in the consensus module, so I assume it will be only used for consensus, thus could handle the deserialization as well.

danielSanchezQ: > If I get it correctly, in case of other services using mixnet for other types of messages would use different kind of mixnet adapter? This would allow to have one mixnet service and multiple adapters that would take care of filtering and deserialization.
> 
> In this case MixNetworkAdapter is already defined in the consensus module, so I assume it will be only used for consensus, thus could handle the deserialization as well.

yes it would be better to handle that in the adapter itself. But in this case consensus just writes/sends to mixnet, but it does not read to it. So instead of deserialization the adapter would probably just serialize.

youngjoon-lee: Yeah.. At the beginning, I imagines the case which uses the mix service but not the network service. That's why I clearly decoupled the mix service from the network service. But, if that is not the case that we should consider, I can couple the mix service and the network service..

bacv: > Mix service sends fully-unwrapped messages to the network service.

I still think this should be handled in the consensus mixnet adapter, as it is consensus related - the block needs to be broadcast in a specific topic, that libp2p network adapter knows about. Mixnet adapter in consensus could require a libp2p network adapter when initializing it.

> But, if that is not the case that we should consider, I can couple the mix service and the network service

We could still keep services uncoupled and agnostic to the details that consensus requires when two adapters in consensus?

youngjoon-lee: > the block needs to be broadcast in a specific topic, that libp2p network adapter knows about. Mixnet adapter in consensus could require a libp2p network adapter when initializing it.

Ah, true. That's what the mix service doesn't know. I think you're right. Let me try to implement it and check which way is the most reasonable.

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -446,6 +469,17 @@ where
                         .await;
                     }
 
+                    Some(mixed_msg) = incoming_mixed_msgs.next() => {
+                        match wire::deserialize::<Block<ClPool::Item, DaPool::Item>>(&mixed_msg) {
+                            Ok(block) => {
+                                adapter.broadcast(NetworkMessage::Block(block)).await;
```
bacv, line 475: Nit: Now that we have two different networks, this `adapter` could use a more descriptive name.

nomos-services/cryptarchia-consensus/src/mix/adapters/mock.rs
youngjoon-lee: Just moved the mock mix backend. We don't need that. Now we use the real mix backend for DA integration tests
b48ce841c7707c9d3f6756a27bbb44f48d3ccb04

nomos-services/cryptarchia-consensus/src/mix/adapters/libp2p.rs
danielSanchezQ: > True. Fixed in [14b8de1](https://github.com/logos-co/nomos-node/commit/14b8de118e5eda5e8d1c3a20631e3ea6eff4fd84). I adopted this style from the network adapter in the consensus service. But yes, we don't need to use the `tokio::sync::broadcast` channel redundantly here in the adapter level because the mix network backend is already using the broadcast channel. If my fix looks good, I will update the network adapter in consensus as well in the next PR, to have the same style over our codebase.

yep, this is it, much better like this.

nomos-services/cryptarchia-consensus/src/lib.rs
danielSanchezQ: > > Mix service sends fully-unwrapped messages to the network service.
> 
> I still think this should be handled in the consensus mixnet adapter, as it is consensus related - the block needs to be broadcast in a specific topic, that libp2p network adapter knows about. Mixnet adapter in consensus could require a libp2p network adapter when initializing it.
> 
> > But, if that is not the case that we should consider, I can couple the mix service and the network service
> 
> We could still keep services uncoupled and agnostic to the details that consensus requires when two adapters in consensus?

I understand this. But take into consideration that we would be using consensus as a bridge for networking **when is not needed**. For consensus just 2 things matter:
* Receive blocks 
* Sends blocks

It should not bridge two network services just because it understands them. So either we merge the two adapter into a single one or this is dealt in mixnet-service level. Probably the key question here is *is the broadcasting part of the mixnet protocol?* if yes, it can be handle there, if not in the adapter. But consensus should just have to calls to adapter to send and receive **full decoded blocks**.

youngjoon-lee: > Probably the key question here is is the broadcasting part of the mixnet protocol?

It just has been agreed to as YES, in the spec.

youngjoon-lee: Resolved by #829

nomos-services/cryptarchia-consensus/src/lib.rs
youngjoon-lee: Good point! Fixed in 9ab301a92cb2b4f38e4b06f935a98d0b8c426075

danielSanchezQ (approved):
Excellent! Another stone to the wall!

bacv (approved):

PR #825, opened 2024-10-14
Fix: Sampling protocol test error

TakeASwing-420:
@danielSanchezQ, The current version of the testing function is incorrect... Whenever one of the peer(in this case we have two peers) has successfully received 10 samples, it breaks out of the loop and closes the stream. This causes a sampling error for the other peer which ultimately results in an infinite loop for that peer.

I have
- Inserted two labels for the two async runtime threads `leading` and `trailing`.
- Created two separate flags `done1` and `done2` for the two tasks. If both are _true_, then I am exiting the loop.
- Refactored some variable  by using `Cow::borrowed` to mitigate unnecessary copying of a _peerId_.

Please check into this.

### DIFF
--- a/compose.static.yml
+++ b/compose.static.yml
@@ -74,7 +74,7 @@ services:
     ports:
       - "3003:3000/udp"
       - "18083:18080/tcp"
-    entrypoint: /etc/nomos/scripts/run_nomos_node.sh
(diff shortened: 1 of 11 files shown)


danielSanchezQ:
I think something went wrong with this PR, it has 0 changes.

TakeASwing-420:
resubmitted this PR in #840

PR #837, opened 2024-10-21
[CL-2] Move POL into nomos-core

zeegomo:

Move PoL proof definition from ledger to nomos-core to uniform with the rest of the codebase. Also changes proof type from STARK to groth16. This currently requires an x86 machine with docker support (or a cloud provider), but it allows to reduce proof sizes consistenly (from 200kb to a few hundred bytes).

Yes, mainly moved somewhere else.

No spec, myself.

Not applicable
Not applicable
## Checklist

* [x] 1. Description added.


nomos-core/chain-defs/src/header/leader_proof.rs
```diff
@@ -0,0 +1,180 @@
...
+            .write(&public_inputs)
+            .unwrap()
+            .write(&private_inputs)
+            .unwrap()
+            .build()
+            .unwrap();
+
+        let start_t = std::time::Instant::now();
+
+        // ATTENTION: producing a groth16 proof currently requires x86 with docker support
```
danielSanchezQ, line 37: They use virtualization for this? I'm really curious why is the case. 🤔

zeegomo: Not sure why, I guess it's just to bundle dependencies for rapidsnark which is what they run inside the container

danielSanchezQ:
Regarding the docker stuff. Should we wait till we have the CI ready for this? 
@bacv & @zeegomo ?

bacv:
> Regarding the docker stuff. Should we wait till we have the CI ready for this? @bacv & @zeegomo ?

GH runners support running docker, we have docker-build job that checks if nomos image can be built. Or we could use the selfhosted linux runner with custom config tailored to this particular case.

danielSanchezQ (approved):

zeegomo:
> Regarding the docker stuff. Should we wait till we have the CI ready for this?

AFAIK we were using dev mode before as well because it was too slow on GH runners. If we want to build proofs I think the only solution is self-hosted runners

davidrusu (approved):
🚢

PR #840, opened 2024-10-22
test fix

TakeASwing-420:


TakeASwing-420:
@danielSanchezQ, The current version of the testing function is incorrect... Whenever one of the peer(in this case we have two peers) has successfully received 10 samples, it breaks out of the loop and closes the stream. This causes a sampling error for the other peer which ultimately results in an infinite loop for that peer.

I have
- Inserted two labels for the two async runtime threads `leading` and `trailing`.
- Created two separate flags `done1` and `done2` for the two tasks. If both are _true_, then I am exiting the loop.
- Refactored some variable  by using `Cow::borrowed` to mitigate unnecessary copying of a _peerId_.

Please check into this.

danielSanchezQ:
Although its a good catch. I have some concerns:
* It it was stuck, it would finish the tests, but this has been running in CI for some time now 🤔 
* I see value in updating so both of them keep running until the constrains are completed. But I do not think this is the way, it is a bit convoluted in my opinion. We could just signal and wait for one another, or just use a shared counter or something like that probably.

TakeASwing-420:
> * I see value in updating so both of them keep running until the constrains are completed. But I do not think this is the way, it is a bit convoluted in my opinion. We could just signal and wait for one another, or just use a shared counter or something like that probably.

Can you please elaborate 😅...I am not getting the point...

danielSanchezQ:
> > * I see value in updating so both of them keep running until the constrains are completed. But I do not think this is the way, it is a bit convoluted in my opinion. We could just signal and wait for one another, or just use a shared counter or something like that probably.
> 
> Can you please elaborate 😅...I am not getting the point...

Two things. 

One question:
How is it working on CI if it gets stuck:

Comment on the approach. 
I do not think we need to add extra layers for testing. 
* For example no need to use `Cow` when we can clone it (it is just a simple test)
* Why having a `"leading"` and `"trailing"` and why are they `Strings`? This can be solved without any special distinction.
* The atomics seem annoying in this case. It could be just `oneshot` channels probably. 

But the most importante is to see if it is really not working.


I know it has been open some time, but time bandwidth is short atm.

TakeASwing-420:
>One question: 
How is it working on CI if it gets stuck:

Have you set any time limit... ngl Devops I don't know that much, lol 🙌
>Comment on the approach.
I do not think we need to add extra layers for testing.
1.  Okay I will remove `Cow`.
2.  I have inserted `leading` (for the first one as it is indeed leading by 1 second) and `trailing` (for the second one) to identify which peer has finished sampling... otherwise I couldn't identify. I actuallly do not have async coding experience..so I tried to do it 
naively but I do think my style of approach might not have fully aligned with the overall workflow of this project.
4. Do you want to fire a message to the other peer using oneshot channel when the sender has received the targeted amount (in this case 10) of samples and then the receiver will update it's respective flag? Okay I will do it...

TakeASwing-420:
@danielSanchezQ , I have a better proposal...
We can do a kind of `3-way TLS handshake` thing here...
1. let's say peer1 has first completed sampling..it will send a random sequence to peer2
2. peer2 will receive the sequence and echo it and after that it will break out the loop with the res vector.
3. peer1 will receive the echoed sequence and will break out of the loop with it's res vector.

TakeASwing-420:
@danielSanchezQ, I’m encountering difficulties in creating a separate message channel for handling the completion of the sampling task. I have experimented with various methods, including mpsc and oneshot, but _managing two simultaneous asynchronous tasks—one for sampling and another for sending completion messages_—has proven challenging at runtime.

In the code below, I thought of prioritizing the evaluation of completion messages before proceeding with the rest of the logic. But this is not working.
```rust

        async fn test_sampling_swarm(
            mut swarm: Swarm<
                SamplingBehaviour<
                    impl MembershipHandler<Id = PeerId, NetworkId = SubnetworkId> + 'static,
                >,
            >,
            sender: mpsc::Sender<&str>,
            mut receiver: mpsc::Receiver<&str>,
        ) -> Vec<[u8; 32]> {
            let mut res = Vec::with_capacity(MSG_COUNT);
            loop {
                // Evaluate the receipt first
                if let Some(msg) = receiver.recv().await {
                    match msg {
                        "Finished Sampling!" => {
                            if res.len() == MSG_COUNT {
                                sender
                                    .send("Finished Sampling! Closing Connection...")
                                    .await
                                    .unwrap();
                                drop(sender);
                                break res;
                            }
                        }
                        "Finished Sampling! Closing Connection..." => {
                            if res.len() == MSG_COUNT {
                                drop(sender);
                                break res;
                            }
                        }
                        _ => {}
                    }
                } else {
                    // If receiver is None, check if we have enough results
                    if res.len() == MSG_COUNT {
                        sender.send("Finished Sampling!").await.unwrap();
                    }
                }

                // Handle the swarm events
                match swarm.next().await {
                    None => {}
                    Some(SwarmEvent::Behaviour(SamplingEvent::IncomingSample {
                        request_receiver,
                        response_sender,
                    })) => {
                        debug!("Received request");
                        // spawn here because otherwise we block polling
                        tokio::spawn(request_receiver);
                        response_sender
                            .send(BehaviourSampleRes::SamplingSuccess {
                                blob_id: Default::default(),
                                subnetwork_id: Default::default(),
                                blob: Box::new(DaBlob {
                                    column: Column(vec![]),
                                    column_idx: 0,
                                    column_commitment: Default::default(),
                                    aggregated_column_commitment: Default::default(),
                                    aggregated_column_proof: Default::default(),
                                    rows_commitments: vec![],
                                    rows_proofs: vec![],
                                }),
                            })
                            .unwrap()
                    }
                    Some(SwarmEvent::Behaviour(SamplingEvent::SamplingSuccess {
                        blob_id, ..
                    })) => {
                        debug!("Received response");
                        res.push(blob_id);
                    }
                    Some(SwarmEvent::Behaviour(SamplingEvent::SamplingError { error })) => {
                        debug!("Error during sampling: {error}");
                    }
                    Some(event) => {
                        debug!("{event:?}");
                    }
                }
            }
        }
```
 However, I do have reverted the **Cow** variables and renamed the **labels** for convenience.

PR #843, opened 2024-10-22
Mix: Offload transmission rate and message processing from libp2p behaviour/handler

youngjoon-lee:

This PR is based on the PR #841.

This PR offloads transmission rate and message processing (wrap/unwrap/delay) from the libp2p behaviour/handler level. Instead, those are going to be handled in the `MixService` level.

Why?
1. We've updated the spec. Now we don't have connection-level mixing. Now the sequence of messages emitted over each connection doesn't need to be different from each other anymore. So, we're going to have one message processor that emits messages to all connections in the promised rate. The updated spec says [this](https://www.notion.so/Nomos-Blend-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#11c8f96fb65c80ac8bc9e4bf1b201f60). The term `Internal Message` in the spec may be confusing, but please see the additional explanation (with diagram) below.
image
2. Core protocol doesn't need to be implemented in the libp2p level.

Additional explanation:
The green part in the [diagram](https://www.notion.so/Nomos-Blend-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#11c8f96fb65c80ceb3c3ddaae55d1224) is for the persistent transmission and the message processing. Previously, most of that part was implemented in the libp2p custom behaviour and connection handler, because we needed per-connection processing. However, now we don't need it. So, the green part is moved to the `MixService`.

NOMIX-Architecture-PTM(1)

> [!NOTE] 
> This PR leaves the implementation of the green part above (persistent transmission and message processing) as TODO. Upcoming PRs will implement it. Currently this PR just has mock. 

Now, the libp2p custom behaviour in the `nomos-mix` is just for sending messages to all connected peers and receiving messages from them. It's similar as gossipsub. But, I still use it because of the reasons described in the PR #841 description:
> This forwarding is similar as libp2p gossipsub, but I'm keeping the custom Behaviour because of two reasons:
>
> - To introduce the connection maintenance mechanism by monitoring emission rates soon
> - To avoid the complexity caused by the control messages (IWANT, ...) of libp2p gossip. They haven't been analyzed from perspective of privacy.

Yes

@madxor @youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [x] 1. Description added.


madxor:
I have only "naming" comment. Is the "Internal Message" clear enough from the development perspective? For me it's a bit unclear. Looking at the types of messages that we have:
1. Inbound message is a message that is incoming from outside of the node, which is at the same time relayed forward and moved to tier 2.
2. Outbound message is a new message that is received from Tier 2 that is going to be processed by Tier 1 and released to the network.
(Alternatively we can use ingress and egress names)

In the above context, wouldn't "Internal Message" be a bit confusing as it could be "naturally" understood as both?

Moreover, we can use the same naming convention for other tiers. So in Tier 2 the inbound would be a message coming from Tier 1 and outbound message would be received from Tier 3, and so on... But I don't know if this is something we aim for.

Just my 1 cent ;)

nomos-mix/network/src/behaviour.rs
```diff
@@ -58,33 +59,41 @@ impl Behaviour {
...
             .negotiated_peers
             .keys()
             .filter(|&peer_id| {
-                if let Some(propagation_source) = propagation_source {
-                    *peer_id != propagation_source
+                if let Some(excluded_peer) = excluded_peer {
+                    *peer_id != excluded_peer
                 } else {
                     true
                 }
```
danielSanchezQ, line 99: Since we collect. We could collect into a hashset directly. Then just remove the excluded peer.

```suggestion
let mut peer_ids: HashSet<_> = self
            .negotiated_peers
            .keys()
            .collect();
 if let Some(peer) = excluded_peer {
     peer_ids.remove(peer);
 }
```

danielSanchezQ (approved):
This makes me wonder if we would need the behaviour or if we went the extra mile for nothing. Isn't just doing flooding atm? Maybe we could have just wrapped the flooding protocol. But as is working now it doesn't really matter anymore.

youngjoon-lee:
> This makes me wonder if we would need the behaviour or if we went the extra mile for nothing. Isn't just doing flooding atm? Maybe we could have just wrapped the flooding protocol. But as is working now it doesn't really matter anymore.

Yes. Our current implementation is almost the same as floodsub. I didn't check the libp2p floodsub implementation details yet, but yeah, ours is pretty much the same as floodsub. This is a concern I also had and I mentioned about this in the PR description. Only one thing I didn't implement yet is "[connection maintenance](https://www.notion.so/Nomos-Blend-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#11c8f96fb65c806db454e80add388ab7)". I need to monitor the traffic rate of each connection, close connections if their rate is far from the expected rate, and reconnect them or establish new connections with new peers. At this moment, I'm not sure which way is the best to implement this. We may be able to implement this even while using floodsub. Or, we may need our custom behaviour/handler. I didn't try it yet, but I'll do it at the next PR. Until then, I want to keep our custom behaviour/handler.

youngjoon-lee:
> I have only "naming" comment. Is the "Internal Message" clear enough from the development perspective? For me it's a bit unclear. Looking at the types of messages that we have:
> 
>     1. Inbound message is a message that is incoming from outside of the node, which is at the same time relayed forward and moved to tier 2.
> 
>     2. Outbound message is a new message that is received from Tier 2 that is going to be processed by Tier 1 and released to the network.
>        (Alternatively we can use ingress and egress names
> 
> 
> In the above context, wouldn't "Internal Message" be a bit confusing as it could be "naturally" understood as both?
> 
> Moreover, we can use the same naming convention for other tiers. So in Tier 2 the inbound would be a message coming from Tier 1 and outbound message would be received from Tier 3, and so on... But I don't know if this is something we aim for.
> 
> Just my 1 cent ;)

@madxor I liked the naming `Input` and `Output` in Tier level, as we discussed in the [thread](https://www.notion.so/Nomos-Blend-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?d=11f8f96fb65c805e8f32001c813d9658&pvs=4#11c8f96fb65c80ceb3c3ddaae55d1224) in Notion. Let's use those names. `Inbound` and `Outbound` are also good, but I think input/output is better.
And, yes, we can use the name `Input` and `Output` for Tier 2 as well. The input of Tier 2 comes from Tier 1. The output of Tier 2 goes to either Tier 1 or the Broadcast network. 
Also, I think we should redefine the input/output of Tier 1. The input of Tier 1 comes from either Tier 2 or network connnections. The output of Tier 1 comes from either Tier 2 or network connection. This way is better to clearly abstract the Tier 1 and decouple it from the network layer.
The implementation of Tier 1 and 2 can be refactored accordingly

If you're fine with this, I can try to update the [diagram](https://www.notion.so/Nomos-Blend-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#11c8f96fb65c80ceb3c3ddaae55d1224) in the spec. To be honest, I don't think we need to draw "Input Queue" and "Output Queue" in the diagram. We can just simply remove them. Just arrows are fine in my opinion. Actually, I think we don't need the explicit names: input/output, in the spec level. Let me try to update the diagram and show you. Then, we can discuss.

youngjoon-lee:
> I have only "naming" comment. Is the "Internal Message" clear enough from the development perspective? For me it's a bit unclear. Looking at the types of messages that we have:

@madxor Let's move this discussion to Notion. I'll open a thread and ping you there

nomos-mix/network/src/behaviour.rs
youngjoon-lee: Much cleaner. Thank you! 01503d7d29cbe32ea790951ab0c8dfc40a202ac0

youngjoon-lee:
I'm merging this PR because I had a call with Marcin regarding the [naming topic](https://github.com/logos-co/nomos-node/pull/843#issuecomment-2434292472), and we made a conclusion. Other PRs will be slightly updated accordingly, but this PR is ready to be merged.

PR #844, opened 2024-10-22
Mix: Core skeleton used in `MixService`

youngjoon-lee:

This PR is based on the PR #843.

- Implements skeleton and mock for core protocol:
  - [Persistent Transmission](https://www.notion.so/Nomos-Blend-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#11c8f96fb65c80f2bc36e6b0b1f27a9d) (Tier 1)
  - [Cryptographic Processor](https://www.notion.so/Nomos-Blend-Network-Tier-2-Message-Blending-Module-1208f96fb65c80e5bcb9df6e27472339?pvs=4#1208f96fb65c80f8a8d3d2b449953bde), [Temporal Processor](https://www.notion.so/Nomos-Blend-Network-Tier-2-Message-Blending-Module-1208f96fb65c80e5bcb9df6e27472339?pvs=4#1208f96fb65c80dca885dda33fbd599b), and [Message Router](https://www.notion.so/Nomos-Blend-Network-Tier-2-Message-Blending-Module-1208f96fb65c80e5bcb9df6e27472339?pvs=4#1208f96fb65c80de8873e5db0f420caf) (Tier 2)
- Uses them in `MixService`

Background:
This core protocol has been offloaded from the libp2p level in the PR #843, so that it can be used in `MixService`.

Yes

@madxor @youngjoon-lee 

Yes

Here, I defined the parameter `max_transmission_frequency` instead of CEF (Consensus Emission Frequency) defined in the spec. I'd like to suggest to rename CEF to not use the term "consensus" in the mix protocol. I opened a [thread](https://www.notion.so/Nomos-Blend-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?d=1278f96fb65c80c7a9bf001cb9309424&pvs=4#11f8f96fb65c80dfbd23e4400feaaf9c) in Notion for this discussion.

## Checklist

> [!WARNING]  

* [ ] 1. Description added.


madxor:
We have change the CEF to Maximal Emission Frequency in the spec.

nomos-mix/core/src/processor/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// [`Processor`] implements "Message Router" defined in the Tier-2 spec.
+/// - Wraps new messages using [`CryptographicProcessor`]
+/// - Unwraps incoming messages received from network using [`CryptographicProcessor`]
+/// - Pushes unwrapped messages to [`TemporalProcessor`] to delay them
+/// - Releases messages returned by [`TemporalProcessor`]
+pub struct Processor {
+    /// To receive new messages originated from this node
+    new_message_receiver: mpsc::UnboundedReceiver<Vec<u8>>,
+    /// To receive incoming messages from received the network
```
madxor, line 26: "To receive incoming messages from the network"?

nomos-mix/core/src/processor/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub cryptographic_processor: CryptographicProcessorSettings,
+    pub temporal_processor: TemporalProcessorSettings,
+}
+
+/// [`Processor`] implements "Message Router" defined in the Tier-2 spec.
+/// - Wraps new messages using [`CryptographicProcessor`]
+/// - Unwraps incoming messages received from network using [`CryptographicProcessor`]
+/// - Pushes unwrapped messages to [`TemporalProcessor`] to delay them
+/// - Releases messages returned by [`TemporalProcessor`]
+pub struct Processor {
```
madxor, line 23: Not claiming that it is a correct way but have you considered having only `inbound_message_processor` and `outbound_message_processor` and construct a message as `{payload, routing_data}`, where `routing_data` defines the context on how the message should be processed?

nomos-mix/core/src/processor/temporal.rs
```diff
@@ -0,0 +1,16 @@
...
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct TemporalProcessorSettings {
+    pub max_delay_seconds: u64,
+}
+
+impl Default for TemporalProcessorSettings {
+    fn default() -> Self {
+        Self {
+            max_delay_seconds: 10,
```
madxor, line 11: In the future a configuration parameter, right?

nomos-services/data-availability/tests/src/common.rs
```diff
@@ -1,5 +1,8 @@
 // std
 use nomos_da_network_service::backends::libp2p::common::DaNetworkBackendSettings;
+use nomos_mix::processor::{
+    CryptographicProcessorSettings, ProcessorSettings, TemporalProcessorSettings,
```
madxor, line 4: Nit picking. This is a bit confusing, wouldn't it be a bit more clear if we user "RoutingProcessor" instead of the "Processor"? Then instead of "ProcessorSettings" we would have "RoutingProcessorSettings"?

nomos-mix/core/src/processor/mod.rs
madxor: I think I have misunderstood this part. Can you confirm that you are making the Message Router a "governor" or a primary entity that encapsulates cryptographic and temporal processors? It is a bit different than I have imagined. My vision was that the Message Router will be "called" by the Cryptographic Processor, and the CP is called by Tier 1 or Tier 3. I was assuming that neither CP or MR (or even TP) are going to have any clocks that determine their work, they should just process messages when needed. We are going to have clocks at Tier 1 and Tier 3.

Regarding TP: the implementation of the temporal processor might not physically add any delay but add a release time to the message metadata and pass it to the PTM. The PTM would read messages from the input queue when the release time is met.

nomos-mix/core/src/processor/temporal.rs
youngjoon-lee: This is already a configuration parameter. This code just defines the default value that can be easily used in some tests. But, yeah, actually, I don't think it's a good practice to define default values. Instead, I can let users configure this value always manually. I will remove these defaults.

nomos-mix/core/src/processor/mod.rs
youngjoon-lee: > I think I have misunderstood this part. Can you confirm that you are making the Message Router a "governor" or a primary entity that encapsulates cryptographic and temporal processors?

Yes. I'll rename the `Processor` to `MessageBlender`, which is the name of Tier 2 defined in the spec. And, I'll explicitly define the `MessageRouter`. The `MessageBlender` will use it. 

>  I was assuming that neither CP or MR (or even TP) are going to have any clocks that determine their work, they should just process messages when needed. We are going to have clocks at Tier 1 and Tier 3.

Yes. CP doesn't have clocks in this implementation. Also, the part that does the same things as MR also doesn't have clocks. But, as I mentioned above, I will refactor this module to make it look more identical to the spec.

> Regarding TP: the implementation of the temporal processor might not physically add any delay but add a release time to the message metadata and pass it to the PTM. The PTM would read messages from the input queue when the release time is met.

Yes. The code comment here is just for explaining the role of TP shortly. If you think this comment is not correct, I will remove it (the `to delay them` part).

nomos-services/data-availability/tests/src/common.rs
youngjoon-lee: Answered in https://github.com/logos-co/nomos-node/pull/844#discussion_r1813079157

nomos-mix/core/src/processor/mod.rs
youngjoon-lee: I renamed `Processor` to `MessageBlend`. It's much clearer. 

But still, I'm not defining the `MessageRouter` module in this code. The role of `MessageRouter` is very very simple. The `MessageBlend` can perform the role without any overhead. The code looks simple and easy. Even lighter in term of computations.
If we define the `MessageRouter` explicitly in the code, we'll end up adding more async channels and data structures to make proper orders to `MessageRouter`. I think it will make code verbose.
As long as this code implements what `MessageRouter` should do correctly, I don't think the implementation must follow the structure defined in the spec 100% the same.

nomos-services/data-availability/tests/src/common.rs
youngjoon-lee: Renamed to `MessageBlendSettings`: 7091a5ee2210c45d8674f8a3b76bd005440f4e3f

nomos-mix/core/src/processor/mod.rs
youngjoon-lee: Thanks. It's a leftover. Fixed in 7091a5ee2210c45d8674f8a3b76bd005440f4e3f

nomos-mix/core/src/processor/temporal.rs
youngjoon-lee: Removed default values: 7091a5ee2210c45d8674f8a3b76bd005440f4e3f

nomos-mix/core/src/message_blend/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// To receive incoming messages from the network
+    inbound_message_receiver: mpsc::UnboundedReceiver<Vec<u8>>,
+    /// To release messages that are successfully processed but still wrapped
+    outbound_message_sender: mpsc::UnboundedSender<Vec<u8>>,
+    /// To release fully unwrapped messages
+    fully_unwrapped_message_sender: mpsc::UnboundedSender<Vec<u8>>,
+    /// Processor
+    cryptographic_processor: CryptographicProcessor,
+    // TODO: Add TemporalProcessor
+}
```
danielSanchezQ, line 35: Nitpick: Probably the most elegant way of doing this would be to implement `Sink` and `Stream` instead of using channels. This is an idea that would have to be specified but I Imagine the distinct tiers as stream operation that could be mapped maybe 🤔 

Again, nothing incorrect with this, is straight forward. But if at some point you want to explore into rust black magic we could do it.

nomos-services/mix/src/lib.rs
```diff
@@ -77,12 +81,41 @@ where
...
+            mpsc::unbounded_channel();
+        let (emission_sender, mut emission_receiver) = mpsc::unbounded_channel();
+        tokio::spawn(async move {
+            persistent_transmission(
+                mix_config.persistent_transmission,
+                transmission_schedule_receiver,
+                emission_sender,
+            )
+            .await;
+        });
```
danielSanchezQ, line 100: Yep, this is another signal that it could probably be implemented as stream operations. Lets keep this in mind for a further refactor. Maybe open an issue for it.

nomos-services/mix/src/lib.rs
```diff
@@ -77,12 +81,41 @@ where
...
+                mix_config.message_blend,
+                new_message_receiver,
+                processor_inbound_receiver,
+                // Connect the outputs of Message Blend to Persistent Transmission
+                transmission_schedule_sender,
+                fully_unwrapped_message_sender,
+            )
+            .run()
+            .await;
+        });
```
danielSanchezQ, line 118: Third possible confirmation this doesn't look the best way. (But again, I'm not sure if my suggestion would be better for sure).

nomos-mix/core/src/message_blend/mod.rs
youngjoon-lee: True. Actually, I tried to implement `Sink` and `Stream` at the beginning, but turned into this way because of some small issues. But now, I think it's possible (as I did in https://github.com/logos-co/nomos-node/pull/846). Let me try in a new PR. There might be some small issues to discuss, but I agree that it's better to implement Sink and Stream instead of using channels in the library level.

youngjoon-lee: This refactoring will be done as a next PR. First of all, we're going to have a [discussion](https://github.com/logos-co/nomos-node/pull/847#pullrequestreview-2394511285) to make things simpler.

nomos-services/mix/src/lib.rs
youngjoon-lee: https://github.com/logos-co/nomos-node/pull/844#discussion_r1816233929

nomos-services/mix/src/lib.rs
youngjoon-lee: https://github.com/logos-co/nomos-node/pull/844#discussion_r1816233929

danielSanchezQ (approved):
Improvements will be made later on.

PR #845, opened 2024-10-22
Mix: Implement Persistent Transmission (Tier 1)

youngjoon-lee:

This PR is based on the PR #844.

This PR implements the actual [Persistent Transmission](https://www.notion.so/Nomos-Blend-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#11c8f96fb65c804db7ccfd024f8c44d0) (Tier 1). The protocol details can be found in the link, but I explained the core logic as code comments.

For your background,
Messages that were processed in Tier 2 are gathered in the Persistent Transmission module. They are periodically and sequentially sent to all connected peers. If no message is scheduled, a drop message is sent with a probability.

Yes

@madxor @youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 1. Description added.

### DIFF
--- a/nomos-mix/core/Cargo.toml
+++ b/nomos-mix/core/Cargo.toml
@@ -7,7 +7,8 @@ edition = "2021"
 cached = "0.53"
 tokio = { version = "1.39" }
 tracing = "0.1"
-rand = { version = "0.8", features = ["small_rng"] }
+rand = "0.8"
 serde = { version = "1.0", features = ["derive"] }
 nomos-mix-message = { path = "../message" }
 futures = "0.3"
+rand_chacha = "0.3"
--- a/nomos-mix/core/Cargo.toml
+++ b/nomos-mix/core/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2021"
 cached = "0.53"
 tokio = { version = "1.39" }
 tracing = "0.1"
-rand = "0.8"
+rand = { version = "0.8", features = ["small_rng"] }
 serde = { version = "1.0", features = ["derive"] }
 nomos-mix-message = { path = "../message" }
 futures = "0.3"
--- a/nomos-mix/core/src/persistent_transmission.rs
+++ b/nomos-mix/core/src/persistent_transmission.rs
@@ -1,7 +1,7 @@
 use std::time::Duration;
 
 use nomos_mix_message::DROP_MESSAGE;
-use rand::Rng;
+use rand::{distributions::Uniform, prelude::Distribution, rngs::SmallRng, Rng, SeedableRng};
 use serde::{Deserialize, Serialize};
 use tokio::{
     sync::mpsc::{self, error::TryRecvError},
@@ -41,6 +41,8 @@ pub async fn persistent_transmission(
     let mut interval = time::interval(Duration::from_secs_f64(
         1.0 / settings.max_emission_frequency,
     ));
+    let mut coin =
+        Coin::<_>::new(SmallRng::from_entropy(), settings.drop_message_probability).unwrap();
 
     loop {
         interval.tick().await;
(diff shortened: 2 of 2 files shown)


nomos-mix/core/src/persistent_transmission.rs
```diff
@@ -26,13 +33,45 @@ impl Default for PersistentTransmissionSettings {
...
+    let mut interval = time::interval(Duration::from_secs_f64(
+        1.0 / settings.max_emission_frequency,
+    ));
+
+    loop {
+        interval.tick().await;
+
+        // Emit the first one of the scheduled messages.
+        // If there is no scheduled message, emit a drop message with probability.
+        match schedule_receiver.try_recv() {
```
madxor, line 50: Receiving should be independent from the interval. We cannot read messages according to ticks, we must only calculate the rate at which they are arriving but receive them instantly.

youngjoon-lee: Yes. The "schedule" channel is corresponding to the "Input Queue" in the spec. Messages arrive in the schedule channel from other modules (Tier 2, currently). 

This code (`schedule_receiver.try_recv()`) is for reading the schedule channel (Input Queue), as defined in the [spec](https://www.notion.so/Nomos-Blend-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#11c8f96fb65c808cb008d2595105c32b). 
> Message is read from the queue at the CEF frequency.

The term `receiver` might be little confusing to me. That's the naming convention that represents the "receiver end" of async channel in Rust.

danielSanchezQ: This should be rewrite using tokio::select probably. No need to wait for receiving as marcin says. If we received they are stored. Otherwise on interval, pop messages or send DROP ones.

youngjoon-lee: I think the current implementation is correct.

Messages are stored in the schedule channel, as soon as other modules can them to the schedule channel (using the sender half: `schedule_sender`). This channel is a mpsc unbound channel.

In this code, we sleep for an interval and "try" to "pop" one message from the schedule channel. This try is non-blocking. If the channel is empty, send a drop message.

I think the code implements the spec correctly. Please let me know if I'm missing anything. BTW, it's my bad that I didn't write tests. I'll do it today.

youngjoon-lee: 494cf76dce3c21e0aa8d4b9db0a3687247284f4b I just added a test. You can see that multiple messages can be scheduled without blocking. Those messages are emitted with the expected interval. If there is no message scheduled, drop messages are emitted with the expected interval as well.

nomos-mix/core/src/persistent_transmission.rs
```diff
@@ -26,13 +33,129 @@ impl Default for PersistentTransmissionSettings {
...
+            Err(TryRecvError::Disconnected) => {
+                tracing::error!("The schedule channel has been closed");
+                break;
+            }
+        }
+    }
+}
+
+fn coin_flip(probability: f64) -> bool {
+    rand::thread_rng().gen::<f64>() < probability
```
danielSanchezQ, line 76: Is the value here normalized?

youngjoon-lee: Yes. It returns the value in [0, 1), but the [doc] doesn't say that explicitly. So, I just considered using [gen_range](https://docs.rs/rand/latest/rand/trait.Rng.html#method.gen_range) explicitly, but found that it is not optimized for repeated sampling (as written in its doc). So, I ended up deciding to using the [Uniform](https://docs.rs/rand/latest/rand/distributions/struct.Uniform.html) distribution by defining the `Coin` struct. 633b21fd16ebee040efae20c64455b925c1ed8cc
I think this is cleaner, more explicit, and more efficient. Thank you.

nomos-mix/core/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2021"
 cached = "0.53"
 tokio = { version = "1.39" }
 tracing = "0.1"
-rand = "0.8"
+rand = { version = "0.8", features = ["small_rng"] }
```
danielSanchezQ, line 10: Curious, why small? Probably rand_chacha should be quick and secure enough instead?

danielSanchezQ (approved):

nomos-mix/core/Cargo.toml
danielSanchezQ: Yeah, lets do the middle term. It should be easy to change if needed. Thanks!

PR #850, opened 2024-10-23
[CL-1] Add proofs to CL txs

zeegomo:

Complete bundle definition from https://github.com/logos-co/nomos-node/pull/835 by adding missing proofs to ensure transaction validity. 
Also refactor `InputWitness` to include the merkle path to the commitment root.

Main reference is the Coordination Layer data model, for which you can find the spec in [Notion](https://www.notion.so/Coordination-Layer-Data-Model-Specification-9a60bd41167d426abb0cd07a01aab905)

@davidrusu @alvatar and myself

Yes, it has been reviewed after the implementation work and discrepancies have been resolved.

Any changes have been already reflected in the spec.

## Checklist

* [X] 1. Description added.


nomos-core/chain-defs/src/tx/bundle.rs
```diff
@@ -2,55 +2,65 @@
...
-use crate::wire;
-use nomos_proof_statements::bundle::{BundlePrivate, BundlePublic};
-
+use crate::{
+    proofs::{balance::BalanceProof, covenant::CovenantProof, ptx::PtxProof},
+    wire,
+};
 #[derive(Debug, Clone)]
 pub struct Bundle {
+    cm_roots: Vec<[u8; 32]>,
```
danielSanchezQ, line 13: Nitpick: Better explicit naming here. Not sure what is `cm` here atm 😅

danielSanchezQ: I would prefer to. I know its picky, but still 😅

zeegomo: >  it's used fairly extensively in code/pseudocode 

I meant to say it's a common name used in literature and code by other projects such as zerocash/zcash, we were not the ones to introduce this abbreviation.
But again, we can use `committment`

danielSanchezQ: Ok, you know better. So lets go for the common literature then. Thanks!

danielSanchezQ (approved):

ledger/nomos-ledger/src/lib.rs
```diff
@@ -722,8 +711,8 @@ pub mod tests {
 
     #[test]
     fn test_evolved_note_is_eligible_for_leadership() {
-        let note = note(0);
-        let (mut ledger, genesis) = ledger(&[note.note_commitment()]);
+        let note = note();
+        let (mut ledger, genesis) = ledger(&[note.commit(NF_SK.commit())]);
```
davidrusu, line 715:
```suggestion
        let (mut ledger, genesis) = ledger(&[commit(note)]);
```

nodes/nomos-node/src/config.rs
```diff
@@ -317,15 +317,17 @@ pub fn update_cryptarchia_consensus(
...
     if let Some(sk) = note_secret_key {
         let sk = <[u8; 16]>::from_hex(sk)?;
-
-        let value = note_value.expect("Should be available if coin sk provided");
-
-        cryptarchia.notes.push(InputWitness::new(
-            NoteWitness::basic(value as u64, NMO_UNIT, &mut rand::thread_rng()),
-            NullifierSecret::from_bytes(sk),
-        ));
+        cryptarchia.leader_config.nf_sk = NullifierSecret::from_bytes(sk);
```
davidrusu, line 330: Is the coin that will be used by this note in leadership proofs?

Don't we also need the note nonce in consensus_args in order to re-generate the same note commitment?

nomos-core/chain-defs/src/proofs/covenant.rs
```diff
@@ -0,0 +1,91 @@
...
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let (risc0_id, proof) = <([u32; 8], risc0_zkvm::Receipt)>::deserialize(deserializer)?;
+        proof.verify(risc0_id).map_err(serde::de::Error::custom)?;
+
+        Ok(Self { risc0_id, proof })
+    }
+}
```
davidrusu, line 91: So is this the intended way to verify a proof? Deserializing will also verify the risc0 proof?

Where do we check that the public inputs were what we expected?

nomos-core/chain-defs/src/tx/bundle.rs
```diff
@@ -63,35 +73,96 @@ impl Bundle {
...
+                    let covenant_public =
+                        covenant_proof.public_inputs().map_err(D::Error::custom)?;
+                    if covenant_public.ptx_root != ptx.root()
+                        || covenant_public.nf != input.nullifier
+                    {
+                        return Err(D::Error::custom("Invalid covenant proof public inputs"));
+                    }
+                    if covenant_proof.covenant() != input.covenant {
+                        return Err(D::Error::custom("Invalid covenant proof"));
+                    }
```
davidrusu, line 139: What do you think of:
```
covenant_proof.verify_binding(ptx.root(), &input)
```

nomos-services/cryptarchia-consensus/src/leadership.rs
```diff
@@ -1,20 +1,38 @@
...
 use nomos_core::header::HeaderId;
 use nomos_ledger::{leader_proof::Risc0LeaderProof, Config, EpochState, NoteTree};
+use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
 pub struct Leader {
     // for each block, the indexes in the note tree of the notes we control
-    notes: HashMap<HeaderId, Vec<InputWitness>>,
+    notes: HashMap<HeaderId, Vec<NoteWitness>>,
+    nf_sk: NullifierSecret,
```
davidrusu, line 16: The assumption here is that the leader should use the same sk for all notes? is this better than the old implementation that supported different sk's?

davidrusu (approved):
some qs, but looks good. 🚢

nomos-services/cryptarchia-consensus/src/leadership.rs
zeegomo: > The assumption here is that the leader should use the same sk for all notes?
Yes
> is this better than the old implementation that supported different sk's?
No, but proper support would require wallet-facing api/structures which we haven't built yet

nomos-core/chain-defs/src/proofs/covenant.rs
zeegomo: Yes. The idea is that strucutre are valid by type contract, so you don't have to call 'validate' functions (and it's impossible to forget to validate).
Covenant and balance proofs are checked during deserialization of `Bundle` as they have local inputs (and shouldn't be used anywhere else). ptx and leader proofs are checked when applying to the ledger

nomos-core/chain-defs/src/proofs/balance.rs
```diff
@@ -0,0 +1,74 @@
...
+        };
+
+        let env = risc0_zkvm::ExecutorEnv::builder()
+            .write(&bundle_private)?
+            .build()?;
+
+        let start_t = std::time::Instant::now();
+
+        let opts = risc0_zkvm::ProverOpts::groth16();
+        let prove_info = prover.prove_with_opts(env, nomos_cl_risc0_proofs::BUNDLE_ELF, &opts)?;
```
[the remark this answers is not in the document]
zeegomo: +1 for bundle balance

davidrusu (approved):

PR #851, opened 2024-10-23
Cli: Dissemination and retrieval

bacv:

Nomos cli was repurposed to disseminate and retrieve blobs using http api of executor and node. Code that used libp2p or is no longer needed was removed.

This version of `nomos-cli` is mainly updated for testnet interactions and might not reflect the future structure of cli crate.

Yes

N/A

N/A

N/A

## Checklist

* [x] 1. Description added.


nomos-cli/src/cmds/executor/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// Index for the Blob associated with Application ID.
+    #[clap(long)]
+    pub index: u64,
+    /// Executor address which is responsible for dissemination.
+    #[clap(long)]
+    pub addr: Url,
+}
+
+impl Disseminate {
+    pub fn run(&self) -> Result<(), Box<dyn std::error::Error>> {
```
danielSanchezQ, line 33: Why by ref?

nomos-cli/src/cmds/executor/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+            data.clone().into_bytes()
+        } else {
+            let file_path = self.file.as_ref().unwrap();
+            std::fs::read(file_path)?
+        };
+
+        let remainder = bytes.len() % 31;
+        if remainder != 0 {
+            bytes.resize(bytes.len() + (31 - remainder), 0);
+        }
```
danielSanchezQ, line 49: Probably just bubble up an error. Also instead of 31 we could use the definition from the proper crate (da-core iirc).

nomos-cli/src/cmds/executor/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+
+        while let Ok(update) = res_receiver.recv() {
+            match update {
+                Ok(_) => tracing::info!("Data successfully disseminated."),
+                Err(e) => {
+                    tracing::error!("Error disseminating data: {e}");
+                    return Err(e.into());
+                }
+            }
+        }
```
danielSanchezQ, line 73: Is this really doing more than a single loop?

nomos-cli/src/cmds/executor/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+        tokio::spawn(async move {
+            let res = client
+                .get_app_data_range_from_node::<DaBlob, kzgrs_backend::dispersal::Metadata>(
+                    app_id,
+                    from..to,
+                )
+                .await
+                .map_err(|err| format!("Failed to retrieve data form appid {app_id:?}: {err:?}",));
+            res_sender.send(res).unwrap();
+        });
```
danielSanchezQ, line 118: This just pulls data from a single executor as a validator? I do not remember adding this call to the client. Which should not be in the executor client btw.

nomos-cli/src/cmds/executor/mod.rs
bacv: Fixed, as we are not reusing chaining commands it's not necessary.

nomos-cli/src/cmds/executor/mod.rs
bacv: Moved this to validator "command", if following the executor-http-client pattern, we'd create similar library for the validator, but it's better to discuss this and implement in another PR.

nomos-cli/src/cmds/executor/mod.rs
bacv: Removed the loop, no need for it indeed.

nomos-cli/src/cmds/executor/mod.rs
danielSanchezQ: Yeah, we will have to do it, and it should be a different pr. thanks!

PR #855, opened 2024-10-25
Tests: Local debugging setup using grafana

bacv:

This PR adds a docker compose file that contains only the services for debugging, without the node itself. The tests module now has a feature flag `debug`, that will configure nodes in the topology to use loki and tempo while the test is running.

This is just another tool available for debugging interactions between two or more nodes. Integration tests without the `debug` flag will work as before - write logs to file.

Readme.md added in `tests/README.md`

@bacv

N/A

N/A

## Checklist

* [x] 1. Description added.


tests/Cargo.toml
```diff
@@ -60,3 +60,4 @@ path = "src/tests/da.rs"
 
 [features]
 metrics = ["nomos-node/metrics"]
+debug = []
```
danielSanchezQ, line 63: Probably we should add a default feature with nothing activated? 🤔

danielSanchezQ (approved):

tests/Cargo.toml
bacv: Is it really necessary? Not sure what the benefit would be, but maybe it's a convention?

danielSanchezQ: I'm not sure actually. Usually would be to activate other features by default. So probably for empty case is not necessary. Up to you.

bacv: Lets add it when we'll need it, I'll re-request the review, minor change for clippy.

danielSanchezQ (approved):

PR #858, opened 2024-10-26
Closing Connection Indicator

TakeASwing-420:
This is a work on progress. This is an extension of the #840


.gitignore
TakeASwing-420, line 1: It is kept intact.

nomos-da/network/core/Cargo.toml
```diff
@@ -29,4 +29,4 @@ thiserror = "1.0"
 tokio = { version = "1", features = ["macros", "rt-multi-thread", "time"] }
 libp2p = { version = "0.53", features = ["ed25519", "ping", "macros", "quic", "tcp", "yamux", "noise"] }
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
-
+crossbeam-channel = "0.5"
```
TakeASwing-420, line 32: added as a dev dependency.

TakeASwing-420:
@danielSanchezQ, I have created a custom indicator channel to signal for whether to close the sampling stream or not...comments are added accordingly in the code itself. I think, we can use this in other protocol test too, but I did not try to look deep into that.
Please, check it out.

nomos-da/network/core/src/protocols/sampling/mod.rs
```diff
@@ -80,19 +80,30 @@ mod test {
             p1_addresses.into_iter().collect(),
         );
         let mut p2 = sampling_swarm(k2.clone(), neighbours, p2_addresses.into_iter().collect());
-
         let request_sender_1 = p1.behaviour().sample_request_channel();
         let request_sender_2 = p2.behaviour().sample_request_channel();
         const MSG_COUNT: usize = 10;
+        let channel1 = Indicator::new(MSG_COUNT);
+        let channel2 = Indicator::new(MSG_COUNT);
```
danielSanchezQ, line 87: Please use better naming, we already have `Indicator`, why are we using `channelX`? 😅

nomos-da/network/core/src/test_utils.rs
```diff
@@ -27,3 +27,35 @@ impl MembershipHandler for AllNeighbours {
         self.neighbours.clone()
     }
 }
+
+use crossbeam_channel::{bounded, Receiver, Sender};
+
+/// A special-purpose multi-producer, multi-consumer(MPMC) channel to relay messages indicating whether the associated stream should be closed or not.
```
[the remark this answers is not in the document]
TakeASwing-420: Hm..I think I could've first clarified a bit about the usage...<br>
### Oneshot
Suppose `peer1` has finished sampling but the data packet for the last sample has not reached to `peer2`.
According to the `SamplingBehaviour`, on the very next iteration of `peer2` the `IncomingSample` event will be triggered.
But in this same loop, the `peer2` will try to listen to peer1's indicator channel. If you make the channel *oneshot*, then you would not be able to listen to peer1 in the next iteration (where you will actually get the sample in hand i.e. the `SamplingSuccess`) as you can only send message once.
### Tokio::MPSC
I preffered **crossbeam::mpmc** over **tokio::mpsc** in view of the potential need of sending messages across threads.
In future, if I need to send message across two async runtime threads( let's say t1 and t2) then tokio::mpsc is unable to do that.. because in the receiver in mpsc can listen **only in the thread which it was created**. Besides, crossbeam's mpmc is actually a wrapper over mpsc, so there is really no overhead.

nomos-da/network/core/src/test_utils.rs
```diff
@@ -27,3 +27,35 @@ impl MembershipHandler for AllNeighbours {
...
+}
+
+impl Clone for Indicator {
+    fn clone(&self) -> Self {
+        Self {
+            sender: self.sender.clone(),
+            receiver: self.receiver.clone(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
TakeASwing-420: The method name in crossbeam::mpmc is **clone** but in reality it just increases the reference count by 1(like the Arc::clone) so there is no ownership issue at all.

nomos-da/network/core/src/protocols/sampling/mod.rs
```diff
@@ -80,19 +80,30 @@ mod test {
...
-            let mut res = vec![];
+            let mut res = Vec::with_capacity(MSG_COUNT);
             loop {
+                if res.len() == MSG_COUNT {
+                    own_channel.send(true); // Indicate the peer that sampling is finished from this side
+                    if let Some(true) = peer_channel.receive() {
+                        // Break out only when peer has also finished sampling
+                        break res;
+                    }
+                }
```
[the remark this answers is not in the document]
TakeASwing-420: replied in the previous comment...

nomos-da/network/core/src/protocols/sampling/mod.rs
TakeASwing-420: Yeah, this one really needs to  be fixed...I copy-pasted from my previous version..apology

PR #859, opened 2024-10-28
Mix: Packet algorithms

youngjoon-lee:

This PR implements the packet algorithm (packing & packing) in the `nomos-mix-message` crate.
This is basically based on https://github.com/nymtech/sphinx, but MUCH simplified.

I'll rename the crate to `nomos-mix-packet` later. I think that's the more appropriate name. This crate defines all internal structures of the packet. And a packet is serialized to a message. All other parts in the mix protocol use the term "message".

This PR doesn't integrate the new packet implementation to the system. The system still uses a mock. I'll integrate it in the next PR.

The algorithm detail is not specified in the spec, but the basic requirement of the layered encryption can be found at the [Cryptographic Processor](https://www.notion.so/Nomos-Blend-Network-Tier-2-Message-Blending-Module-1208f96fb65c80e5bcb9df6e27472339?pvs=4#1208f96fb65c80f8a8d3d2b449953bde) section. Instead, I wrote detailed code comments. I'll also write the spec soon. For now, I think it's better to show you this code rather than the spec for easy understanding. Please start code review from the tests in `packet.rs`. I'm sure the code is easier to read than the spec doc.

Summary:
- Payload is layered-encrypted using Lioness encryption (as Sphinx does). We can consider other encryption algorithms, but I'd suggest to use Lioness for now and replace it later by proper researches. I wrote detailed code comments for this topic.
- Header is not encrypted in this PR. I will implement it at the next PR. It's not simple because of the length preservation requirement.
- Ephemeral public key derivation algorithm is implemented in this PR to avoid including all ephemeral public keys for all recipients (mix nodes) in the packet header. By using the derivation algorithm, we can include only one ephemeral public key in the packet header and allow recipients to derive the next ephemeral public key for the next recipient on the fly. Please see code comments for details. I explain it with an example there.

All of these ideas have been adopted from https://github.com/nymtech/sphinx, but I wrote the code from scratch by choosing right names for function/variable, removing unnecessary metadata/logic, and making all easy to read. 

No spec is written yet.

@youngjoon-lee 

No spec is written yet.

The spec must be written. @youngjoon-lee 

## Checklist

> [!WARNING]  

* [ ] 1. Description added.


nomos-mix/message/src/keyset.rs
```diff
@@ -0,0 +1,253 @@
...
+    pub fn derive_next_ephemeral_public_key(
+        &self,
+        shared_ephemeral_pubkey: &x25519_dalek::PublicKey,
+    ) -> x25519_dalek::PublicKey {
+        let new_shared_secert = self
+            .ephemeral_private_key
+            .diffie_hellman(shared_ephemeral_pubkey);
+        x25519_dalek::PublicKey::from(new_shared_secert.to_bytes())
+    }
+}
```
danielSanchezQ, line 147: Cant we use all this from the original sphix packet crate? My point is if we do not modify somehting and it can be reused is better. Easier for maintain.

youngjoon-lee: Let me check. Their code is not modularized well. and I'm not sure if they provide useful public functions.

youngjoon-lee: I just tried to use functions defined in https://github.com/nymtech/sphinx: 7e07e8488b68e500ef4f3ae8ecf3aeab25e826df in the [`mix-msg-using-sphinx`](https://github.com/logos-co/nomos-node/tree/mix-msg-using-sphinx) branch. I could remove `keyset.rs` and `payload.rs` by reusing their functions. But, unfortunately, only one function that is necessary for deriving the ephemeral public key for the next recipient is not exposed by the sphinx library: [`blind_the_shared_secret`](https://github.com/nymtech/sphinx/blob/344b902df340e0d5af69c5147b05f76f324b8cef/src/header/mod.rs#L234). So, I copied the function to our codebase: [`derive_next_ephemeral_public_key`](https://github.com/logos-co/nomos-node/blob/7e07e8488b68e500ef4f3ae8ecf3aeab25e826df/nomos-mix/message/src/packet.rs#L117).

That function is tightly coupled with the header implementation in Sphinx. However, their header implementation contains some metadata that isn't necessary to us. Also, we're going to add some new metadata to the header that is going to be defined by us (e.g. incentivization). That's way we cannot use their header implementation.

Hm, let me think. The easiest way may be just using the Sphinx as it is for now, without modifying anything, with filling dummy data into unnecessary metadata. It will work. But, in the end, I think we need to adopt all details from them and modify them on our own.

danielSanchezQ: Well, we can just modify that part. If we cannot reuse it its fine to have a copy of it ourselves. But if we reduced the duplication and maintain of code we do not really own is better. Thanks for the extra effort!!

youngjoon-lee: Okay. I merged the `mix-msg-using-sphinx` branch into this branch with some refactoring. You can see the change 8cb0bc7fb3d65791faeaf99b6e0f94814d3c2975. As mentioned above, I copied one function from their code. Also you may find that there is a "not-natural" piece of code regarding "dummy node address", even though we don't need any node address in our design. It's because their code isn't modularized well. I left explanation in code comment.

So far, it's fine. However, we won't be able to reuse their code for encrypting headers (which I'll do at the next PR). Their code is tightly coupled with their own header definition and its size. So, the next PR will contain some complicated algorithms modified from their code. I'll try my best to reuse pieces from their code, but it's not gonna be as simple as this PR.

nomos-mix/message/src/packet.rs
```diff
@@ -0,0 +1,250 @@
...
+
+#[derive(Debug, Serialize, Deserialize)]
+struct RoutingInfo {
+    // TODO: Change this to `is_final_layer: bool`
+    // by implementing the length-preserved layered encryption.
+    // It's not good to expose the info that how many layers remain to the intermediate recipients.
+    remaining_layers: u8,
+    // TODO:: Add the following fields
+    // header_integrity_hamc
+    // additional data (e.g. incentivization)
```
danielSanchezQ, line 34: Are we gonna asses this in the next PRs? If not, lets open issues.

danielSanchezQ (approved):

nomos-mix/message/src/packet.rs
youngjoon-lee: Additional data is not determined yet in the spec. Marcin is mainly working on designing this mechanism. I'll open an issue for this but it will take some time to kick off the dev.

youngjoon-lee: header_integrity_hmac will be added at the next PR. That's essential for encryption/decryption.

PR #860, opened 2024-10-28
Mix: Persistent rate as stream + Extension

danielSanchezQ:

This PR reformats part of the mix protocols to use streams instead of channels+tasks approach.
It should not modify any previous behaviour and it tries to fullfill the same tests as the previous version. 
This just transforms the tier 1 persistent rate and does not integrate it. Further PRs will be implemented for other tiers and a final one for the switch of systems and removal of old approach.

Hope so

Yes

Not at all.

## Checklist

* [x] 1. Description added.


nomos-mix/core/src/persistent_transmission.rs
```diff
@@ -26,6 +30,86 @@ impl Default for PersistentTransmissionSettings {
...
+        settings: PersistentTransmissionSettings,
+    ) -> PersistentTransmissionStream<Self>
+    where
+        Self: Sized + Unpin,
+    {
+        PersistentTransmissionStream::new(settings, self)
+    }
+}
+
+impl<S> PersistentTransmissionExt for S where S: Stream {}
```
youngjoon-lee, line 111: Oh, I haven't known about this skill, which implements `PersistentTransmissionExt` for any type that implements `Stream`.
If this is a common practice, I think I can adopt this for the `TemporalProcessor` that implements `Stream`, to have the common format.

nomos-mix/core/src/persistent_transmission.rs
```diff
@@ -26,6 +30,86 @@ impl Default for PersistentTransmissionSettings {
...
+                    }
+                }
+            };
+        };
+        if let Poll::Ready(Some(item)) = pin!(stream).poll_next(cx) {
+            queue.push_back(item);
+        }
+        Poll::Pending
+    }
+}
```
youngjoon-lee, line 97: What do you think about removing the `queue: VecDeque` and use the `stream` as a queue, as below? The test passes, but I'd like to ask if I'm missing something.
```rust
impl<S> Stream for PersistentTransmissionStream<S>
where
    S: Stream<Item = Vec<u8>> + Unpin,
{
    type Item = S::Item;

    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
        let Self {
            ref mut interval,
            ref mut stream,
            ref mut coin,
            ..
        } = self.get_mut();
        if pin!(interval).poll_tick(cx).is_ready() {
            if let Poll::Ready(Some(item)) = pin!(stream).poll_next(cx) {
                return Poll::Ready(Some(item));
            } else if coin.flip() {
                return Poll::Ready(Some(DROP_MESSAGE.to_vec()));
            }
        };
        Poll::Pending
    }
}
```

danielSanchezQ: So, this depends on the "backpresure" law that we want to have. For example, we could have a stream wrapper over a bounded channel in where if we do not pull and we block the buffer it will block sends. 
I actually just tried to maintain the original implementation here. But I think that we could go for removing the queue.

youngjoon-lee: Ah, you're right. My suggestion works only when the user uses an unbound channel. The user doesn't care about the internal implementation of the `PersistentTransmissionStream`, which delays consuming messages from the channel. So, yeah, it makes sense to consume messages from the channel (stream) and store them in the queue.

youngjoon-lee: I just saw your commit [12f134c](https://github.com/logos-co/nomos-node/pull/860/commits/12f134c7369d9f65fdb6d0d636da49fc3b7f1351) that removed the queue. I think we can go with the way that you think the optimal. I want to learn the practice.

youngjoon-lee: Yeha, I think that's the issue that most of backpressures have. The producers don't know how fast the consumer will consume the channel.

PR #897, opened 2024-10-31
Mix: Packet header encryption

youngjoon-lee:

This PR implements:
- `RoutingInformation` that tells the node what to do with the packet.
- Encrypting multiple `RoutingInformation`s using `ConsistentLengthLayeredCipher`.
- Decrypting the layered encrypted `RoutingInformation`s one by one.

The cipher algorithm has been adopted from https://github.com/nymtech/sphinx, but
- with removing all unnecessary metadata from `RoutingInformation`.
- with refactoring all scattered code of encryption/decryption into `ConsistentLengthLayeredCipher`, which is created by me, not from their code. 
  - by removing all hardcoded numbers to make it work with a custom max number of layers and a custom size of `RoutingInformation`, which fit our needs.

This implementation is for the [Tier 2: Cryptographic Processor](https://www.notion.so/Nomos-Blend-Network-Tier-2-Message-Blending-Module-1208f96fb65c80e5bcb9df6e27472339#1208f96fb65c80f8a8d3d2b449953bde), but this spec document doesn't contain the header construction yet. You can find my research on Sphinx packet construction from [Notion](https://www.notion.so/Sphinx-Packet-d75ce5baf4344e4880f7098fb8f6a864?pvs=4) but it is tightly coupled with the data structures defined in Sphinx and over-complicated.

I will write the spec as soon as possible. Until then, you can see code comments that I carefully wrote. I've tried my best to write detailed/easy comments to explain how the algorithm works, even though it's not at the level of the spec doc.
Actually, I think it's better to show algorithms by code if it's not simple.

@youngjoon-lee 

Yes, but most of specs are not written yet. I'll write them ASAP.

No. I have to write specs.

## Checklist

> [!WARNING]  

* [ ] 1. Description added.


nomos-mix/message/src/layered_cipher.rs
```diff
@@ -0,0 +1,321 @@
...
+/// The cipher also provides the length-preserved decryption.
+/// Even if one layer of encryptions is decrypted, the length of decrypted data is
+/// the same as the length of the original data.
+/// For example:
+///   len(encrypt(k0, [data0, encrypt(k1, [data1])])) == len(encrypt(k1, [data1]))
+pub struct ConsistentLengthLayeredCipher {
+    /// All encrypted data produced by the cipher has the same size according to the `max_layers`.
+    pub max_layers: usize,
+    /// The size of data that will be included in each layer.
+    pub data_size: usize,
```
youngjoon-lee, line 49: I'm refactoring this using generics. Until then, I mark this PR as draft

youngjoon-lee: Done. Ready to be reviewed.

nomos-mix/message/src/layered_cipher.rs
```diff
@@ -0,0 +1,331 @@
...
+/// Even if the number of data and keys provided for encryption is smaller than `max_layers`,
+/// The cipher always produces the max-size output regardless of the number of data and keys provided,
+/// in order to ensure that all outputs generated by the cipher are the same size.
+///
+/// The cipher also provides the length-preserved decryption.
+/// Even if one layer of encryptions is decrypted, the length of decrypted data is
+/// the same as the length of the original data.
+/// For example:
+///   len(encrypt(k0, [data0, encrypt(k1, [data1])])) == len(encrypt(k1, [data1]))
+pub struct ConsistentLengthLayeredCipher<D: ConsistentLengthLayeredCipherData> {
```
danielSanchezQ, line 47: Sometimes there is no need to include the bound here, just in impl blocks. Maybe just `D` could be enough.

nomos-mix/message/src/layered_cipher.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// The total size of fully encrypted output that includes all layers.
+    /// This size is determined by [`D::size`] and [`Self::max_layers`].
+    pub fn total_size(&self) -> usize {
+        Self::single_layer_size() * self.max_layers
+    }
+
+    /// The size of a single layer that contains a data and a MAC.
+    /// The MAC is used to verify integrity of the encrypted next layer.
+    fn single_layer_size() -> usize {
+        D::size() + HEADER_INTEGRITY_MAC_SIZE
```
danielSanchezQ, line 93: if `size` method is const, then we can make this method const as well.

nomos-mix/message/src/layered_cipher.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// Perform the layered encryption.
+    pub fn encrypt(&self, params: &[EncryptionParam<D>]) -> Result<(Vec<u8>, HeaderIntegrityMac)> {
+        if params.is_empty() || params.len() > self.max_layers {
+            return Err(Error::InvalidParams);
+        }
+
+        params
+            .iter()
+            .take(params.len() - 1) // Exclude the last param that will be treated separately below.
+            .rev() // Data and keys must be used in reverse order to encrypt the inner-most layer first
```
danielSanchezQ, line 105: Is there a way to make sure this actually happens like that? as caller may have already reverse the order of Params.
Not to do now, but in this case I think we may construct some structure for it. wdyt?

nomos-mix/message/src/layered_cipher.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// Fillers are encrypted in accumulated manner by keys.
+    fn build_fillers(&self, params: &[EncryptionParam<D>]) -> Vec<u8> {
+        params
+            .iter()
+            .map(|param| &param.key.stream_cipher_key)
+            .fold(Vec::new(), |mut fillers, key| {
+                fillers.extend(vec![0u8; Self::single_layer_size()]);
+                self.apply_streamcipher(&mut fillers, key, StreamCipherOption::FromBack);
+                fillers
+            })
```
danielSanchezQ, line 187: Nitpick: You can probably allocate the full vec beforehand and just call `apply_stream_cipher` to the region you need?

nomos-mix/message/src/layered_cipher.rs
```diff
@@ -0,0 +1,331 @@
...
+            StreamCipherOption::FromFront => &pseudorandom_bytes[..data.len()],
+            StreamCipherOption::FromBack => {
+                &pseudorandom_bytes[pseudorandom_bytes.len() - data.len()..]
+            }
+        };
+        Self::xor(data, pseudorandom_bytes)
+    }
+
+    // In-place XOR operation: b is applied to a.
+    fn xor(a: &mut [u8], b: &[u8]) {
```
danielSanchezQ, line 250:
```suggestion
    fn xor_in_place(a: &mut [u8], b: &[u8]) {
```

nomos-mix/message/src/lib.rs
```diff
@@ -62,3 +64,26 @@ pub fn unwrap_message(message: &[u8]) -> Result<(Vec<u8>, bool), Error> {
...
     message == DROP_MESSAGE
 }
+
+pub(crate) fn concat_bytes(bytes_list: &[&[u8]]) -> Vec<u8> {
+    let mut buf = Vec::with_capacity(bytes_list.iter().map(|bytes| bytes.len()).sum());
+    bytes_list
+        .iter()
+        .for_each(|bytes| buf.extend_from_slice(bytes));
+    buf
+}
```
danielSanchezQ, line 74: Would be better to:

```suggestion
#[inline]
pub(crate) fn concat_bytes(bytes_list: &[&[u8]]) -> Vec<u8> {
    bytes_list.into_iter().flat_map(|&s| s).copied().collect()
}
```

nomos-mix/message/src/packet.rs
```diff
@@ -135,6 +147,41 @@ impl Packet {
         let new_shared_secret = blinding_factor.diffie_hellman(cur_ephemeral_pubkey);
         x25519_dalek::PublicKey::from(new_shared_secret.to_bytes())
     }
+
+    pub fn to_bytes(&self) -> Vec<u8> {
+        concat_bytes(&[
+            &self.header.ephemeral_public_key.to_bytes(),
+            &self.header.encrypted_routing_info.to_bytes(),
+            &self.payload,
+        ])
```
danielSanchezQ, line 156: Yeah, I think it would be better to [chain](https://docs.rs/itertools/latest/itertools/macro.chain.html) the iterator and collect the bytes. No need to create the extra array and references.

nomos-mix/message/src/layered_cipher.rs
```diff
@@ -0,0 +1,331 @@
...
+/// The cipher also provides the length-preserved decryption.
+/// Even if one layer of encryptions is decrypted, the length of decrypted data is
+/// the same as the length of the original data.
+/// For example:
+///   len(encrypt(k0, [data0, encrypt(k1, [data1])])) == len(encrypt(k1, [data1]))
+pub struct ConsistentLengthLayeredCipher<D: ConsistentLengthLayeredCipherData> {
+    /// All encrypted data produced by the cipher has the same size according to the `max_layers`.
+    pub max_layers: usize,
+    _phantom_data: PhantomData<D>,
+}
```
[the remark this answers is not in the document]
youngjoon-lee: It's static. I was thinking about having it as a constant at the upper layer (e.g. mix service).

nomos-mix/message/src/layered_cipher.rs
youngjoon-lee: Here, the data size (`D::size()`) depends on the max size of a block proposal in our case. So, I think it's hard to set a proper constant in this library level.

youngjoon-lee: Yeah, but, In our codebase, the `D` is always the `RoutingInfo` struct. If we make it as a constant, we need to update the constant when we add more fields to the struct. And, its size depends on how it is serialized.

nomos-mix/message/src/packet.rs
youngjoon-lee: Thank you. [a015485](https://github.com/logos-co/nomos-node/pull/897/commits/a0154853b8f53504fb7fd26e03ea5ab1b471585b)

nomos-mix/message/src/lib.rs
youngjoon-lee: Resolved by https://github.com/logos-co/nomos-node/pull/897#discussion_r1850026700

nomos-mix/message/src/layered_cipher.rs
youngjoon-lee: Right. Thanks. [3d32942](https://github.com/logos-co/nomos-node/pull/897/commits/3d32942725a2d769e6a928b74426c97d287f11ab)

nomos-mix/message/src/layered_cipher.rs
youngjoon-lee: Thanks [3d32942](https://github.com/logos-co/nomos-node/pull/897/commits/3d32942725a2d769e6a928b74426c97d287f11ab)

nomos-mix/message/src/error.rs
```diff
@@ -4,10 +4,16 @@ pub enum Error {
...
     #[error("Payload is too large")]
     PayloadTooLarge,
-    #[error("Too many recipients")]
-    TooManyRecipients,
     #[error("Sphinx packet error: {0}")]
     SphinxPacketError(#[from] sphinx_packet::Error),
+    #[error("Invalid packet")]
+    InvalidPacket,
+    #[error("Invalid routing flag: {0}")]
+    InvalidRoutingFlag(u8),
```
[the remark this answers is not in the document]
youngjoon-lee: Thanks 6fa2071

nomos-mix/message/src/layered_cipher.rs
youngjoon-lee: Thanks. 30eefbb

bacv (approved):

PR #912, opened 2024-11-07
Just use priv/pubkey as Node ID in the mock mix message encoding

youngjoon-lee:

In the `MockMixMessage`, we don't need the actual x25519 private/public key because we don't encrypt/decrypt anything. So, it's simpler to use private/public key (passed as a function parameter) as a node ID for message encoding. Then, users don't need to derive public keys from private keys to use this `MockMixMessage`.

Yes

@youngjoon-lee

n/a

n/a

## Checklist

> [!WARNING]  

* [x] 1. Description added.

### DIFF
--- a/nomos-mix/message/src/mock/mod.rs
+++ b/nomos-mix/message/src/mock/mod.rs
@@ -3,21 +3,20 @@ use crate::{Error, MixMessage};
 //
 /// A mock implementation of the Sphinx encoding.
 
-const PRIVATE_KEY_SIZE: usize = 32;
-const PUBLIC_KEY_SIZE: usize = 32;
+const NODE_ID_SIZE: usize = 32;
 
 const PADDED_PAYLOAD_SIZE: usize = 2048;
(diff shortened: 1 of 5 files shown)


danielSanchezQ (approved):

youngjoon-lee:
I also need to change mix service not to convert privkey to pubkey to make tests work

youngjoon-lee:
> I also need to change mix service not to convert privkey to pubkey to make tests work

done

youngjoon-lee:
I think we don't need to change the MockMixMessage here, though it doesn't do any encryption. The user (simulation) can properly derive public keys to use `MockMixMessage`. It's not a big deal.

youngjoon-lee:
> I think we don't need to change the MockMixMessage here, though it doesn't do any encryption. The user (simulation) can properly derive public keys to use `MockMixMessage`. It's not a big deal.

I just reopened this PR. Still it would be good to just use NODE_ID in MockMixMessage, to reduce the computation overhead for converting private key to public key whenever the `unpack_message` is called.

PR #916, opened 2024-11-08
Add winning_probability param to replace num_hops in CoverTrafficSettings

youngjoon-lee:
Instead of calculating the `winning_probability` from `num_of_hops` in `CoverTrafficSettings`, I think it's better to let users set the `winning_probability` directly. Now we think that the `winning_probability` is related to the `num_of_hops` as written in the [spec](https://www.notion.so/Nomos-Blend-Network-Tier-3-Cover-Traffic-Module-10b8f96fb65c80cab153de10115e0023?pvs=4#1158f96fb65c805c8497c3b2860e7e6d), but we also found (during the simulation) that we sometimes need to set a totally different value. Of course, we're probably gonna recommend users to set it to `1 / num_of_hops`.

Yes

@danielSanchezQ @youngjoon-lee @madxor 

n/a

It doesn't rely on the network paraters, but instead it does take an specific value. We need to make sure this is what we want/need @madxor 

## Checklist

> [!WARNING]  

* [x] 1. Description added.

### DIFF
--- a/nomos-mix/core/src/cover_traffic.rs
+++ b/nomos-mix/core/src/cover_traffic.rs
@@ -12,13 +12,12 @@ use std::task::{Context, Poll};
 #[derive(Copy, Clone, Deserialize)]
 pub struct CoverTrafficSettings {
     pub node_id: [u8; 32],
-    pub number_of_hops: usize,
+    pub winning_probability: f64,
     pub slots_per_epoch: usize,
     pub network_size: usize,
 }
 
 pub struct CoverTraffic<EpochStream, SlotStream, Message> {
-    winning_probability: f64,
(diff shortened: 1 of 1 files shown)


bacv (approved):

danielSanchezQ (approved):

youngjoon-lee:
I'm closing this without merging. The description that I wrote above is still valid, but on the other hand, it's also helpful to have `num_hops` instead of `winning_probability` because we don't need to educate users to set `winning_probability = 1/num_hops`. Therefore, I'm not merging this PR now, but we can reopen this later. I'm running simulations to understand the mechanics of parameters. The result will make clear how to set/organize parameters.

PR #925, opened 2024-11-29
Tracing: Always use Once to initialize tracing_subscriber

youngjoon-lee:
I've tried to use `Tracing` service for some integration tests that runs multiple Overwatch instance(?) that contains the `Tracing` service. But, the instances panic-ed because the tracing_subscriber was initialized more than once. 
I found that the `Tracing` service uses `Once` to initialize the tracing_subscriber only when `#[cfg(test)]` is enabled. But, the `test` configuration flag is enabled only when running unit tests. It is not enabled for integration tests, if I understand it correctly.

So, in this PR, I changed the `Tracing` service to always use `Once` regardless of `#[cfg(test)]`, because we anyway should initialize the tracing_subscriber only once. 

Please let me know if this is not correct, or if there is any better way. I've tested that the integration test with multiple Overwatch instances work well. Also, the standalone `nomos-node` prints logs well.

Yes

@youngjoon-lee @bacv 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 1. Description added.

### DIFF
--- a/nomos-services/tracing/src/lib.rs
+++ b/nomos-services/tracing/src/lib.rs
@@ -1,7 +1,10 @@
 // std
-use std::fmt::{Debug, Formatter};
 use std::io::Write;
 use std::sync::{Arc, Mutex};
+use std::{
+    fmt::{Debug, Formatter},
+    sync::Once,
+};
 // crates
 use futures::StreamExt;
 use nomos_tracing::filter::envfilter::{create_envfilter_layer, EnvFilterConfig};
@@ -152,9 +155,6 @@ impl ServiceData for Tracing {
 #[async_trait::async_trait]
 impl ServiceCore for Tracing {
     fn init(service_state: ServiceStateHandle<Self>) -> Result<Self, overwatch_rs::DynError> {
-        #[cfg(test)]
-        use std::sync::Once;
-        #[cfg(test)]
         static ONCE_INIT: Once = Once::new();
 
         let config = service_state.settings_reader.get_updated_settings();
@@ -218,18 +218,12 @@ impl ServiceCore for Tracing {
             });
         }
 
-        #[cfg(test)]
         ONCE_INIT.call_once(move || {
             tracing_subscriber::registry()
                 .with(LevelFilter::from(config.level))
                 .with(layers)
                 .init();
         });
-        #[cfg(not(test))]
-        tracing_subscriber::registry()
-            .with(LevelFilter::from(config.level))
-            .with(layers)
-            .init();
 
         Ok(Self {
             service_state,


bacv (approved):
> So, in this PR, I changed the Tracing service to always use Once regardless of #[cfg(test)]

I think this change makes sense. But it is still not very clear why couple of Overwatch instances in the same process, but with `#[cfg(test)]` flag failed to initialize it once?

bacv:
> But, the test configuration flag is enabled only when running unit tests.

Which command are you using to run tests?

youngjoon-lee:
> > But, the test configuration flag is enabled only when running unit tests.
> 
> Which command are you using to run tests?

I ran:
```
cargo test -p <crate-name> —tests
```
I wrote a simple test by creating `nomos-services/mix/tests/` directory. I didn’t commit this to the repo because this was just a temporary test to check something on my side, and I think you’re preparing something similar. 
Maybe I’m running tests in a wrong way. Let me also double check about the config flag.

danielSanchezQ:
I do not get the issue here.

If using `Once` it should work for both single process (shared memory) and multiprocess (non-shared memory).

youngjoon-lee:
> I do not get the issue here.
> 
> If using `Once` it should work for both single process (shared memory) and multiprocess (non-shared memory).

Yes, but the issue was that `Once` wasn't used when running integration tests (not unit tests).

youngjoon-lee:
> Aaah, that should not matter. In integration tests we are using independent instances (processes) for each node (iirc). So the issue would be something else entirely 🤔

Ah, let me rephrase my description. Sorry for confusion. @danielSanchezQ @bacv 
I'm not taking about the integration tests in `root/tests/`. I wrote a test under the `nomos-services/mix/tests/`. I didn't open a PR for this test yet. I thought I mentioned this in the PR description, but I just found that I didn't :)
https://github.com/logos-co/nomos-node/blob/mix-conn-monitor-intergration-test/nomos-services/mix/tests/frequency.rs
I call this integration test (as introduced in https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests), but yeah it's confusing.

In this test, we have the [`TestNode`](https://github.com/logos-co/nomos-node/blob/78b8cc585e1864512fc8321e12f59672577663c2/nomos-services/mix/tests/frequency.rs#L94) that has `Tracing` service, and we run multiple `TestNode`s using Overwatch in one process. However, because `#[cfg(test)]` is not enabled when running integration(?) tests, the `Once` is not used.

It may be better to re-open this PR when I open a PR for this test.

youngjoon-lee:
Closing this since nobody feels a need of this now.I will reopen this when I open PRs for testing blend modules.

PR #926, opened 2024-11-29
poc: StorageBackend for SurrealDB

AlejandroCabeza:
Implements StorageBackend for SurrealDB with the intention to verify it works on different platforms.

I didn't know what labels to assing.

A couple relevant quirks worth mentioning:
1. Couldn't find a way to directly use bytes, when using `db.create().content()` objects are required.
2. `StorageBackend::new` is sync. Due SurrealDB's `connect` being async, `block_in_place` is used alongside `block_on` for connection:
    - An alternative to this would be initialising the connection in the `Overwatch` handle, and passing it to the `StorageBackend` via `Settings`.
    - Related [Discord Thread](https://discord.com/channels/1111286067413405788/1312075484611346504)
3. Transactions are not currently supported by the [Rust SDK](https://github.com/surrealdb/surrealdb/blob/c717b27005b7fb2608c9b08c1f6c94fdd6764a84/crates/sdk/src/api/method/mod.rs#L257). 
    - You  still can write transactions by chaining queries (e.g. `db.query("CREATE ...").query("SELECT ...").await`), as those calls would be wrapped in a single transaction. Is it worth adding transactions this way now, or shall we wait until they're fully available?
 4. `opentelemetry`'s version had to be updated due to `SurrealDB` requiring a newer `Tokio` version.

@AlejandroCabeza @danielSanchezQ 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 1. Description added.

### DIFF
--- a/nomos-services/storage/Cargo.toml
+++ b/nomos-services/storage/Cargo.toml
@@ -17,6 +17,7 @@ rocksdb = { version = "0.22", optional = true }
 surrealdb = { version = "2.1.2", optional = true }
 thiserror = "1.0"
 tracing = "0.1"
+serde_json = "1.0.133"
 
 [dev-dependencies]
 tokio = { version = "1", features = ["sync", "macros", "time"] }
--- a/nomos-services/storage/Cargo.toml
+++ b/nomos-services/storage/Cargo.toml
@@ -14,13 +14,14 @@ overwatch-rs = { git = "https://github.com/logos-co/Overwatch", rev = "2f70806" 
 serde = "1.0"
 sled = { version = "0.34", optional = true }
 rocksdb = { version = "0.22", optional = true }
-surrealdb = { version = "2.1.2", optional = true}
+surrealdb = { version = "2.1.2", optional = true }
 thiserror = "1.0"
 tracing = "0.1"
 
 [dev-dependencies]
 tokio = { version = "1", features = ["sync", "macros", "time"] }
 tempfile = "3"
+surrealdb = { version = "2.1.2", features = ["kv-mem"] }
 
 [features]
 default = []
--- a/nomos-services/storage/Cargo.toml
+++ b/nomos-services/storage/Cargo.toml
@@ -21,7 +21,7 @@ tracing = "0.1"
 [dev-dependencies]
 tokio = { version = "1", features = ["sync", "macros", "time"] }
 tempfile = "3"
-surrealdb = { version = "2.1.2", features = ["kv-mem"] }
+surrealdb = { version = "2", features = ["kv-mem"] }
 
 [features]
 default = []
--- a/nomos-services/storage/Cargo.toml
+++ b/nomos-services/storage/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2021"
 [dependencies]
 async-trait = "0.1"
 futures = "0.3"
-tokio = { version = "1.41.1", features = ["sync"] }
+tokio = { version = "1", features = ["sync"] }
 bytes = "1.2"
 overwatch-rs = { git = "https://github.com/logos-co/Overwatch", rev = "2f70806" }
 serde = "1.0"
--- a/nomos-services/storage/Cargo.toml
+++ b/nomos-services/storage/Cargo.toml
@@ -8,12 +8,13 @@ edition = "2021"
 [dependencies]
 async-trait = "0.1"
 futures = "0.3"
-tokio = { version = "1", features = ["sync"] }
+tokio = { version = "1.41.1", features = ["sync"] }
 bytes = "1.2"
 overwatch-rs = { git = "https://github.com/logos-co/Overwatch", rev = "2f70806" }
 serde = "1.0"
 sled = { version = "0.34", optional = true }
 rocksdb = { version = "0.22", optional = true }
+surrealdb = { version = "2.1.2", optional = true}
 thiserror = "1.0"
 tracing = "0.1"
 
@@ -26,6 +27,7 @@ default = []
 mock = []
 sled-backend = ["sled"]
 rocksdb-backend = ["rocksdb"]
+surrealdb-backend = ["surrealdb"]
 
 [[bin]]
 name = "rocks"
--- a/nomos-services/storage/src/backends/mod.rs
+++ b/nomos-services/storage/src/backends/mod.rs
@@ -2,9 +2,10 @@
 pub mod mock;
 #[cfg(feature = "sled")]
 pub mod sled;
-
 #[cfg(feature = "rocksdb")]
 pub mod rocksdb;
+#[cfg(feature = "surrealdb-backend")]
+pub mod surrealdb;
 
 // std
 use std::error::Error;
--- a/nomos-services/storage/src/backends/mod.rs
+++ b/nomos-services/storage/src/backends/mod.rs
@@ -1,9 +1,9 @@
 #[cfg(feature = "mock")]
 pub mod mock;
-#[cfg(feature = "sled")]
-pub mod sled;
 #[cfg(feature = "rocksdb")]
 pub mod rocksdb;
+#[cfg(feature = "sled")]
+pub mod sled;
 #[cfg(feature = "surrealdb-backend")]
 pub mod surrealdb;
 
--- a/nomos-services/storage/src/backends/surrealdb.rs
+++ b/nomos-services/storage/src/backends/surrealdb.rs
@@ -1,73 +1,171 @@
 use crate::backends::{StorageBackend, StorageSerde, StorageTransaction};
+use async_trait::async_trait;
 use bytes::Bytes;
-use surrealdb::{Error, Surreal};
-use tokio::task::spawn_blocking;
+use serde::{Deserialize, Serialize};
+use std::marker::PhantomData;
+use std::str::from_utf8;
+use surrealdb::engine::any::{connect, Any};
+use surrealdb::Surreal;
+use tokio::runtime::Handle;
+use tokio::task::block_in_place;
 
+#[derive(Debug, Serialize, Deserialize)]
+struct Data {
+    pub data: Bytes,
+}
+
+impl Data {
+    fn new(value: Bytes) -> Self {
+        Self { data: value }
+    }
+
+    fn to_bytes(&self) -> Bytes {
+        self.data.clone()
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error(transparent)]
+    Surreal(#[from] surrealdb::Error),
+    #[error(transparent)]
+    Utf8(#[from] std::str::Utf8Error),
+}
+
+#[derive(Clone)]
 pub struct SurrealBackendSettings {
     pub db_path: String,
+    pub namespace: String,
+    pub database: String,
+}
+
+impl SurrealBackendSettings {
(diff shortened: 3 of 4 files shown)


nomos-services/storage/Cargo.toml
```diff
@@ -8,24 +8,27 @@ edition = "2021"
 [dependencies]
 async-trait = "0.1"
 futures = "0.3"
-tokio = { version = "1", features = ["sync"] }
+tokio = { version = "1.41.1", features = ["sync"] }
```
danielSanchezQ, line 11: Why fix the version? We recently changed all deps to use just latest 1.

nomos-services/storage/src/backends/mod.rs
```diff
@@ -1,10 +1,11 @@
 #[cfg(feature = "mock")]
 pub mod mock;
-#[cfg(feature = "sled")]
-pub mod sled;
-
 #[cfg(feature = "rocksdb")]
 pub mod rocksdb;
+#[cfg(feature = "sled")]
+pub mod sled;
+#[cfg(feature = "surrealdb-backend")]
```
danielSanchezQ, line 7: for consistency it could be just `"surrealdb"`

danielSanchezQ (commented):
I'll take a proper look later. But one thing that took my attention is that we are using normal queries. Didn't surreal had a proper `K/V` API?

nomos-services/storage/Cargo.toml
AlejandroCabeza: I thought I needed it when I was fixing some dep incompatibility, but it's not anymore. Removing.

nomos-services/storage/src/backends/mod.rs
AlejandroCabeza: That's fair. That being said, why use the library name instead of the feature name?

AlejandroCabeza:
> I'll take a proper look later. But one thing that took my attention is that we are using normal queries. Didn't surreal had a proper `K/V` API?

I couldn't find it in the Rust SDK. Actually, I'll check whether we can access it by using queries.

nomos-services/storage/src/backends/mod.rs
danielSanchezQ: But you named the feature. All the other backends use the same feature name as the library. Maybe I do not get your question though.

danielSanchezQ:
> > I'll take a proper look later. But one thing that took my attention is that we are using normal queries. Didn't surreal had a proper `K/V` API?
> 
> I couldn't find it in the Rust SDK. Actually, I'll check whether we can access it by using queries.

I see this: https://github.com/surrealdb/surrealkv
But maybe is a different db implementation? or is it an API layer on top?
That is what we probably need to know though.

nomos-services/tracing/Cargo.toml
```diff
@@ -10,8 +10,8 @@ async-trait = "0.1"
 futures = "0.3"
 nomos-tracing = { path = "../../nomos-tracing" }
 opentelemetry = { version = "0.25" }
-opentelemetry-otlp = "0.25"
-opentelemetry_sdk = { version = "0.25", features = ["rt-tokio"] }
+opentelemetry-otlp = "0.26"
```
bacv, line 13: I think these were updated because of previously added tokio version, lets better create a new PR if tracing can be updated, I think couple more crates like `opentelemtry` should also be updated.

AlejandroCabeza:
> > > I'll take a proper look later. But one thing that took my attention is that we are using normal queries. Didn't surreal had a proper `K/V` API?
> > 
> > 
> > I couldn't find it in the Rust SDK. Actually, I'll check whether we can access it by using queries.
> 
> I see this: https://github.com/surrealdb/surrealkv But maybe is a different db implementation? or is it an API layer on top? That is what we probably need to know though.

That was what I was checking out. 
So, all of `surrealdb` is an abstraction on top of a key-value engine. For that key-value engine you could use `rocksdb` (and a couple other supported engines), or leverage the surreal team's own key-value engine, `surrealkv`.

nomos-services/tracing/Cargo.toml
AlejandroCabeza: These were updated due to a constraint that `surrealdb` adds. It requires a newer version of `Tokio`, which conflicts with this `opentelemetry` version's requirements.

danielSanchezQ:
> > > > I'll take a proper look later. But one thing that took my attention is that we are using normal queries. Didn't surreal had a proper `K/V` API?
> > > 
> > > 
> > > I couldn't find it in the Rust SDK. Actually, I'll check whether we can access it by using queries.
> > 
> > 
> > I see this: https://github.com/surrealdb/surrealkv But maybe is a different db implementation? or is it an API layer on top? That is what we probably need to know though.
> 
> That was what I was checking out. So, all of `surrealdb` is an abstraction on top of a key-value engine. For that key-value engine you could use `rocksdb` (and a couple other supported engines), or leverage the surreal team's own key-value engine, `surrealkv`.

Wait, surrealdb uses a kv and the underlaying db? or just the surrealkv?

AlejandroCabeza:
> > > > > I'll take a proper look later. But one thing that took my attention is that we are using normal queries. Didn't surreal had a proper `K/V` API?
> > > > 
> > > > 
> > > > I couldn't find it in the Rust SDK. Actually, I'll check whether we can access it by using queries.
> > > 
> > > 
> > > I see this: https://github.com/surrealdb/surrealkv But maybe is a different db implementation? or is it an API layer on top? That is what we probably need to know though.
> > 
> > 
> > That was what I was checking out. So, all of `surrealdb` is an abstraction on top of a key-value engine. For that key-value engine you could use `rocksdb` (and a couple other supported engines), or leverage the surreal team's own key-value engine, `surrealkv`.
> 
> Wait, surrealdb uses a kv and the underlaying db? or just the surrealkv?

Surrealdb stores data using a kv engine. Currently supported ones are: rocksdb, indexdb, surrealkv (their own key value engine), and probably some others.

nomos-services/storage/src/backends/mod.rs
AlejandroCabeza: Oh, I meant: Is there a purpose in using the library name for the feature name?E.g.: `sled` for the feature, instead of and `sled-backend`

nomos-services/storage/src/backends/surrealdb.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+impl SurrealBackendSettings {
+    pub fn new(db_path: String, namespace: String, database: String) -> Self {
+        Self {
+            db_path,
+            namespace,
+            database,
+        }
+    }
```
danielSanchezQ, line 49: Nipick. You can actually make it so it can consume other stuff into the string itself.
```suggestion
    pub fn new(db_path: impl Into<String>, namespace: impl Into<String>, database: impl Into<String>) -> Self {
        Self {
            db_path: db_path.into(),
            namespace: namespace.into(),
            database: database.into(),
        }
    }
```

That way later on you do not need to do it manually: `String::from`.

nomos-services/storage/src/backends/surrealdb.rs
```diff
@@ -0,0 +1,171 @@
...
+        let connect_to_surrealdb = async {
+            let db = connect(db_path).await?;
+            db.use_ns(namespace).use_db(database).await?;
+            Ok(Self {
+                db,
+                _serde_op: Default::default(),
+            })
+        };
+
+        block_in_place(|| Handle::current().block_on(connect_to_surrealdb))
```
danielSanchezQ, line 93: I would prefer to use the `OverwatchHandle` instead of tokio. The idea is that we can be as agnostic as possible from the runtime. There are methods that can help with this there.

nomos-services/storage/src/backends/surrealdb.rs
```diff
@@ -0,0 +1,171 @@
...
+        let key = from_utf8(&*key)?;
+        let data = Data::new(value);
+        let _record: Option<Data> = self.db.create(("kv", key)).content(data).await?;
+        Ok(())
+    }
+
+    async fn load(&mut self, key: &[u8]) -> Result<Option<Bytes>, Self::Error> {
+        let key = from_utf8(key)?;
+        let record: Option<Data> = self.db.select(("kv", key)).await?;
+        Ok(record.map(|data| data.to_bytes()))
```
danielSanchezQ, line 106: Weird that `clippy` does not throw a warning for this:
 ```suggestion
        Ok(record.map(Data::to_bytes))
```

File: nomos-services/storage/src/backends/surrealdb.rs [hunk 7]
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    async fn load_prefix(&mut self, prefix: &[u8]) -> Result<Vec<Bytes>, Self::Error> {
+        let prefix = from_utf8(prefix)?;
+        let query =
+            format!("SELECT * FROM kv WHERE string::starts_with(record::id(id), \"{prefix}\")");
+        let mut response = self.db.query(query).await?;
+        let records: Vec<Data> = response.take(0)?;
+        let records = records.into_iter().map(|data| data.to_bytes()).collect();
```
danielSanchezQ, line 115: Same `Data::to_bytes`

nomos-services/storage/src/backends/surrealdb.rs
```diff
@@ -0,0 +1,171 @@
...
+        Ok(records)
+    }
+
+    async fn remove(&mut self, key: &[u8]) -> Result<Option<Bytes>, Self::Error> {
+        let key = from_utf8(key)?;
+        let record: Option<Data> = self.db.delete(("kv", key)).await?;
+        Ok(record.map(|data| data.to_bytes()))
```
danielSanchezQ, line 122: Yet another

nomos-services/storage/src/backends/surrealdb.rs
```diff
@@ -0,0 +1,171 @@
+use crate::backends::{StorageBackend, StorageSerde, StorageTransaction};
+use async_trait::async_trait;
+use bytes::Bytes;
+use serde::{Deserialize, Serialize};
+use std::marker::PhantomData;
+use std::str::from_utf8;
+use surrealdb::engine::any::{connect, Any};
+use surrealdb::Surreal;
+use tokio::runtime::Handle;
+use tokio::task::block_in_place;
```
danielSanchezQ, line 10: We use (or try to) some structure for imports:

```
// std
std...
// crates
crates.io crate
// internal
anything that is in-house
```

danielSanchezQ (commented):
A few nitpicks.

nomos-services/storage/src/backends/surrealdb.rs
AlejandroCabeza: Cool, thanks :)

nomos-services/storage/src/backends/surrealdb.rs
AlejandroCabeza: Sounds great to me, didn't realise.

nomos-services/storage/src/backends/mod.rs
danielSanchezQ: Probably because its shorter. If you want to use the `-backend` suffix then better to rename all the other features.

AlejandroCabeza: Oh, okay, I thought there might be some special behaviour, or that it might interfere somehow. I'll stick to standards and go with just `surrealdb` then.

PR #928, opened 2024-12-02
Mix: Add connection maintenance

youngjoon-lee:

For spam protection and resiliency, the Tier 1 design contains the [Connection Maintenance](https://www.notion.so/Nomos-Blend-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#11c8f96fb65c806db454e80add388ab7), which measures [frequencies](https://www.notion.so/Nomos-Blend-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#11f8f96fb65c8093a228d6448e41fa7e) of several incoming messages, and close/open connections to maintain the reasonable quality of delivery. 
How frequencies are measured is defined in the [Frequencies Measurement](https://www.notion.so/Nomos-Blend-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#11c8f96fb65c80f5b047eade5dae58cd). How to close/open connections is defined in [here](https://www.notion.so/Nomos-Blend-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#11c8f96fb65c8018b6c6f7724c3fc494). 

This PR implements all mentioned above.

Plus, this implementation provides an option to disable it. Ideally, all nodes should enable connection maintenance. However, we are not sure what much frequency we should expect yet. If we set the threshold too high or small, the protocol would close too many healthy connections than intended. The estimation is not simple even though we have the [persistent transmission](https://www.notion.so/Nomos-Blend-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#11f8f96fb65c80dfbd23e4400feaaf9c), because of the [immediate forwarding](https://www.notion.so/Nomos-Blend-Network-Tier-1-Persistent-Transmission-Module-10b8f96fb65c807cb1e8f92a7f41a771?pvs=4#13d8f96fb65c8034ac47fca992242988).
I will run tests soon with multiple (virtual) nodes to measure the actual frequencies, so that we can provide the recommended parameters to set for the connection maintenance.

Yes

@youngjoon-lee @madxor 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 1. Description added.

### DIFF
--- a/nomos-mix/core/src/conn_maintenance.rs
+++ b/nomos-mix/core/src/conn_maintenance.rs
@@ -4,6 +4,7 @@ use std::{
     time::Duration,
 };
 
+use fixed::types::U57F7;
 use multiaddr::Multiaddr;
 use nomos_mix_message::MixMessage;
 use rand::RngCore;
@@ -84,7 +85,16 @@ where
     pub fn record_effective_message(&mut self, peer: &Multiaddr) {
         if let Some(monitors) = self.monitors.as_mut() {
             if let Some(monitor) = monitors.get_mut(peer) {
-                monitor.effective_messages += 1;
+                monitor.effective_messages = monitor
+                    .effective_messages
+                    .checked_add(U57F7::ONE)
+                    .unwrap_or_else(|| {
+                        tracing::warn!(
+                            "Skipping recording an effective message due to overflow: Peer:{:?}",
+                            peer
+                        );
+                        monitor.effective_messages
+                    });
             }
         }
     }
@@ -95,7 +105,16 @@ where
     pub fn record_drop_message(&mut self, peer: &Multiaddr) {
         if let Some(monitors) = self.monitors.as_mut() {
             if let Some(monitor) = monitors.get_mut(peer) {
-                monitor.drop_messages += 1;
+                monitor.drop_messages = monitor
+                    .drop_messages
+                    .checked_add(U57F7::ONE)
+                    .unwrap_or_else(|| {
+                        tracing::warn!(
+                            "Skipping recording a drop message due to overflow: Peer:{:?}",
+                            peer
+                        );
+                        monitor.drop_messages
+                    });
             }
         }
     }
@@ -192,8 +211,8 @@ where
 /// Meter to count the number of effective and drop messages sent by a peer
 #[derive(Debug)]
 struct ConnectionMonitor {
-    effective_messages: usize,
-    drop_messages: usize,
+    effective_messages: U57F7,
+    drop_messages: U57F7,
 }
 
 #[derive(Clone, Copy, Debug, Serialize, Deserialize)]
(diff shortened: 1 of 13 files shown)


nomos-services/mix/src/backends/libp2p.rs
```diff
@@ -182,6 +200,22 @@ impl MixSwarm {
...
+                );
+                let successful_dials = self.dial_to_peers(amount, Some(&excludes));
+                if successful_dials < amount {
+                    tracing::warn!(
+                        "Tried to establish {} new connections, but only {} succeeded",
+                        amount,
+                        successful_dials
+                    );
+                }
+            }
```
danielSanchezQ, line 218: I think this is the wrong approach. Mix behaviour should not relay on external swarm to interpret what needs to be done. In this case I think it should generate the dialing events internally. Basically removing the `EstablishNewConnections` event and handle it internally in the behaviour itself.

nomos-mix/network/src/behaviour.rs
```diff
@@ -11,21 +12,32 @@ use libp2p::{
...
     negotiated_peers: HashMap<PeerId, HashSet<ConnectionId>>,
+    /// Connection maintenance
+    /// NOTE: For now, this is optional because this may close too many connections
+    /// until we clearly figure out the optimal parameters.
+    conn_maintenance: Option<ConnectionMaintenance<PeerId>>,
+    /// Peers that should be excluded from connection establishments
+    blacklist_peers: HashSet<PeerId>,
+    /// To maintain address of peers because libp2p API uses only [`PeerId`] when handling
+    /// connection events. But, we need to know [`Multiaddr`] of peers to dial to them.
+    peer_addresses: HashMap<PeerId, Multiaddr>,
```
danielSanchezQ, line 40: I think all this should be part of some abstracted tool instead of something embedded completely in the behaviour itself.

nomos-mix/network/src/behaviour.rs
```diff
@@ -11,21 +12,32 @@ use libp2p::{
...
+    pin::pin,
     task::{Context, Poll, Waker},
 };
 
 /// A [`NetworkBehaviour`]:
 /// - forwards messages to all connected peers with deduplication.
 /// - receives messages from all connected peers.
-pub struct Behaviour<M> {
-    config: Config,
+pub struct Behaviour<M, Interval> {
```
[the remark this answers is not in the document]
youngjoon-lee: It's a measure of time. I tried to make it compatible for both async and sync environment (simulation), as we've done for other modules. 
In async, we'll use tokio::time::IntervalStream, but in simulation, we'll use other approach.

nomos-mix/network/src/behaviour.rs
```diff
@@ -127,18 +163,118 @@ where
...
+            .map(|peer_id| {
+                self.blacklist_peers.insert(peer_id);
+                if self.is_negotiated_peer(&peer_id) {
+                    1
+                } else {
+                    0
+                }
+            })
+            .sum()
+    }
```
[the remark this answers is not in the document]
youngjoon-lee: I've thought about it but actually it's not easy to abstract this clearly into the `nomos-mix` crate, because the user can establish/close connections in the libp2p swarm level, which means that it is hard for the `nomos-mix` crate to calculate exactly how many new connections should be established.
But anyway, I'll try to refactor this.;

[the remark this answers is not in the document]
youngjoon-lee: I've thought so at the beginning, but it was not easy to abstract it into the `nomos-mix/core`. Connection-related code (not even for connection maintenance) is scattered across mix-service and mix-behaviour. And, they are conceptually coupled with each other closely. 
However, I just found that the specification has been restructured, and now all connection-related logic is defined under the connection maintenance part. Then, I can abstract all connection-related code into `noms-mix/core`, and make mix-service and mix-behaviour thinner.
For that, there is something to be more clarified in the specification. I will discuss with Marcin first and modify this PR.

nomos-services/mix/src/backends/libp2p.rs
```diff
@@ -44,38 +43,26 @@ const CHANNEL_SIZE: usize = 64;
...
+            rng,
             swarm_message_receiver,
             incoming_message_sender.clone(),
         );
 
-        swarm
-            .listen_on(config.listening_address)
-            .unwrap_or_else(|e| {
-                panic!("Failed to listen on Mix network: {e:?}");
-            });
```
youngjoon-lee, line 67: Now this is done in `MixSwarm::new`.

nomos-services/mix/src/backends/libp2p.rs
```diff
@@ -44,38 +43,26 @@ const CHANNEL_SIZE: usize = 64;
...
-
-        // Randomly select peering_degree number of peers, and dial to them
-        membership
-            .choose_remote_nodes(&mut rng, config.peering_degree)
-            .iter()
-            .for_each(|node| {
-                if let Err(e) = swarm.dial(node.address.clone()) {
-                    tracing::error!("failed to dial to {:?}: {:?}", node.address, e);
-                }
-            });
```
youngjoon-lee, line 77: Now this is done in `nomos-mix-network::Behaviour`.

nomos-mix/network/src/behaviour.rs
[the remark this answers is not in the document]
youngjoon-lee: @danielSanchezQ I refactored the entire code base and this PR is now ready for review. Now all connection maintenance logic is in the `nomos-mix` crate. The `nomos-mix-network` crate and `nomos-mix-service` crate are thin.

nomos-mix/network/src/behaviour.rs
```diff
@@ -115,82 +150,86 @@ where
...
+            }
+        });
+        // Schedule events to connect to peers
+        peers_to_connect.into_iter().for_each(|addr| {
+            self.events.push_back(ToSwarm::Dial {
+                opts: DialOpts::from(addr),
+            });
+        });
+    }
+
```
danielSanchezQ, line 178: Yes! this is exactly what I mean. I think this is a better way to do this. Thanks!

danielSanchezQ (approved):
i think it is much clearer now! Thanks for the extra effort!

nomos-mix/core/src/conn_maintenance.rs
```diff
@@ -0,0 +1,410 @@
...
+    /// The number of effective (data or cover) messages that a peer is expected to send in a given time window.
+    /// If the measured count is greater than (expected * (1 + tolerance)), the peer is considered malicious.
+    /// If the measured count is less than (expected * (1 - tolerance)), the peer is considered unhealthy.
+    pub expected_effective_messages: f32,
+    pub effective_message_tolerance: f32,
+    /// The number of drop messages that a peer is expected to send in a given time window.
+    /// If the measured count is greater than (expected * (1 + tolerance)), the peer is considered malicious.
+    /// If the measured count is less than (expected * (1 - tolerance)), the peer is considered unhealthy.
+    pub expected_drop_messages: f32,
+    pub drop_message_tolerance: f32,
```
danielSanchezQ, line 212: Btw, I guess this values are nomalized. It would probably be a good idea to use fixed size precision: https://docs.rs/fixed/latest/fixed/

youngjoon-lee: Oh, thanks. I updated code by using fixed::U57F7 because I think it's good enough to have the max 0.99 for the fractional part. Also, I updated the code to handle overflow of U57F7 properly when incrementing counters.
I'm merging this PR now since I think the updated code is safe.

PR #933, opened 2024-12-05
feature(tracing): Bandwidth on Mixnet

AlejandroCabeza:
Add basic bandwidth tracing on Mixnet

Yes

@AlejandroCabeza 
@bacv 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 1. Description added.

### DIFF
--- a/nomos-mix/network/src/handler.rs
+++ b/nomos-mix/network/src/handler.rs
@@ -13,8 +13,23 @@ use libp2p::{
     },
     Stream, StreamProtocol,
 };
+use opentelemetry::{global, KeyValue};
 
 use crate::behaviour::Config;
+
+const SERVICE_NAME: &'static str = "nomos-mix/network/handler";
+const METRIC_DATA_RECEIVED: &'static str = "data_received";
+const METRIC_DATA_SENT: &'static str = "data_sent";
+const METRIC_FAILED_INBOUND_MESSAGES: &'static str = "failed_inbound_messages";
+const METRIC_SUCCESSFUL_INBOUND_MESSAGES: &'static str = "successful_inbound_messages";
+const METRIC_FAILED_OUTBOUND_MESSAGES: &'static str = "failed_outbound_messages";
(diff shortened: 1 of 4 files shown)


nomos-mix/network/src/handler.rs
```diff
@@ -95,6 +129,11 @@ impl ConnectionHandler for MixConnectionHandler {
     ) -> Poll<
         ConnectionHandlerEvent<Self::OutboundProtocol, Self::OutboundOpenInfo, Self::ToBehaviour>,
     > {
+        self.pending_outbound_messages_gauge
+            .record(self.outbound_msgs.len() as u64, &[]);
+        self.pending_events_to_behaviour_gauge
+            .record(self.pending_events_to_behaviour.len() as u64, &[]);
```
AlejandroCabeza, line 135: Is there any way to ensure there's not going to be data truncation here?

bacv:
Thanks for the effort, I see that you are creating a new global meter mixnet handler, and then use opentelemetry crate directly. This is a valid way to record metrics, but maybe you could try use the tracing crate for that if possible. 
In `nomos-tracing` crate, the metrics layer is created, which does the same what you did with the global meter, this meter is then accessable through calls to `tracing::info!` macro. You can see a basic example in https://github.com/logos-co/nomos-node/blob/master/nomos-services/mempool/src/da/service.rs#L155

AlejandroCabeza:
> Thanks for the effort, I see that you are creating a new global meter mixnet handler, and then use opentelemetry crate directly. This is a valid way to record metrics, but maybe you could try use the tracing crate for that if possible. In `nomos-tracing` crate, the metrics layer is created, which does the same what you did with the global meter, this meter is then accessable through calls to `tracing::info!` macro. You can see a basic example in https://github.com/logos-co/nomos-node/blob/master/nomos-services/mempool/src/da/service.rs#L155

I forgot to check out the tracing crate 🤦‍♀️ Thanks for pointing it out. Will do that and update all related tracing PRs.

PR #945, opened 2024-12-16
feature(da): Rust-based protobuf-like protocol

AlejandroCabeza:
Replaces protobuf with a Rust-based custom protocol.
Aims to simplify code and work around some annoying protobuf requirements.

Yes.

@AlejandroCabeza
@danielSanchezQ 
@bacv 

Yes

Yes, we are removing sessions wrappers from the messaging specification. Need to be done.

## Checklist

> [!WARNING]  

* [X] 1. Description added.


nomos-da/network/messages/src/packing.rs
```diff
@@ -0,0 +1,156 @@
...
+{
+    let serialized_message = wire::serialize(message).map_err(into_failed_to_serialize)?;
+
+    let data_length = serialized_message.len();
+    if data_length > MAX_MSG_LEN {
+        return Err(get_message_too_large_error(data_length));
+    }
+
+    let mut buffer = Vec::with_capacity(MAX_MSG_LEN_BYTES + data_length);
+    buffer.extend_from_slice(&(data_length as u16).to_be_bytes());
```
danielSanchezQ, line 51: You have a `LenType` type alias declared on top, why use u16 here?

nomos-da/network/messages/src/packing.rs
```diff
@@ -0,0 +1,156 @@
...
+
+pub async fn unpack_from_reader<Message, R>(reader: &mut R) -> Result<Message>
+where
+    Message: DeserializeOwned,
+    R: AsyncReadExt + Unpin,
+{
+    let data_length = read_data_length(reader).await?;
+    let mut data = vec![0u8; data_length];
+    reader.read_exact(&mut data).await?;
+    wire::deserialize(&data).map_err(into_failed_to_deserialize)
```
danielSanchezQ, line 74: For completeness of this library, we should probably just wrap this call into an `unpack` method (?)

nomos-da/network/messages/src/packing.rs
```diff
@@ -0,0 +1,156 @@
...
+    io::Error::new(
+        io::ErrorKind::InvalidData,
+        format!(
+            "Message too large. Maximum size is {}. Actual size is {}",
+            MAX_MSG_LEN, message_len
+        ),
+    )
+}
+
+pub fn pack<Message>(message: &Message) -> Result<Vec<u8>>
```
danielSanchezQ, line 39: Uhm, I'm not sure if should not go back to the trailing `_message` in this and below methods naming.

nomos-da/network/subnetworks-assignations/src/versions/v1.rs
```diff
@@ -42,7 +42,7 @@ impl FillFromNodeList {
 }
 
 impl MembershipHandler for FillFromNodeList {
-    type NetworkId = u32;
+    type NetworkId = u16;
```
danielSanchezQ, line 45: Don't we have a `SubnetworkId` type already for this? or does it mess circular dependencies?

danielSanchezQ (changes requested):
I think in general this was a good choice. Code is cleaner and clearer.

nomos-da/network/core/src/protocols/dispersal/executor/behaviour.rs
```diff
@@ -52,7 +51,7 @@ pub enum DispersalError {
     #[error("Dispersal response error: {error:?}")]
     Protocol {
         subnetwork_id: SubnetworkId,
-        error: DispersalErr,
+        error: dispersal::DispersalError,
```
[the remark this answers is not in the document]
AlejandroCabeza: Because there are some name conflicts. E.g.: There's two `DispersalError` being used. We need some way to differentiate them, that's why I decided to go with namespacing, although it doesn't make completely clear what they are.
Alternatives would be to add a `Message` suffix to all those types (which makes sense, as they indeed are messages), or to use a larger namespace (e.g.: `messages::dispersal::DispersalError`).

nomos-da/network/core/src/protocols/replication/mod.rs
```diff
@@ -3,20 +3,61 @@ pub mod handler;
...
+            rows_commitments: encoded_data.row_commitments.clone(),
+            rows_proofs: encoded_data
+                .rows_proofs
+                .iter()
+                .map(|proofs| proofs.get(index).cloned().unwrap())
+                .collect(),
+        };
+
+        da_blob
+    }
```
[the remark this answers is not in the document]
AlejandroCabeza: Should I create an issue for this?
Maybe do it now?

nomos-da/network/messages/src/common.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+
+#[repr(C)]
+#[derive(Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub enum CloseMessageReason {
+    GracefulShutdown = 0,
+    SubnetChange = 1,
+    SubnetSampleFailure = 2,
+}
```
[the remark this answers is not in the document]
AlejandroCabeza: Ref: https://rust-lang.github.io/unsafe-code-guidelines/layout/enums.html#layout-of-a-fieldless-enum
```
The `#[repr(C)]` annotation [...] selects the same size as the C compiler would use for the given target for an equivalent C-enum declaration.
```

Ref: https://doc.rust-lang.org/reference/type-layout.html#reprc-field-less-enums
```
For field-less enums, the C representation has the size and alignment of the default enum size and alignment for the target platform’s C ABI.
```

nomos-da/network/core/src/protocols/dispersal/executor/behaviour.rs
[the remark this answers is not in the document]
danielSanchezQ: I think it is ok for now tu use the direct path then. Thanks.

nomos-da/network/messages/src/common.rs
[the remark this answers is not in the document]
danielSanchezQ: Yeah, I think in this specific case it is better `repr(u8)` as all variants fits in that smaller size. But we still need to comply to c so `#[repr(C, u8)]`

nomos-da/network/messages/src/packing.rs
```diff
@@ -0,0 +1,156 @@
...
+        format!("Failed to serialize message: {}", error),
+    )
+}
+
+fn into_failed_to_deserialize(error: wire::Error) -> io::Error {
+    io::Error::new(
+        io::ErrorKind::InvalidData,
+        format!("Failed to deserialize message: {}", error),
+    )
+}
```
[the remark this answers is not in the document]
AlejandroCabeza: Doesn't let me, as `wire::Error` is an alias for `bincode::Error`; both types are foreign. Is there an appropriate workaround?

In the meantime I'll remove the duplicate.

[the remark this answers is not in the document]
danielSanchezQ: Yeah, probably then we would have to make `wire::Error` a wrapper type instead of a type alias. Something to be done in a different PR probably. Could you please open an issue for it as well?

nomos-da/network/messages/src/common.rs
```diff
@@ -0,0 +1,42 @@
...
+    pub fn new(reason: CloseMessageReason) -> Self {
+        CloseMessage { reason }
+    }
+}
+
+#[repr(C)]
+#[derive(Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub enum SessionRequest {
+    CloseMessage(CloseMessage),
+}
```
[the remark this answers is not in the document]
bacv: SessionReq is not used anywhere in implementation at the moment, the original idea was to use it for long lasting executor to subnetwork member connection, but this is handled by the libp2p now. I think it can be removed from spec and code.

nomos-da/network/messages/src/packing.rs
```diff
@@ -0,0 +1,156 @@
...
+
+fn get_message_too_large_error(message_len: usize) -> io::Error {
```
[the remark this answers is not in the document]
AlejandroCabeza: I hadn't thought of the type solution, I like it, very clean.
What would `inline` achieve here? Replace the function's body for the `new`'s body? 
Does that actually provide any benefit? That is: Doesn't the compiler do that automatically?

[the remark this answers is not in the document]
danielSanchezQ: > I hadn't thought of the type solution, I like it, very clean. What would `inline` achieve here? Replace the function's body for the `new`'s body? Does that actually provide any benefit? That is: Doesn't the compiler do that automatically?

Yep, the compiler may do it automatically (and probably will). But you hint it specifically for it to do it. Inlining is like removing the function call and instead the compiler will add the instructions wherever is called.

nomos-da/network/messages/src/packing.rs
AlejandroCabeza: Probably added the alias later, and forgot to update the references.

nomos-da/network/messages/src/packing.rs
AlejandroCabeza: I was tempted about doing that, but I felt the `unpack` would be the counterpart to `pack`. That means I'd expected it to just receive `bytes`,  not a `reader`.

This is probably the reason why the previous implementation followed this same naming scheme.

danielSanchezQ: No, I mean adding the unpack from bytes method and use it here. More or less this:

```rust
pub fn unpack<M: DeserializeOwned>(data: &[u8]) -> Result<M> {
    wire::deserialize(&data).map_err(into_failed_to_deserialize)
}
```

bacv (approved):
This looks really good, much simpler. I don't see any problems with it, it looks good to go after the comments are fixed.

nomos-da/network/messages/src/packing.rs
AlejandroCabeza: What's the benefit?

nomos-da/network/subnetworks-assignations/src/versions/v1.rs
AlejandroCabeza: Exactly.
I'll do the middle ground and create a `SubnetworkId = u16` as we did for messages.

nomos-da/network/core/src/protocols/dispersal/executor/behaviour.rs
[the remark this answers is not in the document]
AlejandroCabeza: Do you mean leave as is, or add the `messages` namespace as well?

nomos-da/network/messages/src/common.rs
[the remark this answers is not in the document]
AlejandroCabeza:
```
Combining a C and Int repr (e.g., #[repr(C, u8)]) is not permitted on a fieldless enum.
```

nomos-da/network/messages/src/packing.rs
AlejandroCabeza: Ah, yes, I thought of doing that and somehow I forgot. Thanks for pointing that out!

AlejandroCabeza: The problem with that approach is it doesn't make sense conceptually. Let me elaborate:
- The `pack` function returns [size + data] as bytes.
  - To me that means I'd expected any `unpack*` function to be analogous, that is, to reconstruct from [size + data].
- The `unpack_from_reader` consumes `SizeLen` bytes, interprets that as data size, retrieves that amount of bytes as data, and then deserializes them.
- I'd expected an `unpack` function to be identical to `unpack_from_reader` but receiving `bytes` instead of a parameter. Meaning: Receiving [size + data]  as bytes.

We wouldn't be able to reuse that `unpack` function inside the `unpack_from_reader`:
- The `unpack_from_reader` would need to do something like:
  ```
  let bytes = reader.read_all()
  unpack(bytes)
  ```
  Which is not possible because we need to read an specific amount of bytes.
- Also, the `unpack` function would need to receive only [data] bytes, which feels slightly counter-intuitive to me, as previously explained.

That being said, I'm not opposed to doing it, just stating my POV :)

nomos-da/network/core/src/protocols/dispersal/executor/behaviour.rs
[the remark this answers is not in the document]
danielSanchezQ: Leave it as it is yes.

nomos-da/network/messages/src/common.rs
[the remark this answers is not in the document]
danielSanchezQ: > ```
> Combining a C and Int repr (e.g., #[repr(C, u8)]) is not permitted on a fieldless enum.
> ```

Just `repr(u8)` then. It is more compact and since this goes on wire its better.

nomos-da/network/messages/src/packing.rs
danielSanchezQ: it is more explicit on where to use it (?). As Message type is really anything that can be serializable, although the type name is a hint 🤔 . Ok, we can leave it for now I guess.

nomos-da/network/subnetworks-assignations/src/versions/v1.rs
danielSanchezQ: Or maybe is a hint that this implementation could be generic 🤔 
I actually think it is ok to have the `u16` plain here. As it is an specific implementation for whatever indexes that are `u16`, doesn't matter what name you give it. I think that was my reasoning to use the previous (`u32`) type anyway. So dismiss this comments.

AlejandroCabeza: Shall I revert the crate-private `SubnetworkId = u16` I did?

nomos-da/network/messages/src/common.rs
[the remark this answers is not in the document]
AlejandroCabeza: I couldn't find the specs to edit them. Could someone point me to them?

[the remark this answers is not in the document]
bacv: https://www.notion.so/NomosDA-Network-Specification-c6664294d630470ba20aefb21a218f8c#228627887181403c8878e747cf748f1a

SessionReq didn't make it to the spec, it's only in the PoC, so no need to change it there. But we still have Protobuf related comment that I'll remove, the can mark this comment as resolved.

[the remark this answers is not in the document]
AlejandroCabeza: Alright! 
Couldn't find the protobuf mention so I'll assume you already deleted it. 
Resolving.

nomos-da/network/messages/src/packing.rs
danielSanchezQ: I get what you mean. Ideally `pack` <=> `unpack` should be bijective, pack receives a message and returns bytes, unpack receives bytes returns a message. Maybe the correct change actually is to remove pack and have a `pack_to_writer<M, W>(message: &M, w: &mut W) -> Result<(), io::Error> where W: AsyncWriter`. Instead to fit the API in a single direction. wdyt?

nomos-da/network/subnetworks-assignations/src/versions/v1.rs
danielSanchezQ: > Shall I revert the crate-private `SubnetworkId = u16` I did?

Yeah, I think so. Sorry for the misslead.

AlejandroCabeza: No worries :)

nomos-da/network/messages/src/packing.rs
AlejandroCabeza: Oh, that actually makes sense and makes everything bijective. Good stuff.

PR #950, opened 2024-12-18
feature(wire): Wrap errors

AlejandroCabeza:
Create standalone Error type for wire and wrap errors with it.
Closes: https://github.com/logos-co/nomos-node/issues/949

Yes

@AlejandroCabeza 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 1. Description added.
* [X 5. Link PR to a specific milestone.


nomos-core/chain-defs/src/wire_errors.rs
```diff
@@ -0,0 +1,16 @@
+//STD
+use std::io;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Failed to serialize message: {0}")]
+    Serialize(bincode::Error),
+    #[error("Failed to deserialize message: {0}")]
+    Deserialize(bincode::Error),
+}
```
danielSanchezQ, line 10: Usually unless it is too convoluted this could live in wire. why? it is usually easier to import everything from the same module:
```rust
use wire::{serialize, deserialize, Error};
```

danielSanchezQ (approved):
Added a nitpick, I think it would be better that way. But having a separated module is ok.
Alternatively you may reexport the module in `wire` directly:
```rust
pub use crate::wire_errors::Error;
```

AlejandroCabeza:
> Added a nitpick, I think it would be better that way. But having a separated module is ok. Alternatively you may reexport the module in `wire` directly:
> 
> ```rust
> pub use crate::wire_errors::Error;
> ```

I generally favour splitting stuff into files, but I do very much agree having `Error` exported from the same file would be ideal.
For the sake of arguing: How does refactoring `wire.rs` into a `wire` module sound?
All this being said, I can certainly put `Error` back in`wire.rs`.

danielSanchezQ:
> > Added a nitpick, I think it would be better that way. But having a separated module is ok. Alternatively you may reexport the module in `wire` directly:
> > ```rust
> > pub use crate::wire_errors::Error;
> > ```
> 
> I generally favour splitting stuff into files, but I do very much agree having `Error` exported from the same file would be ideal. For the sake of arguing: How does refactoring `wire.rs` into a `wire` module sound? All this being said, I can certainly put `Error` back in`wire.rs`.

Yeah, wire could be its own module actually, then it would have the mod with the implementation and an error.rs file. Then reexport the Error from the mod. That would tick all the boxes.

nomos-core/chain-defs/src/wire/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+pub use serialization::{serialize, serializer, serializer_into_buffer};
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serde::{Deserialize, Serialize};
+
+    #[test]
+    fn ser_de() {
```
danielSanchezQ, line 21: Is this serialize - deserialize? I think splicit naming should be better.

danielSanchezQ (approved):

PR #952, opened 2024-12-18
feature(encoder): Implement Iterator over DaBlob for EncodedData

AlejandroCabeza:
Implement Iterator over DaBlob for EncodedData
Closes: https://github.com/logos-co/nomos-node/issues/943

Yes

@AlejandroCabeza 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 1. Description added.

### DIFF
--- a/nomos-da/kzgrs-backend/src/encoder.rs
+++ b/nomos-da/kzgrs-backend/src/encoder.rs
@@ -10,10 +10,11 @@ use kzgrs::{
     bytes_to_polynomial, commit_polynomial, encode, Commitment, Evaluations, GlobalParameters,
     KzgRsError, Polynomial, PolynomialEvaluationDomain, Proof, BYTES_PER_FIELD_ELEMENT,
 };
+
 #[cfg(feature = "parallel")]
 use rayon::iter::{IntoParallelRefIterator, ParallelIterator};
-
 // internal
+use crate::common::blob::DaBlob;
 use crate::common::{hash_column_and_commitment, Chunk, ChunksMatrix, Row};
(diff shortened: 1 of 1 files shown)


nomos-da/kzgrs-backend/src/encoder.rs
```diff
@@ -55,6 +56,65 @@ pub struct EncodedData {
     pub column_commitments: Vec<Commitment>,
     pub aggregated_column_commitment: Commitment,
     pub aggregated_column_proofs: Vec<Proof>,
+
+    /// Iterator State
+    current_column_idx: Option<usize>,
```
danielSanchezQ, line 61: This is not the way to go. Usually what you want to do is to implement `IntoIterator` then return a new struct that handles this state and have a reference to the original Encoded data and that struct implements `Iterator`.

danielSanchezQ (changes requested):
Implementation looks ok, but having an iteration state inside that structure is wrong. What happens if I need 5 iterators over it? You will need to add 5 states then...

nomos-da/kzgrs-backend/src/encoder.rs
AlejandroCabeza: Oh, right! This actually makes waaaay more sense. I didn't have a great feeling about it, but as everyone was suggesting doing it this way I assumed it was the standard.

Thanks for pointing this out! Updating

danielSanchezQ (approved):
Looking good, thanks!
We can use this in the executor now. Probably something to do when merging this. Please add an issue and attach to the ongoing iteration.

PR #957, opened 2024-12-19
fix(blend): correctly check the last layer in `MockBlendMessage`

### LINKED ISSUE #956: Fix `MockBlendMessage::unwrap` to correctly check if the current layer is the last
In the current `MockBlendMessage`, we determine that the current layer is the last if the next node ID is `[0u8; 32]`. However, users can use `[0u8; 32]` as an actual node ID. If users use the `[0u8; 32]` as the 2nd node ID, then the implementation determines that the 1st layer is the last. This is a bug.

`MockBlendMessage` is used in simulations.

youngjoon-lee:

Closes #956

As explained in #956, this PR fixes `MockBlendMessage` to correctly check if the current layer is the last. Previously, it has been used `[0u8; 32]` to represent a dummy node ID, but it is common that users use the `[0u8; 32]` as an actual node ID. Then, the check logic fails. So, this PR uses `[u8::MAX; 32]` as a dummy node ID, and return error if users use it as an actual node ID. I think this fix is good enough because it's just a mock (used in simulations).

Yes

@youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 1. Description added.

### DIFF
--- a/nomos-blend/message/src/mock/error.rs
+++ b/nomos-blend/message/src/mock/error.rs
@@ -1,4 +1,4 @@
-#[derive(thiserror::Error, Debug)]
+#[derive(thiserror::Error, Debug, PartialEq)]
 pub enum Error {
     #[error("Invalid blend message format")]
     InvalidBlendMessage,
@@ -6,6 +6,8 @@ pub enum Error {
     PayloadTooLarge,
     #[error("Invalid number of layers")]
     InvalidNumberOfLayers,
+    #[error("Invalid public key")]
+    InvalidPublicKey,
     #[error("Unwrapping a message is not allowed to this node")]
     /// e.g. the message cannot be unwrapped using the private key provided
     MsgUnwrapNotAllowed,
--- a/nomos-blend/message/src/mock/mod.rs
+++ b/nomos-blend/message/src/mock/mod.rs
@@ -1,7 +1,5 @@
 pub mod error;
 
-use std::u8;
-
 use error::Error;
 
 use crate::BlendMessage;
--- a/nomos-blend/message/src/mock/mod.rs
+++ b/nomos-blend/message/src/mock/mod.rs
@@ -1,5 +1,7 @@
 pub mod error;
 
+use std::u8;
+
 use error::Error;
 
 use crate::BlendMessage;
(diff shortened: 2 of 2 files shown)


nomos-blend/message/src/mock/error.rs
```diff
@@ -1,11 +1,13 @@
...
+#[derive(thiserror::Error, Debug, PartialEq)]
 pub enum Error {
     #[error("Invalid blend message format")]
     InvalidBlendMessage,
     #[error("Payload is too large")]
     PayloadTooLarge,
     #[error("Invalid number of layers")]
     InvalidNumberOfLayers,
+    #[error("Invalid public key")]
+    InvalidPublicKey,
```
youngjoon-lee, line 10: Instead of `InvalidNodeId`, I use this name because the trait `BlendMessage` uses the name `PublicKey` (not `NodeId`). In the `MockBlendMessage`, a node ID is used as a public key, but I think the error should use the generic term: public key.

danielSanchezQ:
wdym it is common that users use `[0u8; 32]` as NodeId? I think it is actually the same case as `u8::MAX`.

youngjoon-lee:
> wdym it is common that users use `[0u8; 32]` as NodeId? I think it is actually the same case as `u8::MAX`.

The simulation code generates node IDs from 0 to N. We can avoid this bug by generating node IDs from 1 or larger, but I wanted to say that people will tend to generate node IDs from 0 when they use this mock message implementation.
So, I used u8::MAX for the dummy. Of course, users can use u8::MAX but then this mock returns an error. I think this is more natural.

danielSanchezQ:
> > wdym it is common that users use `[0u8; 32]` as NodeId? I think it is actually the same case as `u8::MAX`.
> 
> The simulation code generates node IDs from 0 to N. We can avoid this bug by generating node IDs from 1 or larger, but I wanted to say that people will tend to generate node IDs from 0 when they use this mock message implementation. So, I used u8::MAX for the dummy. Of course, users can use u8::MAX but then this mock returns an error. I think this is more natural.

Ah that makes sense now. I would say it really is our fault as we could use random ids in the sims, but using integers is handier for debugging though. 
I understand it now. Thanks.

danielSanchezQ (approved):

PR #975, opened 2024-12-27
feature(node-recovery): Implement Overwatch recovery POC

AlejandroCabeza:
This is a POC for a reusable Recovery system on top of the current Overwatch, without making any modification.
I'd love to get feedback for this as it's more of an initial proposal than anything. 
Also important, I suggest reading the context below.

First step for https://github.com/logos-co/nomos-node/issues/960

*Additional context:*
In order to implement this I had to leverage two existing mechanisms:
- `ServiceState::from_settings` for loading on startup
- `StateOperator::run` for saving every time the state changes

On the one hand, it works.
On the other, it feels way more convoluted than it should. It's usable, but it does require more explaining than it should, which is certainly less than ideal.

I'd argue the best solution would be to offer the Recovery mechanism from within Overwatch:
1. It'd strongly increase understandability. It would be equivalent to the any of the `ServiceCore`'s associated types (e.g. `type State`) which can have a `ServiceState` or a `NoState`. For instance: `type RecoveryOperator` being either an implementation of `trait ServiceRecovery` or an already implemented `NoRecovery`.
2. Doing it this way means the state initialisation (using recovery) wouldn't need to done from `ServiceState::from_settings`. That would decouple the `ServiceState` from the `RecoveryBackend` which makes a lot of sense.
3. Additionally, this means the `RecoveryBackend` would be only used in one place which would be much clearer. E.g.: `type Recovery = RecoveryOperator<Backend>`.
4. Finally, this means leaving the `StateOperator` type unassigned, so that any other functionality could be added in there. In this solution we're leveraging it for recovery; so we couldn't be able to use it for any other task unless we merge them in the same operator.

@AlejandroCabeza 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 1. Description added.

### DIFF
--- a/nomos-utils/src/overwatch_recovery.rs
+++ b/nomos-utils/src/overwatch_recovery.rs
@@ -1,12 +1,10 @@
 // STD
 use std::fmt::Debug;
 use std::io;
-use std::marker::PhantomData;
 // Crates
 use log::error;
 use overwatch_rs::services::state::{ServiceState, StateOperator};
-use serde::de::DeserializeOwned;
-use serde::{Deserialize, Serialize};
+use serde::{de::DeserializeOwned, Serialize};
 
 trait FromSettings {
     type Settings;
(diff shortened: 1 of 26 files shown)


nomos-utils/src/overwatch_recovery.rs
```diff
@@ -0,0 +1,286 @@
...
+    Backend::State: Serialize + DeserializeOwned + Debug + Clone,
+{
+    fn new(state: <Backend as RecoveryBackend>::State) -> Self {
+        Self {
+            state,
+            settings: Default::default(),
+            backend: Default::default(),
+        }
+    }
+}
```
danielSanchezQ, line 52: Is this even needed? Also how does settings and backend are not bind to Default, but this still works 🤔 ?

nomos-utils/src/overwatch_recovery.rs
```diff
@@ -0,0 +1,286 @@
...
+    use overwatch_rs::services::{ServiceCore, ServiceData, ServiceId};
+    use overwatch_rs::DynError;
+    use std::env::temp_dir;
+    use std::path::PathBuf;
+
+    #[derive(Debug, Clone)]
+    struct FileBackend<State> {
+        directory: PathBuf,
+        state: PhantomData<State>,
+    }
```
danielSanchezQ, line 153: This can probably be promoted as part of the utils module itself. It would need some tweaks. The generic is not really necessary here I think. It can be part of the usage methods as long as the type implements `Deserialize`.
For an idea of a fn signature: `fn load<T: DeserializeOwned>(&self) -> T`

nomos-utils/src/overwatch_recovery.rs
AlejandroCabeza: The compiler didn't complain about the `Default` trait for them.
Well, I was pretty sure I was using them somewhere but maybe I removed the usage.

nomos-utils/src/overwatch_recovery.rs
AlejandroCabeza: I thought so as well but I didn't split anything into files yet as I thought of this as a proposal.

nomos-utils/src/overwatch_recovery.rs
```diff
@@ -0,0 +1,226 @@
...
+
+            service_state_handle.overwatch_handle.shutdown().await;
+            Ok(())
+        }
+    }
+
+    #[derive(Services)]
+    pub struct RecoveryTest {
+        recovery: ServiceHandle<Recovery>,
+    }
```
youngjoon-lee, line 200: If I understand this correctly, do we need to have a recovery service for each existing service (DA, Network, Cryptarchia, etc.), as below?
```rust
#[derive(Services)]
pub struct Nomos {
  cryptarchia: ServiceHandle<NodeCryptarchia>,
  cryptarchia_recovery: ServiceHandle<CryptarchiaRecovery>,
  network: ServiceHandle<NetworkService<NetworkBackend>>,
  network_recovery: ServiceHandle<NetworkRecovery>,
  ..
}
```
I guess so because we need a distinct `RecoveryBackend` for each existing service that has its own `State` type (which should be matched with `RecoveryBackend::State`). 

In my opinion, your suggestion in the PR description looks cleaner. Each service can have its own `RecoveryOperator`, which can be implemented in various way using a proper backend for the service. Also, as you said, it sounds good that we can leave the `StateOperator` for other uses. 
I think we may be able to add this concept to the Overwatch but make it as generic/simple as possible, and let services implement details. At the same time, I also understand the Daniel's point. For example, I guess we can implement only the mechanism that notified state updates to the registered `RecoveryOperator` in the Overwatch level, because `ServiceStateHandle::state_updater` knows when the state is being updated. But, the rest could be done in the service level probably by implementing `RecoveryOperator`. I think that each `RecoveryOperator` can choose the interval of calling `save_state` if necessary, for example. 
This is not a strong opinion. This may be missing some points.

AlejandroCabeza: I'll try to address all the points. Let me know if I miss something:
1. I don't know if _every_ service needs it, but I'm pretty positive a few of them do.
2. @danielSanchezQ told me the `Operator`'s original function intention is to serve as a recovery mechanism, so it seems that we don't need an extra entity for this.
3. I also suggested @danielSanchezQ adding this concept (state recovery) to `Overwatch` itself, but he told me the idea is to keep it as barebones as possible.
4. In the beginning I thought services would want to save their state every time the `ServiceStateHandle::state_updater` is called, but you make a good point: It could be interesting to delegate the `save_state` calling interval to the service itself.
    - (CC: @danielSanchezQ) Is this something we should consider?

youngjoon-lee: All the points you addressed make sense to me. Thanks.
Regarding the interval, I mentioned it as an example. I don't think it's really necessary right now.

danielSanchezQ (commented):
This starts looking good. I would move the operators to `nomos-sercices/utils`. Also it looks like it has mixing commits? I see a lot of changes that shouldn't be shown (?)

danielSanchezQ (approved):

PR #977, opened 2025-01-06
feature(packaging): Proof of concept

AlejandroCabeza:
Implement packaging and distribution for main platforms (linux, macos and windows).

This PR includes:

1. **Bundling Code**
- Bundling-related utilities
- Bundling code for `nomos-node`. It servers as template for the other services we need to get bundled.
- Royalty-free icon files (required for the bundling).

2. **Github Actions**
- Install Risc0 action
- Install (build + bundle) dependencies action.
- Packaging workflow that produces bundles for `x86_64-linux`, `i686-linux` and `aarch64-osx`. Risc0 is not available in other OSs.

Closes: https://github.com/logos-co/nomos-node/issues/998
Closes: https://github.com/logos-co/nomos-node/issues/999

Yes

@AlejandroCabeza 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 1. Description added.


nomos-bundler/README.md
```diff
@@ -0,0 +1,16 @@
+# Nomos Bundler
+
+This crate performs the bundling tasks for other Nomos' crates.
+
+## Usage
+
+Any crate that needs to be bundled should have their own file in the `src` directory of this crate,
+with the same name as the bundled crate, and implement their bundling logic.
```
youngjoon-lee, line 8: > with the same name as the bundled crate

Considering our current directory structure: `nomos-bundler/nomos-cli/bundle.rs`,

I think it might be clearer to say like below. What do you think about this?

```suggestion
Any crate that needs to be bundled should have their own directory in the `src` directory of this crate,
with the same name as the bundled crate, and it should contain a file that implements their bundling logic.
```

nomos-bundler/nomos-cli/bundle.rs
```diff
@@ -0,0 +1,149 @@
...
+                resources_path
+                    .join("icons/512x512.png")
+                    .to_string_lossy()
+                    .to_string(),
+            ]),
+            resources: None,
+            resources_map: None,
+            copyright: None,
+            license: None,
+            license_file: None,
```
youngjoon-lee, line 83: Side topic: I just realized that we need to choose a proper open-source license soon.

.github/actions/install-risc0/action.yml
```diff
@@ -0,0 +1,67 @@
+name: "Install Risc0"
```
youngjoon-lee, line 1: It seems that this action is not used in the `workflows/packaging.yml`? Could you teach me where this action is executed from?

AlejandroCabeza: This one is called from the `install-dependencies` action. Do you think it's a better idea to have it as a separate step in the main workflow?

nomos-bundler/nomos-cli/bundle.rs
AlejandroCabeza: Indeed.

nomos-bundler/README.md
AlejandroCabeza: It's indeed much better phrased the way you put it.

That being said, you made me rethink if it's a good idea to be this specific: I feel like when working on the remaining bundles we will arrive at a "single bundling binary" solution. I don't really know, though.

youngjoon-lee: It would be good if we can have the single bundling binary somehow but I'm totally fine with the current approach implemented in this PR :)

.github/actions/install-risc0/action.yml
youngjoon-lee: Ah, I missed it, sorry :) No, the current workflow is good in my opinion.

youngjoon-lee (approved):
👍

danielSanchezQ (approved):

PR #981, opened 2025-01-10
Kms service

danielSanchezQ:

This PR implements the service layer for the [KMS (key management system)](https://www.notion.so/KMS-key-management-system-16f8f96fb65c8064a529cd0908294088).

This is a partial implementation but it is pretty straight forward to understand from the document. Notice that when implementing the backend we may need to modify things a bit (probably the SecureKey trait).

@youngjoon-lee @danielSanchezQ 

Yes

It does not.

## Checklist

> [!WARNING]  

* [x] 1. Description added.

### DIFF
--- a/nomos-services/key-management-system/src/backend/mod.rs
+++ b/nomos-services/key-management-system/src/backend/mod.rs
@@ -3,6 +3,7 @@ use bytes::Bytes;
 use either::Either;
 use overwatch_rs::DynError;
 
+#[async_trait::async_trait]
 pub trait KMSBackend {
     type SupportedKeys;
     type KeyId;
@@ -15,5 +16,5 @@ pub trait KMSBackend {
     ) -> Result<Self::KeyId, DynError>;
     fn public_key(&self, key_id: Self::KeyId) -> Result<Bytes, DynError>;
     fn sign(&self, key_id: Self::KeyId, data: Bytes) -> Result<Bytes, DynError>;
-    fn execute(&mut self, op: KMSOperator);
+    async fn execute(&mut self, op: KMSOperator);
 }
--- a/nomos-services/key-management-system/src/backend/mod.rs
+++ b/nomos-services/key-management-system/src/backend/mod.rs
@@ -1,0 +1,19 @@
+use crate::KMSOperator;
+use bytes::Bytes;
+use either::Either;
+use overwatch_rs::DynError;
+
+pub trait KMSBackend {
+    type SupportedKeys;
+    type KeyId;
+    type Settings;
+
+    fn new(settings: Self::Settings) -> Self;
+    fn register(
+        &mut self,
+        key: Either<Self::SupportedKeys, Self::KeyId>,
+    ) -> Result<Self::KeyId, DynError>;
+    fn public_key(&self, key_id: Self::KeyId) -> Result<Bytes, DynError>;
+    fn sign(&self, key_id: Self::KeyId, data: Bytes) -> Result<Bytes, DynError>;
+    fn execute(&mut self, op: KMSOperator);
+}
--- a/nomos-services/key-management-system/src/lib.rs
+++ b/nomos-services/key-management-system/src/lib.rs
@@ -1,4 +1,4 @@
-use crate::backends::KMSBackend;
+use crate::backend::KMSBackend;
 use crate::secure_key::SecuredKey;
 use bytes::Bytes;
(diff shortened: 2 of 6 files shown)


nomos-services/key-management-system/src/backend/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+#[async_trait::async_trait]
+pub trait KMSBackend {
+    type SupportedKeys;
+    type KeyId;
+    type Settings;
+
+    fn new(settings: Self::Settings) -> Self;
+    fn register(
+        &mut self,
+        key: Either<Self::SupportedKeys, Self::KeyId>,
```
youngjoon-lee, line 15: Question: If the `key` is `Either::Right(key_id)`, this function loads the key from its storage?

danielSanchezQ: Well, it depends on the backend. The idea is that all this needs to be reproducible. So if some system had a key, the node goes down then it is recovered, it should continue operations with the same key right?

nomos-services/key-management-system/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+use tokio::sync::oneshot;
+
+mod backend;
+mod secure_key;
+
+const KMS_TAG: ServiceId = "KMS";
+
+// TODO: Use [`AsyncFnMut`](https://doc.rust-lang.org/stable/std/ops/trait.AsyncFnMut.html#tymethod.async_call_mut) once it is stabilized.
+pub type KMSOperator = Box<
+    dyn FnMut(&mut dyn SecuredKey) -> Box<dyn Future<Output = Result<(), DynError>>> + Send + Sync,
```
youngjoon-lee, line 24: Do we expect that the `KMSBackend` returns the output of this operation through a channel? Is that why the `Future` returns `Ok(())`?
If so, I think that `KMSBackend::execute` doesn't need to be `async`, probably?

youngjoon-lee, line 24: I don't clearly understand how the caller of the `KMSMessage::Execute` receives an output. Do we probably need to add a `oneshot::Sender` to this `Fn`?

nomos-services/key-management-system/src/backend/mod.rs
youngjoon-lee: Yes. Hmm, then I'm thinking about whether we really need `Either::Right(key_id)`. When a node is restarted, the node can load all keys from its storage, so that the subsequent operations (e.g. `sign`, etc), can succeed.

youngjoon-lee: If we have `Either::Right(key_id)` in this API, and if the backend loads a key only when the user calls `register`, then the user must keep all `key_id`s that he needs in the storage, so that he can call the `register` manually when the node is restarted.

danielSanchezQ: > If we have `Either::Right(key_id)` in this API, and if the backend loads a key only when the user calls `register`, then the user must keep all `key_id`s that he needs in the storage, so that he can call the `register` manually when the node is restarted. If this is the intended behavior, I'm okay with having `Either::Right(key_id)`.

Yes, this is the intended behaviour. Also they may be loaded or generated. We do not know yet.

nomos-services/key-management-system/src/lib.rs
youngjoon-lee: Ah, right. That totally makes sense. Thanks. I'm resolving this comment.

youngjoon-lee (approved):

nomos-services/key-management-system/src/lib.rs
bacv, line 24: Wouldn't the caller be able to overwrite they `SecuredKey`? The trait doesn't have mutable methods, maybe this could be just `dyn SecuredKey`?

bacv (approved):

nomos-services/key-management-system/src/lib.rs
danielSanchezQ: Uhm, probably with some unsafe. Maybe we need to restric more the type and make it just `&dyn` but I have the feeling we would have to modify this type when implementing the backend. Also expand the `SecuredKey` trait. Let's see what specifics we need and how restrictive can we be.

PR #986, opened 2025-01-20
refactor(blend): enhance connection management structures

youngjoon-lee:

This PR composes 4 changes:

1. refactor(blend): add PeerId to Membership
2. #987
3. #988 
4. #989

A description for the first change can be found in the following section.

### refactor(blend): add PeerId to Membership
This PR serves as preliminary work for refactoring connection management: https://www.notion.so/Connection-Management-in-Libp2p-1708f96fb65c8019a6edd2a3b89ff8e2?pvs=4#1768f96fb65c80e58f68ea5300deba40.

The main change introduced here is the addition of a generic `NodeId` field to the Membership struct. In the case of `Libp2pNetworkBackend`, `libp2p::PeerId` will be used as the `NodeId`.

Previously, Membership relied solely on `Multiaddr`, but this approach was not well-suited to the libp2p concept. Since libp2p manages connections based on `PeerId` rather than `Multiaddr`, our codebase had to maintain a mapping between `Multiaddr` and `PeerId`. This mapping was both cumbersome and fragile because `Multiaddr` is not designed to serve as a unique identifier.

The core change in this PR is [this](https://github.com/logos-co/nomos-node/compare/blend-membership-node-id?expand=1#diff-4b51a6ac8ff54fd413df6709f5c738fe30289317c93b7ff06b87ddc59ac34c23R21). Other relevant parts of the codebase have been updated accordingly to reflect this adjustment.

It's important to note that the newly introduced `NodeId` is not utilized anywhere in the current implementation. Its functionality will be leveraged in the next PR, which will significantly simplify connection management. I don't want to introduce any change in this PR that is going to be removed in the next PR.

@youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [x] 1. Description added.


danielSanchezQ (approved):
This makes sense. It may have to change later when the RDP format and service is implemented (probably just tying up the types).

AlejandroCabeza (approved):

PR #989, opened 2025-01-21
refactor(blend): establish new connections, or deny new streams

youngjoon-lee:

This PR is the last one of refactoring connection management.
https://www.notion.so/Connection-Management-in-Libp2p-1708f96fb65c8019a6edd2a3b89ff8e2?pvs=4#1768f96fb65c80e58f68ea5300deba40

This PR adds the following actions:
- Dial new peers if needed.
- Deny a stream of a newly established connection in certain conditions.

The detailed logic is:
- `Behaviour` maintains three sets of peers (healthy, unhealthy, and malicious) based on the output of connection monitoring added in the PR #987.
- `Behaviour` determines whether new connections need to be established, based on the number of peers in those three sets. If so, the `Behaviour` emits an event to the `Backend`, so that the `Backend` can select random peers and request `Swarm` to dial them.
- `Behaviour` denies the stream of connections, if `max_peering_degree` reached or if the peer has been detected as malicious. It doesn't deny the connection itself (even though libp2p provides an API for that explicitly), because we want to make our behaviour/handler have control on only streams, not connections.

Why `Backend` selects random peers? not by `Behaviour`?
- We still don't know how to handle membership updates. I want to make membership live in the `Backend` because we may have more control and flexibility on the `Backend` rather than `Behaviour`.
- Also, someone who wants to use the `Behaviour` may have a different kind of membership. So, if possible, I want to decouple `Behaviour` with membership.

As I noted in Notion, I've found that we don't actually need to define our own ConnectionManager abstraction. However, please feel free to share your opinion if you think it's better to define the abstraction for both Blend and DA. I think, their requirements are quite different.

Yes

@youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [x] 1. Description added.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # BUILD IMAGE ---------------------------------------------------------
 
-FROM rust:1.82.0-slim-bookworm AS builder
+FROM rust:1.84.0-slim-bookworm AS builder
 
 WORKDIR /nomos
 COPY . . 
--- a/nomos-blend/network/Cargo.toml
+++ b/nomos-blend/network/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2021"
 cached = "0.53.1"
 futures = "0.3.30"
 futures-timer = "3.0.3"
-libp2p = "0.53"
+libp2p = "0.55"
 tracing = "0.1"
 nomos-blend = { path = "../core" }
 nomos-blend-message = { path = "../message" }
@@ -20,7 +20,7 @@ tokio-stream = { version = "0.1", optional = true }
 [dev-dependencies]
 tokio = { version = "1", features = ["macros", "rt-multi-thread", "time"] }
 tokio-stream = "0.1"
-libp2p = { version = "0.53", features = ["ed25519", "tokio", "quic"] }
+libp2p = { version = "0.55", features = ["ed25519", "tokio", "quic"] }
 tracing-subscriber = "0.3.18"
 fixed = "1"
 
--- a/nomos-blend/network/src/behaviour.rs
+++ b/nomos-blend/network/src/behaviour.rs
@@ -4,11 +4,10 @@ use crate::{
 };
 use cached::{Cached, TimedCache};
 use libp2p::{
-    core::Endpoint,
+    core::{transport::PortUse, Endpoint},
     swarm::{
-        behaviour::ConnectionEstablished, ConnectionClosed, ConnectionDenied, ConnectionId,
-        FromSwarm, NetworkBehaviour, NotifyHandler, THandler, THandlerInEvent, THandlerOutEvent,
-        ToSwarm,
+        ConnectionClosed, ConnectionDenied, ConnectionId, FromSwarm, NetworkBehaviour,
+        NotifyHandler, THandler, THandlerInEvent, THandlerOutEvent, ToSwarm,
     },
     Multiaddr, PeerId,
 };
@@ -16,7 +15,7 @@ use nomos_blend::conn_maintenance::{ConnectionMonitor, ConnectionMonitorSettings
 use nomos_blend_message::BlendMessage;
 use sha2::{Digest, Sha256};
 use std::{
-    collections::{HashSet, VecDeque},
+    collections::{HashMap, VecDeque},
     task::{Context, Poll, Waker},
 };
 use std::{marker::PhantomData, time::Duration};
@@ -30,7 +29,7 @@ where
     IntervalProvider: IntervalStreamProvider,
 {
     config: Config,
-    peers: HashSet<PeerId>,
+    negotiated_peers: HashMap<PeerId, NegotiatedPeerState>,
     /// Queue of events to yield to the swarm.
     events: VecDeque<ToSwarm<Event, FromBehaviour>>,
     /// Waker that handles polling
@@ -40,6 +39,12 @@ where
     duplicate_cache: TimedCache<Vec<u8>, ()>,
     _blend_message: PhantomData<M>,
     _interval_provider: PhantomData<IntervalProvider>,
+}
+
+#[derive(Debug, Eq, PartialEq)]
+enum NegotiatedPeerState {
+    Healthy,
+    Unhealthy,
 }
 
 #[derive(Debug)]
@@ -52,6 +57,10 @@ pub struct Config {
 pub enum Event {
     /// A message received from one of the peers.
     Message(Vec<u8>),
+    /// A peer has been detected as malicious.
+    MaliciousPeer(PeerId),
+    /// A peer has been detected as unhealthy.
+    UnhealthyPeer(PeerId),
     Error(Error),
 }
 
@@ -65,7 +74,7 @@ where
         let duplicate_cache = TimedCache::with_lifespan(config.duplicate_cache_lifespan);
         Self {
             config,
-            peers: HashSet::new(),
+            negotiated_peers: HashMap::new(),
             events: VecDeque::new(),
             waker: None,
             duplicate_cache,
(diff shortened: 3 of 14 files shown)


nomos-blend/network/src/behaviour.rs
```diff
@@ -187,10 +236,21 @@ where
...
+                connection_id,
+                ..
+            }) => {
+                if self.should_deny_stream(&peer_id) {
+                    // Notify the corresponding [`ConnectionHandler`] to close the stream.
+                    self.events.push_back(ToSwarm::NotifyHandler {
+                        peer_id,
+                        handler: NotifyHandler::One(connection_id),
+                        event: FromBehaviour::CloseSubstreams,
+                    });
```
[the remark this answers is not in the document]
youngjoon-lee: Since they don't have a proper documentation, I investigated their code.

This `ConnectionEstablished` event is [triggered](https://github.com/libp2p/rust-libp2p/blob/0b44564bb031a60a360ccc020676d74c784cca11/swarm/src/lib.rs#L801) AFTER the `handle_established_outbound/inbound_connection` is [called](https://github.com/libp2p/rust-libp2p/blob/0b44564bb031a60a360ccc020676d74c784cca11/swarm/src/lib.rs#L702) and the `ConnectionHandler` is returned from that.

It means that the `FromBehaviour::CloseSubstreams` that I'm scheduling here will be defintely delivered to the `ConnectionHandler`. If the `ConnectionHandler` already established inbound/outbound substreams, they will be replaced with the `Dropped` state, as I implemented in #988. Even if they're not initialized yet, they'll be set as `Dropped` and are never going to be initalized.

nomos-blend/network/src/peer_sets.rs
```diff
@@ -0,0 +1,77 @@
...
+        self.negotiated_healthy_peers.remove(&peer_id);
+        self.negotiated_unhealthy_peers.insert(peer_id);
+    }
+
+    /// Move the peer from the healthy or unhealthy set to the malicious set.
+    pub(crate) fn set_malicious(&mut self, peer_id: PeerId) {
+        self.negotiated_healthy_peers.remove(&peer_id);
+        self.negotiated_unhealthy_peers.remove(&peer_id);
+        self.malicious_peers.insert(peer_id);
+    }
```
AlejandroCabeza, line 70: If when setting a peer as malicious they are removed from healthy and unhealthy; shouldn't it happen the other way around? E.g.: Shouldn't `set_unhealthy` remove from `malicious_peers`?

nomos-blend/network/src/behaviour.rs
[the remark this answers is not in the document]
danielSanchezQ: I still think in this case it is easier to just reject the connection in the `handle_stablished_outbound/inbound`from this behaviour anaway. Accepting a connection just to get it closed feels weird. Maybe I'm missing something here.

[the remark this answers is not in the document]
youngjoon-lee: Okay. That was my initial implementation when I opened this PR, but changed it while having discussions in other PRs. Rejecting the connection itself is the recommended way by libp2p, as far as I understand.

nomos-blend/network/src/peer_sets.rs
youngjoon-lee: Ah, good point. Actually, in the current spec, any peer who has been set as malicious can't be removed from the malicious set. But, yeah, as you said, we should handle the case when the peer is in the malicious set when `set_unhealthy` is called. I will re-think about this state machine and refine it.

nomos-blend/network/src/behaviour.rs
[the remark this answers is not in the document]
youngjoon-lee: Let's use the way libp2p recommends. I will update `handle_stablished_outbound/inbound` deny the connection.

nomos-blend/network/src/peer_sets.rs
youngjoon-lee: I removed this `peer_sets.rs` because I found and use [`libp2p-allow-block-list`](https://docs.rs/libp2p-allow-block-list/latest/libp2p_allow_block_list/) and [`libp2p-connection-limits`](https://docs.rs/libp2p-connection-limits/latest/libp2p_connection_limits/) in the end. 4680e946b72d52db6c09695b3f7e8cda67f3c4fc
They simplify this PR a lot. Instead of maintaining the `malicious_peers` set, now we add malicious peers to the libp2p block list, so that the connections from them or to them can be denied / closed.

nomos-blend/network/src/behaviour.rs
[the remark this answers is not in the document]
youngjoon-lee: 4680e946b72d52db6c09695b3f7e8cda67f3c4fc I just updated this PR by using [`libp2p-allow-block-list`](https://docs.rs/libp2p-allow-block-list/latest/libp2p_allow_block_list/) and [`libp2p-connection-limits`](https://docs.rs/libp2p-connection-limits/latest/libp2p_connection_limits/).

nomos-blend/network/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2021"
 cached = "0.53.1"
 futures = "0.3.30"
 futures-timer = "3.0.3"
-libp2p = "0.53"
+libp2p = "0.55"
```
youngjoon-lee, line 10: To use all functions of [`libp2p-allow-block-list`](https://docs.rs/libp2p-allow-block-list/latest/libp2p_allow_block_list/), we need to upgrade libp2p to 0.55: https://github.com/libp2p/rust-libp2p/commit/e63975d7742710d4498b941e151c5177e06392ce.

nomos-libp2p/src/lib.rs
```diff
@@ -117,7 +117,7 @@ impl Swarm {
     /// Unsubscribes from a topic
     ///
     /// Returns true if previously subscribed
-    pub fn unsubscribe(&mut self, topic: &str) -> Result<bool, PublishError> {
+    pub fn unsubscribe(&mut self, topic: &str) -> bool {
```
youngjoon-lee, line 120: This is because the `unsubscribe` API has changed in libp2p v0.55

nomos-services/blend/src/backends/libp2p.rs
```diff
@@ -96,43 +98,72 @@ impl BlendBackend for Libp2pBlendBackend {
...
-struct BlendSwarm {
-    swarm: Swarm<nomos_blend_network::Behaviour<SphinxMessage, TokioIntervalStreamProvider>>,
+struct BlendSwarm<R> {
+    swarm: Swarm<BlendBehaviour>,
     swarm_messages_receiver: mpsc::Receiver<BlendSwarmMessage>,
     incoming_message_sender: broadcast::Sender<Vec<u8>>,
+    // TODO: Instead of holding the membership, we just want a way to get the list of addresses.
+    membership: Membership<PeerId, SphinxMessage>,
+    rng: R,
+    peering_degree: usize,
```
youngjoon-lee, line 108: I'll update this to `u16` by rebasing this branch to the https://github.com/logos-co/nomos-node/pull/987, after all changes are reviewed completely.

Dockerfile
```diff
@@ -1,6 +1,6 @@
 # BUILD IMAGE ---------------------------------------------------------
 
-FROM rust:1.82.0-slim-bookworm AS builder
+FROM rust:1.84.0-slim-bookworm AS builder
```
youngjoon-lee, line 3: libp2p v0.55 requires Rust >=1.83.

Why libp2p v0.55? Please see https://github.com/logos-co/nomos-node/pull/989/files#r1929439649

nomos-blend/network/src/peer_sets.rs
AlejandroCabeza: I'll go over it again, no worries :)

AlejandroCabeza (approved):

danielSanchezQ (approved):

PR #991, opened 2025-01-21
feat(kms): add filesystem backend

youngjoon-lee:

This PR implements `FileSystemKMSBackend`, which stores and loads keys to a disk. Keys are encrypted by a password.

This backend can be used as the most naive option. In the future, it would be good to implement `KeyringKMSBackend` using the [`keyring`](https://docs.rs/keyring/latest/keyring/) crate, or so.

The APIs defined by the #981 have been slightly modified to accommodate the implementations. I added PR comments for details.

Yes

Yes

No

## Checklist

> [!WARNING]  

* [ ] 1. Description added.

### DIFF
--- a/nomos-services/key-management-system/src/backend/fs.rs
+++ b/nomos-services/key-management-system/src/backend/fs.rs
@@ -1,0 +1,223 @@
+use std::{collections::HashMap, path::PathBuf};
+
+use aes_gcm::{aead::Aead, Aes256Gcm, KeyInit, Nonce};
+use bytes::Bytes;
+use either::Either;
+use overwatch_rs::DynError;
+use pbkdf2::pbkdf2_hmac;
+use sha2::Sha256;
+use uuid::Uuid;
+use zeroize::{Zeroize, Zeroizing};
+
+use crate::{keys::ed25519::Ed25519Key, secure_key::SecuredKey, KMSOperator};
+
+use super::KMSBackend;
+
+type UuidBytes = [u8; 16];
+
+pub struct FileSystemKMSBackend {
+    dir: PathBuf,
+    encryption_key: aes_gcm::Key<Aes256Gcm>,
+    keys: HashMap<UuidBytes, Box<dyn SecuredKey + Sync + Send>>,
+}
+
+#[derive(Clone)]
+pub struct FileSystemKMSBackendSettings {
+    pub dir: PathBuf,
+    pub password: String,
+    pub salt: String,
+    pub rounds: u32,
+}
+
+#[async_trait::async_trait]
+impl KMSBackend for FileSystemKMSBackend {
+    type SupportedKeys = SupportedKeys;
+    type KeyId = UuidBytes;
+    type Settings = FileSystemKMSBackendSettings;
+
+    fn new(settings: Self::Settings) -> Self {
+        // Derive an encryption key from the password using PBKDF2
+        let mut derived_key = [0u8; 32];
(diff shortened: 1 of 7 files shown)


nomos-services/key-management-system/src/lib.rs
```diff
@@ -12,16 +12,22 @@ use overwatch_rs::services::{ServiceCore, ServiceData, ServiceId};
...
+mod keys;
 mod secure_key;
 
 const KMS_TAG: ServiceId = "KMS";
 
 // TODO: Use [`AsyncFnMut`](https://doc.rust-lang.org/stable/std/ops/trait.AsyncFnMut.html#tymethod.async_call_mut) once it is stabilized.
 pub type KMSOperator = Box<
-    dyn FnMut(&mut dyn SecuredKey) -> Box<dyn Future<Output = Result<(), DynError>>> + Send + Sync,
+    dyn FnMut(
+            &mut Box<dyn SecuredKey + Send + Sync>,
```
danielSanchezQ, line 27: If you have a `Box<dyn SecuredKey + Send + Sync>` you should be able to use it in a `&mut dyn SecuredKey + Send + Sync`. What was the issue?

nomos-services/key-management-system/src/backend/fs.rs
```diff
@@ -0,0 +1,223 @@
...
+            .ok_or(Error::KeyIdNotRegistered(key_id))?)
+        .await
+    }
+}
+
+#[repr(u8)]
+#[allow(dead_code)] // TODO: Remove this macro once this is used outside
+pub enum SupportedKeys {
+    Ed25519 = 0,
+}
```
danielSanchezQ, line 124: Why `u8` repr?

nomos-services/key-management-system/src/lib.rs
```diff
@@ -12,16 +12,22 @@ use overwatch_rs::services::{ServiceCore, ServiceData, ServiceId};
...
+        ) -> Pin<Box<dyn Future<Output = Result<(), DynError>> + Send + Sync>>
+        + Send
+        + Sync,
```
youngjoon-lee, line 30: 1. The returned `Future` is not `Pin`. If not, the `Future` cannot be awaited in the [`execute`](https://github.com/logos-co/nomos-node/pull/991/files?diff=unified&w=0#diff-ca57af75e81d3c4d324286c2cebbe8a63d1311702390be1a76a4b0a1d0e2c619R111) function, due to an error: `Future cannot be unpinned`.

2. The argument `dyn SecuredKey` is boxed now. I'd tried to keep the previous verison, but then, I had to wrap different `SecuredKey` implementations into a single enum, and implement `SecuredKey` trait for the enum again, for keeping them in memory because they should be used in the `sign`, `public_key`, and `execute` functions. Another way is just implementing the `SecuredKey` trait for the common enum, and place all implementations of each key type into the enum. However, then, those implementations of each key type cannot be reused for other backend (will be added in the future).
Because of these reasons, I decided to update this API to use `Box`.

danielSanchezQ: The pinned future is a good catch. But I do not think the other is necessary :thinking: 
It is just dynamic dispatching, you do not need to wrap everything in an enum. Why?

youngjoon-lee: it's hard to explain in this comment. let me try it again in other ways.

youngjoon-lee:
This PR implements the different understanding

PR #992, opened 2025-01-22
feat(kms): add the preload backend

youngjoon-lee:

This PR implements a naive KMS backend, which preloads all keys from the setting struct provided by an user. As we've been currently doing, the user must specify all keys in the config file, for example.
This backend provides key operations on only the keys preloaded. It doesn't support generating new keys. 

This is the most simplest version of the KMS backend. We'll be able to design an ideal backend once the requirement from the validator protocol is determined clearly. A possible approach may be making the backend hold a single master key and derive all keys from the master key along with a key ID provided by user, deterministically. This approach is almost the same as the HD key derivation, which most of cyrpto wallets are using, or any other deterministic key derivation algorithms. 

But anyway, it's enough to have this simple preload backend for now.

Yes

Yes

No

## Checklist

> [!WARNING]  

* [x] 1. Description added.

### DIFF
--- a/nomos-services/key-management-system/src/backend/preload.rs
+++ b/nomos-services/key-management-system/src/backend/preload.rs
@@ -111,8 +111,6 @@ impl Key {
 
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
-    #[error("Not supported")]
-    NotSupported,
     #[error("Key({0}) was not registered")]
     KeyNotRegistered(String),
     #[error("KeyType mismatch: {0:?} != {1:?}")]
--- a/nomos-services/key-management-system/src/backend/preload.rs
+++ b/nomos-services/key-management-system/src/backend/preload.rs
@@ -1,0 +1,196 @@
+use std::collections::HashMap;
+
+use bytes::Bytes;
+use either::Either;
+use overwatch_rs::DynError;
+use serde::{Deserialize, Serialize};
+use zeroize::ZeroizeOnDrop;
+
+use crate::{keys::ed25519::Ed25519Key, secure_key::SecuredKey, KMSOperator};
+
+use super::KMSBackend;
+
+pub struct PreloadKMSBackend {
+    keys: HashMap<String, Key>,
+}
+
+/// This settings contain all [`Key`]s to be loaded into the [`PreloadKMSBackend`].
+/// This implements [`serde::Serialize`] for users to populate the settings from bytes.
+/// The [`Key`] also implements [`zeroize::ZeroizeOnDrop`] for security.
+#[derive(Serialize, Deserialize)]
+pub struct PreloadKMSBackendSettings {
+    keys: HashMap<String, Key>,
+}
+
+#[async_trait::async_trait]
+impl KMSBackend for PreloadKMSBackend {
+    type SupportedKeys = SupportedKeys;
+    type KeyId = String;
+    type Settings = PreloadKMSBackendSettings;
+
+    fn new(settings: Self::Settings) -> Self {
+        Self {
+            keys: settings.keys,
+        }
+    }
+
+    /// This function just checks if the key_id was preloaded successfully.
+    /// This backend doesn't support generating a new key based on the [`SupportedKeys`] provided,
+    /// because this backend preloads all keys from the settings.
+    /// In other words, the users can execute operations with the key_id
+    /// even without the need to register it, if the key was specified in the settings.
+    fn register(
+        &mut self,
+        key: Either<Self::SupportedKeys, Self::KeyId>,
+    ) -> Result<Self::KeyId, DynError> {
+        match key {
+            Either::Left(_) => Err(Error::NotSupported.into()),
+            Either::Right(key_id) => {
+                if self.keys.contains_key(&key_id) {
+                    Ok(key_id)
+                } else {
+                    Err(Error::KeyNotRegistered(key_id).into())
+                }
+            }
+        }
+    }
+
+    fn public_key(&self, key_id: Self::KeyId) -> Result<Bytes, DynError> {
+        Ok(self
+            .keys
+            .get(&key_id)
+            .ok_or(Error::KeyNotRegistered(key_id))?
+            .as_pk())
+    }
+
+    fn sign(&mut self, key_id: Self::KeyId, data: Bytes) -> Result<Bytes, DynError> {
+        self.keys
+            .get_mut(&key_id)
+            .ok_or(Error::KeyNotRegistered(key_id))?
+            .sign(data)
+    }
+
+    async fn execute(&mut self, key_id: Self::KeyId, mut op: KMSOperator) -> Result<(), DynError> {
+        op(self
+            .keys
+            .get_mut(&key_id)
+            .ok_or(Error::KeyNotRegistered(key_id))?)
+        .await
+    }
+}
+
+// This enum won't be used outside of this module
+// because [`PreloadKMSBackend`] doesn't support generating new keys.
+#[allow(dead_code)]
+pub enum SupportedKeys {
+    Ed25519,
+}
+
+#[derive(Serialize, Deserialize, ZeroizeOnDrop)]
+enum Key {
+    Ed25519(Ed25519Key),
+}
+
+impl SecuredKey for Key {
+    fn sign(&mut self, data: Bytes) -> Result<Bytes, DynError> {
+        match self {
+            Self::Ed25519(key) => key.sign(data),
+        }
+    }
+
+    fn as_pk(&self) -> Bytes {
+        match self {
+            Self::Ed25519(key) => key.as_pk(),
+        }
+    }
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Not supported")]
+    NotSupported,
+    #[error("Key({0}) was not registered")]
+    KeyNotRegistered(String),
+}
+
+#[cfg(test)]
+mod tests {
+    use std::collections::HashMap;
+
+    use rand::rngs::OsRng;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn preload_backend() {
+        // Initialize a backend with a pre-generated key in the setting
+        let key_id = "blend/1".to_string();
+        let key = ed25519_dalek::SigningKey::generate(&mut OsRng);
+        let mut backend = PreloadKMSBackend::new(PreloadKMSBackendSettings {
+            keys: HashMap::from_iter(vec![(
+                key_id.clone(),
+                Key::Ed25519(Ed25519Key(key.clone())),
+            )]),
+        });
+
+        // Check if the key was preloaded successfully
+        assert_eq!(
+            backend.register(Either::Right(key_id.clone())).unwrap(),
+            key_id
+        );
+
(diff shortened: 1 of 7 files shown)


nomos-services/key-management-system/src/secure_key.rs
```diff
@@ -2,6 +2,6 @@ use bytes::Bytes;
 use overwatch_rs::DynError;
 
 pub trait SecuredKey {
-    fn sign(&self) -> Result<Bytes, DynError>;
+    fn sign(&mut self, data: Bytes) -> Result<Bytes, DynError>;
```
AlejandroCabeza, line 5: Does it make sense for a `sign` to need `mut self`?

nomos-services/key-management-system/src/backend/preload.rs
```diff
@@ -0,0 +1,196 @@
...
+        Self {
+            keys: settings.keys,
+        }
+    }
+
+    /// This function just checks if the key_id was preloaded successfully.
+    /// This backend doesn't support generating a new key based on the [`SupportedKeys`] provided,
+    /// because this backend preloads all keys from the settings.
+    /// In other words, the users can execute operations with the key_id
+    /// even without the need to register it, if the key was specified in the settings.
```
danielSanchezQ, line 41: But what if a service wasn't given an id for the key? How would it ask for it the first time?
As this is a contract we should use the established contract (trait) in the services and it should work the same. In the case of registering, instead of returning an error, I would return the corresponding keyid.

nomos-services/key-management-system/src/backend/preload.rs
```diff
@@ -0,0 +1,196 @@
...
+        .await
+    }
+}
+
+// This enum won't be used outside of this module
+// because [`PreloadKMSBackend`] doesn't support generating new keys.
+#[allow(dead_code)]
+pub enum SupportedKeys {
+    Ed25519,
+}
```
danielSanchezQ, line 87: In this keys the supported ones should be probably the ones the services need, ex:

```suggestion
#[allow(dead_code)]
pub enum SupportedKeys {
    Blend,
    DA,
    Consensus
}
```

Then we map those to the ones loaded from the config.

nomos-services/key-management-system/src/secure_key.rs
danielSanchezQ: I think it is because some keys use a random state inside that need to advance. Can you confirm @youngjoon-lee ?

danielSanchezQ (changes requested):

nomos-services/key-management-system/src/backend/preload.rs
youngjoon-lee: This says that we still have a different understanding of the KMS API. Let's clarify it now.

Personally, I think it's better to treat `SupportedKeys` as crypto scheme (like `Ed25519`).
If we treat it as the 'category' like Blend and DA, we also need to think about how to handle the case which each category requires multiple crypto schemes. For example, the Blend may require both X25519 and Ed25519 for different purposes. It would be possible to support the case somehow, but I think treating `SupportedKeys` as crypto scheme is one of the ways.

In my approach, the setting for the KMS service would look like below:
```yaml
kms:
  keys:
    - id: "blend/1"
      key: "Gdiast....."
    - id: "blend/2"
      key: "poijbpnk..."
    - id: "da/1"
      key: "qweiojrpoi..." 
```
A `key` is a byte array serialized from the `enum Key` which contains both the key type information and the key itself. 

With this setting, I think we can cover all cases that we can imagine. What do you think? This is little different from what we discussed in the call recently, but I ended up with this design because of the reason I mentioned above.
Also, it's better because we can decouple this KMS service and backend from other services. If we specify `Blend` or `DA` in the `SupportedKeys`, this KMS backend should be updated whenever we add a new service or remove some.

nomos-services/key-management-system/src/secure_key.rs
youngjoon-lee: It's because I used [`ed25519_dalek::ed25519::signature::SignerMut`](https://docs.rs/signature/2.2.0/signature/trait.SignerMut.html) instead of [`ed25519_dalek::ed25519::signature::Signer`](https://docs.rs/ed25519-dalek/latest/ed25519_dalek/trait.Signer.html#method.sign), but actually we don't need to, right now. 
I removed the `mut`. Thank you.
3d2156a5

nomos-services/key-management-system/src/backend/preload.rs
youngjoon-lee: This comment is related to the other comment below https://github.com/logos-co/nomos-node/pull/992#discussion_r1926855623. It depends on how we define the `SupportedKeys`.

nomos-services/key-management-system/src/backend/preload.rs
youngjoon-lee: Now, I think it would be better to redefine the backend API like below.

```diff
 use crate::KMSOperator;
 use bytes::Bytes;
-use either::Either;
 use overwatch_rs::DynError;

 #[async_trait::async_trait]
 pub trait KMSBackend {
-    type SupportedKeys;
+    type SupportedKeySchemes;
     type KeyId;
     type Settings;

     fn new(settings: Self::Settings) -> Self;
     fn register(
         &mut self,
-        key: Either<Self::SupportedKeys, Self::KeyId>,
+        key_id: Self::KeyId,
+        key_scheme: Self::SupportedKeySchemes,
     ) -> Result<Self::KeyId, DynError>;
     fn public_key(&self, key_id: Self::KeyId) -> Result<Bytes, DynError>;
     fn sign(&self, key_id: Self::KeyId, data: Bytes) -> Result<Bytes, DynError>;
```

The user MUST call the `register` before calling other functions.

When the `register` is called, the internal behavior depends on the backend type.
The `PreloadBackend` may just check if the `key_id` is in the settings with the same `key_scheme`. If not, it can return an error.
If we implement the `KeyDerivationBackend` in near future, it will derive a key bytes by composing its master key and the `key_id` provided, and will convert the key bytes to a specific data structure corresponding to the `key_scheme` provided. (e.g. `ed25519_dalek::SigningKey` of the `key_scheme` is `Ed25519`).

The responsibility of the user is memorizing the `key_id` and `key_scheme` somehow.

The setting would look like below, if the backend is `PreloadBackend`.
```yaml
kms:
  keys:
    - id: "/1/1/0"
      scheme: "x25519"
      key: "aaaaaaaaaa...."
    - id: "/1/2/0"
      scheme: "ed25519"
      key: "bbbbbbbbb..."
    - id: "/2/1/0"
      scheme: "ed25519"
      key: "cccccc..." 

blend:
  crypto_processor:
    key:
      id: "/1/1/0"
      scheme: "x25519"
  something:
    key:
      id: "/1/2/0"
      scheme: "ed25519"

da:
  something:
    key:
      id: "/2/1/0"
      scheme: "ed25519"
```

danielSanchezQ: > Now, I think it would be better to redefine the backend API like below.
> 
> ```diff
>  use crate::KMSOperator;
>  use bytes::Bytes;
> -use either::Either;
>  use overwatch_rs::DynError;
> 
>  #[async_trait::async_trait]
>  pub trait KMSBackend {
> -    type SupportedKeys;
> +    type SupportedKeySchemes;
>      type KeyId;
>      type Settings;
> 
>      fn new(settings: Self::Settings) -> Self;
>      fn register(
>          &mut self,
> -        key: Either<Self::SupportedKeys, Self::KeyId>,
> +        key_id: Self::KeyId,
> +        key_scheme: Self::SupportedKeySchemes,
>      ) -> Result<Self::KeyId, DynError>;
>      fn public_key(&self, key_id: Self::KeyId) -> Result<Bytes, DynError>;
>      fn sign(&self, key_id: Self::KeyId, data: Bytes) -> Result<Bytes, DynError>;
> ```
> 
> The user MUST call the `register` before calling other functions.
> 
> When the `register` is called, the internal behavior depends on the backend type. The `PreloadBackend` may just check if the `key_id` is in the settings with the same `key_scheme`. If not, it can return an error.
> 
> ```rust
> impl KMSBackend for PreloadKMSBackend {
>     fn register(
>         &mut self,
>         key_id: Self::KeyId,
>         key_scheme: Self::SupportedKeySchemes,
>     ) -> Result<Self::KeyId, DynError>;
>         let key = self.keys.get(&key_id).ok_or(Error::KeyNotRegistered(key_id))?;
>         if key.scheme() != key_scheme {
>             return Err(Error::KeySchemeMismatch.into());
>         }
>         Ok(key_id)
>     }
> ```
> 
> If we implement the `KeyDerivationBackend` in near future, it will derive a key bytes by composing its master key and the `key_id` provided, and will convert the key bytes to a specific data structure corresponding to the `key_scheme` provided. (e.g. `ed25519_dalek::SigningKey` of the `key_scheme` is `Ed25519`).
> 
> ```rust
> impl KMSBackend for KeyDerivationKMSBackend {
>     fn register(
>         &mut self,
>         key_id: Self::KeyId,
>         key_scheme: Self::SupportedKeySchemes,
>     ) -> Result<Self::KeyId, DynError>;
>         if let Some(key) = self.keys.get(&key_id) {
>             if key.scheme() == key_scheme {
>                 return Ok(key_id)
>             } else {
>                 return Err(Error::KeySchemeMismatch.into())
>             }
>         }
> 
>         let derived_key = derive_key(self.master_key, key_id);
>         let key = match key_scheme {
>             SupportedKeySchemes::Ed25519 => Key::Ed25519(ed25519_dalek::Signing_key::from_bytes(key)),
>         };
>         self.keys.insert(key_id, key);
> 
>         Ok(key_id)
>     }
> ```
> 
> The responsibility of the user is memorizing the `key_id` and `key_scheme` somehow.
> 
> The setting would look like below, if the backend is `PreloadBackend`.
> 
> ```yaml
> kms:
>   keys:
>     - id: "/1/1/0"
>       scheme: "x25519"
>       key: "aaaaaaaaaa...."
>     - id: "/1/2/0"
>       scheme: "ed25519"
>       key: "bbbbbbbbb..."
>     - id: "/2/1/0"
>       scheme: "ed25519"
>       key: "cccccc..." 
> 
> blend:
>   crypto_processor:
>     key:
>       id: "/1/1/0"
>       scheme: "x25519"
>   something:
>     key:
>       id: "/1/2/0"
>       scheme: "ed25519"
> 
> da:
>   something:
>     key:
>       id: "/2/1/0"
>       scheme: "ed25519"
> ```
> 
> As far as I know, this is the API similar as other HD-path based crypto wallets. Should the KMS service increment the `key_id` automatically? I don't think so. Specifying the `key_id` is the responsibility of users (other services in our case). It's not difficult for users. It's even flexible. Some crypto wallet "apps" support the automatic HD-path increment in the app level just for better UX. but, the core HD key derivation algorithm expects that the HD-path is provided from the user.

Yes, this is actually a good idea. KeyId can be provided by the requester service instead. I see this path cleaner, we can go ahead with this instead. Most services can track their own key by composing their own service id plus a nonce or whatever they choose. The api is better too!

AlejandroCabeza (approved):

danielSanchezQ (approved):
Superb!

PR #1017, opened 2025-02-05
feature(recovery): Save security block

AlejandroCabeza:
Save `HeaderId` of block in `security_param (k)` position to facilitate recovery.
We don't need to rebuild everything but just from `k` which is where we consider the chain to be stable.

Closes: https://github.com/logos-co/nomos-node/issues/1014

Yes

@AlejandroCabeza 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 1. Description added.


nomos-services/cryptarchia-consensus/src/storage/adapters/storage.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub storage_relay: OutboundRelay<<StorageService<Storage> as ServiceData>::Message>,
+    _cl_pool: PhantomData<ClPool>,
+    _da_pool: PhantomData<DaPool>,
+}
+
+#[async_trait::async_trait]
+impl<Storage, ClPool, DaPool> StorageAdapterTrait for StorageAdapter<Storage, ClPool, DaPool>
+where
+    Storage: StorageBackend + Send + Sync,
+    DaPool: MemPool<BlockId = HeaderId, Item: Clone + Eq + Hash + Send + DeserializeOwned> + Sync,
```
danielSanchezQ, line 29: You can separate this here:

```suggestion
    DaPool: MemPool<BlockId = HeaderId> + Sync,
    DaPool::Item:  Clone + Eq + Hash + Send + DeserializeOwned
```

nomos-services/cryptarchia-consensus/src/storage/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+    type DaPool: MemPool<BlockId = HeaderId, Item: Clone + Eq + Hash>;
+
+    async fn new(
+        network_relay: OutboundRelay<<StorageService<Self::Backend> as ServiceData>::Message>,
+    ) -> Self;
+
+    async fn get_block(
+        &self,
+        header_id: HeaderId,
+    ) -> Option<Block<<Self::ClPool as MemPool>::Item, <Self::DaPool as MemPool>::Item>>;
```
danielSanchezQ, line 28: Cant the block be generic here?

nomos-services/cryptarchia-consensus/src/storage/adapters/storage.rs
```diff
@@ -0,0 +1,78 @@
...
+    ) -> Option<Block<<Self::ClPool as MemPool>::Item, <Self::DaPool as MemPool>::Item>> {
+        let mut current_block = current_block;
+        // TODO: This implies fetching from DB `security_param` times. We should optimize this.
+        for _ in 0..*security_param {
+            let parent_block_header = current_block.header().parent();
+            let parent_block = self.get_block(parent_block_header).await?;
+            current_block = parent_block;
+        }
+        Some(current_block)
+    }
```
danielSanchezQ, line 68: I dont think this is necessary, you can load this from the consensus state directly.

nomos-services/cryptarchia-consensus/src/storage/adapters/storage.rs
```diff
@@ -0,0 +1,87 @@
+// STD
```
danielSanchezQ, line 1: I think we don't use caps for this
```suggestion
// std
```

nomos-services/cryptarchia-consensus/src/storage/adapters/storage.rs
```diff
@@ -0,0 +1,87 @@
...
+            _blob_certificate: Default::default(),
+        }
+    }
+
+    async fn get_block(&self, key: &HeaderId) -> Option<Self::Block> {
+        self.get_value(key).await
+    }
+
+    async fn get_security_block_header(&self) -> Option<HeaderId> {
+        self.get_value(&SECURITY_BLOCK_KEY).await
```
danielSanchezQ, line 77: Yeah. This is what I mean. The security block is just a block in the storage. Like this would be duplicated. We just need to save the `block_id` to retrieve it later. For that, I think we should use the recovery system, not the storage itself.

danielSanchezQ (changes requested):

nomos-services/cryptarchia-consensus/src/storage/adapters/storage.rs
AlejandroCabeza: That makes sense, tbf. Somehow I assumed the intention was to use the storage, but this makes a lot of sense.

nomos-services/cryptarchia-consensus/src/storage/adapters/storage.rs
AlejandroCabeza: For some reason it feels very weird to me to have it written in lowercaps, but I'll try not to look at it 😂

nomos-core/chain-defs/src/da/blob/mod.rs
```diff
@@ -12,7 +12,7 @@ pub trait Blob {
 
 pub trait BlobSelect {
     type BlobId: info::DispersedBlobInfo;
-    type Settings: Clone;
+    type Settings: Clone + Send;
```
danielSanchezQ, line 15: Is this bound and the one in txSelect necessary? I dont think so. In this case you may want to add the bound where it is really used.

nomos-core/chain-defs/src/tx/mod.rs
```diff
@@ -25,7 +25,7 @@ pub trait Transaction {
 
 pub trait TxSelect {
     type Tx: Transaction;
-    type Settings: Clone;
+    type Settings: Clone + Send;
```
danielSanchezQ, line 28: Ditto.

nomos-services/cryptarchia-consensus/src/network/mod.rs
```diff
@@ -18,7 +18,7 @@ type BoxedStream<T> = Box<dyn Stream<Item = T> + Send + Sync + Unpin>;
 #[async_trait::async_trait]
 pub trait NetworkAdapter {
     type Backend: NetworkBackend + 'static;
-    type Settings: Clone + 'static;
+    type Settings: Clone + Send + 'static;
```
danielSanchezQ, line 21: Maybe here as well, although I'm not so sure in this one.

danielSanchezQ (commented):
Overall looks quite nice. Just check the issue with the bounds.

nomos-services/cryptarchia-consensus/src/network/mod.rs
AlejandroCabeza: I also wasn't really convinced of adding `Send` to all those `Settings` but let me elaborate on the situation:
1. This is the definition of `RecoveryOperator` for `CryptarchiaConsensus`: `RecoveryOperator<JsonFileBackend<Self::State, Self::Settings>>`
2. The `Backend` needs to be Send in order to use it inside the `StateOperator::run`, which is why it requires `Settings` to be `Send`.

Is there a better way to overcome this?

danielSanchezQ: I think you can then bound it for cryptarchia consensus as cryptarchia consensus need it rather than the trait :thinking: 
Can you give it a shot?

danielSanchezQ (approved):

PR #1018, opened 2025-02-06
chore: add toolchain file

ntn-x2:

Add a toolchain file with the default profile, which also installs `cargo-fmt` and `cargo-clippy` components. The CI workflows have also been updated where relevant, with the exception of the `Setup Rust with GNU toolchain (Windows)` step which is left untouched since it overrides the toolchain.

Yes.

N/A

N/A

N/A
## Checklist

> [!WARNING]  

* [x] 1. Description added.


rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "1.84"
```
ntn-x2, line 2: Right now CI always uses latest, so this is a deviation from that behaviour. I could also change this to `"stable"` if there's consensus about it.

ntn-x2:
It seems that with a toolchain file, the `actions-rs/toolchain` step is not needed anymore: https://github.com/actions-rs/toolchain/issues/126#issuecomment-1463045252.

rust-toolchain.toml
ntn-x2: Did that here: https://github.com/logos-co/nomos-node/pull/1018/commits/3db2eb9eb0df9e70f858295018ce7bf239562976.

bacv (approved):

danielSanchezQ (approved):
image
Good!

ntn-x2:
I realised I have not tested whether the codecov and packaging workflows still work fine, I suspect they don't. I will get back to this tomorrow morning, while in the meanwhile tonight we'll find out whether the codecov workflow goes through or not.

AlejandroCabeza (approved):

youngjoon-lee (approved):

ntn-x2:
Sorry for the spam. It indeed did not work for the workflows that were not triggered in the CI. I replaced those as well. Nevertheless, the self-hosted runners do not have the `rustup` binary installed just yet, so as suggested by @bacv I temporarily switch the codecov workflow to use the hosted one (`ubuntu-latest`). I opened a ticket to switch back to the self-hosted one once they support the new workflow.

bacv (approved):

PR #1043, opened 2025-02-11
feature(consensus): Recovery

AlejandroCabeza:
Rebuild Node state from a previously saved state:
- Implement a `StorageAdapter` wrapper over the storage relay.
- Save `tip`, and `Block` and `LedgerState` and `NoteWitness` in security position.
- Use saved data (if there is) to initialise `Cryptarchia` and `Leader`, defaulting to genesis strategy if no data is available.
- Cleanup for clarity: Renaming, reordering, logs, etc.

Test plans to be implemented: https://github.com/logos-co/nomos/issues/1139

Yes

@AlejandroCabeza 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 1. Description added.


nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -872,6 +854,206 @@ where
...
+                    .unwrap_or_else(|| {
+                        panic!("Could not retrieve block {to} from storage during recovery")
+                    });
+                let parent_header_id = block.header().parent();
+                Some((block, parent_header_id))
+            }
+        });
+
+        // To avoid confusion, the order is reversed so it fits the natural `from..to` order
+        blocks.collect::<Vec<_>>().await.into_iter().rev().collect()
```
danielSanchezQ, line 899: I think it would be better to return the stream here instead of the collected vec. Then collect it externally when used. Is more generic case imo.

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -872,6 +854,206 @@ where
...
+            ClPool,
+            ClPoolAdapter,
+            DaPool,
+            DaPoolAdapter,
+            NetAdapter,
+            SamplingBackend,
+            SamplingRng,
+            Storage,
+            TxS,
+        >,
```
danielSanchezQ, line 942: Now that I see this. CryptarchiaConsensusRelays should be a type alias to hide generics except in a single point.

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -872,6 +854,206 @@ where
...
+                    ledger_config,
+                    relays,
+                    block_subscription_sender,
+                )
+                .await;
+
+                (cryptarchia, leader)
+            }
+        }
+    }
```
danielSanchezQ, line 1002: Nitpick: This could be refactore as an implementation of `CryptarchiaInitialisationStrategy` (?), with independent methods for each case.

```rust
impl CryptarchiaInitializationStrategy {
  pub fn recover(&self, *args) -> (Cryptarchia, Leader) {
      match self {
          CryptarchiaInitialisationStrategy::Genesis => recover_from_genesis(self, *args),
          ...
      }
  }
}
```

Or at least separated methods for each branch.

danielSanchezQ (changes requested):
Pretty clean approach. Added some comments. Probably we want to add a basic test for each recovery strategy.

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -143,7 +143,7 @@ impl<Id> Cryptarchia<Id>
 where
     Id: Eq + std::hash::Hash + Copy,
 {
-    pub fn from_genesis(id: Id, config: Config) -> Self {
+    pub fn new(id: Id, config: Config) -> Self {
```
ntn-x2, line 146: Could there be a clearer name for this method?

nomos-services/cryptarchia-consensus/src/leadership.rs
```diff
@@ -30,17 +30,26 @@ pub struct LeaderConfig {
...
             config,
         }
     }
 
+    pub fn from_genesis(
+        genesis: HeaderId,
+        LeaderConfig { notes, nf_sk }: LeaderConfig,
+        config: Config,
+    ) -> Self {
+        Leader::new(genesis, notes, nf_sk, config)
```
ntn-x2, line 50: My pedantic clippy suggests me this as a better alternative
```suggestion
        Self::new(genesis, notes, nf_sk, config)
```

nomos-services/cryptarchia-consensus/src/leadership.rs
```diff
@@ -126,6 +135,10 @@ impl Leader {
 
         None
     }
+
+    pub(crate) fn notes(&self, header_id: &HeaderId) -> Option<&Vec<NoteWitness>> {
```
ntn-x2, line 139: Would it make sense to rater return an `impl Iterator` here? Then the consumer can collect it however they want to.

nomos-services/cryptarchia-consensus/src/states.rs
```diff
@@ -0,0 +1,132 @@
...
+        security_block: Option<HeaderId>,
+        security_ledger_state: Option<LedgerState>,
+        security_leader_notes: Option<Vec<NoteWitness>>,
+    ) -> Self {
+        Self {
+            tip,
+            security_block,
+            security_ledger_state,
+            security_leader_notes,
+            _txs: Default::default(),
```
ntn-x2, line 56: My Clippy recommends again:
```suggestion
            _txs: PhantomData::default(),
```

nomos-services/cryptarchia-consensus/src/states.rs
```diff
@@ -0,0 +1,132 @@
...
+
+        Self::new(
+            Some(cryptarchia.tip()),
+            security_block_header,
+            security_ledger_state,
+            security_leader_notes,
+        )
+    }
+
+    fn can_recover(&self) -> bool {
```
ntn-x2, line 80: Could be a const function
```suggestion
    fn const can_recover(&self) -> bool {
```

nomos-services/cryptarchia-consensus/src/states.rs
```diff
@@ -0,0 +1,132 @@
...
+}
+
+pub struct SecurityRecoveryStrategy {
+    pub tip: HeaderId,
+    pub security_block_id: HeaderId,
+    pub security_ledger_state: LedgerState,
+    pub security_leader_notes: Vec<NoteWitness>,
+}
+
+pub enum CryptarchiaInitialisationStrategy {
```
ntn-x2, line 24: Would help a lot to have comments explaining the different strategies.

nomos-services/cryptarchia-consensus/src/states.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+}
+
+impl<TxS, BxS, NetworkAdapterSettings, BlendAdapterSettings> ServiceState
+    for CryptarchiaConsensusState<TxS, BxS, NetworkAdapterSettings, BlendAdapterSettings>
+{
+    type Settings = CryptarchiaSettings<TxS, BxS, NetworkAdapterSettings, BlendAdapterSettings>;
+    type Error = Error;
+
+    fn from_settings(_settings: &Self::Settings) -> Result<Self, Self::Error> {
```
ntn-x2, line 129: Nothing to do with this PR, but this is the bit I was talking about elsewhere. All states must do this, and it looks like every time the implementation will always be the same, which to me indicates it's not optimal design.

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -872,6 +854,206 @@ where
...
+    /// # Returns
+    ///
+    /// A vector of blocks in the range from `from` to `to`.
+    /// If no blocks are found, returns an empty vector.
+    /// If any of the HeaderId are invalid, returns an error with the first invalid header id.
+    async fn get_blocks_in_range(
+        from: HeaderId,
+        to: HeaderId,
+        storage_adapter: &StorageAdapter<Storage, TxS::Tx, BS::BlobId>,
+    ) -> Vec<Block<ClPool::Item, DaPool::Item>> {
```
ntn-x2, line 881: Could this also be an iterator and letting the consumers choose?

nomos-services/cryptarchia-consensus/src/lib.rs
AlejandroCabeza: The reason for this is to force it to natural order, because afaik, you can't reverse an stream.
Due to `unfold`'s nature, items are in `to` to `from` order, which could be confusing to callers.

nomos-services/cryptarchia-consensus/src/lib.rs
AlejandroCabeza: Is there a way to do it without an associated type?

nomos-services/cryptarchia-consensus/src/lib.rs
AlejandroCabeza: The thought did cross my mind but I forgot. Thanks!

consensus/cryptarchia-engine/src/lib.rs
AlejandroCabeza: I'd argue it's quite accurate. It's really not doing anything fancy, nor anything attached to a special state (genesis), just getting an instance with a given `id`.

nomos-services/cryptarchia-consensus/src/leadership.rs
AlejandroCabeza: That slipped by, thanks.

nomos-services/cryptarchia-consensus/src/leadership.rs
AlejandroCabeza: `self.notes` is already a `Vec`, do you think it's actually worth casting it to return it here?

nomos-services/cryptarchia-consensus/src/states.rs
AlejandroCabeza: I wonder, how does the compiler resolve this?

nomos-services/cryptarchia-consensus/src/lib.rs
AlejandroCabeza: Dani asked me the same thing. Because I answered there, I'd rather have the convo all in the same place if you don't mind :)
https://github.com/logos-co/nomos-node/pull/1043#discussion_r1975951506

nomos-services/cryptarchia-consensus/src/lib.rs
ntn-x2: There is an [`iter.rev()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.rev) which might be what you are looking for? This works on iterators, not sure if streams are something else?

consensus/cryptarchia-engine/src/lib.rs
danielSanchezQ: Naming is fine I think.

nomos-services/cryptarchia-consensus/src/leadership.rs
danielSanchezQ: Probaly `Option<&[NoteWitness]>` is fine. From that can be transformer into iter if needed, or be indexed (which probably is not needed but anyway...).

nomos-services/cryptarchia-consensus/src/relays.rs
```diff
@@ -108,32 +108,34 @@ impl<
...
-    ClPool::Key: Debug + 'static,
+    ClPool::Key: 'static,
     ClPool::Settings: Clone,
     ClPoolAdapter: MempoolAdapter<Payload = ClPool::Item, Key = ClPool::Key>,
     DaPool: RecoverableMempool<BlockId = HeaderId>,
-    DaPool::RecoveryState: Serialize + for<'de> Deserialize<'de>,
     DaPool::BlockId: Debug,
     DaPool::Item: Debug + DeserializeOwned + Eq + Hash + Clone + Send + Sync + 'static,
     DaPool::Key: Debug + 'static,
+    //
```
danielSanchezQ, line 122: leftover?

nomos-services/cryptarchia-consensus/src/lib.rs
AlejandroCabeza: That's probably what I'm already using in the `into_iter().rev()` section. It makes sense to me that you can't chain a `Stream` with a `rev` without explicitly storing it first: You need to store the whole stream in memory (e.g.: vector) before reversing it, otherwise how do you know what stream item is the first one?

nomos-services/cryptarchia-consensus/src/lib.rs
AlejandroCabeza: As good as it sounded moving that code to the strategy structs, it propagates generics to them as well which seemed a bit messy.

For now I ended up just separating them into methods for each branch, and they can be easily moved into strategy if we want.

Do you guys consider worth it worth? That is, moving them to the strategy section.

nomos-services/cryptarchia-consensus/src/relays.rs
AlejandroCabeza: Yup, from the rebas, still cleaning up stuff.

danielSanchezQ (approved):

nomos-services/cryptarchia-consensus/src/states.rs
danielSanchezQ: The idea behind this iirc is that you may overwrite saved state from settings on some services.

ntn-x2: But state should not contain settings, right? Those are always fetched when the service is started or within the service's `run` loop, so I don't see a case in which a state might do that.

ntn-x2: Cause all I've seen is for the service to have settings for the different components, but having settings for a service's state, to me, sounds less than ideal. But maybe there is a concrete use case in which this is needed, not sure.

danielSanchezQ: > Cause all I've seen is for the service to have settings for the different components, but having settings for a service's state, to me, sounds less than ideal. But maybe there is a concrete use case in which this is needed, not sure.

It is not settings for the service state. It is the settings for the service itself. Its true that it could be done from within the service. Meaning that a service would recover its saved state, it will have the settings and to whatever it need to do from it. Instead of delegating it to the state operator.

ntn-x2: Yeah but the trait defines a `type Settings` for the state, which effectively makes it state's settings. There are then bounds that bound it to the service's state, but that is not part of the trait definition. As for your second sentences, I am not sure I understand if you agree with me, or making the case for having this solution? Let's say some stuff must be recovered from a remote DB which requires a bunch of configs. Wouldn't that all be part of the recovery operator, which would be configured with some of the service's settings? 

What I am arguing here is that having this `type Settings` and `fn from_settings` for a service state is not necessary, but perhaps you can help me better understand your previous point?

danielSanchezQ: > As for your second sentences, I am not sure I understand if you agree with me, or making the case for having this solution?

I partially agree with you. Partially making the case. 
Usually, we imagine the state as the composition of some `settings` + `state_N`. At the start, it would just be a state from settings. Later on, it would be the `state_13` + `settings`. It may not make sense for all services *now*, but, it was how it was conceived. Now, *this* could be done by the state operator or by the service itself. We decided to delegate that to the operator.
I see your point about removing it as it is not used atm. But I think it may as well just be kept here. It doesn't harm having it. Unless you have a better proposition than just removing it.

danielSanchezQ: > What I am arguing here is that having this type Settings and fn from_settings for a service state is not necessary, but perhaps you can help me better understand your previous point?

It is true that maybe this should be the `ServiceSettings` itself :thinking:

PR #1049, opened 2025-02-12
feat: add recovery for CL mempool

ntn-x2:

Integrate recovery mechanism into the CL mempool service.

Yes.

@ntn-x2 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 1. Description added.


nomos-services/mempool/src/backend/mod.rs
```diff
@@ -56,6 +56,13 @@ pub trait MemPool {
     fn status(&self, items: &[Self::Key]) -> Vec<Status<Self::BlockId>>;
 }
 
+pub trait RecoverableMempool: MemPool {
+    type RecoveryState;
+
+    fn recover(settings: Self::Settings, state: Self::RecoveryState) -> Self;
+    fn save(&self) -> Self::RecoveryState;
+}
+
```
danielSanchezQ, line 65: Nice, good option so we do not modify the Mempool trait!

nomos-services/mempool/src/backend/mockpool.rs
```diff
@@ -67,7 +90,9 @@ where
         self.last_item_timestamp = SystemTime::now()
             .duration_since(UNIX_EPOCH)
             .unwrap()
-            .as_millis() as u64;
+            .as_millis()
+            .try_into()
+            .unwrap();
```
danielSanchezQ, line 95: Nitpik: Better to use `expect` with a hint message.

nomos-services/mempool/src/tx/service.rs
```diff
@@ -6,28 +6,71 @@ pub mod openapi {
...
+}
 
 pub struct TxMempoolService<N, P>
 where
     N: NetworkAdapter<Payload = P::Item, Key = P::Key>,
-    P: MemPool,
-    P::Settings: Clone,
+    N::Settings: Send,
+    P: RecoverableMempool,
+    P::RecoveryState: Clone + Send + Sync + Serialize + DeserializeOwned,
```
danielSanchezQ, line 72: Do we really need the clone here? I think it should be safe to remove it and just make the call in the RecoverableMempool implementation where is used.

nomos-services/mempool/src/backend/mockpool.rs
```diff
@@ -128,3 +153,20 @@ where
...
+{
+    type RecoveryState = Self;
+
+    fn recover(_settings: Self::Settings, state: Self::RecoveryState) -> Self {
+        state
+    }
+
+    fn save(&self) -> Self::RecoveryState {
+        self.clone()
+    }
```
danielSanchezQ, line 171: Yeah, since we clone here internally there should be no necessity for a bound in the outer layers (?)

danielSanchezQ (commented):
Looking good!

nomos-services/mempool/src/tx/service.rs
danielSanchezQ: Ah yeah, its true that the state is `Clone`.  Maybe that is not really necessary :thinking: 
@AlejandroCabeza did we introduce that during the latest changes or was it like that originally? I think that bound should not really be necessary.

ntn-x2: I can check whether it's required or not. In general, I am personally not a big fan of having bounds on trait associated types, unless they are required in one of the methods exposed by the trait. Otherwise, those bounds can easily be expressed in the place where the trait is used.

ntn-x2: Yeah I'll give it a crack.

ntn-x2: @danielSanchezQ @AlejandroCabeza I got rid of most of the aforementioned bounds in this PR: https://github.com/logos-co/Overwatch/pull/46. The final result that I can see from the examples and tests in the overwatch repo does not look any worse. The only thing I cannot fully picture yet, is whether removing the inter-trait bounds would make services like cryptarchia more complicated, but I don't think that will be the case. I think the new structure would make simpler services easier to define. Let me know what you guys think!

nomos-services/mempool/src/tx/service.rs
```diff
@@ -6,28 +6,71 @@ pub mod openapi {
...
 };
-use tracing::error;
+
+#[derive(Error, Debug)]
+pub enum StateError {}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct MempoolState<NS, PS, P> {
+    // `pool` is an option so that we don't need it to implement `Mempool`
+    // when implementing `ServiceState` for this type, as otherwise we would need to call `P::new(settings)`.
```
AlejandroCabeza, line 39: There's another nuance to this: You need the attributes to be `Option`s as that's what provides the ability to differentiate whether there was a previously saved value or there was none.
I would suggest updating the comment (or maybe even removing it?) so as to avoid confusion with this behaviour. Or do you feel it makes sense as it is?

nomos-services/mempool/src/tx/service.rs
AlejandroCabeza: @danielSanchezQ I feel like we didn't introduce them recently but I might be wrong. IIRC the Clone behaviour was there to be able to share States among services.

nomos-services/mempool/src/tx/service.rs
ntn-x2: Yes, I will rework this whole PR actually once I merge the two outstanding ones in the overwatch repo, as I feel that is enough of a refactor to justify rewriting the few lines I wrote in this PR. But yours is a good point, and I will change that as well.

nodes/nomos-executor/src/config/mod.rs
```diff
@@ -33,6 +35,7 @@ pub struct Config {
     pub http: <ExecutorApiService as ServiceData>::Settings,
     pub cryptarchia: <crate::ExecutorCryptarchia as ServiceData>::Settings,
     pub storage: <crate::StorageService<RocksBackend<Wire>> as ServiceData>::Settings,
+    pub mempool: MempoolConfig,
```
ntn-x2, line 38: I did not know how else to read this config here. Taking the whole service settings basically means taking an empty tuple for the mock mempool and some libp2p settings which I guess are not important since they are not specified already right now, so I figured we would need a new type to include include the recovery path.

ntn-x2: Also I would like to reduce duplication and move this somewhere where it makes sense, and refer to it from both `nomos-node` and `nomos-executor`, but I would like to get opinion on whether it makes sense, and if it does, where it would go.

nomos-services/mempool/src/backend/mockpool.rs
ntn-x2: Did it just here: https://github.com/logos-co/nomos-node/pull/1049/commits/568c170b6b36b5d30f38ce702b60e1c8026a1974

nodes/nomos-executor/src/config/mod.rs
danielSanchezQ: The configuration is different for both. It would be nice if we could flatten one into the other though. Think of an executor as an extra behaviour on top of a `nomos-node`. Maybe it can be done with a macro :thinking: wdyt?

danielSanchezQ: Ah ok, ok. Didn't understood it. My bad. I think the `MempoolConfig` can probably go service level or in `mockpool` if its the only one that have it. Then it can be accessed from the binaries.

ntn-x2: This is a node-level configuration, so I personally don't think it belongs to the service level. I also think it does not belong to the mock pool itself, as it already specifies a settings type which includes both the pool and the network adapter settings, so this would only introduce confusion. If anything, I would perhaps introduce a new crate, nodes common, where we can put stuff that is common to both validator, executor, and any other node binary we might need in the future? What do you think?

ntn-x2: Yeah makes sense, thanks. I changed it in https://github.com/logos-co/nomos-node/pull/1049/commits/c4cf5329501bdddda1691270971700c59fb1be5b.

bacv (approved):
Thats a great improvement, looks good.

nomos-services/mempool/src/backend/mockpool/mod.rs
```diff
@@ -31,20 +82,37 @@ where
...
+}
+
 impl<BlockId, Item, Key> MockPool<BlockId, Item, Key>
 where
     Key: Hash + Eq + Clone,
 {
+    #[must_use]
     pub fn new() -> Self {
-        Default::default()
+        Self::default()
```
danielSanchezQ, line 107: Is there a lint for this? I've seen this around in other places like phantoms and I'm curious why the change?

nomos-services/mempool/src/backend/mockpool/mod.rs
```diff
@@ -128,3 +198,37 @@ where
...
+        state
+    }
+
+    fn save(&self) -> Self::RecoveryState {
+        self.clone()
+    }
+}
+
+#[derive(Error, Debug)]
+pub enum Error {}
```
danielSanchezQ, line 220: Why and empty error? Maybe you want to use [`Infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html) instead?

nomos-services/mempool/src/tx/service.rs
```diff
@@ -6,117 +6,157 @@ pub mod openapi {
...
-    P: MemPool,
-    P::Settings: Clone,
-    P::Item: Debug + 'static,
-    P::Key: Debug + 'static,
-    P::BlockId: Debug + 'static,
+    Pool: RecoverableMempool,
+    Pool::Settings: Clone,
+    NetworkAdapter: NetworkAdapterTrait,
+    NetworkAdapter::Settings: Clone,
+    RecoveryBackend: RecoveryBackendTrait,
```
danielSanchezQ, line 82: Thanks for the effort!

nomos-services/mempool/src/tx/service.rs
```diff
@@ -6,117 +6,157 @@ pub mod openapi {
...
+        Ok(Self::new(recovered_pool, service_state))
     }
 
     async fn run(mut self) -> Result<(), overwatch_rs::DynError> {
-        let Self {
-            mut service_state,
-            network_relay,
-            mut pool,
-            ..
-        } = self;
```
danielSanchezQ, line 91: I'm also curious, is there a lint for this as well?

nomos-services/mempool/src/tx/state.rs
```diff
@@ -0,0 +1,52 @@
...
+    fn from(value: PoolState) -> Self {
+        Self {
+            pool: Some(value),
+            _phantom: PhantomData,
+        }
+    }
+}
+
+#[derive(Error, Debug)]
+pub enum Error {}
```
danielSanchezQ, line 38: Likewise here.

danielSanchezQ (changes requested):

danielSanchezQ:
Forgot to put the comment in the *request changes* review: I just think the empty `Errors` are not needed. Otherwise everything looks ok.

nomos-services/mempool/src/backend/mockpool/mod.rs
ntn-x2: Yes. `'-W clippy::default-trait-access' implied by '-W clippy::pedantic'`

nomos-services/mempool/src/tx/service.rs
danielSanchezQ: No need to change. I was curious as well. But I would prefer that if it is just personal preference, we keep what is there unless there is a specific reason for that. I think in this case, it is ok as we changed some `async` methods to `sync` so borrowing of `self` is not affected. But in most services, you will end up having problems as you need different parts of `self` in different async contexts.

nomos-services/mempool/src/backend/mockpool/mod.rs
danielSanchezQ: I wonder at what point did this changes. Using `Default` was the pattern at some point iirc. Anyway, doesn't really matter too much. Thanks!

danielSanchezQ (approved):

nomos-services/mempool/src/backend/mockpool/mod.rs
ntn-x2: I think `..Default::default()` is expected when we want to initialise some fields and use the type's `Default` implementation for the rest. If we need the whole type's default, from within the same type impl block, then this has been there for a while, as far as I remember.

PR #1057, opened 2025-02-17
chore: setup proper workspace dependencies

ntn-x2:

First step towards a unified workspace with workspace dependencies. This PR does some basic groundwork, including:

* Adds the [`taplo`](https://taplo.tamasfe.dev/) formatter to format TOML files. For those not using VSCode, it can be downloaded as a binary and run as part of regular formatting procedures.
* Add all workspace-local dependencies to the workspace root `Cargo.toml` (except for the risc0 stuff), and references those in all depending crates
* Fix tracing feature flag mishandling in `nomos-executor` cc @bacv 
* Fix minor missing features for some dependencies
* Remove some leftover files regarding RISC0 proof of leadership cc @zeegomo 

Yes.

@ntn-x2 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 1. Description added.


Cargo.toml
```diff
@@ -1,50 +1,99 @@
...
+  "nomos-services/mempool",
+  "nomos-services/network",
+  "nomos-services/storage",
+  "nomos-services/system-sig",
+  "nomos-services/tracing",
+  "nomos-services/utils",
+  "nomos-tracing",
+  "nomos-utils",
+  "testnet/cfgsync",
+  "tests",
```
danielSanchezQ, line 45: Hell yeah! thanks

danielSanchezQ (approved):
Nice and clean. I like. Thanks

bacv (approved):

nomos-core/risc0_proofs/ptx/Cargo.toml
```diff
@@ -1,18 +1,17 @@
...
+name    = "ptx"
+version = "0.1.0"
 
 [workspace]
 
 [dependencies]
-risc0-zkvm = { version = "1.0", default-features = false, features = ['std'] }
-serde = { version = "1.0", features = ["derive"] }
-cl = { path = "../../cl" }
+cl                     = { path = "../../cl" }
```
AlejandroCabeza, line 10: Can we use the workspace one?

AlejandroCabeza (approved):
I would advise against using `align_entries=true`. Were one of the entries have a different `=` position, whole chunks of files would change.

That being said I won't oppose if people are fine with that, ofc.

danielSanchezQ:
> I would advise against using `align_entries=true`. Were one of the entries have a different `=` position, whole chunks of files would change.

This is actually a really good point. It looks cleaner aligned, but changes counter will go `brrrrrr`

danielSanchezQ:
> > I would advise against using align_entries=true. Were one of the entries have a different = position, whole chunks of files would change.
> 
> That's true, but since the relative order between the items is not changing, I've never experience the issue where the changeset was very hard to follow. VSCode but also GitHub do a nice job at highlighting that only the space changed between the two versions. Also, this only happens when a new dependency with a longer name or whatever comes in, and that only ever grows, so I don't see it happening _that_ often. But if you both are in favour of reverting that change I could do that. I am in favor of keeping it.

I think we can give it a try. Both work for me. But I prefer aligned though. 
If we see it's bothersome we can change it back.

Btw +1 adding the pre-commit hook @AlejandroCabeza , can be done here or in a different pr if @ntn-x2 is not familiar with it.

ledger/nomos-ledger/Cargo.toml
```diff
@@ -1,26 +1,30 @@
...
+rpds                   = "1"
+serde                  = { version = "1.0", features = ["derive"], optional = true }
+sha2                   = "0.10"
+thiserror              = "1"
 
 [features]
-serde = ["dep:serde", "nomos-utils/serde", "rpds/serde"]
\ No newline at end of file
+serde = [
+  "cryptarchia-engine/serde",
```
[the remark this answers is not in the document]
ntn-x2: Yes, there was a compilation error when trying to compile the `nomos-ledger` crate, as it was assuming this feature was included, which was not, so this change was necessary. I went to each crate and fixed the ones that were not compiling with the `--all-features` flag.

nomos-core/risc0_proofs/ptx/Cargo.toml
ntn-x2: Unfortunately not. These risc0 proofs require their own workspace, and are not part of the top-level workspace (they are also explicitly excluded from it in the top-level `Cargo.toml` file).

ntn-x2:
Screenshot 2025-02-19 at 09 05 47
FYI just su

danielSanchezQ:
> Screenshot 2025-02-19 at 09 05 47
> 
> FYI just came across one: @danielSanchezQ @AlejandroCabeza

I think it should be fine. As I said, if we change opinion later on we can always bring it back without it.

PR #1058, opened 2025-02-18
chore: optimize CI workflows

ntn-x2:

* Removed a couple of actions which we not really used anymore because we do not build/test for windows machines at the moment
* Add three new actions to restore a cache entry, save a new cache entry, and remove an old cache entry
* The caching strategy always uses the latest entry to restore before deleting it, always keeping a single cache entry for each job at any time. There is a limit of 10GB for the whole cache, and this is likely to be hit with 3 concurrent PRs. What takes the most space is the `cargo test` step, and I will in the future investigate if we can make that faster and take up less resources
* Renamed the `build-test` workflow into `code-check` for more clarity
  * It now uses the new caching strategy for both `cargo clippy` and `cargo test`, while `cargo check` is gone since it was overlapping with the `cargo clippy` run
* Gave a name to all steps, to make it easier to describe what they do when looking at them
* The Docker build workflow is now triggered by a `workflow_dispatch` event, which means it can be triggered manually, or only at the end of the check job, if there are no compilation errors or tests failing. I think this is a good thing to do, as we don't want to build potentially broken Docker images.
  *  It now shows up as failing since the new changes are not yet part of the default branch. But I have tried it locally and it seems to work.

Yes.

@ntn-x2 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 1. Description added.


.github/workflows/code-check.yml
```diff
@@ -0,0 +1,160 @@
...
+          key: ${{ github.ref }}->${{ github.workflow }}->${{ github.job }}
+ 
+  tests:
+    name: Run test suite
+    if: ${{ !startsWith(github.event.pull_request.title, '[WIP]') && !contains(github.event.label.name,  'DO NOT MERGE') }}
+    strategy:
+      fail-fast: false # all OSes should be tested even if one fails (default: true)
+      matrix:
+        feature: [ libp2p ]
+        os: [ self-hosted, macos-latest ] # drop windows for now as risc0 does not support it
```
ntn-x2, line 82: Since we have removed the windows OS here, I also removed all checks against the `"windows"` value since they would never be true at this point. That triggered also the removal of two locally-defined actions that were now unused.

.github/workflows/code-check.yml
```diff
@@ -0,0 +1,160 @@
...
+        run: cargo binstall -y cargo-risczero && cargo risczero install
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      # TODO: Investigate how to get rid of this seemingly necessary build step before running the tests
+      - name: Cargo build
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --all --all-targets --no-default-features --features ${{ matrix.feature }}
+      # We store the result of `cargo build` in the cache, that is then re-used in the next run, and can be picked up by the next `cargo test` step.
```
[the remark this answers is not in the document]
ntn-x2: The next commit on this branch. This is a complicated job actually, since the build provides useful stuff for the tests, but then the `bench` command re-builds everything from scratch. At the same time, we have a max of 10GB of cache across all workflows in the repo, so I did not want to create yet another cache for this job. But to answer your question, the cache is read at the beginning of the workflow, deleted, and then a new key is written, which is then read at the next workflow run (i.e., next batch of commits).

.github/workflows/code-check.yml
danielSanchezQ: Makes sense. We will add it eventually. But for now we can remove supporting it.

.github/workflows/code-check.yml
```diff
@@ -0,0 +1,160 @@
...
+    steps:
+      - name: Trigger "Docker build" workflow
+        uses: actions/github-script@v7
+        with:
+          script: |
+            github.rest.actions.createWorkflowDispatch({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              workflow_id: '${{ env.DOCKER_BUILD_WORKFLOW_FILE }}',
+              ref: '${{ github.ref }}',
```
bacv, line 159: From the failed action output it seems that `github.ref` should be changed to the commit hash here.

ntn-x2: In the codebase I was working on before (https://github.com/KILTprotocol/kilt-node/blob/67c3d65a8adb6b0ac9578492d882c1f9f117fcdd/.github/workflows/check-code.yml#L381), I did exactly the same for a job that is started on both PRs and push events. I believe the error is due to the fact that the new `workflow_dispatch` configuration for the Docker build workflow is not part of the default branch just yet, otherwise I would not be able to explain why this would not work while in the other repo it does.

ntn-x2:
@bacv @danielSanchezQ sorry for the commit spam. The PR is now ready for review. The last working workflow run is: https://github.com/logos-co/nomos-node/actions/runs/13455074805/job/37597206546.

.github/workflows/codecov.yml
```diff
@@ -14,45 +14,49 @@ jobs:
       LLVM_PROFILE_FILE: "nomos-node-%p-%m.profraw"
     runs-on: self-hosted
     steps:
-      - uses: actions/checkout@v4
-        with:
-          submodules: true
-      - uses: arduino/setup-protoc@v3
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Install protoc compiler
```
bacv, line 19: We can remove this step, protobufs were removed from DA.

ntn-x2: For all jobs?

bacv: Yes, no use of protobufs in any part of the node at the moment.

ntn-x2: Good to know! https://github.com/logos-co/nomos-node/pull/1058/commits/6d09c4880d77a8959c7fbba814e6435ed66c6ec5

PR #1060, opened 2025-02-18
feat(DA): Connection monitor behavior

bacv:

Network behaviour for monitoring failures in other sub-behaviours. 
There are two parts:

- `ConnectionMonitorBehaviour` - operates on the provided `ConnectionMonitor` trait implementation and handles disconnects or temporal disconnects of `Malicious` and `Unhealthy` peers.
  - Malicious peers are disconnected and can only be unblocked by calling `unblock_peer` method externally.
  - Unhealthy peers are disconnected, but will be unblocked for dials after the `redial_cooldown` duration has passed.

- `DAConnectionMonitor` - A `ConnectionMonitor` implementation for evaluating Nomos DA peer health based on error rates across three stream protocols.
  - Since there is no formal specification for identifying malicious peers, this implementation treats a consistently high failure rate in any subprotocol as malicious behavior, while an unhealthy peer exhibits occasional errors.
  - Only error events are considered in this evaluation.
  - Implements an exponential weighted moving average to track failure rates over time, applying exponential decay to gradually reduce the impact of older errors.

Connection Monitoring behaviour is only for tracking established connections and preventing new connections to the unhealthy or malicious nodes. 

Connection maintenance behaviour, which will maintain a consistent number connection to the number of subnetwork peers according to the policy will be implemented in a separate PR.

Yes

@bacv 

N/A - might require a separate specification for malicious and unhealthy peer definitions

N/A

## Checklist

* [x] 1. Description added.


nomos-da/network/core/src/maintenance/monitor.rs
```diff
@@ -0,0 +1,373 @@
...
+pub trait ConnectionMonitor {
+    type Event;
+
+    fn record_event(&mut self, event: Self::Event) -> Option<ConnectionMonitorOutput>;
+    fn reset_peer(&mut self, peer_id: &PeerId);
+}
+
+/// A `NetworkBehaviour` that maintains consistent number of  connections to DA subnetwork nodes.
+#[derive(Debug)]
+pub struct ConnectionMonitorBehaviour<Monitor> {
```
danielSanchezQ, line 41: I think this is clean

nomos-da/network/core/src/maintenance/monitor.rs
```diff
@@ -0,0 +1,373 @@
...
+    peer: PeerId,
+}
+
+impl fmt::Display for Blocked {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "peer {} is in the block list", self.peer)
+    }
+}
+
+impl std::error::Error for Blocked {}
```
danielSanchezQ, line 153: Nitpick: Maybe `thiserror`?

nomos-da/network/core/src/maintenance/monitor.rs
```diff
@@ -0,0 +1,373 @@
...
+    ///
+    /// Returns whether the peer was present in the set. Does nothing if the peer
+    /// was not present in the set.
+    pub fn unblock_peer(&mut self, peer: PeerId) -> bool {
+        let removed =
+            self.malicous_peers.remove(&peer) || self.unhealthy_peers.remove(&peer).is_some();
+        if removed {
+            if let Some(waker) = self.waker.take() {
+                waker.wake()
+            }
```
danielSanchezQ, line 106: Probably we want this as a `try_wake` method since we use it in a few places.

nomos-da/network/core/src/swarm/common/monitor.rs
```diff
@@ -0,0 +1,387 @@
...
+            SamplingEvent::SamplingSuccess { .. } => MonitorEvent::Noop,
+            SamplingEvent::IncomingSample { .. } => MonitorEvent::Noop,
+            SamplingEvent::SamplingError { error } => MonitorEvent::Sampling(error.clone()),
+        }
+    }
+}
+
+#[derive(Default, Debug)]
+pub struct PeerStats {
+    // Failure rate tracking (sliding window)
```
danielSanchezQ, line 84: why sliding window? I don't understand the comment

nomos-da/network/core/src/swarm/common/monitor.rs
```diff
@@ -0,0 +1,387 @@
...
+                self.settings.failure_time_window,
+                self.settings.time_decay_factor,
+            );
+
+            if dispersal_rate >= self.settings.malicious_threshold
+                || sampling_rate >= self.settings.malicious_threshold
+                || replication_rate >= self.settings.malicious_threshold
+            {
+                return PeerStatus::Malicious;
+            }
```
danielSanchezQ, line 190: I would extract this as a method. In fact, this should be part of the policies abstraction we talked about (?)

nomos-da/network/core/src/swarm/common/monitor.rs
```diff
@@ -0,0 +1,387 @@
...
+
+            if dispersal_rate >= self.settings.max_dispersal_failures
+                || sampling_rate >= self.settings.max_sampling_failures
+                || replication_rate >= self.settings.max_replication_failures
+            {
+                return PeerStatus::Unhealthy;
+            }
```
danielSanchezQ, line 197: Likewise

danielSanchezQ (commented):
I think this is going where it should go. Structure looks super clean. Just added a few comments, but really nice job!

nomos-da/network/core/src/swarm/common/monitor.rs
bacv: Thats a good idea actually, the policy will be shared by the monitor and maintenance behaviours, and this implementation will provide stats to policy for evaluation. Will add a todo comment for now, implement connection policy in upcoming pr.

nomos-da/network/core/src/maintenance/monitor.rs
bacv: Updated

nomos-da/network/core/src/swarm/common/monitor.rs
bacv: Added a comment to the PeerStats struct and explanation to it's field types

tests/Cargo.toml
```diff
@@ -26,6 +27,7 @@ nomos-cli                = { workspace = true }
 nomos-core               = { workspace = true }
 nomos-da-dispersal       = { workspace = true }
 nomos-da-indexer         = { workspace = true }
+nomos-da-network-core    = { path = "../nomos-da/network/core" }
```
ntn-x2, line 30: Can you switch this to be a workspace dependency instead?

nomos-da/network/core/src/maintenance/monitor.rs
```diff
@@ -0,0 +1,356 @@
...
+    pub fn unblock_peer(&mut self, peer: PeerId) -> bool {
+        let removed =
+            self.malicous_peers.remove(&peer) || self.unhealthy_peers.remove(&peer).is_some();
+        if removed {
+            self.try_wake();
+        }
+        removed
+    }
+
+    pub fn record_event(&mut self, event: Monitor::Event) {
```
ntn-x2, line 112: Wouldn't this be the only method that requires `Monitor` to implement `ConnectionMonitor`? Would it make sense to provide all the other methods for all the possible generics? I think it would make unit testing easier, but I do not have a strong opinion on this. I usually tend to do it, especially when the trait bound applies to a single method.

ntn-x2 (commented):
I just left a couple of minor comments, as the rest is still quite much above my head, but I took the chance to try to make a better picture in my head.

nomos-da/network/core/src/maintenance/monitor.rs
bacv: Just to clarify, the suggested modification would look something like this?:
```
impl<Monitor> ConnectionMonitorBehaviour<Monitor> {
    pub fn new(monitor: Monitor, redial_cooldown: Duration) -> Self
    fn try_wake(&mut self)
    ....
}

impl<Monitor> ConnectionMonitorBehaviour<Monitor>
where
    Monitor: ConnectionMonitor,
{
    pub fn record_event(&mut self, event: Monitor::Event) {...}
}
```

I too don't have a strong opinion on this, for simplicity I would prefer to have it one block, but wouldn't mind have it split too, just that at the moment I don't see reason for it.

nomos-da/network/core/src/swarm/common/monitor.rs
```diff
@@ -0,0 +1,391 @@
...
+            self.last_replication_failure.unwrap_or(now),
+            self.replication_failures_rate,
+            window,
+            factor,
+        )
+    }
+}
+
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+pub struct DAConnectionMonitorSettings {
```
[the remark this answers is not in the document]
bacv: Good catch, will do.

nomos-da/network/core/src/maintenance/monitor.rs
```diff
@@ -0,0 +1,356 @@
...
+        // Close existing connections
+        self.close_connections.push_back(peer);
+        self.try_wake();
+    }
+
+    /// Unblock connections to a given peer.
+    ///
+    /// Returns whether the peer was present in the set. Does nothing if the peer
+    /// was not present in the set.
+    pub fn unblock_peer(&mut self, peer: PeerId) -> bool {
```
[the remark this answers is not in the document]
bacv: This function is supposed to be used by external services in the future, probably could remove it now and add later once this functionality is required.

AlejandroCabeza (commented):
lgtm after all the suggested changes. I'll leave the final approval to those who proposed those changes.

nomos-da/network/core/src/maintenance/monitor.rs
danielSanchezQ: Difference is if `ConnectionMonitorBehaviour` would work with anything that is not a `ConnectionMonitor`. This way we enforce the behaviour not to work under any other bound. But its true that for testing purposes relaxing that would be better. We cannot test the other methods separetely unless we have a mock `ConnectionMonitor`. But since that is tested already it should be fine. Whatever you feel in this case @bacv. But good point @ntn-x2 .

danielSanchezQ (approved):

PR #1062, opened 2025-02-18
feat(TimeService): Implement time service and integrate with cryptarchia

danielSanchezQ:

This PR implements a time service and two backends. First a SystemTime backend that uses the system time and replaces current implementation. Then an `NTP` which uses centralized ntp servers to sync the system clock. 

It looks like a big change but it just spreads a lot of generics because it touches consensus.
It also modifies tests as configuration now loads differently.

@danielSanchezQ 

Not related to any specification yet.

N/A

## Checklist

* [x] 1. Description added.


nomos-services/time/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+edition = "2021"
+license = { workspace = true }
+name    = "nomos-time"
+version = "0.1.0"
+
+[dependencies]
+async-trait        = "0.1"
+cryptarchia-engine = { path = "../../consensus/cryptarchia-engine", features = ["serde", "tokio"] }
```
ntn-x2, line 9: Sorry I just randomly saw this, which can be changed to a workspace dependency 😄

danielSanchezQ: True, I changed a few but some may have missed during the rebase. Thanks!

consensus/cryptarchia-engine/src/time.rs
```diff
@@ -16,6 +19,20 @@ impl Slot {
     pub fn genesis() -> Self {
         Self(0)
     }
+
+    pub fn current_from_offset_and_config(
```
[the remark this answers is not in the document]
danielSanchezQ: Yeah. Makes sense. Will change.

consensus/cryptarchia-engine/src/time.rs
```diff
@@ -57,3 +74,71 @@ impl Add<u32> for Epoch {
...
+    pub fn new(config: SlotConfig) -> Self {
+        SlotTimer { config }
+    }
+
+    pub fn current_slot(&self) -> Slot {
+        Slot::current_from_offset_and_config(OffsetDateTime::now_utc(), self.config)
+    }
+
+    /// Ticks at the start of each slot, starting from the next slot
+    pub fn slot_interval(&self, now: OffsetDateTime) -> Interval {
```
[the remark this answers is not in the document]
danielSanchezQ: This algorithm had the `now` hardcoded to `OffsetDateTime::now_utc()`. This was just parametrized so we can decide where to that `OffsetDateTime` comes from. It may come from the `now_utc` or some external method like in the ntp backend.
The now is used for calculating how much time is pending till the next slot tick (the one that will start the tick). The `tokio::time::Instant::now` in line 137 is just used to offset that delay from the current `Instant`.

nomos-services/time/src/backends/ntp/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub timeout: Duration,
+    pub update_interval: Duration,
+    pub connect_timeout: Duration,
+    pub slot_config: SlotConfig,
+    pub epoch_config: EpochConfig,
+    pub base_period_length: u64,
+    pub ntpclient_settings: NTPClientSettings,
+}
+pub struct Ntp {
+    settings: NtpSettings,
```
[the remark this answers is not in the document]
danielSanchezQ: Maybe, but it wouldn't change nothing. This is a final implementation that depends on nothing else. There is no need to tie up any type as it is the last layer and dependent solely on itself imho.

consensus/cryptarchia-engine/src/time.rs
```diff
@@ -57,3 +74,71 @@ impl Add<u32> for Epoch {
...
+}
+
+#[cfg(feature = "tokio")]
+impl SlotTimer {
+    pub fn new(config: SlotConfig) -> Self {
+        SlotTimer { config }
+    }
+
+    pub fn current_slot(&self) -> Slot {
+        Slot::current_from_offset_and_config(OffsetDateTime::now_utc(), self.config)
```
bacv, line 127: This is minor, but for consistency I think it would be better to pass `now` here too, since system time and the one from the `ntp` backend might be different.

danielSanchezQ: Actually, I may rename it to `current_slot_from_utc`. But good catch indeed.

nomos-services/time/src/backends/ntp/mod.rs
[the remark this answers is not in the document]
AlejandroCabeza: Right, makes sense.

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -421,12 +433,17 @@ where
         )
         .await;
 
+        let time_relay: OutboundRelay<_> = self
```
AlejandroCabeza, line 436: It'd be nice if you could add this to the `CryptarchiaConsensusRelays` struct.

Speaking of which, does it make sense to also move the block broadcaster into it?

danielSanchezQ:
Uhm. @ntn-x2 This branch is rebased and seems like the docker build failed. I think it just happens sometimes.

ntn-x2:
Mhh ok, I will check on that first thing next week. It should not happen, and as far as I can see in other PRs it does not fail all the times...

nomos-services/time/src/backends/common.rs
```diff
@@ -0,0 +1,27 @@
...
+use cryptarchia_engine::time::SlotTimer;
+use cryptarchia_engine::{EpochConfig, Slot, SlotConfig};
+use futures::StreamExt;
+use std::pin::Pin;
+use time::OffsetDateTime;
+use tokio_stream::wrappers::IntervalStream;
+
+pub(crate) fn slot_timer(
+    slot_config: SlotConfig,
+    date: OffsetDateTime,
```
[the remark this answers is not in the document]
danielSanchezQ: done

nomos-services/cryptarchia-consensus/src/lib.rs
danielSanchezQ: In reality this relay is consumed and replaced by the `slot_stream` and not used by anything else. So probably not work to keep a reference to it to be sent to other methods if there is no need. We can add it later on if we add more functionality to it.

bacv (approved):
🚀

nomos-services/time/src/backends/ntp/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub slot_config: SlotConfig,
+    /// Epoch settings in order to compute proper epoch times
+    pub epoch_config: EpochConfig,
+    /// Base period length related to epochs, used to compute epochs as well
+    pub base_period_length: u64,
+}
+pub struct Ntp {
+    settings: NtpSettings,
+    client: AsyncNTPClient,
+}
```
danielSanchezQ, line 35: This is not tested. But as this PR went out of hand with changes already....it will be worked in a different pr.

AlejandroCabeza (approved):

consensus/cryptarchia-engine/src/time.rs
```diff
@@ -57,3 +75,71 @@ impl Add<u32> for Epoch {
...
+    // snapshot has stabilized to take the nonce snapshot.
+    pub epoch_period_nonce_buffer: u8,
+    // This parameter controls how many slots we wait for the nonce snapshot to be considered
+    // stabilized
+    pub epoch_period_nonce_stabilization: u8,
+}
+
+impl EpochConfig {
+    pub fn epoch_length(&self, base_period_length: u64) -> u64 {
+        (self.epoch_stake_distribution_stabilization as u64
```
ntn-x2, line 96: This should be replaced with checked variants, so that we don't risk overflows in release builds, which don't panic in those cases.

ntn-x2: There's also a clippy lints that checks for this, maybe we want to introduce it at some point...

consensus/cryptarchia-engine/src/time.rs
```diff
@@ -57,3 +75,71 @@ impl Add<u32> for Epoch {
...
+            + self.epoch_period_nonce_buffer as u64
+            + self.epoch_period_nonce_stabilization as u64)
+            * base_period_length
+    }
+
+    pub fn epoch(&self, slot: Slot, base_period_length: u64) -> Epoch {
+        ((u64::from(slot) / self.epoch_length(base_period_length)) as u32).into()
```
ntn-x2, line 103: Here as well, there's a division which is risky, and some potentially unsafe casts as well.

nomos-services/time/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+services-utils     = { workspace = true }
+sntpc              = { version = "0.5", optional = true }
+thiserror          = { version = "2.0", optional = true }
+time               = "0.3"
+tokio              = { version = "1", features = ["macros", "sync"] }
+tokio-stream       = "0.1"
+url                = "2.5"
+
+[features]
+default = []
```
ntn-x2, line 24: I guess we don't need to specify a default feature if it's empty?

nomos-services/time/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+ntp     = ["sntpc/tokio-socket", "thiserror/default"]
```
ntn-x2, line 25: I guess we want to use the `dep:` prefix for these deps, as otherwise we would be exposing two additional features called `sntpc` and `thiserror` which I believe we don't want? https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies

ntn-x2 (commented):
I think in general we should reduce potential arithmetic errors that can go unnoticed in release builds. So I wanted to take this review as a chance to ask whether we want to introduce relevant Clippy lints to avoid these errors?

consensus/cryptarchia-engine/src/time.rs
danielSanchezQ: > There's also a clippy lints that checks for this, maybe we want to introduce it at some point...

Yes, we want. Anything that helps catch things earlier the better.

nomos-services/time/Cargo.toml
danielSanchezQ: True. I forgot about this actually. That also removed the need for the default feature. Thanks!

consensus/cryptarchia-engine/src/time.rs
danielSanchezQ: > This should be replaced with checked variants, so that we don't risk overflows in release builds, which don't panic in those cases.

I added `NonZero` wherever possible as well. As it makes sense. It should force values from config to be valid as well. Thanks!

consensus/cryptarchia-engine/src/time.rs
danielSanchezQ: Implemented `try_into`. As this is dependent on config I just basically unwrapped it here and panic but I think its more explicit now.

danielSanchezQ:
@zeegomo I would like you take a look at this if possible.

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -167,7 +167,7 @@ where
     }
 
     pub fn fork_choice(&self) -> Branch<Id> {
-        let k = self.config.security_param as u64;
+        let k = self.config.security_param.get() as u64;
```
ntn-x2, line 170: Not a problem here, but this is what my more pedantic Clippy tells me about this:

> casts from `u32` to `u64` can be expressed infallibly using `From`
an `as` cast can become silently lossy if the types change in the future`

consensus/cryptarchia-engine/src/time.rs
```diff
@@ -16,6 +21,20 @@ impl Slot {
...
+        offset_date_time: OffsetDateTime,
+        slot_config: SlotConfig,
+    ) -> Self {
+        // TODO: leap seconds / weird time stuff
+        let since_start = offset_date_time - slot_config.chain_start_time;
+        if since_start.is_negative() {
+            // current slot is behind the start time, so return default 0
+            Slot::genesis()
+        } else {
+            Slot::from(since_start.whole_seconds() as u64 / slot_config.slot_duration.as_secs())
```
ntn-x2, line 35: This could still underflow, right? Are we leaving them like this, and address them in a different PR where we introduce Clippy lints?

consensus/cryptarchia-engine/src/time.rs
```diff
@@ -57,3 +84,73 @@ impl Add<u32> for Epoch {
...
+    // snapshot has stabilized to take the nonce snapshot.
+    pub epoch_period_nonce_buffer: NonZero<u8>,
+    // This parameter controls how many slots we wait for the nonce snapshot to be considered
+    // stabilized
+    pub epoch_period_nonce_stabilization: NonZero<u8>,
+}
+
+impl EpochConfig {
+    pub fn epoch_length(&self, base_period_length: NonZero<u64>) -> u64 {
+        ((self.epoch_stake_distribution_stabilization.get()
```
ntn-x2, line 105: This can also overflow silently. But I guess we'll tackle them in a different PR.

nodes/nomos-node/src/config.rs
```diff
@@ -126,6 +121,15 @@ pub struct CryptarchiaArgs {
     note_nonce: Option<String>,
 }
 
+#[derive(Parser, Debug, Clone)]
+pub struct TimeArgs {
+    #[clap(long = "consensus-chain-start", env = "CONSENSUS_CHAIN_START")]
+    chain_start_time: Option<i64>,
```
ntn-x2, line 127: Why would this not be a `u64` instad of a `i64`?

ntn-x2: If there's some baked in assumption, it would be good to describe it in a comment.

consensus/cryptarchia-engine/src/time.rs
danielSanchezQ: No, I would prefer to address them here with context. Problem is the same, values should come well from config.

consensus/cryptarchia-engine/src/time.rs
danielSanchezQ: @ntn-x2 , I have some changes that will address the limits from config so this operations should be safe. But it may grow this too much and go out of scope. I think I would add it later in a further PR.

nodes/nomos-node/src/config.rs
danielSanchezQ: Because unix timestamps format are `i64` for the used `OffsetDateTime`.

consensus/cryptarchia-engine/src/time.rs
danielSanchezQ: This was because of a mix of older version with the non-zero. Should be fine now.

consensus/cryptarchia-engine/src/lib.rs
danielSanchezQ: Nice catch by pedantic Clippy. Fixed.

nodes/nomos-node/src/config.rs
ntn-x2: I think it would still make sense to include a comment explaining what a negative value means. Or is that specified elsewhere?

consensus/cryptarchia-engine/src/time.rs
danielSanchezQ: For reference, changes will be integrated later [here](https://github.com/logos-co/nomos-node/pull/1090)

nodes/nomos-node/src/config.rs
danielSanchezQ: Well, as its built from a timestamp in UNIX I actually assumed a negative value would mean datetimes previous to that. iirc it works like that (?)

ntn-x2 (approved):
Nice!

Issue #1072, opened 2025-02-24
Store DA blobs in RocksDb


andrussal:
Currently blobs are stored in filesystem because of the concern that RocksDb is not suitable of storing large blobs. We think that in our case blob size fits into range where RocksDb works well(4 KB to 1 MB). In addition, most of our reads should be targeted to recently inserted keys/values which are still in memory when reading, so avoiding potential issues with large blobs and disk access.

PR #1075, opened 2025-02-24
refactor(DA): Introduce LightBlob and SharedCommitments types

andrussal:

It is part of the work to provide HTTP API to fetch DA blob common commitments.

This PR Introduces `DaLightBlob` and `DaBlobSharedCommitments` types that are stored in RocksDb. `DaBlobSharedCommitments` contains shared commitments from current `DaBlob` and `DaLightBlob` contains the rest.

It replaces current logic where `DaBlob` is stored in filesystem. 

It implements issue: #1072 

@andrussal 

N/A

N/A

## Checklist

* [x] 1. Description added.


nomos-services/data-availability/verifier/src/storage/adapters/rocksdb.rs
```diff
@@ -15,14 +14,15 @@ use overwatch_rs::{
     services::{relay::OutboundRelay, ServiceData},
     DynError,
 };
-// internal
-use crate::storage::DaStorageAdapter;
+// std
```
bacv, line 17: Nit: throughout the codebase we try to group modules in order of std, crates and internal (with respective comments). Unfortunately we don't have a linter for that.

bacv (changes requested):
Looks great and ready for the updated http api. Just few comments to remove `fs.rs` stuff and imports formatting.

nomos-da/kzgrs-backend/src/common/blob.rs
```diff
@@ -58,11 +58,17 @@ impl DaBlob {
     pub fn column_len(&self) -> usize {
         self.column.as_bytes().len()
     }
+
+    pub fn idx(&self) -> [u8; 34] {
+        create_blob_idx(&self.id(), &self.column_idx.to_be_bytes())
+    }
```
danielSanchezQ, line 64: Careful. This is strictly related to the column indexes. We do not need to mix with the id.

nomos-core/chain-defs/src/da/blob/mod.rs
```diff
@@ -5,9 +5,17 @@ pub mod select;
...
 
     fn id(&self) -> Self::BlobId;
     fn column_idx(&self) -> Self::ColumnIndex;
+    fn into_blob_and_shared_commitments(self) -> (Self::LightBlob, Self::SharedCommitments);
+
+    fn from_blob_and_shared_commitments(
+        light_blob: Self::LightBlob,
+        shared_commitments: Self::SharedCommitments,
+    ) -> Self;
 }
```
danielSanchezQ, line 19: I wasn't so sure about this construction. But it is actually a good solution.

nomos-da/kzgrs-backend/src/common/blob.rs
```diff
@@ -71,4 +77,80 @@ impl blob::Blob for DaBlob {
...
+}
+
+// Combines a 32-byte blob ID (`[u8; 32]`) with a 2-byte column index
+// (`u16` represented as `[u8; 2]`).
+pub fn create_blob_idx(blob_id: &[u8], column_idx: &[u8]) -> [u8; 34] {
+    let mut blob_idx = [0u8; 34];
+    blob_idx[..blob_id.len()].copy_from_slice(blob_id);
+    blob_idx[blob_id.len()..].copy_from_slice(column_idx);
+
+    blob_idx
```
danielSanchezQ, line 155: Probably not needed because of the comment above.

nomos-da/storage/src/rocksdb/mod.rs
```diff
@@ -2,6 +2,9 @@ use bytes::{Bytes, BytesMut};
 
 pub const DA_VID_KEY_PREFIX: &str = "da/vid/";
 pub const DA_VERIFIED_KEY_PREFIX: &str = "da/verified/";
+pub const DA_BLOB_PATH: &str = "bl";
+pub const DA_SHARED_COMMITMENTS_PATH: &str = "sc";
+pub const DA_ATTESTATION_PATH: &str = "at";
```
danielSanchezQ, line 7: We should remove any attestation leftover as we do not really have them anymore.

nomos-services/data-availability/indexer/src/storage/adapters/rocksdb.rs
```diff
@@ -128,19 +129,69 @@ where
...
+                    S::deserialize::<DaBlobSharedCommitments>(shared_commitments)
+                        .expect("Failed to deserialize shared commitments");
+
+                let da_blobs: Vec<_> = deserialized_blobs
+                    .into_iter()
+                    .map(|blob| {
+                        DaBlob::from_blob_and_shared_commitments(
+                            blob,
+                            deserialized_shared_commitments.clone(),
+                        )
```
danielSanchezQ, line 190: Maybe this should not have changed. It may be an attempt to optimize that may not work. Reasons are:
* Usually io is gonna be the bottle neck. We are transforming a single serialize/write  into a double one. For each blob is gonna be 2 writes and 2 reads, plus the same serializations/deserializations. I think the easiest way for now would have been to store the whole thing and just what needed from it. `DaBlob` -> `LightBlob` or `DaBlob` -> `DaBlobSharedCommitments`.

nomos-services/data-availability/sampling/src/storage/adapters/rocksdb.rs
```diff
@@ -55,45 +53,45 @@ where
...
+        let blob = match (blob_reply_rx.await?, sc_reply_rx.await?) {
+            (Some(blob), Some(shared_commitments)) => Some(Blob::from_blob_and_shared_commitments(
+                S::deserialize(blob).expect("Failed to deserialize blob"),
+                S::deserialize(shared_commitments)
+                    .expect("Failed to deserialize shared commitments"),
+            )),
+            _ => None,
+        };
+
+        Ok(blob)
```
danielSanchezQ, line 90: Yes, this re-affirms my motives in the previous comment.

nomos-services/data-availability/sampling/src/storage/adapters/rocksdb.rs
```diff
@@ -55,45 +53,45 @@ where
...
-
-// Combines a 32-byte blob ID (`[u8; 32]`) with a 2-byte column index
-// (`u16` represented as `[u8; 2]`).
-fn create_blob_idx(blob_id: &[u8], column_idx: &[u8]) -> [u8; 34] {
-    let mut blob_idx = [0u8; 34];
-    blob_idx[..blob_id.len()].copy_from_slice(blob_id);
-    blob_idx[blob_id.len()..].copy_from_slice(column_idx);
-
-    blob_idx
-}
```
danielSanchezQ, line 99: Ok, so here comes the confusion with the thing before. Where was this used before? @bacv is this indexing related to the indexer?

nomos-services/data-availability/verifier/src/storage/adapters/rocksdb.rs
danielSanchezQ: Neither can be auto fmt iirc. Maybe @ntn-x2 have some magic trick related to that?

nomos-services/data-availability/verifier/src/storage/adapters/rocksdb.rs
```diff
@@ -56,25 +56,45 @@ where
...
+
+        // Store the attestation in the storage backend.
+        let attestation_prefix = format!("{}{}", DA_VERIFIED_KEY_PREFIX, DA_ATTESTATION_PATH);
+        let attestation_key = key_bytes(&attestation_prefix, blob_id);
+        self.storage_relay
+            .send(StorageMsg::Store {
+                key: attestation_key,
                 value: S::serialize(attestation),
             })
             .await
```
danielSanchezQ, line 100: Here they are the 2 writes :sweat_smile:

nomos-services/data-availability/verifier/src/storage/adapters/rocksdb.rs
```diff
@@ -109,16 +132,6 @@ where
...
-// Combines a 32-byte blob ID (`[u8; 32]`) with a 2-byte column index
-// (`u16` represented as `[u8; 2]`).
-fn create_blob_idx(blob_id: &[u8], column_idx: &[u8]) -> [u8; 34] {
-    let mut blob_idx = [0u8; 34];
-    blob_idx[..blob_id.len()].copy_from_slice(blob_id);
-    blob_idx[blob_id.len()..].copy_from_slice(column_idx);
-
-    blob_idx
-}
-
```
danielSanchezQ, line 121: Why did we had this so duplicated everywhere. Any hint @bacv ?

nomos-services/data-availability/verifier/src/storage/adapters/rocksdb.rs
ntn-x2: I think the formatter will maintain relative order if the different groups are spaced with a new line. Honestly I don't see the benefit of adding the comments to signal the different sections. Once we add a new line, any new import will automatically be added to the right section. [This fmt rule](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#StdExternalCrate%5C%3A) is what we could try with. I think that would already be a big step ahead without the need for these custom solution. If you guys agree, I could open a new PR with that and perhaps a few other useful formatting rules, like perhaps [this one](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#Crate%5C%3A).

danielSanchezQ: Yeah, those rules may work I think.  wdyt guys? @bacv @AlejandroCabeza @youngjoon-lee @andrussal

ntn-x2: Sorry I saw the links did not point to the exact section I was referring to. The first link is for the `StdExternalCrate` rule, while the second is for the `imports_granularity = "Crate"` setting.

danielSanchezQ: > Sorry I saw the links did not point to the exact section I was referring to. The first link is for the `StdExternalCrate` rule, while the second is for the `imports_granularity = "Crate"` setting.

Yeah, I just searched them manually from the link hints :sweat_smile:

nomos-services/data-availability/sampling/src/storage/adapters/rocksdb.rs
andrussal: Here I am thinking if it makes sense to provide 2 functions in the next PR where I implement HTTP API? One function retrieves only `LightBlob` and another only `SharedCommitments`?

danielSanchezQ: This adapter is just sampling. I'm not sure we have to pull it from this adapter for the HTTP API. Also the HTTP API endpoint is just for the `SharedCommitments`. So:
* Here we just need a method for the `LightBlob` I think.
* In the endpoint one method for the `ShareCommiments`.

nomos-services/data-availability/indexer/src/storage/adapters/rocksdb.rs
andrussal: `I think the easiest way for now would have been to store the whole thing and just what needed from it.`

Do you mean that we should store `DaBlob` fully in RocksDb and functions return only part of it what they need by extracting it from `DaBlob`? 

Basically we have some storage space overhead by storing shared commitments multiple times but we still implement network level optimisation to transfer either `DaLightBlob` or `DaSharedCommitments`? 

One thing that comes to mind is that we store `DaBlob` only once but read multiple times. Maybe 2 writes per storing is good compromise if we later can read only `LightBlob` or `DaBlob` and not full `DaBlob`.

danielSanchezQ: Agh, it's true. You are absolutely right. We will be duplicating the shared commitments in-store. Yeah, it's a no-go. Just forget what I said actually.

nomos-services/data-availability/verifier/src/storage/adapters/rocksdb.rs
danielSanchezQ: > I am also uncomfortable with it tbh.
> 
> At the same time it replaced writing to the file system. (Also related is comment above about one write(in the sense of storing only once) and reading about multiple times more efficiently).

I'm not sure if its supported by rocksdb but maybe we can try doing a transaction instead? We have an execute method in the storage service API.

nomos-services/data-availability/sampling/src/storage/adapters/rocksdb.rs
andrussal: Makes sense. Actually I wasn't sure what is correct service to get `SharedCommitments`. I was thinking maybe `verifier` because it stores this data, so it felt like a symmetrical operation? What do you suggest @danielSanchezQ ?

nomos-services/data-availability/verifier/src/storage/adapters/rocksdb.rs
andrussal: Good point. I'll take a look. This should be atomic indeed.

nomos-services/data-availability/verifier/src/storage/adapters/rocksdb.rs
bacv: Because it is related to the rocks db adapters only, didn't seem to fit in `nomos-da` core. To have rocksdb specific code in `kzgrs-backend` isn't also ideal, but better for managing. @andrussal maybe lets add a comment about usage of `create_blob_idx` or a more explicit name?

danielSanchezQ: > Because it is related to the rocks db adapters only, didn't seem to fit in `nomos-da` core. To have rocksdb specific code in `kzgrs-backend` isn't also ideal, but better for managing. @andrussal maybe lets add a comment about usage of `create_blob_idx` or a more explicit name?

We have a utils crate maybe can be moved there?
Also, @bacv, this is just for computing the key of the query right? Because in this pr is being use for the `blob_idx` which should be related to the subnetwork/column iirc?

nomos-da/kzgrs-backend/src/common/blob.rs
andrussal: Now it's clear to me that this function doesn't belong here as it's strictly storage related.  Does it make sense to move it to `nomos-da/storage/src/rocksdb/mod.rs`?

nomos-da/storage/src/rocksdb/mod.rs
andrussal: Based on above comment, I have been removing attestation code that is directly affected by my changes. It's probably better to leave full attestation "cleanup" to some other PR in order not to make this PR too big?

danielSanchezQ: > Based on above comment, I have been removing attestation code that is directly affected by my changes. It's probably better to leave full attestation "cleanup" to some other PR in order not to make this PR too big?

Yeah, full attestation cleanup should be done separately.

bacv: Yes, that would be better

nomos-da/kzgrs-backend/src/common/blob.rs
bacv: That would be a good place and protected by rocksdb-backend feature.

nomos-services/data-availability/verifier/src/storage/adapters/rocksdb.rs
AlejandroCabeza: IIRC A couple months ago we talked about this exact options (`StdExternalCrate` and  `crate`) being ideal but still in nightly or beta. If they are released then let's go.
If they're not, removing the comments is still a good idea as that simplifies adoption of those options when they are released.

nomos-services/data-availability/sampling/src/storage/adapters/rocksdb.rs
```diff
@@ -55,45 +52,45 @@ where
         blob_id: <Self::Blob as Blob>::BlobId,
         column_idx: ColumnIndex,
     ) -> Result<Option<Self::Blob>, DynError> {
-        let column_idx = column_idx.to_be_bytes();
-        let blob_idx = create_blob_idx(blob_id.as_ref(), &column_idx);
+        let blob_idx = create_blob_idx(blob_id.as_ref(), column_idx.to_be_bytes().as_ref());
 
-        let (reply_tx, reply_rx) = tokio::sync::oneshot::channel();
+        let blob_prefix = format!("{}{}", DA_VERIFIED_KEY_PREFIX, DA_BLOB_PATH);
```
AlejandroCabeza, line 57: In this file as well.

nomos-services/data-availability/verifier/src/storage/adapters/rocksdb.rs
ntn-x2: You are right, it seems they are still no nightly. Nevertheless, I don't see how that would be a blocker? We can use a nightly toolchain in the CI for the fmt job, and there's way to also configure local machines to run a nightly version as the default formatter. We've done that elsewhere in the past. Would you guys be ok with that? I don't think using nightly for formatting has any security or important implications whatsoever.

nomos-services/data-availability/indexer/src/storage/adapters/rocksdb.rs
```diff
@@ -48,26 +49,24 @@ where
...
             _vid: PhantomData,
         }
     }
 
     async fn add_index(&self, info: &Self::Info) -> Result<(), DynError> {
-        let (app_id, idx) = info.metadata();
-
         // Check if Info in a block is something that the node've seen before.
-        let attested_key = key_bytes(DA_VERIFIED_KEY_PREFIX, info.blob_id());
+        let blob_prefix = format!("{}{}", DA_VERIFIED_KEY_PREFIX, DA_BLOB_PATH);
```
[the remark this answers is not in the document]
andrussal: Do you suggest providing it as function argument? If that's the case, I am not sure about it... It might be better to keep it as storage mechanism internal detail. Or do you mean like providing some helper function that all callers can use?

nomos-services/data-availability/sampling/src/storage/adapters/rocksdb.rs
andrussal: Moved into `nomos-da/storage/src/rocksdb/mod.rs`

nomos-services/data-availability/verifier/src/storage/adapters/rocksdb.rs
andrussal: We decided to address this in a separate PR

bacv (approved):
Looks great! Thanks!

PR #1078, opened 2025-02-25
Provide `store_batch` function to storage backend in order to save batch of values atomically

andrussal:

**Provide `store_batch` function to storage backend in order to save batch of values atomically.**

Not sure if this is the right way to go. Just opened for comments.

The reason of creating this change is related to this PR where we should store multiple keys atomically: #1075 

### Justification why this function is needed in addition to existing storage interface.

Current storage interface provides general purpose `execute` function that allows to run arbitrary code within a context of a transaction. In order to call this functions, caller needs to provide backend specific transaction instance. It seems that currently it’s not straightforward to create such Transaction from an OverWatch service. 

Another justification to add this function to storage interface is that this should be a fairly common use case how transactions are used in the system - storing batch of values together.

Alternately could try to refactor current `execute` solution so that it doesn’t require passing backend specific Transaction instance. I started looking into it but this seems to be much larger effort than providing batch function directly.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 1. Description added.

### DIFF
--- a/nomos-services/storage/src/backends/mock.rs
+++ b/nomos-services/storage/src/backends/mock.rs
@@ -50,6 +50,13 @@ impl<SerdeOp: StorageSerde + Send + Sync + 'static> StorageBackend for MockStora
         Ok(())
     }
 
+    async fn store_batch(&mut self, keys_values: Vec<(Bytes, Bytes)>) -> Result<(), Self::Error> {
+        for (key, value) in keys_values {
+            let _ = self.inner.insert(key, value);
+        }
+        Ok(())
+    }
+
     async fn load(&mut self, key: &[u8]) -> Result<Option<Bytes>, Self::Error> {
         Ok(self.inner.get(key).cloned())
     }
--- a/nomos-services/storage/src/backends/mod.rs
+++ b/nomos-services/storage/src/backends/mod.rs
@@ -44,6 +44,8 @@ pub trait StorageBackend: Sized {
     type SerdeOperator: StorageSerde + Send + Sync + 'static;
     fn new(config: Self::Settings) -> Result<Self, Self::Error>;
     async fn store(&mut self, key: Bytes, value: Bytes) -> Result<(), Self::Error>;
+    /// Store batch of key/value pairs in one transaction
+    async fn store_batch(&mut self, keys_values: Vec<(Bytes, Bytes)>) -> Result<(), Self::Error>;
     async fn load(&mut self, key: &[u8]) -> Result<Option<Bytes>, Self::Error>;
     /// Loads all values whose keys start with the given prefix.
     async fn load_prefix(&mut self, prefix: &[u8]) -> Result<Vec<Bytes>, Self::Error>;
--- a/nomos-services/storage/src/backends/rocksdb.rs
+++ b/nomos-services/storage/src/backends/rocksdb.rs
@@ -113,6 +113,15 @@ impl<SerdeOp: StorageSerde + Send + Sync + 'static> StorageBackend for RocksBack
         self.rocks.put(key, value)
     }
 
+    async fn store_batch(&mut self, keys_values: Vec<(Bytes, Bytes)>) -> Result<(), Self::Error> {
+        let mut batch = rocksdb::WriteBatch::default();
+        for (key, value) in keys_values {
+            batch.put(key, value);
+        }
+        self.rocks.write(batch)?;
+        Ok(())
+    }
+
     async fn load(&mut self, key: &[u8]) -> Result<Option<Bytes>, Self::Error> {
         self.rocks.get(key).map(|opt| opt.map(|ivec| ivec.into()))
     }
@@ -176,6 +185,36 @@ mod test {
         assert_eq!(load_value, Some(value.as_bytes().into()));
         let removed_value = db.remove(key.as_bytes()).await?;
         assert_eq!(removed_value, Some(value.as_bytes().into()));
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_store_batch(
+    ) -> Result<(), <RocksBackend<NoStorageSerde> as StorageBackend>::Error> {
+        let temp_path = TempDir::new().unwrap();
+        let rocks_settings = RocksBackendSettings {
+            db_path: temp_path.path().to_path_buf(),
+            read_only: false,
+            column_family: None,
+        };
+        let key1 = "foo1";
+        let value1 = "bar1";
+
+        let key2 = "foo2";
+        let value2 = "bar2";
+
+        let mut db: RocksBackend<NoStorageSerde> = RocksBackend::new(rocks_settings)?;
+        db.store_batch(vec![
+            (key1.as_bytes().into(), value1.as_bytes().into()),
+            (key2.as_bytes().into(), value2.as_bytes().into()),
+        ])
+        .await?;
+
+        let load_value1 = db.load(key1.as_bytes()).await?;
+        assert_eq!(load_value1, Some(value1.as_bytes().into()));
+        let load_value2 = db.load(key2.as_bytes()).await?;
+        assert_eq!(load_value2, Some(value2.as_bytes().into()));
 
         Ok(())
     }
--- a/nomos-services/storage/src/backends/sled.rs
+++ b/nomos-services/storage/src/backends/sled.rs
@@ -70,6 +70,16 @@ impl<SerdeOp: StorageSerde + Send + Sync + 'static> StorageBackend for SledBacke
         Ok(())
     }
 
+    async fn store_batch(&mut self, keys_values: Vec<(Bytes, Bytes)>) -> Result<(), Self::Error> {
+        self.sled.transaction(|tx| {
+            for (key, value) in &keys_values {
+                tx.insert(key.to_vec(), value.to_vec())?;
+            }
+            Ok(())
+        })?;
+        Ok(())
+    }
+
     async fn load(&mut self, key: &[u8]) -> Result<Option<Bytes>, Self::Error> {
         Ok(self.sled.get(key)?.map(|ivec| ivec.to_vec().into()))
     }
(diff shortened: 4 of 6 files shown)


andrussal:
I feel like that I have some more "fundamental" things unclear about the `execute` functionality and how to use it properly.

I have been thinking how our transaction logic matches with the idea of database transaction. In the sense that it does something atomically from database perspective. I think that our current transaction "framework" actually doesn't really do it by default. For example, in RocksDb test, it makes multiple DB calls in a single closure but this doesn't use database implementation transaction functionality.

```rust
let mut db: RocksBackend<NoStorageSerde> = RocksBackend::new(sled_settings)?;
        let txn = db.txn(|db| {
            let key = "foo";
            let value = "bar";
            db.put(key, value)?;
            let result = db.get(key)?;
            db.delete(key)?;
            Ok(result.map(|ivec| ivec.to_vec().into()))
        });
        let result = db.execute(txn).await??;
```
Here `db.txn` calls `put` and `delete` but they are actually not part of a database transaction, just separate calls. 

Here we could potentially ask actual db transaction from `db`, something like `let tx = db.init_transaction()`, and then do `tx.put`, `tx.delete` and `tx.commit`. Alternatively, instead of `db` argument, closure could already provide actual db implementation transaction type, so it becomes `db.txn(|tx|`.  

Just thinking that our `Transaction` doesn't actually represent a transaction in traditional database sense, it doesn't "hook" with database transaction functionality. It might not add much benefit from caller side to use `execute` instead of doing multiple separate `Store` calls with current implementation.

---

Another thing that is a bit unclear to me is how to create transaction object from an OverWatch service. Execute message takes transaction as argument.

```rust
Execute {
        transaction: Backend::Transaction,
        reply_channel:
            tokio::sync::oneshot::Sender<<Backend::Transaction as StorageTransaction>::Result>,
    }
    
```
  
For example RocksDb transaction can be constructed like this:
  
```rust
  impl<SerdeOp> RocksBackend<SerdeOp> {
    pub fn txn(
        &self,
        executor: impl FnOnce(&DB) -> Result<Option<Bytes>, Error> + Send + Sync + 'static,
    ) -> Transaction {
        Transaction {
            rocks: self.rocks.clone(),
            executor: Box::new(executor),
        }
    }
}
```

This requires caller to have access to RocksBackend and OverWatch service doesn't have it. 

I was thinking maybe it makes sense to change transaction type in `Execute` message to something that doesn't reference db implementation. Maybe only contains `impl FnOnce(&DB) -> Result<Option<Bytes>, Error> + Send + Sync + 'static`.

```rust
Execute {
        transaction: OverWatch::Transaction,
        ...
```

---
Also it seems that we may need to switch to `async` closure.

OverWatch service needs to prepare this `Box<dyn FnOnce(&DB) -> Result<Option<Bytes>, Error> + Send + Sync>` for `Execute` message.

Here `&DB` argument references specific db implementation, like RocksDb `DB`. But OverWatch service probably should have access only to the generic `StorageBackend`. This means that `Box<dyn FnOnce(&DB) -> Result<Option<Bytes>, Error> + Send + Sync>` doesn't work. Instead it needs to be some kind of future because `StorageBackend` functions are `async`, whereas `DB` is sync.

--- 

So I guess my current feeling is that it's not straightforward to use existing `execute` implementation from OverWatch.

danielSanchezQ:
> I feel like that I have some more "fundamental" things unclear about the `execute` functionality and how to use it properly.
> 
> I have been thinking how our transaction logic matches with the idea of database transaction. In the sense that it does something atomically from database perspective. I think that our current transaction "framework" actually doesn't really do it by default. For example, in RocksDb test, it makes multiple DB calls in a single closure but this doesn't use database implementation transaction functionality.
> 
> ```rust
> let mut db: RocksBackend<NoStorageSerde> = RocksBackend::new(sled_settings)?;
>         let txn = db.txn(|db| {
>             let key = "foo";
>             let value = "bar";
>             db.put(key, value)?;
>             let result = db.get(key)?;
>             db.delete(key)?;
>             Ok(result.map(|ivec| ivec.to_vec().into()))
>         });
>         let result = db.execute(txn).await??;
> ```
> 
> Here `db.txn` calls `put` and `delete` but they are actually not part of a database transaction, just separate calls.
> 
> Here we could potentially ask actual db transaction from `db`, something like `let tx = db.init_transaction()`, and then do `tx.put`, `tx.delete` and `tx.commit`. Alternatively, instead of `db` argument, closure could already provide actual db implementation transaction type, so it becomes `db.txn(|tx|`.
> 
> Another thing that is a bit unclear to me is how to create transaction object from an OverWatch service. Execute message takes transaction as argument.
> 
> ```rust
> Execute {
>         transaction: Backend::Transaction,
>         reply_channel:
>             tokio::sync::oneshot::Sender<<Backend::Transaction as StorageTransaction>::Result>,
>     }
>     
> ```
> 
> For example RocksDb transaction can be constructed like this:
> 
> ```rust
>   impl<SerdeOp> RocksBackend<SerdeOp> {
>     pub fn txn(
>         &self,
>         executor: impl FnOnce(&DB) -> Result<Option<Bytes>, Error> + Send + Sync + 'static,
>     ) -> Transaction {
>         Transaction {
>             rocks: self.rocks.clone(),
>             executor: Box::new(executor),
>         }
>     }
> }
> ```
> 
> This requires caller to have access to RocksBackend and OverWatch service doesn't have it.
> 
> I was thinking maybe it makes sense to change transaction type in `Execute` message to something that doesn't reference db implementation. Maybe only contains `impl FnOnce(&DB) -> Result<Option<Bytes>, Error> + Send + Sync + 'static`.
> 
> ```rust
> Execute {
>         transaction: OverWatch::Transaction,
>         ...
> ```
> 
> Also it seems that we may need to switch to `async` closure.
> 
> OverWatch service needs to prepare this `Box<dyn FnOnce(&DB) -> Result<Option<Bytes>, Error> + Send + Sync>` for `Execute` message.
> 
> Here `&DB` argument references specific db implementation, like RocksDb `DB`. But OverWatch service probably should have access only to the generic `StorageBackend`. This means that `Box<dyn FnOnce(&DB) -> Result<Option<Bytes>, Error> + Send + Sync>` doesn't work. Instead it needs to be some kind of future because `StorageBackend` functions are `async`, whereas `DB` is sync.
> 
> So I guess my current feeling is that it's not straightforward to use existing `execute` implementation from OverWatch.

Yes. The `transaction` naming is missleading. It should actually just be called `execute` or something similar. In reality only the kv operations are abstracted. For anything else if implementation specific and that means you need to hardcode the type before hand. 
The idea with `execute` is that you get access to a hook to the specific database and then you do whatever you need. 
Changing the call to be async is actually a good idea. We could extend the api to have an `execute_async` in addition to what we have maybe?

andrussal:
> Yes. The `transaction` naming is missleading. It should actually just be called `execute` or something similar. In reality only the kv operations are abstracted. For anything else if implementation specific and that means you need to hardcode the type before hand. The idea with `execute` is that you get access to a hook to the specific database and then you do whatever you need. Changing the call to be async is actually a good idea. We could extend the api to have an `execute_async` in addition to what we have maybe?

I am thinking that ideally we have an abstract type for the `execute` callback argument as well? Otherwise we end up with implementation specific code in services and elsewhere. Then again `execute` loses it's power to support implementation specific functionality. Just wondering how much we are willing to break abstraction to use db specific stuff anyway.

Maybe a good compromise for most use cases is to have something like `in_transaction` function that takes abstract `Transaction` as argument that has the same interface as `StorageBackend`(except `execute`)? Transaction would be a more generic abstraction than storing a batch as initially suggested.

(Actually I suggested to have `async execute` because `StorageBackend` is `async`. Maybe not needed if we go with db specific type in `execute` callback though...)

danielSanchezQ:
I dont think we can abstract an execute generically because it **hard** depends on the actual underlying implementation. The only thing we can do here if abstract in the `StorageTransaction` trait (lets rename this) an executable that does what we need (builds it up). Otherwise it is up to the user of the service. That way we keep both options. But if you have another suggestion on how to do it generically I'm curious about it.

andrussal:
> I dont think we can abstract an execute generically because it **hard** depends on the actual underlying implementation. The only thing we can do here if abstract in the `StorageTransaction` trait (lets rename this) an executable that does what we need (builds it up). Otherwise it is up to the user of the service. That way we keep both options. But if you have another suggestion on how to do it generically I'm curious about it.

I agree it's good to keep KV interface the same. Perhaps another option is to implement an additional trait for db backend implementation and bring it into scope when needed.

```rust
trait StorageExt {
   fn write_in_batch(...)
}

impl StorageExt for rocksdb:DB {
     fn write_in_batch(...)
}
```

```rust
use StorageExt;

fn store<DB: StorageExt>() {
      let exec = |db: DB| {
           db.write_in_batch(...);
      }
}
```

danielSanchezQ:
> > I dont think we can abstract an execute generically because it **hard** depends on the actual underlying implementation. The only thing we can do here if abstract in the `StorageTransaction` trait (lets rename this) an executable that does what we need (builds it up). Otherwise it is up to the user of the service. That way we keep both options. But if you have another suggestion on how to do it generically I'm curious about it.
> 
> I agree it's good to keep KV interface the same. Perhaps another option is to implement an additional trait for db backend implementation and bring it into scope when needed.
> 
> ```rust
> trait StorageExt {
>    fn write_in_batch(...)
> }
> 
> impl StorageExt for rocksdb:DB {
>      fn write_in_batch(...)
> }
> ```
> 
> ```rust
> use StorageExt;
> 
> fn store<DB: StorageExt>() {
>       let exec = |db: DB| {
>            db.write_in_batch(...);
>       }
> }
> ```

Yeah, this similar extending the current `StorageTransaction` trait, but as a separate trait (maybe as an extension trait?). Anyway, you will still need the specific type to use it. So I think extending the trait for building the "execution operations" is a better approach.

danielSanchezQ:
@andrussal as per our last conversation. I think we can simply close this one right?

andrussal:
> @andrussal as per our last conversation. I think we can simply close this one right?

PR #1079, opened 2025-02-25
fix: docker triggering job

ntn-x2:

Use the pull request ref (`head_ref`) or, if absent, the branch name (`ref`) to trigger the Docker build workflow. It also re-adds the `github_token` input for the risczero action as without it the CI is throttled.

Since it seems GitHub is not strictly enforcing rules on stored caches either, I also opted for splitting the test caches into one for each OS, so that they are both faster to run.

### DIFF
--- a/.github/actions/install-risc0/action.yml
+++ b/.github/actions/install-risc0/action.yml
@@ -2,11 +2,20 @@ name: "Install Risc0"
 
 description: "Install the Risc0 toolchain. Requires the Rust toolchain to be installed."
 
+inputs:
+  github_token:
+    description: "The GitHub token to use for rate limiting."
+    required: true
+
 runs:
   using: "composite"
   steps:
     - name: Install cargo-binstall
       uses: cargo-bins/cargo-binstall@v1.11.1
+      env:
+        GITHUB_TOKEN: ${{ inputs.github_token }}
     - name: Install risc0 using cargo-binstall
       run: cargo binstall -y cargo-risczero@1.2.4 && cargo risczero install
       shell: bash
+      env:
+        GITHUB_TOKEN: ${{ inputs.github_token }}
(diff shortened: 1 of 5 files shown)


.github/actions/install-risc0/action.yml
```diff
@@ -2,11 +2,20 @@ name: "Install Risc0"
...
+
 runs:
   using: "composite"
   steps:
     - name: Install cargo-binstall
       uses: cargo-bins/cargo-binstall@v1.11.1
+      env:
+        GITHUB_TOKEN: ${{ inputs.github_token }}
     - name: Install risc0 using cargo-binstall
       run: cargo binstall -y cargo-risczero@1.2.4 && cargo risczero install
```
bacv, line 18: Seems that node with risc0`1.2.4` fails more often, lets try `1.2.0`?

ntn-x2: Had to do a bit of fire extinguishing due to some deps now requiring rust version 2024, meaning we had to bump the version in the CI, which I though was done automatically via the toolchain file, but it seems not to be the case. Bumping to 1.85 also introduced some more clippy warnings, which I had to address.

ntn-x2: Nevertheless, the tests seem to be failing also with this version installed.

ntn-x2: Should I try a different version? Do you have any hints?

bacv: I'm working on removing `indexer_tests`, the most important for now is that `two_nodes_happy` tests works reliably (it tests consensus). I see that the situation changed, as we have macos green. It seems that now it fails less often - let's try updating to `1.2.4` once I remove `indexer_test`?

bacv (approved):
I suggest to leave 1.2.0 for now (https://github.com/logos-co/nomos-node/pull/1079#discussion_r1971314415)

PR #1084, opened 2025-02-26
chore(nomos-node): update dockerfile to include global params and app…

pradovic:
…ropriate caargo-risczero version

This is an updated version of a `Dockerfile` in `nomos-node`. The structure is based on `testnet/Dockerfile` with the main difference of being a `rust image` as final image. The purpose was to pin down the version of `cargo-risczero@1.2.0` so avoid `please use rzup install` error, and to support mounting `kzgrs` file. 

N/A

@bacv 

N/A

N/A
## Checklist

> [!WARNING]  

* [x] 1. Description added.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,17 +17,13 @@ RUN apt-get update && apt-get install -yq \
 
 RUN cargo install cargo-binstall
 
-# Versions of risczero > 1.2.0 use rzup install instead
+# Versions of cargo-risczero > 1.2.0 use rzup install instead
 RUN cargo binstall cargo-risczero@1.2.0 --no-confirm
 RUN cargo risczero install
 
 RUN cargo build --release -p nomos-node
 
-FROM rust:1.84.0-slim-bookworm AS final
-
-WORKDIR /app
-
-COPY --from=builder /nomos/target/release/nomos-node /usr/bin/nomos-node
+RUN cp /nomos/target/release/nomos-node /usr/bin/nomos-node
 
 # Expose default ports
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,32 +1,35 @@
...
-RUN cargo binstall -y cargo-risczero
+
+# Versions of risczero > 1.2.0 use rzup install instead
+RUN cargo binstall cargo-risczero@1.2.0 --no-confirm
 RUN cargo risczero install
 
 RUN cargo build --release -p nomos-node
 
-# NODE IMAGE ----------------------------------------------------------
+FROM rust:1.84.0-slim-bookworm AS final
```
bacv, line 26: If we use second stage as final image, we also need to install risc0 again. For this reason, for now lets use only one stage (builder) as final image.

bacv (changes requested):
Looks good, but lets use one stage for now. Also maybe all docker images and `shell.nix` in repo could be updated to use 1.85 rust?
There are `ci/Dockerfile` and `testnet/Dockerfile`.

Dockerfile
pradovic: Should be fixed with: https://github.com/logos-co/nomos-node/pull/1084/commits/f5ae2647ff5c9790e5e8cefd765d66e73d4589c2. I am not 100% sure how to test if everything from the node works, but I checked directly on the container as:
```
docker run --rm -it --entrypoint /bin/sh nomos
# cargo risczero --version
cargo-risczero 1.2.0
```

bacv (approved):
🚀

PR #1090, opened 2025-02-27
feat(utils): Add bounded duration

danielSanchezQ:

Implements a type that can be used as a proxy type for deserializing `Duration` with a minimum bound set at compile time. This will be used later in the configuration so non 0 values can be derived.

Yes

@danielSanchezQ 

N/A

N/A

## Checklist

* [ ] 1. Description added.


danielSanchezQ:
@ntn-x2 What do you think of using this to bound settings in cryptarchia so no zero seconds can be used? 
During integration we should use this type as a proxy for deserialization only.

ntn-x2:
I mean if the goal is to only avoid zero seconds, having a check in the constructor and using a `NonZero` type for the relevant field would be sufficient. What you present here is something that goes a little bit beyond, and I cannot say whether it is fully justified.

danielSanchezQ:
> I mean, if the goal is to only avoid zero seconds, having a check in the constructor and using a `NonZero` type for the relevant field would be sufficient. What you present here is something that goes a little bit beyond, and I cannot say whether it is fully justified.

The problem is that Duration cannot be `NonZero '. We have some assumptions on the values that may come from **configuration** and use them with those assumptions. Meaning that things may go wrong with the wrong configuration, but makes the code breezeer just panicking when some config is wrong. I would rather catch up on what is wrong on deserializaton than streghthen the code and make it more complicated just because of configuration being wrong. I'm talking about the operations with time on slots and epochs.

ntn-x2:
Oh yeah I totally agree there, any invalid values should be caught as soon as the config object is created, either via a regular constructor or via deserialization. What about a wrapper type `NonZeroDuration` which guarantees the value is not zero?

danielSanchezQ:
> Oh yeah I totally agree there, any invalid values should be caught as soon as the config object is created, either via a regular constructor or via deserialization. What about a wrapper type `NonZeroDuration` which guarantees the value is not zero?

Yeah, I thought of that, but what if you don't want to don't zero seconds? You still can get non-zero `10ns`. That is why this type ensures that is configured in compile time (checked at runtime deserialization though).

ntn-x2:
Ok so the goal is not to have a zero by which we divide, but forbidding specific fields from being zero? Is that a reasonable ask?

danielSanchezQ:
> Ok so the goal is not to NOT have a zero by which we divide, but forbidding specific fields from being zero? Is that a reasonable ask?

I think it is legit yes. We do not probably want/need/can't have less than one seconds slots for example, but we may want to define them in milliseconds. Does it makes sense? :thinking:

ntn-x2:
Yes, now it does. Still, this type can only prevent a single field from not being less than a given value it seems, right? In any case, it seems to make sense now indeed 😊

nomos-utils/src/bounded_duration.rs
```diff
@@ -0,0 +1,181 @@
...
+#[cfg(test)]
+mod test {
+    use serde_with::serde_as;
+    use time::Duration;
+
+    use crate::bounded_duration::{MinimalBoundedDuration, DAY, SECOND};
+
+    #[test]
+    fn success_deserialize() {
+        let value = serde_json::to_value(Duration::days(1)).unwrap();
```
ntn-x2, line 146: I think it would be better to use a literal JSON string to give an idea of how it would look like when serialized.

ntn-x2 (approved):
Nice ⏰

danielSanchezQ:
Errors are non-related. I'll tackle them in a different PR.

bacv (approved):

PR #1091, opened 2025-02-28
fix(risc0): Align risc0-build and risc0-zkvm with the host server version

bacv:

Nomos ci, docker files and github actions pin the risc0 server to a specific verions. In the process of updating the risc0, our build dependencies needs to be updated as well. 
Caught in `nomos-e2e-tests` run, lost the original error message but it was about incompatible risc0 client and host server.

Since `nomos-core/risc0_proofs` is excluded from the workspace, the versions in that crate will need to be updated manually.

N/A

@bacv @zeegomo 

N/A

N/A

## Checklist

* [x] 1. Description added.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,6 +92,7 @@ tests                    = { path = "./tests" }
 # External
 overwatch-derive = { git = "https://github.com/logos-co/Overwatch", rev = "86c7405" }
 overwatch-rs     = { git = "https://github.com/logos-co/Overwatch", rev = "86c7405" }
+risc0-zkvm       = "1.2.0"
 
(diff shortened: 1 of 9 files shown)


zeegomo (approved):
This change is ok, but the server is installed through rzup, so we should check that as well

bacv:
> This change is ok, but the server is installed through rzup, so we should check that as well

In github workflows we have risc0 pinned ([risc0 action](https://github.com/logos-co/nomos-node/blob/master/.github/actions/install-risc0/action.yml)), same in Dockerimages, but yes, when updating all these places will need to be doublechecked.

PR #1093, opened 2025-02-28
chore: enable clippy::pedantic lints

ntn-x2:

Introduce pedantic lints for everyone to use. I will address the ignored ones about arithmetic operations (`clippy::cast_possible_truncation, clippy::cast_sign_loss, clippy::cast_precision_loss, clippy::cast_possible_wrap`) in a different PR as the changes are not trivial, and this PR gets constant conflicts whenever someone merges into `master`.

Yes.

@ntn-x2 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 1. Description added.


.cargo/config.toml
```diff
@@ -10,4 +10,16 @@ rustflags = [
 rustflags = [
   "-Wclippy::nursery",
   "-Aclippy::large_stack_frames",
+
+  "-Wclippy::pedantic",
+  "-Aclippy::similar_names",
+  # TODO: Tackle before opening PR
+  "-Aclippy::cast_possible_truncation",
```
ntn-x2, line 17: I will tackle this section in a subsequent PR.

.cargo/config.toml
```diff
@@ -10,4 +10,16 @@ rustflags = [
...
+  "-Aclippy::cast_sign_loss",
```
ntn-x2, line 18: I will not tackle these lints here, but we should tackle them before publishing the crates.

.cargo/config.toml
```diff
@@ -10,4 +10,16 @@ rustflags = [
...
+  "-Aclippy::cast_precision_loss",
+  "-Aclippy::cast_possible_wrap",
+  # TODO: Remove at some point
```
ntn-x2, line 21: I will not tackle these lints here, but we should tackle them before publishing the crates.

nomos-blend/network/src/handler.rs
```diff
@@ -244,19 +245,16 @@ where
...
+                    if let Some(msg) = self.outbound_msgs.pop_front() {
+                        tracing::debug!("Sending message to outbound stream: {:?}", msg);
+                        self.outbound_substream = Some(OutboundSubstreamState::PendingSend(
+                            send_msg(stream, msg).boxed(),
+                        ));
+                    } else {
+                        tracing::debug!("Nothing to send to outbound stream");
+                        self.outbound_substream = Some(OutboundSubstreamState::Idle(stream));
+                        self.waker = Some(cx.waker().clone());
+                        return Poll::Pending;
```
danielSanchezQ, line 257: Is there a lint for this? Does it really make sense? :thinking:

nomos-core/chain-defs/src/da/blob/select.rs
```diff
@@ -31,6 +30,14 @@ impl<const SIZE: usize, B: DispersedBlobInfo> BlobSelect for FillSize<SIZE, B> {
         &self,
         certificates: I,
     ) -> impl Iterator<Item = Self::BlobId> + 'i {
-        utils::select::select_from_till_fill_size::<SIZE, Self::BlobId>(|c| c.size(), certificates)
+        #[expect(clippy::redundant_closure_for_method_calls)]
+        // TODO: Replace this redundant closure with `B::size` without triggering compiler errors
+        // about B not living long enough.
```
danielSanchezQ, line 35: What is the issue here. Calling `size`?

nomos-core/chain-defs/src/proofs/leader_proof.rs
```diff
@@ -118,8 +118,10 @@ mod test {
     fn note_commitment_leaves(
         note_commitments: &[cl::NoteCommitment],
     ) -> [[u8; 32]; MAX_NOTE_COMMS] {
-        let note_comm_bytes =
-            Vec::from_iter(note_commitments.iter().map(|c| c.as_bytes().to_vec()));
+        let note_comm_bytes = note_commitments
+            .iter()
+            .map(|c| c.as_bytes().to_vec())
+            .collect::<Vec<_>>();
```
danielSanchezQ, line 124: I would rather use the type in the variable declaration rather than collect though.
`let note_comm_bytes: Vec<_> = ...`
Is this linted as well?

nomos-core/chain-defs/src/tx/select.rs
```diff
@@ -11,18 +11,17 @@ pub struct FillSize<const SIZE: usize, Tx> {
...
+        Self { _tx: PhantomData }
     }
 }
 
 impl<const SIZE: usize, Tx: Transaction> TxSelect for FillSize<SIZE, Tx> {
     type Tx = Tx;
     type Settings = ();
 
-    fn new(_: Self::Settings) -> Self {
+    fn new((): Self::Settings) -> Self {
```
danielSanchezQ, line 24: Is this so in case `Self::Settings` change type the compiler will throw an error about it for this non-usage?

nomos-da/kzgrs-backend/src/common/mod.rs
```diff
@@ -96,21 +104,24 @@ impl AsRef<[Chunk]> for Column {
 }
 
 impl ChunksMatrix {
+    #[must_use]
     pub fn len(&self) -> usize {
         self.0.len()
     }
+    #[must_use]
```
danielSanchezQ, line 111: I see the point, but this feels that is more a spam than other thing :sweat_smile:

nomos-da/kzgrs-backend/src/common/mod.rs
```diff
@@ -96,21 +104,24 @@ impl AsRef<[Chunk]> for Column {
...
     }
     pub fn rows(&self) -> impl Iterator<Item = &Row> + '_ {
         self.0.iter()
     }
     #[cfg(feature = "parallel")]
+    #[must_use]
     pub fn par_rows(&self) -> impl ParallelIterator<Item = &Row> + '_ {
         self.0.par_iter()
     }
     pub fn columns(&self) -> impl Iterator<Item = Column> + '_ {
```
danielSanchezQ, line 123: Why not `must_use` in this method?

nomos-da/kzgrs/benches/rs.rs
```diff
@@ -22,10 +25,10 @@ fn rs_encode(bencher: Bencher, size: usize) {
             let (_, poly) = bytes_to_polynomial::<31>(buff, domain).unwrap();
             let domain = GeneralEvaluationDomain::<Fr>::new(size * 2).unwrap();
             black_box(move || encode(&poly, domain))
-        })
+        });
 }
 
-#[divan::bench(args = [16399, 32798, 65565, 131099, 262167, 524241, 1048606], sample_size = 10, sample_count = 100)]
+#[divan::bench(args = [16399, 32798, 65565, 131_099, 262_167, 524_241, 1_048_606], sample_size = 10, sample_count = 100)]
```
danielSanchezQ, line 31: Let's fill the other as well maybe? `16_399`, `32_798`..?

nomos-da/network/core/Cargo.toml
```diff
@@ -27,6 +27,7 @@ tracing-subscriber       = "0.3.18"
 
 [dev-dependencies]
 blake2             = "0.10"
+kzgrs              = { workspace = true }
```
danielSanchezQ, line 30: Who did it work before? :thinking:

nomos-da/network/core/src/protocols/dispersal/executor/behaviour.rs
```diff
@@ -61,31 +61,33 @@ pub enum DispersalError {
...
 impl DispersalError {
+    #[must_use]
     pub const fn blob_id(&self) -> Option<BlobId> {
         match self {
-            Self::Io { blob_id, .. } => Some(*blob_id),
-            Self::Serialization { blob_id, .. } => Some(*blob_id),
-            Self::Protocol {
+            Self::Io { blob_id, .. }
+            | Self::Serialization { blob_id, .. }
+            | Self::Protocol {
```
danielSanchezQ, line 69: what lint is this? How do they decide when is more readable one than the other version? :thinking:

testnet/cfgsync/src/repo.rs
```diff
@@ -76,36 +77,33 @@ impl ConfigRepo {
...
-                    self.da_params.clone(),
-                    self.tracing_settings.clone(),
-                    hosts,
-                );
-
-                for (host, sender) in waiting_hosts.drain() {
-                    let config = configs.get(&host).expect("host should have a config");
-                    let _ = sender.send(RepoResponse::Config(Box::new(config.to_owned())));
-                }
+        if timeout(timeout_duration, self.wait_for_hosts()).await == Ok(()) {
```
danielSanchezQ, line 80:
```suggestion
        if timeout(timeout_duration, self.wait_for_hosts()).await.is_ok() {
```

tests/src/nodes/executor.rs
```diff
@@ -91,15 +94,20 @@ impl Executor {
...
+                .blob_storage_directory,
+        );
+        dir.path().clone_into(
+            &mut config
+                .da_verifier
+                .storage_adapter_settings
+                .blob_storage_directory,
+        );
+        dir.path()
+            .clone_into(&mut config.da_indexer.storage.blob_storage_directory);
```
danielSanchezQ, line 110: Why? Is there really a lint for this?
This is much more confusing than a single assignation...

danielSanchezQ (commented):
There is a lot of changes. Some really good, some I understand them but they are bothersome. Some I really do not understand them at all. We can probably go ahead. I would like for everyone to check on this. @zeegomo @youngjoon-lee @bacv @AlejandroCabeza @andrussal @pradovic  @CHr15F0x 

Left a few comments but most of them are just out of curiosity.

nomos-da/network/core/Cargo.toml
bacv: We used `Default::default()` for commitments, proofs - https://github.com/logos-co/nomos-node/pull/1093/files#diff-c0e266e9e2ccfbc5a190cf49c9e696a1e023484d68add66e74beacd57bf6cb1bR101-R103

nomos-blend/network/src/handler.rs
ntn-x2: This is the `clippy::single_match_else` lint. Quoting:

> Just readability – `if let` nests less than a match.

I can also revert this if you guys don't like it, but I don't think it's any worse than the previous way.

nomos-core/chain-defs/src/da/blob/select.rs
ntn-x2: Yes, replacing the closure with `B::size` gives lifetime issues. You now how to fix this?

nomos-core/chain-defs/src/proofs/leader_proof.rs
ntn-x2: Yes, I believe that won't trigger warnings. The key here was to replace `::from_iter` with `collect`. I will change that.

nomos-core/chain-defs/src/tx/select.rs
ntn-x2: Yes. The lint is `clippy::ignored_unit_patterns`.

ntn-x2: And also because we would notice the change in type itself, not just because of the missing underscore.

nomos-core/chain-defs/src/da/blob/select.rs
danielSanchezQ: I can take a look. It kinda triggers something from when I implemented this piece. But I do not recall correctly. It doesn't really matter much.

nomos-da/kzgrs/benches/rs.rs
ntn-x2: Clippy did not complain about those, but sure I can do that.

nomos-da/network/core/Cargo.toml
danielSanchezQ: Ah yeah, makes sense!

ntn-x2: Exactly. I would argue using concrete types is easier to understand especially for people that do not deal with this specific part of the code that often. But again, if most people prefer the old way, I can revert the changes.

danielSanchezQ: No, using the types is more readable. We can keep this yes.

nomos-da/network/core/src/protocols/dispersal/executor/behaviour.rs
ntn-x2: This is a `clippy::match_same_arms`. Again, I think it highlights better the fact that all those variants should do the same operation, and it's easier to see. Otherwise you'd have to compare each variant to understand whether they all do the same thing.

nomos-da/kzgrs-backend/src/common/mod.rs
danielSanchezQ: I'm not against it. But it kinda tickles my brain when we have to do something repetitive. This should probably be some kinda flag instead of a directive to use manually. But lets see what the guys opinion is.

tests/src/nodes/executor.rs
ntn-x2: This is a `clippy::assigning_clones`, and it has do to with avoiding unnecessary allocations.

nomos-da/network/core/src/protocols/dispersal/executor/behaviour.rs
danielSanchezQ: Yeah, it makes sense why. But you could argue against it as well. But if it is the consensus, then ahead.

tests/src/nodes/executor.rs
danielSanchezQ: I understand it is more optimal...but sincerly if clippy catches this the compiler should optimize it away, I wounder if it would :thinking: 
The point is that indeed, this is pedantic :joy:

danielSanchezQ: The assignation seems more optimized at first sight: https://rust.godbolt.org/z/War11carx

bacv:
To me all lints makes sense, the only being bit annoying `#[must_use]` but it will automatically handled by editor, so I'll get used to it. Looks good!

nomos-blend/network/src/handler.rs
ntn-x2: Clippy is not very happy with wildcard patterns, nor am I to be honest. It is bothersome to consider all possible cases, but I find this to be a strength of Rust as it makes you aware (willingly or not) that stuff is changing, and you are forced to react to them. In this case, we are indeed matching over an option, so there won't be any more branches in the future, making your suggestion of using a `_` equally valid, but I find the suggestion from Clippy on point and a better approach that indeed leads to a layer less of indentation, which never hurts.

nomos-core/chain-defs/src/da/blob/select.rs
ntn-x2: Sure, this is anyway a very minor details. We do ignore lints in few other places (especially the `too_many_lines` and the `cognitive_complexity` ones), so this would not be the only one we ignore and definitely not the "worst" one to ignore. But I'm curious to see how the solution would look like, as I tried adding a lifetime bound on `B` but it did not fix the compilation issue.

nomos-core/chain-defs/src/proofs/leader_proof.rs
ntn-x2: Let me mention we have a mixed approach throughout our codebase. There is plenty of places in which we use turbofish syntax, so this would not be the only instance. But I changed the approach here: https://github.com/logos-co/nomos-node/pull/1093/commits/20ebe764a10ae25cf4c5777ac7940e6d0db212b1.

nomos-da/kzgrs-backend/src/common/mod.rs
ntn-x2: Co-pilot answer:

> The columns method in the ChunksMatrix implementation returns an iterator over Column objects. Clippy might not have suggested `#[must_use]` for this method because it is common for iterators to be used immediately in a method chain or passed directly to other functions. The #[must_use] attribute is more typically recommended for methods where the return value is crucial to be used, like computations or transformations.

So it seems to me Clippy runs some heurstics on the source code to understand what are the places where the result is easier to be "misused", and suggest to add a `#[must_use]` there.

testnet/cfgsync/src/repo.rs
ntn-x2: Again I think in this case is better to match on the expected result inside the `Ok()`. Pedantic Clippy wont' complain, but the restriction lints I am working on actually will do. If the `timeout` function changes for a reason, I think it would be the least to understand "why", and take action based on that. Using `is_ok()` basically assumes the result is `()` or something useless, but this assumption might not hold true forever in the future. What do you think?

tests/src/nodes/executor.rs
ntn-x2: Taken from the lint description:

> Custom `Clone::clone_from()` or `ToOwned::clone_into` implementations allow the objects to share resources and therefore avoid allocations.

I believe it depends on how the `Clone` trait is implemented, and I understand that some types are given the chance to implement `clone_from` and `clone_into` in a more performant way. So I think the lint as such makes sense, in this case, if you think we should revert to the old way, we can add an ignore, although I don't see a strong reason to do so.

danielSanchezQ: I do not see a strong reason to revert other than the one I gave. But I guess they know better. Also, I'm surprised it is the only place it occurs, though.

nomos-da/kzgrs/benches/rs.rs
ntn-x2: Did that in https://github.com/logos-co/nomos-node/pull/1093/commits/a95373539ea37b68c60b52d0b366c17f8f5a453e. Although because there's no Clippy lint, 4 and 5 digit numbers won't be enforced to use this rule in the future.

pradovic (approved):

nomos-blend/network/src/handler.rs
zeegomo: My point was that I would be ok with replacing `_` patterns but a bit less keen on explicit ones. Anyway, not a blocker, just not sure it's that convenient

zeegomo (approved):

youngjoon-lee (approved):
Thank you for taking the time to review the entire codebase. These are incredibly useful lints.

nomos-core/chain-defs/src/da/blob/select.rs
ntn-x2: FYI I will go ahead and merge this PR. Any changes to this bit here would be introduced in a new, small PR 😊

PR #1096, opened 2025-03-03
feat(DA): get_blob http api

andrussal:

Add HTTP API to get blobs for light clients.

---

Also updated `/da/get-shared-commitments` endpoint request type to match the spec.

Yes

@andrussal 

N/A

N/A

## Checklist
* [x] 1. Description added.


nomos-services/api/src/http/storage.rs
```diff
@@ -1,17 +1,19 @@
...
     backends::{rocksdb::RocksBackend, StorageSerde},
     StorageMsg, StorageService,
 };
-use serde::de::DeserializeOwned;
+use serde::{de::DeserializeOwned, Serialize};
 
 pub async fn block_req<S, Tx>(
     handle: &overwatch_rs::overwatch::handle::OverwatchHandle,
     id: HeaderId,
 ) -> Result<Option<Block<Tx, full_replication::BlobInfo>>, super::DynError>
```
bacv, line 14: This should be `kzgrs-backend::dispersal::BlobInfo`, the types are identical, but slipped when moving from fullreplication to kzgrs. Could you double check if we use full replication anywhere else in the http api?

tests/src/tests/api/storage.rs
```diff
@@ -24,13 +42,22 @@ pub struct TestApiStorageCallsNode {
 }
 
 #[test]
-fn test_get_shared_commitments() {
+fn test_get_blob_data() {
```
bacv, line 45: Would it be possible to move these tests to `da` module (maybe split `da/{api.rs,dispersal.rs}`, and instead of `TestApiStorageCallsNode` use real node?

tests/src/tests/api/storage.rs
```diff
@@ -97,3 +126,28 @@ fn rand_data(elements_count: usize) -> Vec<u8> {
...
+    relay
+        .send(StorageMsg::Store {
+            key: Bytes::from(key.as_ref().to_vec()),
+            value: Wire::serialize(value),
+        })
+        .await
+        .unwrap();
+}
+
+async fn fetch_from_api<T: DeserializeOwned, R: Serialize + ?Sized>(url: &str, req: &R) -> T {
```
bacv, line 144: These could be moved to `clients/executor-http-client`, or `clients/common-http-client`, and use in integration tests. Or this functionality could be added to `tests/src/node/{validator,executor}` just like we have similar calls for the consensus api.

nomos-services/api/src/http/storage.rs
```diff
@@ -55,3 +57,36 @@ where
...
+    let (reply_tx, reply_rcv) = tokio::sync::oneshot::channel();
+    relay
+        .send(StorageMsg::Load {
+            key: blob_key,
+            reply_channel: reply_tx,
+        })
+        .await
+        .map_err(|(e, _)| e)?;
+
+    let result = reply_rcv.await.unwrap();
```
bacv, line 89: Can we map these unwraps to http errors?

bacv (commented):
I my opinion, the since we have this added to the node http api, and it affects the DA functionality, the tests should be performed on a nomos node.

tests/src/tests/api/storage.rs
andrussal: I went with `clients/common-http-client`.

bacv (approved):
🚀

PR #1102, opened 2025-03-04
refactor(da/network): drop custom connection handler and use libp2p-stream instead

CHr15F0x:

The aim of this PR is to refactor [`ReplicationBehaviour`](https://github.com/logos-co/nomos-node/blob/7281b38322017645129a6288a72b944f73f61690/nomos-da/network/core/src/protocols/replication/behaviour.rs#L90) to use [`libp2p-stream`](https://crates.io/crates/libp2p-stream) instead of using a [custom connection handler](https://github.com/logos-co/nomos-node/blob/7281b38322017645129a6288a72b944f73f61690/nomos-da/network/core/src/protocols/replication/handler.rs#L61).

Additionally local test execution time on my machine is reduced from ~12 seconds to ~0.5 seconds by using a [pre-computed test fixture](https://github.com/logos-co/nomos-node/pull/1102/files#diff-69e530f4d0f643b365530e5735137d697f575e6e5dd57763b893b4ebcfa3acf3), so I expect a similar improvement in CI.

Yes, the entirety of prod code changes is kept in `behaviour_v2.rs`, so that it can be easily compared against the current implementation. The plan is to swap out the current implementation once all work on the new `ReplicationBehaviour` is done. There are 3 tasks that will land on this branch in consecutive PRs:
1. the current PR (refactor into using `libp2p-stream`, maintain the same functionality)
2. optimize stream usage: reuse libp2p bidir Stream-s for both incoming and outgoing messages
3. optimize emitting `ReplicationEvent::IncomingMessage`: emit once per unique message, put message in seen-cache when initiating sending a message, avoid unnecessary replication

Spec: @danielSanchezQ 
PR: @CHr15F0x

Yes, except for the fact that the at most 2 bidir libp2p `Stream`-s are used right now in a connection between 2 peers, which will be addressed in a follow up PR (_optimize stream usage_).

No, the implementation is functionally-wise equivalent to the current implementation. 

## Checklist

> [!WARNING]  

* [x] 1. Description added.

### DIFF
--- a/nomos-da/network/core/src/protocols/replication/behaviour_v2.rs
+++ b/nomos-da/network/core/src/protocols/replication/behaviour_v2.rs
@@ -36,13 +36,17 @@ pub enum ReplicationError {
         peer_id: PeerId,
         error: OpenStreamError,
     },
+    #[error("Error dequeuing outbound message, peer not found: {peer_id}")]
+    DequeueOutbound { peer_id: PeerId },
 }
 
 impl ReplicationError {
     #[must_use]
     pub const fn peer_id(&self) -> Option<&PeerId> {
         match self {
-            Self::OpenStream { peer_id, .. } | Self::Io { peer_id, .. } => Some(peer_id),
+            Self::OpenStream { peer_id, .. }
+            | Self::Io { peer_id, .. }
+            | Self::DequeueOutbound { peer_id } => Some(peer_id),
         }
     }
 }
@@ -69,6 +73,7 @@ impl Clone for ReplicationError {
                     )),
                 },
             },
+            Self::DequeueOutbound { peer_id } => Self::DequeueOutbound { peer_id: *peer_id },
         }
     }
 }
@@ -127,6 +132,8 @@ struct PendingOutbound {
     last_scheduled: Option<PeerId>,
 }
 
+struct PeerNotFound;
+
 impl PendingOutbound {
     /// Appends a message to the peer's pending queue. Lazily creates the queue
     /// for the peer upon the first call.
@@ -134,15 +141,10 @@ impl PendingOutbound {
         self.messages.entry(peer_id).or_default().push_back(message);
     }
 
-    /// ### Panics
-    ///
-    /// [`Self::enqueue_message`] must be called at least once for the peer
-    /// before calling this method or it will panic.
-    fn dequeue_message(&mut self, peer_id: &PeerId) -> Option<DaMessage> {
-        self.messages
-            .get_mut(peer_id)
-            .expect("Peer is in the map")
-            .pop_front()
+    /// Dequeues the next message from the peer's pending queue.
+    fn dequeue_message(&mut self, peer_id: &PeerId) -> Result<Option<DaMessage>, PeerNotFound> {
+        let messages = self.messages.get_mut(peer_id).ok_or(PeerNotFound)?;
+        Ok(messages.pop_front())
     }
 
     /// Do internal housekeeping when a peer disconnects.
@@ -438,7 +440,15 @@ where
         // we can write the next pending message to this stream if there is one
         match self.outbound_tasks.poll_next_unpin(cx) {
             Poll::Ready(Some(Ok((peer_id, stream)))) => {
-                match self.pending_outbound.dequeue_message(&peer_id) {
+                let Ok(message) = self.pending_outbound.dequeue_message(&peer_id) else {
+                    return Poll::Ready(ToSwarm::GenerateEvent(
+                        ReplicationEvent::ReplicationError {
+                            error: ReplicationError::DequeueOutbound { peer_id },
+                        },
+                    ));
+                };
+
+                match message {
                     Some(message) => {
                         self.outbound_tasks
                             .push(Box::pin(Self::try_write_message(peer_id, message, stream)));
@@ -469,10 +479,13 @@ where
         });
 
         if let Some(peer_id) = next_peer {
-            let message = self
-                .pending_outbound
-                .dequeue_message(&peer_id)
-                .expect("Message is in the queue, ensured by PendingOutbound::next_peer()");
+            let Ok(message) = self.pending_outbound.dequeue_message(&peer_id) else {
+                return Poll::Ready(ToSwarm::GenerateEvent(ReplicationEvent::ReplicationError {
+                    error: ReplicationError::DequeueOutbound { peer_id },
+                }));
+            };
+            let message =
+                message.expect("Message is in the queue, ensured by PendingOutbound::next_peer()");
 
             match self.outbound_streams.entry(peer_id) {
                 Entry::Occupied(mut entry) => match entry.get_mut().take() {


nomos-da/network/core/src/protocols/replication/behaviour.rs
```diff
@@ -156,361 +218,254 @@ where
...
+        if self.waker.is_none() {
+            self.waker = Some(cx.waker().clone());
         }
 
-        // Assert that no members of other subnets have received the message.
-        for behaviour in &other_behaviours {
-            assert!(behaviour.seen_message_cache.is_empty());
+        // Only wake if we have a reason to
+        if should_wake {
+            self.wake();
```
bacv, line 466: Lets always clone the most recent waker that comes from the context.

nomos-da/network/core/src/protocols/replication/behaviour.rs
```diff
@@ -156,361 +218,254 @@ where
...
+        match self.incoming_tasks.poll_next_unpin(cx) {
+            Poll::Ready(Some(Ok((peer_id, message, stream)))) => {
+                // Replicate the message to all connected peers from the same subnet if we
+                // haven't seen it yet
+                self.replicate_message(message.clone());
+                // Wait for any next incoming message on the same stream
+                self.incoming_tasks
+                    .push(Self::try_read_message(peer_id, stream).boxed());
+                // Signal to the swarm that we've received the message
+                return Poll::Ready(ToSwarm::GenerateEvent(ReplicationEvent::IncomingMessage {
```
bacv, line 366: I know this is WIP, but if we return here, behaviour will only start processing outgoing tasks after all incoming tasks are finished - in theory we might never process outgoing tasks. I would recommend to only process one outgoing + one incoming task during the poll.

nomos-da/network/core/src/protocols/replication/behaviour.rs
```diff
@@ -156,361 +218,254 @@ where
...
+            }
+            Poll::Ready(Some(Err(error))) => {
+                debug!("Error on replication stream {error:?}");
+                // TODO propagate error ?
+            }
+            _ => {}
         }
+        // Schedule writing pending messages to connected peers if possible,
+        // ie. if the streams for those peers are idle or haven't even been opened yet
+        for (peer_id, messages) in &mut self.pending_outgoing_messages {
```
bacv, line 408: Lets also process one `pendiding_outgoing_message` at a time, just in case we have a lot and would like other behaviours to make some progress.

nomos-da/network/core/src/protocols/replication/behaviour.rs
```diff
@@ -156,361 +218,254 @@ where
...
+                Entry::Vacant(entry) => {
+                    entry.insert(StreamState::Busy);
+                    let message = messages
+                        .pop_front()
+                        .expect("At least one message is queued");
+                    let control = self.control.clone();
+                    // Avoid "coercion requires that `'1` must outlive `'static`" compilation error
+                    let peer_id = *peer_id;
+                    self.outgoing_tasks.push(Box::pin(async move {
+                        let stream = Self::try_open_stream(peer_id, control).await?;
```
bacv, line 442: I know that we talked about having stream per request, but since replication request doesn't have response or acknowledgment - we can create one stream when incoming or outgoing connection is established.

bacv (commented):
I know this is WIP and my comments might be what you were planning all along, will review more thoroughly when it's ready for review.

nomos-da/network/core/src/protocols/replication/behaviour.rs
bacv: For sure, it will be easier to see in another PR, for now we can keep it as is.

nomos-da/network/core/src/protocols/replication/behaviour_v2.rs
```diff
@@ -0,0 +1,677 @@
...
+
+        // Schedule reading from any new incoming streams if possible
+        if let Poll::Ready(Some((peer_id, stream))) = self.incoming_streams.poll_next_unpin(cx) {
+            self.incoming_tasks
+                .push(Self::try_read_message(peer_id, stream).boxed());
+            should_wake = true;
+        }
+
+        if let Some(incoming_message) = incoming_message {
+            return incoming_message;
```
bacv, line 506: The future won't be polled if we have incoming message, I would suggest to change all the `should_wake` with the actual call to `waker().wake_by_ref()` - it can be called multiple times, but the future will be polled only once next time.

nomos-da/network/core/src/protocols/replication/behaviour_v2.rs
```diff
@@ -0,0 +1,677 @@
...
+
+        self.connected
+            .iter()
+            .filter(|peer_id| peers.contains(peer_id))
+            .for_each(|peer_id| {
+                self.pending_outbound
+                    .enqueue_message(*peer_id, message.clone());
+            });
+
+        self.try_wake();
```
bacv, line 276: Since `send_message` and `replicate_messge` are being called in poll only, we could just pass the `cx` and wake by ref instead of storing the waker. Storing waker in such network behaviours is a good solution when we want to trigger polling when receiving a new swarm event or something is needed to be reevaluated on new connection.

nomos-da/network/core/src/protocols/replication/behaviour_v2.rs
```diff
@@ -0,0 +1,677 @@
...
+    }
+
+    /// ### Panics
+    ///
+    /// [`Self::enqueue_message`] must be called at least once for the peer
+    /// before calling this method or it will panic.
+    fn dequeue_message(&mut self, peer_id: &PeerId) -> Option<DaMessage> {
+        self.messages
+            .get_mut(peer_id)
+            .expect("Peer is in the map")
```
bacv, line 144: Couldn't the `expect` here be avoided an `None` returned instead?

bacv (changes requested):
Couple requested changes regarding `wake` and `dequeue_message`

nomos-da/network/core/src/protocols/replication/behaviour_v2.rs
danielSanchezQ: Indeed. And probably just logged that the peer is missing in that case.

nomos-da/network/core/src/protocols/replication/behaviour_v2.rs
```diff
@@ -0,0 +1,677 @@
...
+{
+    /// Check if some peer membership lies in at least a single subnetwork that
+    /// the local peer is a member too.
+    fn is_neighbour(&self, peer_id: &PeerId) -> bool {
+        self.membership
+            .membership(&self.local_peer_id)
+            .intersection(&self.membership.membership(peer_id))
+            .count()
+            > 0
+    }
```
danielSanchezQ, line 245: I thought something like this was already in membership. If it is not, we should probably make it part of the membership API. This is because now membership is a local copy of the data, but at some point it is probably gonna be a service itself. We can left it like that now, and open an issue to track this and do it later on.

nomos-da/network/core/src/protocols/replication/behaviour_v2.rs
```diff
@@ -0,0 +1,677 @@
...
+    connected: HashSet<PeerId>,
+    /// Pending outbound messages are stored here, they are then consumed by
+    /// each respective outbound stream
+    pending_outbound: PendingOutbound,
+    /// Indicates which outbound streams are currently idle or busy
+    outbound_streams: HashMap<PeerId, StreamState>,
+    /// Holds tasks for writing messages to the outbound streams
+    outbound_tasks: FuturesUnordered<OutboundTask>,
+    /// Seen messages cache holds a record of seen messages
+    seen_message_cache: IndexSet<(Vec<u8>, SubnetworkId)>,
```
danielSanchezQ, line 197: This probably should be a proper chache. Otherwise it will explode with time. Some RLU with a proper policy. Can be done in a different PR but lets open an issue to track it.

nomos-da/network/core/src/protocols/replication/behaviour_v2.rs
```diff
@@ -0,0 +1,677 @@
...
+
+    fn no_loopback_member_peers_of(&self, subnetwork: SubnetworkId) -> HashSet<PeerId> {
+        let mut peers = self.membership.members_of(&subnetwork);
+        // no loopback
+        peers.remove(&self.local_peer_id);
+        peers
+    }
+
+    fn replicate_message(&mut self, message: &DaMessage) {
+        let message_id = (message.blob.blob_id.to_vec(), message.subnetwork_id);
```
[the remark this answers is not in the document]
bacv: Closing this comment, metadata is not sent with the dispersed blob, it's added to the mempool together with the blob_id, lets leave this as is, the fix needs to be applied in da-services, not networking level.

danielSanchezQ (approved):
This shouldn't really be `v2`. But we can make the move in a different PR. 
Thanks!

PR #1106, opened 2025-03-07
chore: set up clippy::restriction lints

ntn-x2:

Introduces `clippy::restriction` lints in a much more digestible way. It sets up the `.cargo/config.toml` with all the lints we want to allow/tackle, and only tackles the `clippy::allow_attributes` and `clippy::allow_attributes_without_reason` lints in this PR, which should be quite uncontroversial 😃 and would also force us to put in writing the assumptions behind ignoring warnings, where applicable. Some places where the `clippy::type_complexity` simply has to be, we skip adding a message. All other places should comply with this.

Other lints will be introduced in small PRs, where the team can decide whether to enforce the lint, or move it into the set of allowed lints.

It also removes the unused `task: JoinHandle` field since it was a leftover from a previous refactoring.

Yes

@ntn-x2 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 1. Description added.


nomos-services/blend/src/backends/libp2p.rs
```diff
@@ -16,18 +16,13 @@ use nomos_libp2p::{secret_key_serde, NetworkBehaviour};
...
-};
+use tokio::sync::{broadcast, mpsc};
 use tokio_stream::wrappers::BroadcastStream;
 
 use super::BlendBackend;
 
 /// A blend backend that uses the libp2p network stack.
 pub struct Libp2pBlendBackend {
-    #[expect(dead_code)]
-    task: JoinHandle<()>,
```
danielSanchezQ, line 30: I think this is wrong. In this case the task should be wrapped in an abortable and get aborted when stopping the service if possible. That is why it probably was there.

nomos-services/data-availability/indexer/src/lib.rs
```diff
@@ -550,7 +551,10 @@ where
         let storage_adapter = DaStorage::new(storage_relay).await;
 
         let mut lifecycle_stream = service_state.lifecycle_handle.message_stream();
-        #[expect(clippy::redundant_pub_crate)]
+        #[expect(
+            clippy::redundant_pub_crate,
+            reason = "Generated by `tokio::select` macro."
+        )]
```
danielSanchezQ, line 557: I've seen a few of this now. I forgot to ask in the previous prs, why does this triggers with `tokio::select`?

nomos-services/mempool/src/backend/mockpool/mod.rs
```diff
@@ -146,9 +146,10 @@ where
     }
 
     fn view(&self, _ancestor_hint: BlockId) -> Box<dyn Iterator<Item = Self::Item> + Send> {
-        // we need to have an owned version of the iterator to bypass adding a lifetime
-        // bound to the return iterator type
-        #[expect(clippy::needless_collect)]
+        #[expect(
+            clippy::needless_collect,
+            reason = "We need to have an owned version of the iterator to bypass adding a lifetime bound to the return iterator type"
+        )]
```
danielSanchezQ, line 152: This actually could be solved now with the `use` capture keyword.

tests/src/tests/cryptarchia/happy.rs
```diff
@@ -31,7 +31,10 @@ async fn happy_test(topology: &Topology) {
         .floor() as u64;
     let timeout = adjust_timeout(Duration::from_secs(timeout));
     let timeout = tokio::time::sleep(timeout);
-    #[expect(clippy::redundant_pub_crate)]
+    #[expect(
+        clippy::redundant_pub_crate,
+        reason = "Generated by `tokio::select` macro."
+    )]
```
danielSanchezQ, line 37: This is actually bothersome. It feels it really pollutes the code. I guess at some point that will be fixed by tokio guys 🤔

danielSanchezQ (approved):
Left a few comments, but overall I'm ok with the changes. I'm quite happy having messages in the `expect` clauses instead of comments above. Thanks.

tests/src/tests/cryptarchia/happy.rs
ntn-x2: Yes I think so too. An alternative approach would be to declare our own macro that wraps around this, and have the `expect` in a single place. What do you think?

nomos-services/blend/src/backends/libp2p.rs
ntn-x2: Tests kept passing and @youngjoon-lee said it was a leftover from a previous refactoring. I am personally not sure why it's there. Should I revert the change for now?

ntn-x2: Actuall after thinking about this a bit, it wasn't used anyway here, so it was just something for future reference. I would say we remove it here, since the lint enforces it, and then we can re-add it AND make use of it once we want to use it. Sounds good?

youngjoon-lee: @danielSanchezQ Ah, now I can recall the history. Thanks for reminding. 

When I first implemented it, I saved JoinHandle for later use for [Abortable](https://docs.rs/futures/latest/futures/future/struct.Abortable.html). At that time, I referred to DA network backend code. Later, [Abortable was added to DA code](https://github.com/logos-co/nomos-node/pull/853/files#diff-618f063bf12ec5d0d7faae32b79afd2d86d2a6db75d46b8abe0d4fb78489afd2L74), but I didn't change this Blend code yet. I should have added TODO comment (as we did in DA). Sorry for the confusion.

First of all, I agree to remove this for making clippy happy.I will open issue right now, review if adding Abortable is necessary and work on it.

nomos-services/data-availability/indexer/src/lib.rs
ntn-x2: Because the macro generates a `pub(crate) enum` which then triggers the warning.

tests/src/tests/cryptarchia/happy.rs
ntn-x2: There has been discussion going on, and it seems the issue [has been fixed and merged recently](https://github.com/rust-lang/rust-clippy/pull/13952), so I would expect it to be gone in one of the upcoming Rust updates. In the meanwhile, I think it would still be ok to refactor the macro by wrapping it into another macro that will ignore the lint.

ntn-x2: Ok it does not seem to be possible, at the moment, and as I said in another comment, it seems to have been fixed in an upcoming Clippy version, so I would suggest we leave it as it is for now, as bothersome as it might seem.

nomos-services/mempool/src/backend/mockpool/mod.rs
ntn-x2: This is a bigger change because it seems to imply a change in the function signature of the trait, which would have cascading effects in other parts of the code as well, so I will not tackle it here.

PR #1108, opened 2025-03-07
feat(DA): Sampling requests commitments and blobs separately

andrussal:

In this PR, sampling network sends only light blobs and requests shared commitments separately. 
 
**Implementation details**

- Sampling transfers only light blobs(blobs without shared commitments)
- Retrieve commitments separately, either from local storage or over HTTP API
- Introduced `ApiBackend` type and `HttpApiBackend` implementation
- Move shared requests types from nodes to `nomes-da-messages`. This change is probably temporary and these types are better off in `nomos-lib`(not created yet)
- Incomplete Membership and AddressBook handling. It seems better to finish this part after work on "global" AddressBook is finished

---

_PR is a quite bloated because new generic type and switching from blob to light blob. Please let me know if I can somehow simplify it!_

---

**Note on Testing sampling flow**

I am not sure how to trigger the edge case in integration tests when we can’t find shared commitments locally and need to requests these over HTTP API.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 1. Description added.


nomos-core/chain-defs/src/da/blob/mod.rs
```diff
@@ -5,8 +5,8 @@ pub mod select;
 pub trait Blob {
     type BlobId;
     type ColumnIndex;
-    type LightBlob;
-    type SharedCommitments;
+    type LightBlob: Send + Sync;
+    type SharedCommitments: Send + Sync;
```
danielSanchezQ, line 9: This bound should probably not be here but where this bounds are used.

clients/common-http-client/src/lib.rs
```diff
@@ -29,19 +33,20 @@ impl BasicAuthCredentials {
...
-    client: Client,
+    client: Arc<Client>,
     base_address: Url,
     basic_auth: Option<BasicAuthCredentials>,
 }
 
 impl CommonHttpClient {
     #[must_use]
-    pub fn new(base_address: Url, basic_auth: Option<BasicAuthCredentials>) -> Self {
+    pub fn new(base_address: &str, basic_auth: Option<BasicAuthCredentials>) -> Self {
```
danielSanchezQ, line 43: Why not` URL`?

nomos-da/kzgrs-backend/benches/verifier.rs
```diff
@@ -32,25 +32,29 @@ fn verify<const SIZE: usize>(bencher: Bencher, column_size: usize) {
...
                     .iter()
                     .map(|row| row.iter().next().copied().unwrap())
                     .collect(),
             };
-            (verifier, da_blob)
+            let commitments = DaBlobSharedCommitments {
+                aggregated_column_commitment: encoded_data.aggregated_column_commitment,
+                rows_commitments: encoded_data.row_commitments,
+            };
+            (verifier, commitments, da_blob)
```
danielSanchezQ, line 54: Not to do in this PR maybe. But it would be nice to have methods to extract parts of the DA blob. One for each or an `into_parts(self)->(DaBlobSharedCommitments, DaLighBlob)` method.

nomos-da/network/messages/src/common.rs
```diff
@@ -16,6 +16,20 @@ impl Blob {
...
+    pub data: DaLightBlob,
+}
+
+impl LightBlob {
+    #[must_use]
+    pub const fn new(blob_id: BlobId, data: DaLightBlob) -> Self {
+        Self { blob_id, data }
+    }
+}
+
```
danielSanchezQ, line 32: if we changed verifier signature to use this, then this should go into `kzrs-backend` instead.

nomos-services/data-availability/network/src/backends/libp2p/common.rs
```diff
@@ -42,12 +45,15 @@ pub struct DaNetworkBackendSettings<Membership> {
 #[derive(Debug, Clone)]
 pub enum SamplingEvent {
     /// A success sampling
-    SamplingSuccess { blob_id: BlobId, blob: Box<DaBlob> },
+    SamplingSuccess {
+        blob_id: BlobId,
+        light_blob: Arc<DaLightBlob>,
```
danielSanchezQ, line 50: Why `Arc`?

nomos-services/data-availability/sampling/src/api/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+#[async_trait]
+pub trait ApiBackend {
+    type Settings: Clone;
+    type Request: Send;
+    type Response: Send;
+    fn new(settings: Self::Settings) -> Self;
+    async fn run(
+        self,
+    ) -> Result<(UnboundedSender<ApiRequest>, UnboundedReceiver<ApiResponse>), DynError>;
+}
```
danielSanchezQ, line 18: I think this is the wrong abstraction for this. we do not want a running a running task for this. This should be a one-shot client that is used in an async context when needed. Or a Stream/Sink we can poll, but im not sure about we can implement the later easily, so just a wrapper around a reqwest client should be enough.

nomos-services/data-availability/sampling/src/lib.rs
```diff
@@ -114,18 +135,29 @@ where
...
+                let mut sampling_state = sampler.prepare_sampling(blob_id).await;
+                if matches!(sampling_state, SamplingState::WaitingCommitments) {
+                    info_with_id!(blob_id, "WaitingCommitments");
+                    if let Some(commitments) =
+                        Self::request_commitments(storage_adapter, api_request, blob_id).await
+                    {
+                        sampling_state = sampler.init_sampling(blob_id, commitments).await;
+                    }
+                }
+                if let SamplingState::Init(sampling_subnets) = sampling_state {
```
danielSanchezQ, line 159: This feels a bit wrong. Waiting for a state then waiting for the next. Most of this should be handled by the sampler. Once the sampler finishes it would return either a fail state with whatever it has if we need it outside (for example the commitments (?)), or the success with everything as well. Also this part is just changing this week. We can sync about it next week.

danielSanchezQ (changes requested):

nomos-services/data-availability/network/src/backends/libp2p/common.rs
andrussal: I wanted to have single `SharedCommitments` instance hold by `SamplingContext` and share it with verifier without cloning it. I am not 100% sure if my reasoning is correct though.

nomos-da/kzgrs-backend/benches/verifier.rs
andrussal: I ended up changing verifier signature because I wanted to pass `DaBlobSharedCommitments` separately because it's separate in sampler and I tried to avoid cloning. Not sure if necessary though.

nomos-services/data-availability/sampling/src/api/mod.rs
andrussal: I was worried that one slow http request can block the whole sampling flow. But maybe just depend on short timeouts?

nomos-services/data-availability/sampling/src/lib.rs
andrussal: Agree, it's bit convoluted. Idea was that if we find commitments from local storage, we can just jump to next stage.

nomos-services/data-availability/sampling/src/lib.rs
```diff
@@ -197,20 +242,80 @@ where
...
+        sampler: &mut Backend,
+    ) {
+        match api_response {
+            ApiResponse::Commitments {
+                blob_id,
+                commitments: Some(commitments),
+            } => {
+                info_with_id!(blob_id, "ReceivedCommitments");
+                if let SamplingState::Init(sampling_subnets) =
+                    sampler.init_sampling(blob_id, commitments).await
```
bacv, line 277: If ApiRequest included the oneshot response channel, then sampling would be triggered in one place.

nomos-services/data-availability/sampling/src/api/http.rs
```diff
@@ -0,0 +1,194 @@
...
+        commitments: Option<DaBlobSharedCommitments>,
+    },
+    Error {
+        blob_id: BlobId,
+        error: DynError,
+    },
+}
+
+pub struct HttApiBackend {
+    clients: Vec<Arc<CommonHttpClient>>,
```
bacv, line 41: I know that that with current CommonHttpClient we need to provide the base url, but having a client per address feels a bit off. Maybe in future pr we could remove base url per instance, and have one http client, or allow `CommonHttpClient` to be initialized with provided reqwest client.

bacv (changes requested):
Sampling triggering seems bit complex, I think waiting for shared commitments from storage or http response could happen in one place.

nomos-services/data-availability/network/src/backends/libp2p/common.rs
danielSanchezQ: I would have to check deeper. But I do not think that would be needed. That sampling lightblob after validation has no use. It would just be stored. We can revisit this data flow.

andrussal:
@danielSanchezQ and @bacv thank you for taking time to review this pretty big PR. All your requests made sense. I left `client` update to another PR. Also `cargo-deny` `RUSTSEC` stuff keeps coming up. I am not sure how to handle these right now, probably each need to be addressed in a separate PR(if possible).

nomos-da/network/messages/src/common.rs
andrussal: Actually I think verifier signature has DaLightBlob. This is the type that replaced `Blob` in the same file. Please let me know if we should nevertheless move it.

nomos-services/data-availability/sampling/src/api/http.rs
andrussal: Will do in a separate PR 👍

bacv:
> Also cargo-deny RUSTSEC stuff 

The `deny` rules were merged into the master (https://github.com/logos-co/nomos-node/pull/1110), the issue should go away after this branch is updated.

danielSanchezQ:
Andrus. Naming changed in the specification and it would be probably better to have consistency. We can do it in a different PR just for that. But just so you are aware and its done after we merge this. It should be fairly simple.

nodes/nomos-node/config.yaml
```diff
@@ -201,6 +201,9 @@ da_sampling:
   network_adapter_settings: null
   storage_adapter_settings:
     blob_storage_directory: ./
+  api_adapter_settings:
+    addresses:
+      - http://127.0.0.1:8722
```
danielSanchezQ, line 206: Now that I see this. This should be a membership in the same fashion we have in the da-network-backend, as it need to be shared and updated. Can we add that?

danielSanchezQ (commented):
Lets do what I commented in a different PR so I have time to review naming on specs.

danielSanchezQ (approved):
@andrussal leave the duplicated membership configuration as it is now. This will change during next iteration where we will mock the RDP service.

.cargo-deny.toml
```diff
@@ -13,7 +13,7 @@ ignore = [
   { id = "RUSTSEC-2024-0384", reason = "`instant` is unmaintained; consider using an alternative. Use `cargo tree -p instant -i > tmp.txt` to check the dependency tree." },
   { id = "RUSTSEC-2024-0388", reason = "`derivative` is unmaintained; consider using an alternative. Use `cargo tree -p derivative -i > tmp.txt` to check the dependency tree." },
   { id = "RUSTSEC-2024-0436", reason = "`paste` has a security vulnerability; consider using an alternative. Use `cargo tree -p paste -i > tmp.txt` to check the dependency tree." },
-  { id = "RUSTSEC-2025-0014", reason = "humantime is unmaintained" },
+  { id = "RUSTSEC-2025-0014", reason = "Latest humantime crates.io release is four years old and GitHub repository has not seen commits in four years"}
```
danielSanchezQ, line 16: This change is not necessary. I think it comes from missing merge probably.

nomos-da/kzgrs-backend/benches/verifier.rs
```diff
@@ -22,7 +22,8 @@ pub fn rand_data(elements_count: usize) -> Vec<u8> {
     thread_rng().fill_bytes(&mut buff);
     buff
 }
-#[divan::bench(consts = [32, 64, 128, 256, 512, 1024], args = [128, 256, 512, 1024, 2048, 4096], sample_count = 1, sample_size = 30)]
+#[divan::bench(consts = [32, 64, 128, 256, 512, 1024], args = [128, 256, 512, 1024, 2048, 4096], sample_count = 1, sample_size = 30
+)]
```
danielSanchezQ, line 26: Something went wrong here.

danielSanchezQ (changes requested):

.cargo-deny.toml
andrussal: Corrected now, I already had it in my branch and didn't update from master.

danielSanchezQ (approved):

bacv (approved):

PR #1112, opened 2025-03-10
fix: packages and dependencies features

ntn-x2:

This PR does two things:

1. Make sure that no implicit feature is exposed that has the same name as an optional dependency (see [Cargo docs](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies) for more info). This ensures we always have full control over what is exposed to other crates. Unfortunately, I did not find any way to enforce this, so we have to enforce it when PRs are reviewed
2. Introduce [cargo-hack](https://github.com/taiki-e/cargo-hack) to check for all feature combinations for all crates in the workspace. This caught some feature mishandling that has also been addressed in this PR. The check is added as a job to the `code-check` GH workflow. How I started checking for features is because when running `cargo clippy -p nomos-mempool --all-features` some other crate that the mempool crate was depending on was throwing an error because of a feature mishandling. This checks take a while the first time they are run (~ 30 min), but they are very cache-friendly so subsequent runs (including in the CI) take much less time (~ 5 min).

I could also run tests with all feature combinations, which I feel it's something we should do at some point, but that would take too much time so for now we don't do that.

Yes.

@ntn-x2 

N/A

N/A

## Checklist

* [x] 1. Description added.

### DIFF
--- a/.github/workflows/code-check.yml
+++ b/.github/workflows/code-check.yml
@@ -1,9 +1,9 @@
-name: Code checks
-
 on:
   pull_request:
   push:
     branches: [ master ]
+
+name: Code checks
 
 concurrency:
   group: ${{ github.workflow }}@${{ github.ref }}
--- a/.github/workflows/code-check.yml
+++ b/.github/workflows/code-check.yml
@@ -150,6 +150,7 @@ jobs:
     needs:
       - formatting
       - cargo-deny
+      - features
       - lints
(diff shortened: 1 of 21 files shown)


bacv (approved):
This will help running individual tests more easily, I think it's good addition.

danielSanchezQ (approved):

PR #1121, opened 2025-03-11
chore: RUSTSEC-2025-0014 humantime is unmaintained

bacv:

Allow humantime in cargo-deny.

https://github.com/tailhook/humantime/issues/31

N/A

N/A

N/A

## Checklist

* [x] 1. Description added.


ntn-x2:
Just found this related issue from the Apache Arrow project: https://github.com/apache/arrow-rs/issues/7264.

bacv:
> Just found this related issue from the Apache Arrow project

We could consider the migration.

danielSanchezQ:
> Just found this related issue from the Apache Arrow project: [apache/arrow-rs#7264](https://github.com/apache/arrow-rs/issues/7264).

Yeah. This are little things that we could tackle now that we have the time. Can we actually open an issue so we can fix it during next iteration?

ntn-x2:
Yeah I think we should aim to have an empty `cargo-deny` config file before we reach production. Some dependencies, like this one, are easier to replace. Some others might be more complicated, but I think we should aim to address all the warnings cargo deny raises. I created a ticket for this one dependency in the meanwhile: `https://github.com/logos-co/nomos-node/issues/1125`.

PR #1124, opened 2025-03-11
fix(cfgsync): Dispersal mempool push strategy in cfgsync sample yaml

bacv:

Sample config parameters for cfgsync were incorrect.

## Checklist

* [x] 1. Description added.

### DIFF
--- a/testnet/cfgsync.yaml
+++ b/testnet/cfgsync.yaml
@@ -21,12 +21,12 @@ balancer_interval_secs: 5
 # Dispersal mempool publish strategy
 mempool_publish_strategy: !SampleSubnetworks
   sample_threshold: 2
-  dispersal_timeout:
+  timeout:
     secs: 2
     nanos: 0
   cooldown:
     secs: 0
-    nanos: 100_000_000
+    nanos: 100000000
 
 # Tracing
 tracing_settings:


testnet/cfgsync.yaml
```diff
@@ -21,12 +21,12 @@ balancer_interval_secs: 5
...
 mempool_publish_strategy: !SampleSubnetworks
   sample_threshold: 2
-  dispersal_timeout:
+  timeout:
     secs: 2
     nanos: 0
   cooldown:
     secs: 0
-    nanos: 100_000_000
+    nanos: 100000000
```
danielSanchezQ, line 29: Why not seconds?

bacv: This one is used for retrying failed samples after the dispersal but before publishing into the mempool - we expect verification to be performed very fast, but still might want to wait before retrying immediatly, 1 second seems too high, but 100ms just about right. This config is used in testnet, up to the user to decide optimal cooldown.

https://github.com/logos-co/nomos-node/pull/1104/files#diff-abd93eb3d6f07e5ade2afcd4ecb67b381a5a11da1957e5a0baddc4729547c9cbR204-R209

danielSanchezQ: Ah, I did bad math. Makes sense now.

PR #1129, opened 2025-03-12
fix(DA): Disperse succeeds when data and metadata is the same

bacv:

When the same data and metadata are dispersed repeatedly - dispersal succeeds, but response from the executor mempool is `Err(ExistingItem)`. Before, this error was considered as failure by Dispersal service, now it is considered as success.

Yes

@bacv, @danielSanchezQ 

Yes

No

## Checklist

* [x] 1. Description added.

### DIFF
--- a/nomos-services/data-availability/dispersal/src/adapters/mempool/mod.rs
+++ b/nomos-services/data-availability/dispersal/src/adapters/mempool/mod.rs
@@ -1,9 +1,27 @@
 pub mod kzgrs;
 
+use std::error::Error;
+
+use nomos_mempool::backend::MempoolError;
 use overwatch::{
     services::{relay::OutboundRelay, ServiceData},
     DynError,
 };
+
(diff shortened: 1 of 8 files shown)


nomos-services/data-availability/dispersal/src/adapters/mempool/mod.rs
```diff
@@ -1,10 +1,28 @@
...
+
+impl DaMempoolAdapterError {
+    fn to_other<E>(e: E) -> Self
+    where
+        E: Error + Send + Sync + 'static,
+    {
+        Self::Other(Box::new(e) as DynError)
+    }
+}
+
```
danielSanchezQ, line 25: I think it would be better to implement `impl<E: Error + Send + Sync + 'static> From<E> for DaMempoolAdapterError`. It can just be added.

pradovic (approved):

PR #1131, opened 2025-03-12
feat(api): peer blocking

pradovic:

This PR exposes http api calls for block peer, unblock peer and listing blacklisted peers for both executor and validator.

HTTP call timeouts and support for non-existing peers will be implemented in a subsequent PR.

Yes.

1. Http API for block/unblock and list blacklisted peers 
2. There is a new unit test for `ConnectionMonitorBehavior`
3. Slight changes to `ConnectionMonitorBehavior` poll to fix  poll hanging bugs found by integration tests
4. Integraton tests for these api calls

@bacv , @pradovic 

Yes.

It adds 3 new http calls.

## Checklist

> [!WARNING]  

* [x] 1. Description added.


tests/src/tests/da/api.rs
```diff
@@ -52,3 +56,62 @@ async fn test_get_blob_data() {
...
+    let unblocked = executor.unblock_peer(existing_peer_id.to_string()).await;
+    assert!(unblocked);
+
+    let blacklisted_peers = executor.blacklisted_peers().await;
+    assert!(blacklisted_peers.is_empty());
+
+    let unblocked = executor.unblock_peer(existing_peer_id.to_string()).await;
+    assert!(!unblocked);
+
+    // try blocking/unblocking non existing peer id
```
pradovic, line 102: These tests allow blocking/unblocking non existing peers. The support for peers (via membership) could be added in a different PR.

nomos-services/api/src/http/da.rs
```diff
@@ -269,3 +280,89 @@ where
...
+    B::Message: PeerMessagesBuilder,
+{
+    let relay = handle.relay::<NetworkService<B>>().connect().await?;
+    let (sender, receiver) = oneshot::channel();
+    let message = B::Message::create_block_message(peer_id, sender);
+    relay
+        .send(DaNetworkMsg::Process(message))
+        .await
+        .map_err(|(e, _)| e)?;
+    Ok(receiver.await?)
```
pradovic, line 296: Timeout (internal server error) should be added here imho, but I left it for a different PR as I would like to add it to other http calls as well, if it makes sense.

nomos-services/api/src/http/da.rs
```diff
@@ -269,3 +280,89 @@ where
...
+}
+
+// Builder for generating messages for peers (validator and executor).
+pub trait PeerMessagesBuilder {
+    type MessageType;
```
ntn-x2, line 331: What's the goal of this associated type here? All methods return `Self`, and this type is `Self` for all impls, so I am not sure I understand its purpose?

pradovic: There is a duplication of messages in Validator and Executor, so they have "different messages. I talked with the guys and consensus was to keep those separated. So the message that needs to be created is either `ExecutorDaNetworkMessage` or `DaNetworkMessage` (validator one), so there are correspond implementations for those, which are the same as they both create Self::PeerRequest, which is defined in `nomos/nomos-services/data-availability/network/src/backends/libp2p/common`.
So I added a simple builder trait to build appropriate message based on the Message type.

pradovic: I edited my message, I see what you mean now, yeah the type does not have to be included at all. Good catch!

pradovic: Fixed. thank you!

ntn-x2: The way I would do this, is to rename the associated type to `Message` and for the methods to return `Self::Message`. For a builder to return `Self` means, at least in theory, that the builder is returning itself. I guess what we want from the builder are messages, right?

ntn-x2: Then you can add different trait bounds in different places where such message type is needed, without having to constraint the builder itself.

pradovic: Maybe I am missing something, feel free to show the example, pls. The way I just fixed eliminates the associated type completely, and fits the calling in handlers perfectly. I am not sure if we need to support any more use cases for this, tbh.

ntn-x2: What I mean is that it feels a bit unconventional for a builder to return `Self`, rather than `Message`. I think it would work just fine with a `Message` associated type and the methods returning `Self::Message` instead. But maybe we can wait for someone else to provide their input. I am not saying your solution does not work, as it clearly seems it does, but I am saying that the semantics of what I see here are perhaps not 100% "to the book".

pradovic: I see what you mean now. Yes, I think that for builders it is common to return self and have a `.build()` method in the end. But I have also seen across Rust places where builder type is actually a Message type, which is seen from the implementation and I feel like it is contributing to simplicity. That is why I have chosen the pattern, as we don't have a parametrized builder. 

But I don't have any strong feelings about it, happy to change if it's a bad idiom.

@bacv, @danielSanchezQ  let us know what do you guys think.

ntn-x2: Sorry I just noticed that the trait is not implemented for the builder, but for the actual message type already, for instance `DaMessage`. I think I was confused at first, but now I see it again, there is definitely not need for an associated type. Sorry for this misunderstanding, I think the current versions looks ok as it is 👍

ntn-x2: But I'll still let someone also approve the PR as a whole 😊

pradovic: Thinking about this more, maybe this is more of a factory then builder, so I will change the name :)

nomos-services/api/src/http/da.rs
danielSanchezQ: Let's open an issue to track it. We can add it to the next iteration. Thanks

nomos-services/api/src/http/da.rs
```diff
@@ -269,3 +280,83 @@ where
...
+        .map_err(|(e, _)| e)?;
+    Ok(receiver.await?)
+}
+
+// Builder for generating messages for peers (validator and executor).
+pub trait PeerMessagesBuilder {
+    fn create_block_message(peer_id: PeerId, sender: oneshot::Sender<bool>) -> Self;
+    fn create_unblock_message(peer_id: PeerId, sender: oneshot::Sender<bool>) -> Self;
+    fn create_blacklisted_message(sender: oneshot::Sender<Vec<PeerId>>) -> Self;
+}
```
danielSanchezQ: Was the long issue here? Somehow I cannot see the messages anymore, at least during this review. Anyway, maybe the confusion comes from a trait returning itself? In traits the `Self` type refers to the type implementing it not the trait itself. So I think this is fine.

nomos-services/api/src/http/da.rs
danielSanchezQ: > What I mean is that it feels a bit unconventional for a builder to return `Self`, rather than `Message`. I think it would work just fine with a `Message` associated type and the methods returning `Self::Message` instead. But maybe we can wait for someone else to provide their input. I am not saying your solution does not work, as it clearly seems it does, but I am saying that the semantics of what I see here are perhaps not 100% "to the book".

Ok, I can see the convo now after the review.
Check my comment here even if it is not relevant anymore: https://github.com/logos-co/nomos/pull/1131#discussion_r1995121942

danielSanchezQ (approved):
This looks clean. Thanks!

nomos-services/api/src/http/da.rs
pradovic: Yes, this is where the issue is. @ntn-x2 was pointing out that "Builder" was returning `Self` and not `Message`. I ended up renaming it to factory in the end, but actually good observation, Self is the actual type. Thanks for the input, and thank you @ntn-x2 for the correction as well.

bacv (approved):
Great work.

nomos-services/api/src/http/da.rs
pradovic: https://github.com/logos-co/nomos/issues/1137

nomos-services/api/src/http/da.rs
pradovic: @ntn-x2 Are you ok to merge this now?

ntn-x2: Sure, go ahead!

PR #1132, opened 2025-03-12
chore: remove unnecessary trait bounds on DaSamplingService impls

ntn-x2:

The `ServiceData` trait is what is shared between services to create relays between each other. When working on the DA mempool, I realized the trait bounds for the `DaSamplingService` to implement `ServiceData` were too many and not really necessary. Hence, I though about cleaning up the service, while keeping the same functionalities. Hence, this PR does the following:

* Switch from single-letter generics to self-descriptive names
* Remove unnecessary trait bounds on all the service's impl blocks
* Remove unused settings from the `DaSamplingServiceSettings`. Specifically, `network_adapter_settings` and `storage_adapter_settings` have been removed since they were not used in the previous implementation anyway
* Move logic from the `init` to the `run` method, since it's `async` and we can push more stuff into it that we do not necessarily have to do in the `init` method, namely recovering state.

Yes.

@ntn-x2 

N/A

No

## Checklist

* [x] 1. Description added.


ntn-x2:
@danielSanchezQ @bacv this is probably true for other services as well, and it's kinda annoying since those traits are enforced already at the struct level, meaning that interacting with any of these services require all those unnecessary bounds to be enforced. What I did here is to remove the trait bounds from the `ServiceData` (and struct) impls if they are only needed for the `ServiceCore` impl. This means that referring to the service struct from other services requires only the strictly set of trait bounds to be met.

danielSanchezQ:
> @danielSanchezQ @bacv this is probably true for other services as well, and it's kinda annoying since those traits are enforced already at the struct level, meaning that interacting with any of these services require all those unnecessary bounds to be enforced. What I did here is to remove the trait bounds from the `ServiceData` (and struct) impls if they are only needed for the `ServiceCore` impl. This means that referring to the service struct from other services requires only the strictly set of trait bounds to be met.

Yeah, this is a good catch. We can make these changes little by little to other services as well. And reduce the generic hell.

nomos-services/data-availability/sampling/src/lib.rs
```diff
@@ -301,163 +323,143 @@ where
...
+        Ok(Self::new(service_state))
     }
 
-    async fn run(self) -> Result<(), DynError> {
+    async fn run(mut self) -> Result<(), DynError> {
         // This service will likely have to be modified later on.
         // Most probably the verifier itself need to be constructed/update for every
         // message with an updated list of the available nodes list, as it needs
         // his own index coming from the position of his bls public key landing
         // in the above-mentioned list.
```
danielSanchezQ, line 418: Now that I see this comment, is not needed anymore.

nomos-services/data-availability/sampling/src/lib.rs
```diff
@@ -301,163 +323,143 @@ where
...
-            network_relay,
-            storage_relay,
-            verifier_relay,
-            mut service_state,
-            mut sampler,
-            ..
-        } = self;
-
-        let network_relay = network_relay.connect().await?;
-        let mut network_adapter = DaNetwork::new(network_relay).await;
```
danielSanchezQ, line 442: I think we should refrain from changing the style unless there is a specific reason for it. These changes seem meaningless in the context of the PR as well unless I'm missing something.

ntn-x2: Ops you are right. I based this on the tx mempool structure I worked on. I will revert these changes, no problem.

nomos-services/data-availability/sampling/src/lib.rs
ntn-x2: Removed in https://github.com/logos-co/nomos/pull/1132/commits/79c4d0f18b8d7db6c21505339a401f07b39c1940.

nomos-services/data-availability/sampling/src/lib.rs
ntn-x2: Reverted in https://github.com/logos-co/nomos/pull/1132/commits/f040b4004e1f302d7ea105fbcabe59a38788ebf5.

bacv:
This will simplify everything by a lot, thanks for looking into this.

PR #1134, opened 2025-03-13
fix: Return mempool error when AddItem fails

bacv:

In mempool [refactor] one line was missed which returns MempoolError through reply channel.
In this PR I also added the same error reporting to tx mempool. Because of a new line added to the message handling code - clippy complaints about cognitive complexity - we can either allow it or extract function as done in DA.

Please share you preference and I'll update other pool to match the decision.

A line was missed in a refactor https://github.com/logos-co/nomos/pull/1133/files#diff-d5ae27d19a7f2625f3277f74b24650e441ea8b9ac1e0de0f3fe2cadd15cef25bL395

|@bacv @ntn-x2 

N/A

N/A

## Checklist

* [x] 1. Description added.

### DIFF
--- a/nomos-services/mempool/src/tx/service.rs
+++ b/nomos-services/mempool/src/tx/service.rs
@@ -181,6 +181,10 @@ where
     NetworkAdapter::Settings: Clone + Send + 'static,
     RecoveryBackend: RecoveryBackendTrait,
 {
+    #[expect(
+        clippy::cognitive_complexity,
+        reason = "Mempool message handling is convenient to have in one block"
+    )]
     fn handle_mempool_message(
         &mut self,
         message: MempoolMsg<Pool::BlockId, Pool::Item, Pool::Item, Pool::Key>,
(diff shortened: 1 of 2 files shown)


ntn-x2 (approved):
Oh, nice! I guess I missed it in the CL mempool work and propagated it in the DA mempool as well. Thanks for this! As for the style, I'd like the two mempool impls to have the same structure since DA mempool has a single additional line, while keeping the rest the same. I personally prefer not having a separate function for it, but I would also be ok with splitting the `add` message handling into its own function, as long as the style between the two service impls is the same.

nomos-services/mempool/src/tx/service.rs
```diff
@@ -181,6 +181,10 @@ where
     NetworkAdapter::Settings: Clone + Send + 'static,
     RecoveryBackend: RecoveryBackendTrait,
 {
+    #[expect(
+        clippy::cognitive_complexity,
+        reason = "Mempool message handling is convenient to have in one block"
+    )]
```
bacv, line 187: I prefer having mempool message handling in one block in this particular case.

PR #1138, opened 2025-03-14
feat(DA): request shares HTTP API

andrussal:

**Implement HTTP API to get shares(blobs) according to [spec](https://www.notion.so/NomosDA-Reconstruction-1828f96fb65c80b2bbb9f4c5a0cf26a5)** 

As discussed with Daniel, we can try to make it streaming if possible(I think because ideally we don’t need to load all requested blobs into memory at once). At the moment only “frontend” is doing streaming whereas storage backend still returns collection. Should we change it as well...? It seems a bit similar discussion like we had with batch inserts, what we expect our storage backend to support.

For streaming, content type is `application/x-ndjson` but maybe this needs further discussion what we want our API to support.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 1. Description added.


nomos-services/api/src/http/storage.rs
```diff
@@ -96,3 +109,67 @@ where
...
+        + Sync
+        + 'static,
+    StorageOp::Error: Send + Sync,
+{
+    let storage_relay = handle
+        .relay::<StorageService<RocksBackend<StorageOp>>>()
+        .connect()
+        .await?;
+
+    let shares_prefix = key_bytes(DA_BLOB_PREFIX, blob_id.as_ref());
```
danielSanchezQ, line 138: Wait, how are we storing shares? We should be able to ask for a `blob_id` + `share_idx`.

nomos-services/api/src/http/storage.rs
```diff
@@ -96,3 +109,67 @@ where
...
+                        return None;
+                    };
+                    json.push(b'\n');
+                    Some(Ok(Bytes::from(json)))
+                } else {
+                    None
+                }
+            }
+        }))
+}
```
danielSanchezQ, line 175: Does storage has the ability to pick a single share?
What I meant is that we could probably request one by one in an unfold for what we need.

andrussal: We could do it if `return_available` is set to `false`. Then we can only ask `requested_shares`. Otherwise we still need to ask everything because we don't know which shares node has(I think...).

danielSanchezQ: There is something wrong then. Because **we should**.

andrussal: I am thinking if we need a separate extra storage in our key/value store that keeps track of which shares we have per blob? Because at the moment a share is stored under a key: `blob_id+column_idx`. We don't know which such `blob_id+column_idx` exist.

danielSanchezQ: > I am thinking if we need a separate extra storage in our key/value store that keeps track of which shares we have per blob? Because at the moment a share is stored under a key: `blob_id+column_idx`. We don't know which such `blob_id+column_idx` pairs exist.

We do not need a separate extra storage for now. We just need to build the key for storing counting with the `share_idx` probably. I'll review this and come with a solution plan.

danielSanchezQ: > I am thinking if we need a separate extra storage in our key/value store that keeps track of which shares we have per blob? Because at the moment a share is stored under a key: `blob_id+column_idx`. We don't know which such `blob_id+column_idx` pairs exist.

Ok, I understood something else previously, But I checked better and yes, we do probably have to add an stored value to keep track which values do we actually have. Let make a different PR adding that indexing entry. Then we can continue with this one once that is merged. Wdyt @andrussal ?

andrussal: Sounds good, I’ll do it.

nomos-services/api/src/http/da_shares.rs
```diff
@@ -0,0 +1,142 @@
...
+    let blob_shares =
+        load_blob_shares_index::<StorageOp, DaShare>(&storage_relay, &blob_id).await?;
+
+    let filtered_shares: Vec<_> = blob_shares
+        .into_iter()
+        .filter(|idx| {
+            // If requested_shares contains the index, then ignore the filter_shares
+            requested_shares.contains(idx) || (return_available && !filter_shares.contains(idx))
+        })
+        .collect();
```
bacv, line 55:
```suggestion
    let filtered_shares = blob_shares.into_iter().filter(move |idx| {
        // If requested_shares contains the index, then ignore the filter_shares
        requested_shares.contains(idx) || (return_available && !filter_shares.contains(idx))
    });
```

Since we're streaming the iterator later on, I think `collect` is not required here as `Filter` is also an `Iterator`

andrussal: Thanks! Removed unnecessary `collect`

bacv (approved):

nomos-services/api/src/http/da_shares.rs
```diff
@@ -0,0 +1,139 @@
...
+    });
+
+    let blob_id = Arc::new(blob_id);
+    let stream = stream::iter(filtered_shares).then(move |share_idx| {
+        let storage_relay = storage_relay.clone();
+        let blob_id = Arc::clone(&blob_id);
+        load_and_process_share::<StorageOp, DaShare>(storage_relay, blob_id, share_idx)
+    });
+
+    Ok(stream)
```
danielSanchezQ, line 61: Nice, so we get the full streaming. Awesome @andrussal !!

danielSanchezQ (approved):
:rocket:

PR #1143, opened 2025-03-14
fix(DA): Sampling behaviour lost requests

bacv:

It was observed that even in the perfect network conditions some sampling requests wouldn't be sent. There were couple of cases that were not handled as sampling behavior doesn't necessary start the stream between peers:
- When there is no connection between peers `open_stream` returns `ConnectionReset` io error, sample request fails.
- When a stream is being reused, receiving end already has stopped listening to the stream after the first message - sample request fails.
- UnexpectedEof happens when peer closes the stream and in our case this is expected and shouldn't be propagated as an error.

Since dispersal service now has an option to try sampling before publishing into the mempool, a small change was made in sampling service to ensure that sampling service invalidates only blobs that it was requested to sample.

One other change might not be very obvious, but receiving end now tries to listen to the incoming stream as long as it gets the close Eof signal from remote peer. Outgoing and Incoming tasks are put into the same tasks queue.

Yes

@bacv

N/A

N/A

## Checklist

* [x] 1. Description added.

### DIFF
--- a/nomos-da/network/core/src/protocols/sampling/behaviour.rs
+++ b/nomos-da/network/core/src/protocols/sampling/behaviour.rs
@@ -253,12 +253,12 @@ struct ResponseChannel {
     response_receiver: Receiver<BehaviourSampleRes>,
 }
 
-enum SampleStreamSuccess {
+enum SampleStreamResponse {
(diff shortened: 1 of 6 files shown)


pradovic (approved):
Looks great

PR #1146, opened 2025-03-17
chore(api): add timeout to http calls

### LINKED ISSUE #1137: Http requests wait for recevier channels forever
In the HTTP calls that are fetching responses through receiver channels, there is only a `receiver.await` idiom used which waits for the channel forever. It would be useful to add timeouts and return internal server error if the backend does not respond after sensible timeout.

pradovic:

Add 10s timeout to http calls. Resolves #1137.

Yes.

@pradovic 

Yes

HTTP requests should return timeout error (internal server error) if they they don't receive response from the response channel in sensible time.

## Checklist

> [!WARNING]  

* [x] 1. Description added.

### DIFF
--- a/nomos-services/api/src/lib.rs
+++ b/nomos-services/api/src/lib.rs
@@ -1,4 +1,4 @@
-use std::{future::Future, time::Duration};
+use std::{future::Future, sync::OnceLock, time::Duration};
 
 use overwatch::{
     overwatch::handle::OverwatchHandle,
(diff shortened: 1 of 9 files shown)


nomos-services/api/src/lib.rs
```diff
@@ -10,6 +12,8 @@ use overwatch::{
 
 pub mod http;
 
+pub(crate) const HTTP_REQUEST_TIMEOUT: Duration = Duration::from_secs(10);
```
bacv, line 15: I think this would be better to load from configuration, at the moment we don't know how much some longrunning operations could take, it will be a bit easier to test it with different timeout settings.

AlejandroCabeza (approved):
other than the change mentioned by gusto, lgtm :)

nomos-services/api/src/lib.rs
pradovic: Should be fixed now. I am not sure if there was a better way, but I tried to keep everything in the ApiService.

bacv (approved):

PR #1154, opened 2025-03-18
feat(network): add kademlia peer discovery

pradovic:

Add Kademlia peer discovery and routing in base libp2p network.

Yes.

@danielSanchezQ , @pradovic 

Yes.

Preliminary investigation: https://www.notion.so/Preliminary-Research-Network-peer-discovery-membership-1bc8f96fb65c800492acfecb9a6f92dc.

It adds kademila peer discovery.

## Checklist

> [!WARNING]  

* [x] 1. Description added.


nomos-libp2p/src/lib.rs
```diff
@@ -51,14 +60,35 @@ impl Behaviour {
...
                 .build()?,
         )?;
-        Ok(Self { gossipsub })
+
+        let store = kad::store::MemoryStore::new(peer_id);
+        let kad_config = kad::Config::new(StreamProtocol::new("/nomos/kad/1.0.0"));
+        let kademlia = kad::Behaviour::with_config(peer_id, store, kad_config);
+
+        let identify = identify::Behaviour::new(identify::Config::new(
+            "/nomos/id/1.0.0".into(), // Identify protocol name
```
danielSanchezQ, line 69: This should be a `const` item defined outside.

nomos-libp2p/src/lib.rs
```diff
@@ -51,14 +60,35 @@ impl Behaviour {
                 .max_transmit_size(DATA_LIMIT)
                 .build()?,
         )?;
-        Ok(Self { gossipsub })
+
+        let store = kad::store::MemoryStore::new(peer_id);
+        let kad_config = kad::Config::new(StreamProtocol::new("/nomos/kad/1.0.0"));
```
danielSanchezQ, line 65: This too, hardcoded procotols should be `const` items.

nomos-services/network/src/backends/libp2p/command.rs
```diff
@@ -21,6 +21,29 @@ pub enum Command {
         message: Box<[u8]>,
         retry_count: usize,
     },
+
+    // Kademlia commands
+    AddPeer {
+        peer_id: PeerId,
+        addr: Multiaddr,
+        reply: oneshot::Sender<Result<(), String>>,
+    },
```
danielSanchezQ, line 30: Nice, we would have to call this when loading federated trusted nodes from config probably. So we can bootstrap the network.

nomos-services/network/src/backends/libp2p/command.rs
```diff
@@ -21,6 +21,29 @@ pub enum Command {
...
+    PutValue {
+        key: Vec<u8>,
+        value: Vec<u8>,
+        reply: oneshot::Sender<Result<kad::RecordKey, String>>,
+    },
+    GetValue {
+        key: Vec<u8>,
+        reply: oneshot::Sender<Result<Vec<u8>, String>>,
+    },
```
danielSanchezQ, line 43: We have no use for the DHT other than peer discovery for now. @youngjoon-lee could we leverage this for the sync protocol?

nomos-services/network/src/backends/libp2p/command.rs
```diff
@@ -21,6 +21,29 @@ pub enum Command {
...
+    Bootstrap {
+        reply: oneshot::Sender<Result<(), String>>,
+    },
```
danielSanchezQ, line 46: What does this do?

nomos-services/network/src/backends/libp2p/command.rs
pradovic: I have put the stubs just as ideas, so pls feel free to put your 2 c whoever thinks that some command could be useful. But also, new commands should be trivial to add, and I will reduce those to only what we use for now

nomos-services/network/src/backends/libp2p/command.rs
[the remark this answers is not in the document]
pradovic: This PR is a work in progress, and I just added some random stabs for me to test. I will polish the PR and then ping you guys to discuss. The changes here dont reflect my local dev at the moment. Sorry about this, I thought nobody looks at draft PRs :)

pradovic:
This one is not ready for discussion yet. I will reopen it when it is, not to waste other peoples time :)

nomos-libp2p/src/lib.rs
```diff
@@ -51,14 +62,57 @@ impl Behaviour {
                 .max_transmit_size(DATA_LIMIT)
                 .build()?,
         )?;
-        Ok(Self { gossipsub })
+
+        let store = kad::store::MemoryStore::new(peer_id);
+        let mut kad_config = kad::Config::new(StreamProtocol::new(KAD_PROTOCOL_NAME));
+        kad_config.set_periodic_bootstrap_interval(Some(Duration::from_secs(10))); // this interval is only for proof of concept and should be configurable
+        let kademlia = kad::Behaviour::with_config(peer_id, store, kad_config);
+
```
pradovic, line 70: All these we can make configurable. Even the whole kademlia can be optional, i order to cancel it if needed for testing, etc...

pradovic:
> This one is not ready for discussion yet. I will reopen it when it is, not to waste other peoples time outdated stuff :)

Ready for evaluation as proof of concepts. Stylistically not ready for complete review yet.

nomos-libp2p/src/lib.rs
```diff
@@ -51,14 +66,58 @@ impl Behaviour {
                 .max_transmit_size(DATA_LIMIT)
                 .build()?,
         )?;
-        Ok(Self { gossipsub })
+
+        // make config parameters configurable
+        let store = kad::store::MemoryStore::new(peer_id);
```
bacv, line 71: Would be great to implement our own storage to add recovery later on.

nomos-libp2p/src/lib.rs
bacv: At the moment it doesn't seem straightforward to make kademlia optional, maybe you have an idea in mind?

nomos-services/network/src/backends/libp2p/swarm.rs
```diff
@@ -92,6 +131,85 @@ impl SwarmHandler {
...
+                    // we need to add the peer to the kademlia routing table
+                    // in order to enable peer discovery
+                    for addr in &info.listen_addrs {
+                        behaviour.kademlia_add_address(peer_id, addr.clone());
+                    }
+                }
+            }
+
+            SwarmEvent::Behaviour(BehaviourEvent::Kademlia(
+                kad::Event::OutboundQueryProgressed {
```
bacv, line 158: Now with more events, it could be handled per behaviour in one place, similar like it's done here: https://github.com/logos-co/nomos/blob/master/nomos-da/network/core/src/swarm/validator.rs#L265-L267

bacv (commented):
This is great proof of concept! For the first version in master it would be great to have Kademlia optional.

nomos-libp2p/src/lib.rs
pradovic: I think that derive network behavior works for `Option<T>` params, but I am not 100% sure. I will confirm and try to find the new way if not

pradovic: Ah ok, just tried it, maybe not, but I guess a I can always wrap behavior and delegate only the methods for the Some fields. There must be a way :)

bacv: Nice, that should work and could be toggled at runtime! So Identify and Kademlia would both be optional?

pradovic: We should make both optional. We also need to see if we need mdns maybe instead of identify, but I guess identify makes more sense

pradovic: Relevant doc: https://docs.rs/libp2p-kad/latest/libp2p_kad/. I have see  a few impla with mdns instead

danielSanchezQ: they could be composed. For example we could (and I think we should because they serve the same purpose) wrap kademlia+identify in its own behaviour (lets call it `NomosDiscovery`, this can later be expanded for adding other discovery protocols for example) then rewrap with toggle.

nomos-services/network/src/backends/libp2p/command.rs
```diff
@@ -21,6 +21,16 @@ pub enum Command {
...
+
+    // Kademlia commands
+    GetClosestPeers {
+        peer_id: PeerId,
+        reply: oneshot::Sender<Vec<PeerInfo>>,
+    },
+
+    DumpRoutingTable {
+        reply: oneshot::Sender<Vec<PeerId>>,
+    },
```
danielSanchezQ, line 33: I think at this point we should start delegating commands. What I mean is to wrap them into bigger sets:
```rust
pub enum PubSubCommand {
    Broadcast {},
    Subscribre {},
    Unsubscribe {},
    ....
}

pub enum DiscoveryCommand {
    GetClosestPeers {},
    DumpRoutingTable {}
}

pub enum Command {
    PubSub(PubSubCommand),
    Discovery(DiscoveryCommand)
}

```

nomos-libp2p/src/lib.rs
danielSanchezQ: This is a great idea, can we abstract it and still use the `MemoryStore` ourselves so its easier to move later on?

nomos-services/network/src/backends/libp2p/command.rs
pradovic: Yes, this plus handle functions can remove the clippy suppress warnings as well.

nomos-services/network/src/backends/libp2p/command.rs
youngjoon-lee: > We have no use for the DHT other than peer discovery for now. @youngjoon-lee could we leverage this for the sync protocol?

yeah, i guess that's a good approach to try (as we discussed today). I'll give a try and investigate some other approaches.

youngjoon-lee:
A great PoC! (since I've missed the discussions that you guys had, I just reviewed general ideas that this change wants to achieve)

pradovic:
Closing to perform a preliminary investigation to verify if we need kademlia or another approach.

nomos-libp2p/src/lib.rs
pradovic: I am not 100% sure, but I think that identify might be able to help with NAT as well, so we could, in theory, use it without kademlia. It is useful for getting observed address and other addr information upon connection.

bacv: Agree, let's use it.

pradovic:
@danielSanchezQ @CHr15F0x @bacv I think I have addressed all the comments, hopefully. There is a few todos left for another PR or when we need such functionality (like persistent store).

It might be interesting to expose HTTP API for discovery so we can use it in unit tests. Just an idea, probably not a priority.

bacv (approved):
Looks great!

nomos-libp2p/src/lib.rs
```diff
@@ -36,13 +39,27 @@ pub struct Swarm {
...
 
 impl Behaviour {
-    fn new(peer_id: PeerId, gossipsub_config: gossipsub::Config) -> Result<Self, Box<dyn Error>> {
+    fn new(
+        peer_id: PeerId,
+        gossipsub_config: gossipsub::Config,
+        kad_config: Option<KademliaSettings>,
+        identify_config: Option<IdentifySettings>,
+        public_key: identity::PublicKey,
+    ) -> Result<Self, Box<dyn Error>> {
```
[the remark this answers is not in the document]
danielSanchezQ: Yeah, good catch.

nomos-libp2p/src/lib.rs
```diff
@@ -79,11 +182,34 @@ impl Swarm {
             .with_tokio()
             .with_quic()
             .with_dns()?
-            .with_behaviour(|_| Behaviour::new(peer_id, config.gossipsub_config.clone()).unwrap())?
+            .with_behaviour(|keypair| {
+                Behaviour::new(
+                    peer_id,
+                    config.gossipsub_config.clone(),
+                    config.kademlia_config.clone(),
+                    config.identify_config.clone(),
```
danielSanchezQ, line 190: Nitpik: probably no need to clone. Destructure the config and just pass the configs so ownership is splitted.

nomos-libp2p/src/lib.rs
```diff
@@ -79,11 +182,34 @@ impl Swarm {
...
+        if let Some(kademlia_config) = &config.kademlia_config {
+            if !kademlia_config.client_mode {
+                // libp2p2-kad server mode is implicitly enabled
+                // by adding external addressess
+                // <https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/CHANGELOG.md#0440>
+                let external_addr = listen_addr.with(Protocol::P2p(peer_id));
+                swarm.add_external_address(external_addr.clone());
+                tracing::info!("Added external address: {}", external_addr);
+            }
+        }
```
danielSanchezQ, line 212: Uhm. I do not think we have to expose this. It should always be a client and a server. All nodes should participate the same. Does it makes sense?

nomos-services/network/src/backends/libp2p/swarm.rs
```diff
@@ -52,14 +60,44 @@ impl SwarmHandler {
...
+                    tracing::debug!("Added peer to Kademlia: {} at {}", peer_id, peer_addr);
+                } else {
+                    tracing::warn!("Failed to parse peer ID from multiaddr: {}", peer_addr);
+                }
+            } else {
+                tracing::warn!("Multiaddr doesn't contain peer ID: {}", peer_addr);
+            }
+
+            self.swarm.swarm_mut().behaviour_mut().bootstrap_kademlia();
+        }
```
danielSanchezQ, line 95: Maybe lets extract this to a proper function `bootstrap` (?)

nomos-libp2p/src/config.rs
```diff
@@ -1,8 +1,15 @@
...
-use libp2p::{gossipsub, identity::ed25519};
+use libp2p::{
+    gossipsub, identify,
+    identity::{self, ed25519},
+    kad, StreamProtocol,
+};
 use serde::{Deserialize, Serialize};
 
+pub const DEFAULT_KAD_PROTOCOL_NAME: &str = "/test/nomos/kad/1.0.0";
+pub const DEFAULT_IDENTIFY_PROTOCOL_NAME: &str = "/test/nomos/identify/1.0.0";
```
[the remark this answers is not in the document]
bacv: Good catch, I see that the protocol name already has `test` prefix, but we want to have option to select main/test from the config probably.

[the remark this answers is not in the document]
danielSanchezQ: > Good catch, I see that the protocol name already has `test` prefix, but we want to have option to select main/test from the config probably.

Im not sure about the config stuff. In config maybe we should just take a `--test` flag for injecting testing environment capabilities (?)

nomos-libp2p/src/lib.rs
pradovic: This was mentioned in the PoC phase, I believe and there was a todo comment to make it configurable. Basically client mode in kademlia means that node participate networks but is not providing query capabilities, so basically kind of a light node that queries other nodes but does not improve the network by itself. These are mentioned as good candidates for unrealiable devices. 

I am not saying we need this just wanted to clarify.
The default is a server mode, this code just adds an option to be in client_mode as well.

nomos-libp2p/src/config.rs
```diff
@@ -15,6 +22,12 @@ pub struct SwarmConfig {
     // Gossipsub config
     #[serde(with = "GossipsubConfigDef", default = "gossipsub::Config::default")]
     pub gossipsub_config: gossipsub::Config,
+
+    #[serde(default)]
+    pub kademlia_config: Option<KademliaSettings>,
+
+    #[serde(default)]
+    pub identify_config: Option<IdentifySettings>,
```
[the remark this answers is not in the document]
pradovic: I wanted to make them optional. If they are not present that identify and/or kademlia are not enabled. Does it makes sense? Is there a better way to do this maybe?

nomos-libp2p/src/lib.rs
danielSanchezQ: > This was mentioned in the PoC phase, I believe and there was a todo comment to make it configurable. Basically client mode in kademlia means that node participate networks but is not providing query capabilities, so basically kind of a light node that queries other nodes but does not improve the network by itself. These are mentioned as good candidates for unrealiable devices.
> 
> I am not saying we need this just wanted to clarify. The default is a server mode, this code just adds an option to be in client_mode as well.

Ah, yes, I recall now about this. You are correct, thanks for the explanation.

nomos-libp2p/src/config.rs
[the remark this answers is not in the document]
pradovic: I would prefer to make a test default, as we would not like people to run node with an open network by accident (omitting to put the test flag for example). Maybe put a production field in the config? With false being default?

@danielSanchezQ Do you not expect to need to change the protocol version, for example, ever? Sorry for nitpicks, I am just trying to figure out what is the downside of protocol names in the configs?

nomos-libp2p/src/config.rs
```diff
@@ -173,6 +188,273 @@ pub mod secret_key_serde {
     }
 }
 
+fn default_kad_protocol() -> String {
+    DEFAULT_KAD_PROTOCOL_NAME.to_owned()
+}
```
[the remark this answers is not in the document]
pradovic: This was for the serde default as it accepts a fn. I tried to find a way to use it directly, is it possible?

nomos-libp2p/src/config.rs
```diff
@@ -173,6 +188,273 @@ pub mod secret_key_serde {
...
+    }
+}
+
+impl KademliaSettings {
+    #[must_use]
+    pub fn to_libp2p_config(&self) -> kad::Config {
+        // Convert the protocol name to a static string by leaking the memory
+        // This is safe because the protocol name is a static string and will not be
+        // deallocated until the program exits
+        let protocol_name = Box::leak(self.protocol_name.clone().into_boxed_str());
```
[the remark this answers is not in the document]
pradovic: Stream protocol takes `&'static str`. I could not find a better way to do this. Any ideas maybe?

nomos-libp2p/src/config.rs
```diff
@@ -173,6 +188,273 @@ pub mod secret_key_serde {
...
+
+    #[must_use]
+    pub fn get_protocol_name(&self) -> String {
+        self.protocol_name.clone()
+    }
+}
+
+fn default_identify_protocol() -> String {
+    DEFAULT_IDENTIFY_PROTOCOL_NAME.to_owned()
+}
```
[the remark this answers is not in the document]
pradovic: I will wait for the anwer to `DEFAULT_KADEMLIA_PROTOCOL_NAME' discussion and then apply here too.

nomos-libp2p/src/lib.rs
pradovic: Should be fixed now. Needed to change parent configs to be moved instead of borrowed.

nomos-libp2p/src/lib.rs
pradovic: Should be fixed now.

nomos-libp2p/src/config.rs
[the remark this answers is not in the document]
danielSanchezQ: > I would prefer to make a test default, as we would not like people to run node with an open network by accident (omitting to put the test flag for example). Maybe put a production field in the config? With false being default?
> 
> @danielSanchezQ Do you not expect to need to change the protocol version, for example, ever? Sorry for nitpicks, I am just trying to figure out what is the downside of protocol names in the configs?

A protocol upgrade should not come from configuration. In fact it is another weak point to add to the least. This should be hardcoded and just swapped for testnet environment imo.

nomos-libp2p/src/config.rs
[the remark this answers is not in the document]
danielSanchezQ: I think it should be possible to set to a constante but not sure. Anyway, I really think we should come up with a different strategy for this as it should not come from config.

nomos-libp2p/src/config.rs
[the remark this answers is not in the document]
danielSanchezQ: > Stream protocol takes `&'static str`. I could not find a better way to do this. Any ideas maybe?

Exactly, another hint that this should not be really configurable 😉

nomos-libp2p/src/config.rs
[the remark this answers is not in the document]
pradovic: Makes sense, I understand what you mean now, my bad. I will hardcode both identify and kad protocol name, with the test variants.

Maybe we should have one for mainnet, one for testnet and one for unit tests then. We would not like for them to dicover each other.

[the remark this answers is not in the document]
danielSanchezQ: Yeah, those 3 seems what we should have. They can be in a different module btw, so they are not dispersed.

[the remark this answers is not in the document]
pradovic: Should be better now. I have extracted a consts and made procotol name config with 4 values: mainnet, testnet, unittest, integrationtest. The default (if nothing is provided) is unittest.

nomos-libp2p/src/config.rs
```diff
@@ -173,6 +207,220 @@ pub mod secret_key_serde {
...
+
+        // Apply only the settings that were specified
+        if let Some(timeout) = self.query_timeout_secs {
+            config.set_query_timeout(Duration::from_secs(timeout));
+        }
+
+        if let Some(factor) = self.replication_factor {
+            if let Some(factor) = NonZeroUsize::new(factor) {
+                config.set_replication_factor(factor);
+            }
```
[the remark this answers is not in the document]
danielSanchezQ: Actually, the `NonZero` could be used in the settings directly so it breaks when deserializing instead. 🤔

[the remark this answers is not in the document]
pradovic: Agreed then!

nomos-services/network/src/backends/libp2p/swarm.rs
```diff
@@ -264,7 +341,280 @@ impl SwarmHandler {
...
+            }
+            event => {
+                tracing::debug!("Identify event: {:?}", event);
+            }
+        }
+    }
+
+    pub fn handle_kademlia_event(&mut self, event: kad::Event) {
+        match event {
+            kad::Event::OutboundQueryProgressed { id, result, .. } => {
```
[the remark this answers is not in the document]
pradovic: Hey very good observation, I completely forgot to handle it. Thanks!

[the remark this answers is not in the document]
pradovic: Should be handled better now.

nomos-libp2p/src/config.rs
```diff
@@ -173,6 +207,228 @@ pub mod secret_key_serde {
...
+    F: FnOnce(NonZeroUsize),
+{
+    if let Some(val) = value {
+        if let Some(non_zero) = NonZeroUsize::new(val) {
+            f(non_zero);
+        } else {
+            tracing::warn!("{} value is 0, setting to default", field_name);
+        }
+    }
+}
```
danielSanchezQ, line 361: This is what I sad can be replace by:`Option<NonZero<usize>>` during serialization. Forcing the user to set something valid or nothing for default.

pradovic: https://github.com/logos-co/nomos/pull/1154/commits/6ceb44ec3bca2380de5a87992946ac61c9bf072d. That seems much better, thanks!

danielSanchezQ (approved):

Issue #1114, opened 2025-03-10
Use an LRU cache for seen-messages in DA network behaviour


CHr15F0x:
... to avoid the cache ballooning excessively.

In the first implementation: [here](https://github.com/logos-co/nomos-node/blob/b7237ba414f4df308a06cf2bf2b7a8c3ca9dec27/nomos-da/network/core/src/protocols/replication/behaviour.rs#L104)
In the refactored implementation: [here](https://github.com/logos-co/nomos-node/blob/566520702517e3f8eb8f800bb1437bc84245aa81/nomos-da/network/core/src/protocols/replication/behaviour_v2.rs#L199)

PR #1155, opened 2025-03-19
Use an LRU cache for seen-messages in DA network behaviour

CHr15F0x:

https://github.com/logos-co/nomos/issues/1114

Yes.

@CHr15F0x 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 1. Description added.

### DIFF
--- a/nomos-da/network/core/Cargo.toml
+++ b/nomos-da/network/core/Cargo.toml
@@ -16,10 +16,8 @@ libp2p-stream            = "0.3.0-alpha"
 log                      = "0.4"
 nomos-core               = { workspace = true }
 nomos-da-messages        = { workspace = true }
-nomos-utils              = { workspace = true, features = ["time"], optional = true }
 rand                     = "0.8"
 serde                    = "1.0"
-serde_with               = { version = "3.12", optional = true }
 subnetworks-assignations = { workspace = true }
 thiserror                = "1.0"
 tokio                    = { version = "1" }
@@ -36,9 +34,3 @@ libp2p-swarm-test  = { version = "0.5.0", features = ["tokio"] }
 rstest             = "0.25"
 tokio              = { version = "1", features = ["macros", "rt-multi-thread", "time"] }
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
-
-[features]
-time = ["dep:nomos-utils", "dep:serde_with"]
-
-[package.metadata.cargo-machete]
-ignored = ["nomos-utils"]
--- a/nomos-da/network/core/src/protocols/replication/behaviour.rs
+++ b/nomos-da/network/core/src/protocols/replication/behaviour.rs
@@ -180,11 +180,9 @@ impl PendingOutbound {
     }
 }
 
-#[cfg_attr(feature = "time", serde_with::serde_as)]
 #[derive(Clone, Copy, Debug, Deserialize, Serialize)]
 pub struct ReplicationConfig {
     pub seen_message_cache_size: usize,
-    #[cfg_attr(feature = "time", serde_as(as = "MinimalBoundedDuration<1, MINUTE>"))]
     pub seen_message_ttl: Duration,
 }
 
--- a/testnet/cfgsync/src/server.rs
+++ b/testnet/cfgsync/src/server.rs
@@ -41,10 +41,19 @@ pub struct CfgSyncConfig {
     pub monitor_failure_time_window_secs: u64,
     pub balancer_interval_secs: u64,
     pub mempool_publish_strategy: MempoolPublishStrategy,
-    pub replication_settings: ReplicationConfig,
+    pub replication_settings: HumanReadableReplicationConfig,
 
     // Tracing params
     pub tracing_settings: TracingSettings,
+}
+
+// TODO remove this workaround once [`MinimumBoundedDuration`]'s serialized
+// representation is human-readable and deals with [`std::time::Duration`]` or
+// the codebase switches to [`time::Duration`]
+#[derive(Clone, Copy, Debug, Default, Deserialize, Serialize)]
+pub struct HumanReadableReplicationConfig {
+    pub seen_message_cache_size: usize,
+    pub seen_message_ttl_secs: u64,
 }
 
 impl CfgSyncConfig {
@@ -89,7 +98,12 @@ impl CfgSyncConfig {
             },
             balancer_interval: Duration::from_secs(self.balancer_interval_secs),
             redial_cooldown: Duration::ZERO,
-            replication_settings: self.replication_settings,
+            replication_settings: ReplicationConfig {
+                seen_message_cache_size: self.replication_settings.seen_message_cache_size,
+                seen_message_ttl: Duration::from_secs(
+                    self.replication_settings.seen_message_ttl_secs,
+                ),
+            },
         }
     }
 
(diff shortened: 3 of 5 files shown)


nomos-blend/network/Cargo.toml
```diff
@@ -5,7 +5,7 @@ name    = "nomos-blend-network"
 version = "0.1.0"
 
 [dependencies]
-cached              = "0.53.1"
+cached              = { version = "0.55.1", default-features = false }
 futures             = "0.3.30"
```
danielSanchezQ, line 9: Let's add those to the workspace instead.

nomos-da/network/core/Cargo.toml
```diff
@@ -5,6 +5,7 @@ name    = "nomos-da-network-core"
 version = "0.1.0"
 
 [dependencies]
+cached                   = { version = "0.55.1", default-features = false }
```
danielSanchezQ, line 8: Likewise

danielSanchezQ: In fact, @ntn-x2 , what is the heuristic to choose workspace or local deps now? :thinking:

nomos-da/network/core/src/protocols/replication/behaviour.rs
```diff
@@ -213,16 +221,22 @@ pub struct ReplicationBehaviour<Membership> {
...
-    /// Seen messages cache holds a record of seen messages
-    seen_message_cache: IndexSet<(Vec<u8>, SubnetworkId)>,
+    /// The cache for seen messages, evicts entries in two ways:
+    /// - oldest entries when full (Least-Recently-Used),
+    /// - after a certain TTL has passed since the entry was last accessed,
+    ///   regardless of cache utilization.
+    /// The cache allocates memory once up front and any expired and unused
+    /// entries will eventually be overwritten and erasing them via
+    /// [`TimeSizedCache::flush`] does not save space.
+    seen_message_cache: TimedSizedCache<(Vec<u8>, SubnetworkId), ()>,
```
danielSanchezQ, line 231: What is `Vec<u8>` in this context? the bytes of the message before deserialization?
Wouldn't it be more clear to use the `blob_id` in the same fashion as used in storage?

nomos-da/network/core/src/protocols/replication/behaviour.rs
```diff
@@ -177,6 +179,12 @@ impl PendingOutbound {
     }
 }
 
+#[derive(Debug)]
+pub struct ReplicationConfig {
+    pub seen_message_cache_size: usize,
+    pub seen_message_ttl: Duration,
```
danielSanchezQ, line 185: If this is gonna have a minimum duration it would probably be good to use the utils type for this.

danielSanchezQ: Also you may need `Deserialize` for this? How is it loaded otherwise?

nomos-da/network/core/src/protocols/replication/behaviour.rs
```diff
@@ -289,10 +310,10 @@ where
     /// sent again.
     fn send_message_impl(&mut self, waker: Option<&Waker>, message: &DaMessage) {
         let message_id = (message.share.blob_id.to_vec(), message.subnetwork_id);
```
danielSanchezQ, line 312: Oh, so it is actually the blob_id+subnetwork. Makes sense. But don't we have a `BlobId` type somewhere to be used in the declaration above?

nomos-da/network/core/src/swarm/common/mod.rs
```diff
@@ -1,4 +1,14 @@
...
 pub mod balancer;
 pub mod handlers;
 pub mod monitor;
 pub mod policy;
+
+#[derive(Copy, Clone, Debug, Default, Serialize, Deserialize)]
+pub struct DAReplicationSettings {
+    pub seen_message_cache_size: usize,
+    pub seen_message_ttl: Duration,
+}
```
danielSanchezQ, line 14: I don't get why we have two of these for the same attributes.

nomos-da/network/core/src/swarm/validator.rs
```diff
@@ -102,7 +104,17 @@ where
...
-                swarm: Self::build_swarm(key, membership, balancer, monitor, redial_cooldown),
+                swarm: Self::build_swarm(
+                    key,
+                    membership,
+                    balancer,
+                    monitor,
+                    redial_cooldown,
+                    ReplicationConfig {
+                        seen_message_cache_size: replication_settings.seen_message_cache_size,
+                        seen_message_ttl: replication_settings.seen_message_ttl,
```
danielSanchezQ, line 115: If we really need 2 of those. Better to implement `From/Into` or a custom method `into_replication_config`. But I think there should be just one probably.

bacv: Another solution would be to to make Replication generic over `Cached` trait from the `cached` crate. Then we'd just pass the cache constructed from the swarm.

nomos-da/network/core/src/protocols/replication/behaviour.rs
bacv, line 313: I think `ReplicationRequest` could have `id()` method that returns `ReplicationMessageId` type of `[u8;34]` that the cache would expect as it's key.

testnet/cfgsync/src/server.rs
```diff
@@ -86,6 +88,10 @@ impl CfgSyncConfig {
             },
             balancer_interval: Duration::from_secs(self.balancer_interval_secs),
             redial_cooldown: Duration::ZERO,
+            replication_settings: DAReplicationSettings {
+                seen_message_cache_size: 2048 * 100,
+                seen_message_ttl: Duration::from_secs(30 * 60),
+            },
```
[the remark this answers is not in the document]
bacv: > This is some wild guess, I've no idea what should go here.

Would be great to expect that from the `cfgsync.yml`, then we could experiment with the values.

nomos-da/network/core/src/swarm/common/mod.rs
bacv: Since this is used by replication behavior, seems better to define in behavior and then reexport in swarm common. The idea would be similar to `ReplicationEvent`, we have it defined in the behavior too. Thats how policy and monitor settings are organized also.

nomos-da/network/core/src/swarm/validator.rs
bacv: I prefer single struct config at behavior level, in replication we'll most likely use one implementation for cache, so my suggestion is redundant.

danielSanchezQ: > I prefer single struct config at behavior level, in replication we'll most likely use one implementation for cache, so my suggestion is redundant.

:heavy_plus_sign: **1**, I think single struct config is best.

bacv (approved):
Looks good!

testnet/cfgsync/src/server.rs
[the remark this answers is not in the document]
danielSanchezQ: > MinimalBoundedDuration will have to wait: [d0e4885](https://github.com/logos-co/nomos/commit/d0e48852c6d5c54e09f12833d8378474f8edbe0f)

MinimalBoundedDuration is for deserialization only. Why can it not be used?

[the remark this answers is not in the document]
danielSanchezQ: > I'm inclined to actually fix this in a follow up PR once the direction in #1156 is settled - wdyt?

Ok, lets do it like that.

bacv:
good to go 👍

PR #1179, opened 2025-03-20
feature: Add bounded duration compatibility for core::time::Duration

danielSanchezQ:

Added bounded duration compatibility with `core::time::Duration`. Notice that when using `MinimalBoundedTime` duration serialization defaults to human readable (basically delegates into `time::Duration` serialization). This change may make some current config fail. But on the other hand, it would be easier to handle them. cc: @bacv.

Just added trait implementation so we can deserialize into `core::time::Duration`.

N/A
N/A
N/A

## Checklist

* [x] 1. Description added.

### DIFF
--- a/nomos-utils/src/bounded_duration.rs
+++ b/nomos-utils/src/bounded_duration.rs
@@ -1,6 +1,8 @@
 use std::{fmt::Display, marker::PhantomData};
 
-use serde::{de::Error, Deserialize, Deserializer, Serialize, Serializer};
+use serde::{
+    de::Error as DeError, ser::Error as SerError, Deserialize, Deserializer, Serialize, Serializer,
+};
 use serde_with::{DeserializeAs, SerializeAs};
 use time::Duration;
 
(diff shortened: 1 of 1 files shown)


nomos-utils/src/bounded_duration.rs
```diff
@@ -95,6 +97,40 @@ impl<'de, const MIN_DURATION: usize, TAG: TimeTag> DeserializeAs<'de, Duration>
...
+    fn deserialize_as<D>(deserializer: D) -> Result<core::time::Duration, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let duration: Self = Self::deserialize(deserializer)?;
+        if duration.duration.is_negative() {
+            return Err(D::Error::custom(
+                "Negative duration is not supported for std::time::Duration",
+            ));
+        }
```
danielSanchezQ, line 129: This should actually never happen, as we bound the duration. But its not bad to have in case we change or expand the implementation.

bacv (approved):

PR #1183, opened 2025-03-21
feat: migrate to overwatch compile-time checks

ntn-x2:

This PR introduces the Overwatch changes introduced in the following PRs:
* https://github.com/logos-co/Overwatch/pull/60
  * It removes the need to implement `RelayMessage` for all message types
  * The `OverwatchHandle` relay has been unified, meaning that now also requesting a relay is an `async` operation that should be done from within a service's `run` method rather than its `init`. The function can still fail, but it won't fail if the target service does not exist as that is now checked at compile time. Some other functions inside the same `OverwatchHandle` do not fail anymore since we eliminated the error that a target service might not exist.
* https://github.com/logos-co/Overwatch/pull/63
  * Removes `StateOperator` settings, requiring implementers to use the `State`'s settings instead.
* https://github.com/logos-co/Overwatch/pull/58 (biggest changeset)
  * It introduces the notion of `RuntimeServiceId`, which is a compile-time aggregated enum of all services that are part of a specific runtime. This means that requiring a relay for a service that is not part of the same runtime is now a compile time issue. Same with implementing any services on top of those services. For instance, implementing an HTTP API for DA blobs will now fail to compile if the required DA services are not part of the node that exposes such APIs
  * Service-defined string IDs are now replaced by their runtime-aggregated variant, so the name of the service that will show up in tracing matches the name of property of the services when added to the overwatch runtime. For instance, for the `pong: PongService` property, the tracing name will be `RuntimeServiceId::Pong`. If it were to be defined as `random_service: PongService`, then it would be logged as `RuntimeServiceId::RandomService`. This shifts responsibility of declaring tracing logic from the service itself to the runtime, meaning that the same service could be logged differently in two different runtimes if they are named differently when they are added to the runtime.
* The `RuntimeServiceId` type is now almost ubiquitous, and unfortunately @danielSanchezQ and I did not find a way to avoid that to be the case. I think folks will get used quite fast to it, as usually this now requires trait bounds to include `AsServiceId<TargetService>` when a service requires interaction with another service. This `AsServiceId` trait is implemented by the `#[derive_services]` procedural macro, which also simplifies service declaration by removing the need to add `OpaqueServiceHandle`s
* The rest of the changes are mostly a consequence of these new requirements
  * With the new compile-time checks, I also noticed some validator types were re-used in the executor but were still using validator adapters that were hardcoded in the exposed types, which these compile-time checks were now able to catch. I will add a comment in the code for a couple of such cases.

I hope so. I am available to guide reviewers through the codebase if needed.

@ntn-x2 

N/A

No.

## Checklist

* [x] 1. Description added.


ntn-x2:
RISC0 seems broken again, but it has nothing to do with this PR, so please ignore any CI errors...

nodes/nomos-node/src/api/backend.rs
```diff
@@ -214,12 +233,108 @@ where
...
+                SamplingBackend,
+                SamplingNetworkAdapter,
+                SamplingRng,
+                SamplingStorage,
+                DaVerifierBackend,
+                DaVerifierNetwork,
+                DaVerifierStorage,
+                ApiAdapter,
+                RuntimeServiceId,
+            >,
```
danielSanchezQ, line 336: I'm wondering if this could be extracted somehow

nodes/nomos-node/src/lib.rs
```diff
@@ -1,225 +1,184 @@
...
-impl StorageSerde for Wire {
-    type Error = wire::Error;
-
-    fn serialize<T: Serialize>(value: T) -> Bytes {
-        wire::serialize(&value).unwrap().into()
-    }
-
-    fn deserialize<T: DeserializeOwned>(buff: Bytes) -> Result<T, Self::Error> {
-        wire::deserialize(&buff)
-    }
```
danielSanchezQ, line 224: Where did the wire stuff go?

nomos-services/data-availability/indexer/src/lib.rs
```diff
@@ -523,44 +462,83 @@ where
...
+                SamplingNetworkAdapter,
+                SamplingRng,
+                SamplingStorage,
+                DaVerifierBackend,
+                DaVerifierNetwork,
+                DaVerifierStorage,
+                TimeBackend,
+                ApiAdapter,
+                RuntimeServiceId,
+            >>()
```
danielSanchezQ, line 536: Can't the generics be elided? :thinking:

danielSanchezQ:
> RISC0 seems broken again, but it has nothing to do with this PR, so please ignore any CI errors...

@bacv can we try to fix risc0? I would like to see all green in this one.

nomos-services/data-availability/indexer/src/lib.rs
ntn-x2: Yeah good point. I think for some it's not possible, but for other it should be. I will do that 👍

nodes/nomos-node/src/lib.rs
ntn-x2: It's still there, just higher up in the module.

nodes/nomos-node/src/api/backend.rs
ntn-x2: I mean, the only way would be a supertrait, which needs to anyway be defined somewhere. I don't think it makes things any better. Do you have any ideas? It is true that it will not scale well with more services, but I can't think of another way.

danielSanchezQ: Yeah, I do not think there is another way. In this case, I think we should probably migrate everything that has to deal with the node (outer layer) to use the exact types, as we have them defined in the crate already. That would simplify everything at a single point. Not to do in this PR ofc.

nomos-services/blend/src/lib.rs
```diff
@@ -44,58 +47,64 @@ use tokio_stream::wrappers::{IntervalStream, UnboundedReceiverStream};
...
-    Backend: BlendBackend + Send + 'static,
+    Backend: BlendBackend<RuntimeServiceId> + Send + 'static,
     Backend::Settings: Clone,
     Backend::NodeId: Hash + Eq + Unpin,
-    Network: NetworkAdapter + Send + Sync + 'static,
+    Network: NetworkAdapter<RuntimeServiceId> + Send + Sync + 'static,
     Network::BroadcastSettings:
         Clone + Debug + Serialize + DeserializeOwned + Send + Sync + 'static,
+    RuntimeServiceId: AsServiceId<NetworkService<Network::Backend, RuntimeServiceId>>
+        + AsServiceId<Self>
```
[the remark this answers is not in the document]
danielSanchezQ: > Is this a key point that ensures that both `NetworkService` and `Self` (`BlendService`) must be part of a "specific" runtime that implmements the `RuntimeSerivceId` trait?

Yes, now all inter-services interactions through Overwatch are compiled-time checked!!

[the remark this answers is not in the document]
ntn-x2: Correct. The trait is implemented by the runtime macro, so it won't be implemented for the specific `RuntimeServiceId` if they are not part of the same runtime.

nodes/nomos-node/src/api/backend.rs
ntn-x2: Yes, the problem is that these are also used by the executor, so it won't compile anymore if we use concrete types. We would need to either duplicate code, or leave it as it is. As far as I can see.

danielSanchezQ: > Yes, the problem is that these are also used by the executor, so it won't compile anymore if we use concrete types. We would need to either duplicate code, or leave it as it is. As far as I can see.

That will not be a problem. We are gonna have to merge both in the same binary. So we will do it when doing that later on.

ntn-x2: Amazing 👍

ntn-x2: Then all these trait bounds will disappear.

nomos-services/data-availability/indexer/src/lib.rs
ntn-x2: This is the best we can do. We can't completely remove them where there's at least two `AsServiceId` bounds: https://github.com/logos-co/nomos/pull/1183/commits/ed3f51f427c1c41a06f91d4d29746d862bedf47a.

ntn-x2:
I also took the liberty to update the README in https://github.com/logos-co/nomos/pull/1183/commits/0b02212080a42408f0356e2e5c8be012a56b04e5. The formatter I have also updated the ToC, which was nice to discover. Other than that, I update the annotation to use the new procedural macro, and also took the liberty not to include `RuntimeServiceId` in those services. Once we start testing the examples, we will have to do things properly, but for now I think this is good enough.

nomos-services/blend/src/lib.rs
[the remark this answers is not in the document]
ntn-x2: @youngjoon-lee to add to the answers here, the `AsServiceId<Self>` specifically is used by the `should_stop_service` function in the utils crate, and makes sure that the service itself is also part of the runtime, and we can generate something `Display`able for it, which is logged using the tracing library from within the `should_stop_service_ function.

nomos-services/data-availability/indexer/src/lib.rs
danielSanchezQ: > This is the best we can do. We can't completely remove them where there's at least two `AsServiceId` bounds: [ed3f51f](https://github.com/logos-co/nomos/commit/ed3f51f427c1c41a06f91d4d29746d862bedf47a).

This look much better IMO. Thanks!

nomos-services/blend/src/lib.rs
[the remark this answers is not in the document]
youngjoon-lee: Thank you, guys. This is cool. I'm resolving this thread.

youngjoon-lee:
Thanks. This is a cool stuff. I think I got the key point and went through changes. They look good, but I think I'm not the best person to turn a green light on this PR. So, I'll wait for reviews from someone who had reviewed the Overwatch PRs.

danielSanchezQ (approved):
:rocket:

zeegomo (commented):
This PR is ok, but I'm wondering whether getting compile time checks of whether a service is there or not is worth +1k lines and having another generic type in almost every service thing.
A missing service is something that should be very easily caught by some primitive test / just launching the node, and is relatively stable. We don't expect to have millions of different configurations.

The rest of the changes are ok.

This being said, I'm not much in engineering anymore and I don't want to block this, but we knew that this approach had the possibility of exploding into a lot intricate type stuff. 
For me, this is where I'd personally draw the line of this getting a bit too convoluted for the benefits it brings, but it does not drastically change the way we're doing things, so feel free to merge on if you think otherwise

danielSanchezQ:
> This PR is ok, but I'm wondering whether getting compile time checks of whether a service is there or not is worth +1k lines and having another generic type in almost every service thing. A missing service is something that should be very easily caught by some primitive test / just launching the node, and is relatively stable. We don't expect to have millions of different configurations.

Yeah, you are not wrong. We discussed this. But this change should be a one-thing phase. It also simplified some other code. I do not expect to be a nuance from this pr forward (hopefully).

ntn-x2:
@zeegomo yes, as @danielSanchezQ said this is something we discussed (more than once, actually, as also I myself was not sure this is something we want, although I came up with this proposal in the first place). The point, nevertheless, is that we added those check in overwatch, which is potentially a framework other projects will use as well, and having compile-time checks as part of the framework, I think, is a very strong value selling proposition. Going forward, as Daniel also mentioned, it's enough for a service to require that a different service be part of the runtime, or else compilation will fail.

And to your point about testing, there are some more nuanced components which can easily get skipped upon testing, as was the case of the [comment](https://github.com/logos-co/nomos/pull/1183#discussion_r2013614637) I left above. But especially as the team grows bigger, I think it's definitely a very good advantage having the compiler prevent us from doing things wrong, before we catch those with tests, which anyway provide less guarantees about correctness than the compiler. This said, I partially share your view, but I think the value added by this overcomes the (mostly apparent) disruption this PR introduces.

README.md
```diff
@@ -49,13 +53,13 @@ This modular approach allows for easy replacement of components in a declarative
 For example:
 
 ```rust ignore
-#[derive(Services)]
+#[derive_services]
```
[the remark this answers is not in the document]
ntn-x2: A derive macro is not allowed to change the input stream, while we needed that for a couple of things, including removing the unnecessary `OpaqueServiceHandle` wrapper. This is only possible with a full procedural macro.

zeegomo:
> But especially as the team grows bigger, I think it's definitely a very good advantage having the compiler prevent us from doing things wrong,

I would actually make the case that they also 'pollute' in a way the codebase with things that are less about the logic of what the program does and more about some specific tool, raising the learning curve. This is true also for users of Overwatch as a dependency, and btw we don't have intentions to develop/maintain Overwatch more than what we need from it for our needs.

> before we catch those with tests, which anyway provide less guarantees about correctness than the compiler

Compile time checks are really nice (and one of the big points of Rust) but can only go so far, and are especially useful for things that are hard to test for (e.g memory safety). If our test suite can't detect a missing service we have bigger things to worry about.

ntn-x2:
> Compile time checks are really nice (and one of the big points of Rust) but can only go so far, and are especially useful for things that are hard to test for (e.g memory safety). If our test suite can't detect a missing service we have bigger things to worry about.

Yes, I understand your point, and I hope you agree with me it's a matter of tradeoffs. For me, the more we can push at compile time, the better, even at the cost of increased complexity and verbosity: the compiler errors also tend to be quite user-friendly and guide you through the process. I would leave to tests, as you say, anything that we cannot validate at compilation time. Consensus, among the people that engaged with this change, seems to be this is a compromise we are willing to make. I do see your point tho and why you are making it. Of course, we can always switch back to the old model if we see the new one is too cumbersome, it will definitely be much faster to do that. So thank you very much from detailing your feedback, I really appreciate it.

> and btw we don't have intentions to develop/maintain Overwatch more than what we need from it for our needs

This somehow caught me by surprise, since we are making concrete efforts to improve documentations and to publish it. But regardless of that, the same applies as what I mentioned above.

danielSanchezQ:
> This somehow caught me by surprise, since we are making concrete efforts to improve documentations and to publish it. But regardless of that, the same applies as what I mentioned above.

The primary users is always gonna be us. It will be out, and if community engages in improving it in ways we consider valid, so be it.

ntn-x2:
> The primary users is always gonna be us. It will be out, and if community engages in improving it in ways we consider valid, so be it.

Yeah I mistook publishing it with developing it for others 👍

zeegomo:
> I hope you agree with me it's a matter of tradeoffs

I agree!

> For me, the more we can push at compile time, the better, even at the cost of increased complexity and verbosity:

I do not agree  😄 

Btw, I raised the points I wanted to, you have green light from me

youngjoon-lee (approved):
I made another round to look through the code and recent comments.

PR #1189, opened 2025-03-25
chore: add dependency graph freshness checks

ntn-x2:

Add a graph of all dependencies **within** the workspace, and adds a GH job to make sure it never gets outdated.

For an easy comparison inside the CI, we need to use the `.dot` format that `cargo-depgraph` outputs. If we want to generate an image out of it, developers can simply use the [graphviz](https://graphviz.org/) library to generate an image out of it. For instance, `dot -Tpng dependencies_graph.dot > out.png` will generate a PNG, `dot -Tsvg dependencies_graph.dot > out.svg` will generate an SVG, or `dot -Tpdf dependencies_graph.dot > out.pdf` will generate a PDF. The same way, if the graph is needed in some documents or in the README, it can be converted on-the-fly to the required image format using any libraries that can deal with `.dot` files.

Generating an up-to-date version of this graph requires [cargo-depgraph](https://github.com/jplatte/cargo-depgraph) and running the command `cargo depgraph --all-deps --dedup-transitive-deps --workspace-only --all-features > dependencies_graph.dot` from the project root.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 1. Description added.


ntn-x2:
Not very trivial and I do not want to complicate our CI for such small added benefit. Will close for now.

ntn-x2:
I gave it another try, and this is what we could do. Comparing the native `.dot` formats, with the `dependencies_graph.dot` representing the single source of truth inside the repo, that the new job would verify is always up to date. Link to the Discord discussion for more context: https://discord.com/channels/1111286067413405788/1111288368723148941/1352565399669051423.

What do you guys think? If you guys think this would be too disruptive, we don't have to have it committed and I can keep using it on my own whenever I need to. 

@danielSanchezQ @bacv @youngjoon-lee @AlejandroCabeza

ntn-x2:
> So the action would fail when the dependency graph differs from the one in the repository as a reminder to update the file?

Correct, this is to keep the job as simple as possible. As discussed in Discord, we could do more in the future, including pushing a new commit automatically with the up-to-date graph information, or display the new graph (if different from the previous) as part of the PR (description or comment), and more. But I thought I would keep it simple for now.

PR #1191, opened 2025-03-26
chore: pin GH actions

ntn-x2:

https://www.cisa.gov/news-events/alerts/2025/03/18/supply-chain-compromise-third-party-github-action-cve-2025-30066. Tried to use the latest version for everything. If something breaks, I will switch back to the right version. Since Clippy checks skip risc0 building, we do not need to install risc0 in the first place.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 1. Description added.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,7 @@
 [toolchain]
+# Keep this version in sync also in the following places:
+# * Dockerfile
+# * shell.nix
+# * ci/Dockerfile
+# * testnet/Dockerfile
 channel = "1.85.1"
--- a/shell.nix
+++ b/shell.nix
@@ -6,7 +6,7 @@
     overlays = [
       (import (fetchGit {
         url = "https://github.com/oxalica/rust-overlay.git";
-        rev = "8b81b8ed00b20fd57b24adcb390bd96ea81ecd90";
+        rev = "a0e3395c63cdbc9c1ec17915f8328c077c79c4a1";
       }))
     ];
    }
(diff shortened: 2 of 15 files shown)


shell.nix
```diff
@@ -17,7 +17,7 @@ pkgs.mkShell {
 
   buildInputs = with pkgs; [
     pkg-config
-    rust-bin.stable."1.80.0".default
+    rust-bin.stable."1.85.1".default
```
bacv, line 20: For this version to be accessible in here, we need to update the overlay commit hash to the latest (few lines above):
```
overlays = [
      (import (fetchGit {
        url = "https://github.com/oxalica/rust-overlay.git";
        rev = "a0e3395c63cdbc9c1ec17915f8328c077c79c4a1";
      }))
 ]
```
I think comment here would also help.

rust-toolchain.toml
```diff
@@ -1,2 +1,7 @@
 [toolchain]
+# Keep this version in sync also in the following places:
+# * Dockerfile
+# * shell.nix
+# * ci/Dockerfile
+# * testnet/Dockerfile
```
bacv, line 6: 👍

shell.nix
ntn-x2: Is my understanding correct? https://github.com/logos-co/nomos/pull/1191/commits/ec95642864e62b8348edeb7cf7e482417540754c

bacv: > Is my understanding correct?

Yes!

bacv (approved):

PR #1192, opened 2025-03-26
chore: Update risc0 to v2.0.0

bacv:
Use risc0 v2.0.0

### DIFF
--- a/.github/actions/install-risc0/action.yml
+++ b/.github/actions/install-risc0/action.yml
@@ -16,7 +16,7 @@ runs:
         # Required to avoid throttling by GitHub
         GITHUB_TOKEN: ${{ inputs.github_token }}
     - name: Install risc0 using cargo-binstall
-      run: cargo binstall -y cargo-risczero@1.2.0 && cargo risczero install
+      run: cargo binstall -y cargo-risczero@2.0.0 && cargo risczero install --version r0.1.85.0
       shell: bash
       env:
         # Required to avoid throttling by GitHub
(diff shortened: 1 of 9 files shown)


.github/actions/install-risc0/action.yml
```diff
@@ -16,7 +16,7 @@ runs:
         # Required to avoid throttling by GitHub
         GITHUB_TOKEN: ${{ inputs.github_token }}
     - name: Install risc0 using cargo-binstall
-      run: cargo binstall -y cargo-risczero@1.2.0 && cargo risczero install
+      run: cargo binstall -y rzup && ~/.cargo/bin/rzup install cargo-risczero 2.0.0 && ~/.cargo/bin/rzup install rust 1.85.0
```
ntn-x2, line 19: Can you add a a comment in the toolchain file with all the places where the Rust version must be updated as well?

ntn-x2: Also I think we use already 1.85.1, does it matter?

bacv: Risc0 uses it's own toolchain, it downloads it in a separate directory, this is not related to the toolchain used in Nomos repo.

ntn-x2 (approved):

.github/actions/install-risc0/action.yml
bacv: > Also I think we use already 1.85.1, does it matter?

Doublechecked, current `rzup` version downloads required toolchain to `.rustup` directory. Previous version would download it to `~/.risc0/toolchain`. Not sure if there would be a benefit to add both required versions to `nomos` toolchain file, I think it's better keep them seperate, although if we've required both inside nomos toolchain file, we could remove `rzup install rust ...` line. I think it's clearer to have it here.

ntn-x2: No, what I meant is that we write in the toolchain file all places where the Rust version must be manually checked. But I am doing that as part of https://github.com/logos-co/nomos/pull/1191, since it's not really in the scope of this PR anyway, so all good. Perhaps you can add your comments to that PR once you review it.

PR #1200, opened 2025-03-31
feature(sdp): Protocol definitions and state transitions

bacv:

Sdp protocol structures and state transitions defined:
- Newly created provider can only declare service when it's created.
- Active provider becomes inactive after inactivity period defined in service params.
- Inactive provider can request rewards and become active.
- Inactive and active providers can withdraw declaration.
- Withdrawn declaration can't become active or inactive.
- State transition can only happen in the same or future block number.

[Specification](https://www.notion.so/Service-Declaration-Protocol-Specification-17b8f96fb65c80c69c2ef55e22e29506)

@madxor @bacv 

Yes

No deviations from specification

## Checklist

* [x] 1. Description added.


nomos-sdp/src/state.rs
```diff
@@ -0,0 +1,608 @@
...
+    const fn try_into_active(
+        mut self,
+        block_number: BlockNumber,
+        event_type: EventType,
+    ) -> Result<ActiveState, InactiveStateError> {
+        match event_type {
+            EventType::Reward => {
+                self.0.rewarded = Some(block_number);
+                Ok(ActiveState(self.0))
+            }
```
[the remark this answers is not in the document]
bacv: This wasn't mentioned specifically in the specification, but is based on couple of related sentences

[Reward section](https://www.notion.so/Service-Declaration-Protocol-Specification-17b8f96fb65c80c69c2ef55e22e29506?pvs=4#1b38f96fb65c802580d9c334281117a3)
> We presume that an active node periodically requests rewards, while an inactive node does not. We utilize this as a non-intrusive differentiator of node activity. It is crucial to exclude inactive nodes from the set of active nodes, as it enhances the stability of services.

[Inactivity period section](https://www.notion.so/Service-Declaration-Protocol-Specification-17b8f96fb65c80c69c2ef55e22e29506?pvs=4#1b38f96fb65c8025ba1ef9aab0020003)
> `Inactivity_Period` defines the maximum time (as a number of blocks) during which a reward must be requested; otherwise, the declaration is considered inactive.

I agree that this could be defined explicitly, or have a short summary like I've added to the PR description.

nomos-sdp/src/state.rs
```diff
@@ -0,0 +1,608 @@
...
+        }
+
+        // Newly created provider is considered active.
+        if provider_info
+            .created
+            .wrapping_add(service_params.inactivity_period)
+            > block_number
+        {
+            return Ok(ActiveState(*provider_info).into());
+        }
```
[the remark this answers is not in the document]
bacv: This section checks if recently created provider is still considered active even without having requested a reward yet.

The line bellow checks if provider has ever got the reward first (`provider_info.rewarded` is set to `None` when created).

Will add comment

nomos-sdp/src/state.rs
```diff
@@ -0,0 +1,608 @@
...
+        };
+
+        let inactive_state =
+            ProviderState::try_from_info(100, &provider_info, &service_params).unwrap();
+        assert!(matches!(inactive_state, ProviderState::Inactive(_)));
+
+        // Inactive state can't declare a service.
+        let active_state = inactive_state.try_into_active(100, EventType::Declaration);
+        assert!(active_state.is_err());
+
```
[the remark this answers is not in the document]
bacv: Just a few lines bellow this we have:
```
 let inactive_state =
  ProviderState::try_from_info(100, &provider_info, &service_params).unwrap();
let withdrawn_state =
  inactive_state.try_into_withdrawn(115, EventType::Withdrawal, &service_params);
```
Do you mean to add something like this:
```
let inactive_state =
  ProviderState::try_from_info(100, &provider_info, &service_params).unwrap();
let active_state =
  inactive_state.try_into_active(105, EventType::Reward).unwrap();
let withdrawn_state =
  inactive_state.try_into_withdrawn(115, EventType::Withdrawal, &service_params);
```
Maybe you could add a suggestion if I misunderstand.

nomos-sdp/src/state.rs
```diff
@@ -0,0 +1,608 @@
...
+        assert!(matches!(inactive_state, ProviderState::Inactive(_)));
+
+        // Withdrawal should fail before the lock period is over.
+        let withdrawal_attempt =
+            inactive_state.try_into_withdrawn(15, EventType::Withdrawal, &service_params);
+        assert!(withdrawal_attempt.is_err());
+    }
+
+    #[test]
+    fn test_active_cannot_reward_directly() {
```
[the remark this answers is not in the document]
bacv: >  But, if the node wants to withdraw + reward at the same time, the implementation requires to handle EventType::Withdraw first to transit ActiveState to WithdrawState, and then handle EventType::Reward.

Yes this is correct, active state can request withdraw whenever it wants, but when the state is `Withdrawn` it can only request reward during the same block.

I'll rename the test.

nomos-sdp/src/state.rs
```diff
@@ -0,0 +1,608 @@
...
+impl ActiveState {
+    const fn try_into_updated(
+        mut self,
+        block_number: BlockNumber,
+        event_type: EventType,
+    ) -> Result<Self, ActiveStateError> {
+        match event_type {
+            EventType::Declaration => {
+                if self.0.created == block_number {
+                    Ok(self)
```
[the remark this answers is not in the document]
bacv: Is not intended to check for duplicates in the same block, we can actually write something like this in the scope of state transitions (but this is not allowed in the sdp ledger as we check for pending states):
```
let active_state = ProviderState::try_from(new_provider_info_created_at_block_999);
let state1 = active_state.to_active(999, EventType::Declaration);
let state2 = state1.to_active(999, EventType::Declaration);
let state3 = state1.to_active(999, EventType::Declaration);
...
```
As long as it all happens during the block as the provider was created. In other words - provider can declare service only when the provider_info is created.

This test does not allow active state (aka already existing provider) to handle `EventType::Declaration` in future blocs, for example this would fail:
```
let active_state = ProviderState::try_from(new_provider_info_created_at_block_999);
let err_not_state = active_state.to_active(1000, EventType::Declaration); // not the same block
```

nomos-sdp/src/state.rs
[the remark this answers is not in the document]
madxor: I've added a sentence that makes it explicit: https://www.notion.so/Service-Declaration-Protocol-Specification-17b8f96fb65c80c69c2ef55e22e29506?pvs=4#1c98f96fb65c80cfafc9f0890b6ed462

nomos-sdp/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+use std::{
+    collections::{HashMap, HashSet},
+    hash::Hash,
+};
+
+use blake2::{Blake2b, Digest};
+use multiaddr::Multiaddr;
+
+pub type StakeThreshold = u64;
+pub type BlockNumber = u64;
```
danielSanchezQ, line 12: This probably should be `NonZero`.

nomos-sdp/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+    GenericRestaking,
+}
+
+pub type Nonce = [u8; 16];
+#[derive(Clone, Copy, Eq, PartialEq, Hash, Debug)]
+pub struct ProviderId(pub [u8; 32]);
+#[derive(Clone, Copy, Eq, PartialEq, Hash, Debug)]
+pub struct DeclarationId(pub [u8; 32]);
+#[derive(Clone, Copy, Eq, PartialEq, Hash, Debug)]
+pub struct RewardId(pub [u8; 32]);
```
danielSanchezQ, line 47: Give some air between types, this is quite packed and difficult to read 😅

nomos-sdp/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+    pub inactivity_period: u64,
+    pub retention_period: u64,
+    pub reward_contract: ContractAddress,
+    pub timestamp: BlockNumber,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Locator {
+    addr: Multiaddr,
+}
```
danielSanchezQ, line 31: Should this be a wrapper type?
In fact, does this multiaddress matches the restrictions from the spec (in terms of size)?
> The length of a regular libp2p multiaddr is 83 characters in the longest case. However, if we allow to Dynamic DNS for retrieving dynamically set global IP addresses for some nodes, then the longest theoretical DNS name is 253 characters. Now assuming that this lengthy DNS must be part of the multiaddr, then we get 329 characters in the longest case.

nomos-sdp/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+
+impl<Proof> DeclarationMessage<Proof> {
+    fn declaration_id(&self) -> DeclarationId {
+        let mut hasher = Blake2b::new();
+        for locator in &self.locators {
+            hasher.update(locator.addr.as_ref());
+        }
+        DeclarationId(hasher.finalize().into())
+    }
+}
```
danielSanchezQ, line 133: @madxor The spec doesn't specify the hasher. We use blake2b but I think we should specify it in the spec as well. As it is not really implementation details. Someone that has to implement the same should use the same hasher.

nomos-sdp/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+    pub declaration_id: DeclarationId,
+    pub service_type: ServiceType,
+    pub provider_id: ProviderId,
+    pub nonce: Nonce,
+    pub metadata: Option<Metadata>,
+}
+
+impl<Metadata> RewardMessage<Metadata> {
+    pub fn reward_id(&self) -> RewardId {
+        let mut hasher = Blake2b::new();
```
danielSanchezQ, line 155: Same here. I couldn't find a reference in the spec (`ctrl+f`) @madxor .

nomos-sdp/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+
+        Ok(Self {
+            declaration_id: withdraw.declaration_id,
+            service_type: withdraw.service_type,
+            provider_id: withdraw.provider_id,
+            nonce: withdraw.nonce,
+            metadata: withdraw.metadata,
+        })
+    }
+}
```
danielSanchezQ, line 185: If we have a `NoMetadata` construct then it should not be optional in the struct message declaration. I'm not sure if the plan was to change it later on.

nomos-sdp/src/state.rs
```diff
@@ -0,0 +1,615 @@
...
+)]
+impl Into<ProviderInfo> for ProviderState {
+    fn into(self) -> ProviderInfo {
+        match self {
+            Self::Active(active_state) => active_state.0,
+            Self::Inactive(inactive_state) => inactive_state.0,
+            Self::Withdrawn(withdrawn_state) => withdrawn_state.0,
+        }
+    }
+}
```
danielSanchezQ, line 180: Wait, why this cannot be a `from` impl? 
```rust
impl From<ProviderState> for ProviderInfo {
  fn from(state: :ProviderState) -> ProviderInfo {
    match state {
            ProviderState::Active(active_state) => active_state.0,
            ProviderState::Inactive(inactive_state) => inactive_state.0,
            ProviderState::Withdrawn(withdrawn_state) => withdrawn_state.0,
        }
}
}
```

nomos-sdp/src/state.rs
```diff
@@ -0,0 +1,615 @@
...
+    #[error("Provided block is in past, can not transition to state in previous block")]
+    BlockFromPast,
+}
+
+#[derive(Debug, Eq, PartialEq, Hash)]
+pub enum ProviderState {
+    Active(ActiveState),
+    Inactive(InactiveState),
+    Withdrawn(WithdrawnState),
+}
```
danielSanchezQ, line 166: Not saying this is the wrong way to do it. But couldn't we encode it using a type state pattern instead? Are we sure all this errors running around are always due to runtime errors? Or are just because we are encoding all kind of transitions?
I would like to understand it a bit better. Maybe a graph with the state machine transition would be helpful in the module documentation.

nomos-sdp/src/lib.rs
bacv: Why couldn't it be zero?

danielSanchezQ: Wait, totally brainfart...I was thinking of negative values and mixed everything 😵

nomos-sdp/src/state.rs
bacv: It's totally doable with the typestate pattern, but I fear that we'll endup with a large number of weird states like `ActiveRewarded`, `ActiveNotRewarded`, `InactiveRewarded` etc, so chose partial typestate. Even with full list of states we'll need an enum that combines all of them, because we derive the state from the `ProviderInfo` and we are not sure what state it is during the time the message is received.

nomos-sdp/src/state.rs
bacv: Added a clippy comment above: `ProviderState can not be derived from ProviderInfo alone` - we need `ServiceParams` and `BlockNumber` to derive `ProviderState` from `ProviderInfo`, but we can directly derive `ProviderState` into `ProviderInfo`.

nomos-sdp/src/lib.rs
danielSanchezQ: > Reward message can have no metadata and still be rewarded. But - withdrawal message - if it has no metadata, then it can't be rewarded.

Yes, but for me it makes no sense that you can have a `Some(NoMetadata)` in any construct. Or I do not see the case at least.

nomos-sdp/src/state.rs
danielSanchezQ: I do not see the difference. Where are the `ServiceParams` and `BlockNumber` in the current impl?

nomos-sdp/src/lib.rs
bacv: Hm, I'm puzzled, in what case we can have `Some(NoMetadata)`? This method (`TryFrom<Withdrawal> for Reward`) returns a result. So we'll either have `Ok(Reward)` or `Err(NoMetadata)`?

nomos-sdp/src/state.rs
danielSanchezQ: Ok, lets sync about this. I would like to deeply understand this. Thanks.

nomos-sdp/src/state.rs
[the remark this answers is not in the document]
bacv: There are now two tests `test_inactive_to_withdraw` and `test_inactive_to_active_to_withdraw` in https://github.com/logos-co/nomos/pull/1200/commits/a46bcb41d76a18924875c756f0bed6cee1238a0f

nomos-sdp/src/state.rs
```diff
@@ -0,0 +1,636 @@
...
+    ToWithdrawalInvalidEvent(EventType),
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+pub struct ActiveState(ProviderInfo);
+
+impl ActiveState {
+    const fn try_into_updated(
+        mut self,
+        block_number: BlockNumber,
```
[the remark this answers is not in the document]
madxor: If there is no "block number" inside of the block, then the height of the chain is the only option. I think we still need to keep track of the number of blocks somewhere as part of the branch/fork selection logic.

danielSanchezQ (approved):

PR #1215, opened 2025-04-04
Socket per ntp request 

bacv:
Suggestion

### DIFF
--- a/nomos-services/time/src/backends/ntp/async_client.rs
+++ b/nomos-services/time/src/backends/ntp/async_client.rs
@@ -1,6 +1,6 @@
 use std::{
     fmt::{Debug, Formatter},
-    net::SocketAddr,
+    net::{IpAddr, SocketAddr},
     time::Duration,
 };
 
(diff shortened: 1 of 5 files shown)


nomos-services/time/src/backends/ntp/async_client.rs
```diff
@@ -73,12 +80,9 @@ impl AsyncNTPClient {
         &self,
         pool: Addresses,
     ) -> Result<NtpResult, Error> {
-        let socket = &UdpSocket::bind(self.settings.local_socket)
-            .await
-            .map_err(Error::Io)?;
         let hosts = lookup_host(&pool).await.map_err(Error::Io)?;
         let mut checks = hosts
-            .map(move |host| get_time(host, socket, self.ntp_context))
+            .map(move |host| self.get_time(host))
```
danielSanchezQ, line 85: nitpik: no need to move anymore.

danielSanchezQ (approved):

strinnityk (approved):
Fuckin' A. This was some nice catch.
I learned about the port `0` usage, which makes a lot of sense, and a way better explanation for the `ResponseAddressMismatch`.
Lovely

PR #1217, opened 2025-04-06
feat(sdp): Implement sdp service

pradovic:

This is a service layer for SDP service without the concrete implementation of the adapters.

[Specification](https://www.notion.so/Service-Declaration-Protocol-Specification-17b8f96fb65c80c69c2ef55e22e29506)

@pradovic @bacv 

Yes

Yes
## Checklist

> [!WARNING]  

* [x] 1. Description added.


nomos-services/sdp/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+                    tracing::error!("Error processing SDP message: {}", e);
+                }
+            }
+            SdpMessage::MarkInBlock {
+                block_number,
+                result_sender,
+            } => {
+                let result = backend.mark_in_block(block_number).await;
+                // todo: should the result_sender been send to the backend?
+                let _ = result_sender.send(result);
```
bacv, line 112: I think result sender should stay here, the only improvement would be to use tracing to track error just like in line `103` above.

nomos-services/sdp/src/backends/mod.rs
```diff
@@ -0,0 +1,20 @@
+use overwatch::overwatch::handle::OverwatchHandle;
+
+#[async_trait::async_trait]
+pub trait SdpBackend<RuntimeServiceId> {
+    type BlockNumber: Clone + Send + Sync;
+    type Message: Send + Sync;
+    type Error: std::error::Error + 'static + Send + Sync;
+    type Settings: Clone + Send + Sync + 'static;
+
+    fn new(settings: Self::Settings, overwatch_handle: OverwatchHandle<RuntimeServiceId>) -> Self;
```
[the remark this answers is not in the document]
danielSanchezQ: Also lets just add the bounds if they are necessary, we are trying to cleanup those in the trait definition if possible.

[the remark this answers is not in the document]
pradovic: Makes sense. I will clean those def up more when I integrate the `SdpLedger` backend implementation.

nomos-services/sdp/src/backends/sdpledger.rs
```diff
@@ -0,0 +1,61 @@
...
+    Metadata: Send + Sync + 'static,
+    SdpLedgerError<ContractAddress>: Send + Sync,
+{
+    type BlockNumber = BlockNumber;
+    type Message = SdpMessage<Metadata, Proof>;
+    type Error = SdpLedgerError<ContractAddress>;
+    type Settings = (Declarations, Rewards, Services, Stakes);
+
+    fn new(settings: Self::Settings) -> Self {
+        let (declaration_repo, reward_request_sender, services_repo, stake_verifier) = settings;
```
bacv, line 36: It appears that precious design where passed overwatch handle will be required here, instead of getting these from settings, we'll need to create adapters for RewardsService and StakesService, the DeclarationsRepository should be part of SdpService implementation and probably use StorageService. Maybe for this PR lets leave the Mock, and create the SdpLedgerBackend in another one, where we use mock for rewards and staking, but write declarations and providers to real storage?

danielSanchezQ: Yeah, but the overwatch handle should still not be passed. All initialization should be done in the service layer. Backend should get just the adapter for whatever it needs. 
I agree we can go forward with a mock. But better mock adapters than keeping all this generics (which will spread) here.

bacv: That's even better approach - @danielSanchezQ what do you think if we just remove the MockSdpLedgerBackend, pass adapters to the actual SdpLedgerBackend, and mock only rewards and staking adapters.

danielSanchezQ: > That's even better approach - @danielSanchezQ what do you think if we just remove the MockSdpLedgerBackend, pass adapters to the actual SdpLedgerBackend, and mock only rewards and staking adapters.

Yes, I think this is the way to do it.

pradovic: Hopefully I did it the way you guys proposed. I did not implement any mocks for now, just left a dummy todo impl for the actual adapters to keep the PR more contained. And we can see which direction to go next, based on what we have. Hope it makes sense

bacv (approved):
One minor change request, but otherwise looks good.

nomos-services/sdp/src/backends/ledger.rs
```diff
@@ -0,0 +1,98 @@
...
+            .await
+            .map_err(map_ledger_error)
+    }
+
+    fn discard_block(&mut self, block_number: Self::BlockNumber) {
+        self.discard_block(block_number);
+    }
+}
+
+fn map_ledger_error<ContractAddress: Debug + Send + Sync + 'static>(
```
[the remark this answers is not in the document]
pradovic: Ah, yes, nice catch, thanks

nomos-services/sdp/src/adapters/rewards/rewards.rs
```diff
@@ -0,0 +1,49 @@
...
+            phantom: std::marker::PhantomData,
+        }
+    }
+}
+
+#[async_trait]
+impl<
+        Metadata: std::marker::Send + std::marker::Sync,
+        ContractAddress: std::marker::Send + std::fmt::Debug + std::marker::Sync,
+    > RewardsRequestSender for LedgerRewardsAdapter<Metadata, ContractAddress>
```
danielSanchezQ, line 37: Better use `where`.

nomos-services/sdp/src/adapters/stakes/stakes.rs
```diff
@@ -0,0 +1,34 @@
+use async_trait::async_trait;
+use nomos_sdp_core::{
+    ledger::{self, StakesVerifierError},
+    ProviderId,
+};
+
+use super::SdpStakesVerifierAdapter;
+
+pub struct LedgerStakesVerifierAdapter<Proof> {
+    phantom: std::marker::PhantomData<Proof>,
```
danielSanchezQ, line 10:
```suggestion
    _proof: std::marker::PhantomData<Proof>,
```

nomos-services/sdp/src/lib.rs
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    async fn run(mut self) -> Result<(), overwatch::DynError> {
+        let mut lifecycle_stream = self.service_state.lifecycle_handle.message_stream();
+        loop {
+            tokio::select! {
+                Some(msg) = self.service_state.inbound_relay.recv()  => {
+                    {
+                        self.handle_sdp_message(msg).await;
+                    }
```
danielSanchezQ, line 173: why double `{}`?

danielSanchezQ:
@bacv . One thing that we need to think about is how to deal with forks here. Which, btw, I do not think it is stated in the specifaction. Lets maybe discuss in discord. wdyt?

bacv:
>  One thing that we need to think about is how to deal with forks here. Which, btw, I do not think it is stated in the specifaction. Lets maybe discuss in discord. wdyt?

Yes, forks are not mentioned in the SDP spec and service would be a good place to handle that. Even without handling that I think this service base could be merged. Lets start a discussion regarding forks.

nomos-services/sdp/src/lib.rs
pradovic: Accidental leftover while refactoring, good catch

danielSanchezQ (approved):
Good barebones, we can work from here.

PR #1219, opened 2025-04-07
PoC(sync): store block IDs in slot order

youngjoon-lee:

Now the consensus service stores block IDs in slot order, so that they can be scanned in slot range when handling sync requests.
Not tested yet.
Scanning blocks are not implemented yet. That will be implemented soon when implementing the sync request handling.

### DIFF
--- a/nomos-services/cryptarchia-consensus/src/lib.rs
+++ b/nomos-services/cryptarchia-consensus/src/lib.rs
@@ -7,10 +7,21 @@ mod states;
 pub mod storage;
 mod sync;
 
+use crate::network::SyncRequest;
+use crate::storage::sync::SyncBlocksProvider;
+use crate::{
+    leadership::Leader,
+    relays::CryptarchiaConsensusRelays,
+    states::{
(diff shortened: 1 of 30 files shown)


nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -1311,6 +1313,56 @@ where
...
+            .await
+            .map_err(|(e, _)| e)?;
+
+        // Build a key: "block/slot/{slot}/id/{id}"
+        let mut key = Vec::with_capacity(11 + 8 + 4 + 32);
+        key.extend_from_slice(b"block/slot/");
+        key.extend_from_slice(&slot.to_be_bytes());
+        key.extend_from_slice(b"/id/");
+        key.extend_from_slice(&id);
+        let msg = <StorageMsg<_>>::new_store_message(key, id);
```
andrussal, line 1356: I had issues with prefix search if using `new_store_message`, it uses internally `bincode`(I think) and this seems to mess up prefix lookup. So I just used "raw" `StorageMsg::Store`.

andrussal (approved):

PR #1222, opened 2025-04-07
fix(DA): Pass share verification domain size from the configuration in verifier

bacv:

When dispersing data to a network with more than two subnets, the share wouldn't be included in the block because verification would fail because of fixed domain size in verifier. To fix this, size is now passed from the configuration, test was added but ignored, because it seems that in a network of 4 nodes a longer time is needed to wait until all nodes mark the blob in block. This will be fixed in the future when in integration tests we'll be able to use mempool inspection tools to know when tests can proceed with assertions.

Yes

@danielSanchezQ @bacv 

Yes

No

## Checklist

* [x] 1. Description added.

### DIFF
--- a/nomos-services/data-availability/verifier/src/backend/kzgrs.rs
+++ b/nomos-services/data-availability/verifier/src/backend/kzgrs.rs
@@ -26,6 +26,7 @@ impl std::error::Error for KzgrsDaVerifierError {}
 
 pub struct KzgrsDaVerifier {
     verifier: NomosKzgrsVerifier,
+    domain_size: usize,
 }
 
 impl VerifierBackend for KzgrsDaVerifier {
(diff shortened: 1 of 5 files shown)


nomos-services/data-availability/verifier/src/backend/kzgrs.rs
```diff
@@ -51,9 +55,8 @@ impl DaVerifier for KzgrsDaVerifier {
     ) -> Result<(), Self::Error> {
         // TODO: Prepare the domain depending the size, if fixed, so fixed domain, if
         // not it needs to come with some metadata.
-        let domain_size = 2usize;
         self.verifier
-            .verify(commitments, light_share, domain_size)
+            .verify(commitments, light_share, self.domain_size)
```
danielSanchezQ, line 59: Oh DAMN!

danielSanchezQ (approved):
Nice catch, thanks for the test as well!

PR #1228, opened 2025-04-09
PoC(sync): use an explicit peer for backfilling

youngjoon-lee:

We don't need to choose the 'best peer' for backfilling. Instead, it's reasonable to request the backfilling explicitly from the peer who provided the orphan block, because it's highly likely that the peer have all ancestors of the orphan block.
Unlike the [executable specification](https://github.com/logos-co/nomos-specs/blob/dcdb41964864e2fc95f9b8ab4e2f7dfd2a05aab2/cryptarchia/sync.py#L147), which tries to continue fetching from other peers if the first peer is unresponsive, our current implementation fetches a backward chain from a single explicit peer just for simplicity. Using multiple peers is an optimization that will be considered later. It's also the same for the forward synchronization.

To implement this, the following changes have been made:
- The `cryptarchia-sync::adapter::BlockFetcher` (renamed from `NetworkAdapter`) requires the `fetch_blocks_from_slot` function to return a stream of `(Block, ProviderId)` (not just a stream of `Block`). This is reasonable because the `fetch_blocks_from_slot` is meant to fetch blocks from all meaningful peers and return those blocks through a single stream. So, each block will have different `ProviderId` (i.e. `PeerId` in libp2p). 
- The sync algorithm triggers a backfilling using an explicit `ProviderId` by calling the `fetch_chain_backward` function.
- There's a cascade of changes due to the above, but it's easy to understand if you look at the diff.
  - Maybe, the changes on `SyncRequest` or `SyncCommand` may not look so optimal but I didn't want to refactor code substantially. I think this change is simple and fine.

### DIFF
--- a/consensus/cryptarchia-sync/network/src/behaviour.rs
+++ b/consensus/cryptarchia-sync/network/src/behaviour.rs
@@ -388,11 +388,10 @@ mod test {
 
     #[tokio::test]
     async fn test_sync_forward() {
-        let (request_senders, handles) = setup_and_run_swarms(NUM_SWARMS).await;
-        let (request_sender, peer_id) = request_senders[0].clone();
+        let (request_senders, peer_ids, handles) = setup_and_run_swarms(NUM_SWARMS).await;
         let blocks = perform_sync(
-            request_sender,
-            peer_id,
+            request_senders[0].clone(),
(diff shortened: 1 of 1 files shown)


andrussal (approved):
👍

PR #1234, opened 2025-04-10
poc(recovery): Recovery integrity and improvements

strinnityk:
- Implement unchecked functions for `Ledger` and `Cryptarchia` so validations can be bypassed for security recovery.
- Split `CryptarchiaConsensus::process_block` into a normal and unchecked version for security recovery.
- Detected and fixed issue where genesis block was being applied two times during security recovery.
- Documentation and naming improvements

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 1. Description added.

### DIFF
--- a/consensus/cryptarchia-engine/src/lib.rs
+++ b/consensus/cryptarchia-engine/src/lib.rs
@@ -98,12 +98,12 @@ where
     ) -> Self {
         let mut branches = self.branches.clone();
 
-        // Add the new header to the branches. Parent doesn't need to be removed since
-        // it should not be in the tips set.
-        // Maybe we can run the `tips.remove(parent)` nonetheless (since it would be a
-        // no-op). That would allow us to reuse this method in the
-        // [`Self::apply_header`].
         let mut tips = self.tips.clone();
+        // When running this method by itself (not through `Self::apply_header`), the
+        // parent should not in the tips set. That being said, having this call
+        // here allows us to reuse this function in the `Self::apply_header`
+        // method. In the unchecked case it'll be a no-op.
+        tips.remove(&parent);
         tips.insert(header);
 
         branches.insert(
@@ -122,30 +122,19 @@ where
     /// Create a new [`Branches`] instance with the updated state.
     #[must_use = "this returns the result of the operation, without modifying the original"]
     fn apply_header(&self, header: Id, parent: Id, slot: Slot) -> Result<Self, Error<Id>> {
-        let mut branches = self.branches.clone();
-
-        // If the parent was the head of a branch, remove it as it has been superseded
-        // by the new header
-        let mut tips = self.tips.clone();
-        tips.remove(&parent);
-        tips.insert(header);
-
-        let length = branches
+        // Calculating the length here allows us to reuse
+        // `Self::apply_header_unchecked`. Could this lead to length difference
+        // issues? We are calculating length here but the `self.branches` is
+        // cloned in the `Self::apply_header_unchecked` method, which means
+        // there's a risk of length being different due to concurrent operations.
+        let length = self
+            .branches
             .get(&parent)
             .ok_or(Error::ParentMissing(parent))?
             .length
             + 1;
-        branches.insert(
-            header,
-            Branch {
-                id: header,
-                parent,
-                length,
-                slot,
-            },
-        );
 
-        Ok(Self { branches, tips })
+        Ok(self.apply_header_unchecked(header, parent, slot, length))
     }
 
     #[must_use]
--- a/consensus/cryptarchia-engine/src/lib.rs
+++ b/consensus/cryptarchia-engine/src/lib.rs
@@ -99,10 +99,6 @@ where
         let mut branches = self.branches.clone();
 
         let mut tips = self.tips.clone();
-        // When running this method by itself (not through `Self::apply_header`), the
-        // parent should not in the tips set. That being said, having this call
-        // here allows us to reuse this function in the `Self::apply_header`
-        // method. In the unchecked case it'll be a no-op.
         tips.remove(&parent);
         tips.insert(header);
 
--- a/consensus/cryptarchia-engine/src/lib.rs
+++ b/consensus/cryptarchia-engine/src/lib.rs
@@ -85,8 +85,7 @@ where
     ///
     /// * `header` - The ID of the block to be added.
     /// * `parent` - The ID of the parent block. Due to the nature of the method
-    ///   (`unchecked`),
-    /// the existence of the parent block is not verified.
+    ///   (`unchecked`), the existence of the parent block is not verified.
     /// * `slot` - The slot of the block to be added.
     /// * `length`: The position of the block in the chain.
     #[must_use = "Returns a new instance with the updated state, without modifying the original."]
@@ -231,8 +230,7 @@ where
     ///
     /// * `header` - The ID of the block to be added.
     /// * `parent` - The ID of the parent block. Due to the nature of the method
-    ///   (`unchecked`),
-    /// the existence of the parent block is not verified.
+    ///   (`unchecked`), the existence of the parent block is not verified.
     /// * `slot` - The slot of the block to be added.
     /// * `length`: The position of the block in the chain.
     #[must_use = "Returns a new instance with the updated state, without modifying the original."]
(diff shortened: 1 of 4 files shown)


consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -70,19 +70,65 @@ where
...
+    /// * `length`: The position of the block in the chain.
+    #[must_use = "Returns a new instance with the updated state, without modifying the original."]
+    fn apply_header_unchecked(&self, header: Id, parent: Id, slot: Slot, length: u64) -> Self {
+        let mut branches = self.branches.clone();
+
+        // Add the new header to the branches. Parent doesn't need to be removed since
+        // it should not be in the tips set.
+        // Maybe we can run the `tips.remove(parent)` nonetheless (since it would be a
+        // no-op). That would allow us to reuse this method in the
+        // [`Self::apply_header`].
```
danielSanchezQ, line 99: This seems reasonable. It is basically a duplicated of the below for a specific case in where the below one applies anyway? `apply_header` is already valid for this case.

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -78,17 +78,19 @@ struct Cryptarchia {
...
 
 impl Cryptarchia {
-    pub fn new(
-        header_id: HeaderId,
-        ledger_state: LedgerState,
+    /// Initialize a new [`Cryptarchia`] instance.
+    /// [`Cryptarchia`] must always be initialized from genesis.
+    pub fn from_genesis(
+        genesis_id: HeaderId,
+        genesis_ledger_state: LedgerState,
```
danielSanchezQ, line 85: Is this true? You can still create a new instance from an advance ledger state and the current tip right? I think `new` wasn't wrong here.

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -785,11 +826,48 @@ where
...
+            DaPool,
+            DaPoolAdapter,
+            NetAdapter,
+            SamplingBackend,
+            SamplingRng,
+            Storage,
+            TxS,
+            DaVerifierBackend,
+            RuntimeServiceId,
+        >,
```
danielSanchezQ, line 847: Can we extract this type into an alias?

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -937,7 +1034,7 @@ where
         output
     }
 
-    fn validate_block(
+    fn validate_blocks_blobs(
```
danielSanchezQ, line 1037: Block validation will eventually check more stuff than just the blobs, revert to `validate_blobs`

danielSanchezQ (commented):
It is not clear to me how do we set the genesis from the security point.

consensus/cryptarchia-engine/src/lib.rs
strinnityk: I thought so, but I'm left wondering whether, for the unchecked case, it'd make sense to return some sort of error if, when removing the parent, it does exist in `tips`:
- On the one hand, I think it doesn't make sense, as unchecked would cover that. 
- On the other, the specific use case for unchecked wouldn't have the parent, so its existence could point to something being off.

nomos-services/cryptarchia-consensus/src/lib.rs
strinnityk: You certianly can, but at some point we stated that `Cryptarchia` should always be initialised from `genesis` ([Discord](https://discord.com/channels/1111286067413405788/1357607529814691872/1357627429405069383)).

I changed it (back) to `from_genesis` to reinforce the idea that this is the correct use case. You can always make a "wrong" use of it and initialise it to any other arbitrary arguments.

On the other hand, we could revert it to `new` and clarify in the comments that the intended case is to use `genesis`.

nomos-services/cryptarchia-consensus/src/lib.rs
strinnityk: I don't think associated type alias are supported yet in stable.
> Feature `inherent associated types` is not available in the current Rust toolchain.
If there's an alternative I'm all ears :)

nomos-services/cryptarchia-consensus/src/lib.rs
strinnityk: Right, but I assume this is the function that only abstracts this specific validation. I created a higher level `validate_received_block` to group all validations (which is the one called from within `process_block`).
Should I still revert it?

strinnityk:
> It is not clear to me how do we set the genesis from the security point.

In the `recover_from_security` method we initialise `Cryptarchia` to genesis. 
Then create a `Leader` with the security poisition data, and inject the security block into `Cryptarchia` with `unchecked`.
After that, the remaining blocks are applied through `process_block_unchecked`.

strinnityk:
>Then, what do you guys think about having a single function like below?
>
>```rust
>fn apply_header(&self, header: Id, parent: Id, slot: Slot, length: Option<u64>) -> Result<Self, Error<Id>>
>```
>If `length` is `None`, we retrieve the parent's branch and calculate the `length`. If the parent's branch is not found, we return `ParentMissing`. 
>If `length` is `Some`, we skip retrieving the parent's branch because we don't need to calculate the `length`. 
>We do `tips.remove(&parent)` in both cases, even if the `parent` is not in the `tips`. 
>
>This means we're allowing our data structure to contain a partial block tree that may not be fully connected to the genesis.
>In the near future, we’ll need to ensure that any missing parts of the data structure actually exist in storage. I’m not sure how to implement this yet, but we definitely need this check—otherwise, fork choice will fail if any block is missing its parent in memory. For example, when backtracking branches, the fork choice should query storage as needed.
> - @youngjoon-lee 

In my opinion, (and vastly simplifying) a parameter that behaves like a toggle is hiding two functions inside one where there should be two (unless that function is a dispatcher).

All in all, what I mean is I feel compositions is much more clear in this specific instance:
```rust
fn unchecked_function() {}

fn checked_function() {
    do_checks();
    unchecked_function()
}
```

danielSanchezQ:
@youngjoon-lee refering to this: https://github.com/logos-co/nomos/pull/1234#discussion_r2042001581 (I cant reply do not know why).
I think those changes are fair. But, IMO, out of countext atm. They are not necessary for recovery per se. As if the node reaches a state, all previous state should be valid (you trust yourself). If something happens means that you ended up in a corrupted state (hence recovery should fail entirely and should trigger sync).
Now, we can bubble an issue and that can be tackled as part of sync itself, with the proper context. wdyt? Am I missing anything?

nomos-services/cryptarchia-consensus/src/lib.rs
danielSanchezQ: > Right, but I assume this is the function that only abstracts this specific validation. I created a higher level `validate_received_block` to group all validations (which is the one called from within `process_block`). Should I still revert it?

Ah, I missed that. That should be ok then.

danielSanchezQ:
> @youngjoon-lee refering to this: [#1234 (comment)](https://github.com/logos-co/nomos/pull/1234#discussion_r2042001581) (I cant reply do not know why). I think those changes are fair. But, IMO, out of countext atm. They are not necessary for recovery per se. As if the node reaches a state, all previous state should be valid (you trust yourself). If something happens means that you ended up in a corrupted state (hence recovery should fail entirely and should trigger sync). Now, we can bubble an issue and that can be tackled as part of sync itself, with the proper context. wdyt? Am I missing anything?

Lets wait for @youngjoon-lee to finalize this review @strinnityk

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -70,30 +70,71 @@ where
...
+        slot: Slot,
+        chain_length: u64,
+    ) -> Self {
         let mut branches = self.branches.clone();
+
         let mut tips = self.tips.clone();
-        // if the parent was the head of a branch, remove it as it has been superseded
-        // by the new header
+        // When running this method by itself (not through `Self::apply_header`), the
+        // parent should not in the tips set. That being said, having this call
```
danielSanchezQ, line 103: should not **be**? I think you missed something there. It actually do not matter it is or not for this method. Should be check outside right?

PR #1236, opened 2025-04-11
feat(sync): add core algorithm

youngjoon-lee:

- Implements the core algorithm of the chain sync protocol. Checkpoint sync is not implemented yet.

[Spec](https://www.notion.so/Cryptarchia-v1-Chain-Synchronization-Protocol-1a48f96fb65c80468338dd1a8b26a14d?pvs=4#1ad8f96fb65c80dc8b8fff3d64701997) & [Executable spec](https://github.com/logos-co/nomos-specs/blob/master/cryptarchia/sync.py).

This PR focuses on the core algorithm by abstracting all other parts (Cryptarchia, Network, etc.). The abstracted parts will be fully implemented in the upcoming PRs.

@youngjoon-lee 

Yes

Yes

## Checklist

> [!WARNING]  

* [x] 1. Description added.

### DIFF
--- a/consensus/cryptarchia-sync/core/src/sync.rs
+++ b/consensus/cryptarchia-sync/core/src/sync.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{HashMap, HashSet, VecDeque},
+    collections::{HashMap, HashSet},
     error::Error,
 };
 
@@ -174,17 +174,9 @@ where
     }
 
     /// Fetch the fork until it reaches the local block tree.
-    async fn find_missing_part<'a>(
-        &self,
-        mut fork: BoxStream<'a, Self::Block>,
-    ) -> VecDeque<Self::Block> {
-        let mut suffix = VecDeque::new();
-        while let Some(block) = fork.next().await {
-            if self.has_block(&block.id()) {
-                break;
-            }
-            suffix.push_front(block);
-        }
+    async fn find_missing_part<'a>(&self, fork: BoxStream<'a, Self::Block>) -> Vec<Self::Block> {
+        let mut suffix = fork.collect::<Vec<_>>().await;
+        suffix.reverse();
         suffix
     }
 }


danielSanchezQ:
> > Oh, thanks. I forgot and overlooked it. I will change it right now.
> 
> @danielSanchezQ I refactored the structure with applying the pattern you suggested. Previously, we had the `Synchronization` struct, which doesn't have any field but has a static function that accepts `CryptarchiaAdapter` and `BlockFetcher` to start a sync process. That structure worked but I felt it's not natural. So now, I refactored the structure as below:
> 
> * `CryptarchiaSync` trait, which has the `process_block` function that must be implemented by a user.
> * `CryptarchiaSyncExt` trait, which has the default `start_sync` function and is automatically implemented for all types that implement `CryptarchiaSync`.
> 
> Using this structure, a user can simply initialize a struct, which implements the `CryptarchiaSync` trait, and call the `start_sync` function directly on the struct. You can also find a integration code in this PR.
> 
> The `CryptarchiaSyncExt::start_sync` consumes the input structs to mutate them during the entire sync process, and returns them if succeeded. If failed, the structs are lost. The user must pass cloned structs if they want to retain access to the original structs in case the sync process fails. Instead of making `CryptarchiaSyncExt::start_sync` take references and cloning internally, I leave the cloning up to the user. This way, `CryptarchiaSyncExt` doesn’t impose a `Clone` bound on the structs.

This sounds good. I'll take a look. Thanks!

danielSanchezQ (approved):
I think this is a better approach. Pretty clean with nice testing. Thanks!

nomos-services/cryptarchia-consensus/src/sync.rs
```diff
@@ -0,0 +1,93 @@
...
+
+        let header = block.header();
+        match self.cryptarchia.try_apply_header(header) {
+            Ok(cryptarchia) => {
+                self.leader.follow_chain(
+                    header.parent(),
+                    header.id(),
+                    header.leader_proof().nullifier(),
+                );
+                let msg = <StorageMsg<_>>::new_store_message(header.id(), block);
```
andrussal, line 71: Is storing new index(blocks by slot) left intentionally out from this PR?

consensus/cryptarchia-sync/core/src/sync.rs
```diff
@@ -0,0 +1,214 @@
...
+                continue;
+            }
+
+            let slot = block.slot();
+            match self.process_block(block).await {
+                Ok(()) => {
+                    debug!("Processed block {id:?}");
+                    orphan_blocks.remove(&id);
+                }
+                Err(CryptarchiaSyncError::ParentNotFound) => {
```
zeegomo, line 123: How it it possible that when iterating blocks in order you're missing the parent? this should only happen if the starting blocks are not the same, would it be better to handle it separately?

consensus/cryptarchia-sync/core/src/sync.rs
```diff
@@ -0,0 +1,214 @@
...
+        let mut rejected_blocks = HashSet::new();
+        // Run a forward sync, and trigger backfillings if orphan blocks are found.
+        // Repeat the forward sync if any backfilling was performed,
+        // in order to catch up with the updated peers.
+        loop {
+            let ForwardSyncOutput {
+                max_fetched_slot,
+                orphan_blocks,
+            } = self
+                .sync_forward(start_slot, block_fetcher, &mut rejected_blocks)
```
zeegomo, line 38: what would happen if the first block has a missing parent and even though you fetch all the block (i.e. max_start_slot is >> start_slot) you didn't successfully apply any of those?

zeegomo (changes requested):

PR #1244, opened 2025-04-14
refactor(clients): Remove unneeded node clients deps

zeegomo:
Move common api paths and types into a separate crate to reuse the same code across node and executor clients and binaries while reducing deps sharing.
Avoids pulling all node dependencies into the client.

Yes

@zeegomo 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 1. Description added.


Cargo.toml
```diff
@@ -52,10 +53,10 @@ resolver = "2"
 bundler                  = { path = "./nomos-bundler" }
 cfgsync                  = { path = "./testnet/cfgsync" }
 cl                       = { path = "./nomos-core/cl" }
-common-http-client       = { path = "./clients/common-http-client" }
+common-http-client       = { path = "./nodes/nomos-node/client" }
```
ntn-x2, line 56: Why is `common-http-client` inside the `nomos-node` directory now? Isn't it better to put it perhaps still inside the `nodes` folder but in a separate `common` directory?

ntn-x2: Unless what this really should be is `validator-http-client`?

Cargo.toml
```diff
@@ -1,12 +1,13 @@
...
 exclude = ["nomos-core/risc0_proofs"]
 members = [
-  "clients/common-http-client",
-  "clients/executor-http-client",
   "consensus/cryptarchia-engine",
   "ledger/nomos-ledger",
-  "nodes/nomos-executor",
-  "nodes/nomos-node",
+  "nodes/api_common",
+  "nodes/nomos-executor/client",
```
ntn-x2, line 7: Is this better as `http-client`? Also for the `nomos-node` two lines below?

ntn-x2 (commented):
I'm personally just a bit confused about some directory names and some crate names. Otherwise this is a good and much needed change, although this should be all unified under the same binary in the not-so-distant future, right @danielSanchezQ?

danielSanchezQ:
> I'm personally just a bit confused about some directory names and some crate names. Otherwise this is a good and much needed change, although this should be all unified under the same binary in the not-so-distant future, right @danielSanchezQ?

Yes, both binaries are gonna be merged at some point in the near future. 
This was a needed refactor anyway, as I think it will be used later from other crates.

Cargo.toml
zeegomo: I understood it to be essentially the `validator-http-client`, but we can change this otherwise

nodes/api_common/Cargo.toml
danielSanchezQ, line 1: nitpick: `api-common`? I think most of our folders are kebab-case

ntn-x2 (approved):

PR #1250, opened 2025-04-16
chore: update tauri builder version to match the CI 

pradovic:
The compile breaks on the CI because the tauri builder version is automatically updates. The make the code compile and work both locally and on CI, one of the possible solution was to update the version


bacv (approved):
👍

PR #1251, opened 2025-04-16
Use async where possible

danielSanchezQ:
Refactor behaviour to remove unnecessary manual polling on while loops

### DIFF
--- a/consensus/cryptarchia-sync/network/src/behaviour.rs
+++ b/consensus/cryptarchia-sync/network/src/behaviour.rs
@@ -2,16 +2,16 @@ use std::task::{Context, Poll};
 
 use cryptarchia_engine::Slot;
 use futures::{
-    AsyncWriteExt, FutureExt,
     future::BoxFuture,
     stream::{FuturesUnordered, StreamExt},
+    AsyncWriteExt, FutureExt,
 };
 use libp2p::{
-    Multiaddr, PeerId, Stream, StreamProtocol,
     swarm::{
         ConnectionClosed, ConnectionId, FromSwarm, NetworkBehaviour, THandler, THandlerInEvent,
         THandlerOutEvent, ToSwarm,
     },
+    Multiaddr, PeerId, Stream, StreamProtocol,
 };
 use tokio::sync::mpsc::{self, Sender, UnboundedSender};
 use tokio_stream::wrappers::{ReceiverStream, UnboundedReceiverStream};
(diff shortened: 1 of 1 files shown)


consensus/cryptarchia-sync/network/src/behaviour.rs
```diff
@@ -326,10 +316,9 @@ impl NetworkBehaviour for SyncBehaviour {
         &mut self,
         cx: &mut Context<'_>,
     ) -> Poll<ToSwarm<<Self as NetworkBehaviour>::ToSwarm, THandlerInEvent<Self>>> {
-        if let Some(event) = self.handle_outgoing_syncs(cx) {
-            return Poll::Ready(event);
-        }
-        if let Some(event) = self.handle_incoming_syncs(cx) {
+        let _ = self.handle_outgoing_syncs().poll_next_unpin(cx);
+        let _ = self.await_close_streams().poll_next_unpin(cx);
```
bacv, line 320: `await_close_streams` is called here and immediately in the `handle_incoming_syncs`, do we want to close one or all streams, `close_streams` is `FuturesUnordered`, this seems bit confusing.

andrussal: I am trying to finish the refactoring that Daniel helped to start.

bacv: No problem, will focus on the main PR.

andrussal: I'm starting to think this is trickier than it seemed.

Previously, I was polling futures that are part of the Behaviour state:

- `sync_commands_receiver`
- `local_sync_progress`
- `read_sync_requests`
- `sending_data_to_peers`
- `closing_streams`

The issue is that calling async functions directly creates a new future each time they're invoked in poll, and they don’t actually make progress that way.

Initially, I thought I could push more logic into async functions to reduce the number of places using low-level `cx: &mut Context<'_>.` But now I'm thinking the original separation of polling logic might’ve been the right call after all. 

@danielSanchezQ I am not sure what is the right thing to do tbh.

danielSanchezQ: Ah yeah, that is true. Slipped my mind. Ok. Then probably the best way is to keep what we had. But remove the `while` loops and just poll a single item each time. wdyt?

andrussal: Sounds good! I think I will work on removing while loops on the original branch.

PR #1254, opened 2025-04-22
fix(DA): Padded polynomial to next power of two size in encode process

danielSanchezQ:

Fix bug where non-padded polynomial would panic encoding

Simple, yes.

@danielSanchezQ 

N/A

N/A

## Checklist

* [ ] 1. Description added.

### DIFF
--- a/nomos-da/kzgrs-backend/src/encoder.rs
+++ b/nomos-da/kzgrs-backend/src/encoder.rs
@@ -1,3 +1,5 @@
+use std::ops::Div;
+
 use ark_ff::{BigInteger, Field, PrimeField};
 use ark_poly::{EvaluationDomain, Polynomial as _};
 use kzgrs::{
@@ -10,7 +12,6 @@ use kzgrs::{
 };
 #[cfg(feature = "parallel")]
 use rayon::iter::{IntoParallelRefIterator, ParallelIterator};
-use std::ops::Div;
 
 use crate::{
     common::{hash_commitment, share::DaShare, Chunk, ChunksMatrix, Row},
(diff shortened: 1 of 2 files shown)


bacv (approved):

bacv:
These changes are not applicable anymore?

danielSanchezQ:
> These changes are not applicable anymore?

I think they are not really neeeded. I was thinking keeping the test just for the memory of the holy split week.

bacv:
> test just for the memory of the holy split week.

You are right, that test is basically essential.

danielSanchezQ:
I'm closing this PR. Tested locally the fix is not necessary. Will ad the test in a different PR.

bacv:
The problematic case if I remember correctly had 1/3 of payload filled with zeroes in a middle [data ... 000 ... data]. Depending on the subnetwork number (which is column number), some chunks would be just zeroes. Looks like we need this PR even in the v1.5, lets reopen this?

bacv:
https://github.com/logos-co/nomos/pull/1255/files#diff-cb8eca32a71e5a6216f6eede098f92af6424fed0a0821bd483de8e8000b77306R258 This is the actual payload, lets reopen this with unit test for all zeroes and maybe lets add integration test from the linked pr.

danielSanchezQ:
The issue now is in encoding I think. So not same issue. I think when all zeroes the polynomial is zero sized. In this case instead of padding I think is smarter to return zero proofs during the computation. But I need to check with @thomaslavaur and @megonen

megonen:
> The issue now is in encoding I think. So not same issue. I think when all zeroes the polynomial is zero sized. In this case instead of padding I think is smarter to return zero proofs during the computation. But I need to check with @thomaslavaur and @megonen

By 'zero proofs,' do you mean the proof of f(0)? In the example, the zeros correspond to a single row, right? In that case, the corresponding row polynomial seems to become f(x)=0. That would mean the proof always ends up being the identity element. Normally, the code shouldn't throw an error when computing this, as I recall, you wrote it generically. But is the issue happening because something like len(poly) is being used?

PR #1256, opened 2025-04-23
feat(api): return blob id in disperse call

pradovic:
The disperse call for the da is not returning blob id. This information is useful to track blob ids when interacting with da.

Yes

@pradovic 

Yes

There is a change in da dispersal api return object to include blob id.

## Checklist

> [!WARNING]  

* [x] 1. Description added.

### DIFF
--- a/nodes/nomos-executor/executor/src/api/handlers.rs
+++ b/nodes/nomos-executor/executor/src/api/handlers.rs
@@ -12,7 +12,7 @@ use nomos_http_api_common::{paths, types::DispersalRequest};
 use nomos_libp2p::PeerId;
 use nomos_node::make_request_and_return_response;
 use overwatch::{overwatch::handle::OverwatchHandle, services::AsServiceId};
-use serde::de::DeserializeOwned;
+use serde::{de::DeserializeOwned, Serialize};
 use subnetworks_assignations::MembershipHandler;
 
 #[utoipa::path(
@@ -49,6 +49,7 @@ where
         + Sync
         + 'static,
     Backend::Settings: Clone + Send + Sync,
+    Backend::BlobId: Serialize,
     NetworkAdapter: DispersalNetworkAdapter<SubnetworkId = Membership::NetworkId> + Send,
     MempoolAdapter: DaMempoolAdapter,
     Metadata: DeserializeOwned + metadata::Metadata + Debug + Send + 'static,
--- a/nodes/nomos-executor/executor/src/api/handlers.rs
+++ b/nodes/nomos-executor/executor/src/api/handlers.rs
@@ -46,7 +46,8 @@ where
             MempoolAdapter = MempoolAdapter,
             Metadata = Metadata,
         > + Send
-        + Sync,
+        + Sync
+        + 'static,
     Backend::Settings: Clone + Send + Sync,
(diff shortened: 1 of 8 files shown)


nodes/nomos-executor/executor/src/api/handlers.rs
```diff
@@ -46,7 +46,8 @@ where
             MempoolAdapter = MempoolAdapter,
             Metadata = Metadata,
         > + Send
-        + Sync,
+        + Sync
+        + 'static,
```
danielSanchezQ, line 50: Why as this added?

nomos-services/data-availability/dispersal/src/backend/mod.rs
```diff
@@ -17,7 +18,7 @@ pub trait DispersalBackend {
     type NetworkAdapter: DispersalNetworkAdapter;
     type MempoolAdapter: DaMempoolAdapter;
     type Metadata: Debug + metadata::Metadata + Send;
-    type BlobId: AsRef<[u8]> + Send;
+    type BlobId: AsRef<[u8]> + Send + Copy + Serialize;
```
danielSanchezQ, line 21: This probably can be moved where we actually use it. In the implementation we do not use Copy or Serialize. Probably the other constrains could be removed as well (?)

nomos-services/data-availability/dispersal/src/lib.rs
```diff
@@ -88,7 +88,7 @@ where
     type Settings = DispersalServiceSettings<Backend::Settings>;
     type State = NoState<Self::Settings>;
     type StateOperator = NoOperator<Self::State>;
-    type Message = DaDispersalMsg<Metadata>;
+    type Message = DaDispersalMsg<Metadata, Backend>;
```
danielSanchezQ, line 91: In this `impl` block is where the constrains should be.

nomos-services/data-availability/dispersal/src/backend/mod.rs
pradovic: I have added it as the api needs to serialize return type that is dependent on the backend. The point was to return BlobId in the disperse call. I tried several approaches but it looks like that api needs to be able to serialize the response. Is there a better way to do this that I am missing?

nomos-services/data-availability/dispersal/src/lib.rs
pradovic: i will try this approach

nodes/nomos-executor/executor/src/api/handlers.rs
pradovic: The Msg started to be generic over backend. I can try to remove this as well, but the opaque service handled demanded the implementation of backend with enough lifetime

nomos-services/data-availability/dispersal/src/lib.rs
```diff
@@ -26,11 +26,11 @@ pub mod adapters;
 pub mod backend;
 
 #[derive(Debug)]
-pub enum DaDispersalMsg<Metadata> {
+pub enum DaDispersalMsg<Metadata, B: DispersalBackend> {
```
bacv, line 29: Could this be generic over `BlobId` instead of `DispersalBackend`?

PR #1259, opened 2025-04-24
feat(da): DA v1.1

danielSanchezQ:

Implemented aggregated proving for DA encoding.  This schema is a much lightweight design. Encoding times are improved greatly while maintaining verification ones.

Yes. In fact is a much simpler process, a lot of code is removed.

@danielSanchezQ @thomaslavaur @megonen 

~~No, this implies changing both encoding and verification specs and remains to be done.~~
Specification was reviwed and adapted to the new encoding schema.

Yes. But it is in sync now.

## Checklist

* [x] 1. Description added.


nomos-da/kzgrs/src/bdfg_proving.rs
```diff
@@ -0,0 +1,229 @@
...
+/// This function will panic if:
+/// - The hasher fails to be constructed.
+/// - Any commitment fails to serialize properly.
+/// - The hash finalization process fails.
+#[must_use]
+pub fn generate_row_commitments_hash(commitments: &[Commitment]) -> Vec<u8> {
+    let mut hasher = Blake2bVar::new(32).expect("Hasher should be able to build");
+    for c in commitments {
+        let mut buffer = Cursor::new(Vec::new());
+        c.serialize_uncompressed(&mut buffer)
```
thomaslavaur, line 46: Should we add a DST to the hash function for the Fiat-Shamir transform?

nomos-da/kzgrs/src/bdfg_proving.rs
```diff
@@ -0,0 +1,229 @@
...
+    column_idx: usize,
+    column: &[Fr],
+    row_commitments: &[Commitment],
+    column_proof: &Proof,
+    domain: PolynomialEvaluationDomain,
+    global_parameters: &GlobalParameters,
+) -> bool {
+    let row_commitments_hash = generate_row_commitments_hash(row_commitments);
+    let h = Fr::from_le_bytes_mod_order(&row_commitments_hash);
+    let aggregated_elements: Fr = column
```
thomaslavaur, line 210: We can gain a little bit of performance here by computing the powers of h before (to avoid computing it twice)

thomaslavaur (approved):
LGTM = Let's go to the Moon!!

nomos-da/kzgrs/src/bdfg_proving.rs
danielSanchezQ: I'm not sure how it is relate to fiat shamir actually. But it makes sense in case we upgrade protocols. Will do.

nomos-da/kzgrs/src/bdfg_proving.rs
danielSanchezQ: Yes, forgot to optimize that. Thanks for the remainder!

nomos-da/kzgrs/src/bdfg_proving.rs
```diff
@@ -0,0 +1,248 @@
...
+///
+/// This function will panic if:
+/// - The hasher fails to be constructed.
+/// - Any commitment fails to serialize properly.
+/// - The hash finalization process fails.
+#[must_use]
+pub fn generate_row_commitments_hash(commitments: &[Commitment]) -> Vec<u8> {
+    let mut hasher = Blake2bVar::new(32).expect("Hasher should be able to build");
+    // add dst for hashing
+    hasher.update(b"NOMOS_DA_V1");
```
danielSanchezQ, line 45: This needs to be added to specs

bacv (approved):

nomos-da/kzgrs/src/bdfg_proving.rs
megonen: Blake is used in another hashing step to convert the data to 31 bytes. Even though the hash input sizes are different in both cases, using a DST might be a good idea for edge cases (in case of 1 row commitment calculation)

nomos-da/network/core/src/protocols/replication/mod.rs
```diff
@@ -171,6 +171,7 @@ mod test {
     }
 
     #[tokio::test]
+    #[ignore = "Create new fixtures"]
```
danielSanchezQ, line 174: @CHr15F0x , can you generate the new fixtures? Do we have a place which explains how to actually do this?

nomos-da/kzgrs/src/bdfg_proving.rs
danielSanchezQ: > Blake is used in another hashing step to convert the data to 31 bytes. Even though the hash input sizes are different in both cases, using a DST might be a good idea for edge cases (in case of 1 row commitment calculation)

Uhm, it was used for the column hashing, but we do not do that anymore.  I added DST already. Or where you referring to somewhere else?

danielSanchezQ:
@megonen @thomaslavaur , ping when the specs are in sync with this implementation. Otherwise this PR is stalled.

nomos-da/kzgrs/src/bdfg_proving.rs
megonen: Ahh right, it was being used there and no needed anymore. So in that case, Blake will only be used here. When you’re using a hash for just one purpose, there’s actually no need for a DST. What do you think? @thomaslavaur. Since it’s already implemented, we might as well leave it as is

megonen:
> ping when the specs are in sync with this implementation. Otherwise this PR is stalled.

@danielSanchezQ The spec is almost ready. It’s aligned with the implementation. The only difference is in naming: since the new commitments are generated using random linear combinations, they’re referred to as combined polynomial/commitment/proof in the spec. That seemed more appropriate than aggregated.
Should we keep the term aggregated in the implementation, or maybe add a note in the spec saying that aggregated and combined are used interchangeably?

megonen (approved):

danielSanchezQ:
> > ping when the specs are in sync with this implementation. Otherwise this PR is stalled.
> 
> @danielSanchezQ The spec is almost ready. It’s aligned with the implementation. The only difference is in naming: since the new commitments are generated using random linear combinations, they’re referred to as combined polynomial/commitment/proof in the spec. That seemed more appropriate than aggregated. Should we keep the term aggregated in the implementation, or maybe add a note in the spec saying that aggregated and combined are used interchangeably?

I'll change the nomenclature to combined. This must be reflected in any other spec as well, is that the case?

megonen:
> I'll change the nomenclature to combined. This must be reflected in any other spec as well, is that the case?

Yeah, I will also update them 👍

PR #1261, opened 2025-05-06
feat: fork pruning in cryptarchia

ntn-x2:

WIP.

Yes.

@ntn-x2 

N/A

No.

## Checklist

> [!WARNING]  

* [x] 1. Description added.


ntn-x2:
@danielSanchezQ and @strinnityk one more thing that is not clear to me is what would need to be stored in persistent storage? As of now, I have not touched the cryptarchia state mechanism, and I would assume whatever is getting stored is already sufficient for the node to restart again? Pruning all old forks (or only the ones that can be deleted according to the density checks as we are discussing in Discord) would not affect that at all, since those forks are most likely useless anyway.
I feel that deciding how to store the "finalized" history on persistent storage is out of scope for this work, and should be carefully analysed depending on the needs of the different use cases that @youngjoon-lee and @andrussal are working on.

strinnityk:
> @danielSanchezQ and @strinnityk one more thing that is not clear to me is what would need to be stored in persistent storage? As of now, I have not touched the cryptarchia state mechanism, and I would assume whatever is getting stored is already sufficient for the node to restart again? Pruning all old forks (or only the ones that can be deleted according to the density checks as we are discussing in Discord) would not affect that at all, since those forks are most likely useless anyway. I feel that deciding how to store the "finalized" history on persistent storage is out of scope for this work, and should be carefully analysed depending on the needs of the different use cases that @youngjoon-lee and @andrussal are working on.

I don't think this affects recovery in any way.

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -252,17 +276,86 @@ where
...
-    pub fn prune_forks(&mut self, _depth: u64) {
-        todo!()
+    pub fn non_canonical_forks(&self) -> impl Iterator<Item = Branch<Id>> + '_ {
+        self.branches
+            .branches()
+            .filter(|fork_tip| fork_tip.id != self.tip())
+    }
+
+    /// Prune all states strictly deeper than 'depth' with regard to the
+    /// current local chain except for states belonging to the local chain.
```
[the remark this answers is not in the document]
ntn-x2: Sure, makes sense. Did not want to change anything that was previously written and was counting on reviewers to point things like this. I will address this, thanks.

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -643,8 +658,19 @@ where
...
-                                blend_adapter.blend(block).await;
+                            if let Some(new_block) = block {
+                                cryptarchia = Self::process_block_unchecked(
+                                    cryptarchia,
+                                    &mut leader,
+                                    new_block.clone(),
+                                    &relays,
+                                    &mut self.block_subscription_sender,
+                                )
+                                .await;
```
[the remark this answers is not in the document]
ntn-x2: Yes, that is the reason. No need to perform full validation cycle since we are applying the block locally before gossiping to other peers.

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -181,6 +183,18 @@ impl Cryptarchia {
...
+
+    pub fn prune_old_forks(&mut self) {
+        let old_forks_pruned = self.consensus.prune_forks(
+            self.ledger
+                .config()
+                .consensus_config
+                .security_param
+                .get()
+                .into(),
+        );
```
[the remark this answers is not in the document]
ntn-x2: Mhh ok, that would make sense. I will look into it.

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -180,6 +182,39 @@ impl Cryptarchia {
...
+        let mut pruned_states_count = 0usize;
+        // We need to iterate once, then return the unconsumed iterator, so we need to
+        // collect first.
+        let collected_forks = old_forks_pruned.collect::<Vec<_>>();
```
zeegomo, line 202: why not collecting right away?

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -1369,8 +1410,55 @@ where
...
+        .await;
+
+        for old_fork_pruned in old_forks_pruned {
+            let block_id = old_fork_pruned.id();
+            if !leader.prune_notes_at(&block_id) {
+                tracing::warn!(
+                    target: LOG_TARGET,
+                    "Leader proof for block {block_id} was not found in storage."
+                );
+            }
```
zeegomo, line 1444: I thikn this might happen if you haven't set any leader note at all

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -252,17 +276,83 @@ where
...
+    ///
+    /// It returns the blocks that were part of the pruned forks.
+    pub fn prune_forks(&mut self, depth: u64) -> impl Iterator<Item = Branch<Id>> {
+        let local_chain = self.local_chain;
+        let Some(target_height) = local_chain.length.checked_sub(depth) else {
+            tracing::info!(
+                target: LOG_TARGET,
+                "No pruning needed, the canonical chain is not longer than the provided depth. Canonical chain length: {}, provided depth: {}", local_chain.length, depth
+            );
+            return vec![].into_iter();
```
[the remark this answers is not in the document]
ntn-x2: How? The `impl Trait` requires all returns to have the same type, and unfortunately if we return an iterator over a vector down below, we need to return an iterator over an (empty) vector here, unless we go with the `Box<dyn>` approach, which I don't think we need here.

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -252,17 +276,83 @@ where
...
+        for (fork, lca) in &non_canonical_forks_older_than_depth {
+            removed_blocks.extend(self.prune_fork(fork.id, lca.id));
+        }
+        removed_blocks.into_iter()
+    }
+
+    /// Remove the list of blocks from `tip` to and excluding `up_to`, returning
+    /// them in the same order they were encountered while traversing blocks
+    /// from `tip` to `up_to`.
+    fn prune_fork(&mut self, tip: Id, up_to: Id) -> impl Iterator<Item = Branch<Id>> {
```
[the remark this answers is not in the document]
ntn-x2: This is what you are suggesting? https://github.com/logos-co/nomos/pull/1261/commits/01cdbc4126552fd639066e2473ec30b037a42d14. I personally don't think it makes it any easier, but I am not against this way.

consensus/cryptarchia-engine/src/lib.rs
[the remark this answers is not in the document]
ntn-x2: > mismatched types `std::vec::IntoIter<Branch<Id>>` and `std::array::IntoIter<_, 0>` have similar names, but are actually distinct types

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -1283,6 +1321,9 @@ where
             .await;
         }
 
+        // Once the state has been recovered, we get rid of the unnecessary forks.
+        Self::prune_old_forks(&mut cryptarchia, &mut leader, relays.storage_adapter()).await;
```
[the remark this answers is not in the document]
zeegomo: This is not getting new blocks, just loading them from storage to the point where the node was previously, so I don't see how it could get more forks than it had previously. If we don't see a reason for that either we put a comment on it saying it's redundant or I'd remove it, which I think simplifies the code

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -1369,8 +1410,55 @@ where
...
+    ) {
+        let old_forks_pruned = cryptarchia.prune_old_forks().collect::<Vec<_>>();
+        // Trigger parallel block deletion and keep a reference to each block ID for
+        // logging purposes (below).
+        let block_deletion_outcomes = join_all(
+            old_forks_pruned
+                .iter()
+                .map(|f| async move { (f.id(), storage_adapter.remove_block(f.id()).await) }),
+        )
+        .await;
```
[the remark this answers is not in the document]
zeegomo: I think this is more for debugging right? You don't actually do anything except for logging those.  What about returning the list of either missing or deleted blocks?

consensus/cryptarchia-engine/src/lib.rs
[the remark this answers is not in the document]
zeegomo: my bad,  I was thinking about a different thing, the version you have now is ok 👍🏼

nomos-services/cryptarchia-consensus/src/lib.rs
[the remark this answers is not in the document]
ntn-x2: No what I mean is that if we don't put it in the recovery operations, the only other time when pruning would be called (as the PR currently stands) is when a block is received and added to the local chain or when the node itself produces a new block. While it makes sense to run pruning after each of those operations, I think for the case in which a node has been away for a while and comes back online with a few forks, it would make sense, in my opinion, to run the "cleanup" also at the end of the catchup process, without waiting for the next block to be added to the local chain. Is my thinking clearer now?

nomos-services/cryptarchia-consensus/src/lib.rs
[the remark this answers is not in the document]
ntn-x2: I guess the idea was more that you could do something with the returned blocks from storage, since you are giving only the block ID as input. I think we can go with returning an iterator that returns `Option<Block>` and then for now we only log the `None`s. Let me give it a try.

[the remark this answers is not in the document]
andrussal: Maybe it's a naive thought but can we actually have blocks in storage that doesn't have a parent? I mean if we have a fork, it should be a valid in the sense that it forms correct chain of blocks? And in this case we just need last block and first block to remove the whole fork by following parent links?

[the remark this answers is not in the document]
ntn-x2: Well it can happen you get a block for which you don't have the parent, which would then kickoff the catchup procedure @youngjoon-lee explained at the offsite. And until the procedure is complete, if we progress the canonical chain and prune an old fork, we could still have blocks that do not have a parent, if my understanding is correct.

consensus/cryptarchia-engine/src/lib.rs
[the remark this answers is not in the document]
ntn-x2: You suggest we keep the modified version as in https://github.com/logos-co/nomos/commit/01cdbc4126552fd639066e2473ec30b037a42d14 or the original version that takes `from` and `to` block IDs?

nomos-services/cryptarchia-consensus/src/lib.rs
ntn-x2: Addressed in https://github.com/logos-co/nomos/pull/1261/commits/679f95d758e2493e8563b7608f10f92775d656a8. We debug log only if we delete a note, which makes more sense.

nomos-services/cryptarchia-consensus/src/lib.rs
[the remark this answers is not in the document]
andrussal: It might be good idea to wrap deleting a fork into database transaction. If node stops in the middle of deleting a fork, will it stay in consistent state? I haven't looked closely but do we move `Cryptarchia` atomically to new state from "pre-fork deletion" to "after-fork deletion"?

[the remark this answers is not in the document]
zeegomo: It's a storage for a chain, so I don't think it should be completely agnostic. Have we ever considered an append only storage for the immutable part of the chain? 
As far as I'm concerned, we need a better API for removing a bunch of blocks, either by listing them or by providing the branch (or something else entirely), I trust your judgement in picking the one you like the most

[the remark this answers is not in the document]
danielSanchezQ: > It might be good idea to wrap deleting a fork into database transaction. If node stops in the middle of deleting a fork, will it stay in consistent state? I haven't looked closely but do we move `Cryptarchia` atomically to new state from "pre-fork deletion" to "after-fork deletion"?

It should, non removed forks would be removed next time the forking happens. If somehow we transform a fork into a subset of orphan blocks those should be removed as well (?)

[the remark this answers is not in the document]
ntn-x2: Yeah but this is a different thing, I feel. If we wait to delete them from the in-memory storage, then even if it's not a DB transaction and the node stops halfway, we will ask the DB again to delete some previously deleted blocks, some of which do not exist, which is not too bad besides a log line. But putting them into a DB transaction still makes sense regardless, although it does not solve alone the problem of inconsistency between in-memory and storage. Nevertheless, cryptarchia recovery mechanism only tracks the tip and the security block header, no forks. That means that already, upon restarting, we don't know which fork we were tracking before. Supporting consistency across restarts would require dumping the whole engine state into storage, which feels like a more attainable task if we keep it in check by pruning everytime.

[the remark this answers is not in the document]
ntn-x2: And we add a DB function to delete a list of blocks as a batch, then the DB is free to choose how to implement it.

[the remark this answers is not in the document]
ntn-x2: I mean only the info about the forks we were supposed to prune but did not manage to because something happened before the operation was completed. So basically only forks (i.e., only block info, no state) that are older than `k`. It does not feel super clean, but it's a possibility indeed.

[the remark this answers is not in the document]
andrussal: >Could we have pruning logic for the DB @andrussal?

Do you mean if I could implement it? I could do it myself as well also.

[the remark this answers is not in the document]
andrussal: It could save initial pruning request for example atomically and start pruning. At startup some background task would check if these requests are complete, if not would retry. If this simplifies `Cryptarchia` engine logic, it could make sense. But adding actual logic to storage find forks to prune, maybe. Depends on how much logic we are willing put here.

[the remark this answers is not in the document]
danielSanchezQ: > Would there be any benefit, going forward, for the storage to provide a `delete_blocks` vs a `delete_forks` function? If there is no benefit in terms of performance, I would rather have a `delete_blocks` and let cryptarchia engine do the heavylifting. Otherwise, we might need to "inject" the storage layer into the cryptarchia engine so that it can rely on what the storage engine returns when pruning blocks, which I feel might complicate things. Another option would be to turn cryptarchia "engine" into a block storage and simply call `delete_blocks` on it instead based on the blocks the storage layer returned as part of pruning. So from my point of view if a function like `delete_blocks` still allow for atomicity, allow for the mechanism you envisioned in the previous comment, and has no tangible performance downsides, I would go with this which is the least destructive way.

Sounds good. Also part of the logic can be really implemented in the storage adapter, doesnt need to be in the storage itself. So keeping track of stuff like that maybe can be dumped there if it makes sense. I'll let you guys decide what you think is better.

ntn-x2:
Moving back into draft until the changes addressing https://github.com/logos-co/nomos/pull/1261#discussion_r2111484779 are included. Please refrain from starting a new review until pinged.

tests/Cargo.toml
```diff
@@ -46,7 +46,7 @@ subnetworks-assignations = { workspace = true }
 tempfile                 = "3"
 time                     = "0.3"
 tokio                    = "1"
-tracing                  = "0.1"
+tracing                  = { workspace = true }
```
danielSanchezQ, line 49: I'ts done now. But ideally this would have been a separated PR better.

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -252,17 +287,100 @@ where
...
+        };
+        self.non_canonical_forks()
+            .filter_map(|fork| {
+                // We calculate LCA once and store it in `ForkInfo` so it can be consumed
+                // elsewhere without the need to re-calculate it.
+                let lca = self.branches.lca(&local_chain, &fork);
+                (lca.length <= target_height).then_some(ForkDivergenceInfo { tip: fork, lca })
+            })
+            .collect::<Vec<_>>()
+            .into_iter()
```
danielSanchezQ: You do not need to collect. Just let the filter_map. In the other branch simply use [`std::iter::empty`](https://doc.rust-lang.org/std/iter/fn.empty.html)

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -252,17 +287,100 @@ where
...
+                break;
+            };
+            removed_blocks.push(branch.id);
+            current_tip = branch.parent;
+        }
+        tracing::debug!(
+            target: LOG_TARGET,
+            "Pruned {} branches from {tip:#?} to {current_tip:#?}.", removed_blocks.len()
+        );
+        removed_blocks.into_iter()
```
danielSanchezQ: Using a vec here is ok probably. But I'm pretty sure we could avoid it (this looks like a reduce....like everything 😂 )

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -252,17 +287,100 @@ where
...
+    /// canonical chain.
+    ///
+    /// It returns the block IDs that were part of the pruned forks.
+    pub fn prune_forks(&mut self, depth: u64) -> impl Iterator<Item = Id> {
+        let prunable_forks = self.prunable_forks(depth);
+        let mut removed_blocks = vec![];
+        for prunable_fork_info in prunable_forks {
+            removed_blocks.extend(self.prune_fork(&prunable_fork_info));
+        }
+        removed_blocks.into_iter()
```
danielSanchezQ: I think the low level is ok to force the collects, but here is not necessary at all
```suggestion
    pub fn prune_forks(&mut self, depth: u64) -> impl Iterator<Item = Id> {
        self.prunable_forks(depth).iter().flat_map(|prunable_fork_info| self.prune_fork(&prunable_fork_info))
    }
```
I know you would have to deal with lifetimes probably. If its too difficult we can revisit this later.

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -1366,8 +1410,123 @@ where
...
+
+        let Ok(()) = Self::delete_pruned_blocks_from_storage(
+            old_forks_pruned.iter().copied(),
+            storage_adapter,
+        )
+        .await
+        else {
+            // Bail out early (not changing the input `cryptarchia` value if the storage
+            // returned an error).
+            return;
```
danielSanchezQ: Should we log something here?

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -1366,8 +1410,123 @@ where
...
+                        tracing::error!(
+                            target: LOG_TARGET,
+                            "Error deleting block {block_id:#?} from storage: {e}."
+                        );
+                        errors.push((block_id, e));
+                    }
+                }
+                errors
+            },
+        );
```
danielSanchezQ: a map+collect would be better. I think in this case the compiler already optimizes when it knows the size.

andrussal:
> @danielSanchezQ @andrussal @zeegomo please check the new HEAD of this branch. I still want to write some tests to make sure the recovery logic works fine, and need to rebase on top of @zeegomo's [PR](https://github.com/logos-co/nomos/pull/1300) (or simply wait for it to be merged). The basic concept remains the same tho: we keep a backup of the initial state before dispatching ops to the storage layer in case there is a failure, and only after that goes through we update the engine and save the new state into the recovery system. It seems the code will look slightly easier to follow once Giacomo's changes are in. What do you guys think?

Looks great! I think it might be more robust if at some point we could remove all blocks in a single database transaction. Otherwise we may remove only some blocks from storage but cryptarchia state would still contains them if some error happens and we "rollback".

andrussal:
> Wouldn't it be the other way around? Cryptarchia ones would be always removed but storage ones not?

It seems like we try to remove blocks from storage and if this fails(in the middle), we return "old" Cryptarchia. But some blocks may be removed from storage and some not.

danielSanchezQ:
> > Wouldn't it be the other way around? Cryptarchia ones would be always removed but storage ones not?
> 
> It seems like we try to remove blocks from storage and if this fails(in the middle), we return "old" Cryptarchia. But some blocks may be removed from storage and some not.

True, was taking a deeper look at that. I think this is enough for now. It should rarely fail, and by pruning consistently this should be fairly cheap.

ntn-x2:
@danielSanchezQ @andrussal if some DB error occurs, cryptarchia will not prune the fork from memory, and the next time it will ask the DB to delete those blocks again, hopefully this time succeeding. Also, the forks-to-be-pruned are stored in the cryptarchia recovery state, so if the node shuts down, the next time it will re-ask to delete those blocks that previously failed. Also, I agree that having deletions in a DB transaction would increase robustness, but from the point of view of the service this is hidden behind the trait implementation, hence it still needs to implement a rollback mechanism for the cryptarchia engine itself.

Anyway, that is the part I wanted the feedback on. I will write some tests and address the comments, then wait for Giacomo's PR to be merged before finalizing this, since it seems to make more sense. Thanks both for your input.

consensus/cryptarchia-engine/src/lib.rs
ntn-x2: They are still considered two different types, and we can't return two different types as `impl Iterator<Item = ForkDivergenceInfo<Id>>`. This would be solved by `Box<dyn Iterator...>` return type, but I did not deem it necessary here.

```
mismatched types
expected struct `std::iter::Empty<ForkDivergenceInfo<Id>>`
   found struct `std::iter::FilterMap<impl std::iter::Iterator<Item = Branch<Id>> + '_
```

This is the reason for most of the `collect()`, to be honest.

consensus/cryptarchia-engine/src/lib.rs
ntn-x2: Mhh we have a `while` condition, and we are removing nodes from the storage map, I could not think of an alternative way of doing this, but I am of course open to suggestions!

consensus/cryptarchia-engine/src/lib.rs
danielSanchezQ: No need to change the return type. You can just box the iterators before returning them. I would rather box them than collect them so we keep the lazy ness of things. We may incur in lifetime issues though. But that would be a different problem.

consensus/cryptarchia-engine/src/lib.rs
ntn-x2: it's not about lifetimes, it's about borrowing both mutably (`self.prunable_forks`) and immutably (`self.prune_fork`) at the same time.

ntn-x2: I gave it a try first: https://github.com/logos-co/nomos/pull/1261/commits/6b3d08bd871fe04571efe0f2f3aac961edde3a05, but then had to re-introduce at least the `collect` part here: https://github.com/logos-co/nomos/pull/1261/commits/4519b7e428101062fbb7f0f7cf40f3d9bc2e0a07.

nomos-services/cryptarchia-consensus/src/lib.rs
ntn-x2: I went for a `filter_map`, which felt cleaner: https://github.com/logos-co/nomos/pull/1261/commits/53a56eb1aeff44077e1fe438dbf47cd872ba9dd8. Thanks for the tip tho!

nomos-services/cryptarchia-consensus/src/lib.rs
ntn-x2: Yuuup: https://github.com/logos-co/nomos/pull/1261/commits/3d92e3bc784c48590d08f1794be87951e7dff4a2

consensus/cryptarchia-engine/src/lib.rs
ntn-x2: I did need to change return type to `Box<dyn>` in here: https://github.com/logos-co/nomos/pull/1261/commits/4519b7e428101062fbb7f0f7cf40f3d9bc2e0a07. Otherwise there is no way we can return the same thing in both branches without using tools like `itertools`.

consensus/cryptarchia-engine/src/lib.rs
danielSanchezQ: > Mhh we have a `while` condition, and we are removing nodes from the storage map, I could not think of an alternative way of doing this, but I am of course open to suggestions!

Nah, it would be too much work to fit everything and its more readable this way. Hopefully in some not so far future we get the generators syntax (which could help on this kind of things I think).

consensus/cryptarchia-engine/src/lib.rs
danielSanchezQ: Ah yes, sadly there is no workaround for it probably. Thanks for the effort though!

consensus/cryptarchia-engine/src/lib.rs
danielSanchezQ: > I did need to change return type to `Box<dyn>` in here: [4519b7e](https://github.com/logos-co/nomos/commit/4519b7e428101062fbb7f0f7cf40f3d9bc2e0a07). Otherwise there is no way we can return the same thing in both branches without using tools like `itertools`.

You need to cast one of the branches to be `dyn`:

```rust
    /// Get an iterator over the forks that can be pruned given the provided
    /// depth.
    ///
    /// This means that all forks that diverged from the canonical chain before
    /// the provided `depth` height are returned.
    pub fn prunable_forks(&self, depth: u64) -> impl Iterator<Item = ForkDivergenceInfo<Id>> + '_ {
        let local_chain = self.local_chain;
        let Some(target_height) = local_chain.length.checked_sub(depth) else {
            tracing::info!(
                target: LOG_TARGET,
                "No prunable fork, the canonical chain is not longer than the provided depth. Canonical chain length: {}, provided depth: {}", local_chain.length, depth
            );
            return Box::new(std::iter::empty())
                as Box<dyn Iterator<Item = ForkDivergenceInfo<Id>>>;
        };
        Box::new(self.non_canonical_forks().filter_map(move |fork| {
            // We calculate LCA once and store it in `ForkInfo` so it can be consumed
            // elsewhere without the need to re-calculate it.
            let lca = self.branches.lca(&local_chain, &fork);
            (lca.length <= target_height).then_some(ForkDivergenceInfo { tip: fork, lca })
        }))
    }
```

ntn-x2: I learned something new today! Thank you sir, this is very much appreciated! https://github.com/logos-co/nomos/pull/1261/commits/e5530ef5f455e776f2f1827489b7ae040334bfc5

consensus/cryptarchia-engine/src/lib.rs
danielSanchezQ: Ok, I gave it a try, I think this would be the translation if im not mistaken:

```rust
    /// Remove all blocks from `tip` to `lca`, excluding `lca` which belongs to
    /// the canonical chain.
    fn prune_fork(
        &mut self,
        &ForkDivergenceInfo { lca, tip }: &ForkDivergenceInfo<Id>,
    ) -> impl Iterator<Item = Id> + '_ {
        let tip_removed = self.branches.tips.remove(&tip.id);
        if !tip_removed {
            tracing::error!(target: LOG_TARGET, "Fork tip {tip:#?} not found in the set of tips.");
        }

        let current_tip = tip.id;
        let res = (0..)
            .scan(
                (current_tip.clone(), &mut self.branches),
                |(current_tip, branches), _| {
                    branches
                        .branches
                        .remove(&current_tip)
                        .map(move |branch| (branch.parent, branch.id))
                },
            )
            .take_while(move |(current_tip, _)| current_tip != &lca.id)
            .map(|(_, id)| id);
        tracing::debug!(
            target: LOG_TARGET,
            "Pruned branches from {tip:#?} to {current_tip:#?}."
        );
        res
    }
```

The issue is that we capture the `&mut self` and then it breaks in the outer `prune_forks` method. The solution would be to include this and let the `prune_forks` as you had initially, which should be ok since we are collecting there anyway. wdyt?

ntn-x2: This solution gives issues inside `prune_forks` when dealing with

```
forks.into_iter().flat_map(move |prunable_fork_info| self.prune_fork(&prunable_fork_info))`
```

with this error:

```
captured variable cannot escape `FnMut` closure body
`FnMut` closures only have access to their captured variables while they are executing therefore, they cannot allow references to captured variables to escape
```

I would leave the current structure and refactor later on if needed. I also wanted to implement the `Iterator` trait on a bunch of types, but decided it's not worth at the moment, but might be in the future.

danielSanchezQ: > This solution gives issues inside `prune_forks` when dealing with
> 
> ```
> forks.into_iter().flat_map(move |prunable_fork_info| self.prune_fork(&prunable_fork_info))`
> ```
> 
> with this error:
> 
> ```
> captured variable cannot escape `FnMut` closure body
> `FnMut` closures only have access to their captured variables while they are executing therefore, they cannot allow references to captured variables to escape
> ```
> 
> I would leave the current structure and refactor later on if needed. I also wanted to implement the `Iterator` trait on a bunch of types, but decided it's not worth at the moment, but might be in the future.

yeah it was what I mentioned. I agree is probably not worth. The while loop is easier to read as well. Thanks!

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -180,6 +183,42 @@ impl<State: CryptarchiaState> Cryptarchia<State> {
...
+            .consensus
+            .prune_forks(
+                self.ledger
+                    .config()
+                    .consensus_config
+                    .security_param
+                    .get()
+                    .into(),
+            )
+            .collect::<Vec<_>>();
```
youngjoon-lee:
```suggestion
    /// Prune the in-memory storage of forks older than or equal to the LIB.
    ///
    /// The old blocks are removed from the consensus engine and their state
    /// removed from the ledger state.
    pub fn prune_old_forks(&mut self) -> impl Iterator<Item = HeaderId> {
        // We need to iterate once, then return the unconsumed iterator, so we need to
        // collect first.
        let old_forks_pruned = self
            .consensus
            .prune_forks(
                // NOTE: I used `unwrap` here just for simplicity.
                self.consensus.branches().get(&self.consensus.lib()).unwrap().length(), 
            )
            .collect::<Vec<_>>();
```
Now that we defined the [LIB (Latest Immutable Block)](https://www.notion.so/Cryptarchia-v1-Protocol-Specification-7816b67600bb4aad93507723f06ac2e5?source=copy_link#1f58f96fb65c80fea1d4eada76396d18), we should use this for filtering prunable blocks, instead of using the `k`-deep block. 
While the Online fork choice rule is used, the LIB will be always the `k`-deep block. But, if the Bootstrap rule is being used, the LIB may be older than the `k`-deep block. 
Although we haven't implement the fork choice rule transition yet (which will be implemented as part of bootstrapping), it's safer (correct) to use LIB instead of `k` here.

ntn-x2: Mhh ok thanks for the clarification. We are not pruning forks during bootstrapping, which is why I chose `k` as the pruning point. But yeah, I agree it seems safer to do that regardless. Thanks for the input! https://github.com/logos-co/nomos/pull/1261/commits/0a43279da8c305d614966156ee92b898ecef4f98

danielSanchezQ (approved):
Looking good. Any improvement derived from LIB can be a separated PR. Thanks!

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -430,10 +467,93 @@ where
...
+
+    /// Get an iterator over the forks that can be pruned given the provided
+    /// depth.
+    ///
+    /// This means that all forks that diverged from the canonical chain at or
+    /// before the provided `depth` height are returned.
+    pub fn prunable_forks(&self, depth: u64) -> impl Iterator<Item = ForkDivergenceInfo<Id>> + '_ {
+        let local_chain = self.local_chain;
+        let Some(target_height) = local_chain.length.checked_sub(depth) else {
+            tracing::info!(
```
zeegomo: nit: I'd say this is a debug log at most

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -430,10 +467,93 @@ where
...
+                break;
+            };
+            removed_blocks.push(branch.id);
+            current_tip = branch.parent;
+        }
+        tracing::debug!(
+            target: LOG_TARGET,
+            "Pruned {} branches from {tip:#?} to {current_tip:#?}.", removed_blocks.len()
+        );
+        removed_blocks.into_iter()
```
zeegomo: nit: just curious, if this is an internal method (so no API stability worries), why masking the vec that you already have behind an iterator?

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -430,10 +467,93 @@ where
...
+                // If tip is not in branch set, it means this tip was sharing part of its
+                // history with another fork that has already been removed.
+                break;
+            };
+            removed_blocks.push(branch.id);
+            current_tip = branch.parent;
+        }
+        tracing::debug!(
+            target: LOG_TARGET,
+            "Pruned {} branches from {tip:#?} to {current_tip:#?}.", removed_blocks.len()
```
zeegomo: nit: s/branches/block

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -1368,7 +1413,139 @@ where
...
+        // Add prunable blocks, to account for potential DB errors before the node was
+        // shutdown and avoid inconsistent state later on.
+        // We only store the information about the blocks themselves, without any
+        // associated ledger nor leadership-related information.
+        // Each block that has already been deleted by the storage adapter in a previous
+        // iteration will result in a no-op, so it's not a big deal.
+        // Given the right fork choice rules, these forks will fail to extend as they
+        // are past the Last Immutable Block (LIB), hence they will be pruned the next
+        // time the fork choice rule is triggered.
+        for block in prunable_blocks {
```
zeegomo: are these the prunabl blocks from `SecurityRecoveryStrategy`? If so, and we already know this should be pruned, I don't see why we need to add them to cryptarchia

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -1368,7 +1413,139 @@ where
...
+    /// If the interaction with the DB fails, the changes are rolled back, and
+    /// the to-be-pruned forks kept inside the in-memory engine, and a new
+    /// deletion will be performed at the next pruning attempt.
+    async fn prune_old_forks(
+        cryptarchia: &mut Cryptarchia<Online>,
+        leader: &mut Leader,
+        storage_adapter: &StorageAdapter<Storage, TxS::Tx, BS::BlobId, RuntimeServiceId>,
+    ) {
+        // We make a copy of the original state so we don't change the input state
+        // directly. This allows us to rollback in case the storage yields inconsistent
```
zeegomo: that's not really inconsistent state more than inefficient use of resources. If anything, storage not having a block why is needed by cryptarchia would be an inconsistency. 
If this is the reason we're applying prunable blocks to cryptarchia, I'm definitely more worried about using `apply_unchecked` than not deleting something from storage

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -1368,7 +1413,139 @@ where
...
+            .filter_map(|(block_id, outcome)| match outcome {
+                Ok(Some(_)) => {
+                    tracing::debug!(
+                        target: LOG_TARGET,
+                        "Block {block_id:#?} successfully deleted from storage."
+                    );
+                    None
+                }
+                Ok(None) => {
+                    tracing::trace!(
```
zeegomo, line 1519: I'd say this is quite worrying, how could it happen that we miss blocks

nomos-services/cryptarchia-consensus/src/states.rs
```diff
@@ -110,12 +120,44 @@ impl<TxS, BxS, NetworkAdapterSettings, BlendAdapterSettings, TimeBackendSettings
...
+                };
+            cryptarchia
+                .consensus
+                .prunable_forks(pruning_depth)
+                .collect()
+        });
+
+        // Merge all blocks from each prunable fork's tip up until (but excluding) the
+        // fork's LCA with the canonical chain.
+        let mut prunable_blocks = Vec::new();
```
zeegomo: dedup?

nomos-services/cryptarchia-consensus/src/lib.rs
ntn-x2: The problem we tackled here is that we don't want to end up in a situation where we prune forks from the engine, but fail to delete forks from storage. Then those will never be deleted from storage. So what we do here, is that if the storage fails to delete ALL of them for some reason, we revert back and try to delete them again next time. Only once we now none of them returns `Err`, can we prune them also from in-memory storage. This is a workaround for this issue, that we considered the best trade-off in that case. Even implementing the deletion as a transaction with its own rollback logic, we still need to keep cryptarchia in sync with the storage. This was discussed here: https://github.com/logos-co/nomos/pull/1261#discussion_r2109251045.

nomos-services/cryptarchia-consensus/src/lib.rs
ntn-x2: This happens if our pruning request does not fully succeed, and the storage backend implements batch deletion in a non-atomic way (as is the case now). So we account for that case here, and we don't treat them lack of blocks as an issue. Do you think this is a big red flag? I can't think of any other way to keep cryptarchia and the storage in sync. We can also decide we want to tackle this problem separately, at which point we remove this code completely.

ntn-x2:
@zeegomo you had a good point about the old design to support cryptarchia-DB inconsistencies. I have pushed one more commit (https://github.com/logos-co/nomos/pull/1261/commits/8ea32b8c9f6c8c8034dbbf6e790d1013d6529b57), which reverts the changes to the recovery strategy and the force addition of past blocks to cryptarchia engine, and stores the undeleted blocks in the service state. This is a much better design, as we do not need to tweak cryptarchia in any way now. Any feedback on the new approach? Cc @danielSanchezQ, perhaps you want to re-review the changes in the last commit as well.

nomos-services/cryptarchia-consensus/src/lib.rs
zeegomo: I'd say keeping track of which blocks haven't been removed from storage is not an issue of the cryptarchia engine and should be handled somewhere else. That is a critical component and should be as simple as possible.
Anyway, if we return the list of failed blocks we should be able to try to delete only the blocks that failed, right?

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -661,9 +670,10 @@ where
                         )
                         .await;
 
-                        Self::prune_old_forks(&mut cryptarchia, &mut leader, relays.storage_adapter()).await;
-                        // We update cryptarchia state only after attempting to prune blocks, to avoid saving blocks that have been already successfully pruned.
-                        self.service_resources_handle.state_updater.update(Some(Self::State::from_cryptarchia(&cryptarchia, &leader)));
+                        // This will modify `previously_pruned_blocks` to include blocks which are not tracked by Cryptarchia anymore but have not been deleted from the persistence layer.
+                        Self::prune_forks(&mut cryptarchia, &mut leader, relays.storage_adapter(), &mut previously_pruned_blocks).await;
+                        // We dump the undeleted blocks to storage so they are picked up by the service again if it suddenly stops.
+                        self.service_resources_handle.state_updater.update(Some(Self::State::from_cryptarchia_and_unpruned_blocks(&cryptarchia, &leader, previously_pruned_blocks.clone())));
```
zeegomo, line 676: not sure I get this point

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -1428,86 +1434,56 @@ where
...
-    /// deletion will be performed at the next pruning attempt.
-    async fn prune_old_forks(
+    /// Any block that fails to be deleted from the storage layer is added to
+    /// the provided `prunable_blocks` parameter and will be picked up at the
+    /// next invocation of this function.
+    async fn prune_forks(
         cryptarchia: &mut Cryptarchia<Online>,
         leader: &mut Leader,
         storage_adapter: &StorageAdapter<Storage, TxS::Tx, BS::BlobId, RuntimeServiceId>,
+        prunable_blocks: &mut HashSet<HeaderId>,
```
zeegomo, line 1454: nit: I'd take a reference here and return the newly failed blocks

nomos-services/cryptarchia-consensus/src/states.rs
```diff
@@ -102,9 +98,17 @@ impl<TxS, BxS, NetworkAdapterSettings, BlendAdapterSettings, TimeBackendSettings
         }
     }
 
-    pub(crate) fn from_cryptarchia<State: CryptarchiaState>(
+    /// Re-create the cryptarchia state given the engine instance and the leader
+    /// details.
+    ///
+    /// Furthermore, it allows to specify blocks deleted from the cryptarchia
+    /// engine (hence not tracked anymore) but that should be deleted from the
```
zeegomo, line 105: nit: leftover `but`?

nomos-services/cryptarchia-consensus/src/states.rs
ntn-x2: Changed to a `HashSet`, so we do dedup automatically now.

nomos-services/cryptarchia-consensus/src/lib.rs
ntn-x2: I removed them from the strategy.

consensus/cryptarchia-engine/src/lib.rs
ntn-x2: https://github.com/logos-co/nomos/pull/1261/commits/b2f593d89afabaf2157b9437070b1cef06ad7d3d

consensus/cryptarchia-engine/src/lib.rs
ntn-x2: https://github.com/logos-co/nomos/pull/1261/commits/43d7244e5a8033ada99791095fca7c47c2c18ce5

consensus/cryptarchia-engine/src/lib.rs
ntn-x2: Yeah actually good point. It's not necessary in this case: https://github.com/logos-co/nomos/pull/1261/commits/28332946466b22b4e25e8c723ab7f2c385ad9225

nomos-services/cryptarchia-consensus/src/lib.rs
ntn-x2: We include the blocks that failed to be deleted in the service state, if any, to allow for recovery in case of sudden shutdown, what we usually do for services anyway. Was the comment not clear or what?

nomos-services/cryptarchia-consensus/src/lib.rs
ntn-x2: Yeah I thought about that, I thought this would be cleaner since we would not have to do anything from the calling site, which is within the service's `run` function. We only get a mutable reference, and the rest is taken care of from within this function.

nomos-services/cryptarchia-consensus/src/states.rs
ntn-x2: Mhh the point is that the blocks have been removed from cryptarchia but not from storage, which is an unexpected case. I did mean to add the but as that is not the norm.

PR #1267, opened 2025-05-08
(PoC/Storage): New storage abstraction layer

andrussal:

Trying out different storage layer abstraction approach where we **provide high level functions to store and retrieve data**. 

**This change has 2 goals:**

1. Clear overview how storage backend is used by having all interaction behind meaningful functions and in a common place
2. Stronger separation between low level database mechanics and how the services interact with it. Services don’t need call directly Load/Store functions but can use high level functions like `get_light_share` and `store_block`

It introduces `StorageFunctions` trait that is composed of multiple domains specific traits like `StorageDaApi` and `StorageChainApi`.  And provides few example function implementations.

See also Discord [discussions](https://discord.com/channels/1111286067413405788/1368848774696275979/1368859182320844841) and [discussions](https://discord.com/channels/1111286067413405788/1367809683968229427/1367809687696834593)

@andrussal 

N/A

N/A

## Checklist

* [x] 1. Description added.


nomos-services/storage/src/lib.rs
```diff
@@ -144,28 +150,35 @@ impl<Backend: StorageBackend> Debug for StorageMsg<Backend> {
...
     ReplyError { operation: String, key: Bytes },
     #[error("Storage backend error")]
     BackendError(#[source] Backend::Error),
 }
 
 /// Storage service that wraps a [`StorageBackend`]
-pub struct StorageService<Backend: StorageBackend + Send + Sync + 'static, RuntimeServiceId> {
+pub struct StorageService<Backend, RuntimeServiceId>
+where
+    Backend: StorageBackend + StorageBackendApi + Send + Sync + 'static,
```
ntn-x2, line 173: I guess `StorageBackend` bound is not needed since `StorageBackendApi` is a sub-trait.

nomos-services/storage/src/api/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+pub mod chain;
+pub mod da;
+
+#[async_trait]
+pub trait StorageFunctions: StorageChainApi + StorageDaApi {}
+
+#[async_trait]
+pub trait StorageBackendApi: StorageFunctions + StorageBackend
+where
+    Self: StorageChainApi<HeaderId = HeaderId, Block = Bytes>
```
ntn-x2, line 23: Why the need to force the associated types?

nomos-services/storage/src/api/da/mod.rs
```diff
@@ -0,0 +1,171 @@
...
+                })?;
+
+            if response_tx.send(result).is_err() {
+                error!("Failed to send response in get_light_share");
+            }
+
+            Ok(())
+        }
+        _ => {
+            unreachable!("invalid request")
```
ntn-x2, line 95: Would it make more sense to replace these handler functions with a method of the request itself? If we can do that, it would remove the need to use the `unreachable` statements, which to me imply there is something sub-optimal in this design.

ntn-x2 (commented):
I left a small preliminary review, just to help me better understand the context of this PR. I will take a deeper look once few more things are clear to me.

nomos-services/storage/src/api/da/mod.rs
ntn-x2: It would be something akin to the [command pattern](https://rust-unofficial.github.io/patterns/patterns/behavioural/command.html).

nomos-services/storage/src/api/chain/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+
+use async_trait::async_trait;
+use overwatch::DynError;
+use requests::ChainApiRequest;
+use tracing::error;
+
+use crate::{api::StorageBackendApi, StorageServiceError};
+
+#[async_trait]
+pub trait StorageChainApi {
```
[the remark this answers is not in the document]
andrussal: Long term vision is to have all storage related logic here. We may even consider having more specialised functions here like

```rust
async fn prune_fork(from: HeaderId, to: HeaderId);
```

nomos-services/storage/src/api/da/mod.rs
andrussal: I even tried this approach initially but run into issues with trait objects and generics. I had `Box<dyn Command<T:Backend>>`. But maybe there's some other way of doing it. I think about it.

Maybe another option here is just to pass already destructured values from a specific request instead of whole request. I did destructuring here locally in each specific handle function to avoid "explosion" in the main handling function.

Or I could use structs with each enum variant instead of fields in enum variants bodies. That would make it also more concise.

ntn-x2: So no need to use `Box<dyn>` stuff, as far as I can tell.

nomos-services/storage/src/api/backend/mod.rs
```diff
@@ -0,0 +1,2 @@
+#[cfg(feature = "rocksdb-backend")]
+pub mod rocksdb;
```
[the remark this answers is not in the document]
andrussal: Exactly and I also though that this way we can refactor it incrementally.

nomos-services/storage/src/api/da/mod.rs
andrussal: @ntn-x2, @youngjoon-lee What do you guys think of doing it like this?  https://github.com/logos-co/nomos/pull/1267/commits/011c8d728caf68187db8b3b5154d351e5d1db862

ntn-x2: Yes, to me this looks already a much better way to go! I would perhaps just rename the `Executable` trait to something else, like `StorageTransaction` or `StorageOperation` or something else.

nomos-services/storage/src/api/chain/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+    type HeaderId;
+    type Block: Debug;
```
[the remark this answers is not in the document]
andrussal: Would it make sense to have some bounds in trait definition when they are not too restrictive and (sort of) always used? Then we end up having less (boilerplate) code in elsewhere. I have remove `Debug` but right now kept `Send + Sync`.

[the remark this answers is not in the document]
ntn-x2: Yeah for `Debug` specifically, there are already talks to make it a default instead of explicitly requiring it all the time. My question was genuine as if there was a specific need for it. This has been a conversation before also for other team members. For me personally, trait bounds on trait definitions make sense if they are used as part of the trait API or used by auto-implemented methods. But I understand this is more a matter of taste, so I don't have anything against requiring `Debug` here, even though the question would then be "why only for `Block`"? What if we want to debug print also the header?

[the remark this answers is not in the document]
ntn-x2: And debug printing is something that normally the trait implementer does, which is why I think it should be them to implement `Debug` on whatever type they are going to use.

nomos-services/storage/src/api/mod.rs
andrussal: Now I just enforce it like this where necessary:

```rust
    <Storage as StorageChainApi>::Block:
        From<Block<ClPool::Item, DaPool::Item>> + Into<Block<ClPool::Item, DaPool::Item>>
```

I was a bit worried that this causes "explosion" of bounds elsewhere but it seems that there's no better way.

For `HeaderId` I just use `nomos-core::HeaderId` directly as it's easy to use in tests as well.

nomos-services/storage/src/api/chain/mod.rs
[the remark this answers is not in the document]
andrussal: I think it makes sense to have `Send + Sync` for associated types here as these traits are always used in async context(async trait functions). But `Debug` is more arbitrary and I removed it. wdyt?

nomos-services/storage/src/api/da/mod.rs
andrussal: `StorageOperation` name seems perfect! I used that.

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -59,3 +59,28 @@ impl<
         self.bl_blobs.len()
     }
 }
+
+impl<
```
ntn-x2, line 63: Now that I look at this, I think the `Block` struct does not need any bounds in its definition, since `IndexSet` does not require anything either. What do you think about removing those bounds from the `Block` definition, which would then imply remove all bounds but `DeserializeOwned` for the `Tx` type here? I think it would be much cleaner.

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -59,3 +59,28 @@ impl<
...
+    > TryFrom<Bytes> for Block<Tx, BlobCertificate>
+{
+    type Error = wire::Error;
+
+    fn try_from(bytes: Bytes) -> Result<Self, Self::Error> {
+        wire::deserialize(&bytes)
+    }
+}
+
+impl<
```
ntn-x2, line 75: Same here. I think we can get to a point where we only require `Tx: Serialize`.

nomos-services/cryptarchia-consensus/src/storage/mod.rs
```diff
@@ -22,4 +22,10 @@ pub trait StorageAdapter<RuntimeServiceId> {
     ///
     /// The block with the given header id. If no block is found, returns None.
     async fn get_block(&self, key: &HeaderId) -> Option<Self::Block>;
+
+    async fn store_block(
+        &self,
+        header_id: HeaderId,
```
ntn-x2, line 28: Why is this not generic over the `Header` as well? Or perhaps would it make sense for `type: Block` to implement the `Block` trait, which would define what its header would look like?

nomos-services/cryptarchia-consensus/src/storage/adapters/storage.rs
```diff
@@ -1,57 +1,34 @@
...
-}
-
 #[async_trait::async_trait]
 impl<Storage, Tx, BlobCertificate, RuntimeServiceId> StorageAdapterTrait<RuntimeServiceId>
     for StorageAdapter<Storage, Tx, BlobCertificate, RuntimeServiceId>
 where
-    Storage: StorageBackend + Send + Sync,
-    Tx: Clone + Eq + Hash + DeserializeOwned + Send + Sync,
-    BlobCertificate: Clone + Eq + Hash + DeserializeOwned + Send + Sync,
+    Storage: StorageBackend + Send + Sync + 'static,
```
ntn-x2, line 27: Do we need all these trait bounds here?

nomos-services/storage/src/api/da/requests.rs
```diff
@@ -0,0 +1,162 @@
...
+    share_idx: B::ShareIndex,
+    response_tx: Sender<Option<B::Share>>,
+) -> Result<(), StorageServiceError<B>> {
+    let result = backend
+        .get_light_share(blob_id, share_idx)
+        .await
+        .map_err(|e| {
+            error!("Failed to get light share: {:?}", e);
+
+            let key = "blob_id".to_owned().into();
```
ntn-x2, line 63: Should all these values be made constants, to avoid they get out of sync?

nomos-services/storage/src/api/da/requests.rs
```diff
@@ -0,0 +1,162 @@
...
+) -> Result<(), StorageServiceError<B>> {
+    backend
+        .store_light_share(blob_id, share_idx, light_share)
+        .await
+        .map_err(|e| {
+            error!("Failed to store light share: {:?}", e);
+
+            let key = "blob_id".to_owned().into();
+            StorageServiceError::ReplyError {
+                operation: "store_light_share".to_owned(),
```
ntn-x2, line 91: I think for these it might even make more sense to define them as an enum and implement `ToString` for it.

nomos-services/storage/src/api/chain/requests.rs
```diff
@@ -0,0 +1,99 @@
...
+
+async fn handle_get_block<B: StorageBackend>(
+    backend: &mut B,
+    header_id: HeaderId,
+    response_tx: Sender<Option<B::Block>>,
+) -> Result<(), StorageServiceError<B>> {
+    let result = backend.get_block(header_id).await.map_err(|e| {
+        error!("Failed to get block: {:?}", e);
+
+        let key = "header_id".to_owned().into();
```
ntn-x2, line 47: Same point as for the DA operations about unifying the definitions and using enums where it would make more sense.

ntn-x2 (commented):
I'm leaving a second round of review, but I will let someone that has already worked with storage to approve the PR. For me, the only bit I am not sure about, is whether it is ideal to expose a high level API that is to be implemented by each storage layer: for me it seems a mix of responsibilities where a storage is not required to only store/retrieve stuff, but must also have somewhat of business logic. Think about implementing `prune_fork`, where now we would not do it anymore in cryptarchia, or at least the whole logic would not live in cryptarchia anymore but would be spread between cryptarchia and the backend service used. Also for `store_block`, is the backend going to run checks on the block, e.g., if the parent exists? Those checks are already performed by cryptarchia, and if I understand correctly `store_block` would then simply be dumping the serialised block to storage, which would then be exposed in the same interface as `prune_fork`, which most likely implements some more logic then just dumping stuff to storage.

But in general, this probably stems from some concrete needs that were discussed and that I have no context about, so I am just providing my point of view based on this PR and what I have understood of it, but of course I might be missing or misunderstanding some stuff.

EDIT: actually the logic is implemented by the request and not the underlying backend, so it is implemented only once. Still, I think there is a question of how "high-level" each request should be? Do we plan to specify that clearly for each request variant definition?

nomos-core/chain-defs/src/block/mod.rs
ntn-x2: If it's out of scope for this PR, I understand, although it's a pity to introduce all these bounds when they are not really needed, it seems.

nomos-services/cryptarchia-consensus/src/storage/mod.rs
ntn-x2: How do we know that the provided `Block` is compatible with the `HeaderId` we use?

andrussal:
> I'm leaving a second round of review, but I will let someone that has already worked with storage to approve the PR. For me, the only bit I am not sure about, is whether it is ideal to expose a high level API that is to be implemented by each storage layer: for me it seems a mix of responsibilities where a storage is not required to only store/retrieve stuff, but must also have somewhat of business logic. Think about implementing `prune_fork`, where now we would not do it anymore in cryptarchia, or at least the whole logic would not live in cryptarchia anymore but would be spread between cryptarchia and the backend service used. Also for `store_block`, is the backend going to run checks on the block, e.g., if the parent exists? Those checks are already performed by cryptarchia, and if I understand correctly `store_block` would then simply be dumping the serialised block to storage, which would then be exposed in the same interface as `prune_fork`, which most likely implements some more logic then just dumping stuff to storage.
> 
> But in general, this probably stems from some concrete needs that were discussed and that I have no context about, so I am just providing my point of view based on this PR and what I have understood of it, but of course I might be missing or misunderstanding some stuff.
> 
> EDIT: actually the logic is implemented by the request and not the underlying backend, so it is implemented only once. Still, I think there is a question of how "high-level" each request should be? Do we plan to specify that clearly for each request variant definition?

I think new design has many benefits. The main ideas behind the refactoring from my perspective are these:

1. Database neutral interface to the services. Even if we don’t switch database right now, we are much better prepared.
2. Better visibility how storage backend is used by having semantically clear functions in a single location.
3. (I realised also this benefit) - easier to add caching layer if everything is in single location 
4. (And this) - Also access to database specific features is simplified as we don’t need to find cumbersome ways to add these to our database abstraction layer. For example now it’s really easy to use database transactions.

Actually this is completely open yet how specialised functions we feel comfortable putting here. I made ‘prune_fork’ just an example of high level storage function that might make sense to put here as it is should be fairly well defined set of databases operations that can work solely based on its input parameters. But if it’s more complicated than that then it can make more sense to leave it out of course. But I see main benefit having potentially tens or hundreds of sequential database operations in a single place(potentially performed in a database transaction). Also keeping the service layer more clean by moving this heavily storage related functionality out.

Actually I think it would make sense to have database constraints like check parent block exist if database supports it. It would be just an extra safeguard. But I agree here is some granularity how much checks it makes sense to have. We don’t want to update something in multiple places. 

Recently we investigated if SurrealDB would work for us because we come up with situations where we need to store more than just blob of bytes. In addition some extra information like share ids of a blob. Similar requirements probably come later with other “domains”.  This raised few points. Can we use db internal indexes instead of doing this kind things manually. And also it’s very cumbersome to do this stuff with current db abstraction layer if indexing manually.  Even if we can’t use database internal indexes right now, new interface allows to group these operations easily into one transaction.

Refactoring idea came from @danielSanchezQ and @bacv , curious to know their thoughts about the PoC 🙏

nomos-services/cryptarchia-consensus/src/storage/mod.rs
andrussal: I was hoping to reduce the number of places where we need to add constraints and  `HeaderId` is part of `nomos-core` which makes it in some sense part of "`std`" for us. But you make a really good point, it should be compatible with `Block` type. Is it ok to ask all block impls to use `HeaderId` from `nomos-core`.

nomos-services/storage/src/api/da/requests.rs
andrussal: I realised that `ReplyError { operation: String, key: Bytes },` is not actually right thing here(`Bytes`). And it's used only for logging where it is used previously. I just removed its usage in new api. I think that makes sense just to log if sending over channel fails(because there's nothing else to do)?

nomos-core/chain-defs/src/block/mod.rs
ntn-x2: Unfortunately there is not. https://github.com/rust-lang/rust-clippy/issues/1463, https://users.rust-lang.org/t/unused-trait-bounds/37516. Reason being there is risk for high percentage of false positives, since in some cases a marker trait might be required. I have never myself come across such an instance (besides for `Send` and `Sync`, which could be special cases), and I would think an allow-by-default lint might be a good thing to have, but right now it does not exist.

nomos-services/cryptarchia-consensus/src/storage/mod.rs
ntn-x2: I guess doing that would make testing, especially unit testing, somewhat more complicated, as we now need to pull in `HeaderId` from `nomos-core`, which is 1. a new dependency, 2. perhaps the result of calling `<nomos_core::Block as Block>::Id` or something like that. I would rather see this trait definition be generic, but others might have a different opinion about it. Also currently we don't have a trait for a `Block` that would define its header and consequently its ID, although I think we should in the future. So I guess for now it's ok to restrict it, or perhaps introduce a trait that would be implemented by the provided `Block` type that would return its ID, on the way to declare a proper trait for it and use it everywhere.

ntn-x2: But if we require it in the trait impl and not in the trait itself, I guess it would already be better than the current approach.

nomos-services/storage/src/backends/sled.rs
```diff
@@ -56,38 +58,110 @@ impl<SerdeOp: StorageSerde + Send + Sync + 'static> StorageBackend for SledBacke
...
         transaction: Self::Transaction,
-    ) -> Result<<Self::Transaction as StorageTransaction>::Result, Self::Error> {
+    ) -> Result<<Self::Transaction as StorageTransaction>::Result, <Self as StorageBackend>::Error>
+    {
         Ok(self.sled.transaction(transaction))
     }
 }
 
+#[async_trait]
+impl<SerdeOp: StorageSerde + Send + Sync + 'static> StorageChainApi for SledBackend<SerdeOp> {
```
bacv, line 105: Lets just delete sled backend, no need to keep it.

nomos-services/storage/src/api/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+
+pub mod backend;
+pub mod chain;
+pub mod da;
+
+#[async_trait]
+pub trait StorageFunctions: StorageChainApi + StorageDaApi {}
+
+#[async_trait]
+pub trait StorageBackendApi: StorageFunctions
```
ntn-x2, line 20: One last question: why is `where Self: StorageChainApi + StorageDaApi` needed if `StorageBackendApi` is already a sub-trait of `StorageFunctions`, which is itself a subtrait of `StorageChainApi` and `StorageDaApi`? I think we can remove that, just for cleanliness.

nomos-services/cryptarchia-consensus/src/storage/mod.rs
bacv: Consensus service is not generic over block type, I think it's okay to keep it constrained to `Block` and `HeaderId` from nomos-core here as well, since this adapter is used by the consensus service.

bacv (approved):
In my view this changes is exactly what we needed, the adapter code is simplified allowing all optimizations specific to the db backend be performed in the storage service. In general it would be great to minimize the `expect` count in serialization related code, but this could be done in upcoming PRs.

nomos-core/chain-defs/src/block/mod.rs
ntn-x2: Yeah but the `Block` definition is a simple one, and removing trait bounds only from the struct definition is non breaking and relatively simple, in this specific case, that's why I suggested it. On a global scale, we probably have quite a bit of over-requirement of trait bounds, and I will try to pay particular attention to them when I review PRs, so we can remove them as we go, I'd say.

nomos-services/data-availability/verifier/src/storage/adapters/rocksdb.rs
```diff
@@ -59,100 +54,51 @@ where
         shared_commitments: <Self::Share as Share>::SharesCommitments,
         light_share: <Self::Share as Share>::LightShare,
     ) -> Result<(), DynError> {
-        let blob_share_idx = create_share_idx(blob_id.as_ref(), share_idx.as_ref());
-        let share_key = key_bytes(DA_SHARE_PREFIX, blob_share_idx);
-        let shared_commitments_key = key_bytes(DA_SHARED_COMMITMENTS_PREFIX, &blob_id);
-        let index_key = key_bytes(DA_BLOB_SHARES_INDEX_PREFIX, &blob_id);
+        let blob_id: [u8; 32] = blob_id.as_ref().try_into().unwrap();
```
danielSanchezQ, line 57: I think we have a `blobid` typealias. We should use it next time.

PR #1268, opened 2025-05-09
feat(services): membership service sketch + mock implementation

pradovic:

This is a sketch implementation of membership service. It covers minor refactoring of sdp-service to support what is needed for membership service sketch. Membership trait definition and service implementation are at proof of concept level and will be worked on in the subsequent PRs.

todo in subsequent PR(s):
- Implement sdp backend with persistent storage

Yes. 

The code is non-functional sketch of membership service. The purpose of this PR should be to connect most of  components needed for membership service to create and provide membership snapshots based on the finalized events it receives from sdp service.

@danielSanchezQ , @bacv , @pradovic 

The general specification is accurate. Architecture might change slightly during implementation as we find out new cases. The plan is to implement it in iterations until everything is covered.

It adds membership information to he system for the rest of services to use if needed.

## Checklist

> [!WARNING]  

* [x] 1. Description added.


nomos-services/membership/src/backends/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+    // Get the snapshot at the given index for the specified service type.
+    // The snapshot index a sequential order of the service type specific time
+    // snapshots. index = 0 is the latest snapshot
+    // index > 0 is the index of the snapshot in the past
+    // index < 0 is the index of the snapshot relative to the latest snapshot
+    async fn get_snapshot_at(
+        &self,
+        service_type: ServiceType,
+        index: i32,
+    ) -> Result<HashMap<ProviderInfo, DeclarationUpdate>, DynError>;
```
bacv, line 38: We could have `MembershipError` alongside `MembershiptEntry`, it would be more convenient to act on errors on receiving end.

nomos-services/membership/src/backends/sdp.rs
```diff
@@ -0,0 +1,83 @@
+use std::collections::HashMap;
+
+use nomos_sdp::backends::FinalizedBlockEvent;
+use nomos_sdp_core::{DeclarationUpdate, ProviderInfo, ServiceType};
+
+use super::{MembershipBackend, MembershipBackendSettings, MembershipEntry, SnapshotSettings};
+
+pub struct SdpMembershipBackend {
```
bacv, line 8: Could this be called `MockMembershipBackend` and the `Settings` would have Entries from configuration file?

nomos-services/membership/src/backends/sdp.rs
```diff
@@ -0,0 +1,83 @@
...
+            membership: Vec::new(),
+            current_data: HashMap::new(),
+        }
+    }
+
+    async fn get_snapshot_at(
+        &self,
+        service_type: ServiceType,
+        index: i32,
+    ) -> Result<HashMap<ProviderInfo, DeclarationUpdate>, overwatch::DynError> {
```
bacv, line 29: Would it make sense here to return only `ProviderId + Declaration` instead of `ProviderInfo + DeclarationUpdate`, or even better - `ProviderId + Locators`. In this case provider id is `Ed25519_Public_Key` that we'll use to derive `PeerId` from where required, and locators are the addresses to try and connect to. I think other information is not needed by the Membership service users.

nomos-services/membership/src/backends/mod.rs
```diff
@@ -0,0 +1,40 @@
+use std::collections::HashMap;
+
+use async_trait::async_trait;
+use nomos_sdp::backends::FinalizedBlockEvent;
```
bacv, line 4: Would it make sense to move this to `services/nomos-sdp/lib.rs`, or even `nomos-sdp-core`, that way membership would require spd-service dependency only in the sdp adapter, and we wouldn't need sdp service for the mock implementation. I'm not sure about this, but it seems that `FinalizedBlockEvent` fits into `nomos-sdp-core`.

bacv (changes requested):
It would be great to have initial Membership service version fully mocked and independent of sdp service, also most of information related to declarations are not required for membership service users. If the mock membership could be loaded from the config then we could move on with integrating this into DA.

nomos-services/membership/src/backends/sdp.rs
pradovic: Great question, I was not sure about the concrete type (and error type) yet, as I did not fully implemented membership service. 

What does membership service needs relies on what do we need to create a snapshot.  I was thinking that maybe we will need to also fetch service params and remove the expired declarations in order for active providers to be correct in each snapshot.

The idea was to maybe use ServicesAdapter from Sdp to fetch service parameters and use the info from ProviderInfo to expiry old ones. We could also rely on the sdp to provide only up-to-date information, depending on how we implement garbage collection. 

What is your opinion on this? I think if we don't have to check the expiration, then I 200% agree that only ProviderId and Locators is enough.

nomos-services/membership/src/backends/mod.rs
pradovic: I did not implement exact type and error yet, this was still in progress so I can add what I encounter and modify, during implementation of the backend. I will define the error type as well, we will need it for sure, and modify variants in different PRs if needed.

nomos-services/membership/src/backends/sdp.rs
pradovic: The only reason I did not call it MockMembershipBackend but SdpMembershipBackend is because I am getting and event with types from sdp (declarationID, ProviderID, Locators).

But, what you suggested in the PR comment make sense, and mock backend can still use types from sdp, it does not matter, so I will rename it. Thanks!

pradovic:
@bacv Understood. I will remove the sdp stuff then (except for needed types).

nomos-services/membership/src/backends/sdp.rs
bacv: For DA and Blend normal operation we'll need to stream the latest active provider ids and their locators. For DA, we'll need historic data based on BlockNumber. Executor network most likely will need only the latest data.
`get_snapshot_at` could stay as is by providing only ids and locators, and then we still would require a message to subscribe to latest provider ids that became active, or got removed.

pradovic: Great, I would add subscription to the API as well then. Thank you

nomos-sdp/src/lib.rs
```diff
@@ -211,3 +211,9 @@ impl<Metadata, Proof> SdpMessage<Metadata, Proof> {
         }
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct FinalizedBlockEvent {
+    pub block_number: BlockNumber,
+    pub updates: Vec<(ProviderInfo, DeclarationUpdate)>,
```
bacv, line 218: After giving it another though, it seems that it would be better to define a `ProviderInfoUpdate`, since we Membership service doesn't have pruning yet, it's not that obvious, but ProviderInfo in itself doesn't have the state, because the state depends on the block number, in sdp ledger it's derived by converting ProviderInfo with 

``` 
impl ProviderState {
    pub fn try_from_info (
```

in `nomos-sdp/src/state.rs`, but here we could return `ProviderInfoUpdate` defined something like this in `nomos-sdp/src/lib.rs`:

```
enum ProviderState {
  Active, Inactive, Withdrawn
}

struct ProviderUpdate {
  provider_id: ProviderId,
  state: ProviderState,
}
```

nomos-services/membership/src/backends/mock.rs
```diff
@@ -0,0 +1,439 @@
...
+    ) -> MembershipProviders {
+        self.membership
+            .get(&block_number)
+            .and_then(|entry| entry.get(&service_type))
+            .map_or_else(HashMap::new, |data| {
+                data.iter()
+                    .map(|(provider_info, declaration_update)| {
+                        (
+                            provider_info.provider_id,
+                            declaration_update.locators.clone(),
```
bacv, line 126: For mock this will work, when implementing with storage, we just need to keep provider_id, it's active block numbers and latest locators.

bacv (commented):
Sorry that didn't catch the `FinalizedBlockEvent` struct definition, but I think we could use better structure there.

pradovic:
> Sorry that didn't catch the `FinalizedBlockEvent` struct definition, but I think we could use better structure there.

Thanks for the review! No worries, those were actually left like this "on purpose". I explained in the comments. Let me know if it makes sense. I also don't mind changing  them real quick if needed.

nomos-services/membership/src/backends/mock.rs
bacv: For membership service I don't think we need to know service parameters - sdp will determine the state of provider and return the ProviderUpdate. SDP will manage service parameters (updating these parameters is not yet specified as far as I know). Pruning for very old data here should be done by block number, configured separately via membership settings.

nomos-sdp/src/lib.rs
bacv: Not sure if `Event` definition from the spec could be mapped to suggested `ProviderUpdate`. For this PR, lets just define and return what is needed for the Membership service - `ProviderId` + `State` + `BlockNumber` + `Locators`.

pradovic: Sorry, when I said `Event` I met this event `FinalizedBlockEvent`. 

Ok, sounds good.

nomos-services/membership/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+                        MembershipMessage::Subscribe { service_type, result_sender } => {
+                            let tx = if let Some(tx) = self.subscribe_txs.get(&service_type) {
+                                tx.clone()
+                            } else {
+                                let (tx, _) = broadcast::channel(128);
+                                self.subscribe_txs.insert(service_type, tx.clone());
+                                tx
+                            };
+
+                            let stream = make_pin_broadcast_stream(tx.subscribe());
```
bacv, line 142: Shouldn't we write the latest members to the newly created stream after somebody asks for membership?

pradovic: Good point! I will add the latest snapshot to the stream before returning it to the caller.

nomos-services/membership/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+                    if lifecycle::should_stop_service::<Self, RuntimeServiceId>(&msg) {
+                        break;
+                    }
+                }
+                Some(sdp_msg) = sdp_stream.next() => {
+                     match self.backend.update(sdp_msg).await {
+                        Ok(snapshot) => {
+                            for (service_type, snapshot) in snapshot {
+                                if let Some(tx) = self.subscribe_txs.get(&service_type) {
+                                    if tx.send(snapshot).is_err() {
```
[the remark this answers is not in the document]
bacv: Yes, if we send full list on subscription event, here only the changes can be streamed when they happen. Looks good.

bacv (approved):

nomos-sdp/src/ledger.rs
```diff
@@ -396,22 +398,31 @@ where
...
-        Ok(())
+        Ok(result)
     }
 
     pub async fn mark_in_block(
         &mut self,
         block_number: BlockNumber,
-    ) -> Result<(), SdpLedgerError<ContractAddress>> {
+    ) -> Result<FinalizedBlockEvent, SdpLedgerError<ContractAddress>> {
+        let mut result = FinalizedBlockEvent {
```
danielSanchezQ, line 417: What is this event?

nomos-sdp/src/ledger.rs
```diff
@@ -420,12 +431,40 @@ where
...
+                            service_type,
+                            provider_id,
+                            state,
+                            locators: declaration_update.locators,
+                        });
+                    }
+                }
+                Err(err) => {
+                    tracing::error!("Provider information couldn't be updated: {err}");
+                }
```
danielSanchezQ, line 463: Why are we marking in block? Nothing that is not in block should go to sdp...

nomos-sdp/src/lib.rs
```diff
@@ -39,6 +41,18 @@ pub enum ServiceType {
...
+        let service_str = match self {
+            Self::BlendNetwork => "BlendNetwork",
+            Self::DataAvailability => "DataAvailability",
+            Self::ExecutorNetwork => "ExecutorNetwork",
+            Self::GenericRestaking => "GenericRestaking",
+        };
+        write!(f, "{service_str}")
+    }
+}
+
```
danielSanchezQ, line 55: Can't this be derived? 🤔

nomos-sdp/src/lib.rs
```diff
@@ -211,3 +225,17 @@ impl<Metadata, Proof> SdpMessage<Metadata, Proof> {
...
+    pub updates: Vec<FinalizedBlockEventUpdate>,
+}
+
+#[derive(Debug, Clone)]
+pub struct FinalizedBlockEventUpdate {
+    pub service_type: ServiceType,
+    pub provider_id: ProviderId,
+    pub state: ProviderState,
+    pub locators: Vec<Locator>,
+}
```
danielSanchezQ, line 241: I'm not sure this is the way of dealing with this. Maybe I'm missing something.

nomos-sdp/src/state.rs
```diff
@@ -15,7 +15,7 @@ pub enum ActiveStateError {
 }
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
-pub struct ActiveState(ProviderInfo);
+pub struct ActiveState(pub ProviderInfo);
```
danielSanchezQ, line 18: What is the reason of wrapping here if the inner is public? Either is hidden and we implement `AsRef<ProviderInfo>` or we do no need the wrapping type at all.

nomos-services/membership/src/adapters/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+    }
+}
+
+#[async_trait]
+pub trait SdpAdapter {
+    type SdpService: ServiceData;
+
+    fn new(outbound_relay: OutboundRelay<<Self::SdpService as ServiceData>::Message>) -> Self;
+    async fn finalized_blocks_stream(&self) -> Result<FinalizedBlockUpdateStream, SdpAdapterError>;
+}
```
danielSanchezQ, line 31: There is no block that updates the sdp service that is not finalized.

danielSanchezQ (changes requested):
~~I think we have a missconception and we should sync about this. Pinging @pradovic and @bacv.~~ All solved. I'll wait for the cleanup for the approval. Thanks!

nomos-sdp/src/lib.rs
pradovic: Hmm, maybe I am missing something. How do I derive the Display, using `strum` or some lib? Sorry if it's a noob question

nomos-sdp/src/state.rs
pradovic: Good catch, I was sloppy. I just needed to initialize provider state in tests. Solved in the latest commit.

nomos-sdp/src/lib.rs
pradovic: For example:
```
if let Ok(providers) = providers {
                                if tx.send(providers).is_err() {
                                    tracing::error!("Error sending initial membership snapshot for service type: {}", service_type);
                                } else if result_sender.send(stream).is_err() {
                                    tracing::error!("Error sending finalized updates receiver for service type: {}", service_type);
                                }
                            } else {
                                tracing::error!("Failed to get latest providers for service type: {}", service_type);
                            }

 ```

Comment (danielSanchezQ) [reply nomos-sdp/src/lib.rs:55]: In that case you can use the `Debug` version of it directly:
```
tracing::error!("Error sending initial membership snapshot for service type: {:?}", service_type);
```
Which is basically what you implemented anyway iirc.

Comment (pradovic) [reply nomos-sdp/src/lib.rs]: Makes sense, fixed.

File: nomos-services/membership/src/lib.rs [hunk 15]
```diff
@@ -0,0 +1,199 @@
...
+};
+use serde::{Deserialize, Serialize};
+use services_utils::overwatch::lifecycle;
+use tokio::sync::{broadcast, oneshot};
+use tokio_stream::wrappers::BroadcastStream;
+
+mod adapters;
+pub mod backends;
+
+type MembershipProviders = HashMap<ProviderId, Vec<Locator>>;
```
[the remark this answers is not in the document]
pradovic: Might be better. I was not 100% sure until I see the usage. Does it make sense to change this if I see some patterns when I implement the usage from DA?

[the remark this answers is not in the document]
danielSanchezQ: Uhm, it would depend. The idea from the comment is to somehow avoid duplicated locators (just in case). I would use `BTreeSet` just so we have a consistency in ordering.

nomos-services/membership/src/lib.rs
```diff
@@ -124,64 +123,100 @@ where
...
+                        tracing::error!(
+                            "Error sending finalized updates receiver for service type: {:?}",
+                            service_type
+                        );
+                    }
+                } else {
+                    tracing::error!(
+                        "Failed to get latest providers for service type: {:?}",
+                        service_type
+                    );
```
danielSanchezQ, line 194: Should we sent an empty stream of providers back from the api? Otherwise we would be blocking the consuming service.

nomos-sdp/src/state.rs
```diff
@@ -265,6 +265,24 @@ impl From<WithdrawnState> for ProviderState {
     }
 }
 
+impl From<ProviderInfo> for ActiveState {
```
bacv, line 268: Where are these conversions used?

nomos-services/membership/src/lib.rs
pradovic: Hmmm, make sense.. Actually on a second thought, I will return an error, good catch

nomos-sdp/src/state.rs
pradovic: When I removed pub field from State, I need a way to initialize the state for tests. So it was either this or a constructor methods. Does that make sense?

Example:
```
 let updates = vec![FinalizedBlockEventUpdate {
            service_type,
            provider_id: provider_info_1.provider_id,
            state: ProviderState::Active(provider_info_1.into()),
            locators: BTreeSet::from_iter(declaration_update_1.locators.clone()),
        }];
        
```

bacv: Lets add the `#[cfg(test)]` to each of them, these should not be allowed to be used outside the tests. I'll think how to remove them completely in the refactor, will probably make `state.rs` private module to the `nomos-sdp` and expose new enum similar to this `enum State { Active, Inactive, Withdrawn }`

pradovic: I think enum makes sense. I wanted to change it, but did not want to expand the scope of the PR any more. I will add the `#[cfg(test)]`

pradovic: Actually, I dont even need this, I will just use it the same way as in the tests of the state.

pradovic: Done

nomos-services/membership/src/lib.rs
pradovic: Fixed.

pradovic:
@bacv , @danielSanchezQ Thanks for the review!

nomos-services/sdp/src/adapters/activity/mod.rs
```diff
@@ -0,0 +1,7 @@
+use nomos_sdp_core::ledger;
+
+pub mod contract;
+
+pub trait SdpActivityAdapter: ledger::ActivityContract {
+    fn new() -> Self;
+}
```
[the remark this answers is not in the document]
pradovic: I think these will be worked on very soon, when implementing sdp dependencies, cc @bacv

PR #1270, opened 2025-05-12
chore: enforce taplo formatting in CI

ntn-x2:

We run pre-commit hooks, but I somehow forgot to add the taplo check in the CI as well, which means of course some TOML files have not been following the taplo formatting. This PR fixes that.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 1. Description added.

### DIFF
--- a/.github/workflows/code-check.yml
+++ b/.github/workflows/code-check.yml
@@ -48,8 +48,8 @@ jobs:
           args: --all --check
       - name: Install taplo CLI
         run: cargo install taplo-cli --locked --version 0.9.3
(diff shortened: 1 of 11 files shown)


ntn-x2:
> Weird it changed the vertical alignment. Maybe we should double check the config is being picked up?

It is being picked up, since `fmt` also reorded some previously unorder (alphabetically) entries. Also changing the indentation of any TOML file and running `fmt` will re-apply the indentation. So I think that's most likely just a bug.

bacv (approved):

PR #1272, opened 2025-05-15
feature(lifecycles): Integrate Overwatch's lifecycle changes

strinnityk:
Integrate Overwarch's [Lifecycle](https://github.com/logos-co/Overwatch/pull/79) changes, that move the handling of lifecycles from `Service`s to their `ServiceRunner`s.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 1. Description added.


nomos-services/data-availability/network/src/lib.rs
```diff
@@ -79,46 +78,46 @@ where
...
-                    Self::handle_network_service_message(msg, &mut backend).await;
-                }
-                Some(msg) = lifecycle_stream.next() => {
-                    if lifecycle::should_stop_service::<Self, RuntimeServiceId>(&msg) {
-                        // TODO: Maybe add a call to backend to handle this. Maybe trying to save unprocessed messages?
-                        backend.shutdown();
-                        break;
-                    }
-                }
+            if let Some(msg) = inbound_relay.recv().await {
```
ntn-x2, line 108: I guess we could use `while let Some()`

nomos-services/key-management-system/src/lib.rs
```diff
@@ -125,34 +123,28 @@ where
...
-            tokio::select! {
-                Some(msg) = service_state.inbound_relay.recv() => {
-                    Self::handle_kms_message(msg, &mut backend).await;
-                }
-                Some(msg) = lifecycle_stream.next() => {
-                    if lifecycle::should_stop_service::<Self, RuntimeServiceId>(&msg) {
-                        return Ok(());
-                    }
-                }
+            if let Some(msg) = service_resources_handle.inbound_relay.recv().await {
```
ntn-x2, line 146: `while let Some()`

ntn-x2 (commented):
I like how non-intrusive the changes are, so far! Very nice! 

One thing I was thinking is: even if we remove all responsibility for handling lifecycle messages from the services, do we still want to give them the chance to finalise any pending stuff before being shut down? I am thinking about flushing to disk some logs or other similar stuff, instead of (like we do now) calling the `state_updater` after every operation, for instance. They would not need to know WHEN to do it, but only what to do. We could also let overwatch define a timeout, if we really want, after which services are anyway shut down, although I don't think we need to be that strict, for now. Ccing also @danielSanchezQ here.

nomos-services/data-availability/network/src/lib.rs
danielSanchezQ: ➕ 1

danielSanchezQ:
> I like how non-intrusive the changes are, so far! Very nice!
> 
> One thing I was thinking is: even if we remove all responsibility for handling lifecycle messages from the services, do we still want to give them the chance to finalise any pending stuff before being shut down? I am thinking about flushing to disk some logs or other similar stuff, instead of (like we do now) calling the `state_updater` after every operation, for instance. They would not need to know WHEN to do it, but only what to do. We could also let overwatch define a timeout, if we really want, after which services are anyway shut down, although I don't think we need to be that strict, for now. Ccing also @danielSanchezQ here.

In this case I agree we could be softer in how to terminate services. Maybe we could include a terminate channel that signals a service would be shutdown with some timeout so the service (opt-in) does what it needs to do. Anyway, this is separated from what saving state really does, as each service should decide when to actually do it (including on shutdown). Wdyt @ntn-x2 & @strinnityk ? Adding such channel should not be a difficult add-on.

ntn-x2:
> In this case I agree we could be softer in how to terminate services.
+1

> Maybe we could include a terminate channel that signals a service would be shutdown with some timeout so the service (opt-in) does what it needs to do
I was thinking more that overwatch would call `async fn shutdown(self)` on each service runner and wait for all of them to terminate, rather than the other way around. But both are valid approaches.

> Anyway, this is separated from what saving state really does, as each service should decide when to actually do it (including on shutdown)
Yes and no. Yeah state operators are pre-emptively saving state in case of crashes as well, which this case would not cover. But having one more mechanism for flushing stuff to state would give them a choice, while right now if we remove all lifecycle logic from a service, they would have no clue as to when to do cleanups, since it could happen any time. Also, a service being shut down might want to shut down related services as well, instead of the calling service (or even worse overwatch) having to know all the services that need to be shut down.

But in general I'd be fine to tackle that in a separate PR, but perhaps wait until that is also worked on before updating nomos.

danielSanchezQ:
> I was thinking more that overwatch would call `async fn shutdown(self)` on each service runner and wait for all of them to terminate, rather than the other way around. But both are valid approaches.

Uhm, I mean, the service runner is a future that consumed the service itself. At the moment is a cancellable future (iirc). In this case we would have to somehow recover the service itself so then a shutdown operation is called upon it? I think I'm missing something here.

> Yes and no. Yeah state operators are pre-emptively saving state in case of crashes as well, which this case would not cover. But having one more mechanism for flushing stuff to state would give them a choice, while right now if we remove all lifecycle logic from a service, they would have no clue as to when to do cleanups (e.g., deleting a temporary file), since it could happen any time. Also, a service being shut down might want to shut down related services as well, instead of the calling service (or even worse overwatch) having to know all the services that need to be shut down. The alternative to this would be to introduce a shutdown(service_group) message, which would also be a possibility.

The `fn shutdown(&[ServiceId])` should probably be added anyway so we can prepara subgrups of services that work together indeed, I had that in mind. Now, about the flushing mechanism. I think just signaling to the service that is gonna be terminated is enough. Then service implementors may decide what to do with it. Or we can add a method to the service trait (but I think this is more complicated because of the lifetimes going around).


> But in general I'd be fine to tackle that in a separate PR, but perhaps wait until that is also worked on before updating nomos.

Yeah, separated PR and same update on nomos sounds good to me.

ntn-x2:
> I think just signaling to the service that is gonna be terminated is enough. Then service implementors may decide what to do with it. Or we can add a method to the service trait (but I think this is more complicated because of the lifetimes going around).

I'll let @strinnityk find out what method would make most sense and would be easiest to implement and debug. I think my preference would be to have it as part of the service definition, and perhaps provide a default impl for the `ServiceCore` trait which does not do anything, as not to needlessly bother implementers. But yeah maybe we have some lifetime issues, in which case having a channel would be ok. The thing is I would like at least the service implement to know that such a mechanism exist, and to purposefully ignore it if not needed. Just having a channel in there might be a bit "too hard" to see, but if there is no other way, it's still better than having nothing.

strinnityk:
> > I think just signaling to the service that is gonna be terminated is enough. Then service implementors may decide what to do with it. Or we can add a method to the service trait (but I think this is more complicated because of the lifetimes going around).
> 
> I'll let @strinnityk find out what method would make most sense and would be easiest to implement and debug. I think my preference would be to have it as part of the service definition, and perhaps provide a default impl for the `ServiceCore` trait which does not do anything, as not to needlessly bother implementers. But yeah maybe we have some lifetime issues, in which case having a channel would be ok. The thing is I would like at least the service implement to know that such a mechanism exist, and to purposefully ignore it if not needed. Just having a channel in there might be a bit "too hard" to see, but if there is no other way, it's still better than having nothing.

Cognitive load-wise, I'm aligned with @ntn-x2's proposal. A whole different story is how we implement that without trade-offs.

nomos-services/storage/src/lib.rs
```diff
@@ -303,41 +301,40 @@ where
...
+        while let Some(msg) = inbound_relay.recv().await {
+            Self::handle_storage_message(msg, backend).await;
         }
+
         Ok(())
+        // TODO: Implement `Drop` to finish pending transactions and close
+        //  connections gracefully.
+        // Note: Using `Drop` here would only cover the graceful shutdown case.
+        // If the `Stop` is triggered by a user-sent signal, a different
+        // yet-not-implemented-mechanism  is required.
```
danielSanchezQ: Is it? I mean, this should be dropped anyway, the problem is if we move self...

strinnityk: Oh, what I wrote here doesn't make a lot of sense, indeed. I'll double check, just in case.

strinnityk: Oh, yeah, this doesn't make a lot of sense, indeed. I think I was assuming some behaviour in a previous iteration of the Lifecycle. Updating.

When would `self` be moved?

ntn-x2:
🚀🚀🚀

PR #1273, opened 2025-05-15
Use new storage abstraction layer for storage access

andrussal:

Use high level functions(like `get_commitments`) in services to interact with underlying database instead of low level `Load/Store` etc functions.

Followup of [New Storage Abstraction](https://github.com/logos-co/nomos/pull/1267). 

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 1. Description added.

### DIFF
--- a/nomos-da/kzgrs-backend/src/common/share.rs
+++ b/nomos-da/kzgrs-backend/src/common/share.rs
@@ -149,8 +149,9 @@ pub struct DaSharesCommitments {
     pub rows_commitments: Vec<Commitment>,
 }
 
-impl From<Bytes> for DaLightShare {
-    fn from(value: Bytes) -> Self {
-        wire::deserialize::<Self>(&value).expect("Failed to deserialize DaLightShare")
+impl TryFrom<Bytes> for DaLightShare {
+    type Error = wire::Error;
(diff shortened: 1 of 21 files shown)


nomos-da/kzgrs-backend/src/common/share.rs
```diff
@@ -147,3 +148,9 @@ pub struct DaSharesCommitments {
     )]
     pub rows_commitments: Vec<Commitment>,
 }
+
+impl From<Bytes> for DaLightShare {
```
ntn-x2, line 152: Why is this not a `TryFrom` if it can fail?

ntn-x2 (approved):
LGTM

nomos-services/api/src/http/da_shares.rs
```diff
@@ -51,65 +49,68 @@ where
         + Sync
         + Display
         + AsServiceId<StorageService<RocksBackend<StorageOp>, RuntimeServiceId>>,
+    // Service and storage layer types conversions
+    <DaStorageBackend<StorageOp> as StorageDaApi>::BlobId: From<DaShare::BlobId>,
```
danielSanchezQ, line 53: This is not the way to fit generics. There is no need to use `From`, both types should be the same.

PR #1275, opened 2025-05-16
chore: update Rust toolchain to 1.87

ntn-x2:

Fixes https://github.com/logos-co/nomos/issues/1274.

It addresses the new warnings, and explicitly ignores the pedantic `clippy::struct_field_names`, probably for good. 

Nice change: `String::len()` and `Vec::len()` and all methods relying on them are now `const` 🚀

### Note

Risc0 crates are (currently) compiled with rust toolchain 1.85. If we assume that Rust version can differ from the version we use for the rest of the crates, we need to remove them from the workspace, and lint them using the same Rust version that they are built with. The major downside of this is that all the other workspace-wide checks are now not run on those crates anymore.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 1. Description added.


nomos-da/network/core/src/protocols/replication/behaviour.rs
```diff
@@ -80,10 +80,7 @@ impl Clone for ReplicationError {
                     OpenStreamError::Io(error) => {
                         OpenStreamError::Io(std::io::Error::new(error.kind(), error.to_string()))
                     }
-                    err => OpenStreamError::Io(std::io::Error::new(
-                        std::io::ErrorKind::Other,
-                        err.to_string(),
-                    )),
+                    err => OpenStreamError::Io(std::io::Error::other(err.to_string())),
```
danielSanchezQ, line 83: Why does the formatting keep changing? 😅

tests/src/topology/mod.rs
```diff
@@ -160,19 +160,11 @@ impl Topology {
         }
     }
 
-    #[expect(
-        clippy::missing_const_for_fn,
-        reason = "Using `Deref` which is not const."
-    )]
```
danielSanchezQ, line 166: Nice to see this gone!

nomos-services/data-availability/sampling/src/api/http.rs
```diff
@@ -78,10 +78,11 @@ where
...
             api_port: settings.api_port,
-            protocol: settings
-                .is_secure
-                .then(|| "https".to_owned())
-                .unwrap_or_else(|| "http".to_owned()),
+            protocol: if settings.is_secure {
+                "https".to_owned()
+            } else {
+                "http".to_owned()
+            },
```
danielSanchezQ, line 85: why????

danielSanchezQ (approved):
I guess people likes wins....

nomos-da/network/core/src/protocols/replication/behaviour.rs
ntn-x2: No it's not just a formatting change. We changed how we map the inner error. We replaced `std::io::Error::new(std::io::ErrorKind::Other, err.to_string())` with `std::io::Error::other(err.to_string())`

ntn-x2:
FYI since using `self-hosted` does not really play well with switching toolchain, and would require us to re-configure the toolchain every time we switch, I removed all cache and switched to using `ubuntu-latest` for everything but the test step, and it seems everything is green and working, so I will merge with these changes. If stuff starts to break, we will revert the broken step back to using `self-hosted` and adding an explicit override if needed. Going to merge now!

PR #1279, opened 2025-05-20
chore: network-service libp2p backend refactoring

ntn-x2:

This PR introduces the following changes:

* Remove the possibility to get a mutable borrow of the underlying swarm from our libp2p swarm. This means that all operations the swarm provides are exposed via its API (in preparation for the introduction of the NAT traversal black box being worked on)
* Refactor the `nomos-service` crate by removing unnecessary bounds and code blocks
* Add the necessary functions to the swarm to keep it retro-compatible, where they do the same as they did before via `swarm_mut().behaviour_mut()....`.

_Doing any bigger changes to the libp2p swarm (e.g., making it a standalone component which handles all the events and can progress its own state machine) requires a overhaul of the whole crate, which is out of scope for now, but that we can tackle later on to match the design that the DA swarm has followed._

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 1. Description added.


nomos-services/network/src/backends/libp2p/swarm/mod.rs
```diff
@@ -220,13 +218,16 @@ impl SwarmHandler {
     // TODO: Consider a common retry module for all use cases
     fn retry_connect(&mut self, connection_id: ConnectionId) {
         if let Some(mut dial) = self.pending_dials.remove(&connection_id) {
-            dial.retry_count += 1;
+            dial.retry_count = dial
+                .retry_count
+                .checked_add(1)
+                .expect("retry count overflow");
```
pradovic, line 224: Nitpick: there is a check for retry count > MAX_RETRY and retry_count is usize. Do we really need checked_add?

pradovic (approved):
LGTM mod nitpick :)

nomos-services/network/src/backends/libp2p/swarm/mod.rs
ntn-x2: Actually, we only check for `>` relationships, so in the case that 1. `MAX_RETRY` is `usize::MAX` and we reach that value, we would pass the test since `usize::MAX <= usize::MAX` and when we go increment again before storing we overflow. So in the case `MAX_RETRY` is actually `usize::MAX`, but doing a simple addition we would never get out of the recursion.

danielSanchezQ (approved):
Nice! I think this simplifies things around. Thanks!

nomos-services/network/src/backends/libp2p/swarm/mod.rs
pradovic: Maybe I am missing something but in the checked_add case we would panic in this edge case, no?

Maybe we could maybe reorder them like:
```
if dial.retry_count >= MAX_RETRY {  // Check first
    return;
}
dial.retry_count += 1;  
```

ntn-x2: Yes, it would panic instead of being stuck in an infinite loop. But we can change the panic to a return instead. Maybe we should do that.

pradovic:
```
if dial.retry_count >= MAX_RETRY {  // Check first
    return;
}
```

If you do this and then increment after, it should cover the case with MAX_RETRY is usize:MAX and return when it performs usize:MAX retries

ntn-x2: There will soon (not so soon) be a lint coming that will get rid of all potential overflows, since they are indeed quite sneaky in release builds and especially for small numeric types like `u8`s. So I think I'll leave the checked operation.

PR #1281, opened 2025-05-21
feature(http): Http storage adapter

andrussal:

Adds StorageAdapter for http service. Previously http api called storage service directly. This PR adds http service specific StorageAdapter that encapsulates these calls.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 1. Description added.


nomos-services/api/src/http/storage/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+    ) -> Self;
+
+    async fn get_light_share<DaShare>(
+        storage_relay: OutboundRelay<
+            <StorageService<RocksBackend<StorageOp>, RuntimeServiceId> as ServiceData>::Message,
+        >,
+        blob_id: <DaShare as Share>::BlobId,
+        share_idx: <DaShare as Share>::ShareIndex,
+    ) -> Result<Option<<DaShare as Share>::LightShare>, super::DynError>
+    where
```
bacv, line 44: Why we need these constraints on the trait definition, wouldn't it be enough for the implementation to have these constraints?

nomos-services/api/src/http/storage/adapters/rocksdb.rs
```diff
@@ -0,0 +1,297 @@
...
+use tokio::sync::oneshot;
+
+use crate::{http::storage::StorageAdapter, wait_with_timeout};
+
+pub struct RocksAdapter<StorageOp, RuntimeServiceId>
+where
+    StorageOp: StorageSerde + Send + Sync + 'static,
+    RuntimeServiceId: Debug + Sync + Display,
+{
+    pub storage_relay: OutboundRelay<
```
bacv, line 35: In the http backend the `HttpStorageAdapter` is used without initialization, could this be removed? Also if it's not used, the `new` method could also be removed.

bacv (commented):
Just couple questions, otherwise looks good, CI issues seems to be related with runner having old rust version.

nomos-services/api/src/http/storage/mod.rs
bacv: Okay, tried with the associated types, the results would be similar, I'm ok with this solution.

bacv (approved):

PR #1284, opened 2025-05-22
feat(network): Add basic nat behaviour

CHr15F0x:

This PR adds 2 `NetworkBehaviours` to the swarm:
- an autonat v2 server
- a new `Nat` behaviour which for now only includes autonat v2 client without any additional NAT state machine logic.
This PR is the base for a follow up PR which will add the NAT state machine, and will be merged into the [feature branch](https://github.com/logos-co/nomos/tree/feat/nat-traversal).

Yes

@CHr15F0x 

Yes

No.

## Checklist

> [!WARNING]  

* [x] 1. Description added.

### DIFF
--- a/nomos-libp2p/src/behaviour/mod.rs
+++ b/nomos-libp2p/src/behaviour/mod.rs
@@ -30,7 +30,6 @@ pub(crate) struct BehaviourConfig {
     pub kademlia_config: Option<KademliaSettings>,
     pub identify_config: Option<IdentifySettings>,
     pub autonat_client_config: Option<AutonatClientSettings>,
-    pub enable_autonat_server: bool,
     pub protocol_name: ProtocolName,
     pub public_key: identity::PublicKey,
 }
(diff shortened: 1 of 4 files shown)


nomos-libp2p/src/behaviour/nat/behaviour_ext.rs
```diff
@@ -0,0 +1,12 @@
+use libp2p::Multiaddr;
+
+use crate::behaviour::Behaviour;
+
+impl Behaviour {
+    pub(crate) fn add_external_address_candidate(&mut self, addr: Multiaddr) {
+        let Some(nat) = self.nat.as_mut() else {
```
ntn-x2, line 7: Nice to see someone else into `let else` 👌

nomos-libp2p/src/behaviour/nat/swarm_ext.rs
```diff
@@ -0,0 +1,11 @@
+use libp2p::Multiaddr;
+
+use crate::Swarm;
+
+impl Swarm {
+    pub fn add_external_address_candidate(&mut self, addr: Multiaddr) {
```
ntn-x2, line 6: Didn't we say this happens automatically when listening by `Identify`? Why do we need to expose another method? We want to force clients to start listening on a given address, no? I think we can add this method later on when we start dealing with circuit-based NAT traversal. Until then, I don't think we need this, do we?

danielSanchezQ (approved):

PR #1285, opened 2025-05-23
Storage Conversion trait

andrussal:

Move storage and service types conversion behind `Converter` trait. Previously such conversions were done by separate `From` trait implementations which wasn’t very clean approach.

`Converter` is similar to existing `SerdeOp` but works with new storage abstraction layer. Perhaps in future we don’t need  `SerdeOp` and can remove it.

In this PR it’s done only for DA. If this approach seems good then in future I’ll do the same for chain types(which currently is just `Block` though). 

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 1. Description added.

### DIFF
--- a/nomos-services/data-availability/sampling/src/storage/adapters/converter.rs
+++ b/nomos-services/data-availability/sampling/src/storage/adapters/converter.rs
@@ -1,0 +1,60 @@
+use bytes::Bytes;
+use kzgrs_backend::common::share::{DaLightShare, DaShare, DaSharesCommitments};
+use nomos_core::da::{blob::Share, BlobId};
+use nomos_storage::{
+    api::da::DaConverter,
+    backends::{rocksdb::RocksBackend, StorageSerde},
+};
+
+type ShareIndex = [u8; 2];
+
+pub struct DaStorageConverter;
+
+impl<SerdeOP> DaConverter<RocksBackend<SerdeOP>> for DaStorageConverter
+where
+    SerdeOP: StorageSerde + Send + Sync + 'static,
+    <SerdeOP as StorageSerde>::Error: Send + Sync + 'static,
+{
+    type Share = DaShare;
+    type Error = SerdeOP::Error;
+
+    fn blob_id_to_storage(blob_id: BlobId) -> Result<BlobId, Self::Error> {
+        Ok(blob_id)
+    }
+
+    fn blob_id_from_storage(blob_id: BlobId) -> Result<BlobId, Self::Error> {
+        Ok(blob_id)
+    }
+
+    fn share_index_to_storage(
+        share_index: <DaShare as Share>::ShareIndex,
+    ) -> Result<ShareIndex, Self::Error> {
+        Ok(share_index)
+    }
+
+    fn share_index_from_storage(
+        share_index: ShareIndex,
+    ) -> Result<<DaShare as Share>::ShareIndex, Self::Error> {
+        Ok(share_index)
+    }
+
+    fn share_to_storage(service_share: DaLightShare) -> Result<Bytes, Self::Error> {
+        Ok(SerdeOP::serialize(&service_share))
+    }
+
+    fn share_from_storage(backend_share: Bytes) -> Result<DaLightShare, Self::Error> {
+        SerdeOP::deserialize(backend_share)
+    }
+
+    fn commitments_to_storage(
+        service_commitments: DaSharesCommitments,
+    ) -> Result<Bytes, Self::Error> {
+        Ok(SerdeOP::serialize(&service_commitments))
+    }
+
+    fn commitments_from_storage(
+        backend_commitments: Bytes,
+    ) -> Result<DaSharesCommitments, Self::Error> {
+        SerdeOP::deserialize(backend_commitments)
+    }
+}
--- a/nomos-services/data-availability/sampling/src/storage/adapters/mod.rs
+++ b/nomos-services/data-availability/sampling/src/storage/adapters/mod.rs
@@ -1,2 +1,3 @@
+pub mod converter;
 #[cfg(feature = "rocksdb-backend")]
 pub mod rocksdb;
--- a/nomos-services/storage/src/api/da/mod.rs
+++ b/nomos-services/storage/src/api/da/mod.rs
@@ -1,8 +1,42 @@
 use std::{collections::HashSet, error::Error};
 
 use async_trait::async_trait;
+use nomos_core::da::blob::Share;
 
 pub mod requests;
+
+pub trait DaConverter<B: StorageDaApi> {
+    type Share: Share;
+    type Error: Error + Send + Sync + 'static;
+
+    fn blob_id_to_storage(
+        blob_id: <Self::Share as Share>::BlobId,
+    ) -> Result<B::BlobId, Self::Error>;
+
+    fn blob_id_from_storage(
+        blob_id: B::BlobId,
+    ) -> Result<<Self::Share as Share>::BlobId, Self::Error>;
+
+    fn share_index_to_storage(
+        share_index: <Self::Share as Share>::ShareIndex,
+    ) -> Result<B::ShareIndex, Self::Error>;
+
+    fn share_index_from_storage(
+        share_index: B::ShareIndex,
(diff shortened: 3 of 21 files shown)


nomos-services/data-availability/sampling/src/storage/adapters/mod.rs
```diff
@@ -1,2 +1,4 @@
 #[cfg(feature = "rocksdb-backend")]
+pub mod converter;
```
bacv, line 2: Nit, would it make sense to create `/adapters/rocksdb/mod.rs` and have adapter and converter as one module?

bacv (approved):
Just style related nits, otherwise looks great.

nomos-services/data-availability/sampling/src/storage/adapters/converter.rs
```diff
@@ -0,0 +1,60 @@
...
+    ) -> Result<Bytes, Self::Error> {
+        Ok(SerdeOP::serialize(&service_commitments))
+    }
+
+    fn commitments_from_storage(
+        backend_commitments: Bytes,
+    ) -> Result<DaSharesCommitments, Self::Error> {
+        SerdeOP::deserialize(backend_commitments)
+    }
+}
```
danielSanchezQ, line 60: What is the befit of this? This basically grouping `From<Type> for DBType` which would be more generic and less cumbersome to maintain.

andrussal: We could keep `From` style as well. But it seemed cleaner if these kind of conversions are in single location because it's storage specific functionality and adding `From` implementations to each struct separately feels like putting this functionality where it doesn't really belong. I started to think it when I was going to look into zero copy stuff and having all this conversion stuff in one place feels somehow better. But just personal preference I guess.

It might be easier to maintain this way because now we don't have to update so many bounds in type definitions. But not 100% sure.

danielSanchezQ: Yes, I see this may reduce the bounds elsewhere. But I do not see how types are tied-up together. If we change any of the sides we will have to change them here and refactor everything I think. Anyway it's not too much as it is just used for specific storage types.

nomos-services/data-availability/sampling/src/storage/adapters/converter.rs
```diff
@@ -0,0 +1,60 @@
+use bytes::Bytes;
+use kzgrs_backend::common::share::{DaLightShare, DaShare, DaSharesCommitments};
+use nomos_core::da::{blob::Share, BlobId};
+use nomos_storage::{
+    api::da::DaConverter,
+    backends::{rocksdb::RocksBackend, StorageSerde},
+};
+
+type ShareIndex = [u8; 2];
```
ntn-x2, line 9: Isn't this type actually defined by each trait impl?

nomos-services/storage/src/api/da/mod.rs
```diff
@@ -1,9 +1,43 @@
 use std::{collections::HashSet, error::Error};
 
 use async_trait::async_trait;
+use nomos_core::da::blob::Share;
 
 pub mod requests;
 
+pub trait DaConverter<B: StorageDaApi> {
```
ntn-x2, line 8: Wasn't the new convention to avoid single letter generics? Cc @danielSanchezQ

nomos-services/storage/src/api/da/mod.rs
```diff
@@ -1,9 +1,43 @@
...
+    fn blob_id_to_storage(
+        blob_id: <Self::Share as Share>::BlobId,
+    ) -> Result<B::BlobId, Self::Error>;
+
+    fn blob_id_from_storage(
+        blob_id: B::BlobId,
+    ) -> Result<<Self::Share as Share>::BlobId, Self::Error>;
+
+    fn share_index_to_storage(
+        share_index: <Self::Share as Share>::ShareIndex,
```
ntn-x2, line 21: I think one type alias for each of the different share index types would benefit a lot in this case.

ntn-x2 (commented):
I don't have a strong opinion on the strategy here, I just had a couple of nits that might help understand the design choice better. Otherwise, LGTM!

nomos-services/storage/src/api/da/mod.rs
danielSanchezQ: Not sure about the new conversion. But we should avoid (in fact is forbidden) to use single letter generics anymore. And if working on something there are legacy single letter ones, lets please just refactor them in place. Is a common effort.

nomos-services/data-availability/sampling/src/storage/adapters/mod.rs
andrussal: Yes this is nicer, moved to module.

nomos-services/data-availability/sampling/src/storage/adapters/converter.rs
andrussal: It wasn't really needed, removed it altogether, thanks!

andrussal:
I am gonna merge this then. Left some thoughts below.

I think I understand the concern that we have to change these abstractions if we change service data types. This is true for other things as well in the new storage layer, not only conversion stuff. For example, if we merge da share and commitments together like it was before(just hypothetical example), then we’d have to change `DaStorageApi` as well(remove functions to store and retrieve commitments and enums variants to call these). 

But I think the general design stays in place even if we change service types. We don’t have to remove existing abstractions from our design, only to adjust them. (And each domain has its own storage api and corresponding converter. If one domain changes, it doesn’t affect others).

This is sort of the result of not calling KV DB set/get functions directly and having high level functions instead, if service types change, we need to adjust this high level api.

There are maybe some things that aren’t strictly necessary, like helper functions to fill enums for specific api calls. Or maybe we even don’t need enums for each storage api call. We could come up with some generic builder pattern that takes storage function name and its arguments. (But probably helper functions and enums are fine as long we don’t need to change types very often).

One reason why converter(adapter pattern) seemed like a good idea was to keep conversion logic away from locations where types are defined because it’s storage specific logic and location where service types are defined shouldn’t know about storage format. Currently it’s `From<Bytes>` but it could be also something like `From<SomeOtherDbFormat>` , which felt weird if it’s in module where service type is defined. Like `Block` in `nomos-core` would have dependency to some storage thing(`OtherBlockDbFormat`). 

_And I guess if people find it cumbersome to build new storage stuff like this, we still have old KV calls available and this new style would just sort of quietly disappear if not used._

---

**Outlining the new design here as well.**

**Abstract types(for DA):**
- `DaStorageApi` - trait with associated _storage types_ and functions to store and retrieve these types
- `DaStorageConverter` - trait with associated _service types_ to convert to/from `DaStorageApi` types. As `DaStorageConverter<ServiceTypes, StorageTypes>`.  It “matches” `ServiceTypes` with `StorageTypes`

(Above mentioned `ServiceTypes` is actually fictional. We don’t have `ServiceTypes` trait. It’s just a list of individual service types(like `DaShare`)).

**Concrete types(using RocksDb):**
 - `DaStorageApi` impl - works with RocksDb types(Bytes) and its KV api
 - `DaStorageConverter` impl - converts between DA service types and RocksDb types. 

**“Utilities”**
- Enums to send messages for calling respective `DaStorageApi` functions
- Utility functions to instantiate these enums

It's based on existing design how we interact with storage backend by sending messages. It introduces storage API with high level functions and conversion between service types and storage types.

---

**Some common reasons when something here needs to be changed(and where)**

1. Introduce different storage backend.
    1. New `DaStorageApi` and  `DaStorageConverter` implementations
2. Store the same types differently
    1. Add new functions to `DaStorageApi` and to `DaStorageConverter`
3. Modify service types
    1. If introducing new types, need to add these to both traits(related storage functions and converting)
    2. If modifying the types in a way that is externally visible(like splitting or grouping), need to modify both traits
    3. If just adding/removing  fields, may only need to change impl internals


I think the only concerning item is 3.2. because it means refactoring not adding functionality. But if this doesn’t happen often, it should not be an issue. (Also I think in this case it means refactoring absolutely all the other places in codebase as well where these types are used, not only storage. So it would be some bigger refactoring anyway.). 

But the structure/design how everything is working together doesn’t have to change in any of these cases, only details.

PR #1295, opened 2025-05-27
DA and Membership integration

pradovic:

This PR is adding membership service adapter to DA, and is integrating DA with membership service. The goal is for Da MambershipHandler to be updated when membership is updated.

This should be intro to enabling historic members to the DA as well.
The scope of this PR is to integrate membership service with DA and for current functionality to pass (meaning there is still no real updates), but the initial peers should be able to be loaded from membership as well.

Yes. This code offers opinionated solution where arc swap is used to swap the shared Arc<Membership> that is used by da components in a read only fashion, and this Membership is swapped (updated) only by the da network service on an event from membership (when membership latest is changed).

@pradovic

Yes.
No.

## Checklist

> [!WARNING]  

* [x] 1. Description added.


pradovic:
@bacv, @danielSanchezQ This is very unpolished version of the approach I wanted to try to implement membership in DA. I will add a few comments about couple of questions I had along the way. I thought it would be easier to see it in PR and then discuss before moving on to write new tests and try to simplify the types and do some refactoring.

nomos-da/network/subnetworks-assignations/src/lib.rs
```diff
@@ -31,6 +34,9 @@ pub trait MembershipHandler {
     fn last_subnetwork_id(&self) -> Self::NetworkId;
 
     fn get_address(&self, peer_id: &PeerId) -> Option<Multiaddr>;
+
+    #[must_use]
+    fn rebuild_with(&self, members: Vec<PeerId>, addressbook: HashMap<PeerId, Multiaddr>) -> Self;
```
danielSanchezQ, line 39: Why this instead of a replacement? I think this is not the way to go probably. The idea was to implement membership for an adapter of the membership service itself. Should be necessary to have a rebuild as it could be done internally.

nomos-sdp/src/lib.rs
```diff
@@ -46,7 +52,7 @@ pub enum ServiceType {
 
 pub type Nonce = [u8; 16];
 
-#[derive(Clone, Copy, Eq, PartialEq, Hash, Debug)]
+#[derive(Clone, Copy, Eq, PartialEq, Hash, Debug, Serialize, Deserialize)]
```
danielSanchezQ, line 55: We need to take care of how this is done. Probably it should be transparent here, otherwise is gonna be wrapped.

nomos-sdp/src/lib.rs
```diff
@@ -35,9 +36,14 @@ impl Locator {
...
     }
+
+    #[must_use]
+    pub const fn addr(&self) -> &Multiaddr {
+        &self.addr
+    }
 }
 
-#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
```
danielSanchezQ, line 46: In this case we should use some `repr(c)` or `repr(u8)`, this is gonna go through wire right?

nomos-services/data-availability/network/src/backends/libp2p/common.rs
```diff
@@ -179,3 +181,67 @@ pub(crate) async fn handle_balancer_command<Stats: Debug>(
         error!("Error stats request: {cmd:?}");
     }
 }
+
+use arc_swap::ArcSwap;
+
+pub struct SwappableMembershipHandler<T: MembershipHandler> {
```
pradovic, line 187: The main idea of how to update membership is to ArcSwap it in `SwappableMembershipHandler`. Da network service is initializing the handler behind the Arc<SwappableMembershipHandler> and is passing membership.clone() to swarms (and swarm is passing further to components).

The idea is that on membership service update event we can swap this from network service. 

@bacv , @danielSanchezQ Pinging you guys here just to sanity check the idea. It should be simple to use locking instead, but this seemed like a good fit as we have single writer and multiple readers. I will update the tests to verify the swap on initialization (passing the initial peers to membership instead of directly to `FillFromNodeList`. Which should test the startup + the subscribe as subscribe will set the initial latest state.

nomos-services/data-availability/network/src/backends/libp2p/common.rs
```diff
@@ -179,3 +181,67 @@ pub(crate) async fn handle_balancer_command<Stats: Debug>(
...
+        members: Vec<PeerId>,
+        addressbook: std::collections::HashMap<PeerId, Multiaddr>,
+    ) -> Self {
+        let inner = self.inner.load_full();
+        let inner = inner.rebuild_with(members, addressbook);
+        Self {
+            inner: ArcSwap::new(inner),
+        }
+    }
+}
```
danielSanchezQ, line 247: why this? Behaviours gonna share it? Who takes responsability of the swap then?

nomos-services/data-availability/network/src/backends/libp2p/validator.rs
```diff
@@ -68,15 +69,18 @@ pub enum DaNetworkEvent {
...
+{
     task: (AbortHandle, JoinHandle<Result<(), Aborted>>),
     replies_task: (AbortHandle, JoinHandle<Result<(), Aborted>>),
     sampling_request_channel: UnboundedSender<(SubnetworkId, BlobId)>,
     balancer_command_sender: UnboundedSender<ConnectionBalancerCommand<BalancerStats>>,
     monitor_command_sender: UnboundedSender<ConnectionMonitorCommand<MonitorStats>>,
     sampling_broadcast_receiver: broadcast::Receiver<SamplingEvent>,
     verifying_broadcast_receiver: broadcast::Receiver<DaShare>,
-    _membership: PhantomData<Membership>,
+    membership: Arc<SwappableMembershipHandler<Membership>>,
```
danielSanchezQ, line 83: If `SwappableMembershipHandler` is an `Arc` already, why are we wrapping in `Arc` again?

nomos-services/data-availability/network/src/lib.rs
```diff
@@ -53,59 +62,88 @@ impl<B: NetworkBackend<RuntimeServiceId>, RuntimeServiceId> Debug
...
+
+        let membership_adapter = Membership::new(membership_relay);
+
+        let mut stream = membership_adapter
+            .subscribe(ServiceType::DataAvailability)
+            .await
+            .map_err(|e| {
+                tracing::error!("Failed to subscribe to membership service: {:?}", e);
+                e
+            })?;
```
danielSanchezQ, line 141: This adapter could implement `MembershipHandler` as well I think

nomos-services/data-availability/network/src/lib.rs
```diff
@@ -147,6 +190,30 @@ where
                 }),
         }
     }
+
+    fn handle_membership_update_message(backend: &mut Backend, update: &MembershipProviders) {
```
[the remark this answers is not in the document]
danielSanchezQ: No we cannot, the `PeerId` should be derived from the locators themselves iirc.

nomos-services/data-availability/network/src/lib.rs
```diff
@@ -147,6 +190,30 @@ where
...
+        if update.is_empty() {
+            tracing::debug!("Received empty membership update, skipping.");
+            return;
+        }
```
danielSanchezQ, line 198: This is a code smell in my opinion. Probably this situations should not happen.

nomos-services/data-availability/network/src/lib.rs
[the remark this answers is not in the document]
bacv: > No we cannot, the PeerId should be derived from the locators themselves iirc.

This part is not specified and probably would be better to document. I would argue that ProviderId should be used for PeerId, that would minimize data on chain, and since SDP protocol allows addresses to be in `ip:port` format (not only multiaddr), then ProviderId would be used to check the quic connection certs.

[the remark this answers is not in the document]
danielSanchezQ: I don't think it should be like that, you are linking logical identity with network identity which probably you do not want to. Let's discuss in discord.

nomos-services/data-availability/network/src/lib.rs
pradovic: MembershipHandler describes API on the subnetworks assigned membership snapshot.  
Adapter is used to subscribe to membership changes or get snapshot at (later for historical). When the snapshot is received we need to create new subnetwork assigned membership (da membership) for it. That is why there is one trait for MembershipAdapter which is used to subscribe to membership service, and another trait MembershipHandler used by da network components (behaviors) that use membership handler. 
Does that makes sense?

nomos-services/data-availability/network/src/backends/libp2p/common.rs
pradovic: Behaviors share the read only `MembershipHandler`, but da network service sees `SwappableMembershipHandler` and can use update to update it (swap the whole membership with the new one, and behaviors will just see the new one since then).

nodes/nomos-executor/executor/src/lib.rs
```diff
@@ -45,34 +54,73 @@ pub(crate) type TracingService = Tracing<RuntimeServiceId>;
...
+    RuntimeServiceId,
+>;
+
 pub(crate) type BlendService = nomos_blend_service::BlendService<
     BlendBackend,
     BlendNetworkAdapter<RuntimeServiceId>,
     RuntimeServiceId,
 >;
 
+pub(crate) type MembershipService = nomos_membership::MembershipService<
```
bacv, line 69: In `nomos-node` we have `generic_services.rs`, where we could have membership and sdp services defined and then imported here in the executor. Seems that the definitions are the same.

nodes/nomos-executor/executor/src/lib.rs
```diff
@@ -45,34 +54,73 @@ pub(crate) type TracingService = Tracing<RuntimeServiceId>;
...
+        SdpLedger<LedgerDeclarationAdapter, LedgerServicesAdapter, Metadata>,
+        LedgerDeclarationAdapter,
+        LedgerServicesAdapter,
+        Metadata,
+        RuntimeServiceId,
+    >,
+    RuntimeServiceId,
+>;
+
+pub(crate) type SdpService = nomos_sdp::SdpService<
```
bacv, line 93: We are adding sdp service here but it's not actually used by membership service, right? At the moment everything is mocked and read from the file?

nomos-sdp/src/lib.rs
bacv: As I understand the service definition should be 2 bytes, or two chars according current specification.

nomos-services/data-availability/network/src/adapters/membership/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+
+pub mod membership_service;
+
+#[derive(Debug, Error)]
+pub enum MembershipAdapterError {
+    Backend(MembershipBackendError),
+    Other(DynError),
+}
+
+impl std::fmt::Display for MembershipAdapterError {
```
bacv, line 18: Lets use `thiserror` for `MembershipAdapterError` and other similar errors.

nomos-services/data-availability/network/src/backends/libp2p/common.rs
bacv: I think the idea is good, I'm just bit confused by the 3 arcs that used in the `inner`.
Do I understand correctly that `SwappableMembershipHandler` is only updated from the da network service, and swarm with it's behaviours get's clones of this membership, but all clones point to the same `inner`?

testnet/cfgsync/src/config.rs
```diff
@@ -224,6 +241,70 @@ fn update_tracing_identifier(
     }
 }
 
+#[must_use]
+pub fn create_membership_configs(ids: &[[u8; 32]], hosts: &[Host]) -> Vec<GeneralMembershipConfig> {
```
bacv, line 245: Nice

nomos-services/data-availability/network/src/backends/libp2p/common.rs
pradovic: Exactly, that sounds right, network service is doing update and creating assignations (MembershipHandler), while the validator, executor and it's behaviors are reading it.  I don't think we can share it to the behaviors if it is not behind an Arc. There is one Arc in  `SwappableMembershipHandler` that holds `ArcSwap`, and there is one Arc that it inside the `ArcSwap` (by design), and it enables `ArcSwap` to work. The derived Copy on `SwappableMembershipHandler` will do arc clone on `membership.copy()`. It is easy to see from da network lib.rs and executor and validator backend.  I dont see a 3rd arc 🙈 ?

nodes/nomos-executor/executor/src/lib.rs
pradovic: Everything is read from the membership service config, but there is still a dep on sdp service. So, sdp is still not used, just for init of the membership (no actual updates will take place) and membership should contain only the files it read from config.

nomos-sdp/src/lib.rs
pradovic: We probably don’t need to use the same serialization format for over-the-wire communication and for how values are stored/read in SDP, right? So we could use #[repr(u8)] or raw byte codes when sending over the wire, and still implement custom (de)serialization to/from 2-byte codes when interacting with config or the ledger.

bacv: The `nomos-sdp` is core library and it should contain serialization for over the wire comms. If we go this path, then I'd keep definitions here as per spec, this is defined this way in spec for future expansion. The [Mantle PR](https://github.com/logos-co/nomos/pull/1310) already makes the assumption that SDP lib will implement the serialization, maybe we could just revert this change for now if it's not critical for this PR?

danielSanchezQ: Yes, this is already taka care in the mantle pr. Best to just revert this.

pradovic: Hmm, I think I needed to actually add it for my tests to pass. At least Serialization from config files. Do you guys suggest I remove serialization completely from this PR or just wire stuff?

bacv: Maybe lets just copy paste whats done in the MantleTx? That pr will soon be merged, it will be easy to resolve later.

nomos-da/network/subnetworks-assignations/src/lib.rs
danielSanchezQ: You probably can skip the vector of peers as you can just get it from the hashmap keys.

nomos-sdp/src/lib.rs
pradovic: sgtm!

nomos-da/network/subnetworks-assignations/src/lib.rs
danielSanchezQ: > Does that make sense? Could you elaborate more on the implementing membership for an adapter of membership service itself idea?

The membership listing can be updated. But in this case Behaviours that just read them doesnt need to know that it can be updated. They just need to pull info from it. It is not that is not gonna work. We can let it like this for now, but its a flawed design imo.

nomos-sdp/src/lib.rs
danielSanchezQ: > Hmm, I think I needed to actually add it for my tests to pass. At least Serialization from config files. Do you guys suggest I remove serialization completely from this PR or just wire stuff?

Why is this in config files? 🤔 
Anyway, serialization/deserialization shouldnt be necessary for this.  I don't understand why sdp is mixed in this da membership integration when this is supposed to be mocked.

nomos-da/kzgrs-backend/src/encoder.rs
```diff
@@ -409,7 +409,7 @@ pub mod test {
     }
 
     #[test]
-    #[should_panic(expected = "Encoding only zeroes is not allowed in Nomos DA network.")]
+    #[should_panic(expected = "assertion failed: polynomial_degree.is_power_of_two()")]
```
danielSanchezQ, line 412: why is this in this PR? or this change comes from a merge?

pradovic: when I merged master it breaked clippy warning for no message on should panic

danielSanchezQ: But this shouldn't be fixed here. You should have open a pr on master, then re-rebase this branch if its crucial.

PR #1297, opened 2025-05-28
fix: pin the reqwest version until the compilation issue is resolved

pradovic:
Seems to be a new .16 version issue that should be resolved soon, but currently fails our CI.

https://github.com/seanmonstar/reqwest/issues/2695


bacv:
> @bacv I think it was me by accident :) I fixed it

I was just curious, seems that we need that formatting.

PR #1300, opened 2025-05-28
Add Bootstrap / Online modes to Cryptarchia

zeegomo:
This PR adds the Bootstrap and Online modes of operation to Cryptarchia, as specified in https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6 and following the reference spec in https://github.com/logos-co/nomos-specs/pull/127
It also adds a tracker for the Latest Immutable Block for future integration in the implementation

Spec is https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6
Reference implementation is https://github.com/logos-co/nomos-specs/pull/127

@youngjoon-lee @zeegomo @davidrusu 

Yes

Adding an explicit LIB to Cryptarchia as it may be useful for the implementation to avoid recomputing it each time
## Checklist

> [!WARNING]  

* [ ] 1. Description added.

### DIFF
--- a/consensus/cryptarchia-engine/src/lib.rs
+++ b/consensus/cryptarchia-engine/src/lib.rs
@@ -13,12 +13,123 @@ pub use config::*;
 use thiserror::Error;
 pub use time::{Epoch, EpochConfig, Slot};
 
+#[derive(Clone, Debug, Copy)]
+pub struct Boostrapping;
+#[derive(Clone, Debug, Copy)]
+pub struct Online;
+
+pub trait State: Copy + std::fmt::Debug {
+    fn fork_choice<Id>(cryptarchia: &Cryptarchia<Id, Self>) -> Branch<Id>
+    where
+        Id: Eq + std::hash::Hash + Copy;
+    fn lib<Id>(cryptarchia: &Cryptarchia<Id, Self>) -> Id
+    where
+        Id: Eq + std::hash::Hash + Copy;
+}
+
+impl State for Boostrapping {
+    fn fork_choice<Id>(cryptarchia: &Cryptarchia<Id, Self>) -> Branch<Id>
+    where
+        Id: Eq + std::hash::Hash + Copy,
+    {
+        let k = cryptarchia.config.security_param.get().into();
+        let s = cryptarchia.config.s();
+        maxvalid_bg(cryptarchia.local_chain.clone(), &cryptarchia.branches, k, s)
+    }
+
+    fn lib<Id>(cryptarchia: &Cryptarchia<Id, Self>) -> Id
+    where
+        Id: Eq + std::hash::Hash + Copy,
+    {
+        cryptarchia.lib
+    }
+}
+
+impl State for Online {
+    fn fork_choice<Id>(cryptarchia: &Cryptarchia<Id, Self>) -> Branch<Id>
+    where
+        Id: Eq + std::hash::Hash + Copy,
+    {
+        let k = cryptarchia.config.security_param.get().into();
+        maxvalid_mc(cryptarchia.local_chain.clone(), &cryptarchia.branches, k)
+    }
+
+    fn lib<Id>(cryptarchia: &Cryptarchia<Id, Self>) -> Id
+    where
+        Id: Eq + std::hash::Hash + Copy,
+    {
+        cryptarchia
+            .branches
+            .nth_ancestor(
+                &cryptarchia.local_chain,
+                cryptarchia.config.security_param.get().into(),
+            )
+            .id()
+    }
+}
+
+// Implementation of the fork choice rule as defined in the Ouroboros Genesis
+// paper k defines the forking depth of chain we accept without more
+// analysis s defines the length of time (unit of slots) after the fork
+// happened we will inspect for chain density
+fn maxvalid_bg<Id>(local_chain: Branch<Id>, branches: &Branches<Id>, k: u64, s: u64) -> Branch<Id>
+where
+    Id: Eq + std::hash::Hash + Copy,
+{
+    let mut cmax = local_chain;
+    let forks = branches.branches();
+    for chain in forks {
+        let lowest_common_ancestor = branches.lca(&cmax, &chain);
+        let m = cmax.length - lowest_common_ancestor.length;
+        if m <= k {
+            // Classic longest chain rule with parameter k
+            if cmax.length < chain.length {
+                cmax = chain;
+            }
+        } else {
+            // The chain is forking too much, we need to pay a bit more attention
+            // In particular, select the chain that is the densest after the fork
+            let density_slot = Slot::from(u64::from(lowest_common_ancestor.slot) + s);
+            let cmax_density = branches.walk_back_before(&cmax, density_slot).length;
+            let candidate_density = branches.walk_back_before(&chain, density_slot).length;
+            if cmax_density < candidate_density {
+                cmax = chain;
+            }
+        }
+    }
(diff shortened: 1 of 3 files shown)


consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -123,22 +242,31 @@ where
         // issues? We are calculating length here but the `self.branches` is
         // cloned in the `Self::apply_header_unchecked` method, which means
         // there's a risk of length being different due to concurrent operations.
```
youngjoon-lee: I think this comment should be moved to the line 262 where the `length` is calculated.

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -175,7 +303,41 @@ where
...
+        let mut current = b;
+        if a.id == b.id {
+            return true; // `a` is the same as `b`
+        }
+        // Walk up the chain from `b` until we find `a` or reach the root
+        while current.parent != current.id {
+            if current.parent == a.id {
+                return true; // Found `a` in the chain
+            }
+            current = &self.branches[&current.parent];
```
youngjoon-lee, line 319:
```suggestion
            current = &self.branches[&current.parent];
            if current.length() <= a.length() {
                break;
            }
```
What do you think about adding an end condition like this to stop the iteration earlier?

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -316,14 +464,33 @@ where
     }
 }
 
+impl<Id> Cryptarchia<Id, Boostrapping>
+where
+    Id: Eq + std::hash::Hash + Copy,
+{
+    /// Signal transitioning to the online state.
+    pub fn online(self) -> Cryptarchia<Id, Online> {
```
youngjoon-lee, line 472: 👍

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -175,7 +303,41 @@ where
...
+        }
+        false // `a` is not an ancestor of `b`
+    }
+
+    // Returns the min(n, A)-th ancestor of the provided block, where A is the
+    // number of ancestors of this block.
+    fn nth_ancestor(&self, branch: &Branch<Id>, mut n: u64) -> Branch<Id> {
+        let mut current = branch;
+        while n > 0 {
+            n -= 1;
```
ntn-x2, line 329: I think it would be a cleaner approach to use `checked_sub` and break out of the loop when that's `None`.

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -13,21 +13,131 @@ pub use config::*;
...
 pub use time::{Epoch, EpochConfig, Slot};
 
+#[derive(Clone, Debug, Copy)]
+pub struct Boostrapping;
+#[derive(Clone, Debug, Copy)]
+pub struct Online;
+
+pub trait State: Copy + std::fmt::Debug {
+    fn fork_choice<Id>(cryptarchia: &Cryptarchia<Id, Self>) -> Branch<Id>
+    where
```
ntn-x2, line 23: I don't think we need this `where` clauses here, since neither `Cryptarchia` nor `Branch` have them in their struct definitions, right?

ntn-x2: If they are needed, then `Id` could become a generic type for the trait, and the impl block could have the `where` clause instead.

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -186,11 +348,16 @@ pub enum Error<Id> {
...
     #[error("Orphan proof has was not found in the ledger: {0:?}, can't import it")]
     OrphanMissing(Id),
+    #[error("Attempting to fork immutable history at {0:?}")]
+    ImmutableFork(Id),
+    #[error("Invalid slot for block {0:?}, parent slot is greater than child slot")]
+    InvalidSlot(Id),
 }
 
-impl<Id> Cryptarchia<Id>
+impl<Id, St> Cryptarchia<Id, St>
```
danielSanchezQ: Nit: We decided to try to avoid single (or double in this case) letter generics.
```suggestion
impl<Id, State> Cryptarchia<Id, State>
```
Maybe we can call it `Mode` as well.

consensus/cryptarchia-engine/src/lib.rs
zeegomo: mmm, not a fan of that, why would I intentionally trigger an overflow just to check when it's zero?

consensus/cryptarchia-engine/src/lib.rs
zeegomo: The requirement is for the fork choice rule used inside. 
I don't like it being an associated type because the state does not really care about `Id` at all, it makes more sense for it to be local to the invocation of the fc rule instead of being declared as part of the trait impl.
In theory, it could be its own trait like
```
trait Id: Eq + Hash {}
```
but I don't think this is really needed and a simple where clause will do, it's all local after all

consensus/cryptarchia-engine/src/lib.rs
zeegomo: `State` clashes with the trait defined above and I'd personally like to keep the names consistent, not a fan of the trait being called `State` and the type param `Mode`. It's called `State` in the struct def and this should not propagate much. Anyway, open to change ofc

danielSanchezQ: We could rename the trait to `CryptarchiaState` and let the generic be `State`. I know in here is kinda easy to see that `St` meant state, but after some time is better to have it explicit, or someone may think it is saint or something.

consensus/cryptarchia-engine/src/lib.rs
ntn-x2: I was not suggesting to make it an associated type. It would not make sense indeed. I was suggesting to make it a generic, which means it's not the implementer deciding what the type is, but the invoker. That would allow you to remove the `where` clause from the trait definition, since an implementor that wants to implement the trait using a different method that does not require those bounds is forced to introduce the bounds because the trait is defined in that way. Making it a generic means that implementors that needs those bounds, will introduce them, and implementors that do not need them, do not have to care. 

I just wanted to clarify my suggestion to avoid any misunderstandings. This being said, feel free to continue with your proposed solution if you believe it's a better one, in this case.

consensus/cryptarchia-engine/src/lib.rs
ntn-x2: Yeah in terms of efficiency checking for an underflow is probably slightly less performant than checking if `> 0` and then decrementing. It was more about an alternative approach which feels cleaner in my opinion, also foreseeing we might want to remove unsafe arithmetic operations altogether at some point with the enacting of a few lints. But this approach works fine as well.

consensus/cryptarchia-engine/src/lib.rs
zeegomo: Oh I see, I misunderstood what you meant indeed. Yeah that makes sense.
One downside of that approach is that now you have to specify the `Id` bound everywhere, and we already have an issue with generics everywhere. I'd like to keep it like this for now, I don't think it' s a big issue anyway. If you (or anybody else) see a point for changing it in the other PR  or future changes please go ahead

PR #1310, opened 2025-05-29
feature(mantle): Mantle core crate structure

danielSanchezQ:

The new mantle transactions with its Op are abstracted and implemented from the current [spec](https://www.notion.so/Mantle-Specification-1fd261aa09df81bba3d4f99f38e1004b)

This includes:
* Overall `MantleTx` and `SignedMantleTx`
* `Op`'s including an empty `Native`
* Gas prices calculations

It does **not include**:
* The `LedgerTx` implementation, as we may decide to take it into the new cryptarchia ledger structure (cc: @zeegomo ).
* Any verification (which can/will be added later).
* Any zk signatures or proofs.

It should, it is quite directly related to the spec itself. 

@danielSanchezQ , support from @bacv , @zeegomo as the mantle owner.

~~There was some gas calculations that may have changed due to modifications in sdp. @davidrusu should check.~~ Done, it in sync with latest spec.

~~Yes, of the above mentioned gas calculations.~~ This was synced already.

## Checklist

* [x] 1. Description added.


danielSanchezQ:
@bacv this is what I had in mind for mantle implementation. I used the `Blob` tx as example. A lot of stuff remaining ofc. I have other things in mind on how to handle incoming messages. But lets just validate the idea.
cc @zeegomo . 
This would be a replacement of #1307

nomos-mantle/src/ops/blob.rs
```diff
@@ -0,0 +1,14 @@
+use serde::{Deserialize, Serialize};
+
+use crate::ops::{ChannelId, Ed25519PublicKey, TxHash};
+
+pub type BlobId = [u8; 32];
+
+#[derive(Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct BlobOp {
```
bacv, line 8: Shouldn't this be in kzgrs-backend?

bacv: 1) We contain all operations in mantle crate, including SDP messages, that way nomos-sdp crate would import mantle crate and would only define states that are used in ledger. DA verifier, indexer, sampler would also import BlobOp from mantle crate (together with blob_id we need channel_id).
2) SDP crate contains sdp messages, DA contains blob message, that way services using sdp or da blobs wouldn't know anything about mantle crate. Something like this:
```
consensus <--- mantle <--- sdp core ---> sdp services
consensus <--- mantle <---   da core ---> da services
```

danielSanchezQ: > 1. We contain all operations in mantle crate, including SDP messages, that way nomos-sdp crate would import mantle crate and would only define states that are used in ledger. DA verifier, indexer, sampler would also import BlobOp from mantle crate (together with blob_id we need channel_id).

DA verifier, indexer, sampler doesn't need to deal with this (probably). This is just part of the tx validation. DA is only really concerned in the BlobId, or maybe in different parts. About sdp, it depends how it is used. If we do not inmidiatly consume the messages per se, then we could move them to mantle.

>     2. SDP crate contains sdp messages, DA contains blob message, that way services using sdp or da blobs wouldn't know anything about mantle crate. Something like this:

Yeah, this sounds like the optimal. But if you check on the specifications in DA we do not have any message defined per se. But we do in SDP.

nomos-mantle/src/ops/sdp.rs
```diff
@@ -0,0 +1,26 @@
...
+    }
+}
+
+impl GasPrice for SDPActiveOp {
+    fn gas_price<Constants: GasConstants>(&self) -> Gas {
+        Constants::ACTIVE_BASE_GAS
+            + Constants::ACTIVE_BYTE_GAS
+            + self.metadata.as_ref().map_or(0, |m| m.len() as u64)
+    }
+}
```
danielSanchezQ: @davidrusu This are the ones that changed, metadata is present in active not in withdraw. Can you confirm its good? (We should update the specs as well).

nomos-mantle/src/gas.rs
```diff
@@ -0,0 +1,61 @@
...
+    const DECLARE_BASE_GAS: Gas;
+
+    /// Cost of declaration scales with number of locators
+    const DECLARE_LOCATOR_GAS: Gas;
+
+    /// Verify the proof of ownership.
+    const WITHDRAW_BASE_GAS: Gas;
+
+    /// Store the message on the ledger and process it.
+    const WITHDRAW_BYTE_GAS: Gas;
```
danielSanchezQ: We should remove this in case we do modify the spec. This is just a self-note.

nomos-mantle/src/tx.rs
```diff
@@ -0,0 +1,58 @@
...
+
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    gas::{Gas, GasConstants, GasPrice},
+    ops::Op,
+};
+
+// PLACEHOLDERS
+pub type LedgerTx = PhantomData<()>;
```
zeegomo, line 18: why not `()` directly?

nomos-mantle/src/ops/sdp.rs
```diff
@@ -0,0 +1,26 @@
...
+    fn gas_price<Constants: GasConstants>(&self) -> Gas {
+        Constants::WITHDRAW_BASE_GAS
+    }
+}
+
+impl GasPrice for SDPActiveOp {
+    fn gas_price<Constants: GasConstants>(&self) -> Gas {
+        Constants::ACTIVE_BASE_GAS
+            + Constants::ACTIVE_BYTE_GAS
+            + self.metadata.as_ref().map_or(0, |m| m.len() as u64)
```
zeegomo: should this be a `*`?

nomos-mantle/src/tx.rs
danielSanchezQ: Clippy complained about using unit (`()`) for hash...worked this way better.

nomos-mantle/src/ops/sdp.rs
danielSanchezQ: Indeed! good catch!

zeegomo (approved):

PR #1312, opened 2025-05-30
Cryptarchia bootstrapping networking

andrussal:

Cryptarchia Bootstrapping networking according to [Bootsrapping spec](https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6).

Note: This PR doesn't include logic to retrieve blocks from database.

@andrussal 

Yes

N/A

## Checklist

* [x] 1. Description added.

### DIFF
--- a/consensus/cryptarchia-sync/src/behaviour.rs
+++ b/consensus/cryptarchia-sync/src/behaviour.rs
@@ -38,13 +38,14 @@ pub const SYNC_PROTOCOL: StreamProtocol = StreamProtocol::new(SYNC_PROTOCOL_ID);
 
 const MAX_INCOMING_REQUESTS: usize = 4;
 
-type PendingRequestsFuture = BoxFuture<'static, Result<(PeerId, Libp2pStream), ChainSyncError>>;
+type PendingDownloadRequestsFuture =
+    BoxFuture<'static, Result<(PeerId, Libp2pStream), ChainSyncError>>;
 
-type LocallyInitiatedDownloadsFuture = BoxStream<'static, Result<BlocksResponse, ChainSyncError>>;
+type InProgressDownloadRequestsFuture = BoxStream<'static, Result<BlocksResponse, ChainSyncError>>;
 
-type ExternallyInitiatedDownloadsFuture = BoxFuture<'static, Result<(), ChainSyncError>>;
+type InProgressDownloadResponsesFuture = BoxFuture<'static, Result<(), ChainSyncError>>;
 
-type ExternalPendingDownloadRequestsFuture =
+type PendingDownloadResponsesFuture =
     BoxFuture<'static, Result<(PeerId, Libp2pStream, RequestMessage), ChainSyncError>>;
 
 type ToSwarmEvent = ToSwarm<
@@ -127,15 +128,15 @@ pub struct Behaviour {
     peers: HashSet<PeerId>,
     /// Futures for sending download requests. After the request is
     /// read, sending blocks is handled by `locally_initiated_downloads`.
-    pending_download_requests: FuturesUnordered<PendingRequestsFuture>,
+    pending_download_requests: FuturesUnordered<PendingDownloadRequestsFuture>,
     /// Futures for managing the progress of locally initiated block downloads.
-    in_progress_download_requests: SelectAll<LocallyInitiatedDownloadsFuture>,
+    in_progress_download_requests: SelectAll<InProgressDownloadRequestsFuture>,
     /// Futures for reading incoming download requests. After the request is
     /// read, sending blocks is handled by `externally_initiated_downloads`.
-    pending_download_responses: FuturesUnordered<ExternalPendingDownloadRequestsFuture>,
+    pending_download_responses: FuturesUnordered<PendingDownloadResponsesFuture>,
     /// Futures for managing the progress of externally initiated block
     /// downloads.
-    in_progress_download_responses: FuturesUnordered<ExternallyInitiatedDownloadsFuture>,
+    in_progress_download_responses: FuturesUnordered<InProgressDownloadResponsesFuture>,
     /// Futures for closing incoming streams that were rejected due to excess
     /// requests.
     incoming_streams_to_close: FuturesUnordered<BoxFuture<'static, ()>>,


nomos-services/network/src/backends/libp2p/swarm/chainsync.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum ChainSyncCommand {
+    RequestBlocks {
+        info: DownloadBlocksInfo,
+        blocks_stream: mpsc::Sender<BlocksResponse>,
+    },
+}
```
youngjoon-lee, line 52: I'm leaving a question here to gather discussions in this thread.

The PR description says:
> Spec leaves it open how to choose peers for downloading. Initial(this version) chooses a random peer

Could you point where a random peer is chosen in the implementation? I guess the PR description is outdated, because this PR just implements the networking part, not the service logic that starts IBD (initial block download). 

Actually, the spec says:
>  node downloads blocks to catch up with the tip of the local chain $c_{loc}$ of each peer

It means that IBD should be started for all peers. (Of course, we can group peers by tip). 

But, anyway, this can be implemented in a next PR, if I understand the scope of this PR correctly.l

consensus/cryptarchia-sync/src/blocks_downloader.rs
```diff
@@ -0,0 +1,260 @@
...
+    ) -> Poll<Result<(), ChainSyncError>> {
+        match fut.as_mut().poll(cx) {
+            Poll::Pending => {
+                self.current_step = ClosingStream(fut);
+                Poll::Pending
+            }
+
+            Poll::Ready(()) => Poll::Ready(Ok(())),
+        }
+    }
```
danielSanchezQ: Is there a reason why all this cannot be done in an `async` method instead?

consensus/cryptarchia-sync/src/blocks_downloader.rs
```diff
@@ -0,0 +1,260 @@
...
+            SendingDownloadRequest(fut) => self.handle_sending_download_request(cx, fut),
+
+            ReceivingMessage(fut) => self.handle_receiving_message(cx, fut),
+
+            ClosingStream(fut) => self.handle_closing_stream(cx, fut),
+
+            Done => Poll::Ready(Ok(())),
+        }
+    }
+}
```
danielSanchezQ: I think all this could be a simple async method. Then you can poll it outside if needed (?)

danielSanchezQ (commented):
There are many pieces I do not get here. I think it could have been simpler as it should have been just piping streams. But I'm probably missing many points.

nomos-services/network/src/backends/libp2p/swarm/chainsync.rs
andrussal: > because this PR just implements the networking part, not the service logic that starts IBD (initial block download).

This PR only handles low level networking part. It doesn't include logic to start blocks downloading by service.  It assumes that service layer initiates download for each batch of blocks. Sequence diagrams above include Cryptarchia only for clarity. (Please correct me if I am mistaken r.e. the service layer vs networking layer responsibilities). 

> Could you point where a random peer is chosen in the implementation?

https://github.com/logos-co/nomos/blob/e2a707a2943ac904aaae38e7f62bb797346c586f/consensus/cryptarchia-sync/src/behaviour.rs#L155

> But, anyway, this can be implemented in a next PR, if I understand the scope of this PR correctly.

Yes it's maybe better to build it on top of this PR to keep this one lighter. 

~~If I understand correctly, we need to add the following:~~

- ~~Request tip from peer before each `DownloadBlocksRequest` (but only if `target_block` is not given by service). Unless peer can decide it on its own if `target_block` is not given (but I wish to confirm this because it would be different from spec).~~
- ~~When service requests download from networking layer, networking layer sends `DownloadBlocksRequest` to all peers it knows about.~~


~~To clarify the last item r.e Rust implementation - is it fine from spec perspective if these blocks are returned over a single channel in the order they are received from peers? Or should they be sent as single batch of blocks? Python code does it as single batch, I wish to be sure if it's critical.~~

consensus/cryptarchia-sync/src/blocks_downloader.rs
andrussal: Let me see if I can make it async method. I figured there are many ways of doing this and decided to keep it all poll this time.

danielSanchezQ: > Let me see if I can make it async method. I figured there are many ways of doing this and decided to keep it all poll this time.

Not saying this do not work, but I think it would be easier if using `async` syntax instead.

andrussal: But you are right, from clarity perspective async should look much better. I'll give it quick try.

nomos-services/network/src/backends/libp2p/swarm/chainsync.rs
youngjoon-lee: IMO, the low-level network part shouldn't care about a set of peers. For clarity, in the Python code, the [download_blocks](https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-MIGRATED-1e58f96fb65c804aae85f84a2e94392d?source=copy_link#1f18f96fb65c80178dfae9349cc1edc5) function would be implemented in the service layer or in the cryptarchia-sync-core layer. This function has an argument `peer`. That is, this function downloads blocks from only a single peer.
How this function is called is up to the service layer or to the cryptarchia-sync-core layer. 
- This will be called for each of the peers (repeatedly), if a node is starting [IBD](https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-MIGRATED-1e58f96fb65c804aae85f84a2e94392d?source=copy_link#1f18f96fb65c8084b979e9b604dfe05b).
- This will be called for a specific peer, if a node wants to download ancestors of an [orphan block](https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-MIGRATED-1e58f96fb65c804aae85f84a2e94392d?source=copy_link#1f18f96fb65c80c0ad59ccb555b743e6).


I think, the low-level network part should implement only: How the [request and response](https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-MIGRATED-1e58f96fb65c804aae85f84a2e94392d?source=copy_link#1f18f96fb65c80a7a441e0bf51f0b3f9) are communicated between nodes (+ asking the tip of the honest chain. It's represented as [`peer.tip()`](https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-MIGRATED-1e58f96fb65c804aae85f84a2e94392d?source=copy_link#1f18f96fb65c80178dfae9349cc1edc5) in Python since it's trivial).

I haven't looked at the code in detail yet, but leaving this comment to clarify the scope. I will look at details tomorrow to see if I'm understanding the PR correctly.

youngjoon-lee: I just saw you updated the PR description. Now it explains the scope well but I will have a look soon. Thanks.

andrussal: Thank you for the input. This seems also better to me and was my original understanding as well. 

I just allow service to specify `peer_id` and don't follow up with proposed next PR(letting networking layer to request blocks from all peers itself)

andrussal: @youngjoon-lee Followup question: how does service know which peers are connected? Or where this logic should be placed? Asking just for my better understanding.

consensus/cryptarchia-sync/src/blocks_downloader.rs
```diff
@@ -0,0 +1,85 @@
...
+                    self.reply_tx
+                        .send(BlocksResponse::NetworkError(e.to_string()))
+                        .await?;
+                    return Err(Into::into(e));
+                }
+            }
+        }
+
+        stream.close().await.map_err(Into::into)
+    }
```
danielSanchezQ: This piece here should be the stream that we would return to the consumer.

consensus/cryptarchia-sync/src/blocks_provider.rs
```diff
@@ -0,0 +1,39 @@
...
+            }
+        }
+
+        if let Err(e) = pack_to_writer(&DownloadBlocksResponse::NoMoreBlocks, &mut stream).await {
+            return Err(ChainSyncError::PackingError(e));
+        }
+
+        Ok(())
+    }
+}
```
danielSanchezQ: There are utilities to pipe one stream to a sink. But this is still much simpler than previous version already. Thanks.

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,489 @@
...
+            ChainSyncError::StartSyncError("No peers available for chain sync".into())
+        })?;
+
+        let request =
+            DownloadBlocksRequest::new(target_block, local_tip, immutable_block, additional_blocks);
+        let mut task = DownloadBlocksTask::new(request, rx_blocks);
+
+        let control = self.control.clone();
+        self.locally_initiated_downloads
+            .push(async move { task.download_blocks(peer_id, control).await }.boxed());
```
danielSanchezQ: Yeah, this would probably just go away. We do not need to poll from futures here, just return the computation to the one that is gonna actively use it.

danielSanchezQ (commented):
Looking much better indeed

nomos-services/network/src/backends/libp2p/swarm/chainsync.rs
youngjoon-lee: > Followup question: how does service know which peers are connected? Or where this logic should be placed? Asking just for my better understanding.

Actually, that’s what we need to discuss, since it’s an implementation detail. I think we can start from the simplest approach:

- For IBD, a node talks to all nodes specified in the `initial_peers` option.
- While receiving new blocks (after IBD is done), a node should be able to know who sent a block, so that the node can initiate DownloadBlock if the block is orphan. I think the network service (used for the regular block broadcasting) should return not only a block but also the peer where the block is received from. 

I think this is enough to implement the spec. I need to look at the latest master code to check if my intention is possible.

consensus/cryptarchia-sync/src/messages.rs
```diff
@@ -0,0 +1,51 @@
+use bytes::Bytes;
+use nomos_core::header::HeaderId;
+use serde::{Deserialize, Serialize};
+
```
ntn-x2, line 4: These should be pulled from `nomos_core` or then be generic.

consensus/cryptarchia-sync/src/messages.rs
```diff
@@ -0,0 +1,51 @@
...
+pub type Block = Bytes;
+
+/// A request to initiate block downloading from a peer.
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct DownloadBlocksRequest {
+    /// Return blocks up to `target_block` if specified.
+    pub target_block: Option<HeaderId>,
```
ntn-x2: Isn't this a `BlockId` and not an option in the spec?

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,489 @@
...
+
+use crate::{
+    blocks_downloader::DownloadBlocksTask,
+    blocks_provider::{ProvideBlocksTask, BUFFER_SIZE},
+    messages::DownloadBlocksRequest,
+    Block,
+};
+
+/// Cryptarchia networking protocol for synchronizing blocks.
+const SYNC_PROTOCOL_ID: &str = "/nomos/cryptarchia/sync/1.0.0";
```
ntn-x2: Should the protocol ID be specified in the spec?

nomos-services/network/src/backends/libp2p/swarm/chainsync.rs
andrussal: Sounds good to me. Overall I think this design helps to keep core networking logic simple compared to making it download blindly from all peers for every DownloadRequest.

> For IBD, a node talks to all nodes specified in the initial_peers option.

We could add `initial_peers` to cryptarchia config. It doesn't seem straightforward to read network service config from cryptrarchia service.

> While receiving new blocks (after IBD is done), a node should be able to know who sent a block

Gossipsub event contains `propagation_source`, so we can use it here.

consensus/cryptarchia-sync/src/blocks_downloader.rs
andrussal: Do you think it's worth trying to do it? 

We have this events sequence: 
call `behaviour.start_blocks_download`->`open stream`->`send download request`->`start receiving blocks`.

`behaviour.start_blocks_download` stores `Send`.

If trying to use a stream, we could send `ToSwarm` event(with stream) when `start receiving blocks` and match it with correct `behaviour.start_blocks_download` request. I think this would require bunch of redesign on logic where we capture `ToSwarm` events. (Also dunno how easy is to send that stream around using a channel though).

(We also talked about sending stream back with `oneshot`. But it feels like `Send` is easier to work with if we use this pattern).

(But I understand your intention about correct responsibilities, i.e. which part of the system should handle creating a "channel" to return data).

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,489 @@
...
+
+pub struct Behaviour {
+    /// The stream behavior to handle actual networking.
+    stream_behaviour: StreamBehaviour,
+    /// Control to open streams to peers.
+    control: Control,
+    /// A handle to listen to incoming stream requests.
+    incoming_streams: IncomingStreams,
+    /// List of connected peers.
+    peers: Vec<PeerId>,
```
ntn-x2: I guess peers are unique. Why not using a `HashSet`? Removing peers would also be more efficient.

consensus/cryptarchia-sync/src/blocks_downloader.rs
```diff
@@ -0,0 +1,85 @@
...
+                .send(BlocksResponse::NetworkError(e.to_string()))
+                .await?;
+            return Err(Into::into(e));
+        }
+
+        loop {
+            match unpack_from_reader::<DownloadBlocksResponse, _>(&mut stream).await {
+                Ok(response) => match response {
+                    DownloadBlocksResponse::Block(block) => {
+                        if self.received_blocks >= DOWNLOAD_BLOCKS_LIMIT {
```
ntn-x2: I think it makes more sense to run this check at the end, before `await`ing for the next message, so that we can close the stream as soon as we reach the limit instead of waiting for one more message.

ntn-x2: Also, if this is an implementation detail (according to the spec), and not part of the protocol, why would we return a `ProtocolViolation` error? Also, would it make sense for this to be a config rather than a constant?

consensus/cryptarchia-sync/src/blocks_downloader.rs
```diff
@@ -0,0 +1,85 @@
...
+                                "Peer exceeded DOWNLOAD_BLOCKS_LIMIT of {DOWNLOAD_BLOCKS_LIMIT} blocks"
+                            );
+                            self.reply_tx
+                                .send(BlocksResponse::NetworkError(err_msg.clone()))
+                                .await?;
+
+                            return Err(ChainSyncError::ProtocolViolation(err_msg));
+                        }
+                        self.reply_tx.send(BlocksResponse::Block(block)).await?;
+                        self.received_blocks += 1;
```
ntn-x2: Just saying that this could THEORETICALLY overflow, if we use a big enough value as `DOWNLOAD_BLOCKS_LIMIT`. I would use its `checked` variant instead.

consensus/cryptarchia-sync/src/blocks_downloader.rs
danielSanchezQ: I think we use the same idea somewhere in da (maybe sampling) but essentially:

1. Service N requires the blocks so it requests it to network service
2. Network service is just a card dealer and forward the request to the proper behavior (sync one in this case)
3. Behaviour opens a new stream, sends the request, rewraps the stream into something the requester service expects and send it back through the channel that was in the request message.

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,489 @@
...
+            cx.waker().wake_by_ref();
+        }
+    }
+}
+
+impl NetworkBehaviour for Behaviour {
+    type ConnectionHandler = <StreamBehaviour as NetworkBehaviour>::ConnectionHandler;
+    type ToSwarm = Event;
+
+    fn handle_established_inbound_connection(
```
ntn-x2, line 343: What about `handle_pending_inbound_connection` and `handle_pending_outbound_connection` methods?

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,489 @@
...
+    /// Futures for managing the progress of externally initiated block
+    /// downloads.
+    externally_initiated_downloads:
+        FuturesUnordered<BoxFuture<'static, Result<(), ChainSyncError>>>,
+    /// Futures for reading incoming download requests. After the request is
+    /// read, sending blocks is handled by `externally_initiated_downloads`.
+    external_pending_download_requests: FuturesUnordered<
+        BoxFuture<'static, Result<(Stream, DownloadBlocksRequest), ChainSyncError>>,
+    >,
+    /// Futures for closing incoming streams that were rejected due to excess
```
ntn-x2, line 140: Something is missing at the end of this sentence.
```suggestion
    /// Futures for closing incoming streams that were rejected due to excess
```

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,489 @@
...
+                    Event::provide_blocks_request(
+                        request.target_block,
+                        request.known_blocks.local_tip,
+                        request.known_blocks.latest_immutable_block,
+                        request.known_blocks.additional_blocks,
+                        reply_sender,
+                    ),
+                )));
+            }
+            Err(e) => {
```
ntn-x2: Can't we unwrap this result one level up? It feels unnecessary to unwrap it here just to log it.

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,489 @@
...
+pub struct Behaviour {
+    /// The stream behavior to handle actual networking.
+    stream_behaviour: StreamBehaviour,
+    /// Control to open streams to peers.
+    control: Control,
+    /// A handle to listen to incoming stream requests.
+    incoming_streams: IncomingStreams,
+    /// List of connected peers.
+    peers: Vec<PeerId>,
+    /// Futures for managing the progress of locally initiated block downloads.
```
ntn-x2, line 132: Can we call them something else than locally initiated and externally initiated? This can be misleading. What about block requests for locally initiated downloads and block responses for externally initiated downloads?

consensus/cryptarchia-sync/src/blocks_downloader.rs
```diff
@@ -0,0 +1,85 @@
...
+pub struct DownloadBlocksTask {
+    request: DownloadBlocksRequest,
+    reply_tx: mpsc::Sender<BlocksResponse>,
+    received_blocks: usize,
+}
+
+impl DownloadBlocksTask {
+    pub const fn new(
+        request: DownloadBlocksRequest,
+        reply_tx: mpsc::Sender<BlocksResponse>,
```
ntn-x2: Would this be better as a stream? Not sure how we would handle errors tho.

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,489 @@
...
+    locally_initiated_downloads: FuturesUnordered<BoxFuture<'static, Result<(), ChainSyncError>>>,
```
ntn-x2: I think a type alias would be a very good option here.

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,489 @@
...
+        let control = self.control.clone();
+        self.locally_initiated_downloads
+            .push(async move { task.download_blocks(peer_id, control).await }.boxed());
+
+        Ok(())
+    }
+
+    fn handle_download_request(
+        &self,
+        result: Result<(Stream, DownloadBlocksRequest), ChainSyncError>,
```
ntn-x2: Would it make sense to wrap `(Stream, DownloadBlocksRequest)` into our own type, to make sure we don't call it with an invalid stream, for instance?

ntn-x2: We would be able to control that only a stream that has been properly polled as part of `external_pending_download_requests` is used.

nomos-core/chain-defs/src/wire/packing.rs
```diff
@@ -0,0 +1,79 @@
...
+use serde::{de::DeserializeOwned, Serialize};
+use thiserror::Error;
+
+use crate::wire;
+
+type Result<T> = std::result::Result<T, PackingError>;
+
+type LenType = u16;
+const MAX_MSG_LEN_BYTES: usize = size_of::<LenType>();
+const MAX_MSG_LEN: usize = LenType::MAX as usize;
```
ntn-x2, line 13: Why not using a `usize` directly?

consensus/cryptarchia-sync/src/behaviour.rs
danielSanchezQ: > Should the protocol ID be specified in the spec?

Yes it should be there indeed. if its not there then it should be added. Nice catch.

consensus/cryptarchia-sync/src/messages.rs
andrussal: You are right but I would like to get @youngjoon-lee opinion here if we could make it Option because spec also says: "if target_block is None, the peer returns blocks up to the tip of its latest honest chain". That would simplify networking protocol, we don't need to add additional logic to request tip.

```python
def initial_block_download(peers, local_tree):
	# In real implementation, these downloadings can be run in parallel.
	# Also, any optimization can be applied to minimize downloadings, such as grouping peers by tip.
	for peer in peers:
		if peer.tip().height > local_tree.tip().height:
			# if target_block is None, the peer returns blocks up to the tip of its latest honest chain
			download_blocks(local_tree, peer, target_block=None)
```

consensus/cryptarchia-sync/src/behaviour.rs
andrussal: Good point, will change.

consensus/cryptarchia-sync/src/blocks_downloader.rs
andrussal: > I think it makes more sense to run this check at the end, before awaiting for the next message, so that we can close the stream as soon as we reach the limit instead of waiting for one more message.

Currently (low level) protocol expects `NoMoreBlocks` message to handle clean end of stream. I used this approach to distinguish it from just dropping the stream by a peer. But maybe it's not really needed? (Also it's not part of specification but it might be fine, spec just describes high level behaviour).

andrussal: > Also, if this is an implementation detail (according to the spec), and not part of the protocol, why would we return a ProtocolViolation error? Also, would it make sense for this to be a config rather than a constant?

Spec has something in pyhton code(but 100% not sure if this should be considered as part of spec):

```python
class DownloadBlocksResponse:
	# A stream of blocks in parent-to-child order,
	# limited by DOWNLOAD_BLOCKS_LIMIT=1000.
	blocks: Stream[Block]
```

consensus/cryptarchia-sync/src/blocks_downloader.rs
andrussal: Good point, will use checked instead.

consensus/cryptarchia-sync/src/behaviour.rs
andrussal: Actually it might be better to call them. Underlying `libp2p_stream` doesn't handle these but it might safer to call for forward compatibility. 

Now I remember, we also should handle polling `libp2p_stream` for the same reason(although currently it only sends `ToSwarm::Dial`).

consensus/cryptarchia-sync/src/blocks_downloader.rs
ntn-x2: This is then confusing as to whether it is an implementation detail, or part of the spec. If it's the former, we should treat it as such. If it's the latter, I think it should be made more clear and explicit in the text. Let's wait for @youngjoon-lee then.

consensus/cryptarchia-sync/src/behaviour.rs
ntn-x2: Yeah I agree it's better to always forward all implementations to the wrapped type since it might diverge from the default one provided by the trait.

consensus/cryptarchia-sync/src/messages.rs
andrussal: Would it be ok to just assume Bytes? With generics, I think it will end up adding these to the Swarm itself. And even with `Block<Tx, Blob>` from nomos-core, we end up with generics in Swarm. NetworkAdapter could do the serialisation/deserialisation(as it's implemented currently by existing logic).

ntn-x2: Sure, in that case it would make more sense then to highlight the fact that we are dealing with a serialized block instead. And if that's the case, then I am fine with defining it here, as long as it's clear.

consensus/cryptarchia-sync/src/behaviour.rs
ntn-x2: Cc @youngjoon-lee

consensus/cryptarchia-sync/src/blocks_provider.rs
```diff
@@ -0,0 +1,41 @@
...
+            stream.flush().await?;
+        }
+
+        if let Err(e) = pack_to_writer(&DownloadBlocksResponse::NoMoreBlocks, &mut stream).await {
+            return Err(ChainSyncError::PackingError(e));
+        }
+
+        Ok(())
+    }
+}
```
danielSanchezQ: who is providing the `reply_rcv`? I think that ideally that should be the `sink` side of the stream so we can just dump te libp2pstream into that sink.

consensus/cryptarchia-sync/src/messages.rs
youngjoon-lee: In the spec, we don't use Optional for [`DownloadBlocksRequest.target_block`](https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81d2933be066852026ac). 

The [download_block](https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81e8b28fd2bdf3e446b1) function has an optional argument `target_block`, but it's just for emphasizing that the requesting node must ask for the peer's tip first to build a `DownloadBlocksRequest`.

So, here in the `struct DownloadBlocksRequest`, the `target_block` shouldn't be optional.

consensus/cryptarchia-sync/src/behaviour.rs
youngjoon-lee: Thank you. I defined the protocol ID in the spec: https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#20d261aa09df8080ab3ee974d774aef3

@andrussal Let's use `0.1.0` (instead of `1.0.0`) as our other protocols (DA, Blend).

consensus/cryptarchia-sync/src/blocks_downloader.rs
youngjoon-lee: `DOWNLOAD_BLOCKS_LIMIT=1000` is part of the spec. The responder can return at most 1000 blocks. However, if the responder returns more than 1000 blocks, the requestor can simply consume those extra blocks without any complaint because it's actually beneficial to the requestor. This is an implementation detail. I don't think this should be defined in the spec.
I just modified the comment in the [`class DownloadBlocksRequest`](https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81d2933be066852026ac) to avoid confusion, but still I think the behavior of handling extra blocks is implmementation detail.

youngjoon-lee: In this PR, if you think it's safer to handle only 1000 blocks, I think we can go with the ProtocolViolation error. I'm reading through the PR thoroughly and didn't finish it yet. So, I'm not sure which is the best way.

andrussal: > In this PR, if you think it's safer to handle only 1000 blocks, I think we can go with the ProtocolViolation error. I'm reading through the PR thoroughly and didn't finish it yet. So, I'm not sure which is the best way.

I started looking into how to retrieve blocks from storage: [PR](https://github.com/logos-co/nomos/pull/1339). Unless I am doing it wrong, it seems relatively expensive to find the blocks in chain. In a way, it might be actually even better not to have any limit. Once starting block is found, we can keep providing blocks until peer decides to close the stream.

andrussal: > > In this PR, if you think it's safer to handle only 1000 blocks, I think we can go with the ProtocolViolation error. I'm reading through the PR thoroughly and didn't finish it yet. So, I'm not sure which is the best way.
> 
> I started looking into how to retrieve blocks from storage: [PR](https://github.com/logos-co/nomos/pull/1339). Unless I am doing it wrong, it seems relatively expensive to find the blocks in chain. In a way, it might be actually even better not to have any limit. Once starting block is found, we can keep providing blocks until peer decides to close the stream.

But not sure if this actually works. It would mean storing the visited blocks in memory when first walking child->parent path. So it might be ok to do by 1000. May just need to cache some data(like each 1000th block), so when next request comes in, we don't have to walk the whole chain again.

youngjoon-lee: We can add any optimization in the responder side, but later. For now, I'd suggest to focus on the basic behavior. A responder returns up to 1000 blocks. 

In the requester side, I think it's fine to read a stream until it receives NoMoreBlock, regardless of how many blocks are returned. Maybe, some rich responder may want to return more than 1000 blocks, depending on their optimization strategy, and so on. Requesters can just simply consume all blocks returned.

consensus/cryptarchia-sync/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+libp2p-stream = { workspace = true }
+nomos-core    = { workspace = true }
+rand          = "0.9.1"
+serde         = { version = "1.0", features = ["derive"] }
+thiserror     = "1.0"
+tokio         = { version = "1.43", features = ["macros", "rt", "sync"] }
+tracing       = "0.1.41"
+
+[dev-dependencies]
+libp2p-swarm-test = { version = "0.5.0", features = ["tokio"] }
```
youngjoon-lee:
```suggestion
bytes         = { workspace = true }
futures       = "0.3"
libp2p        = { workspace = true, features = ["quic", "tokio", "yamux"] }
libp2p-stream = { workspace = true }
nomos-core    = { workspace = true }
rand          = "0.9"
serde         = { version = "1", features = ["derive"] }
thiserror     = "1"
tokio         = { version = "1", features = ["macros", "rt", "sync"] }
tracing       = "0.1"

[dev-dependencies]
libp2p-swarm-test = { version = "0.5", features = ["tokio"] }
```

I think it's better to use the same version scopes as other crates of us.

nomos-services/network/src/backends/libp2p/swarm/gossipsub.rs
```diff
@@ -101,12 +103,10 @@ impl SwarmHandler {
...
 
-    pub(super) fn handle_gossipsub_event(&self, event: nomos_libp2p::libp2p::gossipsub::Event) {
-        if let nomos_libp2p::libp2p::gossipsub::Event::Message { message, .. } = event {
+    pub(super) fn handle_gossipsub_event(&self, event: gossipsub::Event) {
+        if let gossipsub::Event::Message { message, .. } = event {
             let message = Event::Message(message);
-            if let Err(e) = self.events_tx.send(message) {
-                tracing::error!("Failed to send gossipsub message event: {}", e);
-            }
+            self.event_channels.send_chainsync_event(message);
```
youngjoon-lee:
```suggestion
            self.event_channels.send_gossipsub_message(message);
```

nomos-services/network/src/backends/libp2p/swarm/chainsync.rs
```diff
@@ -0,0 +1,78 @@
...
+pub struct DownloadBlocksInfo {
+    /// Return blocks up to `target_block` if specified.
+    pub target_block: Option<HeaderId>,
+    /// The latest block at the tip of the local chain.
+    pub local_tip: HeaderId,
+    /// The latest immutable block.
+    pub latest_immutable_block: HeaderId,
+    /// The list of additional blocks that the requester has.
+    pub additional_blocks: Vec<HeaderId>,
+}
```
youngjoon-lee: The struct comment is outdated, I think.

Also, what do you think about just using the `DownloadBlocksRequest` struct that you defined in the `consensus/cryptarchia-sync/src/messages.rs` instead of redefining this struct?

andrussal: I was thinking about it but wasn't sure if it's a good idea. One is data structure that is sent over network and this one used in the node internally.  Networking type could be something else, like protobuff for example. For this reason it seemed better to keep them distinct. But not 100% sure if this justifies though having 2 identical types?

consensus/cryptarchia-sync/src/messages.rs
ntn-x2: If not merely for a serialization issue, where other implementations might expect the field not to be optional indeed, leading to deserialization errors.

consensus/cryptarchia-sync/src/blocks_downloader.rs
ntn-x2: Can we explicitly mention in the spec that that's the block limit? Right now is only mentioned in a comment on the pseudocode, which also does not make it very clear whether it belongs to the spec or not.

consensus/cryptarchia-sync/src/messages.rs
danielSanchezQ: > You are right but I would like to get @youngjoon-lee opinion here if we could make it Option because spec also says: "if target_block is None, the peer returns blocks up to the tip of its latest honest chain". That would simplify networking protocol, we don't need to add additional logic to request tip.
> 
> ```python
> def initial_block_download(peers, local_tree):
> 	# In real implementation, these downloadings can be run in parallel.
> 	# Also, any optimization can be applied to minimize downloadings, such as grouping peers by tip.
> 	for peer in peers:
> 		if peer.tip().height > local_tree.tip().height:
> 			# if target_block is None, the peer returns blocks up to the tip of its latest honest chain
> 			download_blocks(local_tree, peer, target_block=None)
> ```

But this changes the specification then right? If specification doesn't stipulate is `Option` then we should bubble up a request for change and pint the necessary owner here.  I see that Youngjoon already replied, but to remark the procedure here.

danielSanchezQ: > If not merely for a serialization issue, where other implementations might expect the field not to be optional indeed, leading to deserialization errors.

Exactly. Lets do as per the spec.

youngjoon-lee: Yes. [`DownloadBlocksRequest.target_block`](https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81d2933be066852026ac) shouldn't be optional.

consensus/cryptarchia-sync/src/blocks_downloader.rs
youngjoon-lee: https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df8135a76cfd7735637795

Actually the spec shortly mentions that the responding peer can limit the number of blocks to be returned. But, the exact number is up to implementer. And, the spec assumes that requesters don't need to know the numbers determined by responders. Each responder can have its own limit.

https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df8100a572e68e49896aba

And here, the spec says that the requesting node should repeat requests for next batches.

In the pseudocode, I just removed the constant `DOWNLOAD_BLOCKS_LIMIT`. This makes implementers think that they have to define a constant commonly used by all nodes, but actually they don't need to do so. So, I think it's better to not specify this number in the spec. Responders can define the limit depending on their architecture and resources. Requesters don't need to know the value. I believe this design can be simple to implement.

Cargo.toml
```diff
@@ -3,6 +3,7 @@
 exclude = ["nomos-core/cl", "nomos-core/proof_statements", "nomos-core/risc0_proofs"]
 members = [
   "consensus/cryptarchia-engine",
+  "consensus/cryptarchia-sync",
```
youngjoon-lee, line 6: I'm leaving this comment here to gather replies under this thread.

@andrussal I feel it will take much time for the reviewers to finish their review. This PR is not super big, but contains changes in many layers. What do you think about moving the service layer changes from this PR to a new PR? Most of comments are currently focusing on the `cryptarchia-sync` crate for now. So, I think you can easily move the upper layer to a new PR, and it will make the review process much faster. 
Already, many comments have been left in a various range of topics. Personally, I (as a PR author) feel it's hard to address all comments and re-engage reviewers to have another look again.
I’m sharing my opinion but the decision is up to you. I will anyway have another look at details again.

nomos-services/network/src/backends/libp2p/swarm/chainsync.rs
youngjoon-lee: Later, if we start using protobuf or something else, we can have separate definitions. For now, as long as two data structures are 100% identical, I personally prefer having a single definition in a right place and using it in multiple places. But, I’ll leave the decision to you and focus instead on reviewing the core logic.

consensus/cryptarchia-sync/src/blocks_downloader.rs
andrussal: > Is there a reason why all this cannot be done in an async method instead?

This was addressed last week. Just adding reply for completeness.

consensus/cryptarchia-sync/src/blocks_downloader.rs
andrussal: > I think all this could be a simple async method.

Same here, was addressed last week.

consensus/cryptarchia-sync/src/blocks_downloader.rs
andrussal: This is now done without using a channel. Blocks are send to service using  ToSwarm.

consensus/cryptarchia-sync/src/blocks_provider.rs
andrussal: Here now I piped `reply_rcv` and libp2p `stream`

Cargo.toml
danielSanchezQ: Yes, too much going around imo too. I agree it would be better to split if possible. Otherwise we can continue working here although is gonna take a long time to get it to be merged. @andrussal next time lets make incremental changes.

consensus/cryptarchia-sync/src/blocks_provider.rs
```diff
@@ -0,0 +1,74 @@
...
+            .fold(
+                Ok(&mut stream),
+                |res: Result<&mut _, ChainSyncError>, block| async move {
+                    let stream = res?;
+                    let message = DownloadBlocksResponse::Block(block);
+                    send_message(peer_id, stream, &message).await?;
+                    Ok(stream)
+                },
+            )
+            .await?;
```
danielSanchezQ, line 67: Yeah, this is not what I meant. This keeps a few problems:

* We still keeps a receiver end of the stream which means that someone else have to push blocks, which means that there is some async task being spawn somewhere (?). When it could be totally avoided.
* Using a `Stream::fold` is not bad per se, but its just like the `while` loop before, but less readeable.

Ideally what we want is to :
1. Get a stream of blocks
2. Separate the libp2pstream and get the `Sink` part of it
3. Feed the block stream to the `Sink` with [`SinkExt::send_all`](https://docs.rs/futures/latest/futures/sink/trait.SinkExt.html#method.send_all)

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,632 @@
...
+    ) -> Self {
+        Self::ProvideBlocksRequest {
+            target_block,
+            local_tip,
+            latest_immutable_block,
+            additional_blocks,
+            reply_sender,
+        }
+    }
+}
```
youngjoon-lee, line 110: It seems that this function doesn't need to be `pub` since it is used only within this file `behaviour.rs`. Also, do we have any benefit to have this transparent wrapper function?

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,632 @@
...
+    /// read, sending blocks is handled by `locally_initiated_downloads`.
+    pending_download_requests: FuturesUnordered<PendingDownloadRequestsFuture>,
+    /// Futures for managing the progress of locally initiated block downloads.
+    in_progress_download_requests: SelectAll<InProgressDownloadRequestsFuture>,
+    /// Futures for reading incoming download requests. After the request is
+    /// read, sending blocks is handled by `externally_initiated_downloads`.
+    pending_download_responses: FuturesUnordered<PendingDownloadResponsesFuture>,
+    /// Futures for managing the progress of externally initiated block
+    /// downloads.
+    in_progress_download_responses: FuturesUnordered<InProgressDownloadResponsesFuture>,
```
youngjoon-lee, line 139:
```suggestion
    /// Futures for sending download requests. After the request is
    /// read, reading blocks is handled by `in_progress_download_requests`.
    pending_download_requests: FuturesUnordered<PendingDownloadRequestsFuture>,
    /// Futures for managing the progress of locally initiated block downloads.
    in_progress_download_requests: SelectAll<InProgressDownloadRequestsFuture>,
    /// Futures for reading incoming download requests. After the request is
    /// read, sending blocks is handled by `in_progress_download_responses`.
    pending_download_responses: FuturesUnordered<PendingDownloadResponsesFuture>,
    /// Futures for managing the progress of externally initiated block
    /// downloads.
    in_progress_download_responses: FuturesUnordered<InProgressDownloadResponsesFuture>,
```
I guess the comments should be fixed like this?

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,632 @@
...
+    pub fn start_blocks_download(
+        &mut self,
+        peer_id: PeerId,
+        download_blocks_info: DownloadBlocksInfo,
+    ) -> Result<(), ChainSyncError> {
+        let peer_id = self.choose_peer().ok_or_else(|| ChainSyncError {
+            peer: peer_id,
+            kind: ChainSyncErrorKind::StartSyncError("No peers available".to_owned()),
+        })?;
+
```
youngjoon-lee, line 193: I guess this code should be removed? If so, the `choose_peer` function is not necessary.

consensus/cryptarchia-sync/src/blocks_downloader.rs
```diff
@@ -0,0 +1,88 @@
...
+        mut control: Control,
+        request: DownloadBlocksInfo,
+    ) -> Result<(PeerId, Libp2pStream), ChainSyncError> {
+        let tip_response = Self::send_tip_request(peer_id, &mut control).await?;
+
+        let result: (PeerId, Libp2pStream) =
+            Self::send_download_request(peer_id, &mut control, request, tip_response.tip).await?;
+
+        Ok(result)
+    }
```
youngjoon-lee, line 31: `send_tip_request` shouldn't be called inside this function. 

The [Python code](https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81e8b28fd2bdf3e446b1) in the spec describes a logic that should be implemented in the service layer.
For IBD, the service layer will ask a peer for the tip. Using the tip as the target block, the service layer sends a DownloadBlocks request via the networking code. The service layer consumes blocks returned from the responder, and applies the blocks to the local block tree. The service layer repeats this process multiple times until the target block is received and applied to the local block tree.
For handling orphan blocks, the service layer uses the orphan block as the target block. So, requesting the tip is not necessary.

youngjoon-lee: Again, this networking crate shouldn't handle any bootstrap/sync logic. This crate should provide only the minimal network operations that can be used by the service layer.

consensus/cryptarchia-sync/src/blocks_downloader.rs
```diff
@@ -0,0 +1,88 @@
...
+    ) -> Result<GetTipResponse, ChainSyncError> {
+        let mut stream = open_stream(peer_id, control).await?;
+
+        let tip_request = RequestMessage::GetTipRequest(GetTipRequest);
+        send_message(peer_id, &mut stream, &tip_request).await?;
+
+        let response: GetTipResponse = unpack_from_reader(&mut stream)
+            .await
+            .map_err(|e| ChainSyncError::from((peer_id, e)))?;
+
```
youngjoon-lee, line 45: Could you please check whether we should close the `stream` here?

consensus/cryptarchia-sync/src/blocks_downloader.rs
ntn-x2: Yes, I also agree not having any mentions of any limits in the spec is the best thing to do. Thanks for fixing that.

nomos-core/chain-defs/src/wire/packing.rs
```diff
@@ -0,0 +1,79 @@
...
+where
+    Message: Serialize + Sync,
+    Writer: AsyncWriteExt + Send + Unpin,
+{
+    let packed_message = wire::serialize(message)?;
+    let data_length = get_packed_message_size(&packed_message)?;
+
+    let length_prefix: LenType = data_length
+        .try_into()
+        .map_err(|_| PackingError::LengthConversion)?;
```
youngjoon-lee, line 51: Instead of defining an additional error kind `LengthConversion`, I think we can throw a `MessageTooLarge` error here. Currently, the `get_packed_message_size` returns the `MessageTooLarge` error, but I think we can remove the `get_packed_message_size` function and just do the following here:
```
let length_prefix: LenType = packed_message.len().try_into.map_err(|_| PackingError::MessageTooLarge...);
```

nomos-core/chain-defs/src/wire/packing.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    writer
+        .write_all(&length_prefix.to_be_bytes())
+        .await
+        .map_err(Into::<PackingError>::into)?;
+
+    writer.write_all(&packed_message).await.map_err(Into::into)
```
youngjoon-lee, line 58:
```suggestion
    writer.write_vectored([
        &length_prefix.to_be_bytes(),
        &packed_message,
    ]).await.map_err(Into::into)?;
```
We may be able to use a vectored IO.

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,632 @@
...
+    ) -> Result<Self::ConnectionHandler, ConnectionDenied> {
+        self.stream_behaviour.handle_established_inbound_connection(
+            connection_id,
+            peer_id,
+            local_addr,
+            remote_addr,
+        )
+    }
+
+    fn handle_established_outbound_connection(
```
youngjoon-lee, line 358: We should also call `self.stream_behaviour.handle_pending_outbound_connection` as well. I know the `StreamBehaviour` doesn't override the `handle_pending_outbound_connection` function, but it may in the future. Similarly, you're already calling the `self.stream_behaviour.handle_pending_inbound_connection` even though it's not overrided by `StreamBehaviour`.
I think it's a safer practice to call all functions of the underlying `NetworkBehaviour` if we define a nested `NetworkBehaviour`. I remember we had discussed about this in a different PR.

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,632 @@
...
+
+        if let Poll::Ready(Some((peer_id, stream))) = self.incoming_streams.poll_next_unpin(cx) {
+            self.handle_incoming_stream(peer_id, stream);
+
+            cx.waker().wake_by_ref();
+        }
+
+        Poll::Pending
+    }
+}
```
youngjoon-lee, line 450:
```suggestion
    fn poll(
        &mut self,
        cx: &mut Context<'_>,
    ) -> Poll<ToSwarm<Self::ToSwarm, THandlerInEvent<Self>>> {
        while self.incoming_streams_to_close.poll_next_unpin(cx) == Poll::Ready(Some(())) {}

        if let Poll::Ready(Some(result)) = self.sending_requests.poll_next_unpin(cx) {
            return self.start_receiving_responses(cx, result);
        }

        if let Poll::Ready(Some(response)) = self.receiving_responses.poll_next_unpin(cx)
        {
            cx.waker().wake_by_ref();

            match response {
                Ok(result) => {
                    return Poll::Ready(ToSwarm::GenerateEvent(Event::DownloadBlocksResponse {
                        result,
                    }))
                }
                Err(e) => {
                    error!("Error while downloading blocks: {}", e);
                    return Poll::Ready(ToSwarm::GenerateEvent(Event::DownloadBlocksResponse {
                        result: BlocksResponse::NetworkError(e),
                    }));
                }
            }
        }

        if let Poll::Ready(Some(result)) = self.sending_responses.poll_next_unpin(cx) {
            if let Err(e) = result {
                error!("Sending response failed: {}", e);
            }

            cx.waker().wake_by_ref();

            return Poll::Pending;
        }

        if let Poll::Ready(Some(result)) = self.receiving_requests.poll_next_unpin(cx) {
            return self.handle_request(cx, result);
        }

        if let Poll::Ready(Some((peer_id, stream))) = self.incoming_streams.poll_next_unpin(cx) {
            self.handle_incoming_stream(peer_id, stream);

            cx.waker().wake_by_ref();
        }

        Poll::Pending
    }
}
```
Names of these futures are confusing. I know this has been discussed in a different thread. And I know that the `pending_*` is a naming convention used by rust-libp2p internally. But, I'm proposing these names, which are more intuitive.

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,632 @@
...
+        })?;
+
+        let control = self.control.clone();
+
+        self.pending_download_requests.push(
+            async move {
+                DownloadBlocksTask::send_request(peer_id, control, download_blocks_info).await
+            }
+            .boxed(),
+        );
```
youngjoon-lee, line 201: A requester should be able to receive responses corresponding to the request. In the current implementation, the `Behaviour` mixes all responses regardless of their corresponding requests. Then, the user of `Behaviour` will not be able to distinguish which responses are for which request. I saw that a response contains `PeerId`, but I think it's not sufficient.

This `start_blocks_download` function should accept (or return) a channel (or a stream) from which the requester can read responses. For that, the `Behaviour` may need to store requests (or request IDs) in memory along with the channel (the sender end). This is the practice used in the `request-response` [example](https://github.com/libp2p/rust-libp2p/blob/d3e88cfc2ec944c3e6beb7117a762452cb855e38/examples/file-sharing/src/network.rs#L386) in rust-libp2p.

consensus/cryptarchia-sync/src/behaviour.rs
youngjoon-lee: I just realized that what I commented here was already mentioned by Antonio in https://github.com/logos-co/nomos/pull/1312/files#r2131695239. So, I'm closing this thread.

consensus/cryptarchia-sync/src/behaviour.rs
danielSanchezQ: Wouldnt it do it already? The request is the doenload_blocks_info. It would just send the requests from that opening a new stream for that single request. Right?

youngjoon-lee: Because @andrussal can't leave comments in Github today, I talked with him in Discord DM. I heard that he initially had channels for the purpose that I'm proposing in this thread. But, I heard he switched to this approach which is similar as gossipsub. If this topic has been discussed like this, I think we can go with this approach. 
But I still think the `PeerId` is not sufficient for request <> response mapping, because a requester can create multiple requests to the same peer. We may need to introduce the concept of request ID but discussing how to do it.

youngjoon-lee: > Wouldnt it do it already? The request is the doenload_blocks_info. It would just send the requests from that opening a new stream for that single request. Right?

For IBD, yes, a requester will send a single request to each peer. But, when [handling orphan blocks](https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df819aac06c1a430b67878) in the regular process, a requester may send multiple requests (for different target blocks) to a single peer. It may or may not. It depends on how the orphan block handling is implemented in the upper layer. I think this network behaviour shouldn't assume 1:1 mapping between request and peer. The network behaviour shouldn't assume anything regarding how upper layers create requests.

danielSanchezQ: > > Wouldnt it do it already? The request is the doenload_blocks_info. It would just send the requests from that opening a new stream for that single request. Right?
> 
> For IBD, yes, a requester will send a single request to each peer. But, when [handling orphan blocks](https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df819aac06c1a430b67878) in the regular process, a requester may send multiple requests (for different target blocks) to a single peer. It may or may not. It depends on how the orphan block handling is implemented in the upper layer. I think this network behaviour shouldn't assume 1:1 mapping between request and peer. The network behaviour shouldn't assume anything regarding how upper layers create requests.

It changes nothing. You do not need a a channel. You take all of those requests, take each of the streams and merge them together and return them. If the upper layer is the one that handle the requests they may actually do it themselves. 
The behavior should just return you a stream per request, that is it, a gateway to the network, nothing else.

youngjoon-lee: Yes. 

I mean, the behaviour should just return a stream **per request**. But, it is not in the current implementation.

danielSanchezQ: > Yes.
> 
> I mean, the behaviour should just return a stream **per request**. But, it is not in the current implementation.

Yeah, I was confused. But indeed it should be like that. No need to keep track of it in the behaviour. It should just reply to a channel in the request with the stream. Thats all.

youngjoon-lee: In the current implementation, the behaviour returns `(PeerId, SerialisedBlock)`, so that the user of Behaviour can distingiush which response is for which request. But, I'm sure that PeerId is not sufficient. I discussed with Andrus to introduce RequestId for correct mapping.

youngjoon-lee: > It should just reply to a channel in the request with the stream. Thats all.

Yes, but the current implementation is not like this, afaik

danielSanchezQ: > > It should just reply to a channel in the request with the stream. Thats all.
> 
> Yes, but the current implementation is not like this, afaik

Yes you are right. I'm just saying we have to change the implementation to use streams and forget about this tracking at all.

youngjoon-lee: > Yes you are right. I'm just saying we have to change the implementation to use streams and forget about this tracking at all.

Ah, I got your point. Sorry for confusion. Yes, I agree.

consensus/cryptarchia-sync/src/blocks_downloader.rs
youngjoon-lee: Thank you for bringing this topic up. Very helpful for making the spec readable.

andrussal:
Closed in favour of [PR](https://github.com/logos-co/nomos/pull/1346)

PR #1319, opened 2025-06-03
refactor(blend): Use 'debug' log level for `MessageUnwrapError::NotAllowed`

### LINKED ISSUE #937: Change log level for `MsgUnwrapNotAllowed` to `DEBUG`.
```
2024-12-09T03:25:07.822251Z ERROR nomos_mix::message_blend: Failed to unwrap message: MsgUnwrapNotAllowed
```
`MsgUnwrapNotAllowed` occurs when the blend node doesn't have the private key to properly unwrap (decrypt) the mix message. This error can happen so often, and that's a normal situation. So, we don't need to log this error with the log level `ERROR`. It's better to use `DEBUG`.
For this, we may need some refactoring.

youngjoon-lee:

Closes #937

All Blend nodes attempt to unwrap incoming Blend messages using their own private key, but only one node can successfully do so. The other nodes are not allowed to unwrap the message. Previously, we logged these failed attempts at the `error` level. This PR changes the log level to `debug`, with some refactoring.

Yes

@youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [x] 1. Description added.

### DIFF
--- a/nomos-blend/message/src/mock/mod.rs
+++ b/nomos-blend/message/src/mock/mod.rs
@@ -62,7 +62,7 @@ impl BlendMessage for MockBlendMessage {
         private_key: &Self::PrivateKey,
     ) -> Result<(Vec<u8>, bool), MessageUnwrapError<Self::Error>> {
         if message.len() != MESSAGE_SIZE {
-            return Err(MessageUnwrapError::Other(Error::InvalidBlendMessage));
+            return Err(Error::InvalidBlendMessage.into());
         }
 
         // In this mock, we don't decrypt anything. So, we use private key as just a
@@ -74,10 +74,7 @@ impl BlendMessage for MockBlendMessage {
 
         // If this is the last layer
         if message[NODE_ID_SIZE..NODE_ID_SIZE * 2] == DUMMY_NODE_ID {
-            return Ok((
-                Self::payload(message).map_err(MessageUnwrapError::Other)?,
-                true,
-            ));
+            return Ok((Self::payload(message)?, true));
         }
 
         let mut new_message: Vec<u8> = Vec::with_capacity(MESSAGE_SIZE);
--- a/nomos-blend/message/src/mock/error.rs
+++ b/nomos-blend/message/src/mock/error.rs
@@ -1,3 +1,5 @@
+use crate::MessageUnwrapError;
+
 #[derive(thiserror::Error, Debug, PartialEq, Eq)]
 pub enum Error {
     #[error("Invalid blend message format")]
@@ -8,4 +10,10 @@ pub enum Error {
     InvalidNumberOfLayers,
     #[error("Invalid public key")]
     InvalidPublicKey,
+}
+
+impl From<Error> for MessageUnwrapError<Error> {
+    fn from(e: Error) -> Self {
+        Self::Other(e)
+    }
 }


nomos-blend/core/src/message_blend/mod.rs
```diff
@@ -98,8 +98,11 @@ where
                     tracing::error!("Failed to send message to the outbound channel: {e:?}");
                 }
             }
+            Err(e @ MessageUnwrapError::NotAllowed) => {
+                tracing::debug!("{e}");
```
ntn-x2, line 102: Should we start using log targets? This feels like `blend` or `blend::core` target to me?

nomos-blend/message/src/lib.rs
```diff
@@ -19,9 +19,17 @@ pub trait BlendMessage {
...
-    ) -> Result<(Vec<u8>, bool), Self::Error>;
+    ) -> Result<(Vec<u8>, bool), MessageUnwrapError<Self::Error>>;
     #[must_use]
     fn is_drop(message: &[u8]) -> bool {
         message == Self::DROP_MESSAGE
     }
 }
+
+#[derive(thiserror::Error, Debug)]
+pub enum MessageUnwrapError<E> {
```
ntn-x2: We should move away from single-letter generics, even if in this case is trivial.

nomos-blend/message/src/mock/mod.rs
```diff
@@ -60,21 +60,24 @@ impl BlendMessage for MockBlendMessage {
     fn unwrap(
         message: &[u8],
         private_key: &Self::PrivateKey,
-    ) -> Result<(Vec<u8>, bool), Self::Error> {
+    ) -> Result<(Vec<u8>, bool), MessageUnwrapError<Self::Error>> {
         if message.len() != MESSAGE_SIZE {
-            return Err(Error::InvalidBlendMessage);
+            return Err(MessageUnwrapError::Other(Error::InvalidBlendMessage));
```
ntn-x2: I guess we could also provide a `From` implementation for any type that implements `Error`, so we could do `return Err(Error::InvalidBlendMessage.into())` here? Does not really matter, but it might improve ergonomics a bit. And we could also use `?` which performs `From` conversion automatically.

ntn-x2 (approved):
Nice! It's already good, but I left a couple of considerations for things we might want to improve on.

nomos-blend/core/src/message_blend/mod.rs
youngjoon-lee: Good idea. What do you think about doing it in a separate PR? I can go through all loggings in the blend-related crates, and add targets. I just looked at the other crates, but it seems that no one is using log targets yet. We may need to think about defining a common way in the `nomos-tracing` crate, but I'm not sure yet.

danielSanchezQ: We are not using targets but spans. This logs would be under the blend span if it was set correctly in the service (I think it is).

nomos-blend/message/src/lib.rs
youngjoon-lee: Thank you! afeaa3de

nomos-blend/message/src/mock/mod.rs
youngjoon-lee: Good point. Thanks! 2ff11918

nomos-blend/core/src/message_blend/mod.rs
youngjoon-lee: Thanks, Daniel. I'm trying to find the code that sets the span, but couldn't find it in the blend service. Should I add something like [this](https://github.com/logos-co/nomos/blob/2ff1191883ebf00df7b5f4cf47283f88e37d5008/nomos-services/cryptarchia-consensus/src/lib.rs#L671) used in the cryptarchia service? It seems that no other service have this. The DA verifier service had it, but it has been removed by [this PR](https://github.com/logos-co/nomos/commit/f83e725b0df7d4fa1d54f9c4b2dca50880c8ec32#diff-b58d1d78d767cdba88f4e99d404d8fa2dc07d6e638bbc44d77f20f74739b5b44). So I guess we're setting spans somewhere else, but I can't find. I'm sorry I've missed those changes.

danielSanchezQ: > Thanks, Daniel. I'm trying to find the code that sets the span, but couldn't find it in the blend service. Should I add something like [this](https://github.com/logos-co/nomos/blob/2ff1191883ebf00df7b5f4cf47283f88e37d5008/nomos-services/cryptarchia-consensus/src/lib.rs#L671) in the cryptarchia service? It seems that no other service have this. The DA verifier service had it, but it has been removed by [this PR](https://github.com/logos-co/nomos/commit/f83e725b0df7d4fa1d54f9c4b2dca50880c8ec32#diff-b58d1d78d767cdba88f4e99d404d8fa2dc07d6e638bbc44d77f20f74739b5b44). So I guess we're setting spans somewhere else, but I can't find. I'm sorry I've missed those changes.

Yeah, all services should have that. Its an easy change and make for better debugging.

bacv: The `target` is automatically added by the tracing crate ([example](https://testnet.nomos.tech/explore?schemaVersion=1&panes=%7B%22i7v%22:%7B%22datasource%22:%22P8E80F9AEF21F6940%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Blevel%3D%5C%22info%5C%22%7D%20%7C%3D%20%60%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22P8E80F9AEF21F6940%22%7D,%22editorMode%22:%22builder%22,%22direction%22:%22backward%22%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D%7D&orgId=1)), not sure if spans per service would add much more info, but maybe it would make bit more sense to have them deeper in the service.

youngjoon-lee: Thanks for the example, Gusto! I think this is good enough.
```
{"message":"Failed to unwrap message: ConsistentLengthLayeredEncryptionError(IntegrityMacVerificationFailed)","_spans":[],"_target":"nomos_blend::message_blend","_module_path":"nomos_blend::message_blend","_file":"nomos-blend/core/src/message_blend/mod.rs","_line":102}
```

ntn-x2: Ok makes sense. Good to know. Does this mean using the `nomos_tracing` crate instead then? I've seen the `tracing` crate being used in some places as well, which would not do what Gusto described.

bacv: Actually, the target is added by unmodified tracing crate, `nomos-tracing` mostly contains open telemetry related layer creation helpers, which are used by `nomos-tracing-service`, and couple of macros to add `trace_id`. I think it would be nice to reexport `tracing` crate through `nomos-tracing`, that way there might be little less confusion.

ntn-x2: Ah ok, so you mean the default target added by tracing which is the path to the module name. It might now be ideal since we might want to grep logs for anything `blend`, for instance, and since log filtering works by matching prefixes, it would be ideal for all `blend`-related stuff to start with `blend`, optionally namespaced by more specific names like `::core`, or `::service`, for instance. Anyway, this is not within the scope of this PR, but if log targets are needed, e.g. to ask user to send their logs with certain targets enabled, then we should have a wider discussion about it.

PR #1321, opened 2025-06-03
feature(services): Notify ready

strinnityk:
Make services change their status to `Ready` when they are fully initialised.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 1. Description added.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,8 +98,8 @@ tests                    = { path = "./tests" }
 bytes            = "1.3"
 cached           = { version = "0.55.1", default-features = false }
 libp2p           = "0.55"
-overwatch        = { git = "https://github.com/logos-co/Overwatch", rev = "6c78ddd" }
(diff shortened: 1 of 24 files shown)


danielSanchezQ:
> What happens if a service forgets to call `notify_ready()`?

Other services that may be awaiting for it would get stuck.

danielSanchezQ:
> Yes, that's what I thought. There is no way to make this implicit I guess, right? `run` itself is async, so that's a no go. Automatically calling it at the end of `init`, perhaps by adding a default impl on the trait also does not seem like the right place.

We would have to refactor service trait. It can be done in the future also with the cleanup changes. For now we would have to live with it. Not something dramatic.

Btw, @strinnityk , could we add an info log below each ready signal?, something like:

```
info!("Service {SERVICE_ID} ready");
```

PR #1323, opened 2025-06-03
Da membership handler

pradovic:

This PR is a subset of this PR: https://github.com/logos-co/nomos/pull/1295. It is the first in a row of smaller PRs to address the same functionality and solution. 

The membership adapter is not yet added. Only the swapable da membership is implemented. The second PR should cover the membership service adapter and subscription.

https://github.com/logos-co/nomos/pull/1295

@pradovic @bacv 

Yes

No

## Checklist

> [!WARNING]  

* [x] 1. Description added.


nomos-da/network/subnetworks-assignations/src/lib.rs
```diff
@@ -11,6 +14,11 @@ pub trait MembershipHandler {
     /// Members Id type
     type Id;
 
+    /// Creates a new instance of the membership handler with the given members
+    /// and address book
+    #[must_use]
+    fn new_with(&self, members: Vec<PeerId>, addressbook: HashMap<PeerId, Multiaddr>) -> Self;
```
pradovic: @danielSanchezQ I am answering here as comments are broken in the main PR.

The reason for this factory method is that v1 and v2 assignation (membership handler) is created generically. The initial members and addressbook are read from the config and the set of config params is not the same for each of those. The config also includes configuration values for v1 and v2 such as dispersal factor etc.. This method just enables the caller to create a new one with the same configuration but difdferent addressbook and members and is used to update the membership when the new snapshot is received from membership service (update call will be in the next PR)

I did not want to create a separate trait for this call only, as:
1. It will introduce a new type and might bloat types a bit
2. The users can not really abuse this as they can not modify anything

nomos-services/data-availability/network/src/backends/mod.rs
```diff
@@ -18,6 +19,8 @@ pub trait NetworkBackend<RuntimeServiceId> {
 
     fn new(config: Self::Settings, overwatch_handle: OverwatchHandle<RuntimeServiceId>) -> Self;
     fn shutdown(&mut self);
+    fn update_membership(&mut self, members: Vec<PeerId>, addressbook: HashMap<PeerId, Multiaddr>);
```
danielSanchezQ: Ok. We do not need members. Members is just the addressbook keys ordered.

nomos-services/data-availability/network/src/backends/libp2p/membership.rs
```diff
@@ -0,0 +1,74 @@
...
+impl<Membership> MembershipHandler for DaMembershipHandler<Membership>
+where
+    Membership: MembershipHandler<Id = PeerId, NetworkId = u16>,
+{
+    type Id = PeerId;
+    type NetworkId = u16;
+
+    fn new_with(&self, _members: Vec<PeerId>, _addressbook: HashMap<PeerId, Multiaddr>) -> Self {
+        unreachable!("DaMembershipHandler does not support new_with")
+    }
```
danielSanchezQ: why it doesnt support it? should you just delegate on the internal update?

nomos-da/network/subnetworks-assignations/src/versions/v1.rs
```diff
@@ -69,6 +69,15 @@ impl MembershipHandler for FillFromNodeList {
...
     type Id = PeerId;
 
+    fn new_with(&self, members: Vec<PeerId>, addressbook: HashMap<PeerId, Multiaddr>) -> Self {
+        Self {
+            assignations: Self::fill(&members, self.subnetwork_size, self.dispersal_factor),
+            subnetwork_size: self.subnetwork_size,
+            dispersal_factor: self.dispersal_factor,
+            addressbook,
+        }
+    }
```
danielSanchezQ: Nit: I think a better naming would be `updated`. Now we just rebuild everything. But implementation wise is wrong. We should just fill up empty spaces from previous subnetwoks.

danielSanchezQ (changes requested):
This is much clearer and shorter. Thanks!

nomos-services/data-availability/network/src/backends/libp2p/membership.rs
pradovic: We can support it but it is not used, so I wanted to make it clear what is actually being called when this one is called in the chain. I can implement it as well, but the actuall new_with is called on an inner which is MembershipHandler

nomos-da/network/subnetworks-assignations/src/versions/v1.rs
pradovic: Actually, then the implementation is wrong and this is very good info. I assumed we rebuild the whole thing when the update is received and this was driving the design decision of this part of solution. If we need to node the state of the previous to update, then update should be inside the assignation and not behind a sharable arc swap. Let's sync on this pls

danielSanchezQ: > Oh ok, then the implementation is wrong and this is very good info. I assumed we rebuild the whole thing when the update is received and this was driving the design decision of this part of solution. If we need to node the state of the previous to update, then update should be inside the assignation and not behind a sharable arc swap. Let's sync on this pls

yes, it is dependant on the assignation algorithm. Which should have the same update actually.

pradovic: Ok, so we synced and actually rebuilding the assignation on update is not entirely correct idea for the future. In the future we will do what Daniel explained above. So, to accommodate this I will re-write this logic to do the update all the way down in the assignations and, for now, just use fill function to re-create it. But in the future we will need to implement those functions to properly handle update

pradovic: I have moved the sync logic and update logic all the way to assignations:
https://github.com/logos-co/nomos/pull/1323/commits/77667bb32b3ce0a43fce88ab4958c92aa174deed.

There is a trait for updating which is visible only to the networking service, and MembershipHandler trait is the same and is visible by behaviors. 

Now every assignation implementation have it's own update method, and sync primitives as well and `new_with` and shared handler is completely removed.

pradovic:
Closing as the design is changing. We will open new PRs with the new approach

PR #1324, opened 2025-06-04
fix(blend): use `Abortable` for swarm task

### LINKED ISSUE #1111: Use `Abortable` for async tasks in the blend service
We have a dead code that stores a `JoinHandle` not used anywhere.
https://github.com/logos-co/nomos-node/blob/4c0d5168e6757e8c6aa5dd2b3e9160f0edea323c/nomos-services/blend/src/backends/libp2p.rs#L28-L30

Originally, it was for a later use for adding [`Abortable`](https://docs.rs/futures/latest/futures/future/struct.Abortable.html). That has been [added to the DA code](https://github.com/logos-co/nomos-node/pull/853/files#diff-618f063bf12ec5d0d7faae32b79afd2d86d2a6db75d46b8abe0d4fb78489afd2L74). But I haven't added it to the blend code yet. I should have added a TODO comment at least.

First of all, I should check if it's really necessary to add `Abortable` to the blend code. If so, I'll add it, as done in DA.

youngjoon-lee:

Closes #1111 

This issue was discussed a few months ago. 
Previously, in the `BlendBackend`, we had a dead code that keeps a `JoinHandle` for the libp2p swarm task. It was not used anywhere. We removed it to make clippy happy: https://github.com/logos-co/nomos/pull/1106.

But, as we discussed in the https://github.com/logos-co/nomos/pull/1106#discussion_r1984846182, this PR re-adds the `JoinHandle` to the `BlendBackend`, but with wrapping it as a `Abortable`. When the `BlendService` (which manages the `BlendBackend`) is dropped, the task is aborted.
This implementation follows the same practice used in the [DA services](https://github.com/logos-co/nomos/blob/dba8a2289ce4e3755d20dce91e4fe6653bc70b48/nomos-services/data-availability/network/src/backends/libp2p/executor.rs#L87).

Yes

@youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [x] 1. Description added.

### DIFF
--- a/nomos-services/blend/src/lib.rs
+++ b/nomos-services/blend/src/lib.rs
@@ -211,6 +211,7 @@ where
     Network: NetworkAdapter<RuntimeServiceId>,
 {
     fn drop(&mut self) {
+        tracing::info!("Shutting down Blend backend");
         self.backend.shutdown();
     }
 }


nomos-services/blend/src/lib.rs
```diff
@@ -208,6 +205,16 @@ where
...
+impl<Backend, Network, RuntimeServiceId> Drop for BlendService<Backend, Network, RuntimeServiceId>
+where
+    Backend: BlendBackend<RuntimeServiceId> + 'static,
+    Network: NetworkAdapter<RuntimeServiceId>,
+{
+    fn drop(&mut self) {
+        self.backend.shutdown();
+    }
+}
+
```
danielSanchezQ: Iirc once the abort handle is droped (which will occur after the older run scope) it just aborts the launched tasks. But I think this is even better as it is explicit and easier to read.

danielSanchezQ (approved):

nomos-services/blend/src/backends/libp2p.rs
```diff
@@ -57,24 +64,37 @@ impl<RuntimeServiceId> BlendBackend<RuntimeServiceId> for Libp2pBlendBackend {
...
             incoming_message_sender.clone(),
         );
 
-        overwatch_handle.runtime().spawn(async move {
-            swarm.run().await;
-        });
+        let (task_abort_handle, abort_registration) = AbortHandle::new_pair();
+        let swarm_task = (
+            task_abort_handle,
+            overwatch_handle
```
ntn-x2, line 78: Do you need to store the spawned task? Isn't it sufficient to store the `task_abort_handle` and simply spawn the `Abortable`?

nomos-services/blend/src/lib.rs
```diff
@@ -208,6 +205,16 @@ where
     }
 }
 
+impl<Backend, Network, RuntimeServiceId> Drop for BlendService<Backend, Network, RuntimeServiceId>
+where
+    Backend: BlendBackend<RuntimeServiceId> + 'static,
+    Network: NetworkAdapter<RuntimeServiceId>,
+{
+    fn drop(&mut self) {
+        self.backend.shutdown();
```
ntn-x2, line 215: Should we add a log line somewhere to detect when the service is stopped?

nomos-services/blend/src/lib.rs
youngjoon-lee: Ah yeah. If so, we may not need this `Drop` impl (though it's easier to read, as you said). I actually followed the same practice used in DA. It implements the `Drop` trait here: https://github.com/logos-co/nomos/blob/707dbbd9bf2cf1743cfa7219aec4e8941fb4e772/nomos-services/data-availability/network/src/lib.rs#L115. Do you think it's better to keep this impl?

youngjoon-lee: Then, I also think we don't need to define the `shutdown` function.

danielSanchezQ: I think its better to keep the current explicit impl with the manual `Drop`. But yeah, I think you may also just remove the `shutdown` function.

youngjoon-lee: I will keep this practice, to have the blend code synced with the DA code. After that, I will deeply look at the details of Abortable and will propose refactorings for the entire codebase if necessary.

nomos-services/blend/src/backends/libp2p.rs
youngjoon-lee: You're right. Good point. 
I think it's better to refactor this in a next PR, because I want to refactor the DA code as well. This PR follows the same practice used in the DA code, just for consistency in the codebase (also as mentioned in https://github.com/logos-co/nomos/pull/1324#discussion_r2125886413).
Soon, I will open a PR that refactors all related code (blend, DA, and any other).

nomos-services/blend/src/lib.rs
youngjoon-lee: Yeah, I just added a log. 16183c70

As mentioned in https://github.com/logos-co/nomos/pull/1324#discussion_r2125886413, we may not need to implement the `Drop` trait. I will take a look at details soon and open a PR for refactoring. Then, this log may be moved to somewhere proper.

nomos-services/blend/src/backends/libp2p.rs
youngjoon-lee: I opened an issue https://github.com/logos-co/nomos/issues/1327 to not forget this.

PR #1326, opened 2025-06-04
fix(blend): change `TimedCache` to `SizedCache` for deduplicating messages in relaying

### LINKED ISSUE #1175: Change `TimedCache` to `TimedSizedCache` in Blend NetworkBehaviour
We need both time and size constraints for a cache.

youngjoon-lee:

Closes #1175 

This PR is a preparation for updating the [message relaying](https://www.notion.so/Blend-Protocol-Version-1-PENDING-MIGRATION-1c48f96fb65c809494efe63019a5ebfb?source=copy_link#1c88f96fb65c803b9e05f8976ed9b0c3) according to the latest spec.
- Each node should be able to deduplicate message during a single [session](https://www.notion.so/Blend-Protocol-Version-1-PENDING-MIGRATION-1c48f96fb65c809494efe63019a5ebfb?source=copy_link#1d08f96fb65c80948d39d98e6924a5db).
- The cache can be [cleared](https://www.notion.so/Blend-Protocol-Version-1-PENDING-MIGRATION-1c48f96fb65c809494efe63019a5ebfb?source=copy_link#2088f96fb65c80be9057c6b4ce6b7023) when a new session is started (precisely, after the [session transition period](https://www.notion.so/Blend-Protocol-Version-1-PENDING-MIGRATION-1c48f96fb65c809494efe63019a5ebfb?source=copy_link#1cb8f96fb65c8037bdb6d36f4cbcd3f4) has passed).

Therefore, we don't need TTL for the cache (provided by `TimedCache`). Theoretically, we can simply use a `HashMap` (or `HashSet`) for this purpose, but I chose to use `SizedCache` to explicitly bound the max size that is expected during a single session.

Since the concept of session is not implemented yet, this PR doesn't implement:
- setting the cache size based on the session length
- clearing the cache when the session transition period has passed.

Yes

@youngjoon-lee @madxor 

Yes

No

## Checklist

> [!WARNING]  

* [x] 1. Description added.

### DIFF
--- a/nomos-services/blend/src/backends/libp2p.rs
+++ b/nomos-services/blend/src/backends/libp2p.rs
@@ -116,7 +116,9 @@ impl BlendBehaviour {
             blend:
                 nomos_blend_network::Behaviour::<SphinxMessage, TokioIntervalStreamProvider>::new(
                     nomos_blend_network::Config {
-                        // This size must be ROUNDS_IN_SESSION * BLEND_HOPS
+                        // TODO: This should be as (ROUNDS_IN_SESSION + BUFFER) * MAX_HOPS,
+                        // once session and round mechanisms are implemented.
+                        // https://www.notion.so/Blend-Protocol-Version-1-PENDING-MIGRATION-1c48f96fb65c809494efe63019a5ebfb?source=copy_link#2088f96fb65c80be9057c6b4ce6b7023
                         seen_message_cache_size: 1_944_000,
                         conn_monitor_settings: config.conn_monitor,
                     },


nomos-services/blend/src/backends/libp2p.rs
```diff
@@ -116,7 +116,8 @@ impl BlendBehaviour {
             blend:
                 nomos_blend_network::Behaviour::<SphinxMessage, TokioIntervalStreamProvider>::new(
                     nomos_blend_network::Config {
-                        duplicate_cache_lifespan: 60,
+                        // This size must be ROUNDS_IN_SESSION * BLEND_HOPS
+                        seen_message_cache_size: 1_944_000,
```
ntn-x2, line 122: I think it would be better for this to be a constant that links to the spec paragraph that defines it.

ntn-x2 (approved):

ntn-x2 (approved):

nomos-services/blend/src/backends/libp2p.rs
youngjoon-lee: Good point. But, I'd like to have these constants (or configurable values) in a right place.
Instead, I added a TODO comment a07163bd3f15fe794a734cf8be3d8905aadef5db, so that I can revisit this once we implement session and round.
FYI, I've discussed further about the cache size in [Discord](https://discord.com/channels/1111286067413405788/1379727129977159720/1379769016285139035). I didn’t ping you in the thread because Marcin already updated the spec accordingly but sharing this here for more information.

PR #1328, opened 2025-06-04
feature(da): DA membership handler

pradovic:

This PR is a first step into using Membership service from DA to construct and update membership. It implements membership handler and it's creation and update logic.

Yes

@pradovic @bacv 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 1. Description added.

### DIFF
--- a/nomos-services/data-availability/network/src/membership/mod.rs
+++ b/nomos-services/data-availability/network/src/membership/mod.rs
@@ -4,11 +4,12 @@ pub mod handler;
 use std::collections::{HashMap, HashSet};
 
 use libp2p::PeerId;
+use nomos_core::block::BlockNumber;
 use nomos_da_network_core::SubnetworkId;
 
 pub type Assignations = HashMap<SubnetworkId, HashSet<PeerId>>;
 
 pub trait MembershipStorage {
-    fn store(&self, block_number: u64, assignations: Assignations);
-    fn get(&self, block_number: u64) -> Option<Assignations>;
+    fn store(&self, block_number: BlockNumber, assignations: Assignations);
+    fn get(&self, block_number: BlockNumber) -> Option<Assignations>;
 }


nomos-da/network/subnetworks-assignations/src/lib.rs
```diff
@@ -31,9 +46,10 @@ pub trait MembershipHandler {
     fn last_subnetwork_id(&self) -> Self::NetworkId;
 
     fn get_address(&self, peer_id: &PeerId) -> Option<Multiaddr>;
-}
 
-use std::sync::Arc;
+    /// Returns all subnetworks with assigned members.
+    fn subnetworks(&self) -> HashMap<Self::NetworkId, HashSet<Self::Id>>;
```
danielSanchezQ: Shouldn't this be references?
```suggestion
    fn subnetworks(&self) -> &HashMap<Self::NetworkId, HashSet<Self::Id>>;
```

nomos-services/data-availability/network/src/membership/mod.rs
```diff
@@ -0,0 +1,12 @@
...
+pub mod handler;
+
+use std::collections::{HashMap, HashSet};
+
+use libp2p::PeerId;
+use nomos_da_network_core::SubnetworkId;
+
+pub trait MembershipStorage {
+    fn store(&mut self, block_number: u64, assignations: HashMap<SubnetworkId, HashSet<PeerId>>);
+    fn get(&self, block_number: u64) -> Option<HashMap<SubnetworkId, HashSet<PeerId>>>;
```
danielSanchezQ: At this point this should be its own type.

```rust
    HashMap<SubnetworkId, HashSet<PeerId>>
```

nomos-da/network/subnetworks-assignations/src/lib.rs
danielSanchezQ: Maybe not as Ids are usually copy 🤔 . Whatever is valid, not a bottleneck for now probably.

bacv: Your suggestion should be applied, the caller can decide if it needs to be cloned.

pradovic: The code did not compile for me with a &, so I removed the reference. I think in order to return a reference we would need to refactor membership handler slightly to store assignations  as `HashMap<Self::NetworkId, HashSet<Self::Id>>`, instead of `Vec<HashSet<PeerId>>`. Should be ezy. Does that makes sense?

bacv: > Shouldn't this be references?

At first it seemed as a good idea, ideally we would return here `impl Iter<Item = (Self::NetworkId, HashSet<Self::Id>>>`, but the purpose of this method is to "dump" all the data inside the membership.

When we think of this method like "dump_subnetworks" the clone makes sense. I encountered problems getting reference to guarded data inside DaMembership, not sure if it's worth to return reference to membership data, since the sole user of this trait implementors will be da network service.

The method could be renamed to "dump_assignations" or similar to justify the clone, it could also consume `self`, seems like a minor detail, but we can resolve it here, or leave as is (forever :D)

nomos-services/data-availability/network/src/membership/handler.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+impl<Membership> DaMembershipHandler<Membership> {
+    pub fn new(membership: Membership) -> Self {
+        // ArcSwap wraps the inner type into Arc when calling `from_pointee`, `load` and
+        // `store` methods instead of T needs to take manually wrapped Arc<T>.
+        Self {
+            membership: Arc::new(ArcSwap::from_pointee(membership)),
+        }
+    }
```
youngjoon-lee, line 17: Could you please briefly explain the context why `Arc` and `ArcSwap` are used? It seems that this has been discussed somewhere else but I couldn't find it.

bacv: `ArcSwap` is very similar to `RWLock`, just works without locking. But if we want to share it between threads, we still need to wrap it, just like other mutexes. In the [documentation example](https://docs.rs/arc-swap/1.7.1/arc_swap/#motivation) it's wrapped inside `Lazy`, in this case it could also be static, but I think Arc works better.

nomos-da/network/subnetworks-assignations/src/lib.rs
bacv: The method could be called `into_assignations(self) -> Assignations` and consume self.

nomos-services/data-availability/network/src/membership/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+
+use std::collections::{HashMap, HashSet};
+
+use libp2p::PeerId;
+use nomos_da_network_core::SubnetworkId;
+
+pub type Assignations = HashMap<SubnetworkId, HashSet<PeerId>>;
+
+pub trait MembershipStorage {
+    fn store(&self, block_number: u64, assignations: Assignations);
```
ntn-x2: Shouldn't `BlockNumber` be generic, or at least pulled in from some common definitions?

nomos-da/network/subnetworks-assignations/src/lib.rs
```diff
@@ -1,10 +1,25 @@
...
+use std::{
+    collections::{HashMap, HashSet},
+    hash::Hash,
+    sync::Arc,
+};
 
 use libp2p::Multiaddr;
 use libp2p_identity::PeerId;
 
+pub trait MembershipCreator: MembershipHandler {
```
ntn-x2, line 14: Why does the creator also have to be a handler itself? Could it not return a handler as an associated type instead?

ntn-x2: I think it would be also easier to mock and unit test

nomos-services/data-availability/network/src/membership/adapter/mock.rs
```diff
@@ -0,0 +1,185 @@
...
+            // was passed in the init method.
+            let mut update = HashMap::new();
+            update.insert(PeerId::random(), Multiaddr::empty());
+
+            adapter.update(1, update);
+        }
+    }
+
+    #[test]
+    fn test_adapter_usage() {
```
ntn-x2, line 178: What is being tested here, exactly? We have a mock implementation, and we test that the mock is implemented correctly? If the mock is supposed to be used as-is, perhaps a different name would be better? Or at least a brief description of what the mock does and what is mocks would be helpful, in my view.

bacv: I'll add the description, it's supposed to demonstrate usage inside the service - it's a minor detail but wanted to validate the api design.

Network backend with swarm is initialized in the service `init` method. Adapter will be initialized in the `run` method, because it needs the handle to the membership service. To allow backend to read the current membership and adapter to update the membership, a shared DaMembershipHandler was created.

It's initialized inside `init` method, and then passed as clone to the adapter in the `run` method.

nomos-da/network/subnetworks-assignations/src/lib.rs
bacv: That's a good suggestion, will update

nomos-services/data-availability/network/src/membership/mod.rs
bacv: Last time I check we didn't have such definition, I would prefer to use type from `nomos-core` here instead of generic one. I can create a new PR with a type in core, `nomos-sdp` would also need an update - it has `BlockNumber` defined for itself.

nomos-da/network/subnetworks-assignations/src/lib.rs
danielSanchezQ: No worries, just let it return the the hashmap (which should be a type itself imo).

nomos-services/data-availability/network/src/membership/handler.rs
danielSanchezQ: I'll check myself because it makes no sense that `ArcSwap` is not send itself, it just contains an inner `Arc`. What it probably doesn't implement is `Clone` because it would make no sense. Which I think would invalidate the whole thing (?)

danielSanchezQ: It is [`Send`](https://docs.rs/arc-swap/1.7.1/arc_swap/struct.ArcSwapAny.html#impl-Send-for-ArcSwapAny%3CT,+S%3E)

nomos-da/network/subnetworks-assignations/src/lib.rs
bacv: After trying to implement the suggestion I realized that Creator and Membership needs to be tightly coupled for the functionality that is currently implemented:
- Idea is to enforce membership implementation to have both traits implemented
- `DaMembershipHandler<Membership>` expects Membership to implement `MembershipCreator` and `MembershipHandler` traits.
- `DaMembershipHandler` only implements `MembershipHandler` and is passed to the "backend" as such trait implementor, so the backend can't make updates to the membership
- Adapter on the other hand has access to the full list of methods that DaMembershipHandler implements - one of them is receiving underlying membership as MembershipCreator to create an updated instance of itself.

If we were to implement the suggestion to have creator and handler decoupled, then `DaMembershipHandler` would need to be generic over two parameters, it's a valid approach, but I don't think it's necessary in this particular case.

nomos-services/data-availability/network/src/membership/handler.rs
bacv: > It is [Send](https://docs.rs/arc-swap/1.7.1/arc_swap/struct.ArcSwapAny.html#impl-Send-for-ArcSwapAny%3CT,+S%3E)

But we need two clones of it, it doesn't implement clone.

bacv: Oh you already mentioned that. We with Petar also searched for the ways to `Clone` it, but didn't find a way provided by the lib.

danielSanchezQ: > Oh you already mentioned that. We with Petar also searched for the ways to `Clone` it, but didn't find a way provided by the lib.

Yeah, that is why I think we are breaking the pattern here. What would happen now is that anyone with access to the any of the clones can actually swap it.

bacv: In case of allowing updates only by those who has full DaMembershipHandler (in this case adapter), we are not braking pattern - I don't see a problem of multiple callers updating the membership, even though in our case it's only one. The backend will never be able to update the membership because it receives DaMembershipHandler as `MembershipHandler` and this trait has no methods to update the inner data.

danielSanchezQ: > In case of allowing updates only by those who has full DaMembershipHandler (in this case adapter), we are not braking pattern - I don't see a problem of multiple callers updating the membership, even though in our case it's only one. The backend will never be able to update the membership because it receives DaMembershipHandler as `MembershipHandler` and this trait has no methods to update the inner data.

True, that makes more sense. Thanks.

youngjoon-lee: > even though in our case it's only one.

So, if I understand the discussion correctly, is this for making `DaMembershipHandler` thread-safe by itself, even though it'll be used by only the NetworkService?

youngjoon-lee: and, the reason why we use `ArcSwap` (instead of `Mutex` or so) is because of the advantage of lock-free data structure?

bacv: > thread-safe by itself, even though it'll be used by only the NetworkService

Network Backend, which is libp2p swarm, is spawned on a different thread and adapter is running in the service loop. DaMembershipHandler will be shared in at least two threads.

> and, the reason why we use ArcSwap (instead of Mutex or so) is because of the advantage of lock-free data structure?

In da swarm, we have multiple behaviours for different protocols such as Replication, Dispersal, Sampling. Each of these operate on membership independently. When DaMembershipHandler will be passed to the swarm, it will be cloned further for each behaviour. This functionality is actually implemented and no changes needed in the swarm. 

The only thing that we had to solve was how do we make updates to this shared structure and ArcSwap seemed like a good solution, similar to what's used in routing related codebases.

youngjoon-lee: It makes sense. Thank you for detailed background.

pradovic: > and, the reason why we use `ArcSwap` (instead of `Mutex` or so) is because of the advantage of lock-free data structure?

I would say correct. It should have a bit better performance in single writer/multiple readers scenario. Something like a lock-free `RWLock`.

danielSanchezQ (approved):
Ok by me. Lets see now how it deals with other layers.

PR #1332, opened 2025-06-05
feature(services): Await for service readiness (Pt. 1)

strinnityk:
- Implement `wait_until_services_are_ready` macro. It waits for `Duration` until all specified services notify they're ready before continuing.
- PoC: Use the `wait_until_services_are_ready` macro in `Membership` service.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 1. Description added.


danielSanchezQ (commented):
Why a macro instead of a function?

danielSanchezQ:
> > Why a macro instead of a function?
> 
> The number of services you have to wait for varies from `Service` to `Service`. I'd have to implement one function per `Service`.

why one per service?

nomos-services/utils/src/overwatch/status/utils.rs
```diff
@@ -0,0 +1,296 @@
...
+                non_ready_services.push(service_status_entry);
+            }
+        )+;
+
+        if !non_ready_services.is_empty() {
+            let error: $crate::overwatch::status::ServiceStatusEntriesError<RuntimeServiceId> = non_ready_services.into();
+            return Err(::overwatch::DynError::from(error));
+        }
+    };
+}
```
danielSanchezQ: Ok, so the macro works but I would change a few behaviors:

* I think you can create a future for each one and then join those futures, then you can collect them. You will be able to reuse some of the utilities you made I think.
* Not sure about the timeout, if at None it just waits forever then I guess there is no harm on having it.
* This should be in a specific scope (just rewrap everything), then return a `Result` from that scope so it can be used a method instead.

nomos-services/utils/src/overwatch/status/utils.rs
danielSanchezQ, line 1: Probably I would rename this to macros if we intend to have them here, a utils inside a utils crate is not that nice

nomos-services/membership/src/lib.rs
```diff
@@ -129,6 +131,12 @@ where
             <RuntimeServiceId as AsServiceId<Self>>::SERVICE_ID
         );
 
+        wait_until_services_are_ready!(
+            &self.service_resources_handle.overwatch_handle,
+            Some(Duration::from_millis(3000)),
+            <S as SdpAdapter>::SdpService
+        );
```
danielSanchezQ: Are you gonna expand later for all services in this PR or different one?

strinnityk: Different. I added that just as PoC (albeit maybe I shouldn't have). I mention that in the PR description.

nomos-services/utils/src/overwatch/status/service_status_entry/status_wrapper.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub const fn new(status: overwatch::services::status::ServiceStatus) -> Self {
+        Self(status)
+    }
+}
+
+impl Display for ServiceStatus {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        use overwatch::services::status::ServiceStatus::{Ready, Starting, Stopped};
+
+        let display_value = match self.0 {
```
ntn-x2: I think it would make more sense to implement `Display` for the wrapped `ServiceStatus` and call `write!(f, self.0)` instead.

nomos-services/utils/src/overwatch/status/service_status_entry/error.rs
```diff
@@ -0,0 +1,40 @@
+//! Provides conversion from a collection of [`ServiceStatusEntry`] to
+//! the general-purpose [`DynError`](overwatch::DynError) type.
+
+use std::fmt::{Debug, Display, Formatter};
+
+use crate::overwatch::status::ServiceStatusEntry;
+
+#[derive(Debug)]
+pub struct ServiceStatusEntriesError<RuntimeServiceId: Display> {
```
ntn-x2, line 9: I think we don't need the `Display` bound at this point but only in the impl block.

nomos-services/utils/src/overwatch/status/service_status_entry/error.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+impl<RuntimeServiceId: Display + Debug> std::error::Error
+    for ServiceStatusEntriesError<RuntimeServiceId>
+{
+}
+
+impl<RuntimeServiceId, Iter> From<Iter> for ServiceStatusEntriesError<RuntimeServiceId>
+where
+    RuntimeServiceId: Display,
```
ntn-x2, line 32: If we remove the bound from the struct, we can remove it from here, since I don't see how implementing `Display` is useful in this context (?)

nomos-services/utils/src/overwatch/status/utils.rs
```diff
@@ -0,0 +1,296 @@
...
+        }
+    }
+
+    struct ServiceA {
+        service_resources_handle: OpaqueServiceResourcesHandle<Self, RuntimeServiceId>,
+    }
+
+    impl ServiceData for ServiceA {
+        type Settings = ();
+        type State = ServiceStateA;
```
ntn-x2: Can't we use `NoState` here instead?

nomos-services/utils/src/overwatch/status/utils.rs
```diff
@@ -0,0 +1,296 @@
...
+        let settings = AppServiceSettings {
+            service_a: (),
+            service_b: (),
+            dependent_service: (),
+        };
+        OverwatchRunner::<App>::run(settings, None).expect("Failed to run overwatch")
+    }
+
+    #[test]
+    fn test_wait_until_services_are_ready_macro() {
```
ntn-x2: Should we test the timeout case as well?

nomos-services/utils/src/overwatch/status/utils.rs
```diff
@@ -0,0 +1,296 @@
...
+///        ServiceB,
+///        ServiceC
+///     );
+/// }
+/// ```
+///
+/// A complete example with working service implementations is available in this
+/// module's tests.
+#[macro_export]
+macro_rules! wait_until_services_are_ready {
```
ntn-x2: Just asking since we've hit a lot of corner cases the `derive_services` macro as well. Does this work with generics, and a mix of generic and non-generic types? Perhaps it makes sense to add a mock generic service in the tests as well to make sure it compiles properly.

danielSanchezQ: It should work as we would have to inject the full type manually for the calls.

ntn-x2: Sure. Just a test where we mix the two to make sure we don't break it in the future would be sufficient then :)

nomos-services/utils/src/overwatch/status/utils.rs
strinnityk: > - Not sure about the timeout, if at None it just waits forever then I guess there is no harm on having it.

As in default to `None`?

nomos-services/utils/src/overwatch/status/service_status_entry/status_wrapper.rs
strinnityk: I'm in favour of that.
That being said, because I didn't wanna bother updating `overwatch` while I was doing this I implemented this small module as a workaround.
If we're fine adding `Display` to `overwatch::ServiceStatus` then great.

nomos-services/utils/src/overwatch/status/service_status_entry/error.rs
strinnityk: `Display` is a supertrait of `std::error:Error`.

nomos-services/utils/src/overwatch/status/service_status_entry/error.rs
strinnityk: Same as above, I believe.

nomos-services/utils/src/overwatch/status/utils.rs
strinnityk: In the existing test case I already included both:
```rust
    &self.service_resources_handle.overwatch_handle,
    Some(Duration::from_secs(5)),
    GenericService, // Generic
    ServiceB // Concrete
)?;
```

nomos-services/utils/src/overwatch/status/service_status_entry/status_wrapper.rs
ntn-x2: Oh, right, sorry, did not notice the other one was in overwatch. Yeah I think in general it would make more sense to add it there, but it's not a big problem if we do it this way, for now.

nomos-services/utils/src/overwatch/status/service_status_entry/error.rs
ntn-x2: Yes but we can require `Display` only when needed, no? Why does it have to be in the struct definition?

nomos-services/utils/src/overwatch/status/utils.rs
ntn-x2: Mhh ok I see. I was more suggesting to have a `Service<Generic>` in there, mostly for syntax reasons. Do you think it's not necessary?

nomos-services/utils/src/overwatch/status/service_status_entry/error.rs
strinnityk: Oh, sorry, my answer was a tad incomplete.
When I attempted to remove the `Display` bound, I got an error due to deriving `Debug`. To derive `Debug`, `RuntimeServiceId` must implement `Display`.

nomos-services/utils/src/overwatch/status/utils.rs
strinnityk: Oh, I see. Sure, let's add one of those as well.

nomos-services/utils/src/overwatch/status/macros.rs
```diff
@@ -0,0 +1,310 @@
...
+                let wait_for_future = async {
+                    if let Err(service_status) = overwatch_handle
+                        .status_watcher::<$service_type>()
+                        .await
+                        .wait_for(::overwatch::services::status::ServiceStatus::Ready, timeout)
+                        .await
+                    {
+                        let service_id = <RuntimeServiceId as ::overwatch::services::AsServiceId<$service_type>>::SERVICE_ID;
+                        let service_status_entry = $crate::overwatch::status::ServiceStatusEntry::<RuntimeServiceId>::from_overwatch(service_id, service_status);
+                        return Err(service_status_entry);
```
danielSanchezQ, line 73: I'm not sure this would work as intended from the scope actually. Reason is that `return` would skip the scope at all and fuse the overall function. Maybe we could wrap the newly added scope into a function:

```rust
move || {
    ... all stuff as it is now
}
```

strinnityk: That inner scope shouldn't have that problem as it's async, but the outer one indeed has that problem. Very nice catch!

danielSanchezQ (approved):

PR #1333, opened 2025-06-05
fix(membership): assert block number and latest block number in update

pradovic:
This PR is a small fix in membership service mock backed. The update method needs to assert that the new finalized block is > then current latest

### DIFF
--- a/nomos-services/membership/src/backends/mock.rs
+++ b/nomos-services/membership/src/backends/mock.rs
@@ -70,6 +70,12 @@ impl MembershipBackend for MockMembershipBackend {
     ) -> Result<HashMap<ServiceType, MembershipProviders>, MembershipBackendError> {
         let block_number = update.block_number;
 
+        if block_number <= self.latest_block_number {
+            return Err(MembershipBackendError::Other(
+                "Block number is not greater than the latest block number".into(),
+            ));
+        }
+
         let mut latest_entry = self
             .membership
             .get(&self.latest_block_number)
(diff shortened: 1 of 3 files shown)


nomos-services/membership/src/backends/mock.rs
```diff
@@ -70,6 +70,12 @@ impl MembershipBackend for MockMembershipBackend {
     ) -> Result<HashMap<ServiceType, MembershipProviders>, MembershipBackendError> {
         let block_number = update.block_number;
 
+        if block_number <= self.latest_block_number {
+            return Err(MembershipBackendError::Other(
```
bacv: This case deserves it's own error I think, like `MembershipBackendError::BlockFromPast`. This change could also introduce `thiserror` to the MembershipBackendError where we could add the description. I would keep `Other` variant for db, io or implementation specific errors.

pradovic: It depends what we want to do in this case, but I agree this is the most flexible way. Will add the suggestion, thanks

pradovic: https://github.com/logos-co/nomos/pull/1333/commits/0865970c8bece87a21d7483d1cb38820aa872ace. @bacv Let me know if this is better

nomos-services/membership/src/backends/mod.rs
```diff
@@ -18,11 +18,13 @@ pub struct Settings {
     historical_block_delta: u64,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Error)]
 pub enum MembershipBackendError {
-    Other(DynError),
-    Sdp(SdpBackendError),
```
bacv, line 24: Nice, if needed we'll reintroduce them.

bacv (approved):
👍

PR #1336, opened 2025-06-06
feature(mantle): SignedMantleTx in node

bacv:

Use SignedMantleTx and MantleTx as mempool payload and item instead of old `Tx`. 

Mempool:
- Mempool receives payloads from network or other services (like http api) as `SignedMantleTx`.
- `SignedMantleTx` is supposed to be verified by the verifier and then passed into the "pool" as `MantleTx`.
- In parallel, the original `SignedMantleTx` is passed back to the network for further propagation.

Consensus:
- When proposing a block - picks items from the pool as `MantleTx`.
- When processing a block (containing only transaction hashes) - passes transaction to a processor (not yet implemented) where pool is queried for `MantleTx`s and further applied to the ledgers.

Mempool is not yet fully specified, this PR proposes to verify `SignedMantleTx` before adding them to the mempool as `MantleTx`.

@bacv @danielSanchezQ 

Yes

No

## Checklist

* [x] 1. Description added.


nomos-services/mempool/src/tx/service.rs
```diff
@@ -200,7 +201,7 @@ where
     )]
     fn handle_mempool_message(
         &mut self,
-        message: MempoolMsg<Pool::BlockId, Pool::Item, Pool::Item, Pool::Key>,
+        message: MempoolMsg<Pool::BlockId, NetworkAdapter::Payload, Pool::Item, Pool::Key>,
         network_relay: OutboundRelay<BackendNetworkMsg<NetworkAdapter::Backend, RuntimeServiceId>>,
     ) {
         match message {
```
bacv, line 207: In the case of `MempoolMsg::Add`, we'd verify SignedMantleTx (aka Payload) before adding it as MantleTx to the pool. Verification would be added in subsequent PRs.

nomos-services/api/src/http/consensus/cryptarchia.rs
```diff
@@ -204,6 +230,7 @@ where
         + 'static
         + AsServiceId<
             Cryptarchia<
+                SignedTx,
                 Tx,
```
bacv, line 234: `SignedTx` here is only required for Overwatch, consensus operates on the MantleTx, similar to the mempool, transaction processing would be performed on all operations in one place.

tests/src/nodes/validator.rs
```diff
@@ -153,15 +154,15 @@ impl Validator {
...
+    pub async fn get_block(&self, id: HeaderId) -> Option<Block<MantleTx, BlobInfo>> {
         CLIENT
             .post(format!("http://{}{}", self.addr, STORAGE_BLOCK))
             .header("Content-Type", "application/json")
             .body(serde_json::to_string(&id).unwrap())
             .send()
             .await
             .unwrap()
-            .json::<Option<Block<Tx, BlobInfo>>>()
+            .json::<Option<Block<MantleTx, BlobInfo>>>()
```
bacv, line 165: For now Block contains full mantle tx, in the future this should be just transaction hash

zeegomo (commented):
> SignedMantleTx is supposed to be verified by the verifier and then passed into the "pool" as MantleTx.

I don't really get this point, why removing auth data from the pool? Is this part of the spec? How can I verify a block offline if auth data is not in there?

danielSanchezQ:
> > SignedMantleTx is supposed to be verified by the verifier and then passed into the "pool" as MantleTx.
> 
> I don't really get this point, why removing auth data from the pool? Is this part of the spec? How can I verify a block offline if auth data is not in there?

There is no spec for mempool afaik (although it should be). The idea is that once you verify the signatures you do not need them anymore (?)

danielSanchezQ:
> Let's say I'm a validator node (i.e. I don't participate in consensus, I just follow the chain and validate), how am I supposed to check whether transactions are valid or not? Do I have to follow the mempool? That would be much more heavy on resources

I get the point. But how are you getting the blocks then and txs then? Can we move this convo to discord, it important I think.

bacv:
> Let's say I'm a validator node (i.e. I don't participate in consensus, I just follow the chain and validate), how am I supposed to check whether transactions are valid or not?

> I get the point. But how are you getting the blocks then and txs then? Can we move this convo to discord? it important I think.

Closing this PR as it's based on misunderstanding of the difference of block proposal and actual block. The updated SignedMantleTx usage in node https://github.com/logos-co/nomos/pull/1358

PR #1337, opened 2025-06-06
chore: migrate lint settings from cargo config to cargo TOML

ntn-x2:

WIP.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 1. Description added.


nomos-core/cl/Cargo.toml
```diff
@@ -12,3 +12,163 @@ rand      = "0.8.5"
 rand_core = "0.6.0"
 serde     = { version = "1.0", features = ["derive"] }
 sha2      = "0.10"
+
+[lints.clippy]
```
youngjoon-lee, line 16: Yeah. I see this downside you mentioned. For now it seems that there's no better way as long as we specify lints in `Cargo.toml`. If it's the recommended way, I wouldn't complain about it.

ntn-x2: Yeah we have few crates that are not part of the workspace, and they have their own `Cargo.toml` definition, which means we need to repeat them there. I looked into pulling a TOML definition from another TOML file, but TOML does not support that, so we cannot have a single definition for the lints and pull them everywhere. But yes, this is the first-class support by cargo for custom clippy and rustc lints. What we currently have with the `.cargo/config.toml` hack is an abuse given the previous lack of a better way.

danielSanchezQ (approved):

PR #1339, opened 2025-06-09
feat(sync): bootstrapping block fetching logic

andrussal:

This PR implements fetching blocks from storage for Cryptarchia Bootstrapping. Additionally, it adds functionality to listen incoming `DownloadBlockRequest`  in Cryptarchia service and request `DownloadBlockRequest` from Cryptarchia service.

- Implement `BlockProvider` to provide stream of blocks from storage(to answer `DownloadBlockRequest` from a peer)
- Add `request_blocks_from_peer` function to `NetworkAdapter` to request stream of blocks from a peer
- Listen incoming block download requests in Cryptarchia service and call `BlockProvider` to provide the block stream

@andrussal 

Yes

N/A

## Checklist

* [x] 1. Description added.


nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// Returns up to MAX_NUMBER_OF_BLOCKS blocks from a known block towards the
+    /// tip, in parent-to-child order.
+    ///
+    /// ## Performance
+    /// - Assumes approximately 1 million blocks per year.
+    /// - Worst case (from genesis): ~100ms
+    pub async fn get_blocks(
+        &self,
+        cryptarchia: &Cryptarchia<State>,
+        target_block: Option<HeaderId>,
```
youngjoon-lee:
```suggestion
        target_block: HeaderId,
```

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
```diff
@@ -0,0 +1,113 @@
...
+        known_blocks: Vec<HeaderId>,
+    ) -> Result<Vec<Block<Tx, BlobCertificate>>, GetBlocksError> {
+        let branches = cryptarchia.consensus.branches();
+        let tip = target_block.unwrap_or_else(|| cryptarchia.tip());
```
youngjoon-lee:
```suggestion
```

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
```diff
@@ -0,0 +1,113 @@
...
+        Self {
+            storage_relay,
+            _phantom: PhantomData,
+            _phantom_tx: PhantomData,
+            _phantom_blob: PhantomData,
+        }
+    }
+
+    /// Returns up to MAX_NUMBER_OF_BLOCKS blocks from a known block towards the
+    /// tip, in parent-to-child order.
```
youngjoon-lee:
```suggestion
    /// target block, in parent-to-child order.
```

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
```diff
@@ -0,0 +1,113 @@
...
+            if known_set.contains(&id) {
+                break;
+            }
+            if path.len() == MAX_NUMBER_OF_BLOCKS {
+                path.pop_back();
+            }
+            path.push_front(id);
+            current = branches.get(&id).and_then(|b| Some(b.parent()));
+        }
+
```
youngjoon-lee: This is wrong. The `KnownBlocks` may contain blocks that are not ancestors of the target block. Please see the [example diagram](https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df8138b041c737c9e0071c) in the spec.

I think this code should look like (pseudocode):

```rust
let start_block = max_lca(cryptarchia, target_block, known_blocks);
let path = compute_path(cryptrarchia, start_block, target_block, LIMIT);

fn max_lca(cryptarchia: cryptarchia_engine::Cryptarchia, target_block: HeaderId, known_blocks: Vec<HeaderId>) -> HeaderId {
    let max_lca = known_blocks.iter().map(|known_block| {
        cryptarchia.lca(
            cryptarchia.get(&target_block).unwrap(),
            cryptarchia.get(&known_block).unwrap(),
        );
    }).max_by(|lca| lca.length).unwrap();

    max_lca.id
}

fn compute_path(cryptarchia: cryptarchia_engine::Cryptarchia, start_block: HeaderId, target_block: HeaderId, limit: usize) {
    let mut path = VecDeque::new();
    let mut current = target_block;
    while let Some(id) = current {
        if id == start_block {
            break;
        }
        if path.len() == limit {
            path.pop_back();
        }
        path.push_front(id);
        current = cryptarchia.get(&id).and_then(|block| block.parent);
    }
}
```

youngjoon-lee: Later, we can consider computing path while finding LCA, as an optimization. But, later.

youngjoon-lee: Also, I expect that the `known_blocks` contains `local_tip` and `latest_immutable_block` as well. as defined in the [spec](https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81d2933be066852026ac).

andrussal: @youngjoon-lee I made changes based on your input. Please take a look.

andrussal: I am not sure if we choose to prune cryptarchia engine state at some point. But it should be cheap to keep some lightweight data around long time because 1mln `HeaderId`s would be just ~30Mb.

nomos-services/cryptarchia-consensus/src/network/adapters/libp2p.rs
```diff
@@ -150,4 +153,46 @@ where
...
+        peer: Self::PeerId,
+        target_block: HeaderId,
+        local_tip: HeaderId,
+        latest_immutable_block: HeaderId,
+        additional_blocks: HashSet<HeaderId>,
+    ) -> Result<BoxedStream<Result<Block<Self::Tx, Self::BlobCertificate>, DynError>>, DynError>
+    {
+        let peer = PeerId::from_str(&peer).map_err(|e| Box::new(e) as DynError)?;
+
+        let (reply_sender, mut receiver) = mpsc::channel(1);
```
danielSanchezQ: Why is this a `mpsc` instead of a `oneshot`?

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// a known block towards the `target_block`, in parent-to-child order.
+    /// The stream yields blocks one by one and terminates early if an error
+    /// is encountered.
+    pub async fn send_blocks<State, Tx, BlobCertificate>(
+        &self,
+        cryptarchia: &Cryptarchia<State>,
+        target_block: HeaderId,
+        peer_tip: HeaderId,
+        latest_immutable_block: HeaderId,
+        additional_blocks: HashSet<HeaderId>,
```
youngjoon-lee:
```suggestion
        peer_tip: HeaderId,
        peer_latest_immutable_block: HeaderId,
        peer_additional_blocks: HashSet<HeaderId>,
```
or, please consider reusing the struct if possible.

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
```diff
@@ -0,0 +1,192 @@
...
+        replay_sender: Sender<BoxStream<'static, Result<Bytes, DynError>>>,
```
youngjoon-lee:
```suggestion
        reply_sender: Sender<BoxStream<'static, Result<Bytes, DynError>>>,
```
reply, probably?

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
youngjoon-lee: or, if you think it's not proper to reuse the existing struct, we can just make this function accept a set of known blocks including peer_tip and peer_lib.

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -1369,6 +1369,45 @@ where
...
+            }
+            ChainSyncEvent::ProvideTipRequest { reply_sender } => {
+                let tip = cryptarchia.tip();
+                info!("Sending tip {tip}");
+                if let Err(e) = reply_sender.send(tip).await {
+                    error!("Failed to send tip header: {e}");
+                }
+            }
+        }
+    }
```
danielSanchezQ: Are we sure we want consensus to serve this? Ideally this could be retrieved from storage. We could cached some stuff if we need. For example caching the tip into storage so it can be retrieved. wdyt?

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
```diff
@@ -0,0 +1,192 @@
...
+        target_block: HeaderId,
+        known_blocks: &[HeaderId],
+    ) -> Option<HeaderId> {
+        let target_branch = branches.get(&target_block)?;
+
+        known_blocks
+            .iter()
+            .filter_map(|known| {
+                let know_branch = branches.get(known);
+                know_branch.map(|known_branch| branches.lca(known_branch, target_branch))
```
youngjoon-lee:
```suggestion
                branches
                    .get(known)
                    .map(|known_branch| branches.lca(known_branch, target_branch))
```
I think we can just do like this? The variable name `know_branch` doesn't looks proper.

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
```diff
@@ -0,0 +1,192 @@
...
+            .max_by_key(Branch::length)
+            .map(|b| b.id())
+    }
+
+    fn compute_path(
+        branches: &Branches<HeaderId>,
+        start_block: HeaderId,
+        target_block: HeaderId,
+        limit: usize,
+    ) -> VecDeque<HeaderId> {
```
youngjoon-lee: I think it's safer to define a rule for this function like below:
"The 1st element in the returned Vec must be the `start_block`"
If not, this function is not so useful because the protocol defines that the responder must return a block stream that starts with a block which can be added to the requester's block tree.
If it's failed to find the `start_block` inside this function, an error should be returned. I didn't mention this in https://github.com/logos-co/nomos/pull/1339#discussion_r2135505897 (because I just wrote a simple pseudo code).

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
```diff
@@ -0,0 +1,192 @@
...
+                path.pop_back();
+            }
+
+            path.push_front(id);
+
+            let parent = branches.get(&id).map(Branch::parent);
+
+            if Some(id) == parent {
+                break;
+            }
```
youngjoon-lee: Is this for checking if the `id` is the genesis block ID? then, at least we need to add a comment. Or, we can probably define a function somewhere to check if a block is the genesis.

nomos-services/cryptarchia-consensus/src/lib.rs
andrussal: Yes I think ideally this is not part of consensus because it can read everything from storage(although it can be quite heavy on storage and slow because each `DownloadBlocksRequest` can trigger million of db get requests to find the slice of blocks to return from `starting block` to `MAX_NUMBER_OF_BLOCKS` by following child->parent pointers. But we can optimise it later).  

I am wondering how it should call storage directly though. Should we change spec to use http? Or create some kind of `ConsensusSyncService`? `ConsensusSyncService` is maybe overkill but fits better with new networking protocol.

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
andrussal: I will add this check as well to be safe. (Currently it just assumes that it's called only when we found `start_block`).

nomos-services/cryptarchia-consensus/src/lib.rs
danielSanchezQ: > Yes I think ideally this is not part of consensus because it can read everything from storage(although it can be quite heavy on storage and slow because each `DownloadBlocksRequest` can trigger million of db get requests to find starting block by following child->parent pointers. But we can optimise it later).
> 
> I am wondering how it should call storage directly though. Should we change spec to use http? Or create some kind of `ConsensusSyncService`? `ConsensusSyncService` is maybe overkill but fits better with new networking protocol.

True, some service would have to take reponsibility on this as it cannot be handled by networking. Uhm, lets leave it to consensus for now. We can refactor it later if we see it's a bottleneck. Thanks.

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
andrussal: This is yes for genesis check. I will add just comment.

nomos-services/cryptarchia-consensus/src/network/adapters/libp2p.rs
andrussal: It should be `oneshot`. Thank you!

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
andrussal: It might be not be possible to use existing(`libp2p`) struct here. I'll use single collection instead.

nomos-services/cryptarchia-consensus/src/lib.rs
youngjoon-lee: > Yes I think ideally this is not part of consensus because it can read everything from storage(although it can be quite heavy on storage and slow because each DownloadBlocksRequest can trigger million of db get requests to find starting block by following child->parent pointers. But we can optimise it later).

Just FYI,
In the current codebase, a `DownloadBlocksRequest` won't trigger millions of DB calls because a node keeps all historical blocks of the honest chain in memory. Once we implement pruning blocks older than LIB from memory in the future, we will encounter the issue that you mentioned. I have some optimization idea. Let's talk later.

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
```diff
@@ -153,9 +157,13 @@ impl<Storage: StorageBackend + 'static> BlockProvider<Storage> {
         start_block: HeaderId,
         target_block: HeaderId,
         limit: usize,
-    ) -> VecDeque<HeaderId> {
+    ) -> Result<VecDeque<HeaderId>, GetBlocksError> {
         let mut path = VecDeque::new();
 
+        if start_block == target_block {
+            return Ok(path);
+        }
```
youngjoon-lee: I think we should return a vector that includes only the target block, because the downloader wants to get a stream of blocks that ends with the target block if available.

youngjoon-lee: This function can be refactored like this. I think this is cleaner, but this is just my suggestion to address https://github.com/logos-co/nomos/pull/1339/files#r2161615762
```rust
    fn compute_path(
        branches: &Branches<HeaderId>,
        start_block: HeaderId,
        target_block: HeaderId,
        limit: usize,
    ) -> Result<VecDeque<HeaderId>, GetBlocksError> {
        let mut path = VecDeque::new();

        let mut current = target_block;
        loop {
            // Push the current block, and evit the last block if the limit is exceeded.
            path.push_front(current);
            if path.len() > limit {
                path.pop_back();
            }

            // If we reached the start block, return the path.
            if current == start_block {
                return Ok(path);
            }

            // Find the parent of the current block.
            match branches.get(&current).map(Branch::parent) {
                Some(parent) => {
                    // If we reacahed the genesis block
                    if parent == current {
                        return Err(GetBlocksError::InvalidState(format!(
                            "Genesis block reached before reaching start_block: {start_block:?}"
                        )));
                    }
                    current = parent;
                }
                None => {
                    return Err(GetBlocksError::InvalidState(format!(
                        "Couldn't reach start_block: {start_block:?}"
                    )));
                }
            }
        }
    }
```

youngjoon-lee: Also, please add tests to check if all logic in `block_provider.rs` is correct.

andrussal: I think above structure encapsulates better parent block edge cases, changed to this.

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
```diff
@@ -206,3 +206,115 @@ where
...
+            .expect("Failed to add block");
+
+        let branches = cryptarchia.branches();
+
+        let start_block = [1; 32];
+        let target_block = [2; 32];
+
+        let path = compute_path(branches, start_block, target_block, MAX_NUMBER_OF_BLOCKS).unwrap();
+
+        assert_eq!(path.len(), 2);
```
youngjoon-lee, line 237: We need to check what the `path` contains.

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
```diff
@@ -0,0 +1,320 @@
...
+
+        let branches = cryptarchia.branches();
+
+        let start_block = [1; 32];
+        let target_block = [limit as u8; 32];
+        let path = compute_path(branches, start_block, target_block, limit).unwrap();
+
+        assert_eq!(path.len(), limit);
+        assert_eq!(path.front().unwrap(), &start_block);
+        assert_eq!(path.back().unwrap(), &target_block);
```
youngjoon-lee:
```suggestion
        let start_block = [1; 32];
        let target_block = [limit + 10 as u8; 32];
        let path = compute_path(branches, start_block, target_block, limit).unwrap();

        assert_eq!(path.len(), limit);
        assert_eq!(path.front().unwrap(), &start_block);
        assert_ne!(path.back().unwrap(), &target_block);
```
In this test, do you probably want to test a case where the target block is not included in the `path` due to the limit? If so, this test should be changed like this. 
If not, we need additional test for the case I mentioned.

nomos-services/cryptarchia-consensus/src/network/adapters/libp2p.rs
```diff
@@ -66,7 +68,7 @@ where
     type Settings = LibP2pAdapterSettings;
     type Tx = Tx;
     type BlobCertificate = BlobCert;
-    type PeerId = PeerId;
+    type PeerId = String;
```
youngjoon-lee: Why do we change this type to `String`? It seems that you anyway convert this `String` to `PeerId` below whenever it is used, right?

andrussal: I wasn't sure what type is convenient to use by `Requestor`. I was concerned that `PeerId` is probably behind `'libp2p` feature. But we need some type mapping anyways between `Requestor` `PeerId` and the adapter `PeerId`. I'll change it back to `PeerId` and we can see later what is the best way.

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
andrussal: I updated the `test_compute_long_path` to have only 3 blocks and check result more precisely.

nomos-services/cryptarchia-consensus/src/sync/block_provider.rs
andrussal: I will add this check as well.

PR #1343, opened 2025-06-09
feat(nat): add nat traversal state machine, integrate the state machine into nat behaviour 

CHr15F0x:

The PR adds the core nat traversal state machine and integrates it into the nat behaviour.

Yes.

Spec: @CHr15F0x @ntn-x2 
PR: @CHr15F0x

Yes.

Specification: [here](https://www.notion.so/Nomos-P2P-Network-NAT-Solution-206261aa09df81c2a8c2ecc0f9365664).

No, however to keep the size of the PR as small as possible, the PR does not implement all the features described in the spec, i.e.:
- The address mapper behaviour is a dummy that always returns a failure.
- The autonat client behaviour does not allow to retest previously successfully confirmed external addresses, which is marked with a `TODO`.

The missing features will be addressed in follow-up PRs.

## Checklist

* [x] 1. Description added.

### DIFF
--- a/nomos-libp2p/src/behaviour/nat/state_machine/transitions/mapped_public.rs
+++ b/nomos-libp2p/src/behaviour/nat/state_machine/transitions/mapped_public.rs
@@ -107,7 +107,7 @@ mod tests {
         assert_eq!(rx.try_recv(), Err(TryRecvError::Empty));
     }
 
-    #[should_panic]
+    #[should_panic = "State<MappedPublic>: Swarm confirmed external address /memory/1, but /memory/0 was expected"]
     #[test]
     fn address_mismatch_in_external_address_confirmed_event_causes_panic() {
         let (tx, _) = unbounded_channel();
@@ -117,7 +117,7 @@ mod tests {
         state_machine.on_test_event(event);
     }
 
-    #[should_panic]
+    #[should_panic = "State<MappedPublic>: Autonat client reported address /memory/1, but /memory/0 was expected"]
     #[test]
     fn address_mismatch_in_autonat_ok_event_causes_panic() {
         let (tx, _) = unbounded_channel();
@@ -127,7 +127,7 @@ mod tests {
         state_machine.on_test_event(event);
     }
 
-    #[should_panic]
+    #[should_panic = "State<MappedPublic>: Autonat client reported address /memory/1, but /memory/0 was expected"]
     #[test]
     fn address_mismatch_in_autonat_failed_event_causes_panic() {
         let (tx, _) = unbounded_channel();
--- a/nomos-libp2p/src/behaviour/nat/state_machine/transitions/public.rs
+++ b/nomos-libp2p/src/behaviour/nat/state_machine/transitions/public.rs
@@ -106,7 +106,7 @@ mod tests {
         assert_eq!(rx.try_recv(), Err(TryRecvError::Empty));
     }
 
-    #[should_panic]
+    #[should_panic = "State<Public>: Swarm confirmed external address /memory/1, but /memory/0 was expected"]
     #[test]
     fn address_mismatch_in_external_address_confirmed_event_causes_panic() {
         let (tx, _) = unbounded_channel();
@@ -116,7 +116,7 @@ mod tests {
         state_machine.on_test_event(event);
     }
 
-    #[should_panic]
+    #[should_panic = "State<Public>: Autonat client reported address /memory/1, but /memory/0 was expected"]
     #[test]
     fn address_mismatch_in_autonat_ok_event_causes_panic() {
         let (tx, _) = unbounded_channel();
@@ -126,7 +126,7 @@ mod tests {
         state_machine.on_test_event(event);
     }
 
-    #[should_panic]
+    #[should_panic = "State<Public>: Autonat client reported address /memory/1, but /memory/0 was expected"]
     #[test]
     fn address_mismatch_in_autonat_failed_event_causes_panic() {
         let (tx, _) = unbounded_channel();
--- a/nomos-libp2p/src/behaviour/nat/state_machine/transitions/test_if_mapped_public.rs
+++ b/nomos-libp2p/src/behaviour/nat/state_machine/transitions/test_if_mapped_public.rs
@@ -78,7 +78,7 @@ mod tests {
         assert_eq!(rx.try_recv(), Err(TryRecvError::Empty));
     }
 
-    #[should_panic]
+    #[should_panic = "State<TestIfMappedPublic>: Autonat client reported address /memory/1, but /memory/0 was expected"]
     #[test]
     fn address_mismatch_in_external_address_confirmed_event_causes_panic() {
         let (tx, _) = unbounded_channel();
@@ -88,7 +88,7 @@ mod tests {
         state_machine.on_test_event(event);
     }
 
-    #[should_panic]
+    #[should_panic = "State<TestIfMappedPublic>: Autonat client reported address /memory/1, but /memory/0 was expected"]
     #[test]
     fn address_mismatch_in_autonat_failed_event_causes_panic() {
         let (tx, _) = unbounded_channel();
--- a/nomos-libp2p/src/behaviour/nat/state_machine/transitions/test_if_public.rs
+++ b/nomos-libp2p/src/behaviour/nat/state_machine/transitions/test_if_public.rs
@@ -80,7 +80,7 @@ mod tests {
         assert_eq!(rx.try_recv(), Ok(Command::MapAddress(ADDR.clone())));
     }
 
-    #[should_panic]
+    #[should_panic = "State<TestIfPublic>: Autonat client reported address /memory/1, but /memory/0 was expected"]
     #[test]
     fn address_mismatch_in_external_address_confirmed_event_causes_panic() {
         let (tx, _) = unbounded_channel();
@@ -90,7 +90,7 @@ mod tests {
         state_machine.on_test_event(event);
     }
 
-    #[should_panic]
+    #[should_panic = "State<TestIfPublic>: Autonat client reported address /memory/1, but /memory/0 was expected"]
     #[test]
     fn address_mismatch_in_autonat_failed_event_causes_panic() {
         let (tx, _) = unbounded_channel();
--- a/nomos-libp2p/src/behaviour/nat/state_machine/transitions/try_map_address.rs
+++ b/nomos-libp2p/src/behaviour/nat/state_machine/transitions/try_map_address.rs
@@ -80,7 +80,7 @@ mod tests {
         assert_eq!(rx.try_recv(), Err(TryRecvError::Empty));
     }
 
-    #[should_panic]
+    #[should_panic = "State<TryMapAddress>: Address mapper reported failure for address /memory/1, but /memory/0 was expected"]
     #[test]
     fn address_mismatch_causes_panic() {
         let (tx, _) = unbounded_channel();


nomos-libp2p/src/behaviour/nat/state_machine.rs
```diff
@@ -0,0 +1,737 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum State {
+    Uninitialized(Uninitialized),
+    TestIfPublic(TestIfPublic),
+    TryAddressMapping(TryAddressMapping),
+    TestIfMappedPublic(TestIfMappedPublic),
+    Public(Public),
+    MappedPublic(MappedPublic),
+    Private(Private),
+}
```
danielSanchezQ: I think it would be better to have this as type state:

```rust
trait NATState {} // marker trait
impl NATState for Uninitialized {}
...
struct State<NATState> {
    state: NATState
}
```
Then we encode the transitions returnin a new `State<NATState>`

```rust
impl State<Uninitialize> {
    pub fn test_public(self) -> State<TestIfPublic>;
}
```
Then this can be wrapped into an abstracted struct that handles everything over a State trait, so in the upper layer you do not have to deal with extra types all the time.

We can help with this if you are not familiar with it.

CHr15F0x: As per the call, the plan is to:
- use dynamic dispatch for states at the upper layer instead of the state wrapper enum,
- use dynamic dispatch for the events at the upper layer instead of `TryInto<EventTypeForStateA> + .. TryInto<EventTypeForStateZ>`

nomos-libp2p/src/behaviour/nat/address_mapper.rs
```diff
@@ -0,0 +1,102 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum Event {
+    AddressMappingFailed(Multiaddr),
+    DefaultGatewayChanged,
+    LocalAddressChanged(Multiaddr),
+    NewExternalMappedAddress(Multiaddr),
+}
+
+impl AddressMapperBehaviour {
+    pub const fn new() -> Self {
```
pradovic: Maybe `default()` instead of new? It could be derived as well I think

nomos-libp2p/src/behaviour/nat/mod.rs
```diff
@@ -1,37 +1,101 @@
...
+        let autonat_client_behaviour = Toggle::from(
+            autonat_client_config.map(|config| autonat::v2::client::Behaviour::new(rng, config)),
+        );
+
+        let (command_tx, command_rx) = tokio::sync::mpsc::unbounded_channel();
+
+        let state_machine = StateMachine::new(command_tx);
 
         Self {
+            static_listen_addr: None,
```
pradovic, line 41: Is `static_listen_addr` configured somewhere? Sorry, I am probably just missing to notice in the changeset.

nomos-libp2p/src/behaviour/nat/address_mapper.rs
CHr15F0x: Yeah I think we can make it `default()` for the time being and switch back to `new()` if there's some config to pass to the behaviour.

nomos-libp2p/src/behaviour/nat/mod.rs
CHr15F0x: If the autonat client config is disabled, we assume that whatever address of the node configured must be a publicly reachable one. So essentially this field is a way to pass data from `on_swarm_event()` to `poll()`:

```
    fn on_swarm_event(&mut self, event: FromSwarm) {
        if self.autonat_client_behaviour.is_enabled() {
            self.state_machine.on_event(event);
            self.autonat_client_behaviour.on_swarm_event(event);
        } else if let FromSwarm::NewListenAddr(NewListenAddr { addr, .. }) = event {
            /// ----> FROM HERE
            self.static_listen_addr = Some(addr.clone());
        }
    }

    fn poll(
        &mut self,
        cx: &mut Context<'_>,
        if let Some(addr) = self.static_listen_addr.take() {
            /// ----> TO HERE
            return Poll::Ready(ToSwarm::ExternalAddrConfirmed(addr));
        }
        
        (...)
    }
```

pradovic: Makes sense. Thanks for the explanation!

nomos-libp2p/src/behaviour/nat/address_mapper.rs
CHr15F0x: https://github.com/logos-co/nomos/pull/1343/commits/778b9e5795cc9ab67e87c4245af1a2a0327b640c

nomos-libp2p/src/behaviour/nat/state_machine.rs
```diff
@@ -0,0 +1,751 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Command {
+    ScheduleAutonatClientTest(Multiaddr),
+    MapAddress(Multiaddr),
+    NewExternalAddrCandidate(Multiaddr),
+}
+
+#[derive(Debug)]
+pub struct StateMachine {
+    inner: Option<Box<dyn OnEvent<Event>>>,
```
danielSanchezQ: Why is this optional?

nomos-libp2p/src/behaviour/nat/state_machine.rs
```diff
@@ -0,0 +1,751 @@
...
+}
+
+#[cfg(test)]
+trait DynPartialEq {
+    fn box_eq(&self, other: &dyn std::any::Any) -> bool;
+    fn as_any(&self) -> &dyn std::any::Any;
+}
+
+impl<S, E> State<S, E> {
+    fn into_boxed<T, C>(self, next_state_ctor: C) -> Box<State<T, E>>
```
danielSanchezQ: nitpick: I think other crates use simply `boxed`.

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: This is a trick that allows a "swap old state for new state" pattern, while holding a `&mut` ref to the wrapping struct, by using `Option::take()` on the current state. Without an `Option<>` we get "cannot move out of `self.inner` which is behind a mutable reference".

I wanted the api of the state machine to operate on `&mut self`, otherwise the caller/user of the state machine would have to do a similar thing manually everywhere an event is passed into the state machine in the behaviour.

You can see the usage of the very `Option<>` in the code a few lines below:

```
impl StateMachine {
    pub fn new(command_tx: UnboundedSender<Command>) -> Self {
        (...)
    }

    pub fn on_event<E>(&mut self, event: E)
    where
        E: TryInto<Event, Error = ()>,
    {
        let current_state = self.inner.take().expect("State to be Some");

        match event.try_into() {
            Err(()) => {
                // Ignore unrecognized events
                self.inner = Some(current_state);
            }
            Ok(event) => {
                self.inner = Some(current_state.on_event(event, &self.command_tx));
            }
        }
    }
}

File: nomos-libp2p/src/behaviour/nat/state_machine.rs (hunk 5 above)
Comment (CHr15F0x) [reply nomos-libp2p/src/behaviour/nat/state_machine.rs]: https://github.com/logos-co/nomos/pull/1343/commits/9613546badfb79081e2cbf0f44eb17954c3eabff

**Review by pradovic — APPROVED** [review]
LGTM!

File: nomos-libp2p/src/behaviour/nat/state_machine.rs [hunk 6]
```diff
@@ -0,0 +1,749 @@
...
+
+impl CommandTx {
+    fn send(&self, command: Command) {
+        self.tx.send(command).expect("Channel not to be closed");
+    }
+}
+
+#[cfg(not(test))]
+trait OnEvent<E>: Debug + Send {
+    fn on_event(self: Box<Self>, event: E, command_tx: &CommandTx) -> Box<dyn OnEvent<E>>;
```
danielSanchezQ: Checking here, why boxed self?

You can implement `OnEvent<E> for Box<T> where T: OnEvent<E>`.

nomos-libp2p/src/behaviour/nat/state_machine.rs
danielSanchezQ: Ah yes. Thanks for the explanation.

danielSanchezQ (commented):
Missed a detail. Added some comment. I think the unboxed solution is better (if possible).

nomos-libp2p/src/config/autonat_client.rs
```diff
@@ -14,7 +14,14 @@ pub struct Settings {
     /// The interval at which we will attempt to confirm candidates as external
     /// addresses.
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub probe_interval_secs: Option<u64>,
+    pub probe_interval_millisecs: Option<u64>,
+
+    #[serde(default = "default_restest_interval_millisecs")]
+    pub retest_successful_external_addresses_interval_millisecs: u64,
```
ntn-x2, line 23: Can you add a comment for this as well?

nomos-libp2p/src/config/autonat_client.rs
```diff
@@ -14,7 +14,14 @@ pub struct Settings {
     /// The interval at which we will attempt to confirm candidates as external
     /// addresses.
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub probe_interval_secs: Option<u64>,
+    pub probe_interval_millisecs: Option<u64>,
```
ntn-x2, line 17: Can we serialize/deserialize this as `Duration` and use the `MinimumBoundDuration` we use elsewhere, so that we can force it to be, for instance, at least 1ms? Cause right now we could accept a value of `0`, which I am not sure what would imply.

nomos-libp2p/src/config/autonat_client.rs
ntn-x2: Also, see my comment for `probe_interval_millisecs`.

nomos-libp2p/src/config/autonat_client.rs
```diff
@@ -14,7 +14,14 @@ pub struct Settings {
...
+
+    #[serde(default = "default_restest_interval_millisecs")]
+    pub retest_successful_external_addresses_interval_millisecs: u64,
+}
+
+fn default_restest_interval_millisecs() -> u64 {
+    60_000 // 1 minute
```
ntn-x2, line 27: Just a personal preference, it would be more readable, I think, to use `Duration::from_millisecs` and simply use a `Duration` for it.

nomos-libp2p/src/behaviour/nat/mod.rs
```diff
@@ -1,37 +1,111 @@
...
+
+type Task = BoxFuture<'static, Multiaddr>;
+
 /// This behaviour is responsible for confirming that the addresses of the node
 /// are publicly reachable.
 pub struct NatBehaviour<R: RngCore + 'static> {
+    /// The static public listen address is passed through this variable to
+    /// the `poll()` method. Unused if the node is not configured with a static
+    /// public IP address.
+    static_listen_addr: Option<Multiaddr>,
```
ntn-x2, line 28: Can't all these properties be grouped inside an enum that is given to the `NatBehaviour`? Instead of using `Option`s and `Toggle`s, we can have a variant which contains these things.

nomos-libp2p/src/behaviour/nat/state_machine.rs
```diff
@@ -0,0 +1,749 @@
...
+    }
+}
+
+impl CommandTx {
+    fn send(&self, command: Command) {
+        self.tx.send(command).expect("Channel not to be closed");
+    }
+}
+
+#[cfg(not(test))]
```
ntn-x2: This will trigger a warning in the future, since it's not recommended to use `cfg(not(test))` attributes. I would suggest to declare the `OnEvent<E>` trait once, and then have a sub-trait that extends `OnEvent<E>` and that requires `DynPartialEq`, something like `OnEventTest<E>: OnEvent<E> + DynPartialEq`, for instance. This second trait will only be available for tests.

nomos-libp2p/src/behaviour/nat/state_machine.rs
```diff
@@ -0,0 +1,749 @@
...
+            }
+            Ok(event) => {
+                self.inner = Some(current_state.on_event(event, &self.command_tx));
+            }
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+struct State<S, E> {
```
ntn-x2: If the `OnEvent<E>` already has the event type, do we need the `E` generic on the struct itself? At a first glance, to me it seems that we don't, but maybe I am missing something? For the `boxed` method, it could be a generic on the function itself. Other than that, I don't see any other reasons to keep the `E` generic around.

nomos-libp2p/src/behaviour/nat/state_machine.rs
```diff
@@ -0,0 +1,749 @@
...
+    {
+        let Self { state, .. } = self;
+        Box::new(State {
+            state: next_state_ctor(state),
+            _phantom_event: PhantomData,
+        })
+    }
+}
+
+impl OnEvent<Event> for State<Uninitialized, Event> {
```
ntn-x2: I think it would help to split this into separate modules, one for each state + `impl OnEvent<Event>` block.

nomos-libp2p/src/behaviour/nat/state_machine.rs
```diff
@@ -0,0 +1,749 @@
...
+            Event::ExternalAddressConfirmed(addr) if self.state.addr_to_test() == &addr => {
+                command_tx.send(Command::ScheduleAutonatClientTest(addr));
+                self.boxed(TestIfPublic::into_public)
+            }
+            Event::AutonatClientTestFailed(addr) if self.state.addr_to_test() == &addr => {
+                command_tx.send(Command::MapAddress(addr));
+                self.boxed(TestIfPublic::into_try_address_mapping)
+            }
+            Event::ExternalAddressConfirmed(addr) | Event::AutonatClientTestFailed(addr) => {
+                tracing::error!(
```
ntn-x2: Why is this an error? Could it be we try to test three different addresses? Or, as far as I can see, we can only test a single address?

nomos-libp2p/src/behaviour/nat/state_machine.rs
```diff
@@ -0,0 +1,749 @@
...
+            }
+            _ => self,
+        }
+    }
+}
+
+impl OnEvent<Event> for State<TryMapAddress, Event> {
+    fn on_event(self: Box<Self>, event: Event, command_tx: &CommandTx) -> Box<dyn OnEvent<Event>> {
+        match event {
+            Event::NewExternalMappedAddress(addr) => {
```
ntn-x2: Why don't we check here that `addr` is what we expect it to be?

nomos-libp2p/src/behaviour/nat/state_machine.rs
```diff
@@ -0,0 +1,749 @@
...
+            {
+                command_tx.send(Command::ScheduleAutonatClientTest(addr));
+                self
+            }
+            Event::AutonatClientTestFailed(addr) if self.state.addr() == &addr => {
+                self.boxed(MappedPublic::into_test_if_public)
+            }
+            Event::ExternalAddressConfirmed(addr) => {
+                tracing::error!(
+                    "Swarm confirmed external address {}, but {} was expected",
```
ntn-x2: Should you use different log messages for different states, so we can easily distinguish between them? Or at the very least the state we are logging for.

nomos-libp2p/src/behaviour/nat/state_machine.rs
```diff
@@ -0,0 +1,749 @@
...
+            }
+            _ => self,
+        }
+    }
+}
+
+impl OnEvent<Event> for State<Private, Event> {
+    fn on_event(self: Box<Self>, event: Event, _: &CommandTx) -> Box<dyn OnEvent<Event>> {
+        match event {
+            Event::LocalAddressChanged(addr) => self.boxed(|state| state.into_test_if_public(addr)),
```
ntn-x2: Are we guaranteed the new address is different than the old one? Should we skip in the unlikely case the new matches the old?

nomos-libp2p/src/behaviour/nat/state_machine.rs
```diff
@@ -0,0 +1,749 @@
...
+#[cfg(test)]
+mod tests {
+    use std::collections::HashSet;
+
+    use fixtures::*;
+    use tokio::sync::mpsc::{error::TryRecvError, unbounded_channel};
+
+    use super::*;
+
+    #[cfg(test)]
```
ntn-x2: We are in the tests already, so this is unnecessary.

ntn-x2: All of them are.

nomos-libp2p/src/behaviour/nat/state_machine.rs
```diff
@@ -0,0 +1,749 @@
...
+        init_state: Box<dyn OnEvent<Event>>,
+        command_tx: UnboundedSender<Command>,
+    ) -> StateMachine {
+        StateMachine {
+            inner: Some(init_state),
+            command_tx: command_tx.into(),
+        }
+    }
+
+    type TestCase<'a> = Vec<(
```
ntn-x2: What does this include? Can you add a comment or use a struct instead?

nomos-libp2p/src/behaviour/nat/state_machine.rs
```diff
@@ -0,0 +1,749 @@
...
+            Option<Command>,
+        )>,
+    )>;
+
+    #[expect(clippy::too_many_lines, reason = "Keep the test readable")]
+    #[test]
+    fn test_transitions_and_emitted_commands() {
+        let expected_transitions: TestCase<'_> = vec![
+            (
+                line!(),
```
ntn-x2: Question: why not using a string identifier instead, independently of the line number?

nomos-libp2p/src/behaviour/nat/address_mapper.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+impl AddressMapperBehaviour {
+    /// Instruct the behaviour to attempt to map the given address.
+    pub fn try_map_address(&mut self, address: Multiaddr) {
+        self.address_to_map = Some(address);
+        self.try_wake();
+    }
+
+    pub fn try_wake(&mut self) {
```
ntn-x2: I guess this is not meant to be public?

nomos-libp2p/src/behaviour/nat/address_mapper.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    fn poll(
+        &mut self,
+        cx: &mut Context<'_>,
+    ) -> Poll<ToSwarm<Self::ToSwarm, THandlerInEvent<Self>>> {
+        if let Some(address) = self.address_to_map.take() {
+            // For now we don't perform any actual mapping, so just generate a failure
+            // event.
+            return Poll::Ready(ToSwarm::GenerateEvent(Event::AddressMappingFailed(address)));
```
ntn-x2, line 89: Why don't we return the same input address to maintain backwards-compatibility?

nomos-libp2p/src/behaviour/nat/state_machine.rs
ntn-x2, line 1: This module is way too big and should be split into multiple sub-modules. I would suggest to split each state, their impl of `OnEvent` and any auxiliary test functions into a sub-module each. Tests should also go into their own module.

nomos-libp2p/src/behaviour/nat/state_machine.rs
```diff
@@ -0,0 +1,749 @@
...
+
+    pub fn on_event<E>(&mut self, event: E)
+    where
+        E: TryInto<Event, Error = ()>,
+    {
+        let current_state = self.inner.take().expect("State to be Some");
+
+        match event.try_into() {
+            Err(()) => {
+                // Ignore unrecognized events
```
ntn-x2: Should we print something in the logs?

nomos-libp2p/src/behaviour/nat/state_machine.rs
```diff
@@ -0,0 +1,749 @@
...
+}
+
+impl From<UnboundedSender<Command>> for CommandTx {
+    fn from(tx: UnboundedSender<Command>) -> Self {
+        Self { tx }
+    }
+}
+
+impl CommandTx {
+    fn send(&self, command: Command) {
```
ntn-x2: Maybe `force_send` would be a better naming since we are `expect`ing the channel not to be closed.

nomos-libp2p/src/behaviour/nat/state_machine.rs
```diff
@@ -0,0 +1,749 @@
...
+    fn on_event(self: Box<Self>, event: E, command_tx: &CommandTx) -> Box<dyn OnEvent<E>>;
+}
+
+#[cfg(test)]
+trait DynPartialEq {
+    fn box_eq(&self, other: &dyn std::any::Any) -> bool;
+    fn as_any(&self) -> &dyn std::any::Any;
+}
+
+impl<S, E> State<S, E> {
```
ntn-x2: We are not using single-letter generics anymore, and I think in this and other instances in this context it would be extremely beneficial not to.

ntn-x2 (commented):
I will take a second look and try to better understand the test cases today or tomorrow, but in the meanwhile I am leaving some feedback here. In general this is a possible way to go. But I think the code should be better modularized, as it feels like a big blob of info right now. I left some suggestions on how I would do it. Also, generic names should be more explicative, and some more comments in key points would help better understand the flow.

This is a good starting point, and I can see how it would ease the transition into the next step, which is port-mapping, later on.

nomos-libp2p/src/behaviour/nat/state_machine.rs
bacv: Any event from swarm is passed to this method, I think even debug logs would be meaningless, I suggest to leave it as is.

nomos-libp2p/src/config/autonat_client.rs
CHr15F0x: Yes I also thought about using `MinimumBoundedDuration` in those configs but it seemed like a good refactor in a separate PR so an not to inflate the current one. Wdyt?

I'd raise an issue and we'd have a separate PR to clean the config stuff up.

nomos-libp2p/src/behaviour/nat/mod.rs
CHr15F0x: Yesterday I suggested having and outer and inner nat behaviours:
- the outer one (the current one) would handle the special case of static public node and the inner behaviour disabled or !(static public) and pass through to the inner behaviour,
- the inner one would be the nat behaviour mod the static public case, all/most of the internals of that inner behaviour will then not require to be optional.
See: https://github.com/logos-co/nomos/pull/1343#issuecomment-2976079856

Pros:
- inner is simple
- outer doesn't contain any toggle/enum matching related stuff in the impls

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: Hmm I'm not sure if this will work out, I need to try it out. Will let you know how it went.

bacv (approved):
Solid work, looks good, only cfg flags left.

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: Yeah I had this on mind as well but just left it as is so as not to delay the review. I think we can just drop the generic param `E` as it's unusable now, not even for testing.

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: Well, I kinda agree and disagree. The concrete state types are simple and mixing them with the impls, which are imo **the most important part of the entire state machine code** will just obfuscate the code. I'd rather do the splitting in the following manner:
(1) keep the state types in a separate module,
(2) keep the impls, which show the flow between states in a separate module (I don't have a good name for it rn: `flow`, `logic`, ???).

(2) should be the first place to go to when browsing the code and smart naming of the member fields of each type state in (1) actually fees the reader of having to even look at (1).

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: Rn we support only 1 node address and this code will need to change if we decide to allow for more than 1 address.

As for now this error indicates an inconsistency in the entire libp2p stack that we're using that we're using - in theory the error condition should not happen.

CHr15F0x: Btw I'm not sure how we should react to such catastrophic conditions, just log and stay in the same state, like it's done rn?

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: This is just a matter of the enum variants being newtypes instead of struct-like. In this case addr means - "the external address of the NAT-box as configured via nat-pmp/pcp/upnp", which is definitely going to be something else than our local address and we must take it for granted, there's no way to predict what's a good expected value here.

Lemme rework the events to be of "struct-variant" style, so that the carried field actually has a meaningful name.

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: I think we should add per-type-state scope to the traces - wdyt?

nomos-libp2p/src/config/autonat_client.rs
danielSanchezQ: if its not gonna be implemented in this PR open an issue for it indeed.

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: Good catch!

It's a matter of what the producer of that event guarantees in his API, and we haven't agreed on that yet :see_no_evil: . Anyway lemme add the check here, just as an additional safeguard, regardless of what we decide later on on the entity that actually does the monitoring.

nomos-libp2p/src/behaviour/nat/state_machine.rs
danielSanchezQ: > This will trigger a warning in the future, since it's not recommended to use `cfg(not(test))` attributes. I would suggest to declare the `OnEvent<E>` trait once, and then have a sub-trait that extends `OnEvent<E>` and that requires `DynPartialEq`, something like `OnEventTest<E>: OnEvent<E> + DynPartialEq`, for instance. This second trait will only be available for tests.

These seems like a nice suggestion. Also all this should be moved into the tests mod

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: Essentially it's an "input + expectations" type, representing a single test case. Lemme document it (or better add a proper struct).

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: When an typical assert fails, it reports a src line. Imo it's better to report a src line because it's easier to jump into it in any ide/editor of choice than grepping for a phrase. And once you jump to a line the name of the state and the event is already there.

nomos-libp2p/src/behaviour/nat/address_mapper.rs
CHr15F0x: Absolutely!

nomos-libp2p/src/behaviour/nat/state_machine.rs
danielSanchezQ: Yeah, state machine can be split. I would go for any of you both suggested. But maybe something like?:

```
/state_machine
    mod.rs
    states.rs
    transitions.rs
```

nomos-libp2p/src/behaviour/nat/state_machine.rs
danielSanchezQ: Nice details. Thanks @ntn-x2.

nomos-libp2p/src/behaviour/nat/address_mapper.rs
CHr15F0x: I'm not sure I follow the question here - we're returning the same address that was passed to the behaviour to be mapped to some publicly-reachable address by the NAT-box, ie. what you call the input address.

The event variant is a newtype-style one so it doesn't explicitly say `Event::AddressMappingFailed{ our_local_input_address_that_we_wanted_to_map_on_the_nat_box }` but I'll fix it as per the [other comment](https://github.com/logos-co/nomos/pull/1343#discussion_r2151748976).

Or did I misunderstand the question?

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: https://github.com/logos-co/nomos/pull/1343#discussion_r2151728817

On top of that comment - I'd rather keep all test-utils in a single module, because they'd obfuscate what's important in the prod code.

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: We're peeking at literally everything that comes from the swarm (and more) so I'd either leave the code as is or use `tracing::trace!`. Wdyt?

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: Absolutely.

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: `S` will get renamed while `E` will be dropped.

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: I really like this suggestion.

ntn-x2: I have a different view on it, but will let you do as you see fit. I don't really see a point of having a file just for states, since in our case states are quite "dumb" and you rightfully pointed out that most of the logic is how we react to events. But, I would rather see this as "this file contains all the logic for when the state machine is in the `Uninitialised` state", meaning the state itself, the functions that operate on that state, and perhaps any other trait impl that belongs to the state. So when you open and scroll through `uninitialised.rs`, you see exactly what it has, and especially how do you move away from it into a different state. Also, by looking at the number and names for those files, you already get a good picture of how many states compose the state machine.
This being said, if you guys think there is a better option, by all means go ahead.

nomos-libp2p/src/behaviour/nat/state_machine.rs
ntn-x2: That is wrong, but also potentially should never happen if we are always dealing with a single address. In that case we could even panic, not sure, or even better move to a different state that supports recovery.

nomos-libp2p/src/behaviour/nat/state_machine.rs
ntn-x2: Yeah maybe renaming the `addr` parameter would help as well :)

nomos-libp2p/src/behaviour/nat/state_machine.rs
ntn-x2: Nah I think we might want to grep all logs for the state machine, so a target like `nomos-network::nat::state-machine` should be sufficient. I don' see any case in which we might want to grep logs for a specific state. If anything, the opposite so we can see what is going on. I would perhaps just add the state in there or use a different message.

nomos-libp2p/src/behaviour/nat/address_mapper.rs
ntn-x2: No what I was asking is why does this have to return an error? Could it just return ok with the input address, so that we could POTENTIALLY use it in the current stack without any breaking changes? I know it would not work since there is a reason why we are trying mapping the address, but just for the sake of "this is a minimal implementation". If you think returning an error is more appropriate, fine by me.

nomos-libp2p/src/behaviour/nat/state_machine.rs
ntn-x2: Yeah ok in that case no need to log.

nomos-libp2p/src/behaviour/nat/state_machine.rs
danielSanchezQ: > I have a different view on it, but will let you do as you see fit. I don't really see a point of having a file just for states, since in our case states are quite "dumb" and you rightfully pointed out that most of the logic is how we react to events. But, I would rather see this as "this file contains all the logic for when the state machine is in the `Uninitialised` state", meaning the state itself, the functions that operate on that state, and perhaps any other trait impl that belongs to the state. So when you open and scroll through `uninitialised.rs`, you see exactly what it has, and especially how do you move away from it into a different state. Also, by looking at the number and names for those files, you already get a good picture of how many states compose the state machine. This being said, if you guys think there is a better option, by all means go ahead.

I think this is a good approach as well. Not voting for anything here.
It is just different views, or having all transitions in a single place  or split them. I see the benefit in both approaches.
Up to you guys.

BTW, one thing that would help is to fill how it is structured in the `mod.rs` and maybe even embedding the state machine transitions diagram on it.

nomos-libp2p/src/behaviour/nat/state_machine.rs
danielSanchezQ: > That is wrong, but also potentially should never happen if we are always dealing with a single address. In that case we could even panic, not sure, or even better move to a different state that supports recovery.

is this transition invalid from the specification? If its an invalid transition then it should panic.

nomos-libp2p/src/behaviour/nat/address_mapper.rs
CHr15F0x: I misunderstood the original question, thanks for clarification.

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: >So when you open and scroll through uninitialised.rs, you see exactly what it has, and especially how do you move away from it into a different state. Also, by looking at the number and names for those files, you already get a good picture of how many states compose the state machine.

After second thought, I'm changing my mind to what @ntn-x2 suggests. Each file for each state can be structured in a way the the most important code is at the top, which solves the issue of "transitions should be instantly visible to the reader".

ntn-x2: If you really want to structure modules properly, you could also let each state have its own folder with a top-level `mod.rs` file, an `on_event.rs` file, and a `tests.rs` file, if you want to highlight the `OnEvent` implementation for each of them.

nomos-libp2p/src/behaviour/nat/address_mapper.rs
CHr15F0x: I think that returning a failure event is more appropriate here, due to the fact how the state machine works:
1. autonat for addr `A` fails
2. try mapping for addr `A`
3. returning `NewMappedExternalAddr(A)` would be misleading, as success here means we definitely should get some other address, say `A* != A`, which we cannot reliably provide at this point, so we just return `MappingFailed(A)`.

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: Added comments, because the tuples are to keep the fixture vector concise and readable at the same time.

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: I resolved this issue a bit differently, because the suggested `OnEventTest` couldn't be contained to test code only.

nomos-libp2p/src/config/autonat_client.rs
CHr15F0x: https://github.com/logos-co/nomos/issues/1384

nomos-libp2p/src/config/autonat_client.rs
CHr15F0x: Let's solve it all in a separate PR to this issue: https://github.com/logos-co/nomos/issues/1384

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: TL;DR I tried and I failed miserably.

----
It did escalate to more boilerplate and the worst part was that I couldn't just return plain `self` for cases where the state does not change. In such cases `self` still has to be boxed again, not to mention that `on_event` itself must take `&self` because you cannot move `dyn OnEvent`. A side effect of `&self` is the requirement of `Clone` for state which I just wanna avoid.

nomos-libp2p/src/behaviour/nat/state_machine.rs
CHr15F0x: Unfortunately, after some tries it turned out I have to go with my original solution:
```
/state_machine
    mod.rs
    states.rs
    transitions.rs
    (...)
```
The reason is that if we split each state into its own respective sub-module, the consuming `into_other_state` methods will not work as internal fields of each state are meant to be private. If we change the visibility of the fields to `pub(super)`, this will defeat any purpose of those methods in the `on_event` transition functions for each state as those functions will not be mandatory to construct the other state (because the fields of each state are now `pub(super)` and the `into_xyz_state` methods can by bypassed).

nomos-libp2p/src/behaviour/nat/inner.rs
```diff
@@ -0,0 +1,211 @@
...
+        state_machine::{Command, StateMachine},
+    },
+    AutonatClientSettings,
+};
+
+type Task = BoxFuture<'static, Multiaddr>;
+
+/// Provides dynamic NAT-status detection, improvement (via address mapping on
+/// the NAT-box), and periodic maintenance capabilities.
+pub struct InnerNatBehaviour<R: RngCore + 'static> {
```
CHr15F0x, line 35: Is `Inner` good enough? Maybe `Core` makes more sense?

```suggestion
pub struct CoreNatBehaviour<R: RngCore + 'static> {
```

nomos-libp2p/src/behaviour/nat/state_machine/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+    #[cfg(test)]
+    fn as_any(&self) -> &dyn core::any::Any {
+        self
+    }
+
+    #[cfg(test)]
+    fn box_eq(&self, other: &dyn core::any::Any) -> bool {
+        other.downcast_ref::<Self>() == Some(self)
+    }
+}
```
danielSanchezQ: Why can t this impl be in a test module if we are just gonna use it for test?

CHr15F0x: It's an empty trait outside tests. I can sort of move the impl to the test module (ie keep an empty impl here and add a new impl with the test specific meat in the test module) but the conditional in the trait itself must remain - I'm running out of ideas how to solve this.

CHr15F0x: Nah, I cannot even do that :arrow_up:  as I'd have conflicting implementations.

nomos-libp2p/src/behaviour/nat/inner.rs
```diff
@@ -0,0 +1,211 @@
...
+                Command::MapAddress(addr) => {
+                    self.address_mapper_behaviour.try_map_address(addr);
+                }
+                Command::NewExternalAddrCandidate(addr) => {
+                    return Poll::Ready(ToSwarm::NewExternalAddrCandidate(addr));
+                }
+            }
+        }
+
+        Poll::Pending
```
ntn-x2, line 209: Do we need to awake the waker here or in some of the branches? What happens if we return `Pending` but do not awake the waker in case, for example, we poll a command from `self.command_rx`?

nomos-libp2p/src/behaviour/nat/inner.rs
ntn-x2: I like inner more.

nomos-libp2p/src/behaviour/nat/state_machine/states.rs
```diff
@@ -0,0 +1,210 @@
...
+pub struct MappedPublic {
+    addr: Multiaddr,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Private {
+    addr: Multiaddr,
+}
+
+impl Uninitialized {
```
ntn-x2, line 21: Do you think it would not be better to have impl blocks right after the type they refer to?

nomos-libp2p/src/behaviour/nat/state_machine/transitions.rs
```diff
@@ -0,0 +1,165 @@
+use crate::behaviour::nat::state_machine::{
+    event::Event,
+    states::{
+        MappedPublic, Private, Public, TestIfMappedPublic, TestIfPublic, TryMapAddress,
+        Uninitialized,
+    },
+    Command, CommandTx, OnEvent, State,
+};
+
+impl OnEvent for State<Uninitialized> {
```
ntn-x2: Do you mind adding a comment on top of each state transition explaining what events they react to, under what conditions, and in which state they end up being?

nomos-libp2p/src/behaviour/nat/state_machine/states.rs
```diff
@@ -0,0 +1,210 @@
...
+impl super::State<Uninitialized> {
+    pub const fn new() -> Self {
+        Self {
+            state: Uninitialized(()),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Uninitialized(
```
ntn-x2, line 16: Can you add a brief comment on top of each state describing what state they actually represent?

nomos-libp2p/src/behaviour/nat/state_machine/tests.rs
```diff
@@ -0,0 +1,454 @@
...
+    )>,
+)>;
+
+#[expect(clippy::too_many_lines, reason = "Keep the test readable")]
+#[test]
+fn test_transitions_and_emitted_commands() {
+    let expected_transitions: TestCases<'_> = vec![
+        (
+            line!(),
+            Box::new(|| Uninitialized::for_test()),
```
ntn-x2: I think it would be great if we could split these tests into separate functions, each testing an expected transition. In case one test case fails, we can still run the others.

ntn-x2: At the very least one for each starting state, but ideally one for each transition. I would have personally also split the tests for each state in a separate module and put it right after the state + transition definition, but this will do fine as well.

nomos-libp2p/src/behaviour/nat/state_machine/mod.rs
ntn-x2: What was the blocker for having a test-only subtrait of `OnEvent`? Anywhere where `OnEvent` is requested, an `OnTestEvent` could be used. I think it would be a nicer choice than having a trait with test-only functions.

ntn-x2 (commented):
Much much cleaner, thank you! A left a few more comments, mostly for aiding with understanding the code, but it looks already much much better.

nomos-libp2p/src/behaviour/nat/state_machine/states.rs
CHr15F0x: No problem, lemme change it, I don't have a preference here.

nomos-libp2p/src/behaviour/nat/state_machine/tests.rs
CHr15F0x: I'm not sure I'm convinced. If this test blows up on some transition, that transition needs to be fixed. I see little utility in having so many tests and I find it useful seeing the entire flow of the entire state machine in roughly 1 screen worth of executed test cases in that single test. It's even more concise than the `transitions.rs` file.

ntn-x2: Yeah we're circling back to the same conversation we had last week. I see as a major benefit to have state, transitions and tests into the same module. You don't, and that's fine :)

CHr15F0x: I could split the modules to have per-state files that contain:
- transition impl,
- tests for that transtion.
But unfortunately those files wouldn't contain the state definitions themselves due to the fact how the typestate pattern works (I addressed this issue in some previous comment).
Are you ok with such partial solution?

nomos-libp2p/src/behaviour/nat/state_machine/mod.rs
danielSanchezQ: > What was the blocker for having a test-only subtrait of `OnEvent`? Anywhere where `OnEvent` is requested, an `OnTestEvent` could be used. I think it would be a nicer choice than having a trait with test-only functions.

➕ **1**

CHr15F0x: I removed the trait altogether. We can just compare `Debug` representations as `OnEvent` dervies from `Debug` in both test and prod code. The `Debug` representation points to concrete objects of concrete type, so there's no ambiguity or room for error. Imo this is a simple solution an frees us from the hassle of casting to `dyn Any` and then downcasting to a concrete type, having additional traits and having the cognitive load of which object has 1 vtable (`dyn OnEvent`) and which has 2 vtables (`dyn OnEventTest`) and how the casting should be done in a correct manner, and are we sure it's correct etc.

nomos-libp2p/src/behaviour/nat/state_machine/transitions/mapped_public.rs
```diff
@@ -0,0 +1,160 @@
...
+        let event = autonat_failed();
+        state_machine.on_test_event(event);
+        assert_eq!(
+            state_machine.inner.as_ref().unwrap(),
+            &TestIfPublic::for_test(ADDR.clone())
+        );
+        assert_eq!(rx.try_recv(), Err(TryRecvError::Empty));
+    }
+
+    #[should_panic]
```
ntn-x2: We don't require it yet, but the `should_panic` should include the panic message, since the function can panic for multiple reasons and we want to be sure it panics for the reason we expect.

ntn-x2 (approved):
Very nice! I think once we add the expected panic messages to the unit tests, we are good to go! Great job! :D I'll approve and let you add the expected panic message, then merge if I am AFK.

nomos-libp2p/src/behaviour/nat/inner.rs
CHr15F0x: IIUC it's [not necessary](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.UnboundedReceiver.html#method.poll_recv): 
```
When the method returns Poll::Pending, the Waker in the provided Context is scheduled to receive a wakeup when a message is sent on any receiver, or when the channel is closed. 
```

danielSanchezQ: Yeah, it doesn't look necessary here as any of the other polls should awake it when they are ready and we do not depend on anything else.

nomos-libp2p/src/behaviour/nat/state_machine/transitions/mapped_public.rs
CHr15F0x: Fixed: https://github.com/logos-co/nomos/pull/1343/commits/c352014f2cf34fa7b904eb9e2f01472fa9f698ec

PR #1345, opened 2025-06-09
feature(services): Await for service readiness (Pt. 2)

strinnityk:
- Make all services await their dependencies

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 1. Description added.


nodes/nomos-node/node/src/api/backend.rs
```diff
@@ -73,7 +75,7 @@ pub struct AxumBackend<
     DaBlobInfo,
     Membership,
     DaVerifiedBlobInfo,
-    DaVerifierBackend,
+    DaVerifierBackend, // TODO:
```
danielSanchezQ: Letfover?

nodes/nomos-node/node/src/api/backend.rs
```diff
@@ -383,6 +385,40 @@ where
...
+                _,
+                SIZE,
+            >,
+            DaVerifier<_, _, _, _, _, _>,
+            DaIndexer<_, _, _, _, _, _, _, _, _, _, _, _, _, _, SIZE>,
+            nomos_da_network_service::NetworkService<_, _>,
+            nomos_network::NetworkService<_, _>,
+            DaStorageService<_, _>,
+            TxMempoolService<_, _, _>,
+            DaMempoolService<_, _, _, _, _, _, _, _, _, _, _>
```
danielSanchezQ, line 416: Holy mother of generics....

nodes/nomos-node/node/src/api/handlers.rs
```diff
@@ -593,6 +593,7 @@ where
         AsServiceId<StorageService<RocksBackend<S>, RuntimeServiceId>> + Debug + Sync + Display,
 {
     let relay = match handle.relay().await {
+        //
```
danielSanchezQ: what?

nomos-services/api/src/lib.rs
```diff
@@ -78,17 +83,24 @@ where
...
         tracing::info!(
             "Service '{}' is ready.",
             <RuntimeServiceId as AsServiceId<Self>>::SERVICE_ID
         );
 
+        endpoint
+            .wait_until_ready(self.service_resources_handle.overwatch_handle.clone())
+            .await?;
+
+        let endpoint = endpoint;
```
danielSanchezQ: wtf

nomos-services/blend/src/lib.rs
```diff
@@ -170,6 +171,13 @@ where
             <RuntimeServiceId as AsServiceId<Self>>::SERVICE_ID
         );
 
+        wait_until_services_are_ready!(
+            &self.service_resources_handle.overwatch_handle,
+            Some(Duration::from_millis(3000)),
+            NetworkService<_, _>
+        )
+        .await?;
```
danielSanchezQ: I would say anything that depends on network should just wait. Maybe lets be pessimistic and wait `10s`?

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -656,6 +658,19 @@ where
...
+            Some(Duration::from_millis(3000)),
+            NetworkService<_, _>,
+            BlendService<_, _, _>,
+            TxMempoolService<_, _, _>,
+            DaMempoolService<_, _, _, _, _, _, _, _, _, _, _>,
+            DaSamplingService<_, _, _, _, _, _, _, _, _>,
+            StorageService<_, _>,
+            TimeService<_, _>
+        )
+        .await?;
```
danielSanchezQ: I would separate network related services from others. Always waiting for network ones first. Then the rest of them. Just because network is actually crucial for any service that depend on it.

danielSanchezQ (commented):
Overall looks nice and simple. We just need to think through a few of the comments. 
In general i'm not sure we want to fuse waiting for services. Maybe something long, network services may take time to do what needs to be done.
Wdyt on just waiting for a long time. Give it a minute for everything to break down apart in the worst case?

nomos-services/api/src/lib.rs
strinnityk: To be doubly clear that mutability's no longer needed. Do you reckon it's unnecessary?

danielSanchezQ: I think it is not necessary to reassing yes, even if you do not need it mutably anymore (and there are not other underlaying problems around it).

nomos-services/blend/src/lib.rs
strinnityk: Increased to 60s, as talked in Discord.

nomos-services/cryptarchia-consensus/src/lib.rs
strinnityk: Is it because you want to specify a different timeout for those? Otherwise I think it matters not. 
As long as one fails, the service won't startup. Which applies the same for `Network` as for any other service. It doesn't really matter if you split the calls.

danielSanchezQ: Yeah, I was thinking about using different timings. But if we go for the simple long wait it should be ok to keep then together.

danielSanchezQ (approved):
I think we can go ahead with this for now. If improvements or modifications are needed can be included later on. Let's see that green CI!

PR #1346, opened 2025-06-10
feat: Cryptarchia bootsrapping networking

andrussal:

Cryptarchia Bootstrapping networking according to [Bootsrapping spec](https://www.notion.so/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6).

Note: This PR doesn't include logic to retrieve blocks from database.

(This is followup PR of [PR](https://github.com/logos-co/nomos/pull/1312))

@andrussal 

Yes

N/A

## Checklist

* [x] 1. Description added.


consensus/cryptarchia-sync/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+mod behaviour;
+mod downloader;
+mod errors;
+mod messages;
+mod provider;
+mod utils;
+
+// TODO: move to nomos-core
```
danielSanchezQ: Probably not. In nomos core only `wire` lives in. All specific can be here.

consensus/cryptarchia-sync/src/messages.rs
```diff
@@ -0,0 +1,70 @@
+use bytes::Bytes;
+use nomos_core::header::HeaderId;
+use serde::{Deserialize, Serialize};
+
+/// Blocks are serialized using nomos-core's packing format.
+pub type SerialisedBlock = Bytes;
+
+/// `HeaderId` is serialized using nomos-core's packing format.
+pub type SerialisedHeaderId = Bytes;
```
danielSanchezQ: There is no packing in nomos-core, just wire (?)

consensus/cryptarchia-sync/src/provider.rs
```diff
@@ -0,0 +1,78 @@
...
+    ) -> Result<RequestStream, ChainSyncError> {
+        let request: RequestMessage = unpack_from_reader(&mut stream)
+            .await
+            .map_err(|e| ChainSyncError::from((peer_id, e)))?;
+
+        Ok(RequestStream::new(peer_id, stream, request))
+    }
+
+    pub async fn provide_tip(
+        reply_rcv: oneshot::Receiver<BoxStream<'static, Result<Bytes, ChainSyncError>>>,
```
danielSanchezQ: If it is just a tip, why do we use a stream here? Is it possible to provide more than a single tip?
Also, why is the stream using bytes already instead of `BlockId/HeaderId`, who is responsible of transforming it to bytes?

consensus/cryptarchia-sync/src/provider.rs
```diff
@@ -0,0 +1,78 @@
...
+            ),
+        })?;
+
+        stream
+            .try_fold(&mut libp2p_stream, |stream, block| async move {
+                let message = DownloadBlocksResponse::Block(block);
+                send_message(peer_id, stream, &message).await?;
+                Ok(stream)
+            })
+            .await?;
```
danielSanchezQ, line 70: Again, I think this is not the best approach. Ideally you take those bytes, zip or flat_map them to include the message size, then write all those bytes into the stream just by piping the data from the incoming block stream into the `Sink` part of the outgoing stream.

consensus/cryptarchia-sync/src/downloader.rs
```diff
@@ -0,0 +1,133 @@
...
+    ChainSyncErrorKind,
+};
+
+pub struct DownloadBlocksTask;
+
+impl DownloadBlocksTask {
+    pub async fn send_tip_request(
+        peer_id: PeerId,
+        control: &mut Control,
+        reply_sender: Sender<BoxStream<'static, Result<ResponseKind, ChainSyncError>>>,
```
danielSanchezQ: Why `ResponseKind`? If its a tip it should return a TipResponse, why a mixed stream of responses?

consensus/cryptarchia-sync/src/downloader.rs
```diff
@@ -0,0 +1,133 @@
...
+
+        let request_stream = RequestStream::new(peer_id, stream, tip_request);
+        Ok((request_stream, reply_sender))
+    }
+
+    pub async fn send_download_request(
+        peer_id: PeerId,
+        mut control: Control,
+        request: DownloadBlocksRequest,
+        reply_sender: Sender<BoxStream<'static, Result<ResponseKind, ChainSyncError>>>,
```
danielSanchezQ: Same here, why not `BlockResponse`. Who then unpacks this? the consumer service is probably expecting `Blocks`

consensus/cryptarchia-sync/src/downloader.rs
```diff
@@ -0,0 +1,133 @@
...
+                        stream::iter(vec![Ok(ResponseKind::Tip(TipResponse::NetworkError(e)))])
+                            .boxed()
+                    }
+                };
+
+            reply_tx.send(tip_stream).await.map_err(|e| ChainSyncError {
+                peer: peer_id,
+                kind: ChainSyncErrorKind::ChannelSendError(format!(
+                    "Failed to send tip stream: {e}",
+                )),
```
danielSanchezQ: I'm sorry if I misslead you in previous reviews, but if we expect a single tip per request then it should not be a stream.

consensus/cryptarchia-sync/src/downloader.rs
```diff
@@ -0,0 +1,133 @@
...
+                },
+            ));
+
+            reply_tx.send(stream).await.map_err(|e| ChainSyncError {
+                peer: peer_id,
+                kind: ChainSyncErrorKind::ChannelSendError(format!(
+                    "Failed to send blocks stream: {e}"
+                )),
+            })
+        })
```
danielSanchezQ: This is beautiful 💟

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,744 @@
...
+        result: TipResponse,
+    },
+}
+
+#[derive(Debug)]
+pub enum ResponseKind {
+    Blocks(BlocksResponse),
+    Tip(TipResponse),
+    NetworkError(ChainSyncError),
+}
```
danielSanchezQ: I think this is not necessary at all.

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,744 @@
...
+        self.sending_requests.push(
+            DownloadBlocksTask::send_download_request(peer_id, control, request, reply_sender)
+                .boxed(),
+        );
+
+        Ok(())
+    }
+
+    fn handle_tip_request(&self, peer_id: PeerId, stream: Libp2pStream) -> Poll<ToSwarmEvent> {
+        let (reply_sender, reply_rcv) = oneshot::channel();
```
danielSanchezQ: I would prefer explicit names as we (try to) have in the codebase.
```suggestion
        let (reply_sender, reply_receiver) = oneshot::channel();
```

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,744 @@
...
+    }
+
+    fn on_connection_handler_event(
+        &mut self,
+        peer_id: PeerId,
+        conn_id: ConnectionId,
+        event: <Self::ConnectionHandler as ConnectionHandler>::ToBehaviour,
+    ) {
+        self.stream_behaviour
+            .on_connection_handler_event(peer_id, conn_id, event);
```
danielSanchezQ, line 458: Using an underlaying stream behaviour has one downside. If you try to open a stream to a peer_id you are not connected to then it generates a dial up even that need to be bubbled up to the upper swarm. I don't remember where this has to be done, but you can check in da sampling probably. Maybe this has been already handled by you but I didn't realize where.

danielSanchezQ (changes requested):
Looking much better but there are 2 main issues:
* The `ResponseKind` type is probably not needed andI think it mix things up. Or I do not understand properly its purpose.
* The tip stuff probably shouldn't use streams as we handle just one item at a time.

consensus/cryptarchia-sync/src/provider.rs
andrussal: > Also, why is the stream using bytes already instead of BlockId/HeaderId, who is responsible of transforming it to bytes?

I tried to avoid adding generics to messages, like `Block<Tx, Blob>` because then they will be part of swarm itself. And used also Bytes for `HeaderId` to have consistent approach. In existing code NetworkAdapter is doing serialisation and I figured here it could do as well. Please let me know if should push serialisation from NetworkAdapter to networking core layer.

consensus/cryptarchia-sync/src/downloader.rs
andrussal: > I'm sorry if I misslead you in previous reviews, but if we expect a single tip per request then it should not be a stream.

Doing both with streams helped me to do fast refactoring because both(tip and block logic) use the same types. I hoped to get quick feedback if this approach is now as expected. I will now iterate to use just a channel to get single `HeaderId`.

consensus/cryptarchia-sync/src/lib.rs
andrussal: Understand, lets keep it here.

consensus/cryptarchia-sync/src/messages.rs
andrussal: I had `packing.rs` previously in `nomos-core`. I copied it from `DA` and put to `nomos-core` so it can be shared. But now I see that we should keep it here. I will update comments accordingly.

consensus/cryptarchia-sync/src/downloader.rs
andrussal: Removed `ResponseKind`.

consensus/cryptarchia-sync/src/provider.rs
danielSanchezQ: > > Also, why is the stream using bytes already instead of BlockId/HeaderId, who is responsible of transforming it to bytes?
> 
> I tried to avoid adding generics to messages, like `Block<Tx, Blob>` because then they will be part of swarm itself. And used also Bytes for `HeaderId` to have consistent approach. In existing code NetworkAdapter is doing serialisation and I figured here it could do as well. Please let me know if should push serialisation from NetworkAdapter to networking core layer.

Ok, it makes sense this understand only about bytes then and the adapter is the bride for serialization, thanks.

consensus/cryptarchia-sync/src/provider.rs
danielSanchezQ: I see that libp2p**stream** doesn't actually implement `Stream`...shame on them. We can keep this solution, is good. Thanks.

consensus/cryptarchia-sync/src/behaviour.rs
andrussal: Thank you for reminding this. It handles now libp2p stream Dial events.

consensus/cryptarchia-sync/src/messages.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// A request to download blocks.
+    DownloadBlocksRequest(DownloadBlocksRequest),
+    /// A request to get the tip of the peer.
+    GetTip,
+}
+
+/// A request to initiate block downloading from a peer.
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct DownloadBlocksRequest {
+    /// Return blocks up to `target_block` if specified.
```
ntn-x2:
```suggestion
    /// Return blocks up to `target_block`.
```

consensus/cryptarchia-sync/src/messages.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// A set of block identifiers the syncing peer already knows.
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct KnownBlocks {
+    /// The local canonical chain latest block.
+    pub local_tip: HeaderId,
+    /// The latest immutable block.
+    pub latest_immutable_block: HeaderId,
+    /// The list of additional blocks that the requester has.
+    pub additional_blocks: Vec<HeaderId>,
```
ntn-x2: Does order matter? If not, should we switch to a `HashSet` here as well?

consensus/cryptarchia-sync/src/downloader.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    pub async fn receive_tip(request_stream: RequestStream) -> Result<(), ChainSyncError> {
+        let RequestStream {
+            mut stream,
+            peer_id,
+            reply_channel,
+            ..
+        } = request_stream;
+
+        let ReplyChannel::Tip(reply_tx) = reply_channel else {
```
ntn-x2: Why can't we enforce this at compilation time instead? We could take a value that is specific to each `RequestStream` variant, for instance, and if we make sure `RequestStream` can only be constructed by calling a specific method, we would solve the issue here.

danielSanchezQ: Yes I agree. Why don't we have different messages for each of the things that we need to handle instead of a generic message for everything?

andrussal: If it seems better to have different set of `FuturesUnordered` for tip and block processing, I will split them into 2. But technically it's required only for this type to solve the compilation issue.

```rust
type SendingRequestsFuture = BoxFuture<'static, Result<RequestStream, ChainSyncError>>;
...
sending_requests: FuturesUnordered<SendingRequestsFuture>,
```

danielSanchezQ: > If it seems better to have different set of `FuturesUnordered` for tip and block processing, I will split them into 2. But technically it's required only for this type to solve the compilation issue.
> 
> ```rust
> type SendingRequestsFuture = BoxFuture<'static, Result<RequestStream, ChainSyncError>>;
> ...
> sending_requests: FuturesUnordered<SendingRequestsFuture>,
> ```

I think its fair to have two sets of futures to be resolved.

andrussal: Got it! Will split into two then.

andrussal:
> Looking much better but there are 2 main issues:
> 
> * The `ResponseKind` type is probably not needed andI think it mix things up. Or I do not understand properly its purpose.
> * The tip stuff probably shouldn't use streams as we handle just one item at a time.

@danielSanchezQ These are solved now. Please take a look again.

danielSanchezQ (approved):
It was a long trip but here we are. Thanks @andrussal

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,803 @@
...
+
+type ReceivingBlocksResponsesFuture = BoxFuture<'static, Result<(), ChainSyncError>>;
+
+type ReceivingTipsResponsesFuture = BoxFuture<'static, Result<(), ChainSyncError>>;
+
+type SendingBlocksResponsesFuture = BoxFuture<'static, Result<(), ChainSyncError>>;
+
+type SendingTipsResponsesFuture = BoxFuture<'static, Result<(), ChainSyncError>>;
+
+type ReceivingRequestsFuture = BoxFuture<'static, Result<ReceivingRequestStream, ChainSyncError>>;
```
youngjoon-lee:
```suggestion
type SendingBlocksRequestsFuture = BoxFuture<'static, Result<BlocksRequestStream, ChainSyncError>>;

type SendingTipRequestsFuture = BoxFuture<'static, Result<TipRequestStream, ChainSyncError>>;

type ReceivingBlocksResponsesFuture = BoxFuture<'static, Result<(), ChainSyncError>>;

type ReceivingTipResponsesFuture = BoxFuture<'static, Result<(), ChainSyncError>>;

type SendingBlocksResponsesFuture = BoxFuture<'static, Result<(), ChainSyncError>>;

type SendingTipResponsesFuture = BoxFuture<'static, Result<(), ChainSyncError>>;

type ReceivingRequestsFuture = BoxFuture<'static, Result<ReceivingRequestStream, ChainSyncError>>;
```
to have the same naming convention.

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,803 @@
...
+        reply_sender: Sender<BoxStream<'static, Result<SerialisedBlock, ChainSyncError>>>,
+    ) -> Result<(), ChainSyncError> {
+        if !self.connected_peers.contains(&peer_id) && !self.initial_peers.contains_key(&peer_id) {
+            return Err(ChainSyncError {
+                peer: peer_id,
+                kind: ChainSyncErrorKind::StartSyncError(
+                    "Peer is neither connected nor known".to_owned(),
+                ),
+            });
+        }
```
youngjoon-lee:
```suggestion
        if !self.connected_peers.contains(&peer_id) {
            return Err(ChainSyncError {
                peer: peer_id,
                kind: ChainSyncErrorKind::StartSyncError(
                    "Peer is neither connected nor known".to_owned(),
                ),
            });
        }
```
I don't think we should check if the provided `peer_id` is not in the `self.initial_peers`. Actually, this `Behaviour` doesn't need to keep the `self.initial_peers`. The user of this `Behaviour` should have a freedom to choose a peer for downloading blocks, as long as the peer is connected.

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,803 @@
...
+        }))
+    }
+
+    fn handle_incoming_stream(&self, peer_id: PeerId, mut stream: Libp2pStream) {
+        let concurrent_requests = self.receiving_requests.len()
+            + self.sending_block_responses.len()
+            + self.sending_tip_responses.len();
+
+        if concurrent_requests >= MAX_INCOMING_REQUESTS {
+            self.incoming_streams_to_close.push(
```
ntn-x2, line 314: Why can't we close it from here directly? We have `mut` access to it.

consensus/cryptarchia-sync/src/packing.rs
```diff
@@ -0,0 +1,69 @@
...
+    let length_prefix: LenType =
+        packed_message
+            .len()
+            .try_into()
+            .map_err(|_| PackingError::MessageTooLarge {
+                max: MAX_MSG_LEN,
+                actual: packed_message.len(),
+            })?;
+
+    // Note: using vectorised io doesn't work
```
youngjoon-lee: thank you for testing the vertorised IO. you can remove this comment :) other people may be confused what this comment means.

consensus/cryptarchia-sync/src/behaviour.rs
andrussal: I see your point.  Actually the only reason I added `initial peers` was to handle `libp2p stream` `Dial` event in a way that seemed useful. It should only want to `Dial` a peer if peer is not connected. And when we try to open a stream to a peer that belongs to `connected` peers, it should never trigger `Dial` event(from `libp2p stream`).  

Also some time ago we  talked that we should add `initial peers` to consensus service config. I know that we dial all initial peers when node starts. But perhaps we should also consider the possibility that connection gets dropped? In that case `libp2p stream` could try to connect again.

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,803 @@
...
+                }
+            }
+
+            return Poll::Pending;
+        }
+
+        if let Poll::Ready(Some(result)) = self.sending_tip_requests.poll_next_unpin(cx) {
+            match result {
+                Ok(request_stream) => {
+                    self.handle_tip_request_available(cx, request_stream);
```
youngjoon-lee: We're passing the `cx` to this function for waking it inside the function. But, all other functions (such as `handle_blocks_request_available`) don't accept the `cx`. Instead, they call the waker stored in the `Behaviour::waker`. I think it's better to use a consistent way across all related funcs.

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,803 @@
...
+            + self.sending_tip_responses.len();
+
+        if concurrent_requests >= MAX_INCOMING_REQUESTS {
+            self.incoming_streams_to_close.push(
+                async move {
+                    let _ = stream.close().await;
+                }
+                .boxed(),
+            );
+            error!("Rejected excess pending incoming request");
```
youngjoon-lee, line 320: I think it's safer to call `self.try_notify_waker()` here as well,  like the `else` statement, because you're adding a future to the set and exiting from this function. The actual behaviour will depend on how this function is used in the `NetworkBehavior::poll`, but I think ti's better to have a consistency inside this function.
Or, you can call `self.try_notify_waker()` at the end of this func.

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -0,0 +1,803 @@
...
+                Err(e) => {
+                    error!("Error while processing incoming request: {}", e);
+                }
+            }
+        }
+
+        if let Poll::Ready(Some((peer_id, stream))) = self.incoming_streams.poll_next_unpin(cx) {
+            self.handle_incoming_stream(peer_id, stream);
+        }
+
```
youngjoon-lee: It would be helpful if you leave a code comment why this branch doesn't return `Poll::Pending` unlike other branches above. If it's trivial, please just leave a reply in thread instead.

ntn-x2 (approved):

consensus/cryptarchia-sync/src/behaviour.rs
youngjoon-lee: We should have the `initial_peer` config in the service layer, yes. But, the service will call `swarm.dial` or something similar. Then, your `NetworkBehaviour` will be able to get events when connections are established. Then, you can just keep those peers in the `connected_peers`. 

What the consensus service will do to initialize IBD is just:
```rust
swarm.cryptarhica_sync().connected_peed().iter().for_each(|peer_id| {
  swarm.cryptarchia_sync().start_blocks_download(peer_id, ...);
});
```
I think this is the simplest approach. 

>  Actually the only reason I added initial peers was to handle libp2p stream Dial event in a way that seemed useful. It should only want to Dial a peer if peer is not connected. And when we try to open a stream to a peer that belongs to connected peers, it should never trigger Dial event(from libp2p stream).

I don't understand this part. Could you please give me more context?

youngjoon-lee: > I don't understand this part. Could you please give me more context?

https://github.com/logos-co/nomos/blob/605a7e29084e594a02480f4e50dce308c5408de6/consensus/cryptarchia-sync/src/behaviour.rs#L538

Okay. I just saw this code. I think this is what you're mentioning about. I don't think this code is really beneficial. The swarm user (service layer) should specify enough information (addresses) when dialing. I think this `Behaviour` doesn't need to enrich the `DialOpt` that was already set by the swarm user. Actually the `initial_peers` information itself was specified by the swarm user. That is, the swarm user can enrich `DialOpt`s by himself if necesary. Am I missing something?

youngjoon-lee: > Also some time ago we talked that we should add initial peers to consensus service config. I know that we dial all initial peers when node starts. But perhaps we should also consider the possibility that connection gets dropped? In that case libp2p stream could try to connect again.

Actually, a node doesn't really need to download blocks from **all** initial peers. Some time ago, I told you that the node will initiate IBD with the initial peers, but only because that's what the node knows when start-up. If some of those peers are offline, then the node can simply skip them in IBD. The `NetworkBehaviour` doesn't have anything special to do with those initial peers.

andrussal: > I don't understand this part. Could you please give me more context?

I added underlying libp2p stream `Dial` event handling to the Behaviour and was concerned how we attach addresses to related `DialOpts`.  And these `Dial` events should be only triggered if node is not connected. And it seems that node is not connected only if we try to download blocks from a initial peer that is not connected(in case connection gets dropped).

youngjoon-lee: > I added underlying libp2p stream `Dial` event handling to the Behaviour and was concerned how we attach addresses to related `DialOpts`. And these `Dial` events should be only triggered if node is not connected. And it seems that node is not connected only if we try to download blocks from a initial peer that is not connected(in case connection gets dropped).

Ah, okay, thanks. I just read https://github.com/logos-co/nomos/pull/1346#discussion_r2139329553. I still think the Behaviour doesn't need to handle initial peers specially.

andrussal: > Ah, okay, thanks. I just read https://github.com/logos-co/nomos/pull/1346#discussion_r2139329553. I still think the Behaviour doesn't need to handle initial peers specially.

In this case we probably don't need to handle libp2p stream `Dial` events and could remove it altogether? We could keep around connected peer's addresses but this doesn't seem useful.

youngjoon-lee: > > Ah, okay, thanks. I just read [#1346 (comment)](https://github.com/logos-co/nomos/pull/1346#discussion_r2139329553). I still think the Behaviour doesn't need to handle initial peers specially.
> 
> In this case we probably don't need to handle libp2p stream `Dial` events and could remove it altogether? We could keep around connected peer's addresses but this doesn't seem useful.

I think we don't need to handle Dial from stream_behaviour. We can just forward whatever event returned from stream_bheaviour.poll() to the caller.

danielSanchezQ: > > > Ah, okay, thanks. I just read [#1346 (comment)](https://github.com/logos-co/nomos/pull/1346#discussion_r2139329553). I still think the Behaviour doesn't need to handle initial peers specially.
> > 
> > 
> > In this case we probably don't need to handle libp2p stream `Dial` events and could remove it altogether? We could keep around connected peer's addresses but this doesn't seem useful.
> 
> I think we don't need to handle Dial from stream_behaviour. We can just forward whatever event returned from stream_bheaviour.poll() to the caller.

:heavy_plus_sign: 1, This is exactly what we need to do indeed.

andrussal: I just tried but it seems that we can't handle it generic way.  `poll` returns `pub enum ToSwarm<TOutEvent, TInEvent>` type.  Here `TOutEvent` for sync behaviour is our custom `Event` and libp2p stream has it just `()`. It seems that we need to handle `Dial` directly.

andrussal: As we discussed in Discord, we can just leave it to the swarm to handle addresses I think and keep it simple though.

consensus/cryptarchia-sync/src/behaviour.rs
andrussal: This was perviously last statement in poll. Now it is before `Dial`. I added return here now as well.

consensus/cryptarchia-sync/src/behaviour.rs
andrussal: I think we can't call it directly here because it's async function.

consensus/cryptarchia-sync/src/behaviour.rs
youngjoon-lee: > I just tried but it seems that we can't handle it generic way. `poll` returns `pub enum ToSwarm<TOutEvent, TInEvent>` type. Here `TOutEvent` for sync behaviour is our custom `Event` and libp2p stream has it just `()`. It seems that we need to handle `Dial` directly.

I think the [a3ac69a](https://github.com/logos-co/nomos/pull/1346/commits/a3ac69a3030019fdee5da88c1cfa165af7e1a7d5) looks good.

youngjoon-lee: > As we discussed in Discord, we can just leave it to the swarm to handle addresses I think and keep it simple though.

Though I didn't see the discussion in Discord yet, I agree. 
This NetworkBehaviour is different from DA sampling Behaviour: https://github.com/logos-co/nomos/blob/deeefe0383f23be0fe9394b5c5d40b2f9ae8c6d8/nomos-da/network/core/src/protocols/sampling/behaviour.rs#L821. 
DA sampling has the membership which is trustable and scoped. However, here we don't have anything. Our behaviour can just toss the Dial event to the swarm, so that the swarm user can do whatever he can.

danielSanchezQ: > > As we discussed in Discord, we can just leave it to the swarm to handle addresses I think and keep it simple though.
> 
> Though I didn't see the discussion in Discord yet, I agree. This NetworkBehaviour is different from DA sampling Behaviour:
> 
> https://github.com/logos-co/nomos/blob/deeefe0383f23be0fe9394b5c5d40b2f9ae8c6d8/nomos-da/network/core/src/protocols/sampling/behaviour.rs#L821
> .
> DA sampling has the membership which is trustable and scoped. However, here we don't have anything. Our behaviour can just toss the Dial event to the swarm, so that the swarm user can do whatever he can.

Pasting here from the discussion:
```markdown
But I think they are 2 different problems:
* Dials from underlaying behaviour should be handled if you are selecting a peer that you trust and you were connected at some point. That should be fine imo.
* How to select which peer to do the sync. This is different as the behaviour will not try to dial a peer which we haven't try to open a stream with in the first place.

TLDR: I think we should keep this event bubbling up. But we need to handle properly which peers to sync from.
```

PR #1350, opened 2025-06-11
refactor(Recovery): Simplify Recovery

zeegomo:

Cryptarchia engine does not need information past the LIB and we can unify genesis and LIB handling.

This removes:
* `unchecked` function variants in cryptarchia engine and consensus
* recovery strategy case work

A change that could influence other nodes is that we don't store anymore the full history in consensus engine. This is on purpose as it's not needed there and we can offload it to colder storage
It's a simplification of the previous procedure
N/A
N/A
N/A
## Checklist

> [!WARNING]  

* [x] 1. Description added.
* [ x] 5. Link PR to a specific milestone.


danielSanchezQ:
> > So i guess this means that we treat genesis as the lib until there an actual lib right? So we do not need to handle any extra situation (?)
> 
> The genesis has exactly the same properties as the Latest Immutable Block, it **is** in fact the LIB for an _empty_ chain. The only difference is that it's the first block, but this information is not really used anywhere in consensus at least

Yeah, nice. Thanks for this!

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -894,8 +854,10 @@ where
             ConsensusMsg::GetHeaders { from, to, tx } => {
                 // default to tip block if not present
                 let from = from.unwrap_or_else(|| cryptarchia.tip());
-                // default to genesis block if not present
-                let to = to.unwrap_or_else(|| cryptarchia.genesis());
+                // default to LIB block if not present
+                // TODO: for a full history, we should use genesis, but we don't want to
+                // keep it all in memory, headers past LIB should be fetched from storage
+                let to = to.unwrap_or_else(|| cryptarchia.lib());
```
youngjoon-lee, line 855: Question:

This comment and the PR description say that the cryptarchia-engine doesn't keep the full history. That probably says that the engine now keeps only history from LIB.
> A change that could influence other nodes is that we don't store anymore the full history in consensus engine. This is on purpose as it's not needed there and we can offload it to colder storage

I agree with this approach because we don't want to keep it all in memory. 

However, it seems that the cryptarchia-engine still hold all blocks from the genesis in memory because the [prune_forks](https://github.com/logos-co/nomos/blob/dc8a216e193e3d586d445af4a22f6f92e74da551/nomos-services/cryptarchia-consensus/src/lib.rs#L1276) function removes blocks only in the forks, not in the honest chain. 

Nevertheless, I see your point in this code (which handles the `GetHeaders` message) because now the cryptarchia-engine doesn't provide the `genesis()` function.

I'd like to ask whether I'm missing something from this PR, regarding removing blocks older than LIB. Or, do you think we can implement the block offloading later? In my opinion, it would be simpler/better for us to implement it as a next step, if it's not implemented in this PR.
Anyway, I will keep this change (or future changes) in mind when I implement the bootstrapping (the responder side).

zeegomo: You're right that we don't remove stuff before the LIB at the moment. My idea is that we probably will/should at some point, as they're not needed for the engine. 
We will also probably need to keep the full history somewhere else for this reason, but that can be on a colder storage (i.e. not memory), and we can add it in a future PR

danielSanchezQ: > You're right that we don't remove stuff before the LIB at the moment. My idea is that we probably will/should at some point, as they're not needed for the engine. We will also probably need to keep the full history somewhere else for this reason, but that can be on a colder storage (i.e. not memory), and we can add it in a future PR

We already store every block in storage. We could safely remove anything older than the LIB. Let's open an issue with some directions and maybe @strinnityk can take on that? wdy guys think?

youngjoon-lee: > We already store every block in storage. We could safely remove anything older than the LIB

True. 

> Let's open an issue with some directions and maybe @strinnityk can take on that? wdy guys think?

I think this task consists of two points:
1. Remove blocks older than LIB from memory
2. Have a logic to load a block from storage when the block is needed but not in memory.

Of course, we can split these into two PRs, but ultimately we need both for supporting bootstrapping or any others in the future. If Alex can take on this, I'd appreciate it. Or, I can also do this.

I'd like to hear @zeegomo's opinion before opnening an issue in github.

strinnityk (approved):
Great stuff! 🔥

nomos-services/cryptarchia-consensus/src/lib.rs
zeegomo: lgtm

zeegomo: A good test for this could be to reliably fix the two nodes happy test, by requesting the full history and verifying one is a prefix of the other up to the LIB

Issue #1288, opened 2025-05-25
Update cover message scheduler


youngjoon-lee:
The new Blend v1 spec introduces the new [Cover Message Schedule](https://www.notion.so/Blend-Protocol-Version-1-1c48f96fb65c809494efe63019a5ebfb?pvs=4#1ca8f96fb65c80fdb4fcd8f068f8e636) that has been substantially updated. 

The new mechanism is based on the following new concepts. We may need to implement these concept first before updating the cover message scheduler.
- [Session](https://www.notion.so/Blend-Protocol-Version-1-1c48f96fb65c809494efe63019a5ebfb?pvs=4#1d08f96fb65c80948d39d98e6924a5db)
- [Interval](https://www.notion.so/Blend-Protocol-Version-1-1c48f96fb65c809494efe63019a5ebfb?pvs=4#1d08f96fb65c80e189e7c1a78fde67e6)
- [Round](https://www.notion.so/Blend-Protocol-Version-1-1c48f96fb65c809494efe63019a5ebfb?pvs=4#1d08f96fb65c80d1bab0f81ed71b3219)

Also, the new cover message schedule includes the concept of [Core Quota](https://www.notion.so/Blend-Protocol-Version-1-1c48f96fb65c809494efe63019a5ebfb?pvs=4#1c88f96fb65c80c3a69cf95562388ae1), which can be fully implemented when our [SDP](https://www.notion.so/Service-Declaration-Protocol-Specification-1fd261aa09df819ca9f8eb2bdfd4ec1d?pvs=4) implementation is ready. For now, we can treat this as a constant since our current implementation works with a static set of core nodes.

PR #1351, opened 2025-06-12
feat: update cover message scheduler based on new spec

ntn-x2:

I updated the cover message schedule to be compliant with the [latest version of the spec](https://www.notion.so/Blend-Protocol-Version-1-PENDING-MIGRATION-1c48f96fb65c809494efe63019a5ebfb#1ca8f96fb65c80fdb4fcd8f068f8e636). So this PR fixes https://github.com/logos-co/nomos/issues/1288.

I decided to keep the changeset as limited as possible while still implementing all the needed changes. Talks about restructuring the whole structure of the blend stack are ongoing and will be tackled in a different set of PRs.

The interface between the service and the cover message module has remained unchanged, with the module implementing the `Stream` trait. The only addition is that whenever the service generates a new data message, it has to communicate it to the cover module so that it can skip generating a new cover message for the next scheduled slot.

I also added some unit tests for the message scheduling logic, and because I made the safety buffer a constant instead of a config element, we have to wait at least 100 intervals before each test case completes. I am open to change the constant to a settings element, if needed, and will use a smaller value for the unit tests to make them run faster.

Yes.

@ntn-x2 dev, @madxor spec writer, @youngjoon-lee code owner.

Yes.

No.

## Checklist

* [x] 1. Description added.


nomos-blend/core/src/membership.rs
```diff
@@ -9,33 +9,34 @@ use rand::{
...
 }
 
-impl<NodeId, M> Membership<NodeId, M>
+impl<NodeId, Message> Membership<NodeId, Message>
 where
-    NodeId: Hash + Eq,
-    M: BlendMessage,
-    M::PublicKey: PartialEq,
+    Message: BlendMessage<PublicKey: PartialEq>,
 {
```
ntn-x2, line 35: I split this impl block into three separate ones, each with its own constraints. It did not make sense to require all the trait bounds to call `.size()` on the membership type.

ntn-x2:
After a follow-up discussion in https://discord.com/channels/1111286067413405788/1382272832616140840, I added a temporary workaround in https://github.com/logos-co/nomos/pull/1351/commits/7564cd5021ddb64c14dfef462f74674dc8b910d0 to differentiate between locally-generated messages and forwarded messages that go through the temporal processor. With the upcoming re-design, all of this logic will be encapsulated into a new scheduler, I think, but for now this is needed since the temporal scheduler makes no difference between the two types of messages.

nomos-services/blend/src/lib.rs
```diff
@@ -178,10 +184,20 @@ where
...
+        // latest v1 of the spec.
+        let mut scheduled_local_messages = HashSet::new();
         loop {
             tokio::select! {
                 Some(msg) = persistent_transmission_messages.next() => {
+                    let is_local_message = scheduled_local_messages.remove(&msg);
                     backend.publish(msg).await;
+                    if is_local_message {
+                        cover_traffic.notify_of_new_data_message().await;
+                    }
```
youngjoon-lee, line 200: (Replying to https://github.com/logos-co/nomos/pull/1351#issuecomment-2966409905)

Thank you. I think this is fine for now.

nomos-services/blend/src/lib.rs
```diff
@@ -209,11 +225,16 @@ where
...
-                    Self::wrap_and_send_to_persistent_transmission(&msg, &mut cryptographic_processor, &persistent_sender);
+                    let wrapped_message = cryptographic_processor.wrap_message(&msg)?;
+                    backend.publish(wrapped_message).await;
                 }
                 Some(msg) = local_messages.next() => {
-                    Self::wrap_and_send_to_persistent_transmission(&msg, &mut cryptographic_processor, &persistent_sender);
+                    let Some(wrapped_message) = Self::wrap_and_send_to_persistent_transmission(&msg, &mut cryptographic_processor, &persistent_sender) else {
+                        continue;
+                    };
+                    scheduled_local_messages.insert(wrapped_message);
```
youngjoon-lee, line 237: Good. I think this implements "releasing a locally generated data message at the next round and skipping one next cover message" very simply, because we can configure the persistent transmission to emit messages at the same interval as rounds. Thanks. As you mentioned, we can refactor this soon to reflect the architecture of the latest spec properly.

nomos-services/blend/src/lib.rs
```diff
@@ -273,58 +296,50 @@ pub struct BlendConfig<BackendSettings, BackendNodeId> {
...
-        futures::stream::Enumerate<IntervalStream>,
-        impl FnMut((usize, time::Instant)) -> usize,
-    > {
-        let slots_per_epoch = self.slots_per_epoch();
-        IntervalStream::new(time::interval(self.slot_duration))
-            .enumerate()
-            .map(move |(i, _)| i % slots_per_epoch)
+            .map(move |(i, _)| SessionInfo {
+                session_number: (i as u64).into(),
+                membership_size,
```
youngjoon-lee, line 342: In the future, this `membership_size` should be retrieved from somewhere like the Membership service, whenever a new session number should be returned, if I understand this code correctly? Then, yes, this is good. We assume that the membership is static for now in this PR since we haven't implemented the membership updates yet.

nomos-blend/core/src/cover_traffic.rs
```diff
@@ -1,132 +1,684 @@
...
-use nomos_blend_message::BlendMessage;
-use serde::Deserialize;
+use nomos_utils::math::NonNegativeF64;
+use tokio::{sync::mpsc, time};
+use tokio_stream::wrappers::IntervalStream;
+use tracing::{debug, error, trace, warn};
 
-#[derive(Copy, Clone, Deserialize)]
+// max: safety buffer length, expressed in intervals
+const SAFETY_BUFFER_INTERVALS: u64 = 100;
```
youngjoon-lee: Yeah. As you said in the PR description, we may need to consider making this configurable. But, it's fine to have this as a constant for now, TBH, because this number was obtained from a simulation (by Alexander) and this won't change in near future.

nomos-blend/core/src/cover_traffic.rs
```diff
@@ -1,132 +1,684 @@
...
+const LOG_TARGET: &str = "blend::core::cover";
```
youngjoon-lee, line 16: Just a dumb question:
I've seen that some code in our codebase use log targets, but some doesn't. I remember that you and I discussed about this topic with Gusto in Discord few weeks ago, but if you think it's better to explicitly set log targets, I will follow this practice for other blend code. Is this for choosing more proper target strings by ourselves instead of relying on target strings automatically generated by `tracing`? Since I don't have production experiences for `tracing`, I'd like to learn the best practice.

nomos-blend/core/src/cover_traffic.rs
```diff
@@ -1,132 +1,684 @@
...
+    /// which such quota will be depleted.
+    fn from_session_info_and_settings<Rng>(
+        session_info: &SessionInfo,
+        rng: &mut Rng,
+        settings: &CoverTrafficSettings,
+    ) -> Self
+    where
+        Rng: rand::Rng,
+    {
+        // C: Expected number of cover messages that are generated during a session.
```
youngjoon-lee:
```suggestion
        // C: Expected number of cover messages that are generated by all the core nodes during a session.
```
It may be helpful for newcomers if we mention this, because this is what I was confused at the beginning.

nomos-blend/core/src/cover_traffic.rs
```diff
@@ -1,132 +1,684 @@
...
+            message_slots,
+            session_number: session_info.session_number,
+        }
+    }
+}
+
+/// As per the spec, it randomly generates rounds at which a cover message will
+/// be generated, over the whole duration of the session, including the
+/// specified safety buffer intervals.
+fn generate_message_slots<Rng>(
```
youngjoon-lee:
```suggestion
fn generate_message_rounds<Rng>(
```
What do you think about renaming this function (and related variables inside the function) like this? just to avoid confusion because now we've defined rounds independently with consensus slots. But, I don't have a strong preference. I guess you uses the word "slots" here just to refer to `(Interval, Round)`. If you think this is better for context, I'm fine with this.

nomos-blend/core/src/cover_traffic.rs
```diff
@@ -1,132 +1,684 @@
...
-    winning_probability: f64,
+pub struct RunningCoverTraffic<SessionStream, Rng> {
+    /// The session stream triggered at every session change, provided from the
+    /// outside.
+    sessions: SessionStream,
+    /// The internal stream that generates a new element at every round change.
+    rounds: Box<dyn Stream<Item = Round> + Send + Unpin>,
+    /// The internal stream that generates a new element at every interval.
+    intervals: Box<dyn Stream<Item = Interval> + Send + Unpin>,
+    /// The current interval value, used to look into the pre-generated slots.
```
youngjoon-lee:
```suggestion
    /// The current interval value, used to look into the pre-generated rounds.
```
We could also change this, as I suggested in https://github.com/logos-co/nomos/pull/1351#discussion_r2144012344.

nomos-blend/core/src/cover_traffic.rs
```diff
@@ -1,132 +1,684 @@
...
-            selected_slots: HashSet::default(),
-            _message: PhantomData,
+            session_info: InternalSessionInfo::from_session_info_and_settings(
+                first_session_info,
+                &mut rng,
+                &settings,
+            ),
+            rng,
+            // Channel is created assuming the node might generate a message for each round, for
+            // each interval, including the safety buffer. This is the absolutely worst case.
```
youngjoon-lee: I think we can move this comment into the `new_data_emission_notification_channel` function because the channel size is determined inside the function.

nomos-blend/core/src/cover_traffic.rs
```diff
@@ -1,132 +1,684 @@
...
+            on_interval_change(
+                session_info,
+                current_interval,
+                new_interval,
+                settings,
+                rounds,
+            );
+        }
+        if let Poll::Ready(Some(new_round)) = rounds.poll_next_unpin(cx) {
+            cx.waker().wake_by_ref();
```
youngjoon-lee: I think we don't probably need to call waker because the `rounds` (specifically `IntervalStream`) probably does it internally with `cx`, afaik? but I'm probably missing something..

nomos-blend/core/src/cover_traffic.rs
```diff
@@ -1,132 +1,684 @@
...
+    data_message_emission_notification_channel: &mut (mpsc::Sender<()>, mpsc::Receiver<()>),
+) where
+    Rng: rand::Rng,
+{
+    debug!(target: LOG_TARGET, "Session {} started.", new_session_info.session_number);
+    *current_session_info =
+        InternalSessionInfo::from_session_info_and_settings(new_session_info, rng, settings);
+    *intervals = intervals_stream(Duration::from_secs(
+        settings.rounds_per_interval.get() * settings.round_duration.as_secs(),
+    ));
```
youngjoon-lee, line 386: Maybe we should reset the `current_interval = Some(Interval::default())` here, as we do in the `RunningCoverTraffic::run`? 

Plus, what do you think about resetting it like below?
```rust
current_interval = intervals.next()
```
because the `intervals` will return the first item immediately?

nomos-services/blend/src/lib.rs
ntn-x2: Yes, I think it makes sense to have a unified scheduler that deals with sub-scheduling algorithms for cover and data messages, and provides a unified `Stream` implementation for the service to consume. I will probably sketch something out next week and we can discuss about the optimal design.

nomos-services/blend/src/lib.rs
ntn-x2: Yes, ideally the membership service would itself provide us with a stream that ticks with a new session info, and we can just pass that to the schedulers. If we are in the middle of a session, it does not make sense to start a new session stream, so there should rather be a single place where the scheduling happens, and that would be the SDP service I think.

nomos-blend/core/src/cover_traffic.rs
ntn-x2: Yeah but I've seen other constants being defined as settings entry as well. The downside of the current approach is that unit tests take quite a while to complete because of this. I am leaning towards making this a settings entry, and in the future we can perhaps define them as associated constants in a overarching settings trait implementation, where some fields can dynamically change, while others are constant and require a code update to be changed. But as I said, I think I'll convert this to a regular settings entry for now, to speed up unit testing.

nomos-blend/core/src/cover_traffic.rs
ntn-x2: The default used by the tracing crate is the module path. That means that if we want to capture logs for all blend-related stuff, from both the core crate and the service crate, we would need to specify two different filters. With this approach (which could also live in a separate crate that contains all the targets for the whole project), we can simply filter logs with `RUST_LOG=blend=trace,off`, for instance, to disable all logs but the blend ones. This is what I did when I ran unit tests to verify the code was working fine. @danielSanchezQ said we use spans, but spans serve a different purpose of tracing the "history" of a given action from start to finish. If we want to collect logs from people in the future in case there is a problem, I think we still need to use custom targets, but I don't have a strong opinion. I can also scrape this off.

nomos-blend/core/src/cover_traffic.rs
ntn-x2: Yeah I copy-pasted the text from the spec so that it can be copy-pasted in there to find the exact match. Maybe I can add your snippet, but at the end, so it does not interfere with pattern matching.

nomos-blend/core/src/cover_traffic.rs
ntn-x2: Mhh yeah, I think this function might deserve a better name. I would say we are not generating rounds, which is just a measure of time, but we are filling some of those rounds with scheduled messages. What do you think about `select_message_rounds` instead?

nomos-blend/core/src/cover_traffic.rs
ntn-x2: Yes, you are right. I will move it. There used not to be a function in an earlier version. Good catch!

nomos-blend/core/src/cover_traffic.rs
ntn-x2: When I was running tests, I got a weird behaviour where the round stream was not ticking anymore after a while. Adding this line solved the issue. But I can investigate that again and see if we can get rid of it.

nomos-blend/core/src/cover_traffic.rs
ntn-x2: Why not letting the `on_interval_change` handle it as is the case now? Resetting the interval stream will, as you say, yield a new item immediately, which is captured by the `on_interval_change` handler right after the `on_session_change`. I feel it better separates the concerns of what each handler should take care of.

nomos-blend/core/src/cover_traffic.rs
danielSanchezQ: > The default used by the tracing crate is the module path. That means that if we want to capture logs for all blend-related stuff, from both the core crate and the service crate, we would need to specify two different filters. With this approach (which could also live in a separate crate that contains all the targets for the whole project), we can simply filter logs with `RUST_LOG=blend=trace,off`, for instance, to disable all logs but the blend ones. This is what I did when I ran unit tests to verify the code was working fine. @danielSanchezQ said we use spans, but spans serve a different purpose of tracing the "history" of a given action from start to finish. If we want to collect logs from people in the future in case there is a problem, I think we still need to use custom targets, but I don't have a strong opinion. I can also scrape this off.

I think this approach is fine as well. We can have both spans and log targets coexisting I think.

nomos-blend/core/src/cover_traffic.rs
ntn-x2: Changed it in https://github.com/logos-co/nomos/pull/1351/commits/dc9bb64186266a03b0f0c8f84c13a97e1b16f83e. Now tests are much faster as well.

nomos-blend/core/src/cover_traffic.rs
ntn-x2: Changed in https://github.com/logos-co/nomos/pull/1351/commits/018702e3562809398f97d2db3bca828dfc1b8dfe.

nomos-blend/core/src/cover_traffic.rs
ntn-x2: Changed in https://github.com/logos-co/nomos/pull/1351/commits/c4d7e7991b88d2a79f78ae81ea91bfb3de4075da.

nomos-blend/core/src/cover_traffic.rs
ntn-x2: I can confirm that not waking the waker results in failed tests. I changed the order around so we first run `on_new_round` and then wake the waker in https://github.com/logos-co/nomos/pull/1351/commits/1094507b519e3b0ec2b018d707d08b7d0f427d0b. I guess we have to wake it since we are implementing the polling ourselves, which is separated from the polling of each stream we run.

nomos-blend/core/src/cover_traffic.rs
```diff
@@ -1,132 +1,685 @@
...
+{
+    /// Waits until the first session info is yielded by the provided stream,
+    /// after which the instance is initialized, started, and returned.
+    pub async fn wait_ready(mut self) -> RunningCoverTraffic<SessionStream, Rng> {
+        let mut next: Option<SessionInfo> = None;
+        while next.is_none() {
+            next = self.sessions.next().await;
+        }
+        let first_session_info =
+            next.expect("We just checked that is was not None. How did this happen?");
```
danielSanchezQ:
```suggestion
        let first_session_info =
            self.sessions.select_next_some().await.expect("We just checked that is was not None. How did this happen?");
```

nomos-services/blend/src/lib.rs
danielSanchezQ: > Yes, ideally the membership service would itself provide us with a stream that ticks with a new session info, and we can just pass that to the schedulers. If we are in the middle of a session, it does not make sense to start a new session stream, so there should rather be a single place where the scheduling happens, and that would be the SDP service I think.

Yes that is the idea, there will be a `Membership` service that you can subscribe to for blend information that will tick with the necessary information every configured session. Is being worked on.

danielSanchezQ (commented):
Nice stuff going on!

nomos-blend/core/src/cover_traffic.rs
ntn-x2: As we discussed internally, this would require the input stream to implement `FusedStream`, so I think we can keep the current implementation. Will mark this as resolved.

nomos-blend/core/src/cover_traffic.rs
youngjoon-lee: Thank you. Perfectly makese sense.

nomos-blend/core/src/cover_traffic.rs
youngjoon-lee: That sounds good to me!

nomos-blend/core/src/cover_traffic.rs
youngjoon-lee: That sounds reasonable because we poll the interval stream **before** polling the round stream.

Previously, I was concern about a case where:
- The `on_session_change` was called. It reset the interval stream.
- But, if the `on_new_round` is called before a new interval is yielded, it may cause unexpected situation in the future.
For now, even if this situation happens, it seems we're fine because the `on_new_round` function won't generate any cover message because `message_slots.remove` will return false.

Anyway, this situation won't happen as long as we poll the interval stream before the round stream.

ntn-x2: Yes, that is always the case. We poll new sessions, then new intervals, then new rounds.

nomos-blend/core/src/cover_traffic.rs
ntn-x2: Done in https://github.com/logos-co/nomos/pull/1351/commits/10b29e6359ad947616701df7dc69b89c8a3fb05e.

nomos-blend/core/src/cover_traffic.rs
ntn-x2: Done in https://github.com/logos-co/nomos/pull/1351/commits/5973711294342166a76d9bb6d242a18fc8ffd551.

nomos-blend/core/src/cover_traffic.rs
youngjoon-lee: Ah, I think I found the reason.

If all three streams (`sessions`, `intervals`, and `rounds`) return Poll::Ready, and if the `data_message_receiver_channel` inside `on_new_round` returns Poll::Ready, the waker may not be registered. In this case, the `on_new_round` returns Poll::Pending, the executor won't re-poll this task again because the waker wasn't registered.

So, to minimize calling waker, I think we can call waker only if `on_new_round` returns Poll::Pending. I changed code in my local and checked the tests pass.

This is my understanding for now, but I still have some missing pieces in my head. I will investigate this more, and will be back soon today or tomorrow. 

For now, I think your code will work, though it might be calling waker frequenctly more than necessary. Let me check.

ntn-x2: Sure, thanks for looking into this, but it seems that what you are suggesting is indeed the right way to go forward.

danielSanchezQ: If I may I suggest something to simplify this. I would create an enum that wraps each stream result, them map each stream to wrap the return type in each variant and merge the resulting streams into a single one which would be easy to poll in this method. wdyt?

youngjoon-lee: > Sure, thanks for looking into this, but it seems that what you are suggesting is indeed the right way to go forward.

yeah. I did a quick research. and i think what i suggested is correct. It is sufficient to call waker only when on-new-round returns Pending because of the case where no waker is registered as I descrived above. 
If Ready is returned, the executor will call our poll_next function again when our application code calls `cover_message_stream.next().await`

ntn-x2: > If I may I suggest something to simplify this. I would create an enum that wraps each stream result, them map each stream to wrap the return type in each variant and merge the resulting streams into a single one which would be easy to poll in this method. wdyt?

Sounds like a sane suggestion. I'll try to implement it.

ntn-x2: @danielSanchezQ I gave it a try, and I did not like how it turned out. I modified the current code to awake the awaker from inside the `on_new_round` function, which gives us more fine-grained control. For instance, when we end up in an invalid state (e.g., when the session is taking too long), we simply skip awaking the waker, since we have anyway no new messages to produce. Once a new session starts, then we start waking it again. I implemented that change here: https://github.com/logos-co/nomos/pull/1351/commits/db7ac4b02052ae9f75c5c165bd719633561928e6. Let me know what you guys think. Cc @youngjoon-lee

ntn-x2: No nevermind, I think we still need to awake it also when we are in an invalid state, or else the function will not be called anymore and we will not get out of the invalid state. Youngjoon is right. I rectified it in here: https://github.com/logos-co/nomos/pull/1351/commits/3c7ab991b95e50d30b9d45c619ad5c394639d6b4. Sorry for the spam, I think we are close to closing this now, thanks.

youngjoon-lee: The new code looks good to me, Thanks!

PR #1352, opened 2025-06-12
feat(membership): return block number with providers in queries

pradovic:
This tiny PR is adding block_number information to the return value of membership service queries. It will be needed by subscribers to track the latest block number. 
Also, it might be useful when we handle skipped blocks.

* [x] 1. Description added.

### DIFF
--- a/nomos-services/membership/src/backends/mock.rs
+++ b/nomos-services/membership/src/backends/mock.rs
@@ -315,57 +315,57 @@ mod tests {
 
         // (1st entry)
         // blocknumber 100 = 105 - k.historical_block_delta
-        let result = backend.get_providers_at(service_type, 105).await.unwrap().1;
+        let (_, providers) = backend.get_providers_at(service_type, 105).await.unwrap();
 
-        assert_eq!(result.len(), 1);
-        assert!(result.contains_key(&provider_info_1.provider_id));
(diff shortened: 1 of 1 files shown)


nomos-services/membership/src/backends/mock.rs
```diff
@@ -313,6 +316,7 @@ mod tests {
         // (1st entry)
         // blocknumber 100 = 105 - k.historical_block_delta
         let result = backend.get_providers_at(service_type, 105).await.unwrap();
```
ntn-x2: I guess you can add `.1` directly here instead of re-defining the `result` variable?

ntn-x2 (approved):

nomos-services/membership/src/backends/mock.rs
```diff
@@ -312,7 +315,8 @@ mod tests {
 
         // (1st entry)
         // blocknumber 100 = 105 - k.historical_block_delta
-        let result = backend.get_providers_at(service_type, 105).await.unwrap();
+        let result = backend.get_providers_at(service_type, 105).await.unwrap().1;
```
andrussal: Maybe easier to read

```rust
let (_, locators) =  backend.get_providers_at(service_type, 105).await.unwrap();
```

PR #1341, opened 2025-06-09
 Mock memberhip adapter in DA

pradovic:
## 1. What does this PR implement?

> [!NOTE]  
> Review  
>
the [Specification to Implementation Process document](https://www.notion.so/Specification-to-Implementation-Process-0cd6547233fa43a1a03d6e360de9eb2b).

## 2. Does the code have enough context to be clearly understood?

> [!IMPORTANT]  
> Add a link to the specification document, pointing at the specific section.  
> Any implementation detail that could be considered idiosyncratic and not direct from the specification, should be  
> explained.  
> Always consider what someone new would need to know for the code to make sense, and add this context to it.

## 3. Who are the specification authors and who is accountable for this PR?

> [!IMPORTANT]  
> Ping the original specification author(s) and add them as reviewers, alongside any other from the engineering team.  
> You are accountable for the PR and its rapid resolution, so make sure anyone involved is aware and actively
> advancing  
> the review.

## 4. Is the specification accurate and complete?

> [!IMPORTANT]
> * If not, engage in a conversation directly with the authors, prompting any updates in the document.
> * You are accountable of this process, and the specification author needs to give priority to unblocking you.

## 5. Does the implementation introduce changes in the specification?

> [!IMPORTANT]
> * Contact the specification author(s) and engage in the necessary conversation to agree on the proposed  
    modification/solutions.
> * Ensure the new changes are on par with the implementation at the end of the process.
> * The PR should not be merged until parity between implementation and specifications has been reached.

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [ ] 1. Description added.
* [ ] 2. Context and links to Specification document(s) added.
* [ ] 3. Main contact(s) (developers and specification authors) added
* [ ] 4. Implementation and Specification are 100% in sync including changes. This is critical.
* [ ] 5. Link PR to a specific milestone.

PR #1353, opened 2025-06-12
feat(da): nomos da mock membership adapter integration

pradovic:

This is a 3rd PR related to implementing membership in da: https://github.com/logos-co/nomos/pull/1341. 

The purpose of this PR is to initialize MockMembershipAdapter in network service and implement usage as describe in mock tests. 
It also covers all the needed glue between sdp, membership and da-network in order to connect mock adapter.

The update logic is still not implemented, and the membership handler is still read from the config, so behavior should be the same.

 The next PR should probably be to implement membership service to read from config and initialize da-membership from membership service instead of it's own config.

Yes

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [x] 1. Description added.


nomos-services/sdp/src/adapters/services/services_repository.rs
```diff
@@ -1,9 +1,10 @@
 use async_trait::async_trait;
 use nomos_sdp_core::ledger;
 
+#[derive(Debug, Clone)]
 pub struct LedgerServicesAdapter;
 
-pub trait SdpServicesAdapter {
+pub trait SdpServicesAdapter: ledger::ServicesRepository {
     fn new() -> Self;
 }
```
danielSanchezQ, line 9: I'm not sure about this traits with new on them. If we need to instantiate maybe lets just add it to the parent trait.

nomos-services/sdp/src/adapters/declaration/repository.rs
```diff
@@ -1,11 +1,15 @@
...
+    ledger::{self, DeclarationsRepository, DeclarationsRepositoryError},
     DeclarationInfo,
 };
 
 pub struct LedgerDeclarationAdapter;
 
+pub trait SdpDeclarationAdapter: DeclarationsRepository {
+    fn new() -> Self;
+}
+
```
danielSanchezQ, line 12: Yeah, likewise I think. I know we had this before and it will make a bit of a refactor in the types.

nomos-services/data-availability/network/src/lib.rs
```diff
@@ -122,29 +178,60 @@ where
...
 
+        let membership_service_relay = overwatch_handle
+            .relay::<MembershipServiceAdapter::MembershipService>()
+            .await?;
+
+        let storage_adapter = StorageAdapter::default();
+
+        let membership_service_adapter = MembershipServiceAdapter::new(
+            membership_service_relay,
+            membership.clone(),
```
danielSanchezQ: why `clone`, do we use it afterwards? I think not.

nomos-services/data-availability/network/src/membership/adapters/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    fn new(
+        relay: OutboundRelay<<Self::MembershipService as ServiceData>::Message>,
+        handler: DaMembershipHandler<Membership>,
+        storage: Storage,
+    ) -> Self;
+
+    async fn subscribe(&self) -> Result<MembershipSnapshotStream, MembershipAdapterError>;
+
+    async fn update(&self, block_number: u64, new_members: HashMap<PeerId, Multiaddr>);
```
danielSanchezQ: Having the update on a comsumer adapter is actually quite confusing.

nomos-services/data-availability/network/src/storage/adapters/mock.rs
```diff
@@ -0,0 +1,19 @@
...
+impl MembershipStorage for MockStorage {
+    fn store(&self, block_number: u64, assignations: Assignations) {
+        let mut storage = self.storage.lock().unwrap();
+        storage.insert(block_number, assignations);
+    }
+
+    fn get(&self, block_number: u64) -> Option<Assignations> {
+        self.storage.lock().unwrap().get(&block_number).cloned()
+    }
+}
```
danielSanchezQ: If this have the store it means the network service is responsible of storing it. I know we had this convo already, it can be done here, but somehow it feels a bit sketchy.

danielSanchezQ (commented):
Welcome to generics hell @pradovic!

nomos-services/sdp/src/adapters/services/services_repository.rs
pradovic: Yes, I remember we talked about it. Let's fix it in different PR, I did not want to bloat this one, so I just made it compile. I need to clean those adapters in membership. It should be easy to clean up, but the changsets look ugly so I thought to avoid changing it here. Does that makes sense?

nomos-services/data-availability/network/src/lib.rs
pradovic: I think because Adapter wants to move it, and it is also in the service, so one ref stays in service and one is going to the adapter. When I remove the membership handler config and start relying on membership service for init then I think this can be cleaned up

nomos-services/data-availability/network/src/membership/adapters/mod.rs
pradovic: I have just added `subscribe()` the update was here from the previous PR and in the test cases that show the functionality. The functionality was that network service calls `update` method. Do you suggest to remove `update` all together and move the loop and task inside the adapter?

danielSanchezQ: No, I was just making a comment. I do not have an alternative actually.

nomos-services/data-availability/network/src/storage/adapters/mock.rs
pradovic: I am not sure I understand. Could you elaborate a bit more please? 
This was also a mock storage and usage from the previous PR. Are you suggesting to remove the storage from the adapter? I think we are using storage so we can get the previous values when needed in order to update, and the previous values are related to the previous assignations, not general membership snapshots

danielSanchezQ: > I am not sure I understand. Could you elaborate a bit more please? This was also a mock storage and usage from the previous PR. Are you suggesting to remove the storage from the adapter? I think we are using storage so we can get the previous values when needed in order to update

Yeah, I mean that the service taking snapshot is the network service. I'm not sure it should actually the one at all. But we can move it later I guess.

nomos-services/data-availability/network/src/membership/adapters/mod.rs
pradovic: Ok, I will think about it more. Another idea was to reduce membership adapter to be just a client to wards membership service, but this was suggested and I started to like this idea as well. If we don't like it in the end, maybe we can try to reduce the adapter only to "client" and code around it in network service.

nomos-services/data-availability/network/src/storage/adapters/mock.rs
pradovic: Hmm, this makes sense. Let's see if this glue code is ok, and I can try to refactor storage later. I think it will be easier to review :)

pradovic: Or, I can first try to solve it in a different PR and then update this one. But it is def related to update implementation

nomos-services/data-availability/network/src/membership/adapters/mod.rs
bacv: > Having the update on a consumer adapter is actually quite confusing.

It's because adapter is also handling storage, we could split this functionality so that adapter only subscribes, `update` and `get_historic_membership` would be in `MembershipStorage` or similar trait.

pradovic: Another approach could be to spawn a task to listen to subscription events inside the adapter, removing the need for update in the trait. The adapter could subscribe in the new, init or bootstrap method(), so the network service wouldn’t need to be aware of update at all. Just an idea  as I’m not sure if this responsibility belongs entirely in the adapter, but it might be a cleaner separation if we are still up for the adapter idea. Here’s an example of what I had in mind: https://github.com/logos-co/nomos/pull/1357. Let me know if this looks less or more confusing.

pradovic: Discarded https://github.com/logos-co/nomos/pull/1357 as bad design.

bacv: The changes in this PR are very close to final solution, the key here is that Adapter got new functionality and now the naming doesn't work anymore - we need to keep subscription in the Adapter, but update and historic data shouldn't stay in here anymore. Will create a suggesion targeting this PR.

nomos-services/data-availability/network/src/storage/adapters/mock.rs
bacv: With the suggestion merged, the storage adapter allows to store it's own historic snapshots of subnetwork assignations. These are different from the snapshots that MembershipService takes - providers with their locators doesn't hold the subnetwork assignation at the given moment. Since this information is only required for DA network service - the service manages the storage of these assignations by itself.

Do you think this thread can be resolved now @danielSanchezQ ?

nomos-services/data-availability/network/src/membership/adapters/mod.rs
bacv: Resolving, fixed in https://github.com/logos-co/nomos/pull/1359

bacv (approved):

nomos-services/membership/src/backends/mod.rs
```diff
@@ -14,6 +14,7 @@ pub struct MembershipBackendSettings {
     pub settings_per_service: HashMap<ServiceType, Settings>,
 }
 
+#[derive(Clone, Debug)]
```
ntn-x2: You could add `Copy` as well.

ntn-x2 (commented):
Just some small comments. I am not very familiar with this parts of the codebase.

danielSanchezQ (approved):
:zap:

PR #1354, opened 2025-06-13
feat: add dry-run capabilities to binaries and add checks in CI

ntn-x2:

We add a boolean flag that allows us to dry-run a binary and exit right after the config file is deserialized, to only check for its validity. This can be enabled with the `--check-config` on both binaries. The new capability is integrated into our CI, in the test job, to re-use the binaries that are anyway built for integration tests.

I also tried to add the [`deny_unknown_fields`](https://serde.rs/container-attrs.html#deny_unknown_fields) to the overarching config objects, but it seems they do not work when `flatten` is used, which we do, and it needs further investigation. So right now the checks fail if a field is missing but do not detect if there are unused fields. Worth looking into it in the future.

I also fixed the config files by using the values used in integration tests.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 1. Description added.

### DIFF
--- a/.github/workflows/code-check.yml
+++ b/.github/workflows/code-check.yml
@@ -197,19 +197,17 @@ jobs:
       - name: Run the node binary to check the config file validity
         uses: actions-rs/cargo@9e120dd99b0fbad1c065f686657e914e76bd7b72  # Version 1.0.1
         env:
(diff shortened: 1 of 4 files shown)


danielSanchezQ (approved):
Nice! Thanks!

PR #1356, opened 2025-06-13
fix(bundler): Safety

strinnityk:
Due to changes in `set_var`'s safety, we need to do some unsafe wrapping.

Yes.

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 1. Description added.

### DIFF
--- a/nomos-bundler/nomos-node/bundle.rs
+++ b/nomos-bundler/nomos-node/bundle.rs
@@ -16,32 +16,33 @@ const CRATE_NAME: &str = "nomos-node";
 const CRATE_PATH_RELATIVE_TO_WORKSPACE_ROOT: &str = "nodes/nomos-node";
 
 /// Prepares the environment for bundling the application
-///
-/// # Safety
-///
-/// This function is unsafe because it sets environment variables,
-/// which can lead to unexpected behaviour in multithreaded contexts.
-///
-/// For more information, see the documentation for [`set_var`].
-unsafe fn prepare_environment(architecture: &str) {
-    // Bypass an issue in the current `linuxdeploy`'s version
-    set_var("NO_STRIP", "true");
+fn prepare_environment(architecture: &str) {
+    // SAFETY: This is a single-threaded context as no threads have been spawned
+    // yet. Due to this, it is safe to call `build_package` which internally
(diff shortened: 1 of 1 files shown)


nomos-bundler/nomos-node/bundle.rs
```diff
@@ -15,19 +15,33 @@ use tauri_utils::platform::target_triple;
...
+///
+/// # Safety
+///
+/// This function is unsafe because it sets environment variables,
+/// which can lead to unexpected behaviour in multithreaded contexts.
+///
+/// For more information, see the documentation for [`set_var`].
+unsafe fn prepare_environment(architecture: &str) {
+    // Bypass an issue in the current `linuxdeploy`'s version
     set_var("NO_STRIP", "true");
```
ntn-x2: Do you think it makes sense to wrap these two calls into an `unsafe` block? Just in case we add more instructions in the future?

nomos-bundler/nomos-node/bundle.rs
```diff
@@ -116,7 +130,7 @@ fn build_package(version: String) {
     prepare_environment(arch);
```
ntn-x2, line 131: Should this also be wrapped in an `unsafe` block?

ntn-x2 (approved):
I think we can `unsafe`ify a smaller surface of the code. Other than that, LGTM!

nomos-bundler/nomos-node/bundle.rs
strinnityk: I don't know what's' the best approach for using `unsafe`, so we could do that, yeah.
The only thing is: The only place from where you can ensure you're in single-thread context is in `main`, so I felt a bit weird using `unsafe` so far below. Do yo think it's a good idea?

ntn-x2: Yeah ideally we want to use `unsafe` blocks only for the unsafe operations, so to draw a very small area of where things could potentially go wrong. Of course this is not the case, but it would still help collaboration if we know exactly where an `unsafe` function is actually unsafe. Also, I think perhaps we don't have to make the `build_package` itself unsafe, could it be?

PR #1357, opened 2025-06-15
Feature nomos da membership adapter mock trait idea

pradovic:


pradovic:
@bacv, @danielSanchezQ Init and bootstrap (from the adapter trait) could also be reduced in single method maybe when we remove membership handler config, if we could initialize adapter in run method.

Currently, membership handler is needed (for behaviors) in swarm init which is happening in the init phase of network service. If I am not mistaken, swarm will be started before network service run method is initialized. Is this by design? Would it be more controlled maybe to initialize swarm in run method? This way we could initialize membership in run as well, and even bootstrap it from membership service before starting the swarm.

nomos-services/data-availability/network/src/membership/adapters/mock.rs
```diff
@@ -58,21 +61,21 @@ where
...
+    async fn bootstrap(self: Arc<Self>) -> Result<(), MembershipAdapterError> {
+        let mut stream = self.subscribe_stream(ServiceType::DataAvailability).await?;
+
+        tokio::spawn(async move {
+            while let Some((block_number, new_members)) = stream.next().await {
+                Self::update(block_number, new_members);
+            }
+        });
+
+        Ok(())
```
danielSanchezQ, line 78: This is a sign of bad design. We used to have a stream which is polled by the upper service, there is no need for more tasks. We should avoid creating this tiny tasks.

pradovic: Ok, it was just an idea, as I thought we want to keep everything in the adapter. 
I will close the PR

PR #1358, opened 2025-06-16
feat(mantle): SignedMantleTx in node

bacv:

Changed previously used Tx type to SignedMantleTx in the node, added simple test of posting such transaction to the mempool. 

Yes

@bacv @danielSanchezQ @zeegomo 

Yes

No

## Checklist

* [x] 1. Description added.

### DIFF
--- a/nomos-core/chain-defs/src/tx/mod.rs
+++ b/nomos-core/chain-defs/src/tx/mod.rs
@@ -19,6 +19,11 @@ pub trait Transaction {
     fn hash(&self) -> Self::Hash {
         Self::HASHER(self)
     }
+    /// Returns the bytes that are used to form a signature of a transaction.
+    ///
+    /// The resulting bytes are then used by the `HASHER`
+    /// to produce the transaction's unique hash, which is what is typically
+    /// signed by the transaction originator.
     fn as_sign_bytes(&self) -> Bytes;
 }
 


nomos-mantle/src/tx.rs
```diff
@@ -68,9 +68,30 @@ impl nomos_core::tx::Transaction for MantleTx {
...
+        self.mantle_tx.hash(state);
+    }
+}
+
+impl nomos_core::tx::Transaction for SignedMantleTx {
+    const HASHER: TransactionHasher<Self> = |tx| blake2::Blake2b::digest(tx.as_bytes()).into();
+    type Hash = TxHash;
+
+    fn as_bytes(&self) -> bytes::Bytes {
+        self.mantle_tx.as_bytes()
```
zeegomo: why just the mantle tx?

zeegomo (changes requested):

nomos-mantle/src/tx.rs
bacv: Didn't find a specification for how SignedMantleTx hash should be computed, but when thinking about it - we only care about MantleTx, I don't see a purpose to have two different hashes for both.

danielSanchezQ: But then it should not be implemented for the Signed version of it (?)

danielSanchezQ: Btw, should the `Transaction` be implemented by the Signed wrapper? I think it should be only for the inner mantle one (?)

bacv: `Transaction` trait is what consensus and mempool are expecting. If we pass SignedMantleTx to consensus and mempool, then `SignedMantleTx` should implement `Transaction`. Transaction trait implementation is actually not needed for the MantleTx, at least it's not used anywhere.
In the end I see both of these transactions as one - they are related via one thing - hash.

zeegomo: are `Hash` and `as_bytes` tied together? I would expect that `as_bytes` returns something I can convert back to the original item (and yes, tx hash ignores signatures)

bacv: > are Hash and as_bytes tied together

Yes, the `Transaction` trait is defined in [`nomos-core`](https://github.com/logos-co/nomos/blob/master/nomos-core/chain-defs/src/tx/mod.rs#L14-L23). Maybe you had different idea for `Transaction` trait when initially defining, but at the moment there's no other use for it than to get the hash.

I agree that `as_bytes` is bit confusing, there were some ideas to rename this trait to `TransactionHash` or similar, but then `as_bytes` method wouldn't make sense even more. I suggest to keep it as is - if this trait is implemented - we can get 1) hash 2) transaction bytes.

zeegomo: I think currently it's quite confusing as `as_bytes` is a common method name in Rust that carries some semantics. What about `as_sign_bytes` implying those are the bytes to be signed?

bacv: Updated here https://github.com/logos-co/nomos/pull/1358/commits/6b56af77761dbcf6a39350933f1e1436f904897d

PR #1362, opened 2025-06-17
feat(sync): integrate cryptarchia-sync to nomos-libp2p

andrussal:

This PR integrates `cryptarchia-sync`(Bootstrapping networking) to `nomos-libp2p`.  
- Add cryptarchia-sync Behaviour to Swarm
- Extend nomos-libp2p::Swarm with `request_tip` function
- Extend nomos-libp2p::Swarm with `start_blocks_download` function

@andrussal 

N/A

N/A

## Checklist

* [x] 1. Description added.


nomos-libp2p/src/behaviour/chainsync/swarm_ext.rs
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    pub fn start_blocks_download(
+        &self,
+        peer_id: PeerId,
+        target_block: HeaderId,
+        local_tip: HeaderId,
+        latest_immutable_block: HeaderId,
+        additional_blocks: HashSet<HeaderId>,
+        reply_sender: Sender<BoxStream<'static, Result<SerialisedBlock, ChainSyncError>>>,
```
danielSanchezQ: I think it would be nice to have at least a type alias for this type as it is a bit bit:


```suggestion
        reply_sender: Sender<SerializedBlockStream>,
```

danielSanchezQ (approved):
Added a tiny suggestion, otherwise nice to go. Thanks!

nomos-libp2p/src/config/chainsync.rs
```diff
@@ -0,0 +1,5 @@
+use serde::{Deserialize, Serialize};
+
+/// Chainsync configuration settings. When set to `None`, chainsync is disabled.
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct Settings;
```
ntn-x2: Is this needed at all?

andrussal: I think it makes sense to remove it. I followed the pattern here but right now I can't think of a situation when sync behaviour should be disabled.

youngjoon-lee: I agree with removing this. We can add it when necessary. 

> I can't think of a situation when sync behaviour should be disabled.

Right. We don't need an option to disable sync

Issue #1289, opened 2025-05-25
Reevaluate unhealthy Blend node connections


youngjoon-lee:
The new Blend v1 spec defines [a step that periodically reevaluate unhealthy connections](https://www.notion.so/Blend-Protocol-Version-1-PENDING-MIGRATION-1c48f96fb65c809494efe63019a5ebfb?source=copy_link#1c48f96fb65c80b3859ce9bd690e398e). 

We had implemented the connection maintenance mechanism that continuously monitors the frequency of incoming messages per connections and marks connections, which has less frequency than expected, as unhealthy.
But, we hadn't implemented a mechanism to keep monitoring unhealthy connections and restore them to the healthy state when their frequency becomes normal.

ntn-x2:
Closed with https://github.com/logos-co/nomos/pull/1365.

PR #1365, opened 2025-06-17
chore: update connection maintenance code based on new spec

ntn-x2:

This PR restructures the connection management component of the Blend stack to account for the new algorithm described in the spec. This PR does not tackle yet https://github.com/logos-co/nomos/issues/1289, which will be tackled as a follow-up PR based on this new structure, which is now trivial. We have the concept of rounds and observation window for a peer, so we introduce those concepts here.

Furthermore, I moved the `conn_maintenance` module inside `nomos-blend-network` from its previous location in `nomos-core`, since it feels like the best place for it. Along the same lines of the cover traffic refactoring, the connection monitor component now expects a stream which periodically return a range of messages to be expected from the remote peer. This range is derived from some parameters and from the size of the core nodes, as described in the [spec section](https://www.notion.so/Blend-Protocol-Version-1-MIGRATED-1c48f96fb65c809494efe63019a5ebfb?source=copy_link#1c88f96fb65c804bae6df7a19c3e3e1f). In the long term, the stream will wrap around the membership info, calculate the right expected number of messages, and pass that along to the connection monitor, which can now be dynamically updated.

Once the next PR will tackle managing unhealthy peers as https://github.com/logos-co/nomos/issues/1289 describes, and once the message encapsulation PR is merged, I think the time is right to refactor the code based on the new spec, which will make managing all these streams much easier in the future.

Yes.

Spec: @madxor, implementor: @ntn-x2, code owner: @youngjoon-lee.

Yes.

No.

## Checklist

* [x] 1. Description added.


nomos-blend/network/src/tokio.rs
```diff
@@ -0,0 +1,376 @@
...
+
+        fn interval_stream(&self) -> Self::IntervalStream {
+            let interval = self.0;
+            let range = self.1.clone();
+            Box::new(
+                tokio_stream::wrappers::IntervalStream::new(tokio::time::interval(interval))
+                    .map(move |_| range.clone()),
+            )
+        }
+    }
```
youngjoon-lee, line 89: Could we probably use`ObservationWindowTokioIntervalProvider` instead of having `TestTokioIntervalStreamProvider` in the tests?

nomos-blend/network/src/conn_maintenance.rs
```diff
@@ -0,0 +1,184 @@
...
+            ConnectionMonitorOutput::Spammy
+        } else if self.is_unhealthy() {
+            ConnectionMonitorOutput::Unhealthy
+        } else {
+            ConnectionMonitorOutput::Healthy
+        };
+        self.reset(new_expected_message_count_range);
+        Poll::Ready(outcome)
+    }
+}
```
danielSanchezQ, line 101: Why not implement `Future` instead? Unless there is an incompatibility usually Future is easier to read and less error phrone (?)

danielSanchezQ: If it is because its supposed to be polled for continuous outcomes and doesn't fuse on resolution then probably `Stream` instead.

danielSanchezQ (commented):
:rocket:

nomos-blend/network/src/conn_maintenance.rs
ntn-x2: Yeah this bit I did not touch, but I still think `Stream` is the right behaviour because the connection monitor is `poll`ed within the connection handler for new events continuously. Every time the observation window ticks, the monitor yields a new result.

nomos-blend/network/src/tokio.rs
ntn-x2: Mh the problem with that is that we would need to find the right set of parameters to yield the expected message window we want to test. So i thought in this case the interval provider is not what we want to test, but rather how two swarms interact with each other, so I thought about having a test implementation that lets us define exactly what is the range of expected messages. Do you want me to switch implementation and find the right set of parameters for the expected message range instead?

youngjoon-lee: I agree with your point. This code is good and simple enough. Thanks.

PR #1366, opened 2025-06-17
feat(sync): Integrate chain sync events/commands to network service

andrussal:

This PR updates `noms-services/network`  to handle `cryptarchia-sync` related events and commands. 
- Add support to send `ChainSyncCommand`
- Add support to subscribe to `cryptarchia-sync` Events

Note: this should be merged to master but for review I set target to `feat/bootstrapping-nomos-libp2p`.  It can be merged after `feat/bootstrapping-nomos-libp2p` is merged.

@andrussal 

N/A

N/A

## Checklist

* [x] 1. Description added.


consensus/cryptarchia-sync/src/errors.rs
```diff
@@ -68,3 +68,41 @@ impl From<(PeerId, PackingError)> for ChainSyncError {
         }
     }
 }
+
+impl Clone for ChainSyncErrorKind {
```
ntn-x2, line 73: Can you add a small comment explaining why manual `Clone` impl is required?

nomos-services/network/src/backends/libp2p/swarm/mod.rs
```diff
@@ -41,7 +44,8 @@ pub struct SwarmHandler {
     pub pending_dials: HashMap<ConnectionId, Dial>,
     pub commands_tx: mpsc::Sender<Command>,
     pub commands_rx: mpsc::Receiver<Command>,
-    pub events_tx: broadcast::Sender<Event>,
+    pub pubsub_events_tx: broadcast::Sender<Message>,
```
ntn-x2: This is not for events anymore, is it?

nomos-services/network/src/backends/mod.rs
```diff
@@ -13,10 +13,12 @@ pub mod mock;
...
     type Message: Debug + Send + Sync + 'static;
-    type EventKind: Debug + Send + Sync + 'static;
-    type NetworkEvent: Debug + Send + Sync + 'static;
+    type PubSubEvent: Debug + Send + Sync + 'static;
+    type ChainSyncEvent: Debug + Send + Sync + 'static;
 
     fn new(config: Self::Settings, overwatch_handle: OverwatchHandle<RuntimeServiceId>) -> Self;
     async fn process(&self, msg: Self::Message);
-    async fn subscribe(&mut self, event: Self::EventKind) -> Receiver<Self::NetworkEvent>;
+    async fn subscribe_to_pubsub(&mut self) -> Receiver<Self::PubSubEvent>;
```
ntn-x2: Wouldn't this be better as a stream that as a broadcast channel?

ntn-x2 (approved):

nomos-services/network/src/backends/mod.rs
andrussal: Would this work with stream? My understanding is that it has multiple consumers.

ntn-x2: There exists a `BroadcastStream` type as well: https://docs.rs/tokio-stream/latest/tokio_stream/wrappers/struct.BroadcastStream.html which is a wrapper around a broadcast receiver. If cloning happens inside, then hiding it behind a `impl Stream` would be sufficient. If cloning happens on the caller side, then returning a `BroadcastStream` is required. This is, of course, if you think using a stream interface would be a better choice than using a channel interface. I have been wondering about the same things.

danielSanchezQ: > There exists a `BroadcastStream` type as well: https://docs.rs/tokio-stream/latest/tokio_stream/wrappers/struct.BroadcastStream.html which is a wrapper around a broadcast receiver. If cloning happens inside, then hiding it behind a `impl Stream` would be sufficient. If cloning happens on the caller side, then returning a `BroadcastStream` is required. This is, of course, if you think using a stream interface would be a better choice than using a channel interface. I have been wondering about the same things.

Yeah, I think this was handled in the adapters which transform this to a stream in most cases.

nomos-services/cryptarchia-consensus/src/network/adapters/libp2p.rs
```diff
@@ -91,18 +91,15 @@ where
         let (sender, receiver) = tokio::sync::oneshot::channel();
         if let Err((e, _)) = self
             .network_relay
-            .send(NetworkMsg::Subscribe {
-                kind: EventKind::Message,
-                sender,
-            })
```
danielSanchezQ, line 97: Why are we changing this? And why in this PR?

nomos-services/network/src/backends/libp2p/mod.rs
```diff
@@ -14,44 +19,42 @@ pub use self::{
...
 const BUFFER_SIZE: usize = 64;
 
 #[async_trait::async_trait]
 impl<RuntimeServiceId> NetworkBackend<RuntimeServiceId> for Libp2p {
     type Settings = Libp2pConfig;
     type Message = Command;
-    type EventKind = EventKind;
-    type NetworkEvent = Event;
+    type PubSubEvent = Message;
+    type ChainSyncEvent = cryptarchia_sync::Event;
```
danielSanchezQ, line 30: This is wrong (?). This should not be another type. But the old `EventKind` should be expanded to support cryptarchia sync ones. Why adding an associated type?

nomos-services/cryptarchia-consensus/src/network/adapters/libp2p.rs
andrussal: Old subscribe function returned a Receiver of type NetworkMessage. In a previous PR we discussed a situation where we don't want receiver side to do patter matching over return returned value. If it subscribes to pubsub, it ideally just receives `pubsub` messages. Or `chainsync` messages. So I decoupled these behind different subscribe functions. Then passing `EventKind` didn't seem anymore needed. And I replaced `NetworkMessage` with `PubSubEvent` and `ChainSyncEvent`.

danielSanchezQ: Yeah, I understood it later on the decoupling and forgot to remove the initial comment. Thanks for the explanation.
But I think we could have kept this and just update the eventkind, instead of introducing more types. We can continue the convo in the other comment though.

nomos-services/network/src/backends/mod.rs
andrussal: > Yeah, I think this was handled in the adapters which transform this to a stream in most cases.

Perhaps it makes sense to change it to a BroadcastStream as we do it anyway in adapters but in another PR?

danielSanchezQ: > > Yeah, I think this was handled in the adapters which transform this to a stream in most cases.
> 
> Perhaps it makes sense to change it to a BroadcastStream as we do it anyway in adapters but in another PR?

Yeah, we can do that. Sounds good.

nomos-services/network/src/backends/libp2p/mod.rs
andrussal: @danielSanchezQ Is this a separate concern or related to https://github.com/logos-co/nomos/pull/1366#discussion_r2152355257? `EventKind` was previously passed along with subscription to receive `NetworkMessage`s. But it supported only one event kind and `NetworkMessage` had only one variant.

danielSanchezQ: Uhm, I see. Well, I think splitting may be more correct. I had a different structure in mind from older code. And I didn't grasp the changes well. Sorry for that. 

TLDR: This version is good.

andrussal: Thank you!

PR #1367, opened 2025-06-17
feat(mempool): Mempool payload processing

bacv:

Proposal for enabling mempool to trigger actions according to the received transaction or operations before adding it to the mempool.

NoOp implementation is added to keep current behaviour, an example `SignedTransactionProcessor` contains a potential code that would trigger sampling for a blob.

Preparation to remove DA mempool

@bacv @danielSanchezQ 

N/A

N/A

## Checklist

* [x] 1. Description added.

### DIFF
--- a/nomos-services/mempool/src/processor/tx.rs
+++ b/nomos-services/mempool/src/processor/tx.rs
@@ -68,6 +68,10 @@ where
             })
             .collect();
 
+        // In this case errors are about sending the `TriggerSampling` message to the DA
+        // Sampling Service. These errors do not represent the result of
+        // sampling itself, but might indicate that sending the message to
+        // sampling service failed.
         let errors: Vec<SignedTxProcessorError> = StreamExt::collect::<Vec<_>>(all_futures)
             .await
             .into_iter()


nodes/nomos-executor/executor/src/lib.rs
```diff
@@ -194,4 +195,5 @@ pub struct NomosExecutor {
     http: ApiService,
     storage: StorageService,
     system_sig: SystemSigService,
+    no_service: NoService,
```
danielSanchezQ: `?` why is this?

nodes/nomos-node/node/src/lib.rs
```diff
@@ -190,4 +191,5 @@ pub struct Nomos {
     http: ApiService,
     storage: StorageService,
     system_sig: SystemSigService,
+    no_service: NoService,
```
danielSanchezQ: leftovers?

nomos-services/mempool/src/processor/tx.rs
```diff
@@ -0,0 +1,60 @@
...
+        for op in &payload.mantle_tx.ops {
+            if let Op::Blob(blob_op) = op {
+                self.sampling_relay
+                    .send(DaSamplingServiceMsg::TriggerSampling {
+                        blob_id: blob_op.blob,
+                    })
+                    .await
+                    .map_err(|(e, _)| SignedTxProcessorError::sampling_error(e))?;
+            }
+        }
```
danielSanchezQ: Nipick:
1. Filter only blobs
2. Map sending the message
3. Join all futures (either join or in a futures_unordered)
4. Either gather error or fuse from the futures_unordered

nodes/nomos-node/node/src/lib.rs
bacv: No, added this to have fully working PoC without propagating types everywhere in mempool related services - when plugging in the `SignedTxProcessor` we'll need to propagate the `<Processor>` type. At the moment it's hardcoded here https://github.com/logos-co/nomos/pull/1367/files#diff-ca797e6e2c0f5ada3af86c41728f417cb7b1fd3428b65780ab51ade0e4dac0b7R36-R58

nodes/nomos-executor/executor/src/lib.rs
bacv: If PR looks good, I won't merge it, but remove Noop stuff and add all the stuff required to `SignedTxProcessor`, would be great @zeegomo to take a look before this.

nomos-services/mempool/src/processor/tx.rs
bacv: Something like this:
```
        let mut all_futures = FuturesUnordered::new();

        for op in &payload.mantle_tx.ops {
            if let Op::Blob(blob_op) = op {
                let fut = async {
                    self.sampling_relay
                        .send(DaSamplingServiceMsg::TriggerSampling {
                            blob_id: blob_op.blob,
                        })
                        .await
                        .map_err(|(e, _)| SignedTxProcessorError::sampling_error(e))
                };
                all_futures.push(fut);
            }
        }

        let mut errors = Vec::new();
        while let Some(result) = all_futures.next().await {
            if let Err(e) = result {
                errors.push(e);
            }
        }

        if errors.is_empty() {
            Ok(())
        } else {
            Err(errors) // Returns Vec<SignedTxProcessorErrors>
        }
   ```

Comment (danielSanchezQ) [reply nomos-services/mempool/src/processor/tx.rs]: Formatting below sucks but just to get the idea maybe is enough:

```rust
let mut all_futures: FuturesUnordered = 
       payload.mantle_tx.ops
            .iter()
            .copied()
            .filter_map(|op| if let Op::Blob(blob_op) = op) {self.sampling_relay
                        .send(DaSamplingServiceMsg::TriggerSampling {
                            blob_id: blob_op.blob,
                        })} else {None}).collect();
let results: Result<(), WhateverError> = StreamExt::collect(all_futures).await;
results
```

nodes/nomos-node/node/src/lib.rs
zeegomo: man these generics are coming after us

nomos-services/mempool/src/tx/service.rs
```diff
@@ -177,16 +218,30 @@ where
...
+                Some((key, item)) = network_items.next() => {
+                    if let Err(e) =processor.process(&item).await {
+                        tracing::debug!("could not process item from network due to: {e:?}");
+                        continue;
+                    }
+                    if let Err(e) = self.pool.add_item(key, item) {
                         tracing::debug!("could not add item to the pool due to: {e}");
-                    });
+                        continue;
+                    }
```
zeegomo, line 287: sampling can definitely take a while, we should run sampling concurrently for different txs as well

zeegomo (changes requested):
Overall looks good, but I feel like we might want to explore some other options rather than generics for everything, 200 lines just to change a type feels a bit overkill

nomos-services/mempool/src/tx/service.rs
bacv: When blob operation is being processed - it only "triggers" the sampling and returns immediately. In here it's assumed that other ops would be checked similarly - if they need to run some background task, processor would trigger it (via service or in a thread) and return imediatly; or - perform some required validation before adding.

Or do you suggest that we should process in parallel and then remove item later if something fails?

zeegomo: so it does not actually wait for the result of the sampling? 
I would add a comment to the `process` method to remind implementors of this then

How are you expected to act on the result of the operation, for example if sampling fails?

bacv: > How are you expected to act on the result of the operation, for example if sampling fails?

In general - processor implementation decides if item can be added to the pool or not. 

In case of blob operation - mempool just cares about triggering sampling - but not the result of sampling. This specific implementation returns result of "did sending trigger message to sampling service succeeded".

Why we don't wait for sampling result here?:
  - because mempool doesn't use this information;
  - sampling result is polled by consensus from sampling service when blob is being validated;
  - sampling result is polled by consensus when a new block proposal is being formed;

zeegomo: well, why would we keep items in the mempool if they're not valid? 
Like, we run sampling, if fails, should we remove it from the mempool entirely instead of waiting for somebody to request its status? What's the rationale for not doing that.
It's not bad per se, I'm just curious to know the reason

bacv: Seems that both solutions would achieve the same goal - not sure how one be better than other apart from that "triggering" solution is already implemented and tested.

Current solution:
  - Mempool receives item and triggers sampling
  - Sampling service samples multiple subnets for blob
  - When consensus is proposing tx with blob ops:
    - pulls list of sampled blobs from sampling service;
    - pulls txs that contain sampled blobs from mempool;
  - When consensus is processing block:
    - verifies list of blob ops via sampling service;
  - Items that stay in pool for too long gets pruned
 
Proposed solution (correct if wrong):
  - Mempool receives items and triggers sampling
  - Sampling service samples multiple subnets for blob
  - Instead of consensus, mempool periodically checks if blob sampling succeeded (or receives sampling result via chan)
  - When consensus is proposing tx with blob ops:
    - pulls any txs from mempool because they are sampled
  - When consensus is processing block
    - verifies list of blob ops via sampling service
  - Only items that are proposed or processed in block are pruned, or some other logic

zeegomo: Indeed one is not necessarily better than the other, however

> Items that stay in pool for too long gets pruned

How's that decided? Since mempools need to be in sync across nodes to validate blocks, we need to be sure that I'm not deleting something that some leader might include later

bacv: > How's that decided? Since mempools need to be in sync across nodes to validate blocks

My mistake, at the moment the transactions are not pruned - we only have a service command to prune, but it's not used anywhere, this part will need to be decided and implemented.

zeegomo: I actually think we need to clean the mempool from invalid transactions regardless of other requests but it can be done in another pr.

Imagine a non-leader node or a node with very little stake which is rarely elected, it will never prune invalid txs from its mempool because those will ofc not be broadcast by incoming blocks.

zeegomo (approved):

danielSanchezQ (approved):
Look good to me. Generics hell, but well, we are used to it.

nomos-services/mempool/src/processor/tx.rs
```diff
@@ -0,0 +1,83 @@
...
+                            .await
+                            .map_err(|(e, _)| SignedTxProcessorError::sampling_error(e))
+                    })
+                } else {
+                    None
+                }
+            })
+            .collect();
+
+        let errors: Vec<SignedTxProcessorError> = StreamExt::collect::<Vec<_>>(all_futures)
```
zeegomo, line 75: does this wait for the sampling result? I think a comment indicating whether it does or not would be helpful

zeegomo (approved):
20% of the PR is actual new code, the remaining 80% (~600 lines) is generics and type params, I feel this is getting out of hand (not that it's the fault of this PR)

nomos-services/mempool/src/processor/tx.rs
bacv: Added a comment https://github.com/logos-co/nomos/pull/1367/commits/8737708b9508ce1c4b4931c1c2dd7f8dc8068b04
```
        // In this case errors are about sending the `TriggerSampling` message to the DA
        // Sampling Service. These errors do not represent the result of
        // sampling itself, but might indicate that sending the message to
        // sampling service failed.
```

bacv:
> 20% of the PR is actual new code, the remaining 80% (~600 lines) is generics and type params, I feel this is getting out of hand (not that it's the fault of this PR)

I think the Alex solution partially reflected in this PR where we keep Service generic, but also provide "less" generic service definition could help, but we'd need to apply this to chain service for best effect.

PR #1368, opened 2025-06-17
feat(membership): Membership service configuration

pradovic:

This PR is adding configuration for membership service. This is prerequisite for initializing da membership from membership service.

Yes.

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [x] 1. Description added.


nomos-sdp/src/lib.rs
```diff
@@ -51,12 +52,71 @@ pub enum ServiceType {
     ExecutorNetwork,
 }
 
+// Add Display and FromStr implementations so we can serialize
```
pradovic: @bacv. @danielSanchezQ I could not make ServiceType and ProviderId serialize nicely from config inside the `HashMap<ServiceType...>` or `HashMap<ProviderId...>`, getting the error message that key must be a string.

I know we already talked about this in the first PR and this solution was rejected in favor of transparent, but I added it in this draft as an illustration and wanted to see if you guys have a better idea instead.

Context:
https://github.com/logos-co/nomos/pull/1295/files#r2108825339

This can be tested by `test_address_book`. It started to fail when I added membership config because it could not serialize `ProviderId` and `ServiceType` inside Hashmap.

I guess if we want to mimic the current tests we should set some initial values in membership config. If we don't have to set initial values in membership config, then we don't need this at all. But I guess it would be useful for tests, at least as long as we don't implement  integrate sdp.

danielSanchezQ: uhm, deriving serde is not enough? I do not like that display is implemented for this purpose. That kind of thags are not really descriptive.  Instead probably we want a wrapper type on the hasmaps where we can control the serialization/deserialization when we send it to storage (if that is the problem as It wasnt clear to me).

pradovic: Tbh I thought it was enough as well. But what seems to happen is that serde would use derive when serializing ServiceType, but will not use it when serializing it inside the HashMap. I found this relevant question on stackoverflow:https://stackoverflow.com/questions/68579809/using-serde-to-deserialize-a-hashmap-with-a-enum-key. 

Wrapper type on Hashmaps could be a good solution. I can also try some serde_with combination.

nomos-sdp/src/lib.rs
```diff
@@ -51,12 +52,71 @@ pub enum ServiceType {
...
+}
+
 pub type Nonce = [u8; 16];
 
-#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
-#[serde(transparent)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
 pub struct ProviderId(pub [u8; 32]);
 
+impl Serialize for ProviderId {
```
pradovic, line 59: @bacv , @danielSanchezQ 
The similar question as ServiceType ⬆️  

The only difference is that it might be useful for ProviderId to be base64 string, if we are going to pass it to configs.

danielSanchezQ: This is fine indeed.

bacv: Could we change that to hex strings? I've seen we use `const-hex` in such cases, would be better to have everything in one format.

nomos-sdp/src/lib.rs
danielSanchezQ: Yeah, just the easiest thing to do would do.

pradovic: I have played a bit more and it seems that these parsing errors related to service type don't appear at all when ProviderId is correctly serialized/deserialized, so I have removed the annotation. I will go back to it when I test configs if I ecounter any problems.

nomos-sdp/src/lib.rs
pradovic: Fixed: https://github.com/logos-co/nomos/pull/1368/commits/522ad16fe107467a430ee4334cf7e045629055a9.

nomos-sdp/Cargo.toml
```diff
@@ -10,11 +10,13 @@ workspace = true
 [dependencies]
 async-trait = "0.1"
 blake2      = "0.10"
+const-hex   = "1"
 multiaddr   = "0.18"
 nomos-core  = { workspace = true }
 serde       = { version = "1.0", features = ["derive"] }
 thiserror   = "1"
-tracing     = { workspace = true }
+
```
bacv: Nit, there are some newlines in different cargo tomls

tests/src/tests/da/api.rs
```diff
@@ -128,7 +140,11 @@ async fn test_get_shares() {
     let app_id: [u8; 32] = app_id.try_into().unwrap();
     let metadata = kzgrs_backend::dispersal::Metadata::new(app_id, 0u64.into());
 
+    tokio::time::sleep(Duration::from_secs(15)).await;
```
bacv: We can remove these timeouts now

bacv (commented):
Looks good, will give another quick pass after it's ready for review.

tests/src/lib.rs
```diff
@@ -70,3 +70,14 @@ pub fn secret_key_to_peer_id(node_key: nomos_libp2p::ed25519::SecretKey) -> Peer
...
+#[must_use]
+pub fn secret_key_to_provider_id(
+    node_key: nomos_libp2p::ed25519::SecretKey,
+) -> nomos_sdp_core::ProviderId {
+    nomos_sdp_core::ProviderId(
+        nomos_libp2p::ed25519::Keypair::from(node_key)
+            .public()
+            .to_bytes(),
+    )
+}
```
danielSanchezQ, line 83: @bacv was this the provider id finally? Just double checking

bacv: Yes, here is provider id definition in the spec:

> provider_id: used to sign the SDP messages and to establish secure links between validators; it is Ed25519PublicKey.

Provider in `nomos-sdp` will soon have it's own module, it'll contain the serialization and conversion between types. Will add in another PR.

[Spec](https://www.notion.so/Service-Declaration-Protocol-Specification-1fd261aa09df819ca9f8eb2bdfd4ec1d?source=copy_link#1fd261aa09df81dfbf07e8d4a01ef0e0)

PR #1371, opened 2025-06-18
feat(blend): new message encapsulation

### LINKED ISSUE #1287: New Blend Message Encapsulation Mechanism
So far, we have used the Sphinx encapsulation mechanism for blend messages. Now that the new [Message Encapsulation Mechanism](https://www.notion.so/Message-Encapsulation-Mechanism-1fd261aa09df81b98f51c84ade85fe02?pvs=4) is finalized, we can start implementing it. 

The new mechanism includes the following components that could **not** be implement right now. We can leave them as **placeholders** until those are ready to be implemented.
- https://github.com/logos-co/nomos/issues/1315 can be implemented when our [SDP](https://www.notion.so/Service-Declaration-Protocol-Specification-1fd261aa09df819ca9f8eb2bdfd4ec1d?pvs=4) implementation is ready.
  - Until then, we can keep using a static set of Blend nodes.
- https://github.com/logos-co/nomos/issues/1316 and https://github.com/logos-co/nomos/issues/1317 can be ready to implemented once we have https://github.com/logos-co/nomos/issues/1315.

youngjoon-lee:

Closes #1287 

This PR implements the new Blend message encapsulation mechanism that will replace the Sphinx encoding. 

- Specs
  - [Message Encapsulation Mechanism](https://www.notion.so/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link)
    - [Message Formatting Specification](https://www.notion.so/Message-Formatting-Specification-215261aa09df81c79e3acd9e921bcc30?source=copy_link)
    - [Payload Formatting Spec](https://www.notion.so/Payload-Formatting-Specification-215261aa09df8153a456c555b7dcbe1c?source=copy_link)
    - [Key Types and Generation Specification](https://www.notion.so/Key-Types-and-Generation-Specification-1c88f96fb65c80cea617c21ae0272f2c?source=copy_link)
  - Parent: [Blend Protocol: Version 1](https://www.notion.so/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link)

- Tips for review
  - Recommended order
    - Basic data structures in `message.rs` (which follows [Message Formatting Specification](https://www.notion.so/Message-Formatting-Specification-215261aa09df81c79e3acd9e921bcc30?source=copy_link))
    - Encapsulation/Decapsulation in `encap.rs`. I recommend to read the tests first. 
  - Performance
    - This PR uses `serde` and `wire`. But the encapsulation requires a lot of memory manipulations (such as encryption and shifting). So, in the future, we can consider the zerocopy serde https://github.com/logos-co/nomos/issues/1372. For now, what this PR implements should be fine considering low message throughput (1 msg/sec).

- What's next (What this PR doesn't implement)
  - Implementing the `BlendMessage` trait for integration
  - Integration with `BlendService`
  - Generating/Verifying ProofOfQuota and ProofOfSelection

Yes

- Spec: @madxor 
- Code: @youngjoon-lee @ntn-x2 

Yes

No

## Checklist

> [!WARNING]  

* [x] 1. Description added.

### DIFF
--- a/nomos-blend/network/src/conn_maintenance.rs
+++ b/nomos-blend/network/src/conn_maintenance.rs
@@ -75,12 +75,14 @@ where
     /// If the interval has elapsed, evaluate the peer's status,
     /// reset the monitor, and return the result as `Poll::Ready`.
     /// If not, return `Poll::Pending`.
-    pub fn poll(&mut self, cx: &mut Context<'_>) -> Poll<ConnectionMonitorOutput> {
-        let Poll::Ready(Some(new_expected_message_count_range)) =
-            self.connection_window_clock.poll_next_unpin(cx)
-        else {
-            cx.waker().wake_by_ref();
-            return Poll::Pending;
+    pub fn poll(&mut self, cx: &mut Context<'_>) -> Poll<Option<ConnectionMonitorOutput>> {
+        let new_expected_message_count_range = match self
+            .connection_window_clock
+            .poll_next_unpin(cx)
+        {
+            Poll::Ready(Some(new_expected_message_count_range)) => new_expected_message_count_range,
+            Poll::Ready(None) => return Poll::Ready(None),
+            Poll::Pending => return Poll::Pending,
         };
         // First tick is used to set the range for the new observation window.
         if self.expected_message_range.is_none() {
@@ -96,7 +98,7 @@ where
             ConnectionMonitorOutput::Healthy
         };
         self.reset(new_expected_message_count_range);
-        Poll::Ready(outcome)
+        Poll::Ready(Some(outcome))
     }
 }
 
@@ -129,7 +131,7 @@ mod tests {
         monitor.record_message();
         assert_eq!(
             monitor.poll(&mut Context::from_waker(&noop_waker())),
-            Poll::Ready(ConnectionMonitorOutput::Healthy)
+            Poll::Ready(Some(ConnectionMonitorOutput::Healthy))
         );
 
         // Recording the maximum expected number of messages,
@@ -138,7 +140,7 @@ mod tests {
         monitor.record_message();
         assert_eq!(
             monitor.poll(&mut Context::from_waker(&noop_waker())),
-            Poll::Ready(ConnectionMonitorOutput::Healthy)
+            Poll::Ready(Some(ConnectionMonitorOutput::Healthy))
         );
 
         // Recording more than the expected number of messages,
@@ -148,21 +150,21 @@ mod tests {
         monitor.record_message();
         assert_eq!(
             monitor.poll(&mut Context::from_waker(&noop_waker())),
-            Poll::Ready(ConnectionMonitorOutput::Spammy)
+            Poll::Ready(Some(ConnectionMonitorOutput::Spammy))
         );
 
         // Recording less than the expected number of messages (i.e. no message),
         // expecting the peer to be unhealthy
         assert_eq!(
             monitor.poll(&mut Context::from_waker(&noop_waker())),
-            Poll::Ready(ConnectionMonitorOutput::Unhealthy)
+            Poll::Ready(Some(ConnectionMonitorOutput::Unhealthy))
         );
 
         // Recording the right number of messages, marks the peer as healthy again.
         monitor.record_message();
         assert_eq!(
             monitor.poll(&mut Context::from_waker(&noop_waker())),
-            Poll::Ready(ConnectionMonitorOutput::Healthy)
+            Poll::Ready(Some(ConnectionMonitorOutput::Healthy))
         );
     }
 
--- a/nomos-blend/network/src/handler.rs
+++ b/nomos-blend/network/src/handler.rs
@@ -163,19 +163,20 @@ where
         // TODO: Refactor this to a separate function.
         if let Poll::Ready(output) = self.monitor.poll(cx) {
             match output {
-                ConnectionMonitorOutput::Spammy => {
+                Some(ConnectionMonitorOutput::Spammy) => {
                     self.close_substreams();
                     self.pending_events_to_behaviour
                         .push_back(ToBehaviour::SpammyPeer);
                 }
-                ConnectionMonitorOutput::Unhealthy => {
+                Some(ConnectionMonitorOutput::Unhealthy) => {
                     self.pending_events_to_behaviour
                         .push_back(ToBehaviour::UnhealthyPeer);
                 }
-                ConnectionMonitorOutput::Healthy => {
+                Some(ConnectionMonitorOutput::Healthy) => {
                     self.pending_events_to_behaviour
                         .push_back(ToBehaviour::HealthyPeer);
                 }
+                None => panic!("Connection monitor stream was closed."),
             }
         }
 
--- a/nomos-blend/network/src/tokio.rs
+++ b/nomos-blend/network/src/tokio.rs
@@ -47,7 +47,10 @@ impl IntervalStreamProvider for ObservationWindowTokioIntervalProvider {
         let expected_message_range = self.calculate_expected_message_range();
         Box::new(
             tokio_stream::wrappers::IntervalStream::new(tokio::time::interval(
-                Duration::from_secs(self.round_duration_seconds.get()),
+                Duration::from_secs(
+                    (self.rounds_per_observation_window.get() as u64)
+                        * self.round_duration_seconds.get(),
+                ),
             ))
             .map(move |_| expected_message_range.clone()),
         )


nomos-blend/message/src/message.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/// A payload that is fully decapsulated.
+/// This must be encapsulated when being sent to the blend network.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct Payload {
+    pub header: PayloadHeader,
+    /// A body is padded to [`MAX_PAYLOAD_BODY_SIZE`],
+    /// Vec is used to not allocate a big array on the stack.
+    pub body: Vec<u8>,
```
ntn-x2: Could we define a `ConstVec` (or maybe there is an already existing implementation) which gives us compile-time guaranteed about its length and capacity?

nomos-blend/message/src/crypto.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// Generates a new Ed25519 private key using the [`ChaCha12Rng`].
+    #[must_use]
+    pub fn generate() -> Self {
+        Self(ed25519_dalek::SigningKey::generate(
+            &mut ChaCha12Rng::from_entropy(),
+        ))
+    }
+
+    #[must_use]
+    pub fn from_bytes(bytes: &[u8; KEY_SIZE]) -> Self {
```
ntn-x2: Should we just provide a `From<&[u8; KEY_SIZE]>` and `From<[u8; KEY_SIZE]>` implementations, so we can do `array.into()`?

nomos-blend/message/src/crypto.rs
```diff
@@ -0,0 +1,180 @@
...
+pub struct SharedKey([u8; KEY_SIZE]);
+
+impl SharedKey {
+    #[must_use]
+    pub const fn as_slice(&self) -> &[u8] {
+        &self.0
+    }
+
+    /// Encrypts data in-place by XOR operation with a pseudo-random bytes.
+    pub fn encrypt(&self, data: &mut [u8]) {
```
ntn-x2, line 116: Why did you decide to mutate the input parameter instead of returning the encrypted payload?

ntn-x2: I mean this might also make sense, like `encrypt_in_place`, but we should also provide a method for returning the result without changing the input argument, I think.

nomos-blend/message/src/crypto.rs
```diff
@@ -0,0 +1,180 @@
...
+    #[must_use]
+    pub const fn as_bytes(&self) -> &[u8; PROOF_OF_SELECTION_SIZE] {
+        &self.0
+    }
+}
+
+/// Generates pseudo-random bytes of the given size
+/// using [`ChaCha20`] cipher with a key derived from the input key.
+#[must_use]
+pub fn pseudo_random_bytes(key: &[u8], size: usize) -> Vec<u8> {
```
ntn-x2, line 187: Why not using a const generic, so we can return `[u8; EXPECTED_SIZE]` instead?

nomos-blend/message/src/encap.rs
```diff
@@ -0,0 +1,481 @@
...
+//
+// TODO: Implement [`BlendMessage`] trait for this type.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct EncapsulatedMessage {
+    /// A header that is not encapsulated.
+    header: Header,
+    /// A public header that is not encapsulated.
+    public_header: PublicHeader,
+    /// An encapsulated private header.
+    private_header: [Vec<u8>; MAX_ENCAPSULATIONS],
```
ntn-x2: Do you think we should define a `PrivateHeader` and a `Payload` type, even if they are just simple wrappers around byte arrays? It would make it harder to use one byte array for another, especially when dealing with a lot of byte arrays. I would define a different type for each of these components.

ntn-x2: Also, isn't this the maximum size of the set, but as the message gets from one hop to the other it will be reduced by one? If that's the case, I think we could use a `BoundedVec` which contains AT MOST `N` elements, or less.

nomos-blend/message/src/encap.rs
```diff
@@ -0,0 +1,481 @@
...
+
+impl EncapsulatedMessage {
+    /// Encapsulates the given payload with the provided encapsulation inputs.
+    /// Encapsulation inputs more than [`MAX_ENCAPSULATIONS`] are not allowed.
+    pub fn new(
+        inputs: &[EncapsulationInput],
+        payload_type: PayloadType,
+        payload_body: &[u8],
+    ) -> Result<Self, Error> {
+        if inputs.len() > MAX_ENCAPSULATIONS {
```
ntn-x2: Do you think we could encode all these constraints into its own type, e.g., `RangeVec<const MIN: usize, const MAX: usize>`? Or something like that? Might be out of scope for this PR but wanted to have your opinion on this.

nomos-blend/message/src/encap.rs
```diff
@@ -0,0 +1,481 @@
...
+            signing_pubkey: signing_key.public_key(),
+            proof_of_quota: proof_of_quota.clone(),
+            signature: sig,
+        };
+
+        Ok(msg)
+    }
+
+    /// An initialization step for preparing encapsulations.
+    fn initialize(
```
ntn-x2, line 131: I think this would be a perfect fit for the builder/typestate pattern. Each step of the encapsulation has a different type, on which you can call a specific set of functions to get something else in return. So here we can have an uninitialised and an initialised message type, which contain different data. These are nits, especially given this is early code and the time constraints, but wanted to raise my point.

nomos-blend/message/src/encap.rs
```diff
@@ -0,0 +1,481 @@
...
+
+        // Zero out the public header.
+        let public_header = Self::zeroized_public_header();
+
+        // Randomize the private header.
+        // The last `inputs.len()` blending headers should be randomized with
+        // resonstructed data. BlendingHeaders[0]: randomized
+        // BlendingHeaders[1]: pseudo-randomized with inputs[1]
+        // BlendingHeaders[2]: pseudo-randomized with inputs[0]
+        assert!(inputs.len() <= MAX_ENCAPSULATIONS);
```
ntn-x2: This should be an error rather than a panic?

nomos-blend/message/src/encap.rs
```diff
@@ -0,0 +1,481 @@
...
+            .map(Some)
+            .chain(repeat_n(None, MAX_ENCAPSULATIONS - inputs.len()))
+            .rev()
+            .map(|input| {
+                input.map_or_else(Self::random_blending_header, |input| {
+                    Self::pseudo_random_blending_header(input.shared_key.as_slice())
+                })
+            })
+            .map(|blending_header| {
+                wire::serialize(&blending_header).expect("BlendingHeader should be serialized")
```
ntn-x2: If we define our own blending header type, this would not be done here anymore, which I think would be a better design.

nomos-blend/message/src/encap.rs
```diff
@@ -0,0 +1,481 @@
...
+                body_len: payload_body
+                    .len()
+                    .try_into()
+                    .map_err(|_| Error::InvalidPayloadLength)?,
+            },
+            body: vec![0u8; MAX_PAYLOAD_BODY_SIZE],
+        };
+        assert!(payload_body.len() <= payload.body.len());
+        payload.body[..payload_body.len()].copy_from_slice(payload_body);
+        let payload = wire::serialize(&payload).expect("Payload should be serialized");
```
ntn-x2: Same for the payload.

nomos-blend/message/src/encap.rs
youngjoon-lee: > Do you think we should define a `PrivateHeader` and a `Payload` type, even if they are just simple wrappers around byte arrays? It would make it harder to use one byte array for another, especially when dealing with a lot of byte arrays. I would define a different type for each of these components.

Makes sense. I will define structs soon. I actually thought about it but forgot to do it. Thank you.

youngjoon-lee: > Also, isn't this the maximum size of the set, but as the message gets from one hop to the other it will be reduced by one? If that's the case, I think we could use a `BoundedVec` which contains AT MOST `N` elements, or less.

One interesting/important property of the spec is that total message length is maintained even after decapsulating one layer. In the decapsulation process, we decrypt this array, shift items to the left, and fill a pseudo-random item at the end. This prevents the next blend node from investigating how many hops remain. So, the length of private header must be preserved.

nomos-blend/message/src/crypto.rs
youngjoon-lee: I see your point. So far, this in-place version can handle all the cases in our codebase. What do you think about adding the other version when we need it? Considering data size to be encrypted, I think the in-place version is more efficient.

nomos-blend/message/src/encap.rs
youngjoon-lee: Does it mean we'd be better to have `BlendingHeader.serialize()` function? probably

youngjoon-lee: Ah I see your point. Yes. I think we can have something like `BlendingHeader.encapsulate(...) -> EncapsulatedBlendingHeader`. This is related to https://github.com/logos-co/nomos/pull/1371#discussion_r2154068431. Let me try

nomos-blend/message/src/message.rs
youngjoon-lee: I guess we can just use `Box<[u8; MAX_PAYLOAD_BODY_SIZE]>` since it's fixed-size. Let me try.

ntn-x2: Oh that's a nice idea indeed.

nomos-blend/message/src/crypto.rs
ntn-x2: Yes it's most likely very efficient, but I am thinking about cases where you might want to maintain the plaintext, which would now require a `clone` before. But as you said, if this is enough, let's push it for a later time.

nomos-blend/message/src/encap.rs
ntn-x2: Yup, that's the part of the spec I did not get to. If the set is constant size, then this makes sense indeed.

ntn-x2: Although I would call it something else than `MAX_ENCAPSULATIONS`, like simply `ENCAPSULATIONS_COUNT` or something like that.

nomos-blend/message/src/encap.rs
ntn-x2: Yeah basically de-coupling the serialization/deserialization logic from the message construction part, indeed. Better isolation. So we build the message and then serialize it, and not that we serialize parts of it, while the rest is serialized later on.

nomos-blend/message/src/message.rs
```diff
@@ -0,0 +1,153 @@
...
+    pub const fn payload_type(&self) -> PayloadType {
+        self.header.payload_type
+    }
+
+    /// Returns the payload body unpadded.
+    /// Returns an error if the payload cannot be read up to the length
+    /// specified in the header
+    pub fn body(&self) -> Result<&[u8], Error> {
+        let len = self.header.body_len as usize;
+        if self.body.len() < len {
```
ntn-x2, line 136: Is this check really necessary? Nobody can mess around with the internal structure of the payload, so how would this ever happen?

nomos-blend/message/src/message.rs
```diff
@@ -0,0 +1,153 @@
...
+
+/// A blending header that is fully decapsulated.
+/// This must be encapsulated when being sent to the blend network.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct BlendingHeader {
+    pub signing_pubkey: Ed25519PublicKey,
+    pub proof_of_quota: ProofOfQuota,
+    pub signature: Signature,
+    pub proof_of_selection: ProofOfSelection,
+    pub is_last: bool,
```
ntn-x2, line 52: This is not included in the pseudo-code. Where was this specified?

nomos-blend/message/src/crypto.rs
ntn-x2: I guess this is still to be addressed?

nomos-blend/message/src/encap.rs
```diff
@@ -0,0 +1,543 @@
...
+    error::Error,
+    input::EncapsulationInputs,
+    message::{BlendingHeader, Header, Payload, PayloadType, PublicHeader},
+};
+
+/// An encapsulated message that is sent to the blend network.
+//
+// TODO: Implement [`BlendMessage`] trait for this type.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct EncapsulatedMessage<const MAX_ENCAPSULATIONS: usize> {
```
ntn-x2: We said this naming is misleading, right? It's not max but the actual amount, no?

nomos-blend/message/src/encap.rs
```diff
@@ -0,0 +1,543 @@
...
+
+        // Construct the public header.
+        let public_header = PublicHeader {
+            signing_pubkey: signing_key.public_key(),
+            proof_of_quota,
+            signature: part.sign(&signing_key),
+        };
+
+        Ok(Self {
+            header: Header::new(),
```
ntn-x2, line 70: Why not letting specifying the header as well?]

nomos-blend/message/src/encap.rs
```diff
@@ -0,0 +1,543 @@
...
+        payload_body: &[u8],
+    ) -> Result<Self, Error> {
+        Ok(Self {
+            private_header: EncapsulatedPrivateHeader::initialize(inputs),
+            payload: EncapsulatedPayload::initialize(&Payload::new(payload_type, payload_body)?),
+        })
+    }
+
+    /// Add a layer of encapsulation.
+    fn encapsulate(
```
ntn-x2, line 143: Would it make sense to return a different type here, on which we can then call `decapsulate`? Else now we could call `EncapsulatedPart::initialize` and then call `.decapsulate` without first calling `.encapsulate`.

nomos-blend/message/src/encap.rs
```diff
@@ -0,0 +1,543 @@
...
+    /// Encapsulates the private header.
+    fn encapsulate(
+        mut self,
+        shared_key: &SharedKey,
+        signing_pubkey: Ed25519PublicKey,
+        proof_of_quota: ProofOfQuota,
+        signature: Signature,
+        proof_of_selection: ProofOfSelection,
+        is_last: bool,
+    ) -> Self {
```
ntn-x2, line 284: Same point as above, about a different return type, if feasible.

nomos-blend/message/src/encap.rs
```diff
@@ -0,0 +1,543 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct EncapsulatedPayload(Vec<u8>);
+
+impl EncapsulatedPayload {
+    fn initialize(payload: &Payload) -> Self {
+        Self(wire::serialize(payload).expect("Payload should be able to be serialized"))
+    }
+
+    fn try_deserialize(&self) -> Result<Payload, Error> {
```
ntn-x2, line 443: Are we sure we want to expose this method and also have a `#[derive(Serialize, Deserialize)]`?

ntn-x2: Maybe we should simply implement `Deserialize` ourselves?

ntn-x2 (commented):
Looks much cleaner and easier to follow, thank you :) I left some more minor comments, but overall it looks very nice.

nomos-blend/message/src/message.rs
youngjoon-lee: I thought so, but I realized that someone in the network can manipulate the encrypted payload bytes maliciously. Then, even if the last blend node decrypts the payload (including header), he will get a wrong number in the `self.header.body_len`. Actually, this scenario should be protected by signature verification (which is TODO yet).

But, another malicious scenario is that, the original message builder (the block proposer) may fill the `body_len` incorrectly (whether intentional or not). So, I think it's safer to check if `body_len` is correct because slicing the `self.body`.

nomos-blend/message/src/message.rs
youngjoon-lee: Oh, this has been added very recently in https://www.notion.so/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#216261aa09df806cb21ef93f2db88615. I didn't realize that this is not yet included in the [Message Formatting Spec](https://www.notion.so/Message-Formatting-Specification-215261aa09df81c79e3acd9e921bcc30?source=copy_link#215261aa09df81e09955f7d3bd08aa28). I will add it since @madxor is OOO.

youngjoon-lee: Ah, I can't since the doc is locked. I will ask Marcin to update it.

nomos-blend/message/src/encap.rs
youngjoon-lee: This module implements the version 1 spec itself. So, I thought we don't need to let users specify the version. When we add more fields in the `Header`, I think we can expose the `Header::new()` function as public. Wdyt?

nomos-blend/message/src/encap.rs
youngjoon-lee: I added this method because a deserialization can fail if the original message sender built the message incorrectly (whether intentional or not). 
I guess, you're asking about why we should have this function even though the `Deserialize` is derived for `Payload`. The reason is that I'm using `wire` when serializing `Payload` into `EncapsulatedPayload`. So, I wanted to have the symmetric method to deserialize it using `wire`. Does this make sense?

nomos-blend/message/src/crypto.rs
youngjoon-lee: Yeah, I actually added [pseudo_random_**sized**_bytes](https://github.com/logos-co/nomos/blob/bfcbee9bd4a2da7ed0e059ef361e52af980b886d/nomos-blend/message/src/crypto.rs#L178) to use in https://github.com/logos-co/nomos/blob/bfcbee9bd4a2da7ed0e059ef361e52af980b886d/nomos-blend/message/src/message.rs#L59.

However, I realized that we still need the dynamic sized version when encrypting [blending headers](https://github.com/logos-co/nomos/blob/bfcbee9bd4a2da7ed0e059ef361e52af980b886d/nomos-blend/message/src/encap.rs#L403) and [payloads](https://github.com/logos-co/nomos/blob/bfcbee9bd4a2da7ed0e059ef361e52af980b886d/nomos-blend/message/src/encap.rs#L428), because the size of their serialized bytes depends on the serde backend (`wire` in our case), which is not known at compile-time.

nomos-blend/message/src/encap.rs
youngjoon-lee: Modified in 05a6c647

I thought `MAX_ENCAPSULATIONS` is appropriate, but now I agree that `ENCAPSULATION_COUNT` is correct, because messages are encapsulated the `ENCAPSULATION_COUNT` number of times no matter how many layers users specify (as long as it's not greater than `ENCAPSULATION_COUNT`).

nomos-blend/message/src/encap.rs
youngjoon-lee: I tried it in different ways, but I haven't been satisfied with any of them. It was not as trivial as I expected because the initialization step includes encapsulations. https://github.com/logos-co/nomos/blob/05a6c6477257b34abb391a4dcb11e56b0a45424f/nomos-blend/message/src/encap.rs#L259

Also, when decapsulating messages, we never know whether the result is fully decapsulated until we try deserialization and check a flag.

So, I defined the `Encapsulated*` structs as "zero or multiple encapsulated", but I should have left comments. And, I agree that this is not a perfect design.
Let me try it again with a fresh mind. Or, it'd be also good to do it as a next PR, to not make reviews of this PR too long.

nomos-blend/message/src/encap.rs
youngjoon-lee: I guess your point is also related to https://github.com/logos-co/nomos/pull/1371#discussion_r2159072418?

nomos-blend/message/src/encap.rs
youngjoon-lee: I gave other tries but couldn't find any way cleaner than the current one, because of the reason I mentioned above. Instead, I added comments in the code: a7df1344. IMO, it's fine to go with the current one. Wdyt?

youngjoon-lee: I also added [578bee8](https://github.com/logos-co/nomos/pull/1371/commits/578bee861f6146d707f1ccae9133e65462335663) comments for better typestate pattern. I will address this later but it might require polishing the spec. For now, I'd like to go as similar as how the spec is defined.

nomos-blend/message/src/message.rs
ntn-x2: Yes, makes sense.

nomos-blend/message/src/encap.rs
ntn-x2: Sounds good. Perhaps a comment for that would also help, since we are not calling the message type `EncapsulatedMessageV1`.

nomos-blend/message/src/encap.rs
ntn-x2: Ah ok I see that `try_deserialize` returns `BlendingHeader` and not `EncapsulatedBlendingHeader`. Ok I think this is good enough, sorry for dragging this for this long. A comment would greatly help, and maybe a name change for the function, since it's not easy to see when to use `EncapsulatedBlendingHeader::deserialize` vs `EncapsulatedBlendingHeader::try_deserialize`.

nomos-blend/message/src/encap.rs
```diff
@@ -0,0 +1,565 @@
...
+    /// Decapsulate a layer.
+    fn decapsulate(
+        self,
+        key: &SharedKey,
+    ) -> Result<PartDecapsulationOutput<ENCAPSULATION_COUNT>, Error> {
+        match self.private_header.decapsulate(key)? {
+            PrivateHeaderDecapsulationOutput::Incompleted((private_header, public_header)) => {
+                let payload = self.payload.decapsulate(key);
+                // TODO: public_header.signature.verify(private_header|payload)
+                // The spec should be clarified on this.
```
danielSanchezQ, line 182: This should be bubbled up to research, not left as a TODO imo. Also it should be mentioned in the PR description that this is needed.

youngjoon-lee: It is recently added to the spec since I informed it to Marcin, but I didn't add it in this PR since the reviews were under going. I will add it in the next PR.

nomos-blend/message/src/encap.rs
youngjoon-lee: Yes. Done in 26c3e9bb

nomos-blend/message/src/encap.rs
youngjoon-lee: Yes. I agree. I had the same thought when I wrote this code. I added comments in [a7df134](https://github.com/logos-co/nomos/pull/1371/commits/a7df1344f4fe6b2cb3241dae742437e7550430ab), but please let me know if this is not informative. 
As discussed in https://github.com/logos-co/nomos/pull/1371#discussion_r2159072418, this could be refactored better. I will try it again as an upcoming PR.

nomos-blend/message/src/encap.rs
youngjoon-lee: I implemented this in https://github.com/logos-co/nomos/pull/1386

PR #1375, opened 2025-06-18
feat(sync): add request tip support to cryptarchia service

andrussal:

This PR updates `nomos-services/cryptarchia-consensus`  to handle get tip request for bootstrapping. 
- Add support to call `request_tip` on NetworkAdapter
- Add support to subscribe to `ChainSyncEvent`s on NetworkAdapter
- Cryptarchia service listens and responds to incoming `ProvideTipRequest`s
- (Also removed SerialisedHeaderId as HeaderId can be easily used directly)

Note: this should be merged to master but for review I set target to `feat/bootstrapping-networking-service`. 

@andrussal 

N/A

N/A

## Checklist

* [x] 1. Description added.

### DIFF
--- a/consensus/cryptarchia-sync/src/provider.rs
+++ b/consensus/cryptarchia-sync/src/provider.rs
@@ -59,16 +59,18 @@ impl Provider {
         })?;
 
         stream
+            .map_err(|e| ChainSyncError {
+                peer: peer_id,
+                kind: ChainSyncErrorKind::ReceivingBlocksError(format!(
+                    "Failed to receive block from stream: {e}"
+                )),
+            })
             .try_fold(&mut libp2p_stream, |stream, block| async move {
                 let message = DownloadBlocksResponse::Block(block);
                 send_message(peer_id, stream, &message).await?;
                 Ok(stream)
             })
-            .await
-            .map_err(|e| ChainSyncError {
-                peer: peer_id,
-                kind: ChainSyncErrorKind::ChannelReceiveError(e.to_string()),
-            })?;
+            .await?;
 
         let request = DownloadBlocksResponse::NoMoreBlocks;
         send_message(peer_id, &mut libp2p_stream, &request).await?;
--- a/consensus/cryptarchia-sync/src/errors.rs
+++ b/consensus/cryptarchia-sync/src/errors.rs
@@ -27,6 +27,9 @@ pub enum ChainSyncErrorKind {
 
     #[error("Failed to send data to channel: {0}")]
     ChannelSendError(String),
+
+    #[error("Service error: {0}")]
+    ReceivingBlocksError(String),
 }
 
 #[derive(Debug, Error, Clone)]


nomos-services/cryptarchia-consensus/src/network/adapters/libp2p.rs
```diff
@@ -113,4 +114,40 @@ where
...
+            .send(NetworkMsg::Process(Command::ChainSync(
+                ChainSyncCommand::RequestTip { peer, reply_sender },
+            )))
+            .await
+        {
+            return Err(Box::new(e));
+        }
+
+        let result = receiver.await.map_err(|e| Box::new(e) as DynError)?;
+        result.map_err(|e| Box::new(e) as DynError)
```
zeegomo, line 151: why is this here? shouldn't be in cryptarchia-sync?

andrussal: This function will trigger `GetTip` request in `cryptarchia-sync`(`cryptarchia-sync` is meant to be only low level networking). It should be used by a Bootstrap `Requestor` that Youngjoon is going to implement in cryptarchia service.

andrussal: In an earlier PR we had discussion how we request block downloads from a peer and it was decided that first we request a peer tip to fill in the `target_block` .

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -690,6 +691,15 @@ where
                     Some(msg) = self.service_resources_handle.inbound_relay.next() => {
                         Self::process_message(&cryptarchia, &self.block_subscription_sender, msg);
                     }
+
+                    Some(event) = chainsync_events.next() => {
+                        if let ChainSyncEvent::ProvideTipRequest{reply_sender} = event {
```
ntn-x2, line 696: I think you can add this check above, like `Some(ChainSyncEvent::ProvideTipRequest { reply_sender }) = chainsync_events.next() => ...`

consensus/cryptarchia-sync/src/behaviour.rs
```diff
@@ -106,19 +106,11 @@ pub enum Event {
...
-    },
-    DownloadBlocksResponse {
-        /// The response containing a block or an error.
-        result: Result<SerialisedBlock, ChainSyncError>,
-    },
-    GetTipResponse {
-        /// The response containing the latest tip or an error.
-        result: Result<SerialisedHeaderId, ChainSyncError>,
+        reply_sender: Sender<HeaderId>,
     },
```
youngjoon-lee, line 114: Could you hint me why you're removing these types?

consensus/cryptarchia-sync/src/provider.rs
```diff
@@ -67,7 +64,11 @@ impl Provider {
                 send_message(peer_id, stream, &message).await?;
                 Ok(stream)
             })
-            .await?;
+            .await
+            .map_err(|e| ChainSyncError {
+                peer: peer_id,
+                kind: ChainSyncErrorKind::ServiceError(format!("Failed to receive blocks: {e}")),
+            })?;
```
youngjoon-lee: If an error was returned from `send_message` (not from reading the `stream`), we may need to map the error differently with a different error message. Maybe we need to use `ChainSyncErrorKind::ChannelSendError` error for that case?

Also I feel we may be able to find a better name than `ServiceError` because this networking code doesn't have any concept about what 'service' means. But I don't have any idea yet.

nomos-services/cryptarchia-consensus/src/network/adapters/libp2p.rs
```diff
@@ -113,4 +114,40 @@ where
             }
         })))
     }
+
+    async fn chainsync_stream(&self) -> Result<BoxedStream<ChainSyncEvent>, DynError> {
```
youngjoon-lee: nit: I think `chainsync_event_stream` is a better name. `chainsync_stream` is too generic.

nomos-services/network/src/backends/libp2p/swarm/chainsync.rs
```diff
@@ -83,8 +83,33 @@ impl SwarmHandler {
...
+                latest_immutable_block,
+                additional_blocks,
+                reply_sender,
+            },
+            cryptarchia_sync::Event::ProvideTipsRequest { reply_sender } => {
+                Self::ProvideTipRequest { reply_sender }
+            }
+        }
+    }
+}
```
youngjoon-lee, line 115: Now that you removed some types in the `cryptarchia_sync::Event`, the `ChainSyncEvent` is identical to `cryptarchia_sync::Event`. Then, what do you think about just having only one of them? We had a similar discussion some time ago, but now I think we can re-consider. Maybe I'm missing something (related to my above comment https://github.com/logos-co/nomos/pull/1375#discussion_r2155837598).

consensus/cryptarchia-sync/src/behaviour.rs
andrussal: I didn't realise that I don't need them earlier(before implementing service part). `Event` enum only needs variants for data that is going from networking to service.

nomos-services/network/src/backends/libp2p/swarm/chainsync.rs
andrussal: This is something I added later when `cargo hack` failed in CI. I couldn't use `libp2p` types directly in service code.

consensus/cryptarchia-sync/src/provider.rs
andrussal: Here inner error comes from reading the stream, just double checked. I renamed `ServiceError` to `ReceivingBlocksError`.  Let me know what you think of this naming.

youngjoon-lee: > Here inner error comes from reading the stream, just double checked.

I mean, the error from `send_message` is already `ChainSyncError`. 
https://github.com/logos-co/nomos/blob/a466f5a5c3383ca04ea07d6c93cecfaacda8bc87/consensus/cryptarchia-sync/src/provider.rs#L64

The doc of [`try_fold`](https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#method.try_fold) says that:
> This method is similar to fold, but will exit early if an error is encountered in either the stream or the provided closure.

So, I think we are converting any `ChainSyncError` (returned from `send_message`) into another `ChainSyncError` (which contains `ReceivingBlocksError`) in the `map_err`.

https://github.com/logos-co/nomos/blob/a466f5a5c3383ca04ea07d6c93cecfaacda8bc87/consensus/cryptarchia-sync/src/provider.rs#L68

If you think we must deal with any error as `ReceivingBlocksError`, I'm fine with it.
If not, I think we need to do some pattern matching in the `map_err`, but yeah, it's not clean because the error returned from `try_fold` is ` Box<dyn Error + Send + Sync>`.

youngjoon-lee: I think we can do like this.

```rust
        stream
            .map_err(|e| ChainSyncError {
                peer: peer_id,
                kind: ChainSyncErrorKind::ReceivingBlocksError(format!(
                    "Failed to receive block from stream: {e}"
                )),
            })
            .try_fold(&mut libp2p_stream, |stream, block| async move {
                let message = DownloadBlocksResponse::Block(block);
                send_message(peer_id, stream, &message).await?;
                Ok(stream)
            })
            .await?;
```
Converting any error from reading the `stream` into `ChainSyncError { ..., ReceivingBlocksError }`, and just bubbling up any error from `send_message` as it is.

andrussal: Thank you! This is definitely the correct way.

youngjoon-lee: > Thank you! This is definitely the correct way.

I just checked [4935ac7](https://github.com/logos-co/nomos/pull/1375/commits/4935ac79f1b0811e948dd6b8b2e6972914390af4) but this is not what i suggested. This commit actually change nothing other than naming

youngjoon-lee: In my proposal, I call map_err before try_fold

zeegomo (commented):
Looks good on a first pass, I'll leave this to @youngjoon-lee

consensus/cryptarchia-sync/src/provider.rs
andrussal: I had your version but in my codebase but didn't commit it. It's now pushed: https://github.com/logos-co/nomos/commit/5dd97f0b6f285eae4dd734c5da6f313b8db733c5

youngjoon-lee: thank you!

PR #1377, opened 2025-06-18
feat(node): Role based service selection

strinnityk:
Implements role based service selection when booting up the node.

As derived from SDP [specification](https://www.notion.so/Service-Declaration-Protocol-Specification-1fd261aa09df819ca9f8eb2bdfd4ec1d).

@strinnityk

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.


.DS_Store
ntn-x2, line 1: Ops!

nodes/nomos-node/node/src/main.rs
```diff
@@ -1,69 +1,32 @@
...
+
+    let config =
+        serde_yaml::from_reader::<_, Config>(std::fs::File::open(cli_args.config_path())?)?
+            .update_from_args(cli_args)?;
 
     #[expect(
         clippy::non_ascii_literal,
         reason = "Use of green checkmark for better UX."
     )]
-    if check_config_only {
```
ntn-x2, line 66: I think this slipped through in my rebase. I don't see any need to change the naming of the variable.

nodes/nomos-node/node/src/main.rs
```diff
@@ -111,7 +74,41 @@ async fn main() -> Result<()> {
...
     app.wait_finished().await;
     Ok(())
 }
+
+async fn get_services_to_start(
+    app: &Overwatch<RuntimeServiceId>,
+    must_blend_service_group_start: bool,
+    must_da_service_group_start: bool,
+) -> Result<Vec<RuntimeServiceId>, OverwatchError> {
+    let mut service_ids = app.handle().retrieve_service_ids().await?;
```
ntn-x2, line 90: I'm thinking, would it make sense for a service to declare the service group they belong to, maybe in their `ServiceData` implementation? Something like `const SERVICE_GROUP: Option<&'static str>`? I am wondering whether the concept of a service belonging to a group should be in the node or in the service itself. What do you think? Since services depend on each other, I think there is an acyclic path between them, so the service that does not depend on anything might want to export its service group, and other services might simply refer to that to avoid one service changing its service group while the others don't. Hope I made it clear enough what I am trying to say...

danielSanchezQ: > I'm thinking, would it make sense for a service to declare the service group they belong to, maybe in their `ServiceData` implementation? Something like `const SERVICE_GROUP: Option<&'static str>`? I am wondering whether the concept of a service belonging to a group should be in the node or in the service itself. What do you think? Since services depend on each other, I think there is an acyclic path between them, so the service that does not depend on anything might want to export its service group, and other services might simply refer to that to avoid one service changing its service group while the others don't. Hope I made it clear enough what I am trying to say...

I dont think so. I think you are probably mixing 2 thing here. One is grouping and the other are **inter-services dependencies**. Grouping is not trying to solve dependencies itself. Although what you comment makes sense, I do not think we can solve it with the proposed solution neither. Reason is that a service do not know who is gonna consume him. By binding himself to a subgroup its forcing consumer services to activate/deactivate the whole subgroup (?), or be a member of that subgroup to function. What then if a services depends on 2 services with no dependencies that are in different groups?

TLDR: Manually handling this in the binary should be enough.

nodes/nomos-node/node/src/main.rs
strinnityk: What should've been the name?

ntn-x2: This was renamed to `dry_run` in this PR, but on `master` it's `check_config_only`. Up to you whatever name you prefer, but I was pointing out this PR changes it, not intentionally (on my side).

nodes/nomos-node/node/src/main.rs
ntn-x2: Mhh yeah, then we would need a service to declare groups instead of a single service identifier, and the node would pick services that have the specified ID in their declared group. In that way, services have ability to "depend" on another service if they are meant to be used together (which is an even stronger guarantee than the `AsServiceId` bound, which requires the two services are simply included in the same runtime). A service will never need to know, even in the previous suggestion, who is gonna consume him, but the opposite is true: a service could include the service ID of another service to make sure that they are both started together.

Now that I am thinking about it, we already introduced inter-service dependencies into each service by making a service wait for another to be ready before doing stuff. Letting the "consumer" service specify that it belong to the same group as the consumed one, would avoid the situation in which the dependency is there, but for some reason the node does not start the service that the consumer depends upon, leading to a duplication in the logic, which can potentially go out of sync for a large number of services.

I am not saying we have to do this, but I am saying that we need to pay particular attention to this, although, as far as I understand, the node CLI flag will be replaced by the SDP protocol, so each service will know when to start and when to stop, and they assume the services they depend upon will do the same.

So TLDR: this will be replaced soon, so probably no need to pay too much attention, but it was a very interesting discussion.

danielSanchezQ: > I am not saying we have to do this, but I am saying that we need to pay particular attention to this, although, as far as I understand, the node CLI flag will be replaced by the SDP protocol, so each service will know when to start and when to stop, and they assume the services they depend upon will do the same.

:100: agree on this. 
This is logic that will have to deal when actually integrating those changes, let's see how it goes.

nodes/nomos-node/node/src/main.rs
```diff
@@ -105,7 +68,41 @@ async fn main() -> Result<()> {
...
+
+    if !must_blend_service_group_start {
+        service_ids.retain(|value| value != &RuntimeServiceId::Blend);
+    }
+
+    if !must_da_service_group_start {
+        let da_service_ids = [
```
ntn-x2, line 97: HashSet rather? I know it's small values, but still... 😊

danielSanchezQ: I think this is explicit enough, otherwise we have to collect into the hashset just to iterate again. But if services groups gets more complicated it may be a good point indeed.

danielSanchezQ (approved):

PR #1378, opened 2025-06-18
Fix da dispersal executor membership change

bacv:

Dynamic membership requires dispersal behaviour in execuror and validator to be ready to accept or open dispersal protocol streams.

Dynamic membership in https://github.com/logos-co/nomos/pull/1368 requires to change our assumptions about dispersal behaviours. [Spec](https://www.notion.so/NomosDA-Network-Specification-1fd261aa09df81188e76cb083791252d)

@pradovic @bacv @danielSanchezQ 

Yes

Specification doesn't have a strict requirement for not accepting connections from membership peers if initially they act as validator but later as executor. [Spec section](https://www.notion.so/NomosDA-Network-Specification-1fd261aa09df81188e76cb083791252d?source=copy_link#1fd261aa09df81a2ab77d8df15d73926)

## Checklist

* [x] 2. Description added.


nomos-da/network/core/src/protocols/dispersal/validator/behaviour.rs
```diff
@@ -155,13 +151,29 @@ impl<M: MembershipHandler<Id = PeerId, NetworkId = SubnetworkId> + 'static> Netw
...
+        addr: &Multiaddr,
+        role_override: Endpoint,
+        port_use: PortUse,
     ) -> Result<THandler<Self>, ConnectionDenied> {
-        Ok(Either::Right(libp2p::swarm::dummy::ConnectionHandler))
+        // Sampling or replication behaviour might open connection to a member peer.
+        // During the lifetime of a connection the remote peer might decide to
+        // disperse data via existing connection - in such case the connection
+        // needs to already have a handler that accepts DA_DISPERSAL_PROTOCOL
+        // messages.
```
danielSanchezQ, line 164: I do not get the extent of this PR. Are we trying to reuse connections from sampling or replication for dispersal?

nomos-da/network/core/src/protocols/dispersal/validator/behaviour.rs
```diff
@@ -155,13 +151,29 @@ impl<M: MembershipHandler<Id = PeerId, NetworkId = SubnetworkId> + 'static> Netw
...
+        }
+        self.stream_behaviour
+            .handle_established_outbound_connection(
+                connection_id,
+                peer,
+                addr,
+                role_override,
+                port_use,
+            )
+            .map(Either::Left)
```
danielSanchezQ, line 176: I belive this is wrong behaviour. Validators dispersal behaviours should not stablish any outbound connections at all.

bacv: Validator behaviour doesn't establish any outbound behaviours, but any other behaviour might establish outbound connection to the peer that is capable of dispersing data. If validator didn't register handler when such outbound connection happened - remote peer needs to make a new connection, even though we already have one and could reuse.

nomos-da/network/core/src/protocols/dispersal/validator/behaviour.rs
bacv: Yes, instead of having multiple connections between same peers (which would defy the purpose of DA protocol), we reuse the connections that happen between validators mostly because of replication (we connect between validators in a subnet)

nomos-da/network/core/src/protocols/dispersal/validator/behaviour.rs
danielSanchezQ: > Validator behaviour doesn't establish any outbound behaviours, but any other behaviour might establish outbound connection to the peer that is capable of dispersing data. If validator didn't register handler when such outbound connection happened - remote peer needs to make a new connection, even though we already have one and could reuse.

Are you sure that is the case? I don't think so. Remote peer would request a new **stream** and if the already have a connection the connection will be used. If they do not have a connection then it will be dialed. Its the whole premisses on how the stack works. Otherwise...well, we are not in a good place.

danielSanchezQ: After sync, this is the correct behaviour. Thanks for the explanation.

pradovic (approved):

PR #1380, opened 2025-06-20
feat(PoLv2): Partial update to PoLv2

zeegomo:
This commit paves the ground for PoLv2 by changing the public and private inputs to the proof and related traits. It's not a complete implementation and does not attempt to build a valid proof as that would require the new mantle note format and we're going to ditch risc0 anyway. Instead, the purpose of this PR is to split the work needed for the new mantle and enable the note changes.

https://www.notion.so/Proof-of-Leadership-Specification-MIGRATED-1438f96fb65c80f6bb1adfeb73473f65

@thomaslavaur is the original author and I'm in charge of this pr

 Yes, https://www.notion.so/Proof-of-Leadership-Specification-MIGRATED-1438f96fb65c80f6bb1adfeb73473f65

This is only a partial implementation of the spec, a throughout check will be conducted on the final impl

## Checklist

* [x] 2. Description added.

### DIFF
--- a/ledger/nomos-ledger/src/lib.rs
+++ b/ledger/nomos-ledger/src/lib.rs
@@ -278,8 +278,8 @@ impl LedgerState {
     {
         assert_eq!(config.epoch(slot), self.epoch_state.epoch);
         let public_inputs = LeaderPublic::new(
-            self.epoch_state.commitments.root(),
-            self.commitments.root(),
+            self.aged_commitments().root(),
+            self.latest_commitments().root(),
             proof.entropy(),
             self.epoch_state.nonce,
             slot.into(),
@@ -317,7 +317,7 @@ impl LedgerState {
     fn update_nonce(self, contrib: [u8; 32], slot: Slot) -> Self {
         Self {
             nonce: <[u8; 32]>::from(
-                Blake2b::new_with_prefix(b"epoch-nonce")
+                Blake2b::new_with_prefix(b"EPOCH_NONCE")
                     .chain_update(self.nonce)
                     .chain_update(contrib)
                     .chain_update(slot.to_be_bytes())
--- a/nomos-services/cryptarchia-consensus/src/leadership.rs
+++ b/nomos-services/cryptarchia-consensus/src/leadership.rs
@@ -5,7 +5,7 @@ use nomos_ledger::{EpochState, NoteTree};
 use nomos_proof_statements::leadership::{LeaderPrivate, LeaderPublic};
 use serde::{Deserialize, Serialize};
 
-/// TODO: this is a temprorary solution until we have a proper wallet
+/// TODO: this is a temporary solution until we have a proper wallet
 /// implementation. Most notably, it can't track when initial notes are spent
 /// and moved
 #[derive(Clone)]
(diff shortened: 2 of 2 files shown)


nomos-services/cryptarchia-consensus/src/leadership.rs
```diff
@@ -1,20 +1,17 @@
...
 use cryptarchia_engine::Slot;
-use nomos_core::{header::HeaderId, proofs::leader_proof::Risc0LeaderProof};
+use nomos_core::proofs::leader_proof::Risc0LeaderProof;
 use nomos_ledger::{EpochState, NoteTree};
 use nomos_proof_statements::leadership::{LeaderPrivate, LeaderPublic};
 use serde::{Deserialize, Serialize};
 
+/// TODO: this is a temprorary solution until we have a proper wallet
+/// implementation. Most notably, it can't track when initial notes are spent
+/// and moved
```
danielSanchezQ: :+1:

ledger/nomos-ledger/src/lib.rs
```diff
@@ -342,47 +278,33 @@ impl LedgerState {
     {
         assert_eq!(config.epoch(slot), self.epoch_state.epoch);
         let public_inputs = LeaderPublic::new(
-            proof.merke_root(),
+            self.epoch_state.commitments.root(),
+            self.commitments.root(),
+            proof.entropy(),
```
youngjoon-lee: It'd be more readable if we use the `self.aged_commitments` function that you added below?
```suggestion
            self.aged_commitments().root(),
            self.commitments.root(),
            proof.entropy(),
```
https://github.com/logos-co/nomos/blob/a9ffd6032d3be87a5dca0416955f7df845b5695b/ledger/nomos-ledger/src/lib.rs#L376

ledger/nomos-ledger/src/lib.rs
```diff
@@ -392,12 +314,12 @@ impl LedgerState {
         self.nullifiers.contains(nullifier)
     }
 
-    fn update_nonce(self, nullifier: Nullifier, slot: Slot) -> Self {
+    fn update_nonce(self, contrib: [u8; 32], slot: Slot) -> Self {
```
danielSanchezQ, line 317: Should the `[u8; 32]` by its own type? maybe an alias?

ledger/nomos-ledger/src/lib.rs
danielSanchezQ: I changed the suggestion to the actual `self.aged_commitments()`` call, I think the `()` was missing @youngjoon-lee .

nomos-services/cryptarchia-consensus/src/leadership.rs
```diff
@@ -27,87 +24,83 @@ pub struct LeaderConfig {
...
+                }
+                (_, Some(_)) => {
+                    tracing::debug!("Found note in latest tree only: {:?}", note);
+                    continue;
+                }
+                (None, None) => {
+                    tracing::debug!("Note not found in either tree: {:?}", note);
+                    continue;
+                }
+            }
```
danielSanchezQ: Nit: This may be a function. Although having it there is immediate info, it also pollutes the loop a bit. Up to you anyway.

danielSanchezQ (approved):
Removing stuff before adding helps a lot. Thanks.

nomos-services/cryptarchia-consensus/src/leadership.rs
zeegomo: this is actually a debug leftover, I'll remove everything but the success case

ledger/nomos-ledger/src/lib.rs
zeegomo: Not sure, it's really just a bunch of bytes and we don't have a separate nonce type for instance. Not to say that it's not a good idea, but I'd like to see the full new structure before this

bacv (approved):

PR #1381, opened 2025-06-20
feat(blend): new blend message scheduler

ntn-x2:

This PR introduces the revised message scheduler to be used in the blend service. Right now, it is added as an additional module for review, but I will work on a stacked PR that removes the unnecessary modules (e.g., the temporal processor), and integrates the new, easier-to-use module inside the blend service.

I renamed the `nomos-core` create into `nomos-scheduling`, as that it what it really contains: scheduling logic.

The main elements introduced in this PR are the `MessageScheduler`, which will be the interface between the blend service and the scheduling sub-streams. It can be initialized with the required info such as maximum delay between two release windows, how many rounds are in interval, etc. It exposes two main functions to the outside: `notify_new_data_message` to inform the cover message sub-stream NOT to emit a new cover message next time because a new data message has been generated, and `schedule_message` to add a new processed message to the queue to be released when the next (randomly selected) release round ticks, as per the spec.

The two sub-stream implemented are `SessionCoverTraffic` (temporarily in a `cover_traffic_2` module, which will replace the previous cover traffic component), and `SessionReleaseDelayer` (sorry for the lack of a better name). They are only visible to the `MessageScheduler` and hence cannot be used by any other component, as it should be.

`MessageScheduler` implements the `Stream` trait, as usual, and will be polled from within the blend service, which will then forward the messages to all the peers, as described in the spec.

A follow-up PR will remove unnecessary components (i.e., everything but the cryptographic processor), and update the logic of the service to deal with the new component.

Yes.

Spec: @madxor, code owner: @youngjoon-lee, PR author: @ntn-x2.

Yes.

No.

## Checklist

* [x] 2. Description added.


nomos-blend/scheduling/src/message_scheduler/mod.rs
```diff
@@ -0,0 +1,487 @@
...
+pub mod round_info;
+pub mod session_info;
+
+const LOG_TARGET: &str = "blend::scheduling";
+
+/// A message scheduler waiting for the session stream to provide consumable
+/// session info, to compute session-related components to pass to its
+/// constituent sub-streams.
+pub struct UninitializedMessageScheduler<SessionClock, Rng> {
+    /// The random generator to select slots for message releasing.
```
youngjoon-lee:
```suggestion
    /// The random generator to select rounds for message releasing.
```

nomos-blend/scheduling/src/release_delayer.rs
```diff
@@ -0,0 +1,221 @@
...
+
+use tracing::{debug, trace};
+
+use crate::message::OutboundMessage;
+
+const LOG_TARGET: &str = "blend::scheduling::delay";
+
+/// A scheduler for delaying processed messages and batch them into release
+/// rounds, as per the specification.
+pub struct SessionReleaseDelayer<Rng> {
```
youngjoon-lee: What do you think about renaming this to `SessionProcessedMessageDelayer` if this is only for processed messages?

If you chose this generic name to cover any case where we would delay any other type of messages (as well as processed messages) in the future, I'm fine with this name.

nomos-blend/scheduling/src/message_scheduler/mod.rs
```diff
@@ -0,0 +1,487 @@
...
+            cover_traffic.current_round() == release_delayer.current_round(),
+            "The two sub-streams should never get out of sync."
+        );
+        // We poll the sub-stream and return the right result accordingly.
+        let cover_traffic_output = cover_traffic.poll_next_round().map(|()| vec![].into());
+        let release_delayer_output = release_delayer.poll_next_round();
+
+        match (cover_traffic_output, release_delayer_output) {
+            // If none of the sub-streams is ready, we do not return anything.
+            (None, None) => Poll::Pending,
```
youngjoon-lee: Here, I think we should call waker if `session_clock.poll_next_unpin` in the line 132 returned `Poll::Ready(Some)` (which means that the waker was not registered inside `session_clock.poll_next_unpin`).

nomos-blend/scheduling/src/message_scheduler/mod.rs
```diff
@@ -0,0 +1,487 @@
...
+        // We update session info on new sessions.
+        if let Poll::Ready(Some(new_session_info)) = session_clock.poll_next_unpin(cx) {
+            setup_new_session(
+                cover_traffic,
+                release_delayer,
+                round_clock,
+                settings,
+                &new_session_info,
+            );
+        }
```
youngjoon-lee: We may need to `panic!()` if `session_clock.poll_next_unpin` returns `Poll::Ready(None)`, which means that the session clock is terminated. 
If it returns `Poll::Ready(None)`, and if it is called again later, then it will anyway panic. So, if we're really sure that the `session_clock` never end, we can `panic!` proactively when it returns `Poll::Ready(None)`.

The same topic is being discussed in https://github.com/logos-co/nomos/pull/1382#discussion_r2160545653.

nomos-blend/scheduling/src/cover_traffic_2.rs
```diff
@@ -0,0 +1,216 @@
...
+    /// advancing its internal round counter. Rounds follow 0-based indexing, so
+    /// the very first round in a session is considered to be `0`.
+    ///
+    /// **This function is meant to be externally called at every round, as
+    /// failing to do so might result in missed or delayed cover messages**.
+    pub fn poll_next_round(&mut self) -> Option<()> {
+        let current_round = self.current_round;
+        // We can safely saturate, since even if we ever reach the end and there's a
+        // scheduled message, it will be consumed the first time the value is hit.
+        self.current_round = self.current_round.saturating_add(1);
```
youngjoon-lee: Could you please help me understanding what this comment means? A round is removed from `self.scheduled_message_rounds` once it is hit. For example, `self.scheduled_message_rounds.remove(usize::MAX)` is called when the round `usize::MAX` is hit. Nevertheless, this `poll_next_round` function can be called (e.g. accidentally). So, this is just a safe guard to avoid overflow?

nomos-blend/scheduling/src/release_delayer.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+}
+
+impl<Rng> SessionReleaseDelayer<Rng> {
+    /// Add a new message to the queue to be released at the next release round
+    /// along with any other queued message.
+    pub fn schedule_message(&mut self, message: OutboundMessage) {
+        self.unreleased_messages.push(message);
+    }
+}
```
youngjoon-lee: nit: maybe we could merge this impl block with the previous block? if there's any reason to split

nomos-blend/scheduling/src/message_scheduler/mod.rs
ntn-x2: But that would require the session stream to implement `FusedStream`, no? Here we are requiring just `Stream`, which means that while it can return `Ready(None)`, once, it might return `Ready(Some)` again in the future. I think it's ok to handle this case explicitly, but perhaps panicking is not the best we can do. We can simply return `Ready(None)` ourselves perhaps?

nomos-blend/scheduling/src/cover_traffic_2.rs
ntn-x2: Yeah so basically `poll_next_round` is expected to be called once per round. If this invariant is not respected, meaning the outside callers calls it more often than that, we might end up in a situation where `self.current_round` (hypothetically) quickly reaches `usize::MAX`. So in this case it's ok to saturate instead of panicking for overflow (with `.checked_add` alternative) since even if we had one release scheduled for `usize::MAX`, that would be consumed the first time we reach that value. Subsequent calls to `poll_next_round` would yield `Poll::Pending` until the scheduler itself is reset by the outside caller. In short, I explained why it's fine to saturate vs panicking.

nomos-blend/scheduling/src/release_delayer.rs
ntn-x2: Mmh the previous impl block has the `where Rng: rand::Rng` trait bound, which we don't need in this case, so I split them into two separate blocks. No need to have a random number generator if we simply want to schedule a new message.

nomos-blend/scheduling/src/message_scheduler/mod.rs
ntn-x2: Why do you think that's the case? If there is no scheduled release, we will wait for the next round clock tick to be polled again (which is done automatically) which will trigger this function.

nomos-blend/scheduling/src/message_scheduler/mod.rs
ntn-x2: https://github.com/logos-co/nomos/pull/1381/commits/a1a5fb83d2042fde6d14002a6c38903c5803157b

nomos-blend/scheduling/src/release_delayer.rs
ntn-x2: I like your suggestion! https://github.com/logos-co/nomos/pull/1381/commits/7a3fb8c58ef7dc66c07410b8a022315b394df430

nomos-blend/scheduling/src/message_scheduler/mod.rs
ntn-x2: Sorry, you are right. I had to read up a bit more on polling and awaking. Of course if we consume a round tick but return `Pending`, we need to re-register the awaker: I've done that in https://github.com/logos-co/nomos/pull/1381/commits/19923d60e338034f1213f21c9598ae18f8c0de7e. Good catch!

nomos-blend/scheduling/src/message_scheduler/mod.rs
ntn-x2: I changed it to return `Ready(None)` if either the session or the round substreams return `Ready(None)`: https://github.com/logos-co/nomos/pull/1381/commits/f289aa002cc8ea9ada677466c91bd4be1de3d000.

nomos-blend/scheduling/src/cover_traffic_2.rs
danielSanchezQ: Ok I see the issue here. Although this should be enough, as it is just consumed when the round is ticking, I would suggest to use the stream extesions pattern we used previously. You could transform this into a stream that consumes a stream that ticks. That way there is no misscall at all. Also it is automaticall reusable if instead of rounds we decide its any other tick.  wdyt?

nomos-blend/scheduling/src/message_scheduler/mod.rs
```diff
@@ -0,0 +1,498 @@
...
+
+impl<SessionClock, Rng> UninitializedMessageScheduler<SessionClock, Rng> {
+    pub const fn new(session_clock: SessionClock, settings: Settings, rng: Rng) -> Self {
+        Self {
+            rng,
+            session_clock,
+            settings,
+        }
+    }
+}
```
danielSanchezQ: Nice typing here from Uniinitialize to wait to Message scheduler!

nomos-blend/scheduling/src/message_scheduler/mod.rs
```diff
@@ -0,0 +1,498 @@
...
+
+        // Sub-streams should never get out of sync. We check that in debug builds to
+        // catch any issues early on.
+        debug_assert!(
+            cover_traffic.current_round() == release_delayer.current_round(),
+            "The two sub-streams should never get out of sync."
+        );
+        // We poll the sub-stream and return the right result accordingly.
+        let cover_traffic_output = cover_traffic.poll_next_round().map(|()| vec![].into());
+        let release_delayer_output = release_delayer.poll_next_round();
```
danielSanchezQ: Yeah, I think all this would be much clearer if we just compose streams.

nomos-blend/scheduling/src/release_delayer.rs
```diff
@@ -0,0 +1,221 @@
...
+                .expect("Overflow when scheduling next release round offset.");
+            // Return the unreleased messages and clear the buffer.
+            let messages = mem::take(&mut self.unreleased_messages);
+            debug!(target: LOG_TARGET, "Round {current_round} is a release round.");
+            return Some(messages);
+        }
+        trace!(target: LOG_TARGET, "Round {current_round} is not a release round.");
+        None
+    }
+}
```
danielSanchezQ: I'm getting confused with this methods called *poll* and not returning `Poll`. I think we should rename them so avoid this confusion. We can use `check_next_round_ready` or something like that.

nomos-blend/scheduling/src/release_delayer.rs
```diff
@@ -0,0 +1,221 @@
...
+        }
+    }
+
+    /// Poll the scheduler for a potential new release round, while also
+    /// advancing its internal round counter. Rounds follow 0-based indexing, so
+    /// the very first round in a session is considered to be `0`.
+    ///
+    /// **This function is meant to be externally called at every round, as
+    /// failing to do so might result in missed or delayed release rounds**.
+    pub fn poll_next_round(&mut self) -> Option<Vec<OutboundMessage>> {
```
danielSanchezQ: More polling naming that we should change imo.

danielSanchezQ (changes requested):
I think this could have maintained the stream composition we used to have and code would be simpler as well. Right now code depends on some calls to be done at the right time which may be tricky if this need to be change in the future. I also commented on a few nitpicks on naming.

nomos-blend/scheduling/src/message_scheduler/mod.rs
ntn-x2: Yeah, I partially agree with you. The problem with streams is that they are harder to test. When looking at each component separately, it's easy to see what they do and what state they operate on. Not having any stream, I feel, makes for code easier to understand. Also, since streams are not clonable, each would run its own clock, while in this case (on purpose), the idea was to have a single place where syncronization happen, and that would be the scheduler stream, which "polls" substreams syncronously without each having its own clock. So I personally think having each sub-stream consume its own clock would be less clear, but I can change this if you you and Youngjoon think the old format was a better choice.

nomos-blend/scheduling/src/release_delayer.rs
ntn-x2: Yes, probably the naming should change here. I agree with you.

nomos-blend/scheduling/src/cover_traffic_2.rs
ntn-x2: I explained my reasoning below: https://github.com/logos-co/nomos/pull/1381#discussion_r2161054006. The idea here is to have a single, unified "clock" which is set up by the scheduler itself, and sub-components would not need to run their own streams/clocks but would rather expose a traditional API for the scheduler to interact with. Here, I've been wanted to be extremely careful, but even if we were to use a stream, we would need to check what happens if the stream reaches its maximum value (since we need the value of the current round), in which case I think it would still be fine to saturate until the stream is reset, instead of panicking.

nomos-blend/scheduling/src/message_scheduler/mod.rs
danielSanchezQ: > When looking at each component separately, it's easy to see what they do and what state they operate on. 
It sure wasn't for me at first sight 😅 

Not having any stream, I feel, makes for code easier to understand. Also, since streams are not clonable, each would run its own clock, while in this case (on purpose), the idea was to have a single place where syncronization happen, and that would be the scheduler stream, which "polls" substreams syncronously without each having its own clock. So I personally think having each sub-stream consume its own clock would be less clear, but I can change this if you you and Youngjoon think the old format was a better choice.

I don't see it that way. The idea of the streams is that they work on top of another. Each of them doing a single tasks. If two streams need the same clocking that is a different problem for a different abstraction imo. The question in this case is who and how the clock is provided.

> Also, since streams are not clonable.

Uhm, are we sure? In this case if you think you need the upper ticking stream to be clonable we could bound it. But again it depends where does it come from. Is it another service? Because in that case it is probably be a channel wrapped in a stream (?)


TLDR: I see the point but I think is more confusing this way. Now, I may not be seeing the other point of view. So it will depend on your decision (+ @youngjoon-lee ).

ntn-x2: The session stream will come from a different service, but once that is provided, the blend service would run its own sub-streams. At the moment, different sub-streams run their own clock: message scheduling runs one, connection management runs one, but ideally there would be a single unified `round` clock that runs at the service level and is bound to all sub-streams.

danielSanchezQ: Ok, I have an hybrid approach. instead of maual polling the round clock. Just map the operation to that stream and the stream impl would be just polling that inner clock stream with the mapped operations for both outputs. wdyt? I can give it a try as it should not be complicated. And open a PR on top of this one.

nomos-blend/scheduling/src/cover_traffic_2.rs
ntn-x2: Just pointing out that even exposing a stream interface would require the scheduler to poll both sub-streams, and failing to do so would bring the overall scheduler in an inconsistent state, so that problem is not solved by using streams instead. In this case we can actually verify the states of both sub-streams, instead.

nomos-blend/scheduling/src/message_scheduler/mod.rs
ntn-x2: So letting the scheduler interact with sub-streams via the `Stream` interface instead of direct API calls?

danielSanchezQ: Yeah. 2 changes, make cover and release not following round but instead just getting the round tick itself. Wrapp them in a stream with the computation that is just polled from the wrapper `MessageScheduler`. You can check on this branch: https://github.com/logos-co/nomos/commit/d6adcbbc12f8bbdeb4206ee4babc5e89cd56f76f
⚠️ : It fucked up tests and I have a question which is what is the problem when both return empty. this is also part of the current behaviour that is inherited there. But just for showing the approach should be enough.

ntn-x2: Cool, will check it out.

As for your question, if both sub-streams return `None`, then it means it should return `Pending`. If it returns `Ready` with an empty vector and no cover message, then that is a totally valid case, and it means the clock ticked for a new release window (since cover message is `None`), but no messages were scheduled for this release window.

ntn-x2: I checked your suggestion. While I might be on board with the changes to the cover scheduler, I feel a bit less happy about the changes to the delayer component. In there, once the new delayed round is reached, the component computes the next scheduled round. Offering an API which allows to query for a random round value might mess up the state, since the component has no more control over what happens internally, so I don't think this is a good approach. I think we are left with two things: leave things as they are, where each component is still fully responsible for its own state management and offer an API which can only be misused by being called more often than it should, or switching everything to using streams, where each component would use a stream internally and implement the `Stream` trait itself: this is also not mis-usable since the component will simply keep returning `Pending` if the next item is not ready, so basically the implementation we had before. Downside is that now the two sub-components are not really in sync, as each would use its own stream.

danielSanchezQ: > In there, once the new delayed round is reached, the component computes the next scheduled round. Offering an API which allows to query for a random round value might mess up the state, since the component has no more control over what happens internally, so I don't think this is a good approach. 

But it doesn't force the update call, so you could end up with a de-sync state anyway. There is no much difference on that side as the caller still have the responsibility  :thinking: 

> I think we are left with two things: ...

Its really up to you. I will strongly suggest moving to streams even if that means that we bound it to clone so the input stream can be "safely" shared.

ntn-x2: > But it doesn't force the update call, so you could end up with a de-sync state anyway. There is no much difference on that side as the caller still have the responsibility 🤔

The caller can only increase the round counter, so the API will never result in the stored counter to go back in time. Allowing for a generic value as the function input means that once we reach round `5`, the component computes a new delayed window in range `[6, 10]`, for instance, then the caller can still call the function with values `0` to `5`, which should not be possible since those rounds have already been "consumed". While exposing a `poll_next` (or similar), internally guarantees that the next value is always greater than the last polled one. Yes, the caller can still misuse this, but it will not mess with the internal invariants of the component. It will just render the output of the component meaningful.

> Its really up to you. I will strongly suggest moving to streams even if that means that we bound it to clone so the input stream can be "safely" shared.

 I will try to investigate if I can make it work with a stream-based interface that would allow the same stream to be consumed in more than one place.

danielSanchezQ: > I will try to investigate if I can make it work with a stream-based interface that would allow the same stream to be consumed in more than one place.

Well one thing would be to kept them separated as implementation, then merged both in a single stream than extends the round stream. Similar to whats implemented here actually, but keeping them as streams (?)
Will be looking forward for some magic here :wink:

ntn-x2:
@danielSanchezQ please take a look now. I managed to implement a solution based on broadcast channels for a multi-consumer stream. I did not yet make it its own type, but I think it would be extremely helpful also in other contexts. Now both sub-streams implement the `Stream` trait, and I am quite happy with the results. @youngjoon-lee please take a second round of reviews when you can, thanks!

nomos-blend/scheduling/src/message_scheduler/mod.rs
```diff
@@ -0,0 +1,298 @@
...
+
+        // We poll the sub-stream and return the right result accordingly.
+        let cover_traffic_output = cover_traffic.poll_next_unpin(cx);
+        let release_delayer_output = release_delayer.poll_next_unpin(cx);
+
+        match (cover_traffic_output, release_delayer_output) {
+            // If none of the sub-streams is ready, we do not return anything.
+            (Poll::Pending, Poll::Pending) => {
+                // Awake to trigger a new round clock tick.
+                cx.waker().wake_by_ref();
```
danielSanchezQ, line 264: Do we really need this waker? If this waker was passed to the stream polls above then it should be awake wenever they are ready. Othewise this future is marked as reade to be polled inmedately 🤔

nomos-blend/scheduling/src/release_delayer.rs
```diff
@@ -0,0 +1,248 @@
...
+                &mut self.rng,
+            );
+            // Return the unreleased messages and clear the buffer.
+            let messages = mem::take(&mut self.unreleased_messages);
+            debug!(target: LOG_TARGET, "Round {new_round} is a release round.");
+            return Poll::Ready(Some(messages));
+        }
+        trace!(target: LOG_TARGET, "Round {new_round} is not a release round.");
+        // Awake to trigger a new round clock tick.
+        cx.waker().wake_by_ref();
```
danielSanchezQ, line 158: Another worry thing here. This will work, but it will continously mark the stream to be polled . So, when should this really be awaken, when the round clock ticks?
I think it may be ok as the second time polled will go through to the clock poll anyway.

nomos-blend/scheduling/src/message_scheduler/mod.rs
```diff
@@ -0,0 +1,298 @@
...
+            Poll::Pending => {}
+        }
+
+        // We do not return anything if a new round has not elapsed.
+        let new_round = match round_clock.poll_next_unpin(cx) {
+            Poll::Pending => return Poll::Pending,
+            Poll::Ready(None) => return Poll::Ready(None),
+            Poll::Ready(Some(new_round)) => new_round,
+        };
+        trace!(target: LOG_TARGET, "New round {new_round} started.");
```
danielSanchezQ, line 254: Is this needed if cover_traffic and release_delayer do nothing depending on the same clock?

nomos-blend/scheduling/src/message_scheduler/mod.rs
ntn-x2: We are at this point when the round clock has returned `ready`, so we have consumed that. If we don't awake the waker, we won't get the next round tick. So we should not awake it if round is pending, since the round clock will do that, but once that is ready, it won't awake the waker unless we re-register it.

So what happens here is that we awake it once, and then won't awake it anymore until the next round, since we short circuit if the round clock returns `Pending`.

nomos-blend/scheduling/src/message_scheduler/utils.rs
```diff
@@ -0,0 +1,119 @@
...
+        async move {
+            while let Some(new_round) = new_round_clock.next().await {
+                if let Err(send_error) = round_clock_stream_sender_clone.send(new_round) {
+                    error!(target: LOG_TARGET, "Failed to send round tick to consumers. Error: {send_error:?}");
+                }
+            }
+            trace!(target: LOG_TARGET, "Round clock terminated.");
+        },
+        new_abort_registration,
+    ));
```
danielSanchezQ, line 56: Well, I guess this is inevitable 😢

nomos-blend/scheduling/src/release_delayer.rs
ntn-x2: Same as above. It won't continuously awake the waker. It will do it only once after the round clock ticks, and only if we don't have anything to yield. Otherwise, we return `Ready` and the caller will trigger us again by calling `next().await`

nomos-blend/scheduling/src/message_scheduler/mod.rs
ntn-x2: Well, yes, that is a good point, but as far as this component is concerned, it does not know how the inner streams handle this case. It only cares about what they return via their `poll` interface, which we handle by also returning `Ready(None)`, if any the sub-streams returns that.

nomos-blend/scheduling/src/message_scheduler/mod.rs
danielSanchezQ: Yeah I see, we do not know if the other streams return pending because of the clock. Makes sense.

nomos-blend/scheduling/src/message_scheduler/utils.rs
ntn-x2: If we want to maintain a multi-consumer `Stream` interface, I think there is no way around this, unfortunately. At least I could not think of a better way for it, without each sub-stream have its own clock.

danielSanchezQ: This gives me an itch I cannot explain. But It will have to be. No time to scratch 😅

danielSanchezQ (approved):

nomos-blend/scheduling/src/message_scheduler/utils.rs
ntn-x2: Yeah I think it would make you and me feel better if it was hidden behind its own type, which would be clonable, and that implements `Stream` and `Drop`. I will leave it for the future if we need to use it in more places, where a single clock should be used for multiple streams.

ntn-x2:
Thanks for reviewing! I will pause this PR until the message encapsulation integration into the Blend service is merged: https://github.com/logos-co/nomos/pull/1385. It has a lot of interesting new stuff which will require small changes in this PR. Then, I can integrate this PR into the service as well. I think this is the best plan forward to avoid unnecessary merge conflicts and reworks.

PR #1382, opened 2025-06-21
fix(blend): Global observation window rounds

bacv:

@zeegomo found that proposals are not being sent through blend network. The issue was that node considered peer unhealthy after the first tick. Seems that observation duration was set to 1 round, even though expected message rate during observation window is between 6 and 60. Spec defines global observation window to 30 rounds, updated code to reflect that.

Another issue was that blend connection monitor was in busy loop and generating many log messages, when polling stream we don't need to wake wakers manually, stream itself calls it when new item is ready.

Since proposer get it's own proposal, two_nodes_happy test seemed working, giving us false positive that consensus is being achieved but unreliably. We should probably add some test to cover this case.

Feel free to update merge or close this PR, two_nodes_happy still fails, maybe we should keep CI with it running after all other tests.

Yes

@ntn-x2 @youngjoon-lee 

Yes

No

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/network/src/conn_maintenance.rs
+++ b/nomos-blend/network/src/conn_maintenance.rs
@@ -79,7 +79,6 @@ where
         let Poll::Ready(Some(new_expected_message_count_range)) =
             self.connection_window_clock.poll_next_unpin(cx)
         else {
-            // Waker will be called by the `poll_next_unpin` when a new item is ready.
             return Poll::Pending;
         };
         // First tick is used to set the range for the new observation window.
--- a/nomos-blend/network/src/conn_maintenance.rs
+++ b/nomos-blend/network/src/conn_maintenance.rs
@@ -75,11 +75,14 @@ where
     /// If the interval has elapsed, evaluate the peer's status,
     /// reset the monitor, and return the result as `Poll::Ready`.
     /// If not, return `Poll::Pending`.
-    pub fn poll(&mut self, cx: &mut Context<'_>) -> Poll<ConnectionMonitorOutput> {
-        let Poll::Ready(Some(new_expected_message_count_range)) =
-            self.connection_window_clock.poll_next_unpin(cx)
-        else {
-            return Poll::Pending;
+    pub fn poll(&mut self, cx: &mut Context<'_>) -> Poll<Option<ConnectionMonitorOutput>> {
+        let new_expected_message_count_range = match self
+            .connection_window_clock
+            .poll_next_unpin(cx)
+        {
+            Poll::Ready(Some(new_expected_message_count_range)) => new_expected_message_count_range,
+            Poll::Ready(None) => return Poll::Ready(None),
+            Poll::Pending => return Poll::Pending,
         };
         // First tick is used to set the range for the new observation window.
         if self.expected_message_range.is_none() {
@@ -95,7 +98,7 @@ where
             ConnectionMonitorOutput::Healthy
         };
         self.reset(new_expected_message_count_range);
-        Poll::Ready(outcome)
+        Poll::Ready(Some(outcome))
     }
 }
 
@@ -128,7 +131,7 @@ mod tests {
         monitor.record_message();
         assert_eq!(
             monitor.poll(&mut Context::from_waker(&noop_waker())),
-            Poll::Ready(ConnectionMonitorOutput::Healthy)
+            Poll::Ready(Some(ConnectionMonitorOutput::Healthy))
         );
 
         // Recording the maximum expected number of messages,
@@ -137,7 +140,7 @@ mod tests {
         monitor.record_message();
         assert_eq!(
             monitor.poll(&mut Context::from_waker(&noop_waker())),
-            Poll::Ready(ConnectionMonitorOutput::Healthy)
+            Poll::Ready(Some(ConnectionMonitorOutput::Healthy))
         );
 
         // Recording more than the expected number of messages,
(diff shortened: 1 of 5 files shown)


nomos-blend/network/src/conn_maintenance.rs
```diff
@@ -79,7 +79,7 @@ where
         let Poll::Ready(Some(new_expected_message_count_range)) =
             self.connection_window_clock.poll_next_unpin(cx)
         else {
-            cx.waker().wake_by_ref();
+            // Waker will be called by the `poll_next_unpin` when a new item is ready.
```
zeegomo: who's going to call `poll_next_unpin` if the above call returns `Poll:Ready(None))`?

bacv: When `poll_next_unpin` is called, a context with `waker` is passed to the stream handler, stream calls waker when item is ready.

bacv: ~~Oh if it's `Poll::Ready(None)` then it executes this poll method further and it's up tho this whole `poll` method to decide what happens next.~~ It returns `Poll::Pending`

bacv: So looked at the docs - when stream returns `Poll::Ready(None)` - it doesn't call waker anymore, stream is closed. Not sure what should happen then. Probably Poll::Pending, effectively never executing poll method again?

bacv: If we poll again this stream as is - it will panic, either we propagate the `Poll::Ready(None)` or use `fuse()` so that it always return `Poll::Ready(None)` after closure.

youngjoon-lee: @ntn-x2 What do you think about just returning `Poll::Ready(None)` in this case,  so that `ConnectionMonitor::poll` will never be called again? 

In this `ConnectionMonitor::poll` implementation, we actually expect that `self.connection_window_clock.poll_next_unpin` never returns `Poll::Ready(None)` because we will anyway build the `self.connection_window_clock` to tick every N seconds and never end. However, from the perspective of `ConnectionMonitor`, this clock is a generic. So, it shouldn't expect any behavior that is not implemented in this module.

So, I'd like to suggest: Just stopping the operation once `self.connection_window_clock.poll_next_unpin` has terminated (returns `Poll::Ready(None)`). It means that the connection monitor stops working if there is no clock left. I think it makes sense, though we will anyway build the never-ending clock stream.

(or we can even panic)

An advantage of this approach is that we don't need to introduce any other mechanism (such as `fuse()`) to handle the case where the clock stream returns `Poll::Ready(None)`.

One side effect of this is that [`self.reset`](https://github.com/logos-co/nomos/blob/a198e4379d739cbb94d104192b8516b1b280b53c/nomos-blend/network/src/conn_maintenance.rs#L98) is not going to be called if the polling is stopped. Then, the message counter (`usize`) will increase indefinitely. We may need to stop increasing the counter as well.

nomos-blend/network/src/tokio.rs
```diff
@@ -47,7 +50,7 @@ impl IntervalStreamProvider for ObservationWindowTokioIntervalProvider {
         let expected_message_range = self.calculate_expected_message_range();
         Box::new(
             tokio_stream::wrappers::IntervalStream::new(tokio::time::interval(
-                Duration::from_secs(self.round_duration_seconds.get()),
+                Duration::from_secs(OBSERVATION_WINDOW_ROUNDS * self.round_duration_seconds.get()),
```
youngjoon-lee: Thank you so much, @bacv! I'm sorry about this bug. Your fix is correct!

One thing I can suggest is:
We don't need to define a constant `OBSERVATION_WINDOW_ROUNDS` since we already have `self.rounds_per_observation_window`.
@ntn-x2 Could you double check this? I think we can just simply modify this code as below, but I'd like to ask for your confirmation :)
```suggestion
                Duration::from_secs(
                    (self.rounds_per_observation_window.get() as u64)
                        * self.round_duration_seconds.get(),
                ),
```

nomos-blend/network/src/tokio.rs
```diff
@@ -9,6 +9,9 @@ pub use tokio_stream::StreamExt as _;
 
 use crate::IntervalStreamProvider;
 
+/// Number of rounds that the observation window lasts.
+const OBSERVATION_WINDOW_ROUNDS: u64 = 30;
+
```
youngjoon-lee: If we apply my suggestion in https://github.com/logos-co/nomos/pull/1382/files#r2160547732, we can remove this constant.

```suggestion
```

nomos-blend/network/src/conn_maintenance.rs
ntn-x2: Yes, I answered something similar to a similar point in a related PR. I think it's fine to return `Ready(None)` if any of the substreams returns `Ready(None)` indeed, and let the caller decide whether to panic or do something else.

nomos-blend/network/src/tokio.rs
```diff
@@ -47,7 +47,10 @@ impl IntervalStreamProvider for ObservationWindowTokioIntervalProvider {
         let expected_message_range = self.calculate_expected_message_range();
         Box::new(
             tokio_stream::wrappers::IntervalStream::new(tokio::time::interval(
-                Duration::from_secs(self.round_duration_seconds.get()),
+                Duration::from_secs(
```
ntn-x2, line 50: Ops, this slipped through. Nice catch!

nomos-blend/network/src/conn_maintenance.rs
zeegomo: oh right, I missed the part where `ConnectionWindowClock` is a stream, all good to return `Ready(None)` in this case, as we shouldn't schedule it anymore

PR #1385, opened 2025-06-23
feat(blend): integrate new blend encapsulation to the service

youngjoon-lee:
## 1. What does this PR implement?

This PR integrates the new blend message encapsulation (implemented in #1371) to the blend service. So, this is a stacked PR based on #1371.

28 files changed, but don't worry. Most of them are just generics changes.

Key changes:
- In the `nomos-blend-message` crate,
  - Removed the `mock` module
  - Removed the `sphinx` module  
  - Removed `BlendMessage` trait 
- In the `nomos-blend` crate, 
  - Updated `CryptographicProcessor` to use `EncapsulatedMessage` implemented in #1371 
  - Removed unnecessary generics
  - Refactoring functions and return types to handle data/cover messages cleaner.
- In the `nomos-blend-service` crate,
  - Nothing special. Just applying the changes from the `CryptographicProcessor`.

## 2. Does the code have enough context to be clearly understood?

Yes

## 3. Who are the specification authors and who is accountable for this PR?

- Spec: @madxor 
- Code: @youngjoon-lee @ntn-x2 

## 4. Is the specification accurate and complete?

Yes

## 5. Does the implementation introduce changes in the specification?

No

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [x] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [x] 2. Description added.
* [x] 3. Context and links to Specification document(s) added.
* [x] 4. Main contact(s) (developers and specification authors) added
* [x] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 6. Link PR to a specific milestone.

PR #1386, opened 2025-06-23
feat(blend): verify signature in public header

youngjoon-lee:

This is a incremental PR based on #1371.

This implements verifying a signature included in a public header.

This verification must be performed in the following 2 cases:
1. when [reconstructing](https://www.notion.so/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#215261aa09df81de89a2fae9ce5dadce) a public header while decapsulating a private header.
2. as soon as a node [receives](https://www.notion.so/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81f6b2ffdbda5054a191) a message from a peer.

This PR implements only the case 1, since this new encapsulation is not integrated in the blend service yet (#1385 is in review).

While implementing this PR, I found a bug in `EncapsulatedPrivateHeader::initialize`. This bug can be detected when verifying signatures. So, this PR fixes the bug as well.

Yes

- Spec: @madxor 
- Code: @youngjoon-lee @ntn-x2 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.


nomos-blend/message/src/message.rs
```diff
@@ -41,6 +41,16 @@ pub struct PublicHeader {
     pub signature: Signature,
 }
 
+impl PublicHeader {
```
ntn-x2, line 44: Do you think it makes sense to have this return a `VerifiedPublicHeader` that does not have the `signature` field anymore?

ntn-x2 (approved):
Looks good! Just one nit.

nomos-blend/message/src/message.rs
youngjoon-lee: Aha. This `PublicHeader` struct (the `signature` field) should be preserved even after the `verify_signature` is called. 

When decapsulating the private header (a set of blending headers), the first blending header is popped out from the private header. And, we [reconstruct the public header using the popped-out blending header](https://github.com/logos-co/nomos/blob/28d4cf6222c2830d503989346e16b30c3eb430b2/nomos-blend/message/src/encap.rs#L180). Then, we [verify that the reconstructed public header](https://github.com/logos-co/nomos/blob/28d4cf6222c2830d503989346e16b30c3eb430b2/nomos-blend/message/src/encap.rs#L182) is valid (e.g. verifying the signature). If the public header is valid, then we use it to build the new EncapsulatedMessage and send it to the peers.

PR #1387, opened 2025-06-23
chore: Add separate http testing api server behind feature flag 

pradovic:
The purpose of this PR is to add a separate HTTP endpoint for integration testing and use it to test membership service. All http part should be behind feature flag while the service logic (ex added message) is not but will be used only from http testing backend.

PR where this is used: https://github.com/logos-co/nomos/pull/1400.

Yes

@pradovic, @bacv 

Yes

No
## Checklist

> [!WARNING]  

* [x] 2. Description added.


tests/src/nodes/executor.rs
```diff
@@ -377,5 +377,14 @@ pub fn create_executor_config(config: GeneralConfig) -> Config {
         },
         membership: config.membership_config.service_settings,
         sdp: (),
+
+        #[cfg(feature = "testing")]
+        testing_http: nomos_api::ApiServiceSettings {
```
pradovic, line 416: I am not sure should we read testing http settings from `config.yaml` as well, or define it here as it is only for testing.

bacv: In here, we are not reading from config file - but using "general config" which is common for validators and executors, and after this part is configured, only when we spawn the executor or validator we write it's config to the file. This looks right to me as is.

The only nit is to dynamically assign port here, in some tests we might launch multiple executors.

nomos-services/api/src/http/membership.rs
```diff
@@ -0,0 +1,46 @@
+#![cfg(feature = "axum")]
```
bacv: Why is this needed here?

bacv (commented):
Looks good, can we just modify so that our integration tests would have no initial membership, and we'd use this endpoint in this PR to show that it works?

tests/src/nodes/executor.rs
pradovic: Sounds good, I have just added the proper logic in create configs (did not commit yet) that uses get_available_port

nomos-services/api/src/http/membership.rs
pradovic: These are the features of api in Cargo.toml:
```
[features]
axum            = ["dep:axum", "dep:hyper", "utoipa-swagger-ui/axum"]
default         = ["axum"]
instrumentation = []

```

And I think this `axum` feature is used to support different backends as well. Since this is the only part where I actually need axum as dep, I have added it to be conditional with the axum feature as well. Does that makes sense?

pradovic:
@bacv I am already adding a test just to test this http endpoint. Is that ok for this PR?
And then I would open a different PR to show the tests with membership changes, as I expect I might find some bugs along the way, so I don't want to bloat this PR.
So my plan was:
1. Only add the test to show that testing http call works in this PR
2. Separate PR for integration tests that uses this functionality

nomos-services/api/src/http/membership.rs
bacv: Hm, we don't feature gate here other handlers, and if we would have done that, it will probably happen in the `mod.rs`, not here. Lets remove, unless I still miss the point.

bacv:
> Only add the test to show that testing http call works in this PR

That's good plan, since this requires our CI to be updated, I would like to see it in action.

nomos-services/api/src/http/membership.rs
pradovic: You are right, sorry, I missed to clean it up, good catch!

danielSanchezQ (approved):
We will have to remove this once SDP and new ledger is integrated. Can we open an issue to remove this and adapt tests later on?
Thanks!

pradovic:
@danielSanchezQ Sounds good: https://github.com/logos-co/nomos/issues/1413. Thanks!

PR #1390, opened 2025-06-24
fix(blend): make keys compatible with serde_yaml

youngjoon-lee:

This is an incremental PR based on #1385. 

@ntn-x2 I'm sorry. In the PR #1385, I didn't check if a byte array in `config.yaml` can be deserialized into `Ed25519PublicKey` defined in the `nomos-blend-message` crate.
Precisely, a byte array in a YAML file cannot be deserialized into `ed25519-dalek::VerifyingKey`, which is used by our `Ed25519PublicKey`.

Reasons:
- `serde_yaml` crate doesn't support [`deserialize_bytes`](https://github.com/dtolnay/serde-yaml/blob/2009506d33767dfc88e979d6bc0d53d09f941c94/src/de.rs#L1502). 
  ```
  serialization and deserialization of bytes in YAML is not implemented
  ```
- But, the `ed25519-dalek::VerifyingKey` uses the function internally.

It's my bad that I didn't check it earlier. I'm opening this stacked PR instead of pushing a commit in the previous PR.

This PR:
- overrides the serde for Ed25519PublicKey with the custom `ed25519_pubkey_hex` module.
- implements the serde for Ed25519PrivateKey with the custom `ed25519_privkey_hex` module.

Nevertheless, this PR keeps `#[derive(Serialize, Deserialize)]` for Ed25519PublicKey because it should be serialized as bytes when encapsulating messages.

You may feel that it's too verbose to implement custom serde modules. I agree. We will soon implement the key-management service to simplify all of these key serde in the entire codebase.
 

Yes

@ntn-x2 @youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/core/src/serde.rs
+++ b/nomos-blend/core/src/serde.rs
@@ -1,7 +1,7 @@
 use nomos_blend_message::crypto::KEY_SIZE;
 use serde::{de::Error as _, Deserialize as _, Deserializer};
 
-/// Ed25519PrivateKey <> Hex string hex/deserialization
+/// [`Ed25519PrivateKey`] <> Hex string hex/deserialization
 pub mod ed25519_privkey_hex {
     use nomos_blend_message::crypto::Ed25519PrivateKey;
     use serde::{Deserializer, Serialize as _, Serializer};
@@ -23,7 +23,7 @@ pub mod ed25519_privkey_hex {
     }
 }
 
-/// Ed25519PublicKey <> Hex string serialization/deserialization
+/// [`Ed25519PublicKey`] <> Hex string serialization/deserialization
 pub mod ed25519_pubkey_hex {
     use nomos_blend_message::crypto::Ed25519PublicKey;
     use serde::{de::Error as _, Deserializer, Serialize as _, Serializer};
@@ -43,7 +43,7 @@ pub mod ed25519_pubkey_hex {
     {
         deserialize_hex_to_key(deserializer)?
             .try_into()
-            .map_err(|e| D::Error::custom(format!("{e}")))
+            .map_err(D::Error::custom)
     }
 }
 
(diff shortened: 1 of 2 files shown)


nomos-blend/core/src/serde.rs
```diff
@@ -0,0 +1,58 @@
...
+
+fn deserialize_hex_to_key<'de, D>(deserializer: D) -> Result<[u8; KEY_SIZE], D::Error>
+where
+    D: Deserializer<'de>,
+{
+    hex::decode(String::deserialize(deserializer)?)
+        .map_err(|e| D::Error::custom(format!("Invalid hex: {e}")))?
+        .try_into()
+        .map_err(|_| D::Error::custom("Invalid key length"))
+}
```
danielSanchezQ, line 58: I have the feeling we have this somewhere else too? :thinking:

danielSanchezQ (approved):

nomos-blend/core/src/serde.rs
youngjoon-lee: we have this:
https://github.com/logos-co/nomos/blob/be2a4f58fc66c55d8358e93163794a53746bb976/nomos-core/chain-defs/src/utils/mod.rs#L17
but this can be used only for `Foo([u8; N])` types. In this PR, we want to deserialize a value into `struct Ed25519PublicKey(ed25519_dalek::VerifyingKey)`, which is more complicated than `Foo([u8; N])`.
Also the existing code doesn't handle any case where `[u8; N]` holds invalid bytes. In this PR, I want to return an error if the value is invalid for a public key.

We also have this:
https://github.com/logos-co/nomos/blob/be2a4f58fc66c55d8358e93163794a53746bb976/nomos-libp2p/src/config/mod.rs#L68
But, this deserializes a value into `libp2p::identity::ed25519::SecretKey`.

PR #1391, opened 2025-06-24
feat(nat-traversal): merge the feature branch feat/nat-traversal into main

CHr15F0x:

The aim of this PR is to merge the feature branch [feat/nat-traversal](https://github.com/logos-co/nomos/tree/feat/nat-traversal) into `main`.

Yes.

Spec: @CHr15F0x @ntn-x2 
PR: @CHr15F0x 

Yes.

No.

## Checklist

* [x] 2. Description added.


nomos-libp2p/src/config/nat/autonat_client.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// The interval at which we will attempt to confirm candidates as external
+    /// addresses, only used for new candidates.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub probe_interval_millisecs: Option<u64>,
+
+    /// The interval at which we will retest successful external addresses.
+    /// This is used to ensure that the external address is still valid and
+    /// reachable.
+    #[serde(default = "default_restest_interval_millisecs")]
+    pub retest_successful_external_addresses_interval_millisecs: u64,
```
ntn-x2: Can we make this a `Duration`?

nomos-libp2p/src/config/nat/autonat_client.rs
```diff
@@ -0,0 +1,53 @@
...
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, Default)]
+pub struct Settings {
+    /// How many candidates we will test at most.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub max_candidates: Option<usize>,
+
+    /// The interval at which we will attempt to confirm candidates as external
+    /// addresses, only used for new candidates.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub probe_interval_millisecs: Option<u64>,
```
ntn-x2, line 17: Can we make this a `Duration`?

nomos-libp2p/src/config/nat/autonat_client.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    /// The interval at which we will retest successful external addresses.
+    /// This is used to ensure that the external address is still valid and
+    /// reachable.
+    #[serde(default = "default_restest_interval_millisecs")]
+    pub retest_successful_external_addresses_interval_millisecs: u64,
+}
+
+const fn default_restest_interval_millisecs() -> u64 {
```
ntn-x2: I guess this could also return a `Duration`

nomos-libp2p/src/config/nat/mapping.rs
```diff
@@ -0,0 +1,61 @@
+use std::time::Duration;
+
+use nomos_utils::{
+    bounded_duration::{MinimalBoundedDuration, SECOND},
+    math::PositiveF64,
+};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+
+const GW_RECV_TIMEOUT_SEC: u64 = 1;
```
ntn-x2: This should also be a `Duration`.

nomos-libp2p/src/config/nat/mapping.rs
```diff
@@ -0,0 +1,61 @@
...
+const LEASE_LIFETIME_SECS: u32 = 7200; // 2 hours
```
ntn-x2: Same.

nomos-libp2p/src/config/nat/mapping.rs
```diff
@@ -0,0 +1,61 @@
...
+const DEFAULT_RENEWAL_DELAY_FRACTION: f64 = 0.8;
+const DEFAULT_RETRY_INTERVAL_SEC: u64 = 30;
```
ntn-x2: Same.

nomos-libp2p/src/config/nat/mapping.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub renewal_delay_fraction: PositiveF64,
+    #[serde(default = "default_retry_interval")]
+    pub retry_interval: Duration,
+}
+
+const fn default_timeout() -> Duration {
+    Duration::from_secs(GW_RECV_TIMEOUT_SEC)
+}
+
+const fn default_lifetime() -> u32 {
```
ntn-x2: This should also be a `Duration`.

nomos-libp2p/src/config/mod.rs
```diff
@@ -1,56 +1,64 @@
...
+    /// Identify config (required)
     #[serde(default)]
     pub identify_config: identify::Settings,
 
     /// Chain sync config
     #[serde(default)]
     pub chain_sync_config: cryptarchia_sync::Config,
+
+    /// Nat config
+    /// When the value is None, NAT is disabled.
```
ntn-x2:
```suggestion
    /// When the value is None, NAT traversal is disabled.
```

ntn-x2: But I think we don't want to have this optional, do we?

andrussal: > But I think we don't want to have this optional, do we?

And if user wants, they can basically just set static public address. I'll remove Option.

nomos-libp2p/src/config/nat/mod.rs
```diff
@@ -4,9 +4,13 @@ pub mod autonat_client;
...
-#[derive(Debug, Clone, Copy, Serialize, Deserialize, Default)]
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub struct Settings {
     pub autonat: autonat_client::Settings,
     pub mapping: mapping::Settings,
     pub gateway_monitor: gateway::Settings,
+    /// Static external address for nodes with fixed public IPs
+    /// When set, NAT traversal is disabled and this address is used directly
+    #[serde(default)]
+    pub static_external_address: Option<libp2p::Multiaddr>,
```
ntn-x2, line 15: @andrussal I just saw this now. I think for this it would be best to use an enum, so we can let a user choose EITHER a static IP, OR the NAT settings that we have here. Else, it would be confusing to specify a gateway monitoring interval duration and also a static external address. Would you mind opening a new PR that fixes this? If we still want to maintain a `Default` state for the settings, then the equivalent of what we have now is to use the non-static version.

andrussal: > @andrussal I just saw this now. I think for this it would be best to use an enum, so we can let a user choose EITHER a static IP, OR the NAT settings that we have here. Else, it would be confusing to specify a gateway monitoring interval duration and also a static external address. Would you mind opening a new PR that fixes this? If we still want to maintain a `Default` state for the settings, then the equivalent of what we have now is to use the non-static version.

Good point, these are actually exclusive. I'll open PR

PR #1392, opened 2025-06-24
feat(mantle): Mantle Update/1: LedgerTx

zeegomo:

First of a series of PRs to update the mantle spec. This PR:
  * focuses on LedgerTx and related stuff
  * removes outdated proofs and structures
  * move nomos-mantle inside nomos-core

No additional functionality is added, in particular, we still don't process txs. This will be added in a separate PR.

Yes?

@davidrusu @thomaslavaur spec authors, @zeegomo implementing

https://www.notion.so/Mantle-Specification-21c261aa09df810c8820fab1d78b53d9

Yes
No

## Checklist

> [!WARNING]  

* [x] 2. Description added.


nomos-core/chain-defs/src/mantle/ledger.rs
```diff
@@ -0,0 +1,159 @@
...
+
+        hasher.update(b"NOMOS_UTXO_ID_V1");
+        hasher.update(self.tx_hash.0);
+        hasher.update(self.output_index.to_le_bytes());
+        hasher.update(self.note.value.to_le_bytes());
+        hasher.update(self.note.pk.as_bytes());
+
+        let hash = hasher.finalize();
+        NoteId(hash.into())
+    }
```
youngjoon-lee: Wdyt about making this function return `UtxoId` instead of `NoteId`, since you're already using the tag `NOMOS_UTXO_ID_V1`?

I see that we haven't defined the type `UtxoId`, but I guess we can simply rename the `NoteId` type to `UtxoId`, since this is only one place where the `NoteId` is constructed, if I didn't miss anything?

If we change the name like that, 
https://github.com/logos-co/nomos/blob/0a555e2c8bbed1385d4a3ceecb308b65163436c6/ledger/nomos-ledger/src/utxotree.rs#L10-L12
We can use `UtxoId` as tree keys because the tree name is `utxos`.
When starting my review, I needed some time to understand the difference between Note and Utxo. Now I understand it, but just asking what you think. If you think these names are fine, I'm fine with them. You have a better perspective.

danielSanchezQ:
image

https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExbGFmbzZjYTJ0aXdwZ2NwamtsNDM3ZW5qMXZ3cmY5Y2s1Y2ltYmxvcyZlcD12MV9naWZzX3NlYXJjaCZjdD1n/oPDgbJSjRIdoqwWudK/giphy.gif

ledger/nomos-ledger/src/utxotree.rs
```diff
@@ -0,0 +1,75 @@
...
+const MAX_UTXOS: usize = 1 << 14;
+
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Clone, Debug, PartialEq, Eq, Default)]
+pub struct UtxoTree {
+    utxos: rpds::RedBlackTreeMapSync<NoteId, Note>,
+}
+
+// TODO: change to sparse computation
+fn note_id_leaves(utxos: &rpds::RedBlackTreeMapSync<NoteId, Note>) -> [[u8; 32]; MAX_UTXOS] {
```
danielSanchezQ: Nitpick: `[[u8; 32]; MAX_UTXOS]` Hashes? I think at least a type alias would make it better to read? `MerkleLeaves`?

nodes/nomos-node/node/src/config/mod.rs
```diff
@@ -98,25 +97,32 @@ pub struct HttpArgs {
...
     #[clap(
-        long = "consensus-note-sk",
-        env = "CONSENSUS_NOTE_SK",
-        requires("note_value")
+        long = "consensus-utxo-sk",
+        env = "CONSENSUS_UTXO_SK",
+        requires = "value"
     )]
-    note_secret_key: Option<String>,
+    pub secret_key: Option<String>,
```
danielSanchezQ, line 104: To be changed by the wallet at some point?

nodes/nomos-node/node/src/config/mod.rs
```diff
@@ -98,25 +97,32 @@ pub struct HttpArgs {
...
     )]
-    note_nonce: Option<String>,
+    tx_hash: Option<String>,
+
+    #[clap(
+        long = "consensus-utxo-output-index",
+        env = "CONSENSUS_UTXO_OUTPUT_INDEX",
+        requires = "value"
+    )]
+    output_index: Option<usize>,
```
danielSanchezQ, line 125: Nitpick: Maybe we can encapsulate this in its own utxo related config? Then re-embedded it here

nomos-core/chain-defs/src/mantle/ledger.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub output_index: usize,
+    pub note: Note,
+}
+
+impl Utxo {
+    #[must_use]
+    pub fn id(&self) -> NoteId {
+        let mut hasher = crate::crypto::Hasher::new();
+
+        hasher.update(b"NOMOS_UTXO_ID_V1");
```
danielSanchezQ: Nipick: Maybe an outside const def? It can be in Utxo level:

```suggestion
impl Utxo {
    const UTXO_ID: &[u8] = &b"NOMOS_UTXO_ID_V1";
    #[must_use]
    pub fn id(&self) -> NoteId {
        let mut hasher = crate::crypto::Hasher::new();
        hasher.update(Self::UTXO_ID);
        ...
```

Use better naming, utxo_id is not really a good suggestion maybe.

nomos-core/chain-defs/src/mantle/ledger.rs
danielSanchezQ: I think naming is fine. It returnts the NoteId of the note tha defined the utxo :thinking:

nomos-core/chain-defs/src/mantle/ledger.rs
```diff
@@ -0,0 +1,159 @@
...
+
+impl Tx {
+    #[must_use]
+    pub const fn new(inputs: Vec<NoteId>, outputs: Vec<Note>) -> Self {
+        Self { inputs, outputs }
+    }
+
+    #[must_use]
+    pub fn as_sign_bytes(&self) -> Bytes {
+        let mut bytes = BytesMut::from("NOMOS_LEDGER_TXHASH_V1");
```
danielSanchezQ: nitpick: Same same, I would extract this to `const` tags.

nomos-core/chain-defs/src/mantle/ledger.rs
```diff
@@ -0,0 +1,159 @@
...
+
+        for input in &self.inputs {
+            bytes.extend(input.as_bytes());
+        }
+
+        bytes.extend_from_slice(b"INOUT_SEP");
```
danielSanchezQ: :heavy_plus_sign: 1

nomos-core/chain-defs/src/mantle/tx.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+impl Transaction for MantleTx {
+    const HASHER: TransactionHasher<Self> =
+        |tx| <[u8; 32]>::from(crate::crypto::Hasher::digest(tx.as_sign_bytes())).into();
+    type Hash = TxHash;
+
+    fn as_sign_bytes(&self) -> bytes::Bytes {
+        let mut buff = bytes::BytesMut::new();
+        buff.extend_from_slice(b"NOMOS_MANTLE_TXHASH_V1");
```
danielSanchezQ: :heavy_plus_sign: 1

nomos-core/chain-defs/src/mantle/tx.rs
```diff
@@ -0,0 +1,111 @@
...
+        for op in &self.ops {
+            buff.extend_from_slice(op.as_sign_bytes().as_ref());
+        }
+        buff.extend_from_slice(b"END_OPS");
```
danielSanchezQ: :heavy_plus_sign: 1

danielSanchezQ (commented):
Nice to see some red once in a while. I left a few comments, most of them related to const tag extractions. This looks really good otherwise, thanks a lot!

nomos-core/chain-defs/src/mantle/ledger.rs
zeegomo: FIY, I'm using names as defined in the spec when possible, although we don't have a Utxo struct there

nodes/nomos-node/node/src/config/mod.rs
zeegomo: yes exactly

nodes/nomos-node/node/src/config/mod.rs
zeegomo: I wanted to, but you can't flatten a struct and make it optional with clap, so this is a bit of machinery to achieve something like:

```
struct Utxo {
  value: ...
  sk: ...
  ....
}


struct CryptarchiaArgs {
    utxo: Option<Utxo> // clap does not understand this is optional and still requires all fields
}
```
if that's what you meant

nomos-core/chain-defs/src/mantle/ledger.rs
zeegomo: I kinda liked that you could instantly see that it matched the spec instead of going to the constant definition but yes, it's a trivial change

ledger/nomos-ledger/src/utxotree.rs
```diff
@@ -0,0 +1,75 @@
...
+
+type MerkleLeaf = [u8; 32];
+
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Clone, Debug, PartialEq, Eq, Default)]
+pub struct UtxoTree {
+    utxos: rpds::RedBlackTreeMapSync<NoteId, Note>,
+}
+
+// TODO: change to sparse computation
```
davidrusu, line 16: By sparse, do you mean that the tree leaves are in insertion order with new insertions filling in any empty slots that were created when a note was spent?

ledger/nomos-ledger/src/utxotree.rs
```diff
@@ -0,0 +1,75 @@
+use nomos_core::mantle::{
+    merkle::{self, PathNode},
+    Note, NoteId, Utxo,
+};
+// up to 2^14 commitments
+const MAX_UTXOS: usize = 1 << 14;
```
davidrusu, line 6: This limitation is temporary right? I think we specced 2^32 max utxos?

nomos-core/chain-defs/src/mantle/keys.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    #[must_use]
+    pub const fn as_bytes(&self) -> &[u8; 16] {
+        &self.0
+    }
+
+    #[must_use]
+    pub const fn to_public_key(&self) -> PublicKey {
+        // Placeholder for actual public key derivation logic
+        PublicKey([0; 32])
```
davidrusu: uh, either add a `// TODO:` or preferably panic with `unimplemented!()`. Don't want to somehow miss this later

nomos-core/chain-defs/src/mantle/ledger.rs
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        bytes.extend_from_slice(INOUT_SEP);
+
+        for output in &self.outputs {
+            bytes.extend_from_slice(output.to_bytes().as_ref());
+        }
+
+        bytes.freeze()
+    }
```
davidrusu, line 104: Are you using this as the serialization format of the ledger tx? If so, then it's a bit wasteful with the strings inline.

davidrusu (approved):

nomos-core/chain-defs/src/mantle/ledger.rs
danielSanchezQ: > Are you using this as the serialization format of the ledger tx? If so, then it's a bit wasteful with the strings inline.

Why?

nomos-core/chain-defs/src/mantle/keys.rs
danielSanchezQ: Yeah, actually maybe opening an issue is better unless this is gonna be tackled in a soon further pr.

nomos-core/chain-defs/src/mantle/ledger.rs
zeegomo: no it's not the serialization format, `as_sign_bytes` is meant to be the bytes on which you build a signature (which is used to compute the hash), here it's a bit of overload for convenience since we don't use the ledger tx hash for signing.

davidrusu: This is putting the string constants NOMOS_LEDGER_TXHASH_V1 and INOUT_SEP in the serialized encoding, that's like ~50bytes of overhead

ledger/nomos-ledger/src/utxotree.rs
zeegomo: By sparse I mean that we don't need to actually allocate the full max sized tree to compute the root.
Actually, do we define how to compute the root (as in, order of notes)?

ledger/nomos-ledger/src/utxotree.rs
zeegomo: yes, we currently allocate all the leaves so it would be quite inefficient, we can tackle this in a separate PR as this is already quite big

nomos-core/chain-defs/src/mantle/ledger.rs
danielSanchezQ: > This is putting the string constants NOMOS_LEDGER_TXHASH_V1 and INOUT_SEP in the serialized encoding, that's like ~50bytes of overhead

oh, ok, i dont think this goes through wire as explained in the comment above. thanks for the explanation though, I didn't understood the comment.

nomos-core/chain-defs/src/mantle/keys.rs
zeegomo: added the `unimplemented!`

Issue #1309, opened 2025-05-29
`two_nodes_happy` flaky test


strinnityk:
https://github.com/logos-co/nomos/actions/runs/15323180269/job/43111271182?pr=1306
https://github.com/logos-co/nomos/actions/runs/15352452565/attempts/1
https://github.com/logos-co/nomos/actions/runs/15731421345/job/44333243454?pr=1374

Issue #1304, opened 2025-05-29
`happy_test` flaky test on Linux


strinnityk:
https://github.com/logos-co/nomos/actions/runs/15319363009/job/43099458231?pr=1302

PR #1395, opened 2025-06-25
fix(cryptarchia): Fix misbehaviour due to pruning condition and flaky `two_nodes_happy` test.

strinnityk:
After investigating the `two_nodes_happy` test for a while, I realised that when the two nodes proposed blocks for the same slot their forks would get immediately pruned off. That'd happen even if the forks were happening from LIB (or genesis). Thus:
- Fix a block pruning condition that was pruning forks older than (and including) selected depth (e.g.: LIB), instead of strictly older.
- Fix flaky behaviour on `two_nodes_happy` test.

Closes: https://github.com/logos-co/nomos/issues/1309 
Closes: https://github.com/logos-co/nomos/issues/1304

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/consensus/cryptarchia-engine/src/lib.rs
+++ b/consensus/cryptarchia-engine/src/lib.rs
@@ -399,42 +399,43 @@ where
         &self.local_chain
     }
 
-    /// Prune all blocks that are included in forks that diverged at or before
-    /// the `depth`th block from the current local chain.
-    ///
-    /// For example, if the tip of the canonical chain is at height 10 (i.e., 11
-    /// blocks long), calling `self.prune_forks(10)` will remove any forks
-    /// stemming from the genesis block, with height `0`, which is the 10th
-    /// block in the past.
-    ///
-    /// This function does not apply any particular logic when evaluating forks
-    /// other than the height at which they diverged from the local
-    /// canonical chain.
-    ///
+    /// Prune all blocks that are included in forks that diverged before
+    /// the `max_div_depth`-th block from the current local chain tip.
     /// It returns the block IDs that were part of the pruned forks.
-    pub fn prune_forks(&mut self, depth: u64) -> impl Iterator<Item = Id> + '_ {
+    ///
+    /// For example,
+    /// Given a block tree:
+    ///               b6
+    ///             /
+    /// G - b1 - b2 - b3 - b4 - b5 == local chain tip
+    ///                  \
+    ///                    b7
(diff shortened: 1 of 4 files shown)


consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -443,7 +443,7 @@ where
             // We calculate LCA once and store it in `ForkInfo` so it can be consumed
             // elsewhere without the need to re-calculate it.
             let lca = self.branches.lca(&local_chain, &fork);
-            (lca.length <= target_height).then_some(ForkDivergenceInfo { tip: fork, lca })
+            (lca.length < target_height).then_some(ForkDivergenceInfo { tip: fork, lca })
```
ntn-x2: Can you please also update the function comment? It says that it deletes chains at or before the provided `depth`, but this should change. There is nothing wrong with this function, but rather what `depth` is provided elsewhere, but I agree this change will make it easier to provide a value of `depth` which is simply derived from the height of the LIB.

nomos-services/cryptarchia-consensus/src/states.rs
```diff
@@ -199,25 +199,22 @@ mod tests {
...
+        // `0` to `3`), it means that all forks diverging before 2
+        // blocks in the past are considered prunable.
+        // That is:
+        // - Blocks `3` and `4`, belonging to the first fork from genesis.
+        // - Block `6` belonging to the second fork from genesis.
+        // On the other hand:
+        // - Block `7` is not pruned since it diverged from block `1`, which is the LIB.
+        // - Block `8` is not pruned since it diverged from block `2`, which is 1 block
+        //   younger
+        // than LIB.
```
ntn-x2: I guess this belongs to the previous line.

ntn-x2 (approved):
This seems to make sense to me. My bad, something got messy with the cryptarchia changes and the introduction of the LIB halfway through this PR, since the problem is not so much with the prune function itself, but how it is called from the cryptarchia service, which should have used a larger value for `depth`.

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: I opened a PR https://github.com/logos-co/nomos/pull/1397 before reading this thread :) and I just saw you approved it. Thanks!

PR #1398, opened 2025-06-26
fix(cryptarchia): don't include not-yet-pruned blocks in the service state

youngjoon-lee:

As described in many code comments, we defined the `CryptarchiaConsensusState` (which is for recovery) to store `prunable_blocks` as well as the tip and the LIB.
Here, the `prunable_blocks` means the blocks which were pruned from memory (cryptarchia engine) but not from storage.

It's good. But in the current implementation, the `CryptarchiaConsensusState` was also storing blocks that are **not yet pruned** in the cryptarchia engine. 
Whenever we rebuild the `CryptarchiaConsensusState`, we fetch **prunable** blocks from the cryptarchia engine and include them into the `CryptarchiaConsensusState`.
However, **those blocks don't need to be included** because they are not yet pruned in the cryptarchia engine.

This PR simply **removes fetching prunable blocks** from the cryptarchia engine.
We should build `CryptarchiaConsensusState` only with the provided `prunable_blocks`, which were already pruned from memory but not from storage.

Even after this change, the consensus service should work as before, because we anyway prune blocks right before rebuliding the `CryptarchiaConsensusState`. In other words, the previous implementation has worked well, but had an unnecessary call of fetching prunable blocks, which may cause side effects in the future.

(FYI, this PR is one of preparations for implementing chain bootstrapping.)

Yes

- Spec: The state recovery is not directly related to the Cryptarchia spec.
- Code owners: @zeegomo @youngjoon-lee 
but adding @ntn-x2 @strinnityk as reviewers who have heavily contributed to the state recovery and pruning.

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/cryptarchia-consensus/src/lib.rs
+++ b/nomos-services/cryptarchia-consensus/src/lib.rs
@@ -624,12 +624,12 @@ where
                         Self::log_received_block(&block);
 
                         // Process the received block and update the cryptarchia state.
-                        cryptarchia = Self::process_block_and_update_state(
+                        (cryptarchia, storage_blocks_to_remove) = Self::process_block_and_update_state(
                             cryptarchia,
                             &leader,
                             block.clone(),
 
-                            &mut storage_blocks_to_remove,
+                            &storage_blocks_to_remove,
                             &relays,
                             &self.block_subscription_sender,
                             self.service_resources_handle
@@ -663,12 +663,12 @@ where
 
                             if let Some(block) = block {
                                 // apply our own block
-                                cryptarchia = Self::process_block_and_update_state(
+                                (cryptarchia, storage_blocks_to_remove) = Self::process_block_and_update_state(
                                     cryptarchia,
                                     &leader,
                                     block.clone(),
 
-                                    &mut storage_blocks_to_remove,
+                                    &storage_blocks_to_remove,
                                     &relays,
                                     &self.block_subscription_sender,
                                     self.service_resources_handle
@@ -881,7 +881,7 @@ where
         mut cryptarchia: Cryptarchia<Online>,
         leader: &Leader,
         block: Block<ClPool::Item, DaPool::Item>,
-        storage_blocks_to_remove: &mut HashSet<HeaderId>,
+        storage_blocks_to_remove: &HashSet<HeaderId>,
         relays: &CryptarchiaConsensusRelays<
             BlendAdapter,
             BS,
@@ -908,14 +908,14 @@ where
                 >,
             >,
         >,
-    ) -> Cryptarchia<Online> {
+    ) -> (Cryptarchia<Online>, HashSet<HeaderId>) {
         cryptarchia =
             Self::process_block(cryptarchia, block, relays, block_subscription_sender).await;
 
         // This will modify `previously_pruned_blocks` to include blocks which are not
         // tracked by Cryptarchia anymore but have not been deleted from the persistence
         // layer.
-        Self::prune_forks(
+        let storage_blocks_to_remove = Self::prune_forks(
             &mut cryptarchia,
             relays.storage_adapter(),
             storage_blocks_to_remove,
@@ -935,7 +935,7 @@ where
             }
         }
 
-        cryptarchia
+        (cryptarchia, storage_blocks_to_remove)
     }
 
     /// Try to add a [`Block`] to [`Cryptarchia`].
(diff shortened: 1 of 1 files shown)


nomos-services/cryptarchia-consensus/src/states.rs
```diff
@@ -47,41 +46,13 @@ impl<TxS, BxS, NetworkAdapterSettings, BlendAdapterSettings>
         let lib_block_length = lib.length();
         let lib_leader_utxos = leader.utxos().to_vec();
```
ntn-x2, line 47: This was accounting for any inconsistencies between the provided prunable blocks and what's in cryptarchia, e.g., if cryptarchia had in the meanwhile moved on and it would need to prune more forks. As far as this function is concerned, cryptarchia might still have stale blocks, that is why I decided that it does not hurt to call `prunable_forks` again in here. In the best case, that will return an empty iterator and the input hashset is not changed. Maybe can we add a comment that explains we are assuming cryptarchia does not have any unpruned forks?

ntn-x2 (commented):
I understand the change, but I think the function comment could be improved a bit to indicate that `prunable_blocks` refers to previously processed blocks and that we are not looking into the engine state at this point.

nomos-services/cryptarchia-consensus/src/states.rs
youngjoon-lee: I will enhance the comments. 

Sorry, I didn't clearly understand your explanation. I think the inconsistencies that you mentioned are fine from the perspective of recovery state. I think it's fine for the recovery state to just store `blocks_pruned_from_engine_but_not_from_storage`. In the previous code that fetches prunable blocks from engine when rebuilding the recovery state, someone might be confused in the future when he loads the recovery state from storage and tries to decide what to do. He should be able to clearly know that the blocks loaded from the recovery state were already removed from the engine but not from the storage. If the recovery state mixes them with blocks that are still in the engine, it may be confusing in the future. Do I understand the purpose of the recovery correctly? I guess I'm on the same page with you, but checking if I'm missing anything.

youngjoon-lee: a801a973 I changed the variable name to `storage_blocks_to_remove`, which is clearer, I think.
The current [function comment](https://github.com/logos-co/nomos/blob/a801a973b798e0ee39c43ceca2968705a9335def/nomos-services/cryptarchia-consensus/src/states.rs#L32) looks clear to me, even with the new variable name. I don't think we should mention in the comment that this function doesn't fetch prunable blocks from engine, because that's the previous behavior that is removed by this PR. I hope the new variable name makes things much clearer. Please let me know your opinion.

ntn-x2: I think it's fine, this is simpler to follow, so we can go ahead with this.

ntn-x2 (approved):

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -1240,35 +1240,37 @@ where
...
             // All blocks, past and present, have been successfully deleted from storage.
-            Ok(()) => *prunable_blocks = HashSet::new(),
+            Ok(()) => *storage_blocks_to_remove = HashSet::new(),
             // We retain the blocks that failed to be deleted.
             Err(failed_blocks) => {
-                *prunable_blocks = failed_blocks
+                *storage_blocks_to_remove = failed_blocks
                     .into_iter()
                     .map(|(block_id, _)| block_id)
                     .collect();
```
zeegomo: nit: comment says 'Any block that fails to be deleted from the storage layer is added to the provided `storage_blocks_to_remove` parameter and will be picked up at the next invocation of this function' but you overwrite its current contents

zeegomo (approved):

nomos-services/cryptarchia-consensus/src/lib.rs
youngjoon-lee: Yeah. The comment was based on the previous implementation. That doesn't make sense for the current implementation. Instead, I updated the function to return the new `storage_blocks_to_remove` instead of overwriting the previous one. I think it's cleaner. Of course, I updated the comment as well. [a5de61c](https://github.com/logos-co/nomos/pull/1398/commits/a5de61ceb4c634d6dff03193ccdb2580ed7b05bf)

PR #1400, opened 2025-06-26
chore(membership): integration tests update membership and dissiminate

pradovic:
This PR adds integration tests for membership. It does the same thing as regular da dissiminate test but uses update membership instead of membership configuration to update membership before dissiminate.

Context: https://github.com/logos-co/nomos/pull/1387.


tests/src/topology/mod.rs
```diff
@@ -18,20 +18,37 @@ use crate::{
...
-    network_params: NetworkParams,
+    pub n_validators: usize,
+    pub n_executors: usize,
+    pub consensus_params: ConsensusParams,
+    pub da_params: DaParams,
+    pub network_params: NetworkParams,
+}
+
+#[derive(Clone)]
+pub struct NodeConfig {
```
bacv: Hm, I'm not really getting this NodeConfig, it seems it's exactly the same as `GeneralConfig` just instead having `Vec<GeneralConfig>` we now have
```
NodeConfig {
  consensus: Vec<GeneralConsensus>,
  da: Vec<GeneralDa>,
  ....
}
```

pradovic: I extracted fn to reuse logic with configs, and clippy warned about too many params, so I added a new struct to solve this. I willi rewrite it to Vec<GeneralConfig>

bacv (commented):
I would leave Topology as is, add a method `create_blank` to `tests/src/topology/configs/membership.rs ` and use that method in `spawn_validator_executor_wo_membership` when creating topology without membership. After that's spawned, I would iterate through the generated node ids and construct membership from that

tests/src/topology/mod.rs
pradovic: Fixed: https://github.com/logos-co/nomos/pull/1400/commits/113b2f2b8aeb160339918732a40059db16fe51e9.

danielSanchezQ (approved):
Nice! Thanks!

PR #1403, opened 2025-06-27
refactor(cryptarchia): prune blocks automatically in engine

youngjoon-lee:

(This is a stacked PR based on #1398. This doesn't need to be necessarily a stacked PR, but it's much easier for reviews.)

### Changes
- Prune blocks automatically in `cryptarchia-engine` when LIB is updated.
  
### Backgrounds / Details

- So far, we've triggered pruning in the consensus service after `process_block`. That works well, but this PR is a preparation to address a [TODO comment](https://github.com/logos-co/nomos/blob/5f57d55059a3ddf1dd57829486c2b3dcf25c088a/consensus/cryptarchia-engine/src/lib.rs#L222) in the `cryptarchia-engine`.
    - In the block validation, we call `is_ancestor(lib, parent)` to ensure a new block is a descendant of LIB. It works well, but not so efficient if LIB is deep. To optimize this, the Cryptarchia spec suggests to just [compare heights](https://www.notion.so/Cryptarchia-v1-Protocol-Specification-21c261aa09df810cb85eff1c76e5798c?source=copy_link#21c261aa09df81978b9df055c4b8e67a), under the assumption that unnecessary forks have been pruned immediately.
    - So, this PR changes the `cryptarchia-engine` to automatically prune blocks when LIB is updated. No pruning is necessary if LIB is not updated.
    - This PR doesn't change the block validation yet. That will be implemented in the next PR.
- This is the last preparation before implementing the chain bootstrapping.
  - The chain bootstrapping should work even if we don't optimize the block validation. But, considering a long chain during bootstrapping, I'd like to minimize triggering block pruning, by this PR.

Yes

@zeegomo @youngjoon-lee @ntn-x2 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/chain-service/src/lib.rs
+++ b/nomos-services/chain-service/src/lib.rs
@@ -567,10 +567,10 @@ where
             &self.block_subscription_sender,
         )
         .await;
-        let mut storage_blocks_to_remove = Self::remove_pruned_blocks_from_storage(
+        let mut storage_blocks_to_remove = Self::delete_pruned_blocks_from_storage(
             &pruned_blocks,
-            relays.storage_adapter(),
             &storage_blocks_to_remove,
+            relays.storage_adapter(),
         )
         .await;
 
@@ -909,10 +909,10 @@ where
         let (cryptarchia, pruned_blocks) =
             Self::process_block(cryptarchia, block, relays, block_subscription_sender).await;
 
-        let storage_blocks_to_remove = Self::remove_pruned_blocks_from_storage(
+        let storage_blocks_to_remove = Self::delete_pruned_blocks_from_storage(
             &pruned_blocks,
-            relays.storage_adapter(),
             storage_blocks_to_remove,
+            relays.storage_adapter(),
         )
         .await;
 
@@ -1238,29 +1238,27 @@ where
 
     /// Remove the pruned blocks from the storage layer.
     ///
-    /// Also, this removes the `storage_blocks_to_remove` from the storage
+    /// Also, this removes the `additional_blocks` from the storage
     /// layer. These blocks might belong to previous pruning operations and
     /// that failed to be removed from the storage for some reason.
     ///
     /// This function returns any block that fails to be deleted from the
     /// storage layer.
-    async fn remove_pruned_blocks_from_storage(
-        newly_pruned_blocks: &PrunedBlocks<HeaderId>,
+    async fn delete_pruned_blocks_from_storage(
+        pruned_blocks: &PrunedBlocks<HeaderId>,
+        additional_blocks: &HashSet<HeaderId>,
         storage_adapter: &StorageAdapter<Storage, TxS::Tx, BS::BlobId, RuntimeServiceId>,
-        storage_blocks_to_remove: &HashSet<HeaderId>,
     ) -> HashSet<HeaderId> {
-        // We try to delete both newly pruned blocks as well as the blocks that
-        // were pruned in the past but failed to be deleted from storage.
-        match Self::delete_pruned_blocks_from_storage(
-            newly_pruned_blocks
+        match Self::delete_blocks_from_storage(
+            pruned_blocks
                 .iter()
-                .chain(storage_blocks_to_remove.iter())
+                .chain(additional_blocks.iter())
                 .copied(),
             storage_adapter,
         )
         .await
         {
-            // All blocks, past and present, have been successfully deleted from storage.
+            // No blocks failed to be deleted.
             Ok(()) => HashSet::new(),
             // We retain the blocks that failed to be deleted.
             Err(failed_blocks) => failed_blocks
@@ -1270,20 +1268,20 @@ where
         }
     }
 
-    /// Send a bulk deletion request to the storage adapter.
+    /// Send a bulk blocks deletion request to the storage adapter.
     ///
     /// If no request fails, the method returns `Ok()`.
     /// If any request fails, the header ID and the generated error for each
     /// failing request are collected and returned as part of the `Err`
     /// result.
-    async fn delete_pruned_blocks_from_storage<Headers>(
-        pruned_block_headers: Headers,
+    async fn delete_blocks_from_storage<Headers>(
+        block_headers: Headers,
         storage_adapter: &StorageAdapter<Storage, TxS::Tx, BS::BlobId, RuntimeServiceId>,
     ) -> Result<(), Vec<(HeaderId, DynError)>>
     where
         Headers: Iterator<Item = HeaderId> + Send,
     {
-        let blocks_to_delete = pruned_block_headers.collect::<Vec<_>>();
+        let blocks_to_delete = block_headers.collect::<Vec<_>>();
         let block_deletion_outcomes = blocks_to_delete.iter().copied().zip(
             storage_adapter
                 .remove_blocks(blocks_to_delete.iter().copied())


consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -373,18 +373,35 @@ where
...
-    pub fn update_lib(&mut self) {
-        self.branches.lib = <State as CryptarchiaState>::lib(&*self);
+    fn update_lib(&mut self) -> impl Iterator<Item = Id> + '_ {
+        let new_lib = <State as CryptarchiaState>::lib(&*self);
+        // Trigger pruning only if the LIB has changed.
+        if self.branches.lib == new_lib {
+            Box::new(core::iter::empty()) as Box<dyn Iterator<Item = Id>>
+        } else {
+            self.branches.lib = new_lib;
+            Box::new(self.prune_forks(self.lib_depth()))
```
zeegomo: why the need for `Box`?

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -170,8 +167,29 @@ impl<State: CryptarchiaState> Cryptarchia<State> {
...
+///
+/// This also contains the set of pruned blocks for post-processing,
+/// such as removing them from the storage layer.
+struct UpdatedCryptarchia<State> {
+    cryptarchia: Cryptarchia<State>,
+    pruned_blocks: HashSet<HeaderId>,
+}
+
+/// Convert a [`Cryptarchia`], which is not updated, to an
+/// [`UpdatedCryptarchia`].
```
zeegomo: nit: I find this comment a bit confusing (and `UpdatedCryptarchia` naming in general). It does not make the data any more updated by wrapping in into another struct

youngjoon-lee: I agree. I updated the comment dec500544f7a64c86d87d85972c5bd4234b65bf2

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: It's to avoid the error: `if and else have incompatible types`. The type of `core::iter::empty()` is `Empty`, which is different from what is returned from the `self.prune_forks` in the `else` clause.
The same technique has been used in: https://github.com/logos-co/nomos/blob/600d3788831857b9a54193049f3d95de7fc1cf2f/consensus/cryptarchia-engine/src/lib.rs#L447

But now, I think this function doesn't necessarily need to return `impl Iterator<Item = Id>` in our codebase. It's okay for it to return a HashSet because we anyway do `self.update_lib().collect()` in the callers.

nomos-services/cryptarchia-consensus/src/lib.rs
youngjoon-lee: > (and UpdatedCryptarchia naming in general).

Let me think about this.

youngjoon-lee: Alright, I agree that the naming is confusing. 
I realized that it's better to remove the `UpdateCryptarchia` and instead, return `(Cryptarchia, PrunedBlocks)` explicitly. 1d316e2a. This also solves https://github.com/logos-co/nomos/pull/1403#discussion_r2174766486.

Thanks for your feedback. I'm re-requesting your review. (Sorry for asking another round).

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: This topic is resolved by https://github.com/logos-co/nomos/pull/1403#discussion_r2176993319

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -373,18 +373,32 @@ where
...
+        parent: Id,
+        slot: Slot,
+    ) -> Result<(Self, PrunedBlocks<Id>), Error<Id>> {
         let mut new: Self = self.clone();
         new.branches = new.branches.apply_header(id, parent, slot)?;
         new.local_chain = new.fork_choice();
-        new.update_lib();
-        Ok(new)
+        let pruned_blocks = new.update_lib();
+        Ok((new, pruned_blocks))
```
danielSanchezQ, line 393: Would not this be a single block always? :thinking:

danielSanchezQ (approved):
We do still keep the blocks in storage which is fine I guess. Or do we have any plans for storaged blocks?

consensus/cryptarchia-engine/src/lib.rs
zeegomo: not necessarily, there could be three-headed (or more) forks in theory

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -386,21 +386,21 @@ where
...
-        let pruned_blocks = new.update_lib().collect();
+        let pruned_blocks = new.update_lib();
         Ok(UpdatedCryptarchia {
             cryptarchia: new,
             pruned_blocks,
         })
     }
 
-    fn update_lib(&mut self) -> impl Iterator<Item = Id> + '_ {
+    fn update_lib(&mut self) -> HashSet<Id> {
```
zeegomo: maybe a comment on what is returned by this function, it's not immediately obvious from the signature

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -373,18 +373,32 @@ where
         }
     }
 
-    /// Create a new [`Cryptarchia`] instance with the updated state.
+    /// Process a block and returns a [`UpdatedCryptarchia`] that contains
+    /// an updated [`Cryptarchia`] instance
+    /// and the blocks pruned while processing the provided block.
```
zeegomo: comment is outdated now

nomos-services/cryptarchia-consensus/src/lib.rs
```diff
@@ -142,23 +146,10 @@ impl<State: CryptarchiaState> Cryptarchia<State> {
...
-        // collect first.
-        let lib_depth = self
-            .consensus
-            .tip_branch()
-            .length()
-            .checked_sub(self.consensus.lib_branch().length())
-            .expect("Canonical chain tip heigh must be higher than LIB height.");
-        let old_forks_pruned = self.consensus.prune_forks(lib_depth).collect::<Vec<_>>();
+    /// Remove the ledger states of the pruned blocks.
+    pub fn prune_ledger_states(&mut self, pruned_blocks: &PrunedBlocks<HeaderId>) {
```
zeegomo: I'd expand on the comment explain which blocks should be removed and which one shouldn't, and maybe remove the pub vis as this is called only internally anyway

zeegomo (approved):
Left a few comments, good to go after those are fixed!

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: Yes. Multiple forks could be pruned. And, each fork could have multiple blocks.

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: Yeah. I had already updated this signature to return `PrunedBlock` but the commit ID that I shared might not show you this change :) In the current `HEAD`, we return `PurnedBlock`, which is clearer.

Also, I added a comment for better understanding 7362a0ec. Thank you!

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: Thank you! I forgot to update this.... 
7362a0ec

nomos-services/cryptarchia-consensus/src/lib.rs
youngjoon-lee: I agree. I improved the comment. In the new comment, I don't explain which blocks are pruned from the engine because it depends on the engine implementation. But, I emphasize that this function should be called with the blocks **pruned from the engine**. Also, I added a link in the comment to point the pruning details can be found.
7362a0ec

I also remove the `pub` visibility. Thanks!

PR #1404, opened 2025-06-27
feat(cryptarchia): optimize LIB-descendant check

youngjoon-lee:

(This is a stacked PR based on https://github.com/logos-co/nomos/pull/1403.)

This PR resolves a [TODO comment](https://github.com/logos-co/nomos/blob/5f57d55059a3ddf1dd57829486c2b3dcf25c088a/consensus/cryptarchia-engine/src/lib.rs#L222) in the `cryptarchia-engine`, optimizing checking whether a new block is a descendant of the LIB, as part of block validation.
Previously, we had used `is_ancestor(lib, block)`. However, the [spec](https://www.notion.so/nomos-tech/Cryptarchia-v1-Protocol-Specification-21c261aa09df810cb85eff1c76e5798c?source=copy_link#21c261aa09df81978b9df055c4b8e67a) proposes to optimize it by just comparing heights, with the assumption that all old forks (diverged before LIB) are pruned completely.

#1403 implements pruning forks whenever LIB is updated. So, this PR optimizes the LIB descendant check based on #1403.

This PR has a long history, but I believe I’ve come up with the most reasonable solution. If you're busy, please read only the last bullet point below.

- At the beginning, I just simply replaced `is_ancestor(lib, block)` with `lib.height < block.height` in the `Branches::apply_header`: https://github.com/logos-co/nomos/commit/618efc2d40bbd9c7b4762f5437a60939c35d8a25.
- Although that was super simple, I thought it's not safe, as discussed in the [comment](https://github.com/logos-co/nomos/pull/1404#discussion_r2172268014) below. So, I refactored `Branches` and `Cryptarchia` in a way that I believe is safer. What I had refactored can be found in the [gist](https://gist.github.com/youngjoon-lee/fa4f8de63d4632db7eb720aa1992e75c) and [92bebae](https://github.com/logos-co/nomos/pull/1404/commits/92bebae700b7b7b31c97fa583ee12f261ebe8616). This removes `Branches` and merge everything into `Cryptarchia`.
- But, @zeegomo shared me the rationale of `Branches` in the [comment](https://github.com/logos-co/nomos/pull/1404#pullrequestreview-2979364399) below. I respect it and don't want to break it by this PR. 
- So, I've come up with the final version, which is again super simple and safer than the initial version.
  - It moves `Branches::lib` to `Cryptarchia::lib`. It makes sense because `local_chain` is also defined in the `Cryptarchia`. Now, `Branches` manages only a tree. All protocol-specific information is managed in the `Cryptarchia`.
  - Whenever a new block is received, `Cryptarchia` first checks if the block is a descendant of LIB, by just comparing its height with LIB's height. It works because `Cryptarchia` always prune forks diverged before LIB.
  - After that, `Cryptarchia` lets `Branches` to do the rest. If everything is good, `Cryptarchia` updates `local_chain` and `lib`, and prunes forks according to the updated `lib`.

  

Yes

@zeegomo @youngjoon-lee 

I'm also adding @ntn-x2 since he's reviewing #1403.

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/consensus/cryptarchia-engine/src/lib.rs
+++ b/consensus/cryptarchia-engine/src/lib.rs
@@ -348,7 +348,7 @@ where
     ) -> Result<(Self, PrunedBlocks<Id>), Error<Id>> {
         // Check if the block is a descendant of the LIB,
         // by checking if parent.height >= LIB.height.
-        // This works because all forks diverged before LIB has been pruned.
+        // This works because all forks diverged before LIB have been pruned.
         let parent_branch = self
             .branches
             .get(&parent)


consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -219,11 +219,12 @@ where
...
-        // sufficient to check the header height.
-        // We might relax this check in the future and get closer to the
-        // Cryptarchia spec once we stabilize the pruning logic.
-        if !self.is_ancestor(self.lib(), parent_branch) {
+        // Check if the parent is not deeper than LIB by just checking heights,
+        // under the assumption that all forks diverged deeper than LIB have
+        // been already pruned when the LIB was updated.
+        // If the assumption is not true, this check must be replaced with
+        // `is_ancestor(lib, parent)`.
+        if parent_branch.length() < self.lib().length() {
```
ntn-x2: Can you maybe call `prunable_forks` and check whether that returns an empty iterator? If that's the case, we don't need to introduce external assumptions, no?

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -383,6 +369,15 @@ where
         parent: Id,
         slot: Slot,
     ) -> Result<UpdatedCryptarchia<Id, State>, Error<Id>> {
+        // TODO: Try to remove this by refactoring.
+        // Currently, the `Branches::apply_header` requires forks to be pruned
+        // beforehand, but pruning is triggered in `Self`.
+        // It may be worth trying to make this safer by leveraging type systems.
+        debug_assert!(
```
ntn-x2: Ah I see. Would it not make more sense, instead of panicking (only for debug builds, what happens with production builds?) to decide which strategy to apply based on the result of `prunable_forks`?

youngjoon-lee: Let me think about it. I avoided calling `prunable_forks` in this logic because it's more expensive than `is_ancestor` (which this PR tries to remove).

consensus/cryptarchia-engine/src/lib.rs
danielSanchezQ: Wait, but, why would we be applying a header for a fork that diverges at an older point than *lib*?

youngjoon-lee: Yeah, Daniel. Here, in the line 227, we're trying to not apply a header for a fork diverged deeper than LIB. Previously we ensures that by calling `is_ancestor(lib, parent)`. But, this PR is suggesting to replace the `is_ancestor` check with the height comparison, which is much lighter.
However, the height comparison can be done only with the assumption that all old forks have been pruned completely. Pruning is done in the upper layer, not in this `Branches` struct. That's why Antonio was suggesting to (at least) check whether there is really no old fork not pruned yet, before performing the height comparison.

This discussion is continued in https://github.com/logos-co/nomos/pull/1404#discussion_r2172268014. Let's close this thread and move to the other one. Currently I'm thinking about how to make this safer.

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: I have a nice idea. Please hold on reviews. I will mark this PR as draft and come back soon. @ntn-x2 @danielSanchezQ @zeegomo

youngjoon-lee: @ntn-x2 @danielSanchezQ @zeegomo 
I'm re-requesting your reviews. Please see the updated PR description above that explains what has been changed. 

I guess these changes may bring some discussions, but I believe the updated code is quite safe and easy to review.

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -122,10 +127,30 @@ where
...
+/// Whenever a new block is received, it performs the fork choice rule
+/// and update the LIB if needed (depending on [`CryptarchiaState`]).
+///
+/// This guarantees that only blocks descending from the LIB are accepted,
+/// by proactively pruning forks that are diverged before the LIB.
 #[derive(Clone, Debug)]
 pub struct Cryptarchia<Id, State: ?Sized> {
+    /// Blocks that are ancestors of the [`Self::lib`].
+    /// It is guaranteed that these blocks constitute a single, fork-less chain.
+    branches_before_lib: HashMap<Id, Branch<Id>>,
```
zeegomo: why call it branches then? also, we probably can avoid this entirely, anything before the LIB can be discarded by the engine although we don't need to solve it now

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -122,10 +127,30 @@ where
...
+    /// Blocks descending from the [`Self::lib`].
+    /// These blocks may form multiple forks.
+    branches_from_lib: HashMap<Id, Branch<Id>>,
+    /// Tips of the branches descending from the [`Self::lib`],
+    /// including the local chain tip.
+    /// It is guaranteed that these tips are in [`Self::branches_from_lib`].
+    tips_from_lib: HashSet<Id>,
```
zeegomo: it's redundant to specify `from_lib`, we can't have tips or branches before the LIB anyway

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -386,8 +255,52 @@ where
...
         slot: Slot,
     ) -> Result<(Self, PrunedBlocks<Id>), Error<Id>> {
-        let mut new: Self = self.clone();
-        new.branches = new.branches.apply_header(id, parent, slot)?;
+        // The parent must be searched from the [`branches_from_lib`]
+        // to ensure that the new block is a descendant of the LIB.
+        //
+        // If the parent is not found, it means two cases:
+        // 1. The block is part of forks diverged from deeper than LIB.
+        // 2. Not the case 1, but the parent is just not known to the node.
```
zeegomo: nit: remove 'not the case 1', it's implicit in the fact there are two cases

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -403,16 +316,32 @@ where
...
+        if new_lib.id() == self.lib {
+            // No change in LIB, nothing to prune.
+            return PrunedBlocks::new();
+        }
+
+        let lib_depth_from_tip = self
+            .tip_branch()
+            .length()
+            .checked_sub(new_lib.length())
+            .expect("Local chain height must be >= LIB height");
```
zeegomo: nit: the separate function was nice

zeegomo (changes requested):
Not necessarily against merging `Branches` and `Cryptarchia` but providing some background on the design.
`Branches` follows a design similar to the ledger, it's just a tree of child-parent relations. The protocol-specific bits and the knowledge on how to act on those branches (e.g. fork-choice) is instead in `Cryptarchia`. 

The fact that they're separate structs does not mean they can't rely on one another. As long as the functionality to prune / add blocks is closely controlled (e.g. the methods are private and can only be called by `Cryptarchia`), we can assume they're correctly used.

What if instead of merging them we move the check on the LIB back to `Cryptarchia` (just an idea, didn't put to much though into it)?

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -122,10 +127,30 @@ where
...
+    /// including the local chain tip.
+    /// It is guaranteed that these tips are in [`Self::branches_from_lib`].
+    tips_from_lib: HashSet<Id>,
+    /// The tip of the local honest chain.
     local_chain: Branch<Id>,
-    branches: Branches<Id>,
+    /// The latest immutable block (LIB) in the local honest chain.
+    lib: Id,
+    /// Configuration
     config: Config,
```
youngjoon-lee: @zeegomo I'm replying to https://github.com/logos-co/nomos/pull/1404#pullrequestreview-2979364399, in this thread to gather conversations.

I respect the rationale why we designed both `Branches` and `Cryptarchia`. Then, I don't want to break the design by this PR.

> The fact that they're separate structs does not mean they can't rely on one another. As long as the functionality to prune / add blocks is closely controlled (e.g. the methods are private and can only be called by Cryptarchia), we can assume they're correctly used.

I agree with this at some point. Yes, the `Branches` is controlled by `Cryptarchia`. No outside actor mutates `Branches`. So, we can assume that the `Branches` is mutated always correctly by `Cryptarchia`. In other words, each method of `Cryptarchia` mutates the `Branches` **completely**  without introducing any intermediate (temporary) state.

For example, even in the current `master` branch, `Cryptarchia.receive_block` performs the following mutations:
```rust
let mut new: Self = self.clone();
new.branches = new.branches.apply_header(id, parent, slot)?;
new.local_chain = new.fork_choice();
let pruned_blocks = new.update_lib();
```
`Branches.apply_header` adds a new block to the tree, but doesn't update `Branches::lib`. Please note that `lib` is defined in the `Branches` (not `Cryptarchia`). 
But we've assumed it's fine because `Branches::lib` is updated in the end in the same method (`Cryptarchia.receive_block`).

In that case, my initial commit 618efc2d40bbd9c7b4762f5437a60939c35d8a25 is fine. 

However, as you suggested, I think it's better to move `Branches::lib` to `Cryptarchia::lib`. It makes more sense because the `local_chain` is similarly defined in `Cryptarchia`. Then, `Branches` defines only a tree of child-parent relations, without containing any protocol-specific information (such as `lib` and `local_chain`). Also, the `lib` should be always part of `local_chain`.
I think this change makes more sense. Let me try and will get back soon.

youngjoon-lee: I just pushed a new commit, as I said above https://github.com/logos-co/nomos/pull/1404#discussion_r2181671369. @zeegomo 

Please see the entire diff again. It's so so short. This commit addresses all of your short comments above automatically.
(I didn't force-push because I just wanna keep history for someone who's interested in.)

I also updated the PR description.
Thank you for your feedback.

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -386,6 +346,17 @@ where
         parent: Id,
         slot: Slot,
     ) -> Result<(Self, PrunedBlocks<Id>), Error<Id>> {
+        // Check if the block is a descendant of the LIB,
+        // by checking if parent.height >= LIB.height.
+        // This works because all forks diverged before LIB has been pruned.
```
zeegomo: s/has/have

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -386,6 +346,17 @@ where
...
+        let parent_branch = self
+            .branches
+            .get(&parent)
+            .ok_or(Error::ParentMissing(parent))?;
+        if parent_branch.length() < self.lib.length() {
+            return Err(Error::ImmutableFork(parent));
+        }
+
         let mut new: Self = self.clone();
         new.branches = new.branches.apply_header(id, parent, slot)?;
```
zeegomo, line 361: Ok I see why we went the other way, we're now accessing the parent branch twice. 
I don't want to block this forever but I'd also like we keep the quality high here since it's a critical piece, a few ideas:
* branches only contain blocks after the LIB so that it's automatically guaranteed that we don't fork from before (without the additional check)
* merge `Branches` and `Cryptarchia` as you did before but taking into accounts the comments

I trust your judgment, feel free to merge when you're satisfied, we can iterate later

zeegomo (approved):

consensus/cryptarchia-engine/src/lib.rs
danielSanchezQ: :heavy_plus_sign: 1, to me the first option sounds better.

youngjoon-lee: Yeah, I also had the same feeling when writing the parent branch access twice, though it's not a big overhead.

> branches only contain blocks after the LIB so that it's automatically guaranteed that we don't fork from before (without the additional check)

This is the cleanest way but we also need to implement accessing old blocks from storage when necessary (e.g. to provide blocks with bootstrapping nodes). So, I'd suggest implementing this ultimate pruning as a next step soon.

> merge Branches and Cryptarchia as you did before but taking into accounts the comments

IMHO, both my last version and the merge version look reasonable to me. Hm, let me give a thought one more time..

youngjoon-lee: @zeegomo @danielSanchezQ 

Alright, let's go with the first option (what I call _ultimate pruning_). Then the existing [BlockProvider](https://github.com/logos-co/nomos/blob/ddf6875c419c23fd46b212733b8752fc5637061d/nomos-services/chain-service/src/sync/block_provider.rs#L36) logic should be updated to access storage for blocks pruned by engine, but I have an idea how to efficiently do that. That would be even more efficient than the current [BlockProvider](https://github.com/logos-co/nomos/blob/ddf6875c419c23fd46b212733b8752fc5637061d/nomos-services/chain-service/src/sync/block_provider.rs#L36) logic.

I wanted to go with the temporary solution first (either my last version or the merge version) because I don't want to delay implementing the bootstrapping anymore. But I agree that we anyway have to implement the ultimate pruning in near future. So, let's go for it. The bootstrapping can be implemented in parallel, even without the ultimate pruning and the LIB descendant check optimization, but we need to keep in mind that the bootstrapping will cause a lot of long chain traversals if we don't have the LIB check optimization (because LIB is not updated during bootstrapping). Anyway, I will tackle both topics concurrently.

First of all, I will close this PR. Then, I will open a new PR that implements the ultimate pruning. It could seem to be simple to implement the ultimate pruning. But, actually it's not so simple. We need to deal with deleting ledger states and storage blocks carefully. In short, if a pruned block is part of the honest chain, it shouldn't be deleted from storage, but it's okay to remove it from the ledger state. I will nicely implement it in a new PR. 

(FYI, the new PR won't move anything between `Branches` and `Cryptarchia`. That's another topic, and not really necessary to implement the ultimate pruning.)

Then, this PR (LIB descendant check optimization) is not necessary.  

And then, I will modify the existing [BlockProvider](https://github.com/logos-co/nomos/blob/ddf6875c419c23fd46b212733b8752fc5637061d/nomos-services/chain-service/src/sync/block_provider.rs#L36) to handle blocks that are not in memory. I have an idea, as I said above. In short, we may need to store a new key-value `<Slot, HeaderId>` when a block becomes immutable. I will share details soon.

I will leave this PR open by next Monday. Please ping me if you think we need more discussions. Thank you again for your through review. Implementing this PR was simple, but it was not simple to make this safe in our code base. Let's try to jump to the ultimate way. It may seems to delay development, but I believe it will save our time in the end.

PR #1406, opened 2025-06-28
feat(blend): use fork stream instead of custom spawnable solution

danielSanchezQ:

I had an idea on making a forkable stream extension...but seens someone else already had it, so I just integrated it.

Its simpler than previous solition.

@ntn-x2 @danielSanchezQ 

Implementation details only here.

No

## Checklist

* [x] 2. Description added.


nomos-utils/src/lib.rs
```diff
@@ -4,9 +4,6 @@ pub mod math;
 #[cfg(feature = "time")]
 pub mod bounded_duration;
 
-#[cfg(feature = "stream")]
```
ntn-x2, line 7: OK then the `stream` feature and its optional dependencies can go as well.

nomos-blend/scheduling/src/message_scheduler/mod.rs
```diff
@@ -260,12 +219,10 @@ where
                 if Box::pin(setup_new_session(
```
ntn-x2, line 219: Nice! Making this non `async` solves a big problem, so we can just get rid of polling from inside here, which would otherwise require an additional synchronization primitive to avoid polling a new session before the previous setup was completed, which I took notes and wanted to resolve on Monday.

ntn-x2 (commented):
Simple and elegant. I love it. I don't know why I did not find this crate when I did my research. Anyway, looks much better now. Do you want me to take over? Or the you could merge it into the base branch and I can take it from there. But this looks lovely, thank you very much!

danielSanchezQ:
> Simple and elegant. I love it. I don't know why I did not find this crate when I did my research. Anyway, looks much better now. Do you want me to take over? Or the you could merge it into the base branch and I can take it from there. But this looks lovely, thank you very much, although the crate has no commits for two years, so we might need to find an alternative one in the future, but for now it looks fine.

Yeah, it looks a bit unmaintained, but in reality meanwhile it works its fine. Otherwise we may fork it, or just implement the stuff ourselves, its actually quite easy. 

> Do you want me to take over?

We can merge to your branch as this derives from there. So you can take continue doing stuff on top of it. wdyt?

ntn-x2:
Will merge into base branch and pick it up from there. Thanks @danielSanchezQ!

PR #1407, opened 2025-06-28
feat(da): historic membership in subnetwork assignations

pradovic:

This PR implements historic querying for subnetwork assignations in the DA network, addressing the todo! left in previous PRs. It also adds a message to get the subnetwork assignations for a block number in `da_network` service.

Initializing subnetworks from historical data required loading the corresponding addressbook. In this implementation, the addressbook is stored together with the assignations.

There’s an open question: should we rely on the latest addressbook (assuming it retains all past addresses), or is it safer to store historical addressbooks? Additionally, using the latest address might be preferable since peer addresses can change. 

The approach in this PR looked decent for now. Let’s sync on the correct approach, or let's revisit when we implement the incremental update as we should have all needed information how the `FillFromNodeList` will look like. Maybe it would be useful for address book to be a separate entity with a different lifecycle instead of saving it as historic snapshots.

Another option would be to just use `latest_membership_handler.addressbook()` and use it for `init()`, if we can guarantee that latest address book would also contain the addresses for historic peers.

cc: @danielSanchezQ @bacv

Yes
@pradovic 

Yes

Maybe

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/data-availability/network/src/lib.rs
+++ b/nomos-services/data-availability/network/src/lib.rs
@@ -308,6 +308,8 @@ where
                         );
                     });
                 } else {
+                    // todo: handle errors properly when the usage of this function is known
+                    // now we are just logging and returning an empty assignations
                     tracing::warn!("No membership found for block number {block_number}");
                     sender.send(SubnetworkAssignations::default()).unwrap_or_else(|_| {
                         tracing::warn!(


nomos-da/network/subnetworks-assignations/src/lib.rs
```diff
@@ -25,9 +29,9 @@ pub trait MembershipCreator: MembershipHandler {
 
 pub trait MembershipHandler {
     /// Subnetworks Id type
-    type NetworkId: Eq + Hash;
+    type NetworkId: Eq + Hash + Send + Sync + 'static;
     /// Members Id type
-    type Id: Debug;
+    type Id: Debug + Send + Sync + 'static;
```
danielSanchezQ: This probably is not needed in trait level, but when we implement it.

nomos-da/network/subnetworks-assignations/src/lib.rs
```diff
@@ -52,6 +56,8 @@ pub trait MembershipHandler {
 
     /// Returns all subnetworks with assigned members.
     fn subnetworks(&self) -> SubnetworkAssignations<Self::NetworkId, Self::Id>;
+
+    fn addressbook(&self) -> HashMap<Self::Id, Multiaddr>;
```
danielSanchezQ, line 60: This should probably be a reference. Or if we want to clone it for whatever reason an `Arc`, it makes no sense to return something that should be inmutable and just for reference.

nomos-services/data-availability/network/src/lib.rs
```diff
@@ -272,6 +292,26 @@ where
...
+            DaNetworkMsg::SubnetworksAtBlock {
+                block_number,
+                sender,
+            } => {
+                if let Some(membership) = membership_storage.get_historic_membership(block_number) {
+                    let assignations = membership.subnetworks();
+                    sender.send(assignations).unwrap_or_else(|_| {
+                        tracing::warn!(
+                            "client hung up before a subnetwork assignations handle could be established"
+                        );
```
danielSanchezQ, line 308: This should be an error probably (?), it means that the other service was terminated. Maybe not and a warning is enough.

nomos-services/data-availability/network/src/storage/adapters/mock.rs
```diff
@@ -11,17 +11,20 @@ pub struct MockStorageService;
...
+    addressbooks: HashMap<BlockNumber, HashMap<PeerId, Multiaddr>>,
+}
+
 #[derive(Default)]
 pub struct MockStorage {
-    storage: Mutex<HashMap<BlockNumber, Assignations<PeerId, SubnetworkId>>>,
+    state: Mutex<StorageState>,
 }
 
 impl MembershipStorageAdapter<PeerId, SubnetworkId> for MockStorage {
```
danielSanchezQ, line 30: This is fine. I think there is something missing which is a call for prunning old states that wont be required anymore. I can't say now when and how it should be triggered, but we would have to do it eventually.

danielSanchezQ (commented):
This looks good. I would say that next step would be to have the storage plugged to actual storage instead of keeping everything in memory. As it is a simple thing to implement already.

nomos-services/data-availability/network/src/lib.rs
pradovic: I agree. I was thinking about this as well and I am not sure. Maybe we can solve it when we implement a logic that uses this message, it should be more clear then. It should be straightforward to introduce an error

nomos-da/network/subnetworks-assignations/src/lib.rs
pradovic: Membership is behind `ArcSwap`. I am not sure if we can return a reference of a field in the thread safe data structures. Even if it was behind a regular `Mutex` we would still have a same problem I think. 

But this method should be called only on update event, when saving addresbook to storage. Does that makes sense? I'll try to fiddle with it a bit more, maybe we could add addressbook behind an `Arc`, but it would propagate to the trait implementation.

We have a methods on addresbook for reference access, such as:
```
    fn get_address(&self, peer_id: &libp2p::PeerId) -> Option<libp2p::Multiaddr> {
        self.membership.load().get_address(peer_id)
    }
```

danielSanchezQ: Ok, we can clone it for now I guess. I remember we already had this convo anyway :sweat_smile:

nomos-services/data-availability/network/src/lib.rs
danielSanchezQ: Add a todo on what would be needed, so when coming back to the code we relate.

danielSanchezQ (approved):

bacv (approved):

nomos-da/network/subnetworks-assignations/src/lib.rs
pradovic: Yes, I think so as well. I still don't have a better approach, but I will think about it. I am not sure we can avoid clone because we are actually saving it to storage. This is where we use this call. Let's see :)

PR #1414, opened 2025-07-01
feat(blend): connection handlers for core-edge and edge-core communication

ntn-x2:

WIP. PR description will follow.

> [!IMPORTANT]  
> explained.  

> [!IMPORTANT]  
> advancing  
> the review.

> [!IMPORTANT]

> [!IMPORTANT]
    modification/solutions.

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


nomos-blend/network/src/handler/edge/core_edge/starting.rs
```diff
@@ -0,0 +1,75 @@
...
+use crate::handler::edge::core_edge::{
+    dropped::DroppedState, ready_to_receive::ReadyToReceiveState, ConnectionEvent, ConnectionState,
+    FailureReason, PollResult, StateTrait, LOG_TARGET,
+};
+
+/// Entrypoint to start receiving a single message from an edge node.
+pub struct StartingState {
+    /// Time after which the connection will be closed. It is started as soon as
+    /// the inbound stream is negotiated.
+    connection_timeout: Duration,
```
youngjoon-lee, line 18: What do you think about naming this as `idle_timeout` or `idle_connection_timeout`, which matches well with: (not a strong suggestion.) https://github.com/logos-co/nomos/blob/e0f76d58e9ea249acd68f3ed0e87677f903902c7/nomos-blend/network/src/handler/edge/core_edge/ready_to_receive.rs#L18-L20


Also, libp2p has the [`with_idle_connection_timeout`](https://github.com/libp2p/rust-libp2p/blob/1aa43375c3b05b84de985b84a2ae0caa9efa3940/examples/file-sharing/src/network.rs#L64) which can be set to `Swarm`. But, if I understand correctly, **our timeout** is for different purposes. Instead of closing the **connection**, we just move the connection to the `DroppedState`, which does nothing with the connection. If the connection is not used for a long time by any `NetworkBehaviour`s, it will be closed by `Swarm` if we set `with_idle_connection_timeout`. I think this is a good design. Do I understand this correctly?

nomos-blend/network/src/handler/edge/core_edge/ready_to_receive.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn from(value: ReadyToReceiveState) -> Self {
+        Self::ReadyToReceive(value)
+    }
+}
+
+impl StateTrait for ReadyToReceiveState {
+    // If the timer elapses, moves the state machine to `DroppedState` with a
+    // timeout error. Otherwise, it moves it to `ReceivingState` with the `recv_msg`
+    // future to fetch the first message from the stream.
+    fn poll(mut self, cx: &mut Context<'_>) -> PollResult<ConnectionState> {
```
youngjoon-lee, line 51: Asking just to understand this correctly.

So, as long as the timeout is not 0, does `ReadyToReceiveState` transition immediately to the `ReceivingState`? 
(assuming that `poll()` is called immediately without any delay).

nomos-blend/network/src/handler/edge/core_edge/receiving.rs
```diff
@@ -0,0 +1,84 @@
...
+            return (
+                Poll::Pending,
+                DroppedState::new(Some(FailureReason::Timeout), Some(cx.waker().clone())).into(),
+            );
+        };
+        let Poll::Ready(message_receive_result) = self.incoming_message.poll_unpin(cx) else {
+            // We don't wake here since the `incoming_message` future will wake the waker
+            // when completed.
+            return (Poll::Pending, self.into());
+        };
```
youngjoon-lee, line 63: Probably, do we need to reset the `self.timeout_timer` here once receiving a message?

nomos-blend/network/src/handler/edge/core_edge/starting.rs
ntn-x2: Yeah that serves a different purpose. I also thought about implementing the `connection_keepalive` method on `ConnectionHandler`, but that also was for a different reason. In our case, regardless of the activity going on inside the substream, if the message is not received by the time the timeout triggers, the stream is dropped, and consequently the connection closed by the swarm, as you say. While the other methods apply to idle streams, which is slightly different to what we need.

nomos-blend/network/src/handler/edge/core_edge/ready_to_receive.rs
ntn-x2: Yes, that is correct. This state takes care of "spawning" the download task and mapping its return value, then transition into the next state.

nomos-blend/network/src/handler/edge/core_edge/receiving.rs
ntn-x2: If we receive a message before the timer goes off, we move to the next state and the timer is dropped, so we don't need to manually reset. If, on the other hand, we have still not received the full message, we still need to poll the timer since we want to close the stream if the message takes too long to receive. In either case, I think we don't need to care about resetting the timer.

youngjoon-lee: Ah right. I forgot that this is an oneshot connection from an edge node. Yeah, I just checked that the state transitions to the DroppedState as soon as a message is received. Looks good

PR #1415, opened 2025-07-01
feat(chain): network adapter logic for orphan blocks download

andrussal:

This PR adds functionality to consensus service network adapter to request downloading missing blocks in order to connect an orphan block with local chain. 

It chooses a random peer from all discovered peers that are not connected, and falls back to connected peers if such peers don’t exist. 

Note: This PR not implementing the actual logic to trigger downloading missing blocks for an orphan block and adding them to the block tree.

@andrussal 

N/A

N/A

## Checklist

* [x] 1. Description added.

### DIFF
--- a/nomos-libp2p/src/behaviour/kademlia/behaviour_ext.rs
+++ b/nomos-libp2p/src/behaviour/kademlia/behaviour_ext.rs
@@ -1,6 +1,9 @@
 use std::collections::HashMap;
 
-use libp2p::{kad::QueryId, Multiaddr, PeerId, StreamProtocol};
+use libp2p::{
+    kad::{PeerInfo, QueryId},
+    Multiaddr, PeerId, StreamProtocol,
+};
 
 use crate::behaviour::Behaviour;
 
@@ -30,5 +33,30 @@ impl Behaviour {
 
     pub(crate) fn kademlia_get_closest_peers(&mut self, peer_id: PeerId) -> QueryId {
         self.kademlia.get_closest_peers(peer_id)
+    }
+
+    pub(crate) fn kademlia_discovered_peers(&mut self) -> Vec<PeerInfo> {
+        let Some(kademlia) = self.kademlia.as_mut() else {
+            tracing::error!("kademlia is not enabled");
+            return Vec::new();
+        };
+
+        // get all buckets and in each buket, peers with addresses
+        kademlia
+            .kbuckets()
+            .flat_map(|bucket| {
+                bucket
+                    .iter()
+                    .filter_map(|entry| {
+                        let peer_id = *entry.node.key.preimage();
+                        let addresses: Vec<_> = entry.node.value.iter().cloned().collect();
+                        (!addresses.is_empty()).then_some(PeerInfo {
+                            peer_id,
+                            addrs: addresses,
+                        })
+                    })
+                    .collect::<Vec<_>>()
+            })
+            .collect()
     }
 }
--- a/nomos-libp2p/src/behaviour/kademlia/behaviour_ext.rs
+++ b/nomos-libp2p/src/behaviour/kademlia/behaviour_ext.rs
@@ -36,13 +36,8 @@ impl Behaviour {
     }
 
     pub(crate) fn kademlia_discovered_peers(&mut self) -> Vec<PeerInfo> {
-        let Some(kademlia) = self.kademlia.as_mut() else {
-            tracing::error!("kademlia is not enabled");
-            return Vec::new();
-        };
-
         // get all buckets and in each buket, peers with addresses
-        kademlia
+        self.kademlia
             .kbuckets()
             .flat_map(|bucket| {
                 bucket
--- a/nomos-services/chain-service/src/network/adapters/libp2p.rs
+++ b/nomos-services/chain-service/src/network/adapters/libp2p.rs
@@ -1,10 +1,13 @@
-use std::{collections::HashSet, marker::PhantomData};
+use std::{collections::HashSet, fmt::Debug, hash::Hash, marker::PhantomData};
 
 use cryptarchia_sync::GetTipResponse;
 use futures::TryStreamExt as _;
 use nomos_core::{block::Block, header::HeaderId, wire};
 use nomos_network::{
-    backends::libp2p::{ChainSyncCommand, Command, Libp2p, PeerId, PubSubCommand::Subscribe},
+    backends::libp2p::{
+        ChainSyncCommand, Command, DiscoveryCommand, Libp2p, NetworkCommand, PeerId,
+        PubSubCommand::Subscribe,
+    },
     message::{ChainSyncEvent, NetworkMsg},
     NetworkService,
 };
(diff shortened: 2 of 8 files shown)


nomos-services/cryptarchia-consensus/src/network/mod.rs
```diff
@@ -42,4 +42,12 @@ pub trait NetworkAdapter<RuntimeServiceId> {
...
         additional_blocks: HashSet<HeaderId>,
     ) -> Result<BoxedStream<Result<Block<Self::Tx, Self::BlobCertificate>, DynError>>, DynError>;
+
+    async fn request_missing_blocks_for_orphan(
+        &self,
+        target_block: HeaderId,
+        local_tip: HeaderId,
+        latest_immutable_block: HeaderId,
+        additional_blocks: HashSet<HeaderId>,
+    ) -> Result<BoxedStream<Result<Block<Self::Tx, Self::BlobCertificate>, DynError>>, DynError>;
```
youngjoon-lee: The function name could be more generalized, like `request_blocks_from_new_peers` or so. From the perspective of the network adapter, it doesn't need to know that this function is called for handling orphan blocks. The caller may want to download blocks from a newly discovered peer for different purposes (in the future). So, I think we don't need to limit the purpose of this function.

nomos-services/cryptarchia-consensus/src/network/adapters/libp2p.rs
```diff
@@ -187,4 +233,147 @@ where
...
+
+    let remaining = max.saturating_sub(selected.len());
+    let remaining_available = remaining.min(connected_peers.len());
+
+    if remaining_available > 0 {
+        let connected_vec: Vec<_> = connected_peers.into_iter().collect();
+
+        let indexes = sample(&mut thread_rng(), connected_vec.len(), remaining_available);
+        selected.extend(indexes.into_iter().map(|i| connected_vec[i]));
+    }
```
youngjoon-lee:
```suggestion
    let mut selected = discovered_only
        .iter()
        .copied()
        .choose_multiple(&mut thread_rng(), max);

    let remaining = max.saturating_sub(selected.len());
    if remaining > 0 {
        selected.extend(
            connected_peers
                .iter()
                .copied()
                .choose_multiple(&mut thread_rng(), remaining),
        );
    }
```

This could be simplified like this.

andrussal: Thank you, this is easier to read.

nomos-libp2p/src/behaviour/kademlia/behaviour_ext.rs
```diff
@@ -46,4 +49,29 @@ impl Behaviour {
         };
         Ok(kademlia.get_closest_peers(peer_id))
     }
+
+    pub(crate) fn kademlia_discovered_peers(&mut self) -> Vec<PeerInfo> {
+        let Some(kademlia) = self.kademlia.as_mut() else {
+            tracing::error!("kademlia is not enabled");
+            return Vec::new();
+        };
```
danielSanchezQ: This should actually never happen unless in testing mode right? Maybe at some point we would have to remove this stuff. I do not like that we introduce this stuff than would give us headaches later on releasing. I think we should probably panic here (?)

nomos-libp2p/src/behaviour/kademlia/behaviour_ext.rs
```diff
@@ -46,4 +49,29 @@ impl Behaviour {
...
+                    .iter()
+                    .filter_map(|entry| {
+                        let peer_id = *entry.node.key.preimage();
+                        let addresses: Vec<_> = entry.node.value.iter().cloned().collect();
+                        (!addresses.is_empty()).then_some(PeerInfo {
+                            peer_id,
+                            addrs: addresses,
+                        })
+                    })
+                    .collect::<Vec<_>>()
```
danielSanchezQ, line 53: unnecessary collect

nomos-services/chain-service/src/network/adapters/libp2p.rs
```diff
@@ -1,25 +1,32 @@
...
 use tokio::sync::oneshot;
 use tokio_stream::{wrappers::errors::BroadcastStreamRecvError, StreamExt as _};
+use tracing::debug;
 
 use crate::{
     messages::NetworkMessage,
     network::{BoxedStream, NetworkAdapter},
 };
 
+const MAX_PEERS_TO_TRY_FOR_ORPHAN_DOWNLOAD: usize = 3;
```
danielSanchezQ, line 29: Is this defined in the specification or it is just an arbitrary number?

nomos-services/chain-service/src/network/adapters/libp2p.rs
```diff
@@ -187,4 +233,132 @@ where
...
+            {
+                Ok(stream) => {
+                    debug!("Requested orphan parents from peer: {peer}");
+                    return Ok(stream);
+                }
+                Err(err) => {
+                    tracing::warn!("Failed to request orphan parents from peer {peer}: {err}");
+                }
+            }
+        }
```
danielSanchezQ, line 274: Shouldn't we just try all of them and then fuse the futures on first arrival instead of doing it sequentially?

nomos-libp2p/src/behaviour/kademlia/behaviour_ext.rs
andrussal: I think we should be able to remove them, can't see why it should break test. But perhaps in another PR.

nomos-services/chain-service/src/network/adapters/libp2p.rs
andrussal: This is just arbitrary.

nomos-services/chain-service/src/network/adapters/libp2p.rs
andrussal: I think it makes sense to try all of them. But it might be better to keep it simple in first iteration and do it another PR after we have integration tests for sync.

danielSanchezQ: ~~Let's open an issue for it then. Otherwise it will just be forgotten.~~ it was done, so better :smiley:

nomos-services/chain-service/src/network/adapters/libp2p.rs
youngjoon-lee: I think this doesn't need to be defined in the spec, because this number only affects this node. In the spec, we're mentioning that implementers can introduce their own retry logic.

nomos-services/chain-service/src/network/adapters/libp2p.rs
zeegomo: won't the local tip change after receiving blocks from a peer? also, what's the target block? Isn't it each peer's tip?

andrussal: Here the goal is to try multiple peers and return first successful response stream. We assume that other peers can also miss parents of the orphan because we received orphan from one of them, so we try many. 

> won't the local tip change after receiving blocks from a peer?

The caller of this function will be responsible giving each time new local tip.

> also, what's the target block? Isn't it each peer's tip?

Caller needs to set `taget_block` to the orphan block, so we should get back stream of blocks from a know block to the orphan block.

PR #1419, opened 2025-07-04
feat(cryptarchia): prune all blocks before LIB

youngjoon-lee:

This is the new version of https://github.com/logos-co/nomos/pull/1404 (to be closed soon). After [long discussions](https://github.com/logos-co/nomos/pull/1404#discussion_r2183189878), I came up with this final version, which I believe is optimal.

### Purpose
We want to optimize the LIB descendant check in the block validation. It's important because the current way (`is_ancestor(lib, block)`) is expensive especially during bootstrapping (which doesn't update LIB for a long time).
Similar as proposed in the [spec](https://www.notion.so/nomos-tech/Cryptarchia-v1-Protocol-Specification-21c261aa09df810cb85eff1c76e5798c?source=copy_link#21c261aa09df81978b9df055c4b8e67a), we can optimize the check to O(1).

### How to optimize?
This PR updates `cryptarchia_engine::Cryptarchia` to prune all blocks deeper than the new LIB whenever the LIB is updated. Previously, we have been pruning only stale forks diverged from deeper than LIB, but this PR prunes more blocks which are in the honest chain. 
By doing this, we can guarantee that `cryptarchia_engine::Cryptarchia` always hold only blocks descending from LIB (all active forks as well as the honest chain).
It means that we don't need any additional mechanism for the LIB descendant check. If the parent of a new block can be found in the `Cryptarchia`, it means that the new block is a descendant of LIB. It's even more efficient than what the [spec](https://www.notion.so/nomos-tech/Cryptarchia-v1-Protocol-Specification-21c261aa09df810cb85eff1c76e5798c?source=copy_link#21c261aa09df81978b9df055c4b8e67a) is proposing.

### Important notes
Now we have two kinds of pruned blocks:
https://github.com/logos-co/nomos/blob/81a4f3ec3051b36fa9d8fa8e130775715e19e65f/consensus/cryptarchia-engine/src/lib.rs#L543-L551
Both block sets can be pruned from the ledger state (in the upper layer). But, when pruning blocks in storage, we delete only `stale_blocks` from storage. We don't want to delete `immutable_blocks` from storage because we need historical blocks for some purpose (e.g. providing blocks with bootstrapping peers). This PR handles this carefully.

Another important note is that this PR breaks [sync::BlockProvider](https://github.com/logos-co/nomos/blob/81a4f3ec3051b36fa9d8fa8e130775715e19e65f/nomos-services/chain-service/src/sync/block_provider.rs#L36), which accesses historical blocks only from memory for supporting the bootstrapping peers. I left a [TODO comment](https://github.com/logos-co/nomos/blob/81a4f3ec3051b36fa9d8fa8e130775715e19e65f/nomos-services/chain-service/src/sync/block_provider.rs#L69) that should be done in the next PR. I will take ownership of it.

Yes

spec: @davidrusu 
code: @zeegomo @youngjoon-lee 

Yes

This PR takes a better way than what the [spec](https://www.notion.so/nomos-tech/Cryptarchia-v1-Protocol-Specification-21c261aa09df810cb85eff1c76e5798c?source=copy_link#21c261aa09df81978b9df055c4b8e67a) is proposing. But, I think we don't need to update the spec because the spec delivered the optimization idea clearly in a generalized way.

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/consensus/cryptarchia-engine/src/lib.rs
+++ b/consensus/cryptarchia-engine/src/lib.rs
@@ -799,12 +799,12 @@ pub mod tests {
     // canonical chain length.
     #[test]
     fn pruning_too_back_in_time() {
-        // Create a chain with 50+1 blocks with k=51.
+        // Create a chain with 50+1 blocks with k=50.
         // b0(LIB) - b1 - ... - b49
         //         \
         //          b100
         let (cryptarchia, pruned_blocks) =
-            create_canonical_chain(50.try_into().unwrap(), Some(config_with(51)))
+            create_canonical_chain(50.try_into().unwrap(), Some(config_with(50)))
                 // Add a fork from genesis block
                 .receive_block(hash(&100u64), hash(&0u64), 1.into())
                 .expect("test block to be applied successfully.");
@@ -823,6 +823,25 @@ pub mod tests {
         assert!(pruned_blocks.all().next().is_none());
         assert!(cryptarchia.branches.tips.contains(&hash(&100u64)));
         assert!(cryptarchia.branches.branches.contains_key(&hash(&100u64)));
+
+        // Add two new blocks to the local honest chain,
+        // and check if the LIB is updated and blocks are pruned.
+        let (cryptarchia, pruned_blocks) = cryptarchia
+            .receive_block(hash(&50u64), hash(&49u64), 50.into())
+            .expect("test block to be applied successfully.")
+            .0
+            .receive_block(hash(&51u64), hash(&50u64), 51.into())
+            .expect("test block to be applied successfully.");
+        // The LIB was updated to b1.
+        assert_eq!(cryptarchia.lib(), hash(&1u64));
+        // The stale fork b100 was pruned.
+        assert_eq!(pruned_blocks.stale_blocks, [hash(&100u64)].into());
+        assert!(!cryptarchia.branches.tips.contains(&hash(&100u64)));
+        assert!(!cryptarchia.branches.branches.contains_key(&hash(&100u64)));
+        // The immutable block b0 was pruned.
+        assert_eq!(pruned_blocks.immutable_blocks, [hash(&0u64)].into());
+        assert!(!cryptarchia.branches.tips.contains(&hash(&0u64)));
+        assert!(!cryptarchia.branches.branches.contains_key(&hash(&0u64)));
     }
 
     #[test]


consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -408,7 +383,12 @@ where
             PrunedBlocks::new()
         } else {
             self.branches.lib = new_lib;
-            self.prune_forks(self.lib_depth()).collect()
+            PrunedBlocks {
+                // TODO: Eliminate the need of `lib_depth` by refactoring `prune_stale_forks`,
+                //       similar as `prune_immutable_blocks`.
```
danielSanchezQ, line 388: Lets open an issue if its not gonna be done in this PR.

danielSanchezQ (approved):

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -549,7 +540,53 @@ where
...
 
-pub type PrunedBlocks<Id> = HashSet<Id>;
+/// Represents blocks that have been pruned because they are no longer needed
+/// for future block validations.
+pub struct PrunedBlocks<Id> {
+    /// Blocks from the stale forks diverged before the LIB.
+    stale_blocks: HashSet<Id>,
+    /// Immutable blocks that were deeper than the LIB,
+    /// excluding the LIB itself.
+    immutable_blocks: HashSet<Id>,
```
zeegomo, line 550: just a thought, could it be helpful to maintain order here

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -593,7 +630,7 @@ pub mod tests {
         length: NonZero<u64>,
         c: Option<Config>,
     ) -> Cryptarchia<[u8; 32], Boostrapping> {
-        let mut engine = Cryptarchia::from_lib([0; 32], c.unwrap_or_else(config));
+        let mut engine = Cryptarchia::from_lib(hash(&0u64), c.unwrap_or_else(config));
```
zeegomo, line 633: why this?

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -811,203 +799,298 @@ pub mod tests {
...
 
-        // But, no block is pruned because `security_param` is
+        // But, no block was pruned because `security_param` is
         // greater than local chain length.
-        assert!(pruned_blocks.is_empty());
-        assert!(cryptarchia.branches.tips.contains(&[100; 32]));
-        assert!(cryptarchia.branches.branches.contains_key(&[100; 32]));
+        assert!(pruned_blocks.all().next().is_none());
+        assert!(cryptarchia.branches.tips.contains(&hash(&100u64)));
+        assert!(cryptarchia.branches.branches.contains_key(&hash(&100u64)));
```
zeegomo, line 825: nit: can we try to apply one more block and confirm the LIB advances?

zeegomo:
Good work!

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: Yeah, thanks. Opened https://github.com/logos-co/nomos/issues/1446

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: This is just for using the consistent way for building header IDs in this testing module. So far, we've use both `[X; 32]` and `hash(&Xu64)` interchangeably. I thought it's confusing. So I standardized everything to `hash(&Xu64)`. 

Now, I think we don't actually need to have the `hash()` unless we create more than 256 block in the tests. But, I'm not removing `hash()` in this PR. I will try it as a separate PR and check if it's better.

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: Currently, the order is not so useful. But it depends on the needs of callers in the future. 

I improved this to `Vec<(Slot, Id)>` in [the next PR](https://github.com/logos-co/nomos/pull/1424/files#diff-9c893a274d5c71bc8958e4aa3df21f25402092aa00fc14ecf8a9b33d61b2bf29R550) (to where Daniel already left a comment).

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: Good point. I added it: 4357a4081fa34adeb2762a5e24bb9c55f740cee5

PR #1421, opened 2025-07-04
fix(storage): check upper bound in load_prefix

youngjoon-lee:

`RocksBackend::load_prefix` was not working as intended.

For example, if there are several keys as below:
```
boo/0
foo/0
foo/1
zoo/0
```
If we call `load_prefix("foo/")`, it currently returns 3 items that start with `foo/` or `zoo/`). However, we want to get only 2 items that start with `foo/`.

I found this bug while implementing the chain bootstrapping.

We didn't have any unit test for `load_prefix`. I just added a test, so that you can easily reproduce this based on the previous code.

The reason is that we're using [`rockdb.DBCommon.prefix_iterator`](https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.prefix_iterator), but it works only if prefix extractors are configured for column families. If not, it works the same as the regular [`DBCommon.iterator`](https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.iterator), which doesn't check any upper bound. That's why `zoo/0` is included in the result. 

Since the [doc](https://docs.rs/rocksdb/latest/rocksdb/struct.ReadOptions.html#method.set_prefix_same_as_start) has a very short description about this, I don't fully understand yet how to make it work. 

Instead, it seems clearer to use the regular iterator and check the upper bound manually.

Later, we can dig into the [Prefix Seek](https://github.com/facebook/rocksdb/wiki/Prefix-Seek) to see if there's any optimized way. Also, in the current code base, we're anyway [misusing](https://github.com/logos-co/nomos/blob/cf416848620d28e9ede8d948b0df4ed9f3672c3f/tests/src/nodes/executor.rs#L378) the column family. We need to either drop column family or figure out how to use it properly.
Thus, for now, I believe just using regular iterator is fine.

(FYI, this change is also used for a next PR that will add a new argument `start_key: Option<&[u8]>`, which is used along with `prefix`). 

 
Yes

No codeowner, but asking @bacv for his review 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/storage/src/backends/rocksdb.rs
+++ b/nomos-services/storage/src/backends/rocksdb.rs
@@ -2,7 +2,7 @@ use std::{marker::PhantomData, path::PathBuf, sync::Arc};
 
 use async_trait::async_trait;
 use bytes::Bytes;
-use rocksdb::{Error, Options, DB};
+use rocksdb::{Direction, Error, IteratorMode, Options, DB};
 use serde::{Deserialize, Serialize};
 
 use super::{StorageBackend, StorageSerde, StorageTransaction};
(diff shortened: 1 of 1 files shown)


bacv (approved):
A very good. My bad for not having a test for it. Thank you!

PR #1422, opened 2025-07-04
refactor: remove behaviour toggles

andrussal:

This PR removes `Toggle` from Identity and Kademlia behaviours because we don’t have a use case when they should be disabled.

Yes.

@andrussal 

N/A

N/A

## Checklist

* [x] 1. Description added.

### DIFF
--- a/nomos-services/network/src/backends/libp2p/swarm/kademlia.rs
+++ b/nomos-services/network/src/backends/libp2p/swarm/kademlia.rs
@@ -45,25 +45,15 @@ impl SwarmHandler {
     pub(super) fn handle_discovery_command(&mut self, command: DiscoveryCommand) {
         match command {
             DiscoveryCommand::GetClosestPeers { peer_id, reply } => {
-                match self.swarm.get_closest_peers(peer_id) {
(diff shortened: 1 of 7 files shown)


nomos-services/network/src/backends/libp2p/swarm/kademlia.rs
```diff
@@ -45,25 +45,15 @@ impl SwarmHandler {
...
-                    }
-                    Err(err) => {
-                        tracing::warn!(
-                            "Failed to get closest peers for peer ID: {peer_id}: {:?}",
-                            err
-                        );
-                        let _ = reply.send(Vec::new());
-                    }
-                }
+                let query_id = self.swarm.get_closest_peers(peer_id);
```
ntn-x2, line 48: What was the reason to remove the error trace and sending back the result?

andrussal: Here the only reason for error was `Toggle` returning `None` for the behaviour, then we returned `OperationNotSupported`.  `get_closest_peers` itself just returns `QueryId`.

PR #1423, opened 2025-07-04
feat(nat): upnp address mapping protocol

andrussal:

This PR adds UpnP address mapping protocol to nat address mapper. 

Libp2p has official UpnP behaviour but this doesn’t fit with our state machine logic. We request mapping after autonat client fails to detect access. But libp2p UpnP runs automatically on `NewListenAddr` event from Swarm. This PR uses the same underlying `igd-next` crate though.

Note on testing: Comes out terribly hard. Best approach is to install `miniupnpd` because it  should support all our desired mapping protocols. But I haven’t been able to get it working so far. Have to run in Docker container(works only on linux), and even Docker environment seems to require some Linux specific host features.

Yes. [NAT Spec](https://www.notion.so/Nomos-P2P-Network-NAT-Solution-1d28f96fb65c80198b03f777be835967)

@andrussal 

N/A

N/A

## Checklist

* [x] 1. Description added.


nomos-libp2p/src/behaviour/nat/address_mapper/protocol.rs
```diff
@@ -0,0 +1,49 @@
+use libp2p::Multiaddr;
+
+use crate::behaviour::nat::address_mapper::{errors::AddressMapperError, upnp::UpnpProtocol};
+
+/// Trait for NAT mapping protocols
+#[async_trait::async_trait]
+pub trait MappingProtocol: Send + Sync + 'static {
+    /// Initialize the protocol
+    async fn initialize(&mut self) -> Result<(), AddressMapperError>;
```
ntn-x2: Mhh, why not something like `initialize() -> Result<Self, AddressMapperError>`?

nomos-libp2p/src/behaviour/nat/address_mapper/protocol.rs
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    pub async fn try_map_address(
+        &mut self,
+        address: &Multiaddr,
+    ) -> Result<Multiaddr, AddressMapperError> {
+        let external_address = self.upnp.map_address(address).await?;
+
+        tracing::info!(
+            "Successfully mapped {} to {} using {}",
```
ntn-x2, line 34: But what we get is a `Result` right? So it's not necessarily a success here.

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+/// UPnP-based TCP address mapper
+#[derive(Default)]
+pub struct AddressMapperBehaviour {
+    mapping_future: Option<BoxFuture<'static, Result<Multiaddr, AddressMapperError>>>,
+    original_address: Option<Multiaddr>,
+    waker: Option<Waker>,
+}
+
+impl AddressMapperBehaviour {
+    pub fn try_map_address(&mut self, address: Multiaddr) {
```
ntn-x2: Ideally, this would also be typesafe so that it would not be able to be called more than once. But, if we do not want to go that far, at the very least we should return a `Result` with an error if we try to map more than once, since this is a no-op if `self.mapping_future.is_some()`.

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+    ) {
+    }
+
+    fn poll(
+        &mut self,
+        cx: &mut Context<'_>,
+    ) -> Poll<ToSwarm<Self::ToSwarm, THandlerInEvent<Self>>> {
+        self.waker = Some(cx.waker().clone());
+
+        if let Some(mut mapping_future) = self.mapping_future.take() {
```
ntn-x2: I would suggest to use a `if let Some(mut mapping_future) = self.mapping_future.take() else {...}` so we early return `Poll::Pending` and avoid one layer of indentation.

nomos-libp2p/src/behaviour/nat/address_mapper/upnp.rs
```diff
@@ -0,0 +1,97 @@
...
+
+#[async_trait::async_trait]
+impl MappingProtocol for UpnpProtocol {
+    async fn initialize(&mut self) -> Result<(), AddressMapperError> {
+        let gateway = igd_next::aio::tokio::search_gateway(SearchOptions::default()).await?;
+
+        let gateway_external_ip = gateway
+            .get_external_ip()
+            .await
+            .map_err(|e| AddressMapperError::ExternalIpFailed(e.to_string()))?;
```
ntn-x2: I think we implement `From` for this error type, right? So propagating the error with `?` should be enough as for the gateway above.

nomos-libp2p/src/behaviour/nat/address_mapper/upnp.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        self.gateway = Some(gateway);
+        self.gateway_external_ip = Some(gateway_external_ip);
+
+        Ok(())
+    }
+
+    async fn map_address(&mut self, address: &Multiaddr) -> Result<Multiaddr, AddressMapperError> {
+        let gateway = self
```
ntn-x2: We wouldn't have to do any of this if we return `Self` in `initialize`, and we don't use `Option`s for the struct definition.

nomos-libp2p/src/behaviour/nat/address_mapper/upnp.rs
```diff
@@ -0,0 +1,97 @@
...
+            Protocol::Ip4(addr) => Some(IpAddr::V4(addr)),
+            Protocol::Ip6(addr) => Some(IpAddr::V6(addr)),
+            _ => None,
+        })
+        .ok_or(AddressMapperError::NoIpAddress)?;
+
+    let port = addr
+        .iter()
+        .find_map(|protocol| match protocol {
+            Protocol::Tcp(p) => Some(p),
```
ntn-x2: Do we assume TCP is the only transport protocol supported/expected? If so, why? And a comment on top of the function would help.

nomos-libp2p/src/behaviour/nat/address_mapper/upnp.rs
```diff
@@ -0,0 +1,97 @@
...
+            .add_port(
+                PortMappingProtocol::TCP,
+                // Request the same port as the internal address
+                mapped_port,
+                internal_address,
+                0,
+                "libp2p UPnP mapping",
+            )
+            .await
+            .map_err(|e| AddressMapperError::PortMappingFailed(e.to_string()))?;
```
ntn-x2: I guess we can implement `From<AddPortError>` as well?

ntn-x2 (commented):
This is good! I have a few comments about some design decisions, and a question about the inner behaviour of UPnP.

nomos-libp2p/src/behaviour/nat/address_mapper/protocol.rs
andrussal: Here it returns early if error.

nomos-libp2p/src/behaviour/nat/address_mapper/upnp.rs
```diff
@@ -0,0 +1,86 @@
...
+                // Request the same port as the internal address
+                mapped_port,
+                internal_address,
+                0,
+                "libp2p UPnP mapping",
+            )
+            .await?;
+
+        let port = mapped_port;
+        let external_addr = format!("/ip4/{}/tcp/{port}", self.gateway_external_ip);
```
ntn-x2: I guess if we use UDP as well this address will be different, right?

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
danielSanchezQ: yeah, at least a log that we are trying to call it when was called already

danielSanchezQ (approved):
Looking pretty good. Just a single comment with a +1 on what @ntn-x2 mentioned. Otherwise is good to go. Thanks!

PR #1424, opened 2025-07-07
feat(chain): store pruned immutable blocks permanently

youngjoon-lee:

(This is a stacked PR based on #1419).

#1419 prunes all immutable blocks before LIB from the engine. 
This PR stores them to the storage.

- Key: `immutable_block/slot/{slot.to_be_bytes()}`
- Value: `HeaderId`

This key-value will be used by the bootstrapping block provider that should access historical blocks. Using `Slot` as a key is important for that.

Yes

@zeegomo @youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/.github/actions/install-risc0/action.yml
+++ b/.github/actions/install-risc0/action.yml
@@ -27,7 +27,7 @@ runs:
       # * .github/workflows/code-check.yml (the `risc0-clippy` within the `lints` job)
       run: |
         cargo install rzup --version 0.4.1 --locked \
-          && ~/.cargo/bin/rzup install cargo-risczero 2.2.0 \
+          && ~/.cargo/bin/rzup install cargo-risczero 2.3.0 \
           && ~/.cargo/bin/rzup install rust 1.88.0
       shell: bash
       env:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -107,7 +107,7 @@ libp2p-stream    = "0.3.0-alpha"
 overwatch        = { git = "https://github.com/logos-co/Overwatch", rev = "f5a9902" }
 overwatch-derive = { git = "https://github.com/logos-co/Overwatch", rev = "f5a9902" }
 reqwest          = "0.12"
-risc0-zkvm       = "2.2.0"
+risc0-zkvm       = "2.3.0"
 serde_with       = "3.12.0"
 tracing          = "0.1"
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ RUN apt-get update && apt-get install -yq \
 
 RUN cargo install cargo-binstall --locked
 RUN cargo install rzup --version 0.4.1 --locked
-RUN rzup install cargo-risczero 2.2.0
+RUN rzup install cargo-risczero 2.3.0
 RUN rzup install rust 1.88.0
 
 RUN cargo build --release -p nomos-node
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -25,5 +25,5 @@ USER jenkins
 
 RUN cargo install cargo-binstall --locked
 RUN cargo install rzup --version 0.4.1 --locked
-RUN /home/jenkins/.cargo/bin/rzup install cargo-risczero 2.2.0
+RUN /home/jenkins/.cargo/bin/rzup install cargo-risczero 2.3.0
 RUN /home/jenkins/.cargo/bin/rzup install rust 1.88.0
--- a/nomos-core/risc0_proofs/Cargo.toml
+++ b/nomos-core/risc0_proofs/Cargo.toml
@@ -5,7 +5,7 @@ name    = "nomos_risc0_proofs"
 version = "0.1.0"
 
 [build-dependencies]
-risc0-build = "2.2.0"
+risc0-build = "2.3.0"
 
 [package.metadata.risc0]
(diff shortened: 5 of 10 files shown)


consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -547,7 +547,7 @@ pub struct PrunedBlocks<Id> {
     stale_blocks: HashSet<Id>,
     /// Immutable blocks that were deeper than the LIB,
     /// excluding the LIB itself.
-    immutable_blocks: HashSet<Id>,
+    immutable_blocks: Vec<(Slot, Id)>,
```
danielSanchezQ: Why vec instead of the set?

nomos-core/chain-defs/src/header/mod.rs
```diff
@@ -147,6 +147,19 @@ impl From<HeaderId> for [u8; 32] {
     }
 }
 
+impl TryFrom<&[u8]> for HeaderId {
+    type Error = ();
```
danielSanchezQ: `()` doesn't implement error which will make it difficult later if we have to use `?`. Even something generic would be better. We could return the length of the input as a hint on where we fucked up :sweat_smile:

nomos-services/chain-service/src/storage/adapters/storage.rs
```diff
@@ -105,4 +106,19 @@ where
...
+        blocks: impl Iterator<Item = (Slot, HeaderId)> + Send,
+    ) -> Result<(), overwatch::DynError> {
+        for (slot, header_id) in blocks {
+            self.storage_relay
+                .send(StorageMsg::store_immutable_block_id_request(
+                    slot, header_id,
+                ))
+                .await
+                .map_err(|_| "Failed to send store_immutable_block_id request to storage relay")?;
+        }
```
danielSanchezQ: Create a future for each and then await them all. This sets should be small anyway if we always call them per lib change.

nomos-services/storage/src/api/backend/rocksdb/blocks.rs
```diff
@@ -34,6 +41,42 @@ impl<SerdeOp: StorageSerde + Send + Sync + 'static> StorageChainApi for RocksBac
...
+            })
+            .transpose()
+    }
+
+    async fn scan_immutable_block_ids(
+        &mut self,
+        _slot_range: RangeInclusive<Slot>,
+        _limit: NonZeroUsize,
+    ) -> Result<Vec<HeaderId>, Self::Error> {
+        todo!("implement this by updating load_prefix to accept more arguments")
```
danielSanchezQ, line 83: Lets open an issue.

nomos-services/chain-service/src/storage/adapters/storage.rs
andrussal: This could be also done as transaction in RocksDb and send all ids at once: `StorageMsg::store_immutable_block_ids_request`

andrussal (approved):

nomos-core/chain-defs/src/header/mod.rs
zeegomo: yes, either use infallible if it can't fail or some real error in this case when there's a length mismatch

nomos-services/chain-service/src/storage/adapters/storage.rs
zeegomo: It's going to be just one most of the times but I see both points, if order is important go for batching

nomos-services/storage/src/api/backend/rocksdb/blocks.rs
youngjoon-lee: Ah, this has been already implemented by the next PR: https://github.com/logos-co/nomos/pull/1427/files#diff-4d1c22bad6b511d0fd957ab6f04145724b2046616cf781c09640dff89ee419ea

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: Right. I just changed it to `BTreeMap<Slot, Id>` which is safer. [de1140a](https://github.com/logos-co/nomos/pull/1424/commits/de1140afab839304e6fb7489fc6e3fd2d165878d)

I could use `HashMap` instead, because the order is not super important for now. The caller use this `immutable_blocks` only to store `Slot->Id` key-values, and Rocksdb manages ordering internally. Nevertheless, it's more efficient from the perspective of Rocksdb if I provide an ordered list of key-values. Also, in the future, we may have another reason to have the ordered map. 

So, I chose `BTreeMap`.

nomos-core/chain-defs/src/header/mod.rs
youngjoon-lee: Right. I just defined a concrete error type: cde97941e75e2007b83e3668064cac6b323d738f

nomos-services/chain-service/src/storage/adapters/storage.rs
youngjoon-lee: I implemented batch write: [d26f538](https://github.com/logos-co/nomos/pull/1424/commits/d26f538b455a1a6d62aa709bbf5f04076f36b0aa)

danielSanchezQ (approved):

PR #1427, opened 2025-07-07
feat(sync): block provider: load historical blocks from storage

youngjoon-lee:

(This is a stacked PR based on #1424.)

(This is a collaborative work between @youngjoon-lee and @andrussal.)

Now that cryptarchia-engine prunes all immutable blocks before LIB (#1419) and those blocks are stored in storage in slot order (#1424), this PR updates the bootstrap block provider to access historical blocks from storage.

This PR:
- Updates the logic for finding the start block.
  - Instead of just calling `max_lca`, this PR implements additional ways to find the start block if `max_lca` fails because the target block or the known blocks have been pruned from the engine.
- Updates the logic for computing the path between the start block and the target block.
  - The previous logic works only when all blocks between `[start_block, target_block]` exist in the engine. This PR introduces additional logic to load immutable blocks from storage. 

Yes

@youngjoon-lee @zeegomo @andrussal 

Yes

NO

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/storage/src/api/backend/rocksdb/blocks.rs
+++ b/nomos-services/storage/src/api/backend/rocksdb/blocks.rs
@@ -106,7 +106,7 @@ mod tests {
     use crate::backends::{rocksdb::RocksBackendSettings, testing::NoStorageSerde};
 
     #[tokio::test]
-    async fn store_immutable_block_ids() {
+    async fn immutable_block_ids() {
         let temp_dir = TempDir::new().unwrap();
         let mut backend = RocksBackend::<NoStorageSerde>::new(RocksBackendSettings {
             db_path: temp_dir.path().to_path_buf(),
@@ -115,6 +115,7 @@ mod tests {
         })
         .unwrap();
 
+        // Store
         backend
             .store_immutable_block_ids(
                 [(0.into(), [0u8; 32].into()), (1.into(), [1u8; 32].into())].into(),
@@ -122,6 +123,7 @@ mod tests {
             .await
             .unwrap();
 
+        // Get
         assert_eq!(
             backend.get_immutable_block_id(0.into()).await.unwrap(),
             Some([0u8; 32].into())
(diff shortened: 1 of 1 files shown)


nomos-services/chain-service/src/sync/block_provider.rs
```diff
@@ -33,301 +37,907 @@ pub enum GetBlocksError {
...
-                async move {
-                    let (tx, rx) = oneshot::channel();
-                    if let Err((e, _)) = storage.send(StorageMsg::get_block_request(id, tx)).await {
-                        error!("Failed to send block request for {id:?}: {e}");
-                    }
+                let storage = storage.clone();
 
-                    let block = rx
+                async move {
+                    let block = Self::load_block(id, &storage)
```
zeegomo, line 149: can't we use the scan blocks method here? the path should always be in order

nomos-services/chain-service/src/sync/block_provider.rs
```diff
@@ -33,301 +37,907 @@ pub enum GetBlocksError {
...
+        cryptarchia: &cryptarchia_engine::Cryptarchia<HeaderId, State>,
+        known_blocks: &HashSet<HeaderId>,
+        target_info: &BlockInfo,
+    ) -> Result<BlockInfo, GetBlocksError> {
+        if target_info.location == BlockLocation::Engine {
+            if let Some(start_info) =
+                Self::find_optimal_start_block_from_engine(cryptarchia, known_blocks, target_info)?
+            {
+                return Ok(start_info);
+            }
```
zeegomo: should we return an error if we can't find the block here even though the location is set to engine or do we just rely on the error at the end?

zeegomo (approved):

nomos-services/chain-service/src/sync/block_provider.rs
andrussal: Currently we have stored only header ids in slot order: `slot->id`, to avoid storing full blocks twice. This mapping allows us to avoid doing walk in storage from `target_block` to a `know_block` in order to find `MAX_NUMBER_OF_BLOCKS` slice of blocks to send: `known_block...MAX_NUMBER_OF_BLOCKS...target_block`. 

We could also store full blocks: `slot`->`block` but I think we don't win very much in terms of performance and would use twice as space for blocks.

nomos-services/chain-service/src/sync/block_provider.rs
andrussal: Here location is only `target_block` specific. If target block is in engine, we first try to find start block from engine using lca. If this fails, we fall back to storage. And if storage path also fails, we return error.

nomos-services/chain-service/src/sync/block_provider.rs
zeegomo: could you remember how are full blocks stored? I guess incoming blocks are stored by id and pruned ones eliminated so that immutable blocks are the ones that remain after pruning? could it make sense to have a immutable specific store which is append only so that streaming is easier? Not to be done in this PR, just a note for the future

andrussal: > could you remember how are full blocks stored? I guess incoming blocks are stored by id and pruned ones eliminated so that immutable blocks are the ones that remain after pruning? 

Yes exactly!

> could it make sense to have a immutable specific store which is append only so that streaming is easier? Not to be done in this PR, just a note for the future

I think this is a good idea.  When we "mark" a block as immutable, we could store it by slot instead of id so we can scan it at one go, instead of getting ids first and then fetching blocks.

youngjoon-lee: The immutable-specific store sounds good. But we need to think about removing the `HeaderId -> Block` scheme. From the perspective of sync block provider (this PR), the `Slot` key scheme is enough. But we might need to access immutable blocks by `HeaderId` in the future. 
This is why the current implementation keeps the original storage (`HeaderId -> Block`) and imitates the secondary index (`Slot -> HeaderId`) separately only for the immutable blocks.
However, I see the point. If we don't have any specific use case for the `HeaderId` key scheme, we can go with only the Slot key scheme for now, and imitate the secondary index (`HeaderId -> Slot`) later when necessary. I just opened an issue for this topic: https://github.com/logos-co/nomos/issues/1445.

PR #1428, opened 2025-07-09
feat(chain): select/switch fork choice rule

youngjoon-lee:
## 1. What does this PR implement?

This implements:
- [Fork Choice Rule Selection](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81299066c768c56a06f1), which is the first step of chain bootstrapping.
  - As a part of this, [force_bootstrap](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81e38821d6911f78dba3) option is added.
- [Prolonged Bootstrap Period](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df8162be49e5aa02199378), which is a period to be waited before switching to Online after IBD is complete.

If you want to see the entire flow, please see the [flow chart](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81f6bf8cf7246f99d36e) in the spec.

To handle the two modes — Bootstrap and Online — in a clean and maintainable way, a slight refactor of the chain-service was inevitable. But I believe this refactoring is very simple and doesn't change the fundamental structures. I have more refactoring idea, but for next PRs. I don't want to make this PR large. 

At next PRs, I will implement [IBD](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81bd899ff05e97d66375) and the rest of criteria ([offline grace period](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df8168a95af51c8be83732)) for the fork choice rule selection.

## 2. Does the code have enough context to be clearly understood?

Yes

## 3. Who are the specification authors and who is accountable for this PR?

@youngjoon-lee @zeegomo 

## 4. Is the specification accurate and complete?

Yes

[...]

PR #1431, opened 2025-07-09
refactor(chain): extract `BlockProcessor` and `StorageAdapterExt` from `CryptarchiaConsensus`

youngjoon-lee:

The current `CryptarchiaConsensus` struct (aka. chain service) contains many methods. It's not bad, but I found that we need to extract some methods to separate modules, so that we can use them easily from other modules (e.g. IBD which will be implemented after #1428 is merged). 

So, this PR introduces:
- `BlockProcessor` trait and impl
  - This will be useful for implementing/testing IBD. 
- `StorageAdapterExt` trait, which provides utility methods that are not directly provided by `StorageAdapter` but useful for `BlockProcessor`.
  - We had been carrying around the blocks that failed to be deleted from storage at the service level, and it was quite a hassle. So I decided to maintain those failed operations within `StorageAdapterExt` instead. I believe this is more natural.

Yes

@zeegomo @youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/chain-service/src/lib.rs
+++ b/nomos-services/chain-service/src/lib.rs
@@ -949,23 +949,11 @@ where
         );
     }
 
-    /// Initialize cryptarchia
-    /// It initialize cryptarchia from the LIB (initially genesis) +
-    /// (optionally) known blocks which were received before the service
-    /// restarted.
+    /// Recover [`Cryptarchia`] by loading blocks from the storage.
     ///
-    /// # Arguments
-    ///
-    /// * `initial_state` - The initial state of cryptarchia.
-    /// * `lib_id` - The LIB block id.
-    /// * `lib_state` - The LIB ledger state.
-    /// * `leader` - The leader instance. It needs to be a Leader initialised to
-    ///   genesis. This function will update the leader if needed.
-    /// * `ledger_config` - The ledger configuration.
-    /// * `relays` - The relays object containing all the necessary relays for
-    ///   the consensus.
-    /// * `block_subscription_sender` - The broadcast channel to send the blocks
-    ///   to the services.
+    /// It loads blocks from the storage starting from the LIB block
+    /// up to the tip block stored in the initial service state.
+    /// Loaded blocks are applied to the [`Cryptarchia`].
     async fn recovery_cryptarchia<State>(
         &self,
         mut cryptarchia: Cryptarchia<State>,


nomos-services/chain-service/src/storage/adapters/storage.rs
```diff
@@ -106,3 +106,17 @@ where
...
+    Storage: StorageBackend + Send + Sync + 'static,
+    <Storage as StorageChainApi>::Block:
+        TryFrom<Block<Tx, BlobCertificate>> + TryInto<Block<Tx, BlobCertificate>>,
+    Tx: Clone + Eq + Serialize + DeserializeOwned + Send + Sync + 'static,
+    BlobCertificate: Clone + Eq + Serialize + DeserializeOwned + Send + Sync + 'static,
+{
+    fn parent_id(block: &Self::Block) -> HeaderId {
+        block.header().parent()
+    }
+}
```
danielSanchezQ: uhm, is this really needed? Why can't whatever not be able to access the parent?

danielSanchezQ (approved):

nomos-services/chain-service/src/storage/adapters/storage.rs
youngjoon-lee: I noticed the same thing when I was writing this method, but I didn't want to touch the existing `StorageAdapter` in this PR, since I was worried how many cascaded changes I had to make.

However, I just made a try and came up with this simple commit 6d427f40.

The previous `StorageAdapter` trait has abstracted the `Block` as an associate type. I guess the reason of this was the [`StorageChainApi::Block`](https://github.com/logos-co/nomos/blob/4110eaae15f5a38de2412d1a04bc1e262e4d3a21/nomos-services/storage/src/api/chain/mod.rs#L11), which is also an associate type. It makes sense to have an associate type in the `StorageChainApi` because it's defined in the storage service. But, from the perspective of the chain service, the Block doesn't need to be abstracted. It's clear that the chain service uses `nomos_core::block::Block`. In other words, the `StorageAdapter` should be able to deal with the `nomos_core::block::Block`.

My commit addresses this issue. I believe this commit doesn't make this PR complicated.

youngjoon-lee: Hm, I guess the reason why we introduced the associate type `Block` for `StorageAdapter` was for testing, since it's hard to build a `Block` which contains a proof. Though, we don't have any test yet.

youngjoon-lee: I just reverted my last commit https://github.com/logos-co/nomos/commit/6d427f40d41c3e663114a6d79d97b3ee8108560f because of the reason I mentioned [above](https://github.com/logos-co/nomos/pull/1431#discussion_r2197394123). We need to write tests using this abstraction soon.

nomos-services/chain-service/src/lib.rs
```diff
@@ -1165,141 +952,36 @@ where
...
-        >,
-        ledger_config: nomos_ledger::Config,
-        leader_config: LeaderConfig,
-        relays: &CryptarchiaConsensusRelays<
-            BlendAdapter,
-            BS,
+    async fn recovery_cryptarchia<State>(
+        &self,
+        mut cryptarchia: Cryptarchia<State>,
+        block_processor: &mut NomosBlockProcessor<
```
zeegomo, line 960: the comment is now outdated

nomos-services/chain-service/src/processor.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// The updated [`Cryptarchia`] is returned if the block is valid and added
+    /// successfully. If not, [`Error`] is returned along with the unchanged
+    /// [`Cryptarchia`].
+    #[instrument(level = "debug", skip(self, cryptarchia, storage))]
+    async fn process_block<CryptarchiaState, Storage>(
+        &mut self,
+        cryptarchia: Cryptarchia<CryptarchiaState>,
+        block: Self::Block,
+        storage: &mut Storage,
+    ) -> Result<Cryptarchia<CryptarchiaState>, (Error, Cryptarchia<CryptarchiaState>)>
```
zeegomo: `Cryptarchia` should be clonable, we don't need to return it in the event of an error

nomos-services/chain-service/src/storage/adapters/storage.rs
```diff
@@ -118,3 +123,55 @@ where
...
+                    tracing::error!(target: LOG_TARGET, "Failed to remove block: {e}");
+                    Some(*id)
+                }
+            })
+            .collect();
+    }
+
+    fn parent_id(block: &Self::Block) -> HeaderId {
+        block.header().parent()
+    }
```
zeegomo, line 154: why is this a method on `StorageAdapterExt`?

nomos-services/chain-service/src/storage/mod.rs
```diff
@@ -67,3 +69,48 @@ pub trait StorageAdapter<RuntimeServiceId> {
         blocks: BTreeMap<Slot, HeaderId>,
     ) -> Result<(), overwatch::DynError>;
 }
+
+const LOG_TARGET: &str = "cryptarchia::service::storage";
+
+/// Extension trait for [`StorageAdapter`] that provides additional
+/// functionalities.
+#[async_trait::async_trait]
+pub trait StorageAdapterExt<RuntimeServiceId>: StorageAdapter<RuntimeServiceId> {
```
zeegomo: what's the reason for an extension trait instead of adding these to the `StorageAdapter` trait?

nomos-services/chain-service/src/processor.rs
```diff
@@ -0,0 +1,267 @@
...
+    where
+        CryptarchiaState: cryptarchia_engine::CryptarchiaState + Send,
+        Storage: StorageAdapter<RuntimeServiceId, Block = Self::Block>
+            + StorageAdapterExt<RuntimeServiceId>
+            + Send
+            + Sync
+            + 'static;
+}
```
zeegomo: do we really need a trait here? we don't use it in the parent module

zeegomo (changes requested):

nomos-services/chain-service/src/processor.rs
youngjoon-lee: We will use this for testing IBD in #1442 because I want to test IBD without creating real `Blocks` and spawning dependent services such as mempool and storage.
I will mention this clearly in the PR description (as later PRs)

nomos-services/chain-service/src/storage/adapters/storage.rs
youngjoon-lee: This is necessary for the `StorageAdapterExt::get_blocks_in_range` method. This will be anyway replace with the Block trait that I added in #1442.

nomos-services/chain-service/src/storage/mod.rs
youngjoon-lee: I will move these method to the `StorageAdapter` trait. At the beginning, I thought these methods may not be needed to be implemented for all `StorageAdapter` impls, but I was wrong. These are necessary for the chain service and all impls should implements these.

youngjoon-lee: Done 962ff5ef

nomos-services/chain-service/src/processor.rs
youngjoon-lee: Yes. Done 4a3b1c314d3bd503c7bafdf2ac0a22ad6281a48f

nomos-services/chain-service/src/lib.rs
youngjoon-lee: Thanks. Updated d13b85b7

PR #1435, opened 2025-07-10
feat(da): move ApiAdapter  from da-sampling to da-network

pradovic:

The `da-sampling` was still using different `MembershipHandler` from config, rather then one from the `da-network`, to get peer for commitments call. Also, get commitments call should be moved to `da-network` and moved from http to quic, in the future. This is the first step towards this.

This PR is moving http adapter from `da-sampling` to `da-network`. The next PR will introduce `GetCommitments` message to `da-network`, and `da-sampling` will use this to fetch commitments.  `da-network` will still use http to serve this call.

Yes

@pradovic @bacv 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.


nomos-services/data-availability/sampling/src/lib.rs
```diff
@@ -291,10 +271,10 @@ where
         }
 
         // Fall back to API request
-        let (reply_sender, reply_channel) = oneshot::channel();
-        if let Err(e) = api_request.request_commitments(blob_id, reply_sender).await {
-            error!("Error sending request to API backend: {e}");
-        }
+        let (_reply_sender, reply_channel) = oneshot::channel();
+        // if let Err(e) = api_request.request_commitments(blob_id, reply_sender).await
```
bacv: The new service message wasn't introduced in this PR because we'd need to propagate the Commitment type?

pradovic: No, just for the reason of making small PRs. If you think it would still be ok to review, I don't mind updating this PR with that.
We would need to add the message to `da-network` and to update the network adapter with the call to get commitments. I thought it might be more clean to review separately. 
I can also open a PR branching from this one with this functionality

bacv (approved):

danielSanchezQ (approved):
This is a temporary step, but so stuff do not break. Thanks

PR #1436, opened 2025-07-10
feat(ledger): Use efficient UtxoTree in ledger

zeegomo:
This PR integrates the newly introduced UtxoTree in the ledger, using blake2 as the hashing alg.
In addition it:
* adds a compact serialization format for the UtxoTree that only store the utxos and their position (this also remove the `item_to_key` field in favor of requesting directly a key like in map structures)
* fixes an issue in the holes where we were using a FIFO queue for holes, not inserting in the smallest one (and adds a regression test)

Yes

@zeegomo 

N/A

We're still using blake2 here for convenience since we still have to select a suitable library for poseidon2. I don't anticipate an issue in changing this later as the code accepts a generic type for the hash.

## Checklist

> [!WARNING]  

* [x] 2. Description added.
* [ ] 5. Implementation and Specification are 100% in sync including changes. This is critical. It's not, for the reasons explained above


danielSanchezQ (approved):

utxotree/src/lib.rs
```diff
@@ -1,34 +1,55 @@
...
     // key -> (item, position in merkle tree)
-    items: HashTrieMap<Key, (Item, usize)>,
-    item_to_key: F,
+    items: HashTrieMapSync<Key, (Item, usize)>,
+}
+
+impl<Key, Item, Hash> Default for UtxoTree<Key, Item, Hash>
+where
+    Key: AsRef<[u8]> + Clone + std::hash::Hash + Eq,
+    Hash: Digest<OutputSize = digest::typenum::U32>,
```
thomaslavaur, line 32: Isn't this typenum::U32 specific to Blake? or binary hashes?

zeegomo: it's the number of bytes produced by the output of the hash indeed, I guess we could make it even more flexible when we will add poseidon or just cram it by serializing

PR #1444, opened 2025-07-13
feat(da): split addressbook and membership

pradovic:
This PR is splitting addressbook and membership. The reason for this is that membership saves and reads snapshots from storage, while the addressbook has a different usage pattern. 

The idea is to keep cumulative addressbook and support `get_address` call using storage. 
The mock implementation in this PR is not using storage but a simple hash map. The real implementation can be implemented very soon in some of the next PRs, if we agree on this design.

Yes.

@pradovic @bacv 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/data-availability/network/src/lib.rs
+++ b/nomos-services/data-availability/network/src/lib.rs
@@ -1,3 +1,4 @@
+pub mod addressbook;
 pub mod api;
 pub mod backends;
 pub mod membership;
@@ -16,7 +17,7 @@ use futures::Stream;
 use kzgrs_backend::common::share::{DaShare, DaSharesCommitments};
 use libp2p::{Multiaddr, PeerId};
 use nomos_core::{block::BlockNumber, da::BlobId};
-use nomos_da_network_core::addressbook::{mock::MockAddressBook, AddressBookMut as _};
+use nomos_da_network_core::addressbook::AddressBookMut as _;
 use overwatch::{
     services::{
         state::{NoOperator, ServiceState},
@@ -31,6 +32,7 @@ use tokio::sync::oneshot;
 use tokio_stream::StreamExt as _;
 
 use crate::{
+    addressbook::mock::MockAddressBook,
     api::ApiAdapter as ApiAdapterTrait,
     membership::{handler::DaMembershipHandler, MembershipAdapter},
 };
--- a/nomos-da/network/core/src/addressbook.rs
+++ b/nomos-da/network/core/src/addressbook.rs
@@ -1,5 +1,3 @@
-pub mod mock;
-
 use std::{collections::HashMap, fmt::Debug};
 
 use libp2p::Multiaddr;
--- a/nomos-services/data-availability/network/src/addressbook/mock.rs
+++ b/nomos-services/data-availability/network/src/addressbook/mock.rs
@@ -4,8 +4,7 @@ use std::{
 };
 
 use libp2p::{Multiaddr, PeerId};
-
-use crate::addressbook::{AddressBookHandler, AddressBookMut};
+use nomos_da_network_core::addressbook::{AddressBookHandler, AddressBookMut};
 
 #[derive(Default, Debug, Clone)]
 pub struct MockAddressBook {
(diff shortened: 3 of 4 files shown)


nodes/nomos-node/node/src/lib.rs
```diff
@@ -82,7 +82,7 @@ impl StorageSerde for Wire {
 /// Membership used by the DA Network service.
 pub type NomosDaMembership = FillFromNodeList;
 pub type DaMembershipStorage = MockStorage;
-pub type DaNetworkApiAdapter = HttApiAdapter<DaMembershipHandler<NomosDaMembership>>;
+pub type DaNetworkApiAdapter = HttApiAdapter<DaMembershipHandler<NomosDaMembership>, DaAddressbook>;
```
pradovic, line 85: I did not create `da-network` generic over Addresbook, but have exported public type `DaAddressbook` which is set to the mock implementation. 
Underlying behaviors are generic over addressbook.
The reason was to avoid the type propagation on top of network service. Let me know what you think @danielSanchezQ

nomos-services/data-availability/network/src/lib.rs
```diff
@@ -417,11 +432,16 @@ where
...
+        update: HashMap<Membership::Id, Multiaddr>,
         storage: &MembershipStorage<StorageAdapter, Membership>,
+        addressbook: &DaAddressbook,
     ) {
-        storage.update(block_numnber, update);
+        storage.update(block_number, update.clone());
+        // since addressbook access is different then membership (get address vs
+        // snapshots) addressbook real implementation would have storage inside
+        // for update and get address
+        addressbook.update(update);
```
pradovic, line 446: My idea here is to access and update addressbook in a different them membership. 
The reason is that membership is working on snapshots and the whole snapshots are propagated.

The addressbook update would just need to bulk insert new ids in the underlying storage (rewriting the possible old ones), while `self.addressbook.get_address` would use the underlying storage to get the address.
This is not implemented in the mock implementation, but the real implementation would use a storage to get the address. 
@danielSanchezQ , @bacv Please, let me know what you think.

nomos-da/network/core/src/addressbook/mock.rs
```diff
@@ -0,0 +1,29 @@
...
+use libp2p::{Multiaddr, PeerId};
+
+use crate::addressbook::{AddressBookHandler, AddressBookMut};
+
+#[derive(Default, Debug, Clone)]
+pub struct MockAddressBook {
+    peers: Arc<Mutex<HashMap<PeerId, Multiaddr>>>,
+}
+
+impl AddressBookHandler for MockAddressBook {
```
bacv: `nomos-da-network-core` is proper place for the Trait definition, maybe it could be moved just to `addressbook.rs` or left as is, but Mock and other implementations shouldn't be here in my opinion. 

Just like MembershipHandler, we should have the implementation in DaNetwork service, real implementation will have to communicate with the storage.
For tests in nomos-da-network-core we could just redefine it in test_utils

nomos-services/data-availability/network/src/lib.rs
bacv: This place for update seems good.

bacv (approved):
With seeing the code it's now evident that the split was a good idea, just one comment about implementation moving out from core to service.

nomos-da/network/core/src/addressbook/mock.rs
pradovic: That sounds better to me as well, I will move it as suggested. Thanks!

pradovic: If I understood correctly this should be it: https://github.com/logos-co/nomos/pull/1444/commits/9073b63ebf4a55ae596e8ef7c2112d07e9a5a992.

bacv: Yes!

nodes/nomos-node/node/src/lib.rs
danielSanchezQ: Lets go with this and see nuances later.

PR #1448, opened 2025-07-16
feat(bootstrap): IBD integration testing

andrussal:

This PR adds integration test for IBD(Initial Block Download). The test starts 2 nodes and lets them to create blocks. Later 2 other nodes try to join. First one fails to download blocks because initial nodes are still in Bootstrapping mode. After nodes are switched to Online mode, another node tries to download blocks, and succeeds.

- This PR also adds `mode` to `CryptarchiaInfo` returned by http consensus info endpoint for easy mode tracking

I also added few fixes that seemed necessary. I didn't want to put these to another PRs (where they could fit logically better) because there they probably would be not noticed.

--- 
Note about consensus - when configuring consensus for 3 nodes but starting only 2 initially, they seem to build often separate chains, including 2 initial nodes. Many missing parent entries in log. 

Although it's not IBD related, perhaps I should find root cause of this in this PR.

Yes. 

@andrussal 

N/A

N/A

## Checklist

* [x] 1. Description added.


nomos-services/chain-service/src/lib.rs
```diff
@@ -952,15 +962,27 @@ where
                     ).await;
                 }
 
+                Some(event) = chainsync_events.next() => {
+                    Self::reject_chainsync_event(event).await;
```
youngjoon-lee: Please link https://github.com/logos-co/nomos/issues/1451 to this PR if you think this PR addresses the issue completely.

nomos-services/chain-service/src/lib.rs
```diff
@@ -1547,6 +1571,24 @@ where
...
+                    Err(DynError::from(
+                        "Node is bootstrapping, cannot provide blocks",
+                    ))
+                });
+                if let Err(e) = reply_sender.send(Box::pin(error_stream)).await {
+                    tracing::warn!("Failed to send error response for ProvideBlocksRequest: {e}");
+                }
+            }
+            ChainSyncEvent::ProvideTipRequest { .. } => {
+                tracing::warn!("Node is bootstrapping, cannot provide tip");
```
youngjoon-lee: (Though this is not a key change of this PR.)
In this case, doesn't we need to return an error (like `ProvideBlocksRequest`)? This is the point I thought we may need refactoring for implementing this rejecting, as I mentioned https://github.com/logos-co/nomos/issues/1451. But, you may know better how the requester works if the responder silently. doesn't respond like this.

nomos-services/chain-service/src/processor.rs
```diff
@@ -298,7 +298,16 @@ where
...
+                if let Err(e) = self
+                    .storage
+                    .store_immutable_block_ids(pruned_blocks.immutable_blocks().copied())
+                    .await
+                {
+                    error!("Could not store immutable blocks as immutable: {e}");
+                }
+
+                self.storage_blocks_to_remove
+                    .extend(pruned_blocks.stale_blocks());
```
youngjoon-lee: Could you please reorganize the base branch ([ibd_and_pruning_merge](https://github.com/logos-co/nomos/tree/ibd_and_pruning_merge)) based on the new `master` and other base PRs? I know it's hard to maintain this temporary branch, but at least, I think this change shouldn't be displayed in this PR.

(I’ll need to refactor some of the dependent PRs soon (as discussed in the comments Giacomo left). It will take some time. I’ll let you know when another reorganization is needed. In the meantime, this base branch can already be cleaned up now.)

nomos-services/chain-service/src/sync/block_provider.rs
```diff
@@ -145,6 +145,9 @@ where
     ) -> BoxStream<'static, Result<Bytes, DynError>> {
         let storage = self.storage_relay.clone();
 
+        // Skip known block
+        let path = path.into_iter().skip(1);
```
youngjoon-lee: Wdyt about updating this comment?
```
Skip a block already known to the requester.
```

nomos-services/chain-service/src/sync/block_provider.rs
```diff
@@ -290,7 +293,33 @@ where
...
+    }
+
+    async fn find_immutable_genesis_block(
+        &self,
+        known_blocks: &HashSet<HeaderId>,
+    ) -> Result<Option<BlockInfo>, GetBlocksError> {
+        let genesis_block_id = self
+            .scan_immutable_block_ids(
+                Slot::genesis()..=Slot::genesis(),
+                NonZeroUsize::new(1).expect("MAX_NUMBER_OF_BLOCKS should be > 0"),
```
youngjoon-lee: I think the `expect` message is not proper. I think we don't need to mention `MAX_NUMBER_OF_BLOCKS` in this function because we're not using it. We're just creating a `NonZeroUsize` from `usize` that should be always successful.

nomos-services/chain-service/src/sync/block_provider.rs
```diff
@@ -290,7 +293,33 @@ where
...
+            )
+            .await?
+            .into_iter()
+            .next()
+            .ok_or(GetBlocksError::StartBlockNotFound)?;
+
+        if !known_blocks.contains(&genesis_block_id) {
+            return Ok(None);
+        }
```
youngjoon-lee: I don't understand the purpose of this `if` statement. Can we just remove this `if`? I think the purpose of this function is just check whether the genesis block has been stored as immutable. 

Also, FYI, the spec recommends nodes to not include their genesis block in the `known_blocks` (but yeah, they can include it if they want). So, this `if` will be `true` in most cases (due to `!`).

nomos-services/chain-service/src/sync/block_provider.rs
```diff
@@ -659,8 +687,8 @@ mod tests {
 
         assert_eq!(
             retrieved_blocks.len(),
-            MAX_NUMBER_OF_BLOCKS,
-            "Retrieved blocks should not exceed MAX_NUMBER_OF_BLOCKS"
+            MAX_NUMBER_OF_BLOCKS - 1, // Subtract 1 because a known block is skipped
+            "Retrieved blocks should not exceed MAX_NUMBER_OF_BLOCKS minus the skipped known block"
         );
```
youngjoon-lee: Do we need this change because of the setup of this test? Or, has the actual limit changed to `MAX_NUMBER_OF_BLOCKS - 1` (because of the change of "skipping the first block")? 
I think the limit should remain at `MAX_NUMBER_OF_BLOCKS`. The limit doesn't need to be changed according to the internal logic of the provider.
Maybe, I'm missing something in this test. Could you confirm?

nomos-services/chain-service/src/lib.rs
```diff
@@ -1199,10 +1221,11 @@ where
         cryptarchia
     }
 
-    fn process_message<State: CryptarchiaState>(
+    fn process_message<State: CryptarchiaState + 'static>(
         cryptarchia: &Cryptarchia<State>,
         block_channel: &broadcast::Sender<Block<ClPool::Item, DaPool::Item>>,
         msg: ConsensusMsg<Block<ClPool::Item, DaPool::Item>>,
+        mode: ConsensusMode,
     ) {
```
youngjoon-lee: I understand why you introduced `ConsensusMode` even though we already have `CryptarchiaState` (because Rust doesn't provide runtime type information for generic types). 
I think we can at least rename `ConsensusMode` to `ConsensusModeInfo` to imply that this enum is only for building the REST API response (`CryptarchiaInfo`), not for any core logic.

tests/src/nodes/validator.rs
```diff
@@ -91,50 +93,48 @@ impl Drop for Validator {
...
-        let node = Self {
-            addr: config.http.backend_settings.address,
-            testing_http_addr: config.testing_http.backend_settings.address,
-            child,
-            tempdir: dir,
-            config,
-        };
-        tokio::time::timeout(adjust_timeout(Duration::from_secs(10)), async {
+        tokio::time::timeout(adjust_timeout(Duration::from_secs(10)), async move {
+            let mut file = NamedTempFile::new().unwrap();
```
youngjoon-lee: Could you please give me a note why we need this change? (moving everything into this code block)

tests/src/tests/cryptarchia/bootstrap.rs
```diff
@@ -0,0 +1,123 @@
...
+    let target_blocks = security_param * CHAIN_LENGTH_MULTIPLIER;
+
+    let min_blocks = 5;
+
+    println!(
+        "Waiting for initial validators to reach at least min_blocks ({min_blocks}) but still be in bootstrapping mode...",
+    );
+
+    wait_for_validators_mode_and_height(&validators, min_blocks, ConsensusMode::Bootstrapping)
+        .await;
```
youngjoon-lee: Could you leave a comment to explain what the `target_blocks` is for? When reading this test from top to bottom, it's not easy to understand what it is for (because the meaning of `CHAIN_LENGTH_MULTIPLIER` is also not clear). 

I think you can move the declaration of `target_block` to the line 60 where it is actually used. Also, I think you can rename this to `target_height` (which is clearer I think) because it is used for the `wait_for_validators_mode_and_height` function. 

With the same reason, I think we can also rename `min_blocks` to `min_height`. 

Plus, we may need to ensure that `target_blocks` is higher than `min_blocks`.

youngjoon-lee: Also, I'd like to understand why we need to wait until `target_blocks` has been reached? Isn't it okay for us to just wait until the node switches to the Online mode?

tests/src/tests/cryptarchia/bootstrap.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    let initial_node_min_height = initial_heights.iter().min().unwrap();
+
+    // 1 second is enough to download the initial blocks and catch up
+    tokio::time::sleep(adjust_timeout(Duration::from_secs(1))).await;
+
+    let behind_node_info = behind_node.consensus_info().await;
+    println!("behind node info: {behind_node_info:?}");
+
+    assert!(behind_node_info.height >= *initial_node_min_height);
```
youngjoon-lee: In addition to this check, we may need to clearly check whether the behind node has caught up with the **current** tip of the initial nodes, because the chains of the initial nodes grow while the IBD is running (though this test is super short).
For example,
```rust
const MARGIN = 1;
let current_min_height =  initial_heights.iter().min().unwrap();
assert!(behind_node_info.height >= current_min_height - MARGIN);
```

nomos-services/chain-service/src/lib.rs
andrussal: I will open another PR to solve rejecting more cleanly that will close #1451

nomos-services/chain-service/src/sync/block_provider.rs
andrussal: Make sense, I will remove it from here. It should be checked where this function is called. 

> Also, FYI, the spec recommends nodes to not include their genesis block in the known_blocks (but yeah, they can include it if they want). So, this if will be true in most cases (due to !).

It will be necessary to include it when lib is genesis, right? Maybe I am missing something though.

nomos-services/chain-service/src/sync/block_provider.rs
andrussal: Good point, it should still be the `MAX_NUMBER_OF_BLOCKS` even if we remove known block. I will update.

tests/src/nodes/validator.rs
andrussal: Previously `Validator::spawn` just returned Validator or panicked with timeout. I changed it to return Timeout. This made it easier to test the case when Validator failed to start. Now I think that returning Result would be better though.

tests/src/tests/cryptarchia/bootstrap.rs
andrussal: Good idea, I will add this check too.

nomos-services/chain-service/src/processor.rs
andrussal: I updated the base branch to exclude this change from here.

tests/src/tests/cryptarchia/bootstrap.rs
andrussal: I think we can safely remove `target_block` as we only need to know if node is Online. I used the same approach as in two_nodes_happy test but here we can just use mode.

nomos-services/chain-service/src/lib.rs
andrussal: I renamed it to `ConsensusModeInfo`. And also in function signature replaced `ConsensusMode` with a generic, perhaps this way is a bit cleaner.

nomos-services/chain-service/src/sync/block_provider.rs
youngjoon-lee: > It will be necessary to include it when lib is genesis, right? Maybe I am missing something though. 

Right.

PR #1450, opened 2025-07-16
feat(da): Commitments request and response in sampling protocol

bacv:

Sampling protocol specification allows peers to query for shared commitments. Sampling behaviour was expanded to handle such request and response.

Sampling behaviour was also split into several modules for easier navigation.

[Spec](https://www.notion.so/nomos-tech/NomosDA-Sampling-1fd261aa09df81dcb3c6ebc99206a0b1?source=copy_link#22d261aa09df8012b81de31d3d1db354)

@danielSanchezQ @bacv 

Yes

No

## Checklist

* [ ] 2. Description added.

### DIFF
--- a/nomos-da/network/core/src/protocols/sampling/behaviour.rs
+++ b/nomos-da/network/core/src/protocols/sampling/behaviour.rs
@@ -8,319 +8,38 @@ use either::Either;
 use futures::{
     channel::{
         oneshot,
-        oneshot::{Canceled, Receiver, Sender},
+        oneshot::{Receiver, Sender},
     },
-    future::BoxFuture,
     stream::{BoxStream, FuturesUnordered},
     AsyncWriteExt as _, FutureExt as _, StreamExt as _,
 };
-use kzgrs_backend::common::{share::DaLightShare, ShareIndex};
 use libp2p::{
     core::{transport::PortUse, Endpoint},
     swarm::{
         dial_opts::DialOpts, ConnectionDenied, ConnectionId, DialFailure, FromSwarm,
         NetworkBehaviour, THandler, THandlerInEvent, THandlerOutEvent, ToSwarm,
     },
-    Multiaddr, PeerId, Stream,
+    Multiaddr, PeerId,
 };
-use libp2p_stream::{Control, IncomingStreams, OpenStreamError};
-use nomos_core::{da::BlobId, wire};
-use nomos_da_messages::{
-    common,
-    packing::{pack_to_writer, unpack_from_reader},
-    sampling,
-};
+use libp2p_stream::{Control, IncomingStreams};
+use nomos_core::da::BlobId;
+use nomos_da_messages::sampling;
 use rand::{rngs::ThreadRng, seq::IteratorRandom as _};
-use serde::{Deserialize, Serialize};
 use subnetworks_assignations::MembershipHandler;
-use thiserror::Error;
 use tokio::sync::mpsc::{self, UnboundedSender};
 use tokio_stream::wrappers::UnboundedReceiverStream;
-use tracing::{error, warn};
+use tracing::warn;
 
-use super::connections::Connections;
+use super::{
+    connections::Connections,
+    errors::SamplingError,
+    streams::{self, SampleStream},
+    BehaviourSampleReq, BehaviourSampleRes, ResponseChannel, SampleStreamResponse, SamplingEvent,
+    SamplingStreamFuture, SubnetsConfig,
+};
 use crate::{addressbook::AddressBookHandler, protocol::SAMPLING_PROTOCOL, SubnetworkId};
 
 type AttemptNumber = usize;
-
-#[derive(Debug, Error)]
-pub enum SamplingError {
-    #[error("Stream disconnected: {error}")]
-    Io {
-        peer_id: PeerId,
-        error: std::io::Error,
-        message: Option<sampling::SampleRequest>,
-    },
-    #[error("Dispersal response error: {error:?}")]
-    Protocol {
-        subnetwork_id: SubnetworkId,
-        peer_id: PeerId,
-        error: sampling::SampleError,
-    },
-    #[error("Error opening stream [{peer_id}]: {error}")]
-    OpenStream {
-        peer_id: PeerId,
-        error: OpenStreamError,
-    },
-    #[error("Unable to deserialize blob response: {error}")]
-    Deserialize {
-        blob_id: BlobId,
-        subnetwork_id: SubnetworkId,
-        peer_id: PeerId,
-        error: wire::Error,
-    },
-    #[error("Error sending request: {request:?}")]
-    RequestChannel {
-        request: BehaviourSampleReq,
-        peer_id: PeerId,
-    },
-    #[error("Malformed blob id: {blob_id:?}")]
-    InvalidBlobId { peer_id: PeerId, blob_id: Vec<u8> },
-    #[error("Blob not found: {blob_id:?}")]
-    BlobNotFound {
-        peer_id: PeerId,
-        blob_id: Vec<u8>,
-        subnetwork_id: SubnetworkId,
-    },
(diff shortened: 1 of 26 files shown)


nomos-da/network/core/src/protocols/sampling/behaviour.rs
```diff
@@ -418,129 +155,14 @@ where
         Ok(SampleStream { stream, peer_id })
     }
 
-    /// Get a hook to the sender channel of the sample events
+    /// Get a hook to the sender channel of the share events
     pub fn sample_request_channel(&self) -> UnboundedSender<BlobId> {
```
pradovic: Nitpick: should we rename it to `shares_request_channel`?

nomos-da/network/core/src/protocols/sampling/behaviour.rs
```diff
@@ -600,37 +222,56 @@ impl<Membership: MembershipHandler<Id = PeerId, NetworkId = SubnetworkId> + 'sta
         }
     }
 
-    fn try_peer_sample(&mut self, peer_id: PeerId) {
-        if let Some((subnetwork_id, blob_id)) = self
+    fn sample_commitments(&mut self, blob_id: BlobId) {
+        if let Some((_, peer_id)) = &self.sampling_peers.iter().choose(&mut rand::thread_rng()) {
```
pradovic:
```suggestion
        if let Some((_, &peer_id)) = &self.sampling_peers.iter().choose(&mut rand::thread_rng()) {
```

Nitpick: This might eliminate: `let peer_id = **peer_id`

nomos-services/data-availability/network/src/backends/libp2p/common.rs
```diff
@@ -147,6 +125,186 @@ pub(crate) async fn handle_validator_events_stream(
...
+        sampling::SamplingEvent::SamplingSuccess {
+            blob_id,
+            light_share,
+            ..
+        } => {
+            if let Err(e) = sampling_broadcast_sender.send(SamplingEvent::SamplingSuccess {
+                blob_id,
+                light_share,
+            }) {
+                error!("Error in internal broadcast of sampling success: {e:?}");
```
pradovic, line 143: Just an idea: We could maybe use `.inspect_err(|e| error!("Error in internal broadcast of sampling success: {e:?}");`, for these kind of stuff. I also did not use it, but it just crossed my mind now reading the code as it has lot of if err patterns. Feel free to ignore if you like `if let Err(e)` better.

pradovic (approved):
LGTM! I left a few nitpick ideas.

nomos-da/network/messages/src/common.rs
```diff
@@ -30,6 +30,20 @@ impl LightShare {
     }
 }
 
+#[repr(C)]
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct Commitments {
+    pub blob_id: BlobId,
```
danielSanchezQ: This is not needed. The blob id is the hash of the inner commitments. So this wrapper need to be removed. We can add a convenient method to the `DaSharesCommitments` that returns the blob id instead.

nomos-da/network/messages/src/sampling.rs
```diff
@@ -12,46 +12,92 @@ pub enum SampleErrorType {
...
     pub error_description: String,
 }
 
+#[repr(C)]
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct CommitmentsError {
+    pub blob_id: BlobId,
+    pub error_type: SampleErrorType,
+    pub error_description: String,
+}
```
danielSanchezQ, line 28: Do we reply with this through the wire? 🤔

testnet/cfgsync/src/config.rs
```diff
@@ -358,7 +358,8 @@ mod cfgsync_tests {
                     seen_message_ttl: Duration::ZERO,
                 },
                 subnets_refresh_interval: Duration::from_secs(1),
-                retry_subnets_limit: 1,
+                retry_shares_limit: 1,
+                retry_commitments_limit: 1,
```
danielSanchezQ, line 337: I think the correct configuration is 5.

testnet/cfgsync/src/server.rs
```diff
@@ -98,7 +98,8 @@ impl CfgSyncConfig {
             redial_cooldown: Duration::ZERO,
             replication_settings: self.replication_settings,
             subnets_refresh_interval: Duration::from_secs(30),
-            retry_subnets_limit: 1,
+            retry_shares_limit: 1,
+            retry_commitments_limit: 1,
```
danielSanchezQ: ditto

nomos-services/data-availability/network/src/backends/libp2p/common.rs
danielSanchezQ: I didn't know about `inspect_err` is it from std?

pradovic: Yes, from result: https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect_err. I think it was introduced in 1.66.  I have seen it by accident yesterday in some code.

nomos-da/network/messages/src/sampling.rs
bacv: Everything that is in `nomos-da/messages` goes through the wire, in this case the `SampleError` became `enum` with two variants: ShareError and CommitmentsError

testnet/cfgsync/src/config.rs
bacv: These are in tests for cfgsync config generation, but this is a good catch, I need to update the `yaml` whitch is missing.

nomos-services/data-availability/network/src/backends/libp2p/common.rs
danielSanchezQ: Ok I see, `inspect_error` does not consume the result. I think it is mostly used for when you want to handle the error in different places, bubbling up the same type/value to other layers. In this case we just consume everything here as it is the only layer it deals with it. I think it is more explicit with `let Err`.

pradovic: Yes, it would still need `let _ =` to explicitly consume the result.

danielSanchezQ: But we want the `e` (?)

pradovic: It would look like this, if I am not mistaken:
```
 let _ = sampling_broadcast_sender
                .send(SamplingEvent::SamplingSuccess {
                    blob_id,
                    light_share,
                })
                .inspect_err(|e| error!("Error in internal broadcast of sampling success: {e:?}"));
```
            
But yes, I think it comes down to personal preference, they both look good to me. I just like to avoid parenthesis where possible. It was really a nitpick idea

danielSanchezQ: Ah, I get it now. Uhm, I would personally keep the `let Err` still. But if we have consensus with the team I'm down to change it. Now keep in mind you (caller, keeper 😉 ) will be doing a whole mother refactor on the whole codebase.

bacv: Tried it, will update these code changes, saves a little bit of boilerplate.

pradovic: I don't have a too strong opinion, but if we end up with consensus  I don't mind a refactor 😅

bacv: > Yes, it would still need let _ = to explicitly consume the result.

Ah, this is then the same, will keep `if let`, this is the only place where we use err. But nice discovery, not sure where we currently could use it, because we usually propagate errors and then consume it at one place.

PR #1457, opened 2025-07-16
feat(da): Implement subnetworks assignations algorithm

danielSanchezQ:

This PR introduce the implementation of the [DA subnetworks assignations algorithm](https://www.notion.so/nomos-tech/DA-subnetworks-assignations-217261aa09df80fc8bb9cf46092741ce). It also adds same testing as per spec and a set of benches.

It is a much more complicated than older versions. But it should be easily redable and it matches pretty much with the [running python spec](https://github.com/logos-co/nomos-specs/pull/130).

@danielSanchezQ @madxor 

Yes

No introduction was changed. It deviates a bit from the python implementation to reduce allocations as much as possible. Otherwise its pretty much relatable.

## Checklist

* [x] 2. Description added.


nomos-utils/src/fisheryates.rs
```diff
@@ -4,12 +4,24 @@ use rand::{Rng as _, RngCore};
...
         }
     }
+
+    pub fn sample<R: RngCore, T: Ord>(
+        elements: impl IntoIterator<Item = T>,
+        sample_size: usize,
+        rng: &mut R,
+    ) -> impl Iterator<Item = T> {
+        let mut elements: Vec<T> = elements.into_iter().collect();
+        elements.sort();
```
pradovic, line 22: Is this for determinism?  Why do we need to sort before shuffle?

danielSanchezQ: ~~yes it was the idea, although now looking at it, it may not be really necessary :thinking:~~ It is, see my comment below.

pradovic: I am not 100% sure but if shuffle is rng then we probably dont need it. Seems like fisher yates should output random shuffle

bacv: Are we planning to use seedable rng? In that case sort would make sense, @pradovic lets both take a look at this [spec](https://www.notion.so/nomos-tech/DA-subnetworks-assignations-217261aa09df80fc8bb9cf46092741ce)

danielSanchezQ: I defined it like that in the [python spec](https://github.com/logos-co/nomos-specs/blob/c763d473399204b4672deecaf97d81ad533002f8/da/assignations/refill.py#L58) as well. I think is better to sort so its independent of the input order what to chose. This is pseudo-randomness here. But most important is reproducibility. It could be tricky if you pass different collections here, with the same rng and they will return different results. We do not want this (it could also happen depending implementation). So we chose to be safe and sort.

danielSanchezQ: > Are we planning to use seedable rng? In that case sort would make sense, @pradovic lets both take a look at this [spec](https://www.notion.so/nomos-tech/DA-subnetworks-assignations-217261aa09df80fc8bb9cf46092741ce)

Yes, we will use the `BlakeRng` custom rng with a different seed each session.

pradovic: Nice, thanks for the explanation. The requirement is clearly defined in the spec.

bacv (approved):

pradovic (approved):
LGTM!

PR #1470, opened 2025-07-17
feat(ledger): Process LedgerTx

zeegomo:

LedgerTxs are now applied to the state of the ledger as specified in the mantle spec. This is only one step in the full implementation of the mantle and as such:
* all fees are currently burned
* no support for locked notes

https://www.notion.so/nomos-tech/Mantle-Specification-21c261aa09df810c8820fab1d78b53d9#21c261aa09df81e09db4ec4cf0f4e27b

@davidrusu @thomaslavaur , @zeegomo 

Yes

No, but it's only a partial implementation

## Checklist

> [!WARNING]  

* [x] 2. Description added.


ledger/nomos-ledger/src/cryptarchia/mod.rs
```diff
@@ -188,6 +188,60 @@ impl LedgerState {
...
+                .ok_or(LedgerError::Overflow)?;
+            pks.push(note.pk.into());
+        }
+
+        if !tx.ledger_tx_proof().verify(&ZkSignaturePublic {
+            pks,
+            tx_hash: tx.hash().into(),
+        }) {
+            return Err(LedgerError::InvalidProof);
+        }
```
danielSanchezQ: Shouldn't we verify this first? or is more costly than computing the things above?

zeegomo: it's more costly, we could even move to the end as the last step, logically it does not make a difference

ledger/nomos-ledger/src/cryptarchia/mod.rs
```diff
@@ -188,6 +188,60 @@ impl LedgerState {
...
+
+        for (i, output) in ledger_tx.outputs.iter().enumerate() {
```
davidrusu: Since "output_index" is such an important concept in utxo's, what do you think if the ledger_tx should provide something like a `ledger_tx.utxos() -> Iterator<Item=Utxo>` that encapsulates that output index logic.

ledger/nomos-ledger/src/cryptarchia/mod.rs
```diff
@@ -188,6 +188,60 @@ impl LedgerState {
...
+                .utxos
+                .insert(
+                    Utxo {
+                        tx_hash: tx.hash(),
+                        output_index: i,
+                        note: *output,
+                    }
+                    .id(),
+                    *output,
+                )
```
davidrusu: So after we compute the note id, we throw away the index and tx hash? So we won't be able to re-construct the utxo from the information stored in the node right? we will only be able to verify a utxo by checking it against the note id?

ledger/nomos-ledger/src/cryptarchia/mod.rs
```diff
@@ -188,6 +188,60 @@ impl LedgerState {
             .update_nonce(proof.entropy(), slot))
     }
 
+    pub fn try_apply_tx<Id, Constants: GasConstants>(
+        mut self,
+        tx: impl AuthenticatedMantleTx,
```
davidrusu, line 193: Any reason why we need the full MantleTx here? can we pass only the LedgerTx instead?

nomos-core/proof_statements/src/zksig.rs
```diff
@@ -0,0 +1,7 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ZkSignaturePublic {
+    pub tx_hash: [u8; 32],
```
davidrusu: This "ZkSignature" was supposed to be more general in that it accepted an arbitrary message as input (to model a general signature scheme).

But I'm not sure if we use these ZkSigs outside of the mantle tx..

I would say that since this is in Nomos Core, then these zk sigs should not be tied directly to mantle tx through this hash. It would be nice to have these be generic signature schemes

ledger/nomos-ledger/src/cryptarchia/mod.rs
zeegomo: You can reconstruct every utxo if you have the full blockchain history (which we do), but that's not of particular interest in the ledger and we don't keep it ready in memory. Is there a usecase where do you think it's useful?

ledger/nomos-ledger/src/cryptarchia/mod.rs
zeegomo: We also need to
* verify the tx proof (and we only know at this point the pks to use as pub input)
* calculate gas fees, for which we need information about the total size of the tx (so proofs and ops)
It felt easier to pass the full mantle tx instead of adding these pieces somehow

ledger/nomos-ledger/src/cryptarchia/mod.rs
davidrusu: No, just was a bit surprising that we throwaway the utxo after generating the note id. No reason to change anything, just confirming

ledger/nomos-ledger/src/cryptarchia/mod.rs
davidrusu: You mean verifying the ledger tx proof? I think an API like this is reasonable:

    ledger.try_apply_tx(ledger_tx, ledger_tx_proof)

And I feel like gas fees should be computed when we are processing the mantle transaction in the mantle (not in the ledger) since the gas fees will be added to the mantle service pooled fees there and that's where we evaluate all the mantle operations.

zeegomo: > ledger.try_apply_tx(ledger_tx, ledger_tx_proof)

I had this before indeed, but then I changed when I had to add fees (we need to pass the calculated value of the fee one way or the other anyway). I'm more than open to change, but I think this will become clearer in the future anyway

ledger/nomos-ledger/src/cryptarchia/mod.rs
youngjoon-lee: I have a question. Is it okay to remove input notes from the UTXO tree, even if the proof is invalid?

https://github.com/logos-co/nomos/blob/fe8a3720c34c5b4413eba4de70c74f013463432b/ledger/nomos-ledger/src/cryptarchia/mod.rs#L198-L203

nomos-core/proof_statements/src/zksig.rs
```diff
@@ -0,0 +1,7 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ZkSignaturePublic {
+    pub msg_hash: [u8; 32],
+    pub pks: Vec<[u8; 32]>,
```
youngjoon-lee, line 6: nit: might be better to rename this to `public_keys`, although `pks` is idiomatic?

ledger/nomos-ledger/src/cryptarchia/mod.rs
zeegomo: Overall, the whole ledger is structured so that either all changes are applied or none at all. 
In this case we consume the ledger state and if there's an error somewhere we throw it away, that ensures that you can't be in a state where you have removed the utxos but the proof is invalid

nomos-core/proof_statements/src/zksig.rs
zeegomo: my personal preference would be `pks` since it's shorter and I think it's pretty clear from the context (e.g. a signature struct) what it means. Open to change though

ledger/nomos-ledger/src/cryptarchia/mod.rs
youngjoon-lee: Ah, I got it. Thank you.

PR #1468, opened 2025-07-17
refactor(blend): separate core/edge modules clearly

youngjoon-lee:
## 1. What does this PR implement?

This is a preparation for implementing the edge node service: https://github.com/logos-co/nomos/issues/1463.

This PR defines `nomos-blend-network::core` and `nomos-blend-network::edge` clearly. It would be useful because we decided to define the separate edge node service and proxy service. 

Previously, the `EdgeToCoreConnectionHandler` (for edge nodes) was defined next to the `CoreToEdgeConnectionHandler` (for core nodes). However, `EdgeToCoreConnectionHandler` is to be used by `NetworkBehaviour` for edge nodes (not core nodes). 
So, I think it's better to restructure modules like below:
```
nomos-blend/network
└── src
    ├── core
    │   ├── behaviour.rs
    │   ├── conn_maintenance.rs
    │   ├── error.rs
    │   ├── handler
    │   │   ├── core_edge
    │   │   │   ├── mod.rs
    │   │   │   └── ...
    │   │   ├── core.rs
    │   │   └── mod.rs
    │   ├── mod.rs
    │   └── ...
    ├── edge
    │   ├── handler
    │   │   ├── mod.rs
    │   │   └── ...
    │   └── mod.rs
    ├── lib.rs
    └── tests
        └── ...
```

This PR doesn't change any logic. Just moves files into modules and updates some imports.

## 2. Does the code have enough context to be clearly understood?

Yes

## 3. Who are the specification authors and who is accountable for this PR?

@ntn-x2 @youngjoon-lee 

## 4. Is the specification accurate and complete?

Yes

## 5. Does the implementation introduce changes in the specification?

No

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [x] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [x] 2. Description added.
* [x] 3. Context and links to Specification document(s) added.
* [x] 4. Main contact(s) (developers and specification authors) added
* [x] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 6. Link PR to a specific milestone.

Issue #1463, opened 2025-07-17
Edge node service


youngjoon-lee:
Add an Overwatch service that implements the edge node logic. This is for separating the edge node logic from the core node logic for easy maintenance, considering the huge difference between them.

Use the connection handler implemented by https://github.com/logos-co/nomos/pull/1414.

This service will be integrated with the blend proxy service that will be implemented by https://github.com/logos-co/nomos/issues/1464.

PR #1473, opened 2025-07-17
feat(blend): NetworkBehaviour for edge nodes

youngjoon-lee:

This is a preparation for implementing the edge node service: https://github.com/logos-co/nomos/issues/1463.
This is based on the PR #1468.

This PR implements `NetworkBehaviour` for edge nodes, which will be used by the edge node service. Previously, we've had only `NetworkBehaviour` for core nodes.

The behavior of edge nodes is simple: [Spec](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81f58af4c606db92f20a)
- They establish an ad-hoc connection to a core node randomly selected, whenever they want to send a data message to the Blend network.
- After sending a message, the connection should be closed. This has been already implemented by https://github.com/logos-co/nomos/pull/1414.
- No connection maintenance needed.

Yes

@madxor @youngjoon-lee @ntn-x2 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/network/src/tests/mod.rs
+++ b/nomos-blend/network/src/tests/mod.rs
@@ -5,7 +5,7 @@ use core::{
 
 use edge_core::dummy_edge_swarm;
 use futures::{task::noop_waker_ref, StreamExt as _};
-use libp2p::swarm::{dial_opts::DialOpts, ListenError, SwarmEvent};
+use libp2p::swarm::{dial_opts::DialOpts, DialError, ListenError, SwarmEvent};
 use nomos_blend_scheduling::membership::Membership;
 use tokio::time::sleep;
 
@@ -17,6 +17,7 @@ use crate::{
 mod core_edge;
 mod edge_core;
 
+/// Tests that the [`crate::edge::Behaviour`] denies inbound connections.
 #[test_log::test(tokio::test)]
 async fn core_to_edge_connection_failure() {
     let (mut edge_node, edge_node_address) = edge_sender_swarm(None).await;
@@ -63,6 +64,9 @@ async fn core_to_edge_connection_failure() {
     .await;
 }
 
+/// Tests that the [`crate::edge::Behaviour`] dials a core node
+/// as soon as a new message is scheduled, and sends the message
+/// throught the established connection.
 #[test_log::test(tokio::test)]
 async fn message_sending() {
     let (mut core_node, core_node_info) = core_receiver_swarm(Duration::from_secs(1)).await;
@@ -114,11 +118,69 @@ async fn message_sending() {
     .await;
 }
 
+/// Tests that the [`crate::edge::Behaviour`] closes the connection
+/// that was established without any message assigned to it.
+#[test_log::test(tokio::test)]
+async fn edge_to_core_connection_without_message_assigned() {
+    let (mut edge_node, _) = edge_sender_swarm(None).await;
+    let (mut core_node, core_node_info) = core_receiver_swarm(Duration::from_secs(1)).await;
+    let core_node_peer_id = *core_node.local_peer_id();
+    let edge_node_peer_id = *edge_node.local_peer_id();
+
+    // As a Swarm user, dial the core node without any message assigned to it,
+    // expecting that it is closed by [`crate::edge::Behaviour`].
+    edge_node
+        .dial(DialOpts::from(core_node_info.address))
+        .expect("Failed to connect to core node.");
+
+    let mut core_loop_done = false;
+    let mut edge_loop_done = false;
+    let mut cx = Context::from_waker(noop_waker_ref());
+    async {
+        loop {
+            if !core_loop_done {
+                let core_node_event = core_node.poll_next_unpin(&mut cx);
+                if let Poll::Ready(Some(SwarmEvent::ConnectionEstablished { peer_id, .. })) =
+                    core_node_event
+                {
+                    if peer_id == edge_node_peer_id {
+                        core_loop_done = true;
+                    }
+                }
+            }
+
+            if !edge_loop_done {
+                let edge_node_event = edge_node.poll_next_unpin(&mut cx);
+                if let Poll::Ready(Some(SwarmEvent::OutgoingConnectionError {
+                    peer_id,
+                    error: DialError::Denied { .. },
+                    ..
+                })) = edge_node_event
+                {
+                    if peer_id == Some(core_node_peer_id) {
+                        edge_loop_done = true;
+                    }
+                }
+            }
+
+            if core_loop_done && edge_loop_done {
+                break;
+            }
+            sleep(Duration::from_millis(100)).await;
+        }
+    }
+    .await;
+}
+
+/// Tests that a core node gets a timeout error if a connected edge node
+/// doesn't send any message within the specified timeout period.
 #[test_log::test(tokio::test)]
 async fn sender_timeout() {
     // Set timeout to 1. If an edge node doesn't send any message within 1 second,
     // the core node should get a timeout error.
     let (mut core_node, core_node_info) = core_receiver_swarm(Duration::from_secs(1)).await;
+    // Initialize an edge node with a dummy behaviour that establishes a connection
+    // but doesn't send any message.
     let (mut edge_node, _) = dummy_edge_swarm().await;
     let core_node_peer_id = *core_node.local_peer_id();
     let edge_node_peer_id = *edge_node.local_peer_id();


nomos-blend/network/src/core/error.rs
```diff
@@ -2,11 +2,11 @@ use std::io;
...
 use libp2p::{swarm::ConnectionId, PeerId};
 
-#[derive(Debug)]
+#[derive(Debug, thiserror::Error)]
 pub enum Error {
-    /// There were no peers to send a message to.
+    #[error("No peers to send a message to")]
     NoPeers,
-    /// IO error from peer
+    #[error("IO error from peer {peer_id:?} connection {connection_id:?}: {error}")]
```
ntn-x2:
```suggestion
    #[error("IO error from peer {peer_id:?} and connection {connection_id:?}: {error}")]
```

nomos-blend/network/src/edge/behaviour.rs
```diff
@@ -0,0 +1,306 @@
...
+    fn handle_established_outbound_connection(
+        &mut self,
+        _connection_id: ConnectionId,
+        peer_id: PeerId,
+        _addr: &Multiaddr,
+        _role_override: Endpoint,
+        _port_use: PortUse,
+    ) -> Result<THandler<Self>, ConnectionDenied> {
+        // If no membership is provided (for tests), we assume all peers are core nodes.
+        // nodes.
```
ntn-x2:
```suggestion
```

nomos-blend/network/src/edge/handler/mod.rs
```diff
@@ -135,10 +137,13 @@ pub enum FailureReason {
 
 #[derive(Debug)]
 pub enum ToBehaviour {
+    /// Notify the behaviour that the connection handler is ready to send a
+    /// message.
+    ReadyToSend,
```
ntn-x2: Why do we need this and `DropSubstream` in `FromBehaviour`? Can't we treat the connection handler as a black box which returns `Ok` or `Err` and react on it accordingly? What I am thinking is that whenever we need to send a new message, we create a new connection handler, and if it returns `Ok`, we consider the message sent. If it returns `Err`, we assume the substream is closed and we can re-schedule sending with a different connection handler. I think it would make it easier to understand. So we would not need the additional overhead of signaling `Ready` to the behaviour or for the behaviour to signal `DropSubstream` to the connection handler. Each connection handler has "exclusive" access to the message being sent, which is returned back to the behaviour in case something happens. Failure cases, e.g. failed to establish a connection, should be handled by the connection handler state machine, in my opinion. What is your opinion on this?

nomos-blend/network/src/edge/handler/message_set.rs
```diff
@@ -30,6 +32,16 @@ impl From<MessageSetState> for ConnectionState {
 }
 
 impl StateTrait for MessageSetState {
+    fn on_behaviour_event(self, event: FromBehaviour) -> ConnectionState {
+        match event {
+            FromBehaviour::Message(_) => self.into(),
```
ntn-x2: I mean, the goal of the state machine is to avoid invalid transitions. I think in this case we should log a warning line before returning `self.into()` to explicitly log that we are ignoring the input message.

nomos-blend/network/src/edge/handler/message_set.rs
```diff
@@ -30,6 +32,16 @@ impl From<MessageSetState> for ConnectionState {
...
+            FromBehaviour::DropSubstream => {
+                tracing::trace!(target: LOG_TARGET, "MessageSetState -> DroppedState by request from behaviour.");
+                DroppedState::new(None, self.waker).into()
```
ntn-x2:
```suggestion
                DroppedState::new(None, self.waker.take()).into()
```

nomos-blend/network/src/edge/behaviour.rs
```diff
@@ -0,0 +1,306 @@
...
+
+        let opts = self.new_dial_opts(node.id, &node.address);
+        self.requested_dials.insert((node.id, opts.connection_id()));
+        self.events.push_back(ToSwarm::Dial { opts });
+        self.try_wake();
+        Ok(())
+    }
+
+    /// Creates a new [`DialOpts`] for the given peer.
+    /// Repeats until a new connection ID (not exist in [`requested_dials`]) is
```
ntn-x2:
```suggestion
    /// Repeats until a new connection ID (not existing in [`requested_dials`]) is
```

nomos-blend/network/src/edge/behaviour.rs
```diff
@@ -0,0 +1,306 @@
...
+    /// generated.
+    fn new_dial_opts(&self, peer_id: PeerId, address: &Multiaddr) -> DialOpts {
+        loop {
+            let opts = DialOpts::peer_id(peer_id)
+                .condition(PeerCondition::DisconnectedAndNotDialing)
```
ntn-x2: Why this and not always? Each message should be independently scheduled, right? So why do we care if we are already connected to the peer we are dialing? Is this because of the `During the session, it selects at random (without repetition) a node from that set.` in the spec? That part of the spec feels to me not very clear, since it also states that connections should be opened at the beginning of each session, when that is not really the case. I think we can do better, perhaps after talking to Marci. If we have 10 messages to send but only 5 core nodes, with the current approach it would takes us 2x the time, when we could simply randomly select a core node for each message, which might result in the highly unlikely scenario in which we send all 10 messages to the same core node. If this has already been discussed that's ok, else I would like to highlight this in this case.

nomos-blend/network/src/edge/behaviour.rs
```diff
@@ -0,0 +1,306 @@
...
+                .addresses(vec![address.clone()])
+                .build();
+            if !self
```
ntn-x2: I feel this check belongs rather to the caller site, where the set is also updated. This function should return a simple dial result, and the caller should then do whatever they please with it, including checking if the set already contains such an element.

nomos-blend/network/src/edge/behaviour.rs
```diff
@@ -0,0 +1,306 @@
...
+
+    fn try_wake(&mut self) {
+        if let Some(waker) = self.waker.take() {
+            waker.wake();
+        }
+    }
+
+    // TODO: Remove this method once the session stream is implemented.
+    //       https://github.com/logos-co/nomos/issues/1462
+    pub fn set_membership(&mut self, membership: Membership<PeerId>) {
```
ntn-x2: I think this is only used for tests, at least in this PR, right? Has this call been integrated into the libp2p stack in a different PR? Otherwise I'd mark this as a `test`-only function, if you agree.

nomos-blend/network/src/tests/mod.rs
```diff
@@ -181,8 +148,6 @@ async fn sender_timeout() {
                 }
             }
 
-            // We stop after verifying that the connection was established, so by not
```
ntn-x2, line 184: Is the comment not relevant anymore?

nomos-blend/network/src/tests/edge_core.rs
```diff
@@ -1,121 +1,130 @@
...
+    Behaviour: NetworkBehaviour,
+{
+    loop {
+        if let Some(SwarmEvent::NewListenAddr { address, .. }) = swarm.next().await {
+            break address;
         }
     }
+}
+
+pub struct NoMessageSendBehaviour {
```
ntn-x2: Can you add a comment on what this test behaviour does, and what it allows to test?

nomos-blend/network/src/edge/handler/mod.rs
ntn-x2: This would also solve the ["not so nice thing"](https://github.com/logos-co/nomos/pull/1473/files#diff-8516153a181e8a1aa94b54b09ebd2362976f0e6fbdfff25bcfbefac6006f7cdeR250), currently in the PR, that we might close a substream right after opening it because there are no messages to send, which is a bit counter-intuitive and can be avoided if we only open a connection when a new message comes in, and do not allocate that message to a different one until the first connection handler either succeeds or fails.

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: Now I think we can use `Always`. I will change it. It's better and simpler.
My initial intention was the sentence that you referred to. But, the sentence is not so clear. 
Let's go simple.

nomos-blend/network/src/edge/handler/mod.rs
youngjoon-lee: I wanted to discuss about this topic.
All stories are started from the async dialings (which also might be not initiated by this NetworkBehaviour).
I had the same thought as you explained, but it was not simple. Let me try some changes in the code now before leaving details in this comment.

youngjoon-lee: Improved in 8674163d3b249398d98637b01fb3fc28b4607d6b. I recommend you to do a full review again, which might be easier, instead of reviewing this specific commit.

I simplified the NetworkBehaviour and ConnectionHandler a lot. 
In short, now NetworkBehaviour assigns a message to a specific dialing. Once the dialing succeeded, we creates a ConnectionHandler and assign the message to it. If the dialing fails, NetworkBehaviour re-requests another dialing for the specific message.
If any error occurs from the ConnectionHandler, NetworkBehaviour re-requests another dialing, similarly.

I found we don't need two ConnectionHandler states: ReadyToSend and MessageSet, because now we assign a message to a ConnectionHandler when it is created. So, I removed them.

nomos-blend/network/src/edge/handler/message_set.rs
youngjoon-lee: The MessageSetState has been removed in [8674163](https://github.com/logos-co/nomos/pull/1473/commits/8674163d3b249398d98637b01fb3fc28b4607d6b)

nomos-blend/network/src/edge/handler/message_set.rs
youngjoon-lee: The MessageSetState has been removed in [8674163](https://github.com/logos-co/nomos/pull/1473/commits/8674163d3b249398d98637b01fb3fc28b4607d6b)

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: Resolved in [8674163](https://github.com/logos-co/nomos/pull/1473/commits/8674163d3b249398d98637b01fb3fc28b4607d6b)

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: We don't have this check anymore [8674163](https://github.com/logos-co/nomos/pull/1473/commits/8674163d3b249398d98637b01fb3fc28b4607d6b)

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: We don't have this method anymore [8674163](https://github.com/logos-co/nomos/pull/1473/commits/8674163d3b249398d98637b01fb3fc28b4607d6b)

nomos-blend/network/src/tests/mod.rs
youngjoon-lee: Updated the comment [8674163](https://github.com/logos-co/nomos/pull/1473/commits/8674163d3b249398d98637b01fb3fc28b4607d6b)

nomos-blend/network/src/tests/edge_core.rs
youngjoon-lee: We don't have this test behaviour anymore [8674163](https://github.com/logos-co/nomos/pull/1473/commits/8674163d3b249398d98637b01fb3fc28b4607d6b)

nomos-blend/network/src/edge/behaviour.rs
```diff
@@ -0,0 +1,250 @@
...
+
+    fn handle_established_inbound_connection(
+        &mut self,
+        _connection_id: ConnectionId,
+        _peer_id: PeerId,
+        _local_addr: &Multiaddr,
+        _remote_addr: &Multiaddr,
+    ) -> Result<THandler<Self>, ConnectionDenied> {
+        tracing::debug!(target: LOG_TARGET, "Connection established from a non-core node. Creating a connection handler with DroppedState.");
+        Ok(EdgeToCoreBlendConnectionHandler::new_dropped())
```
ntn-x2: Why this instead of denying the connection as it was before?

nomos-blend/network/src/edge/behaviour.rs
```diff
@@ -0,0 +1,250 @@
...
+        } else {
+            self.waker = Some(cx.waker().clone());
+            Poll::Pending
+        }
+    }
+}
+
+impl<Rng> Behaviour<Rng>
+where
+    Rng: RngCore,
```
ntn-x2: I think we should split the impl into two blocks, one where `RngCore` is required and one for things like `new`, `is_core_node` etc. where it's not.

nomos-blend/network/src/edge/behaviour.rs
```diff
@@ -0,0 +1,250 @@
...
+    fn handle_established_outbound_connection(
+        &mut self,
+        connection_id: ConnectionId,
+        peer_id: PeerId,
+        _addr: &Multiaddr,
+        _role_override: Endpoint,
+        _port_use: PortUse,
+    ) -> Result<THandler<Self>, ConnectionDenied> {
+        if !self.is_core_node(&peer_id) {
+            tracing::debug!(target: LOG_TARGET, "Connection established to a non-core node. Creating a connection handler with DroppedState.");
```
ntn-x2: Do you think this should rather be a warning? Technically since we control the dialing, this should never happen. In case this happens is probably only on the edge of a session rotation, so probably a warning or just a trace should be sufficient. What do you think?

nomos-blend/network/src/edge/behaviour.rs
```diff
@@ -0,0 +1,250 @@
...
+            self.pending_messages.remove(&(peer_id, connection_id));
+            return Ok(EdgeToCoreBlendConnectionHandler::new_dropped());
+        }
```
ntn-x2: Also here, why do we need to do this? If we instantiate it with `(None, None)` it will always return `Poll::Pending`. If we need to return an error so that the message should be scheduled again for sending, the error should be `Some`. But in general I don't see why this is needed, maybe you can help me understand that.

nomos-blend/network/src/edge/behaviour.rs
```diff
@@ -0,0 +1,250 @@
...
+        // Reschedule a new dial only if the failure is from
+        // the connection that this NetworkBehaviour has requested.
+        let Some(peer_id) = failure.peer_id else {
+            return;
+        };
+        let Some(message) = self
+            .pending_messages
+            .remove(&(peer_id, failure.connection_id))
+        else {
+            return;
```
ntn-x2: Should we add an `error` log here?

nomos-blend/network/src/edge/behaviour.rs
```diff
@@ -0,0 +1,250 @@
...
+    // TODO: Replace with the session stream and make this a non-Option
+    //       https://github.com/logos-co/nomos/issues/1462
+    current_membership: Option<Membership<PeerId>>,
+    rng: Rng,
+}
+
+#[derive(Debug)]
+pub enum EventToSwarm {
+    /// Notify the swarm that the message was sent successfully.
+    MessageSuccess(Vec<u8>),
```
ntn-x2: Should we return the peer to which the message was forwarded?

nomos-blend/network/src/edge/behaviour.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// Handles [`ToBehaviour::SendError`] event from the connection handler
+    /// by rescheduling a new dial.
+    fn handle_send_error(
+        &mut self,
+        peer_id: PeerId,
+        connection_id: ConnectionId,
+        reason: &FailureReason,
+    ) {
+        tracing::error!(target: LOG_TARGET, "Failed to send message. Reason: {reason:?}. Rescheduling dial.");
+        if let Some(message) = self.pending_messages.remove(&(peer_id, connection_id)) {
```
ntn-x2: Should we still add error logs in case this is not true? I feel it would help us with testing and debugging in the future.

nomos-blend/network/src/edge/handler/mod.rs
```diff
@@ -154,10 +124,7 @@ impl ConnectionHandler for EdgeToCoreBlendConnectionHandler {
         SubstreamProtocol::new(DeniedUpgrade, ())
     }
 
-    fn on_behaviour_event(&mut self, event: Self::FromBehaviour) {
-        let state = self.state.take().expect("Inconsistent state");
```
ntn-x2: Why was this removed? Do you think it might not make sense to still have it in, to avoid we change `FromBehaviour` in the future but don't update the `on_behaviour_event` impl, which is now just `{}`?

nomos-blend/network/src/tests/mod.rs
```diff
@@ -87,8 +44,8 @@ async fn core_to_edge_connection_failure() {
 
             if !edge_loop_done {
                 let edge_node_event = edge_node.poll_next_unpin(&mut cx);
-                if let Poll::Ready(Some(SwarmEvent::IncomingConnectionError {
-                    error: ListenError::Denied { .. },
+                if let Poll::Ready(Some(SwarmEvent::ConnectionClosed {
+                    cause: Some(ConnectionError::KeepAliveTimeout),
```
ntn-x2: Yeah this is a consequence of creating `Dropped(None, None)` connection handlers, I think. We should not really fail with a timeout here, since a connection to an edge node should be forbidden altogether.

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: I considered the case where this `NetworkBehaviour` is used along with other `NetworkBehaviour`s (which potentially can dial additional connections). I didn't want to let this `NetworkBehaviour` close the connection by returning `ConnectionDenied`. But, it's up to us. We can declare that "This `NetworkBehaviour` can deny connections that are not necessary for the blend edge node".  I'd like to hear your opinion regarding this aspect.

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: I will change this to warn.

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: This is because of the same reason I commented above: https://github.com/logos-co/nomos/pull/1473/files/8674163d3b249398d98637b01fb3fc28b4607d6b#r2224855950.

But, we could simply return `ConnectionDenied`. Let's discuss there.

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: I don't think so, because we can never control how Swarm initiates dials. No matter how we design this `NetworkBehaviour`, the Swarm user can initiate a lot of dials that are not relevant to this `NetworkBehaviour`. If some of those dials fail, this `NetworkBehaviour` will be notified anyway.
I don't want to print `error` logs for the failures of those dials.

youngjoon-lee: Okay, I just saw another comment: https://github.com/logos-co/nomos/pull/1473/files/8674163d3b249398d98637b01fb3fc28b4607d6b#r2222147755. I will add a debug log.

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: I will add a debug log, as mentioned in https://github.com/logos-co/nomos/pull/1473/files/8674163d3b249398d98637b01fb3fc28b4607d6b#r2222143205

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: It's not necessary right now, but I agree it will be useful, at least for debugging purposes. I will add it.

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: f4e2bebad843a241567dbcd560e344327313a8c5 Now we deny connections.

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: f4e2bebad843a241567dbcd560e344327313a8c5 i changed it to warn.

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: f4e2bebad843a241567dbcd560e344327313a8c5 now we deny connections.

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: f4e2bebad843a241567dbcd560e344327313a8c5 I added a warn log

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: Sorry, this is a bug. I fixed it 6daf6f8a795e3f304d4ed9802af1c905cc7c8bca. The `pending_messages` should be only for managing pending dials. After the connection is established, and after the `ConnectionHandler` is created, the message should be removed from the `pending_messages`, and should be assigned to the `ConnectionHandler`. If the `SendError` occurs from the `ConnectionHander`, the message should be contained in the `SendError`. We shouldn't try to find the message from the `pending_messages` (because it has been already removed). So, in the commit, I renamed `pending_messages` to `pending_dials`, and added the `message` field to the `SendError`.

nomos-blend/network/src/tests/mod.rs
youngjoon-lee: Reverted https://github.com/logos-co/nomos/commit/6daf6f8a795e3f304d4ed9802af1c905cc7c8bca

nomos-blend/network/src/edge/handler/mod.rs
youngjoon-lee: You're right. Reverted b16498cc

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: Refactored 1595793b

nomos-blend/network/src/tests/edge_core.rs
youngjoon-lee: My previous commit made a bug in the `sender_timeout` test. It makes the test occasionally failed.
I fixed the test by simply defining a dummy behaviour and a dummy connection handler for an edge node, which opens a substream but doesn't send any data. 35fe4cae

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: done e0bed213

nomos-blend/network/src/edge/handler/mod.rs
```diff
@@ -82,16 +71,14 @@ impl ConnectionState {
...
-            Self::ReadyToSend(s) => s.poll(cx),
             Self::Sending(s) => s.poll(cx),
             Self::Dropped(s) => s.poll(cx),
         }
     }
 }
 
 trait StateTrait: Into<ConnectionState> {
-    fn on_behaviour_event(self, _event: FromBehaviour) -> ConnectionState {
+    fn on_behaviour_event(self) -> ConnectionState {
```
ntn-x2:
```suggestion
    fn on_behaviour_event(self, _event: FromBehaviour) -> ConnectionState {
```

nomos-blend/network/src/edge/handler/mod.rs
```diff
@@ -154,9 +135,9 @@ impl ConnectionHandler for EdgeToCoreBlendConnectionHandler {
         SubstreamProtocol::new(DeniedUpgrade, ())
     }
 
-    fn on_behaviour_event(&mut self, event: Self::FromBehaviour) {
+    fn on_behaviour_event(&mut self, _event: Self::FromBehaviour) {
         let state = self.state.take().expect("Inconsistent state");
-        self.state = Some(state.on_behaviour_event(event));
+        self.state = Some(state.on_behaviour_event());
```
ntn-x2:
```suggestion
        self.state = Some(state.on_behaviour_event(event));
```

nomos-blend/network/src/edge/behaviour.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    fn handle_established_inbound_connection(
+        &mut self,
+        _connection_id: ConnectionId,
+        _peer_id: PeerId,
+        _local_addr: &Multiaddr,
+        _remote_addr: &Multiaddr,
+    ) -> Result<THandler<Self>, ConnectionDenied> {
+        let reason = ConnectionDeniedReason::InboundNotAllowed;
+        tracing::warn!(target: LOG_TARGET, "Denying the connection: {reason:?}");
```
ntn-x2, line 73: Now that I better think about this, probably a trace would be the best target. Debug is not really useful, and I think info would be too much. But it's probably not a warning since we can't really control what happens with our peers.

nomos-blend/network/src/edge/behaviour.rs
```diff
@@ -0,0 +1,275 @@
...
+        &mut self,
+        connection_id: ConnectionId,
+        peer_id: PeerId,
+        _addr: &Multiaddr,
+        _role_override: Endpoint,
+        _port_use: PortUse,
+    ) -> Result<THandler<Self>, ConnectionDenied> {
+        if !self.is_core_node(&peer_id) {
+            let reason = ConnectionDeniedReason::OutboundToEdgeNodeNotAllowed;
+            tracing::warn!(target: LOG_TARGET, "Denying the connection: {reason:?}");
```
ntn-x2, line 87: Here warning makes more sense since we know we should not be connected to another edge node.

nomos-blend/network/src/edge/behaviour.rs
```diff
@@ -0,0 +1,275 @@
...
+        membership.contains_remote(peer_id)
+    }
+
+    /// Schedules a [`EventToSwarm`] to be sent to the swarm.
+    fn schedule_event_to_swarm(&mut self, event: EventToSwarm) {
+        self.events.push_back(ToSwarm::GenerateEvent(event));
+        self.try_wake();
+    }
+
+    fn get_messsage_of_failed_dial(&mut self, failure: DialFailure) -> Option<Vec<u8>> {
```
ntn-x2:
```suggestion
    fn get_message_of_failed_dial(&mut self, failure: DialFailure) -> Option<Vec<u8>> {
```

nomos-blend/network/src/edge/behaviour.rs
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    /// Handles [`FromSwarm::DialFailure`] event from the swarm
+    /// by rescheduling a new dial if necessary.
+    fn handle_dial_failure(&mut self, failure: DialFailure) {
+        tracing::error!(target: LOG_TARGET, "Dial failure: {failure:?}");
+
+        // Reschedule a new dial only if the failure is from
+        // the connection that this NetworkBehaviour has requested.
+        if let Some(message) = self.get_messsage_of_failed_dial(failure) {
```
ntn-x2: I think the case where this is `None` should never happen. If it does, in this case probably an error or at least a warning should be logged. Something is going wrong.

nomos-blend/network/src/tests/edge_core.rs
```diff
@@ -1,44 +1,70 @@
...
 
-impl NetworkBehaviour for TestEdgeSenderBehaviour {
-    type ConnectionHandler = EdgeToCoreBlendConnectionHandler;
-    type ToSwarm = THandlerOutEvent<Self>;
+/// A dummy [`NetworkBehaviour`] that creates a [`DummyConnectionHandler`]
+/// for outbound connections.
+pub struct DummyBehaviour;
+
+impl NetworkBehaviour for DummyBehaviour {
+    type ConnectionHandler = DummyConnectionHandler;
```
ntn-x2, line 66: The idea here was to test edge->core and core->edge connection handlers, which is something in between a unit and an integration test. Why did you think it would be better to use a dummy connection handler instead? Where do we test `EdgeToCoreBlendConnectionHandler` now?

ntn-x2 (commented):
Thanks for this! Apart from some minor comment, one question I have left is: how do we test `EdgeToCoreBlendConnectionHandler` now? I don't think we do anymore, do we?

nomos-blend/network/src/edge/handler/mod.rs
youngjoon-lee: Hm. I removed the `FromBehaviour` struct because now we don't have any `ConnectionState` that receives it. Do you think it's better to have an empty enum `FromBehaviour` as a placeholder in the future?

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: That case can happen if a swarm user explicitly dial to a certain node and if the dial fails, regardless of the implementation of this NetworkBehaviour. I think that failure can be silently ignored by this NetworkBehaviour because it's out of interest.
Unlike https://github.com/logos-co/nomos/pull/1473#discussion_r2226841399, I didn't print any log because this behaviour doesn't do anything special for the failure. I think there's nothing to notify to the swarm user.
Probably, a debug log would be useful, but TBH, I think it's not really necessary.

nomos-blend/network/src/edge/behaviour.rs
youngjoon-lee: TBH, warning is fine. We're closing a connection here. If a swarm user (who doesn't clearly understand the details of this behaviour) wanted to listen for an inbound connection, this warning log is useful to inform that this behaviour is closing the connection aggressively.

nomos-blend/network/src/edge/handler/mod.rs
youngjoon-lee: 8e8e0e14 I understand your point. The method `on_behaviour_event` doesn't make sense without the `event` argument.

nomos-blend/network/src/tests/edge_core.rs
youngjoon-lee: The rationale is that I wanted to use our new `crate::edge::Behaviour` for testing, instead of introducing the previous `TestEdgeSenderBehaviour`.

This `DummyBehaviour` and `DummyConnectionHandler` are for the `sender_timeout` test, which expects that a core node gets a timeout error if an edge node doesn't send any message via a substream for a long time. Since our new `crate::edge::Behaviour` always does the right thing, it's not possible to simulate the timeout scenario. So, I introduces a very simple `DummyBehaviour`, which I think it's fine because the purpose of the `sender_timeout` test is for testing the behaviour of core node, not edge node.

Another reason is that I removed the `edge_to_core_connection_establishment` test. The edge->core connection establishment is now automatically tested by the `message_sending` test if we use `crate::edge::Behaviour`. Now the dial is automatically requested as soon as a user calls `edge_node.behaviour_mut().send_message(...)`. 

On the other hand, I left `TestCoreReceiverBehaviour` as it is, because all of our tests don't actually test our `crate::core::Behaviour`.

youngjoon-lee: I just added an additional test: `edge_to_core_connection_without_message_assigned`: [41abb4a](https://github.com/logos-co/nomos/pull/1473/commits/41abb4a8bffa74bbb7b467e38bf74910e205347a).

BTW, I feel we can move these tests into each `core` and `edge` modules.

youngjoon-lee: I wanna close this PR and replace this with https://github.com/logos-co/nomos/pull/1493, which looks optimal in my opinion. Then, all of these discussions can be closed.

ntn-x2: Let's close this then, the linked PR is a better solution.

youngjoon-lee: Closing this. 

NOTE: This PR has been replaced with https://github.com/logos-co/nomos/pull/1493

PR #1477, opened 2025-07-18
feat(membership): Integrate subnetwork assignations from spec

danielSanchezQ:

This PR includes the new HistoryAware membership handler into the does themselves. It is lacking pluging proper randomness source which will be included later on.

Quite simple

@danielSanchezQ @madxor

It is accurate

No changes

## Checklist

> [!WARNING]  

* [x] 2. Description added.


bacv (approved):
Looks good!

nodes/nomos-executor/config.yaml
```diff
@@ -77,13 +77,8 @@ da_network:
...
     num_subnets: 2
   membership:
-    assignations:
-      - - 12D3KooWA5LkQcuefjUBTktxCXUQ4rx4PcqRY1tgbSpV9eZDHWir
-        - 12D3KooWKD62Q1nAAMHeV2Cx3rNNMzjgHkTT2edFMzHZ5ur7sGnh
-      - - 12D3KooWKD62Q1nAAMHeV2Cx3rNNMzjgHkTT2edFMzHZ5ur7sGnh
-        - 12D3KooWA5LkQcuefjUBTktxCXUQ4rx4PcqRY1tgbSpV9eZDHWir
     subnetwork_size: 2
-    dispersal_factor: 2
+    replication_factor: 5
```
danielSanchezQ: @bacv will this break integration tests? As no nodes are registered via sdp yet, all DA would fail...

bacv: Integration tests doesn't use this yaml, this is example config file. In `tests/nodes/{executor,validator}` module we should set this to what is reasonable for testing.

danielSanchezQ: what I mean is that what assignations would we be using? Because this takes no initial ones (as the old versions where doing). cc @pradovic

pradovic: Just wanted to add: integration tests fetch the membership from membership service which is loaded from config as the sdp is still not ready. So, the membership handler only needs the params.

danielSanchezQ: > Just wanted to add: integration tests fetch the membership from membership service which is loaded from config as the sdp is still not ready. So, the membership handler only needs the params.

Nice then!, we can go with that until we have all the other pieces ready. Thanks!

danielSanchezQ: two nodes happy is consistenly failing. Its probably something to deal with the assignations :thinking:

bacv: Seems that the node is not starting up, taking a look

bacv:
```
2025-07-18T13:57:45.189973Z ERROR nomos_tracing::panic: A panic occurred panic.payload="The network size is smaller than the replication factor" panic.location="/Users/bacv/code/logos/nomos-node/nomos-da/network/subnetworks-assignations/src/versions/history_aware_refill/assignations.rs:207:9" panic.backtrace="disabled backtrace" panic.note="run with RUST_BACKTRACE=1 environment variable to display a backtrace"
```
in tests lets try a smaller replication factor

danielSanchezQ: Ah fuck, I changed the node only, not the executor finally. Let me try.

danielSanchezQ: Yeah, this is never gonna work. Because there are no participation. We have two options:
* Define the minimum participation and disable everything if that threshold is not reach. This is no good for testing anything really...
* I can bring back temporary the initial membership and load the nodes. With proper config, those nodes will be added to the subnetworks and it should work. Later we would have to migrate to have the declarations in the genesis block.

wdyt @bacv @pradovic ?

pradovic: https://github.com/logos-co/nomos/pull/1481 - mostly assertion fixes.

PR #1476, opened 2025-07-18
refactor(blend): separate core/edge service modules

youngjoon-lee:
## 1. What does this PR implement?

(Based on PR #1475)

A preparation for https://github.com/logos-co/nomos/issues/1464.

Restructure blend service modules clearly: `core` and `edge`. Github doesn't show the diffs efficiently, but I didn't change any logic. I moved files across folders and updates some imports. So, you can skip reviewing file bodies.
Only one change I made is leaving only common structs in the `crate::message`. Core-specific message structs were moved to a new `crate::core::message`.
```
nomos-services/blend
└── src
    ├── core
    │   ├── backends
    │   │   ├── ...
    │   │   └── mod.rs
    │   ├── message.rs
    │   ├── mod.rs
    │   ├── network
    │   │   ├── ...
    │   │   └── mod.rs
    │   └── settings.rs
    ├── edge
    │   ├── backends
    │   │   ├── ...
    │   │   └── mod.rs
    │   ├── mod.rs
    │   └── settings.rs
    ├── lib.rs
    └── message.rs
```

We're going to add the `nomos-service/blend/proxy` module when implementing https://github.com/logos-co/nomos/issues/1464.

## 2. Does the code have enough context to be clearly understood?

Yes

## 3. Who are the specification authors and who is accountable for this PR?

@ntn-x2 @youngjoon-lee 

## 4. Is the specification accurate and complete?

Yes

## 5. Does the implementation introduce changes in the specification?
No

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [x] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [x] 2. Description added.
* [x] 3. Context and links to Specification document(s) added.
* [x] 4. Main contact(s) (developers and specification authors) added
* [x] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 6. Link PR to a specific milestone.

PR #1480, opened 2025-07-21
feat(blend): add/integrate proxy service (no core<>edge switching yet)

youngjoon-lee:

(Based on PR #1476)

- Adds `BlendProxyService`, which currently relays all messages to the Blend core service. At a next PR, I will implement turning on/off core and edge adapters.
- Integrates it into the system.

Yes

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/proxy/core/mod.rs
+++ b/nomos-services/blend/src/proxy/core/mod.rs
@@ -8,8 +8,6 @@ use crate::{
     message::ServiceMessage,
 };
 
-const LOG_TARGET: &str = "blend::service::proxy::core";
-
 #[async_trait::async_trait]
 pub trait Adapter<RuntimeServiceId> {
     type Backend: BlendBackend<Self::NodeId, RuntimeServiceId> + 'static;
--- a/nomos-services/blend/src/proxy/edge/mod.rs
+++ b/nomos-services/blend/src/proxy/edge/mod.rs
@@ -5,8 +5,6 @@ use overwatch::services::relay::OutboundRelay;
 
 use crate::{edge::backends::BlendBackend, message::ServiceMessage};
 
-const LOG_TARGET: &str = "blend::service::proxy::edge";
-
 #[async_trait::async_trait]
 pub trait Adapter<RuntimeServiceId> {
     type Backend: BlendBackend<Self::NodeId, RuntimeServiceId> + 'static;
--- a/nomos-services/api/Cargo.toml
+++ b/nomos-services/api/Cargo.toml
@@ -18,7 +18,6 @@ bytes                    = { workspace = true }
 chain-service            = { workspace = true, features = ["libp2p"] }
 futures                  = "0.3.31"
 kzgrs-backend            = { workspace = true }
-nomos-blend-service      = { workspace = true, features = ["libp2p"] }
 nomos-core               = { workspace = true }
 nomos-da-dispersal       = { workspace = true }
 nomos-da-indexer         = { workspace = true, features = ["rocksdb-backend"] }
--- a/nomos-services/blend/src/proxy/core/libp2p.rs
+++ b/nomos-services/blend/src/proxy/core/libp2p.rs
@@ -8,8 +8,10 @@ use crate::{
         BlendService,
     },
     message::ServiceMessage,
-    proxy::core::{Adapter, LOG_TARGET},
+    proxy::core::Adapter,
 };
+
+const LOG_TARGET: &str = "blend::service::proxy::core::libp2p";
 
 pub struct Libp2pAdapter<RuntimeServiceId> {
     relay: OutboundRelay<<BlendServiceType<RuntimeServiceId> as ServiceData>::Message>,
--- a/nomos-services/blend/src/proxy/edge/libp2p.rs
+++ b/nomos-services/blend/src/proxy/edge/libp2p.rs
@@ -5,8 +5,10 @@ use crate::{
     core::network::{libp2p::Libp2pAdapter as Libp2pNetworkAdapter, NetworkAdapter},
     edge::{backends::libp2p::Libp2pBlendBackend, BlendService},
     message::ServiceMessage,
-    proxy::edge::{Adapter, LOG_TARGET},
+    proxy::edge::Adapter,
 };
+
+const LOG_TARGET: &str = "blend::service::proxy::edge::libp2p";
 
 pub struct Libp2pAdapter<RuntimeServiceId> {
     relay: OutboundRelay<<BlendServiceType<RuntimeServiceId> as ServiceData>::Message>,


nomos-services/blend/src/proxy/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+    },
+    DynError, OpaqueServiceResourcesHandle,
+};
+use services_utils::wait_until_services_are_ready;
+
+use crate::message::ServiceMessage;
+
+const LOG_TARGET: &str = "blend::service::proxy";
+
+pub struct BlendProxyService<CoreAdapter, EdgeAdapter, RuntimeServiceId>
```
ntn-x2, line 23: I would actually make this a transparent proxy, and simply call THIS service `BlendService`: this is all that downstream services will see. They do not need to know that they are dealing with a proxy, as long as they can send messages to it.

nomos-services/blend/src/proxy/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+where
+    CoreAdapter: core::Adapter<RuntimeServiceId>,
```
ntn-x2, line 25: I think there's no need for this trait bound in this struct definition.

nomos-services/blend/src/proxy/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+        state::{NoOperator, NoState},
+        AsServiceId, ServiceCore, ServiceData,
+    },
+    DynError, OpaqueServiceResourcesHandle,
+};
+use services_utils::wait_until_services_are_ready;
+
+use crate::message::ServiceMessage;
+
+const LOG_TARGET: &str = "blend::service::proxy";
```
ntn-x2, line 21: Same here. I would personally not include the wording proxy in anything that is visible from outside this and the core/edge services. But we can discuss about this.

nomos-services/blend/src/proxy/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+            target: LOG_TARGET,
+            "Service '{}' is ready.",
+            <RuntimeServiceId as AsServiceId<Self>>::SERVICE_ID
+        );
+
+        wait_until_services_are_ready!(
+            &self.service_resources_handle.overwatch_handle,
+            Some(Duration::from_secs(60)),
+            crate::core::BlendService<_, _, _, _>,
+            crate::edge::BlendEdgeService<_, _, _, _>
```
ntn-x2: I'm seeing it here, sorry for not commenting in the right PR. Should we call `BlendEdgeService` also `BlendService`, since it's already namespaced by the proper module name?

nomos-services/blend/src/proxy/edge/mod.rs
```diff
@@ -0,0 +1,17 @@
...
+
+use crate::{edge::backends::BlendEdgeBackend, message::ServiceMessage};
+
+#[async_trait::async_trait]
+pub trait Adapter<RuntimeServiceId> {
+    type Backend: BlendEdgeBackend<Self::NodeId, RuntimeServiceId> + 'static;
+    type BroadcastSettings: 'static;
+    type NodeId: Clone;
+
+    fn new(relay: OutboundRelay<ServiceMessage<Self::BroadcastSettings>>) -> Self;
```
ntn-x2, line 14: We don't really need to deal with broadcast settings for an edge node. So to me this screams that we might need to refactor `ServiceMessage` into two separate structs. Then we can also get rid of the `BroadcastSettings` associated type on this trait.

nomos-services/blend/src/proxy/core/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+use crate::{
+    core::{backends::BlendBackend, network::NetworkAdapter},
+    message::ServiceMessage,
+};
+
+#[async_trait::async_trait]
+pub trait Adapter<RuntimeServiceId> {
+    type Backend: BlendBackend<Self::NodeId, RuntimeServiceId> + 'static;
+    type Network: NetworkAdapter<RuntimeServiceId> + 'static;
+    type NodeId;
```
ntn-x2, line 15: I would make `NodeId` a generic instead, since both core and edge are supposed to work with the same type. This generic can then be passed to be proxy service and down to each "proxied" service. `Backend` and `Network` should remain associated types.

nomos-services/chain-service/src/blend/libp2p.rs
```diff
@@ -12,41 +13,46 @@ use serde::{de::DeserializeOwned, Deserialize, Serialize};
...
+    relay: OutboundRelay<<Service<RuntimeServiceId> as ServiceData>::Message>,
+    _phantom: PhantomData<(Tx, BlobCert)>,
 }
 
+type Service<RuntimeServiceId> = BlendProxyService<
+    proxy::core::libp2p::Libp2pAdapter<RuntimeServiceId>,
+    proxy::edge::libp2p::Libp2pAdapter<RuntimeServiceId>,
+    RuntimeServiceId,
+>;
+type BroadcastSettings<RuntimeServiceId> =
```
ntn-x2, line 32: I think we are leaking a bit too much information here. Can we do better so to hide everything behind the proxy service itself?

nomos-services/chain-service/src/blend/mod.rs
```diff
@@ -1,23 +1,24 @@
...
 use nomos_core::block::Block;
 use overwatch::services::{relay::OutboundRelay, ServiceData};
 use serde::{de::DeserializeOwned, Serialize};
 
 #[async_trait::async_trait]
 pub trait BlendAdapter<RuntimeServiceId> {
     type Settings: Clone + 'static;
-    type Backend: BlendBackend<Self::NodeId, RuntimeServiceId> + 'static;
-    type Network: NetworkAdapter<RuntimeServiceId> + 'static;
+    type CoreAdapter: proxy::core::Adapter<RuntimeServiceId>;
```
ntn-x2, line 12: Can we pull these traits directly from inside the proxy service? Instead of requiring all these core and edge stuff in all services that depend on it? I think that was one of my hopes with this new service.

nomos-services/blend/src/proxy/mod.rs
ntn-x2: Also, do you think it would make sense to pull the definitions from the services directly based on the runtime information, rather than delegate those to a trait? All depending services would now need to "inject" adapters for both core and edge modes. Do you think we could abstract over that and use internally whatever was used to instantiate those two services in the node itself? In this way depending services would not need to know about specific backends or anything like that, but I am not sure that would be possible with the current structure.

nomos-services/blend/src/proxy/mod.rs
danielSanchezQ: I think it would be nicer too.

nomos-services/blend/src/proxy/mod.rs
danielSanchezQ: @ntn-x2 you mean getting the types info somehow from the `RuntimeServiceId` type?

danielSanchezQ (commented):
Looking good, I was just wondering why extra services instead of layer them down a single one. Is that so we can turn off the core one if we the node does not participate actively in blend?

ntn-x2:
@danielSanchezQ this was discussed in our Discord. We had to choice: either keep a single service but somehow two different libp2p swarms, or have two separate services. Core vs edge behaviour are very different from each other, and cramming everything into a single service would have been a recipe for disaster, in my opinion. So we opted for having two different services, and we now need a third service that can expose a stable API for all other services.

nomos-services/blend/src/proxy/mod.rs
ntn-x2: Yes, I did not spend too much time thinking about this yet, but having to define an adapter for each proxies state partially defeats the purpose.

danielSanchezQ: Yes I think so. We can make this one service to define the adapter types directly. As it is supossed to be the single consumer of the inner ones (?)

nomos-services/blend/src/proxy/edge/mod.rs
youngjoon-lee: I had the same feeling. The issue is that `BroadcastSettings` is the one which is set outside of the blend (core) service. It is set by the chain service because it knows the `topic` that it wants to subscribe from the gossip network. We might be able to refactor this, though.

ntn-x2: Yes, if I understand correctly it would be enough to separate the service message that the core and edge services accept, since we know for sure edge services do not deal with gossipsub broadcasting.

PR #1475, opened 2025-07-18
feat(blend): add BlendEdgeService

youngjoon-lee:
## 1. What does this PR implement?

(Based on PR #1473)

Closes #1463 

Adds a new service for Blend edge mode, which uses ~~a separate `nomos_blend_network::edge::Behaviour` implemented by #1473.~~ `libp2p_stream::Behaviour` added by https://github.com/logos-co/nomos/pull/1493.

This PR doesn't integrate this service into the system.
As mentioned in a next issue https://github.com/logos-co/nomos/issues/1464, we are going to introduce another new service, which acts as a proxy between core and edge services.
```
chain-service -------> blend-proxy-service -------> core-blend-service
                                           \
                                            ------> edge-blend-service
```

Also, in a next PR, I will reorganize the module structure for blend services (under `nomos-services/blend/`), similar as the PR https://github.com/logos-co/nomos/pull/1468 (which refactors `nomos-blend/`).

## 2. Does the code have enough context to be clearly understood?

Yes

## 3. Who are the specification authors and who is accountable for this PR?

@madxor @ntn-x2 @youngjoon-lee 

## 4. Is the specification accurate and complete?

Yes

## 5. Does the implementation introduce changes in the specification?

No

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [x] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [x] 2. Description added.
* [x] 3. Context and links to Specification document(s) added.
* [x] 4. Main contact(s) (developers and specification authors) added
* [x] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 6. Link PR to a specific milestone.

PR #1493, opened 2025-07-24
feat(blend): replace `edge::Behaviour` with `libp2p_stream::Behaviour` 

youngjoon-lee:

(The base branch is the PR https://github.com/logos-co/nomos/pull/1475)
 
In the PR https://github.com/logos-co/nomos/pull/1473, we've discussed a lot about:
- How to handle connections initiated outside of `edge::Behaviour`.
- How to assign messages to connections.
- How to deal with the life cycle of connections.

But, I realized that we don't actually need to define our own Behaviour and ConnectionHandler for edge nodes.
The logic of edge nodes is simple.
- Establishing a connection
- Opening a substream
- Sending a message
- Dropping the substream

That's it.
These functionalities are already provided by https://crates.io/crates/libp2p-stream.
I think we won't add any additional functionality for edge nodes in the future. Even if we do, such features will likely be implemented in the `EdgeService` layer, and there would be no need to design a new `Behaviour` or `ConnectionHandler`. If it ever becomes truly necessary, we can design them at that point.

So, this PR does:
- Replace `edge::Behaviour` with `libp2p_stream::Behaviour`.
  - This is a key change. Please see the `nomo-services/blend/edge` first.
- Remove `edge::ConnectionHandler`.
  - Also removed `nomos-blend/network/src/tests` which is not necessary anymore. Instead, moved one of those tests in `nomos-blend/network/src/core`.

If this approach looks good, we can close the PR #1473 (which implements `edge::Behaviour`).

Yes

@ntn-x2 @youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/network/Cargo.toml
+++ b/nomos-blend/network/Cargo.toml
@@ -15,15 +15,15 @@ futures-timer          = { version = "3.0.3" }
 libp2p                 = { workspace = true }
 nomos-blend-message    = { workspace = true }
 nomos-blend-scheduling = { workspace = true }
-nomos-utils            = { optional = true, workspace = true }
+nomos-utils            = { workspace = true, optional = true }
 sha2                   = "0.10"
-tokio                  = { features = ["time"], optional = true, version = "1" }
-tokio-stream           = { optional = true, version = "0.1" }
+tokio                  = { version = "1", features = ["time"], optional = true }
+tokio-stream           = { version = "0.1", optional = true }
 tracing                = { workspace = true }
 
 [dev-dependencies]
-libp2p       = { features = ["ed25519", "quic", "tokio"], workspace = true }
-tokio        = { features = ["macros", "rt-multi-thread", "time"], version = "1" }
+libp2p       = { workspace = true, features = ["ed25519", "quic", "tokio"] }
+tokio        = { version = "1", features = ["macros", "rt-multi-thread", "time"] }
 tokio-stream = "0.1"
 
 [features]


nomos-blend/network/src/core/behaviour.rs
```diff
@@ -73,6 +76,8 @@ pub enum Event {
     UnhealthyPeer(PeerId),
     /// A peer has been detected as healthy.
     HealthyPeer(PeerId),
+    /// Failed to receive a message from an edge node.
+    EdgeNodeFailedReception(FailureReason),
```
ntn-x2: I don't think we really need this, do we? Like, there's not really anything we can do about it anyway. It's more useful for the sender to know about this.

nomos-services/blend/src/edge/backends/libp2p/swarm.rs
```diff
@@ -16,9 +20,14 @@ pub(super) struct BlendEdgeSwarm<SessionStream, Rng>
...
 {
-    swarm: Swarm<Behaviour<Rng>>,
+    swarm: Swarm<libp2p_stream::Behaviour>,
+    stream_control: libp2p_stream::Control,
     command_receiver: mpsc::Receiver<Command>,
     session_stream: SessionStream,
+    current_membership: Option<Membership<PeerId>>,
+    rng: Rng,
+    pending_dials: HashMap<(PeerId, ConnectionId), Vec<u8>>,
+    _phantom: PhantomData<Rng>,
```
ntn-x2: No need for this.

nomos-services/blend/src/edge/backends/libp2p/swarm.rs
```diff
@@ -33,58 +42,159 @@ where
...
+        }
+    }
+
+    fn handle_send_message_command(&mut self, msg: Vec<u8>) {
+        self.dial_and_schedule_message(msg);
+    }
+
+    fn dial_and_schedule_message(&mut self, msg: Vec<u8>) {
+        let Some(peer) = self.choose_peer() else {
+            error!("No peers available to send the message to");
```
ntn-x2:
```suggestion
            error!(target: LOG_TARGET, "No peers available to send the message to");
```

nomos-services/blend/src/edge/backends/libp2p/swarm.rs
```diff
@@ -33,58 +42,159 @@ where
...
         }
     }
 
-    fn handle_send_message_command(&mut self, msg: Vec<u8>) {
-        if let Err(e) = self.swarm.behaviour_mut().send_message(msg) {
-            tracing::error!(target: LOG_TARGET, "Failed to schedule sending message: {e}");
+    async fn send_message_to_stream(message: Vec<u8>, stream: libp2p::Stream) {
+        match send_msg(stream, message).await {
+            Ok(_) => {
+                debug!(target: LOG_TARGET, "Message sent successfully");
```
ntn-x2, line 164: I guess we need to close the stream after sending the message to avoid waiting for the core node's timeout to elapse?

ntn-x2 (commented):
Thanks for coming up with this very simple solution! I was not aware of this generic stream protocol, but it looks exactly what we need. I think we just need to explicitly close the stream after sending the message without waiting for the timeout to elapse, then we're good!

nomos-blend/network/src/core/behaviour.rs
youngjoon-lee: > Like, there's not really anything we can do about it anyway.

Right.

I added this for testing. but let me check if I can simplify the test without this.

nomos-services/blend/src/edge/backends/libp2p/swarm.rs
youngjoon-lee: Thanks. Resolved ea19b2c8

nomos-services/blend/src/edge/backends/libp2p/swarm.rs
youngjoon-lee: Thanks. Resolved ea19b2c8

nomos-services/blend/src/edge/backends/libp2p/swarm.rs
youngjoon-lee: Thanks. Resolved ea19b2c8

nomos-blend/network/src/core/behaviour.rs
youngjoon-lee: 2c71f6cb I removed the test and the event. Later, it's better to add tests in the connection handler level.

Issue #1486, opened 2025-07-22
Validate decapsulated header


ntn-x2:
As per Blend v1 spec: https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81fa8369dc90a4daedd6.

Issue #1467, opened 2025-07-17
Validate message public header before relaying, and update conn maintenance accordingly


youngjoon-lee:
The [spec](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81d691acebe6bfede0fc) defines that a node who relays a message with an invalid format must be marked as malicious, and the connection with that node must be closed.

The current implementation checks only the duplication (by calculating a SHA-256 message ID). However, we should toss messages to the `CryptographicProcessor` ~~to try to decapsulate it and~~ to [check if its public header is valid](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df8156bb69c89f3140164c). After all checks are done, the message can be finally relayed to the peers.

In addition, we should update the message ID calculation. Instead of SHA-256, we should use the <[signing_pubkey](https://github.com/logos-co/nomos/blob/d52ced63504559108cc0d6f2f51a2ee109677edf/nomos-blend/message/src/message.rs#L39), nullifier>, as defined in the [spec](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df813d9aabfe1e99f9e94d). Since a nullifier can be obtained from a Proof of Quota, which is not implemented yet, we can replace it with <[signing_pubkey](https://github.com/logos-co/nomos/blob/d52ced63504559108cc0d6f2f51a2ee109677edf/nomos-blend/message/src/message.rs#L39), [proof_of_quota](https://github.com/logos-co/nomos/blob/d52ced63504559108cc0d6f2f51a2ee109677edf/nomos-blend/message/src/message.rs#L40)> for now. (Currently, the `proof_of_quota` is a dummy.)

With these changes, we can address the TODO comment for updating the `seen_message_cache`.
https://github.com/logos-co/nomos/blob/d52ced63504559108cc0d6f2f51a2ee109677edf/nomos-blend/network/src/behaviour.rs#L43-L44
This cache can be used for implementing the [connection maintenance for peers who send duplicate messages](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81fc86bdce264466efd3) (which is not implemented in the `master` yet).

PR #1494, opened 2025-07-24
feat(blend): update cache keys and classify peers for malformed and duplicated messages

ntn-x2:

This PR fixes two tickets.

It fixes https://github.com/logos-co/nomos/issues/1467 by updating the cache to store a set of seen identifiers per peer, so that we consider them malicious if they send two messages with the same identifier, and that we avoid sending them duplicate messages and be considered malicious by them. I also added a trait only used internally to check on the public header, and that will include in the future a method to also validate the proof of quota. This enables us to use a `Validated` variant of an encapsulated message in the `forward` method so that we cannot forget to validate the public header of a message before sending it out, be it a message we send before or after decapsulation.

This PR also fixes https://github.com/logos-co/nomos/issues/1486 by using the mechanism above: once we are able to decapsulate a message, we propagate it to the Blend service for scheduling. When the time comes and the scheduler instructs the swarm to propagate the new messages, the public header of the now decapsulated messages is checked, as well as the unicity of the message. Messages that contain an invalid decapsulated header or a duplicate identifier will not be forwarded to the rest of the network. In the future, we can also run additional checks in the service itself and avoid adding invalid messages to the scheduler in the first place, although we will anyway keep running public header checks in the swarm to avoid sending out invalid messages which can have catastrophic effects for the Blend node.

### TL;DR of the changes

* A few convenience methods were added to the cryptographic processor module to allow other places in the code to perform substeps at different times instead of once, e.g., deserializing a message before trying to decapsulate it, or verifying its public header before decapsulating it.
* The Blend behaviour has been updated to have a cryptographic processor that decapsulates messages, and an updated cache which tracks message identifiers exchanged (i.e., sent and received) with peers. The new cache is not bound in size, and will be cleaned after a session and its transition period have elapsed. When the connection with a peer is closed, the entry in the cache is deleted as well.
  * Also, the behaviour now checks the public header of encapsulated messages it receives from the service, before they are propagated to the rest of the Blend network
  * It also implements logic to mark peers as malicious, according to the Blend v1 spec.
* The tokio tests for the behaviour were temporarily removed as they were not trivial to adapt due to the new requirements introduced by having a cryptographic processor inside the behaviour. They will be re-introduced in a PR based on top of this one.
* The adapter trait between the service and the libp2p stack now handles structured data instead of `Vec<u8>`s: `EncapsulatedMessage`s to blend to the rest of the network, and `UnwrappedMessage`s (i.e., partially or completely decapsulated messages) received by Blend peers

Yes.

Spec: @madxor, PR: @ntn-x2 & @youngjoon-lee.

Few changes were made along the way, but now spec and implementation are in sync.

Few changes were made along the way, but now spec and implementation are in sync.

## Checklist

* [x] 2. Description added.


nomos-blend/scheduling/src/lib.rs
```diff
@@ -1,117 +1,12 @@
...
 mod release_delayer;
-
-pub use self::message_scheduler::UninitializedMessageScheduler;
-
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
-pub enum BlendOutgoingMessage {
-    CoverMessage(CoverMessage),
-    DataMessage(DataMessage),
-    EncapsulatedMessage(EncapsulatedMessage),
-}
```
youngjoon-lee, line 16: Thank you for removing this type.

nomos-blend/network/src/behaviour.rs
```diff
@@ -168,30 +197,144 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
+            };
         }
 
         // Forward the message immediately to the rest of connected peers
-        // without any processing for the fast propagation.
-        if let Err(e) = self.forward_message(&message, from) {
-            tracing::error!("Failed to forward message: {e:?}");
+        // before any further processing for fast propagation.
+        if let Err(e) =
+            self.forward_message(&deserialized_encapsulated_message, source.sender_address())
```
youngjoon-lee: I'm sorry for the confusion caused by my incorrect issue description: https://github.com/logos-co/nomos/issues/1467. I think we need to [validate the public header](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df8156bb69c89f3140164c) to decide whether it's okay to forward the message.

Here comes a confusion.

Before forwarding (relaying), we should validate the **before-decapsulation public header**, if I understand the spec correctly. It's possible because the public header is **public**, which means that it's not encrypted (encapsulated).

I also see another public header validation below [after decapsulation](https://github.com/logos-co/nomos/blob/0d0929cc92b9c40fd29d64ad342c6c83998462fd/nomos-blend/network/src/behaviour.rs#L307-L317). That is also defined in the spec, as you mentioned in the [issue](https://github.com/logos-co/nomos/issues/1486). But, even if the after-decapsulation public header is invalid, the peer shouldn't be marked as malicious, because the peer was not able to decapsulate the message and check whether the public header is invalid. We just need to ignore the message without marking the peer as malicious.

Let's go back to the before-decapsulation public header validation.
The spec defines 4 steps to validate the public header. 
image

The `a` and `d` are just the duplication check that we're doing in this PR. 
The `b` (signature validation) could be done. I can help if it's not clear in the current `nomos-blend-message` implementation. Or, we can leave this as a placeholder, since it might be out of scope of this PR.
The `c` (PoQ validation) could be also a placeholder. Yet I don't clearly understand how to verify the PoQ and what's needed for that.

Or, we can simply add a TODO comment, so that we can address it later soon.

nomos-blend/network/src/behaviour.rs
```diff
@@ -168,30 +197,144 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
+                malicious_error @ (
+                // Mark peer as malicious if it sends a message with either:
+                // * an invalid signature: https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81859cebf5e3d2a5cd8f
+                // * an invalid proof of quota: https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81b593ddce00cffd24a8
+                | nomos_blend_message::Error::SignatureVerificationFailed
+                | nomos_blend_message::Error::ProofOfQuotaVerificationFailed),
+            ) => {
+                tracing::debug!(target: LOG_TARGET, "Failed to verify proofs on message from peer source {source:?}. Error: {malicious_error:?}");
+                self.mark_peer_as_malicious(source);
+                Err(())
```
youngjoon-lee: If my understanding is correct, we shouldn't mark this peer as malicious even if this **after-decapsulation public header validation** fails. The reason is that the peer wasn't able to decapsulate the message and check the public header. I left more details about this in another thread above. We can continue this discussion there.

nomos-blend/network/src/behaviour.rs
```diff
@@ -168,30 +197,144 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
-impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider>
+// TODO: Replace this with the proof of quota nullifier key proof of quota
+// verification is implemented.
+const fn message_id(message: &EncapsulatedMessage) -> MessageIdentifier {
+    let message_public_header = message.public_header();
+    (
+        message_public_header.signing_pubkey,
+        message_public_header.proof_of_quota,
+    )
+}
```
youngjoon-lee: nit: we can probably move this method into `nomos-blend-message::EncapsulatedMessage`.

ntn-x2: Did that in https://github.com/logos-co/nomos/pull/1494/commits/76d32dfbe932a133f49835e5d18f112fe04a9c7f.

PR #1496, opened 2025-07-25
chore(lints): enforce low-pain rustc and Clippy lints

ntn-x2:

Enforce the last lints that make sense to enforce AND are low pain since they do not require big changes. Next, we will start enabling one lint at the time, most likely.

### Clippy lints

* [clippy::cfg_not_test](https://rust-lang.github.io/rust-clippy/master/index.html#/cfg_not_test): we had one instance. In that case, the tracing crate would anyway panic if called more than once, so it does not hurt to run the initialization function once only, also outside of tests.
* [clippy::dbg_macro](https://rust-lang.github.io/rust-clippy/master/index.html#/dbg_macro): we have debug logs instead. In the past, one of these slipped in the Overwatch crate and was polluting all the logs because there is no way of filtering them. `debug_assert`s can still be used. No affected code.

### Rustc lints

* [explicit_outlives_requirements](https://doc.rust-lang.org/beta/rustc/lints/listing/allowed-by-default.html#explicit-outlives-requirements): less code to write, the better. No affected code.
* [if_let_rescope](https://doc.rust-lang.org/beta/rustc/lints/listing/allowed-by-default.html#if-let-rescope): the explanation is quite exhaustive. No affected code.
* [keyword-idents](https://doc.rust-lang.org/beta/rustc/lints/listing/allowed-by-default.html#keyword-idents), [keyword-idents-2018](https://doc.rust-lang.org/beta/rustc/lints/listing/allowed-by-default.html#keyword-idents-2018), [keyword-idents-2024](https://doc.rust-lang.org/beta/rustc/lints/listing/allowed-by-default.html#keyword-idents-2024): same as previous point. No affected code.
* [meta_variable_misuse](https://doc.rust-lang.org/beta/rustc/lints/listing/allowed-by-default.html#meta-variable-misuse): less error-prone macros. No affected code.
* [non_ascii_idents](https://doc.rust-lang.org/beta/rustc/lints/listing/allowed-by-default.html#non-ascii-idents): No affected code.
* [unnameable_types](https://doc.rust-lang.org/beta/rustc/lints/listing/allowed-by-default.html#unnameable-types): exhaustive explanation in the lint description. No affected code.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-services/tracing/src/lib.rs
+++ b/nomos-services/tracing/src/lib.rs
@@ -165,9 +165,8 @@ where
         service_resources_handle: OpaqueServiceResourcesHandle<Self, RuntimeServiceId>,
         _initial_state: Self::State,
     ) -> Result<Self, overwatch::DynError> {
-        #[cfg(test)]
         use std::sync::Once;
-        #[cfg(test)]
+
         static ONCE_INIT: Once = Once::new();
 
         let config = service_resources_handle
@@ -237,18 +236,12 @@ where
             });
         }
 
-        #[cfg(test)]
         ONCE_INIT.call_once(move || {
             tracing_subscriber::registry()
                 .with(LevelFilter::from(config.level))
                 .with(layers)
                 .init();
         });
-        #[cfg(not(test))]
-        tracing_subscriber::registry()
-            .with(LevelFilter::from(config.level))
-            .with(layers)
-            .init();
 
         panic::set_hook(Box::new(nomos_tracing::panic::panic_hook));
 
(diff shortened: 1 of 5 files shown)


nomos-services/tracing/src/lib.rs
```diff
@@ -237,18 +236,12 @@ where
             });
         }
 
-        #[cfg(test)]
         ONCE_INIT.call_once(move || {
```
strinnityk, line 239: While I understand having it this way, I don't understand why we had them under macros in the first place. Is there some difference?

bacv (approved):

pradovic (approved):

nomos-services/tracing/src/lib.rs
ntn-x2: The way it was, we tried to call it once overall when running tests, while we did not enforce that otherwise. But, the tracing crate can anyway only be initialized once by calling `init` (which would panic when called more than once), else we would need to use `try_init` and ignore the initialisation error. I opted for keeping things the same way, and simply unify behaviour between test and not test environment, since what we want in any case, is to initialise the tracing stuff only once.

Issue #1456, opened 2025-07-16
IBD: Abort downloading early on block processing errors


youngjoon-lee:
https://github.com/logos-co/nomos/pull/1442# implements the basic IBD, but doesn't stop downloading even if the block processing is failed. 
Once a block is failed to be processed, we don't need to download subsequent blocks from the peer.

This will require refactoring of `BlockProcessor` (added in https://github.com/logos-co/nomos/pull/1431) since it is not returning any error.

PR #1497, opened 2025-07-25
feat(chain): orphan blocks handling in chain service

andrussal:

This PR adds orphan blocks handling to the chain service according to the requirement in the _Synchronization_ [_spec_](https://nomos-tech.notion.site/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6)

> If the parent of the block is missing from the local block tree, the block cannot be fully validated and added. These blocks are called orphan blocks. To handle an orphan block, the node downloads missing blocks from a randomly selected peer

- It introduces `OrphanBlocksDownloader` component that tries to download orphan block parents from peers.
- Received blocks are received by chain service and applied to local chain

Node may receive multiple orphan blocks from the same fork. If a block is received by regular gossip or part of a orphan stream, downloader pre-emptively removes pending orphans from its cache to avoid downloading the same blocks multiple times. 

Because we can only apply blocks since LIB, all streams are relatively short and should be processed fast. Also normally we shouldn't have multiple orphans at the same time. Plus we may receive the same orphans of the same fork and we want to avoid duplicate downloads. All these suggest that sequential downloading for orphan blocks seems good approach.

Yes

@andrussal 

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-services/chain-service/src/lib.rs
+++ b/nomos-services/chain-service/src/lib.rs
@@ -59,10 +59,6 @@ use tokio::{
 use tracing::{debug, error, info, instrument, span, Level};
 use tracing_futures::Instrument as _;
 
-pub use crate::{
-    bootstrap::config::{BootstrapConfig, IbdConfig},
-    sync::config::{OrphanConfig, SyncConfig},
-};
 use crate::{
     blend::BlendAdapter,
     bootstrap::{ibd::InitialBlockDownload, state::choose_engine_state},
@@ -71,6 +67,10 @@ use crate::{
     states::CryptarchiaConsensusState,
     storage::{adapters::StorageAdapter, StorageAdapter as _},
     sync::{block_provider::BlockProvider, orphan_handler::OrphanBlocksDownloader},
+};
+pub use crate::{
+    bootstrap::config::{BootstrapConfig, IbdConfig},
+    sync::config::{OrphanConfig, SyncConfig},
 };
 
 type MempoolRelay<Payload, Item, Key> = OutboundRelay<MempoolMsg<HeaderId, Payload, Item, Key>>;


nodes/nomos-executor/config.yaml
```diff
@@ -307,6 +307,8 @@ cryptarchia:
     force_bootstrap: false
     ibd:
       peers: []
+  orphan:
+    max_orphan_cache_size: 5
```
youngjoon-lee: nit: I think we can consider moving this config under a new field 
```suggestion
  sync:
    orphan:
      max_orphan_cache_size: 5
```
I think this change is reasonable because we have another parameters that could be included under the `sync` in the future.
https://github.com/logos-co/nomos/blob/e4461c6d1639304a3f925d2518fc8611141ccadd/nomos-services/chain-service/src/sync/block_provider.rs#L22

nomos-services/chain-service/src/lib.rs
```diff
@@ -682,6 +701,10 @@ where
                         state_updater,
                     )
                     .await
+                    .unwrap_or_else(|e| {
+                        error!("Error processing block during IBD: {:?}", e);
+                        (cryptarchia_clone, storage_blocks_to_remove_clone)
+                    })
```
youngjoon-lee, line 705: Just FYI:

Yesss. Actually, I opened an issue https://github.com/logos-co/nomos/issues/1456 which is similar as this change. I see that this change (which makes the `process_block`return an error) is necessary for this PR to cancel the orphan downloader in some cases.

I think we can go with your changes, which look good. But, let's keep the issue https://github.com/logos-co/nomos/issues/1456 open. We probably should categorize errors more in the `process_block` method, and should handle those errors differently in IBD. These are tech debts that have accumulated so far.
Therefore, We can do it as a next PR, which may involve some refactoring https://github.com/logos-co/nomos/issues/1505. I'd like to try it since I have some idea.

Anyway, these changes in this PR are looking good.

nomos-services/chain-service/src/lib.rs
```diff
@@ -791,6 +845,36 @@ where
                            Self::handle_chainsync_event(&cryptarchia, &sync_blocks_provider, event).await;
                         }
                     }
+
+                    Some(block) = orphan_downloader.next(), if orphan_downloader.should_poll() => {
+                        let header_id  =  block.header().id();
```
youngjoon-lee:
```suggestion
                        let header_id = block.header().id();
```

nomos-services/chain-service/src/sync/orphan_handler.rs
```diff
@@ -0,0 +1,846 @@
...
+}
+
+/// Fetches blocks for orphan blocks by requesting their ancestors from the
+/// network.
+pub struct OrphanBlocksDownloader<NetAdapter, RuntimeServiceId>
+where
+    NetAdapter: NetworkAdapter<RuntimeServiceId>,
+{
+    /// Queue of blocks waiting to be fetched
+    pending_orphans_queue: BTreeMap<HeaderId, OrphanInfo>,
```
youngjoon-lee: Could we probably just use a `HashMap` instead of `BTreeMap` because it might not be so meaningful to sort `HeaderId`s which are byte arrays?

nomos-services/chain-service/src/sync/orphan_handler.rs
```diff
@@ -0,0 +1,846 @@
...
+    lib: HeaderId,
+}
+
+impl OrphanInfo {
+    const fn new(block_id: HeaderId, local_tip: HeaderId, immutable_tip: HeaderId) -> Self {
+        Self {
+            orphan_id: block_id,
+            tip: local_tip,
+            lib: immutable_tip,
+        }
```
youngjoon-lee:
```suggestion
    const fn new(block_id: HeaderId, local_tip: HeaderId, lib: HeaderId) -> Self {
        Self {
            orphan_id: block_id,
            tip: local_tip,
            lib,
        }
```
I think we can use the term `lib` which is official. The `immutable_tip` is an incorrect term because the LIB may not be the tip in most cases.

nomos-services/chain-service/src/sync/orphan_handler.rs
```diff
@@ -0,0 +1,846 @@
...
+        Self {
+            pending_orphans_queue: BTreeMap::new(),
+            network_adapter,
+            state: DownloaderState::Idle,
+            max_pending_orphans,
+            max_blocks_per_orphan: security_param,
+            waker: None,
+            _phantom: std::marker::PhantomData,
+        }
+    }
```
youngjoon-lee: I have two comments regarding the `security_param` and the `max_blocks_per_orphan`.

1. It's better to rename the argument `security_param` to `max_blocks_per_orphan`, because the term `security_param` is too `cryptarchia-engine`-specific. In other words, it's not so safe to use the `security_param` directly for some purposes in the `OrphanBlocksDownloader` because the exact ways how to use the `security_params` are abstracted in the `cryptarchia-engine`. Let me explain the reasons in the `2` below.
```suggestion
    pub const fn new(
        network_adapter: NetAdapter,
        max_pending_orphans: NonZeroUsize,
        max_blocks_per_orphan: NonZero<u32>,
    ) -> Self {
        Self {
            pending_orphans_queue: BTreeMap::new(),
            network_adapter,
            state: DownloaderState::Idle,
            max_pending_orphans,
            max_blocks_per_orphan,
            waker: None,
            _phantom: std::marker::PhantomData,
        }
    }
```

2. I think we don't really need to have this `max_blocks_per_orphan` parameter. I see that this parameter is used for stopping a download at some point. But, it is not actually correct if I understand the logic correctly.
  - One good example to explain the reasons is that the LIB could be veryyyy old (e.g. the genesis block) if the Cryptarchia is still in the Bootstrapping mode. Please keep in mind that the default Prolonged Bootstrapping Period is 24h. Nevertheless, the orphan blocks should be (theoretically) downloaded without giving up in the middle. 
  - In other words, the fact that the $$k$$-th block is the LIB is abstracted in the `cryptarchia-engine`. It is not safe to assume that the $$k$$-th block is always the LIB. It depends on the current **mode**. Also, the details can change in the future in the `cryptarchia-engine`. We never know that in this layer.

https://github.com/logos-co/nomos/blob/e4461c6d1639304a3f925d2518fc8611141ccadd/nomos-services/chain-service/src/sync/orphan_handler.rs#L261-L271

With these reasons, I want to suggest the following behavior:
- Remove the `max_blocks_per_orphan` parameter.
- `OrphanBlocksDownloader` just reads a stream and returns any block received without checking the `download.total_blocks_received `.
- Then, the chain service will try to apply the block to the chian. If the block is older than LIB, it will be rejected by the `cryptarchia-engine`, because it is a responsibility of the `cryptarchia-engine`. 
- Then, the chain service will get the error and request the `OrphanBlocksDownloader` to cancel the download.

nomos-services/chain-service/src/sync/orphan_handler.rs
```diff
@@ -0,0 +1,846 @@
...
+                            let request_blocks_stream_fut = Self::request_blocks_stream(
+                                self.network_adapter.clone(),
+                                orphan_info,
+                                known_blocks,
+                            );
+
+                            self.state =
+                                DownloaderState::Requesting(Box::pin(request_blocks_stream_fut));
+
+                            cx.waker().wake_by_ref();
```
youngjoon-lee: Can we probably **do not** initiate the next download if `last_block_id == orphan_info.orphan_id`? (although the peer won't respond with more than 1 block)

nomos-services/chain-service/src/sync/orphan_handler.rs
youngjoon-lee: @andrussal Is this ready for review again?

andrussal: > @andrussal Is this ready for review again?

Yes please take a look again, I didn't have anything to discuss further, all suggestions seemed good.

nomos-services/chain-service/src/sync/orphan_handler.rs
youngjoon-lee: @andrussal I approved this PR, but I think this comment hasn't been resolved yet. If you think this suggestion is not relevant, or if I'm missing anything, please reply here.

andrussal: > @andrussal I approved this PR, but I think this comment hasn't been resolved yet. If you think this suggestion is not relevant, or if I'm missing anything, please reply here.

Thank you @youngjoon-lee 

I added this check to slightly different location where we receive blocks https://github.com/logos-co/nomos/blob/3850078d7cbf0a7499b9b28161a55bbd38632ebb/nomos-services/chain-service/src/sync/orphan_handler.rs#L253 But maybe you were considering situation when peer sends us also blocks after orphan block and we have no reason to stop downloading if we can apply them to local chain. If you think it's better to be more relaxed here, I can move it to your suggested location.

youngjoon-lee: Aha, I see that your fix resolves the case that I suggested. Thanks.

> But maybe you were considering situation when peer sends us also blocks after orphan block and we have no reason to stop downloading if we can apply them to local chain.

No, I don't think we need to do so. 

Thank you. I think we can merge this PR.

Issue #1484, opened 2025-07-22
Enforce peering degrees for core->edge and core->core connections


ntn-x2:
From the spec: steps [6](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81b8bc75cde60ce4e4d3), and [7](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df8152a2f5f5b3ddc97aa6).

PR #1503, opened 2025-07-29
chore(blend): separate network behaviour for core->core and core->edge functionalities

ntn-x2:

This PR lays the ground for https://github.com/logos-co/nomos/issues/1484, by splitting the existing core network behaviour into two separate ones, one for core->core and one for core->edge functionalities. This will allow for better separation of concerns as the two protocols will evolve independently.

Main recap of changes:

* Refactored the folder structure so that both behaviours have a very similar structure.
* Remove anything that was dealing with edge peers from the core->core network behaviour
* Since the core->edge behaviour does not have the list of connected peers and it does not forward a received message automatically to all other peers, such behaviour has been removed also from the core->core network behaviour. Now the swarm, upon receiving a message from either behaviour, will propagate it to all (if from an edge peer) or all but the sender (if from a core peer) before bubbling it up to the Blend service for further processing
* Also, as we have already discussed previously, instead of denying a connection from another peer, which might be too harsh at times, we use in both network behaviour either our own connection handler if we want a substream to be opened, or a `DummySubstream` which will not allow the stream to be upgraded at all, so that the same peer is able to open other substreams if other network behaviours allow it.
* The core->edge connection handler generates two additional events to signal that a sub-stream has been opened. We use that to populate the map of connected edge peers in the core->edge network behaviour. Although we would allow for multiple sub-streams, we still map a single connection to a single peer, so no more than the required peers can establish inbound connections to a core node (to be tackled in a follow-up PR)

I believe so.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/network/src/core/with_edge/behaviour/mod.rs
+++ b/nomos-blend/network/src/core/with_edge/behaviour/mod.rs
@@ -149,6 +149,7 @@ impl NetworkBehaviour for Behaviour {
         match event {
             Either::Left(ToBehaviour::Message(message)) => {
                 self.handle_received_serialized_encapsulated_message(&message);
+                self.try_wake();
             }
             Either::Left(ToBehaviour::SubstreamOpened) => {
                 self.connected_edge_peers
@@ -168,7 +169,6 @@ impl NetworkBehaviour for Behaviour {
                 }
             }
         }
-        self.try_wake();
     }
 
     fn poll(
--- a/nomos-blend/network/src/core/with_core/behaviour/mod.rs
+++ b/nomos-blend/network/src/core/with_core/behaviour/mod.rs
@@ -109,20 +109,27 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
     ///
     /// Before the message is propagated, its public header is validated to make
     /// sure the receiving peer won't mark us as malicious.
-    pub fn validate_and_publish(&mut self, message: EncapsulatedMessage) -> Result<(), Error> {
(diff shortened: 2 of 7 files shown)


nomos-blend/network/src/core/with_edge/behaviour/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+#[derive(Debug)]
+pub struct Config {
+    pub connection_timeout: Duration,
+}
+
+/// A [`NetworkBehaviour`]:
+/// - receives messages from edge nodes and forwards them to the swarm.
+pub struct Behaviour {
+    /// Queue of events to yield to the swarm.
+    events: VecDeque<ToSwarm<Event, Either<Infallible, Infallible>>>,
```
danielSanchezQ, line 46: Curious, why not infallible instead of the either wrapping?

danielSanchezQ (approved):
Nice stuff 💯

nomos-blend/network/src/core/with_edge/behaviour/mod.rs
ntn-x2: Because the `poll` method of the behaviour must return `Poll<ToSwarm<Self::ToSwarm, THandlerInEvent<Self>>>`, where `type THandlerInEvent<TBehaviour> = <THandler<TBehaviour> as ConnectionHandler>::FromBehaviour;`. So basically we need to return whatever event type the connection handler defines, and in our case the `Either` connection handler defines a `Either` event with either variant inside, and each variant is an `Infallible`.

Issue #1505, opened 2025-07-30
Refactor chain-service block processing


youngjoon-lee:
The current block processing is implemented in as an associate method of the chain-service.
Because of that, a lot of features have been also implemented as associate methods of the chain-service.
- Recovery
- IBD
- Orphan block handling (WIP)
- Proposing blocks

It's not bad but not so convenient for testing and modularization. 

It'd be great if we can modularize the block processing as a trait.

Also, we maintain the `storage_blocks_to_remove` HashSet, which is always passed to the `process_block`. We can probably abstract this into the trait that processes blocks.

PR #1442, opened 2025-07-11
feat(chain): IBD (Initial Block Download)

youngjoon-lee:
## 1. What does this PR implement?

Finally, here is a PR for IBD.

Specs:
- [IBD](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81bd899ff05e97d66375)
- [Block Downloading](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81e8b28fd2bdf3e446b1)

This PR is based on a (virtual) branch `ibd-prep` that merges all preparation PRs. This strategy makes this PR smaller, so that you can focus on the core logic.
```  
master - #1428 --------------- ibd-prep - ibd
       └ #1431 - #1438 ------┘       
       └ #1439 --------------┘
```

This PR also includes plenty of tests.

This PR doesn't implement downloading blocks from **multiple** peers concurrently. Instead, it downloads blocks from only one peer. For your review, please focus on the core logic, which download blocks from a peer repeatedly until the peer's tip is reached (as described in the [spec](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81e8b28fd2bdf3e446b1)). The current download will be implemented in a next PR.

## 2. Does the code have enough context to be clearly understood?

Yes

## 3. Who are the specification authors and who is accountable for this PR?

@youngjoon-lee @zeegomo 

## 4. Is the specification accurate and complete?

Yes

## 5. Does the implementation introduce changes in the specification?

No 

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [x] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [x] 2. Description added.
* [x] 3. Context and links to Specification document(s) added.
* [x] 4. Main contact(s) (developers and specification authors) added
* [ ] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 6. Link PR to a specific milestone.

PR #1478, opened 2025-07-18
feat(sync): sync protocol improvements

andrussal:
## 1. What does this PR implement?

This PR addresses some sync networking protocol leftovers what I discovered during IBD and orphan blocks testing

- Add timeouts to requests
- Close streams properly 

## 2. Does the code have enough context to be clearly understood?

Yes. 

## 3. Who are the specification authors and who is accountable for this PR?

@andrussal 

## 4. Is the specification accurate and complete?

N/A

## 5. Does the implementation introduce changes in the specification?

N/A

## Checklist

* [x] 1. Description added.
* [x] 2. Context and links to Specification document(s) added.
* [x] 3. Main contact(s) (developers and specification authors) added
* [x] 4. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 5. Link PR to a specific milestone.

PR #1506, opened 2025-07-30
feat(chain): IBD

youngjoon-lee:

This implemented the [IBD](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81bd899ff05e97d66375).
Previously, I opened a PR https://github.com/logos-co/nomos/pull/1442 for IBD, but I'm replacing it with this PR.
Also, I had a refactoring PR https://github.com/logos-co/nomos/pull/1431 but closed it, so that we can first focus on the IBD logic. 

@zeegomo I have only one TODO comment, saying we need to refactor the block processing in the chain service. I also opened an issue https://github.com/logos-co/nomos/issues/1505 for that. I have an idea but didn't do it in this PR, since I want to focus on only the IBD logic. Nevertheless, I believe this PR is simple and readable. 

@andrussal Please review this, although you'd already reviewed the previous PR #1442. This PR is much simpler.

@danielSanchezQ @bacv @pradovic 
This PR contains a one-line change in DA code because I had to add a trait bound. (As we discussed, we will remove these trait bounds with substantial refactoring in the future). 

Yes

@youngjoon-lee @zeegomo 

Yes

~~- Add `download_limit` to the [`DownloadBlocksRequest`](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df8157959fd9f30c9f99ea)~~

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/chain-service/src/bootstrap/ibd.rs
+++ b/nomos-services/chain-service/src/bootstrap/ibd.rs
@@ -85,6 +85,8 @@ where
         &self,
         cryptarchia: &Cryptarchia,
     ) -> Downloads<NetAdapter::PeerId, NetAdapter::Block> {
+        // TODO: Don't initiate multiple downloads for the same target.
+        //       https://github.com/logos-co/nomos/issues/1455
         let mut downloads = HashMap::new();
         for peer in &self.config.peers {
             match self.initiate_download(*peer, cryptarchia, None).await {


nomos-services/chain-service/src/bootstrap/config.rs
```diff
@@ -1,13 +1,33 @@
...
+pub struct IbdConfig<NodeId>
+where
+    NodeId: Clone + Eq + Hash,
+{
+    /// Peers to download blocks from.
+    pub peers: HashSet<NodeId>,
+    /// The max number of blocks to download by a single request.
+    // TODO: This is not used anywhere yet. Add this to the download request
+    // by modifying related implemn
+    pub download_limit: NonZeroUsize,
```
andrussal: I wonder if this is actually useful if peers can ignore it(I think not in spec)?

nomos-services/chain-service/src/network/adapters/libp2p.rs
```diff
@@ -203,7 +203,7 @@ where
         local_tip: HeaderId,
         latest_immutable_block: HeaderId,
         additional_blocks: HashSet<HeaderId>,
-    ) -> Result<BoxedStream<Result<Self::Block, DynError>>, DynError> {
+    ) -> Result<BoxedStream<Result<(HeaderId, Self::Block), DynError>>, DynError> {
```
andrussal, line 206: Just curious what is the reason of returning tuple: `(HeaderId, Self::Block)`?

nomos-services/chain-service/src/bootstrap/ibd.rs
```diff
@@ -1,21 +1,678 @@
...
+                cryptarchia.lib(),
+                latest_downloaded_block.map_or_else(HashSet::new, |id| HashSet::from([id])),
+            )
+            .await
+            .map_err(Error::BlockProvider)?;
+
+        Ok(Some(Download::new(peer, iteration, target, stream)))
+    }
+
+    fn should_download(target: &HeaderId, cryptarchia: &Cryptarchia) -> bool {
```
andrussal, line 147: I started to think if malicious peer sends a tip that is not a real block, it made it up. We will try to repeat downloads without ever reaching it. But actually `max_download_iterations` eventually would stop this.

andrussal (approved):

nomos-services/chain-service/src/bootstrap/config.rs
```diff
@@ -1,13 +1,33 @@
...
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct IbdConfig<NodeId>
+where
+    NodeId: Clone + Eq + Hash,
+{
+    /// Peers to download blocks from.
+    pub peers: HashSet<NodeId>,
+    /// The max number of blocks to download by a single request.
+    // TODO: This is not used anywhere yet. Add this to the download request
+    // by modifying related implemn
```
zeegomo: typo

nomos-services/chain-service/src/bootstrap/config.rs
```diff
@@ -1,13 +1,33 @@
...
+    pub download_size: NonZeroUsize,
+    /// The max number of downloads to repeat for a single peer.
+    pub max_download_iterations: NonZeroUsize,
```
zeegomo: is this the number of retries? can we provide a default?

nomos-services/chain-service/src/bootstrap/ibd.rs
```diff
@@ -1,21 +1,678 @@
...
+    RuntimeServiceId,
+> where
+    NodeId: Clone + Eq + Hash,
+    ProcessBlockFn: Fn(Cryptarchia, Block) -> ProcessBlockFut,
+    ProcessBlockFut: Future<Output = Cryptarchia>,
+{
+    config: IbdConfig<NodeId>,
+    network: NetAdapter,
+    process_block: ProcessBlockFn,
+    _phantom: PhantomData<(NodeId, Block, Cryptarchia, RuntimeServiceId)>,
```
zeegomo: aren't `NodeId`, `Block` and `Cryptarchia` already captured by `config` and `process_block`?

nomos-services/chain-service/src/bootstrap/ibd.rs
```diff
@@ -1,21 +1,678 @@
...
+    /// Proceeds downloads by reading/processing blocks from the streams.
+    /// It initiates/proceeds additional downloads if needed.
+    /// It returns the updated [`Cryptarchia`].
+    async fn proceed_downloads(
+        &self,
+        mut downloads: Downloads<NodeId, Block>,
+        mut cryptarchia: Cryptarchia,
+    ) -> Cryptarchia {
+        // Read all streams in a round-robin fashion
+        // to avoid forming a long chain of blocks received only from a single peer.
```
zeegomo, line 161: one slow peer could block all progress?

nomos-services/chain-service/src/bootstrap/ibd.rs
```diff
@@ -1,21 +1,678 @@
...
+        Self {
+            peer,
+            iteration,
+            target,
+            last: None,
+            stream,
+        }
+    }
+
+    async fn next_block(&mut self) -> Result<Option<(HeaderId, Block)>, DynError> {
```
zeegomo, line 243: nit: we could also implement `Stream` on `Download`, but this is perfectly fine

nomos-services/chain-service/src/bootstrap/ibd.rs
```diff
@@ -1,21 +1,678 @@
...
+            }
+        }
+    }
+}
+
+pub trait Cryptarchia {
+    fn tip(&self) -> HeaderId;
+    fn lib(&self) -> HeaderId;
+    fn contain(&self, id: &HeaderId) -> bool;
+}
```
zeegomo: what's the reason for this trait?

nomos-services/chain-service/src/lib.rs
```diff
@@ -649,7 +660,36 @@ where
         .await?;
 
         // Run IBD (Initial Block Download).
-        let mut cryptarchia = InitialBlockDownload::run(cryptarchia, network_adapter).await;
+        // TODO: Currently, we're passing a closure that processes each block.
+        //       It needs to be replaced with a trait, which requires substantial
+        // refactoring.       https://github.com/logos-co/nomos/issues/1505
```
zeegomo, line 665: could this be a function?

nomos-services/chain-service/src/lib.rs
```diff
@@ -1554,3 +1601,43 @@ where
...
+/// A wrapper struct that implements [`bootstrap::ibd::Cryptarchia`],
+/// which is for block processing closure used by IBD.
+//
+// TODO: This needs to be replaced with a trait that implements
+//       block processing methods.
+//       https://github.com/logos-co/nomos/issues/1505
+struct IbdCryptarchia {
+    cryptarchia: Cryptarchia,
+    storage_blocks_to_remove: HashSet<HeaderId>,
+}
```
zeegomo: not sure why you need the trait and this impl

nomos-services/chain-service/src/bootstrap/config.rs
youngjoon-lee: Yeah, this is not in the spec. But I think we need to consider adding this to the spec, because it's hard to set `max_download_iterations` (that has been discussed in https://github.com/logos-co/nomos/pull/1442#discussion_r2221942105) if we don't know how many blocks will be returned from a single request. Each responder has its own limit, but the requester doesn't know it (especially when he set the parameter at start-up). 
With this parameter, the requester can estimate the total number of blocks he will download before giving up (`download_limit * max_download_iterations`). Of course, as you said, the responder can ignore the `download_limit` even if we include it in the request. The requester has nothing to do against it. 

But we can have further discussions in Discord. I don't think this solution is optimal.

nomos-services/chain-service/src/network/adapters/libp2p.rs
youngjoon-lee: It's because the IBD module needs to access the block ID but the `Block` is abstracted as an associate type of `NetworkAdapter`. In my previous PR https://github.com/logos-co/nomos/pull/1442, I introduced the `BlockTrait` but I think it's simpler to modify the `NetworkAdapter` methods.
Soon, in the refactoring PRs, I will try to devise the optimal way, not only for the block abstraction, but also for restructuring the chain service. Until then, I don't want to introduce any temporary block abstraction.
I think this modification is not so optimal but okay for now, though I'm open to change.

nomos-services/chain-service/src/bootstrap/ibd.rs
youngjoon-lee: True. The `max_download_iterations` is the last safety barrier. It's tested by the `max_download_iterations_reached` test.

We could add more logic to this `should_download` function, such as checking if the block is older the LIB. But, I'm careful to add any logic, which is part of `cryptarchia-engine`, to here. Those kinds of cases will be detected when we call methods of `cryptarchia-engine`, and we'll end up closing those downloads.

Still there may be many cases that makes requesters consume a lot of iterations. For example, a responder could repeatedly send the same block over and over. We could introduce any logic to avoid those cases in near future, but I think it's fine to just have the `max_download_iterations_reached` for now.

nomos-services/chain-service/src/bootstrap/config.rs
youngjoon-lee: > is this the number of retries? 

Yes, but we have some discussions in https://github.com/logos-co/nomos/pull/1506#discussion_r2241700517.

nomos-services/chain-service/src/bootstrap/ibd.rs
youngjoon-lee: Let's discuss in https://github.com/logos-co/nomos/pull/1506#discussion_r2242370044

nomos-services/chain-service/src/lib.rs
youngjoon-lee: I could actually remove this trait, but let me explain the reason here.

In our current code base, the input of block processing is `(cryptarchia, storage_blocks_to_remove)`. I personally don't like that the `storage_blocks_to_remove` flies around. Especially, the IBD module doesn't need to know anything about `storage_blocks_to_remove`. So, I abstracted those inputs in a single struct (trait).
I'll introduce the optimal refactoring (https://github.com/logos-co/nomos/issues/1505) but this is a temporary solution. 

Another reason is for easy testing. The real `Cryptarchia` accepts the real `Header` struct as an input, but I didn't want to build `Header`s that includes a proof and so on. By having this trait, we can easily build a mock struct that accepts a simple mocked header (block) in tests. 

However I agree that we don't necessarily need to have this temporary solution, which might be confusing. I will try to remove this and see how the entire code looks like.

nomos-services/chain-service/src/bootstrap/config.rs
youngjoon-lee: > can we provide a default?

Yes. I will.

nomos-services/chain-service/src/bootstrap/ibd.rs
youngjoon-lee: Good point. Let me think how to improve this.

nomos-services/chain-service/src/lib.rs
youngjoon-lee: Sorry, I think I didn't fully understand your comment. 
Is what you're proposing extracting this big code block as a function?
```
InitialBlockDownload::new(...).run(...).await?.into_inner();
```
I considered it but didn't do it, because I didn't want to define a new function anymore, which accepts many arguments that have a lot of generics. But, let me think again if this is what you're proposing.

nomos-services/chain-service/src/lib.rs
youngjoon-lee: a2cf216e I just removed the `ibd::Cryptarchia` trait. I could make tests work without the trait.

nomos-services/chain-service/src/bootstrap/ibd.rs
youngjoon-lee: 6f4e5386 I just removed redundant generic types. Now we have only one phantom (for `RuntimeServiceId`).

nomos-services/chain-service/src/bootstrap/ibd.rs
youngjoon-lee: @andrussal added a timeout parameter in the PR https://github.com/logos-co/nomos/pull/1478. It means that we won't wait more than the timeout for each peer. Once the timeout error occurs, we give up the stream. I think this approach is fine, though one slow peer could block all progress for a while, but not more than the timeout.
I've also considered other approaches (such as spawning tasks) but I think this is the simplest approach.

nomos-services/chain-service/src/bootstrap/config.rs
youngjoon-lee: Ok, I just removed the `download_limit` parameter 6f76555b, because I don't wanna add a new parameter that is not used yet and needs to be discussed about. I think it's still meaningful to have the other parameter `max_download_iterations`, but let's have further discussions in Discord. I think this PR can be merged independently with the discussion.

nomos-services/chain-service/src/bootstrap/config.rs
youngjoon-lee: > > can we provide a default?
> 
> Yes. I will.

e7cb8856 I just added a default value: 1000. If a peer returns at most 1000 blocks per response (as implemented in [here](https://github.com/logos-co/nomos/blob/e7cb8856d02b5c85ac1d73a4ec4e248596046ac1/nomos-services/chain-service/src/sync/block_provider.rs#L22)), a requester can receive at most 1000*1000 blocks (approx. a chain of 1 year) if he uses the default value.
But, I feel we need to have a further discussion regarding the ultimate stop-gap solution, as I mentioned in https://github.com/logos-co/nomos/pull/1506#discussion_r2244337719, though I think this PR can be merged independently.

nomos-services/chain-service/src/bootstrap/config.rs
youngjoon-lee: I just removed this comment because of the reason mentioned in https://github.com/logos-co/nomos/pull/1506#discussion_r2244337719

nomos-services/chain-service/src/bootstrap/config.rs
zeegomo: just an idea for the future, we can put a limit on the progress instead of on the raw number of retries

nomos-services/chain-service/src/lib.rs
zeegomo: you could extract the closure as a function, but I guess the issue is that it has lots of generics. It was just an idea, it's fine like this as well

nomos-services/chain-service/src/bootstrap/config.rs
youngjoon-lee: I also removed the `max_download_iterations` parameter 5a033226 as we discussed in [Discord](https://discord.com/channels/1111286067413405788/1400347387314307073). That parameter is not so reasonable. Even without that, node operators can terminate their nodes and restart by changing peers, if IBD takes a very long time.

nomos-services/chain-service/src/bootstrap/ibd.rs
youngjoon-lee: Yes. I just found a point to be improved for handling multiple downloads. I'm working on it as a next PR, which will refactor the `Download` as a `Stream`. I think this PR can be merged without that improvement.

nomos-services/chain-service/src/lib.rs
```diff
@@ -212,11 +216,13 @@ pub struct CryptarchiaSettings<Ts, Bs, NetworkAdapterSettings, BlendAdapterSetti
     pub network_adapter_settings: NetworkAdapterSettings,
     pub blend_adapter_settings: BlendAdapterSettings,
     pub recovery_file: PathBuf,
-    pub bootstrap: BootstrapConfig,
+    pub bootstrap: BootstrapConfig<NodeId>,
```
danielSanchezQ, line 219: Is this `NodeId` the peer id or the associated pk?

danielSanchezQ (approved):
I see there are improvements incoming. We should focus on simplifying if possible. It looks pretty good anyway.

nomos-services/chain-service/src/lib.rs
youngjoon-lee: It's PeerId. The reason why it's a generic is that the `NetworkAdapter` abstracts it.

PR #1508, opened 2025-07-30
chore: Reduce the size of some error enums

c410-f3r:

A large enum used in several places can cause a negative runtime impact. In fact, this is so common that the community created several lints to prevent such a scenario.

- [large_enum_variant](https://rust-lang.github.io/rust-clippy/master/?groups=perf#large_enum_variant)
- [result_large_err](https://rust-lang.github.io/rust-clippy/master/?groups=perf#result_large_err)
- [variant_size_differences](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/types/static.VARIANT_SIZE_DIFFERENCES.html)

This PR cuts 8 bytes from 3 different `TodoError` enum through the use of `Box<str>`, which doesn't change any intended behavior and ~100 bytes from the `BehaviourError` through the use of the same strategy applied to the largest variant.

### DIFF
--- a/tests/src/tests/cryptarchia/bootstrap.rs
+++ b/tests/src/tests/cryptarchia/bootstrap.rs
@@ -6,30 +6,31 @@ use tests::{
     adjust_timeout,
     nodes::validator::{create_validator_config, Validator},
     secret_key_to_peer_id,
-    topology::configs::create_general_configs,
+    topology::configs::{create_general_configs, GeneralConfig},
 };
 
 #[tokio::test]
 async fn test_ibd_behind_nodes() {
     let n_validators = 4;
+    let n_initial_validators = 2;
 
     let general_configs = create_general_configs(n_validators);
 
-    let mut validators = vec![];
-    for config in general_configs.iter().take(2) {
+    let mut initial_validators = vec![];
+    for config in general_configs.iter().take(n_initial_validators) {
         let config = create_validator_config(config.clone());
-        validators.push(Validator::spawn(config).await.unwrap());
+        initial_validators.push(Validator::spawn(config).await.unwrap());
     }
 
     println!("Testing IBD while initial validators are still bootstrapping...");
 
     let initial_peer_ids: HashSet<PeerId> = general_configs
         .iter()
-        .take(2)
+        .take(n_initial_validators)
         .map(|config| secret_key_to_peer_id(config.network_config.swarm_config.node_key.clone()))
         .collect();
 
-    let mut config = create_validator_config(general_configs[2].clone());
+    let mut config = create_validator_config(general_configs[n_initial_validators].clone());
     config.cryptarchia.bootstrap.ibd.peers = initial_peer_ids.clone();
 
     let failing_behind_node = Validator::spawn(config).await;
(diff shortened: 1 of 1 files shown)


danielSanchezQ (changes requested):
Sadly there are a few changes required to pass CI yet.

c410-f3r:
I am sorry, it should be working now

nomos-libp2p/src/behaviour/mod.rs
```diff
@@ -7,7 +7,6 @@ use std::error::Error;
 
 use cryptarchia_sync::ChainSyncError;
 use libp2p::{identity, kad, swarm::NetworkBehaviour, PeerId};
-use thiserror::Error;
```
danielSanchezQ: why remove this?

nomos-libp2p/src/behaviour/mod.rs
```diff
@@ -22,12 +21,18 @@ pub mod kademlia;
 // this limit so large. Remove this once we transition to smaller proofs.
 const DATA_LIMIT: usize = 1 << 18; // Do not serialize/deserialize more than 256 KiB
 
-#[derive(Debug, Error)]
```
danielSanchezQ: why?

nomos-libp2p/src/behaviour/mod.rs
c410-f3r: Because that is what I always do. Fewer dependencies reduce compilation times as well as the surface area of supply chain attacks. Not only for `thiserror` but also for any other applicable crate.

But that is only me. If you want I can revert this change, that is not a problem.

danielSanchezQ: I would recommend for future stuff to maybe ask. The amount of time for compiling this is probably negligible. The  time a dev spends on checking why an error is not an `Error` is probably worst. I would rather have it derive whats needed and keep it as it was. Thanks.

nomos-services/api/tests/todo.rs
```diff
@@ -165,13 +165,13 @@ mod todo {
     pub enum TodoError {
         /// Todo already exists conflict.
         #[schema(example = "Todo already exists")]
-        Conflict(String),
+        Conflict(Box<str>),
```
danielSanchezQ, line 170: Both String and Box<str> are heap allocated. What does it change and improve here?

c410-f3r: > A large enum used in several places can cause a negative runtime impact

`Box<str>` is made up of 2 words while `String` has 3 words. They have the same semantic and this conversion trick is commonly used in several projects.

Every time a function returns `Result<T, TodoError>`, the program allocates `max(size_of(T), size_of(TodoError)) + padding` bytes of stack memory, which can become a runtime problem if such a thing is a hot path called in several locations.

Of course, this particular scenario probably won't make such of a difference but it is about the message, the same message I tried to propagate trying to reduce the usage of third-party crates.

But again that is just me. If you want, I can revert this change.

danielSanchezQ: Cool. Thanks for the explanation. Not sure this is totally worth here as its just for testing. But we can go ahead with it. Thanks.

c410-f3r:
Thanks. The code has been updated

Issue #1460, opened 2025-07-16
Max retries for connection establishments


youngjoon-lee:
The current codebase doesn't retry connecting to a core node.

However, the spec defines the max retries parameter:
- [Core node parameter](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81c9aad4d6ac7521a286)
- [Edge node parameter](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81c9aad4d6ac7521a286)

The behaviors for each parameter are defined in:
- [Core node](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df8179845dc7a33bf1e462)
- [Edge node](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81a5b4cef6c9b446cee5)

In short, core/edge nodes should retry dialing.

PR #1511, opened 2025-07-31
feat(blend): enforce peering degrees for core->core and core->edge behaviours

ntn-x2:

Fixes https://github.com/logos-co/nomos/issues/1484.

This PR is built on top of the core->core and core->edge behaviour split PR: https://github.com/logos-co/nomos/pull/1503.

It introduces the logic to rate-limit incoming and outgoing connections between core nodes, as well as incoming connections from edge nodes towards core nodes.

The new logic simplify things quite a bit in terms of event handling for both behaviour as well as swarm. Starting from the swarm, it gets now notified whenever a peer goes from healthy->unhealthy, unhealthy->healthy, and when a peer is disconnected, along with their latest negotiated state. The main difference with the previous logic is that we let the connection handler drop the stream and the swarm to close the connection, and only when the connection is closed, the behaviour captures the event and notifies the swarm about it, so that the swarm can react to this event.

The core->core behaviour does not hard reject a connection, since at that stage we do not really know who the other peer is, and rejecting it might hinder the logic of other behaviours within the same swarm. So we simply add every new connection to the connection map, removing them when the connection is dropped: in the case the peer is not "relevant" for us, a dummy handler is used which will instruct the swarm to close the connection right away, releasing a connection slot for a new peer to utilise. The logic for the core->edge behaviour is very similar, but it tracks and operates only on incoming connections.

Yes. There are a lot of comments to ease understanding of the new parts, although I think the new changes make the code easier to follow and understand, overall.

@ntn-x2 

Yes.

No. Any unclear bits have already been rectified in the Blend v1 spec, linked to the issue this PR closes.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/network/src/core/with_core/behaviour/mod.rs
+++ b/nomos-blend/network/src/core/with_core/behaviour/mod.rs
@@ -53,13 +53,12 @@ pub struct Config {
 /// they are propagated to the rest of the network, making sure no peer marks
 /// this node as malicious due to an invalid Blend message.
 pub struct Behaviour<ObservationWindowClockProvider> {
-    /// Tracks ALL established connections between this node and other core
-    /// nodes.
+    /// Tracks connections between this node and other core nodes.
     ///
     /// Only connections with other core nodes that are established before the
     /// specified connection limit is reached will be upgraded and the state of
     /// the peer negotiated, monitored, and reported to the swarm.
-    established_connections: HashMap<(PeerId, ConnectionId), Option<NegotiatedPeerState>>,
+    negotiated_peers: HashMap<(PeerId, ConnectionId), NegotiatedPeerState>,
     /// Queue of events to yield to the swarm.
     events: VecDeque<ToSwarm<Event, Either<FromBehaviour, Infallible>>>,
     /// Waker that handles polling
@@ -112,7 +111,7 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
         current_membership: Option<Membership<PeerId>>,
     ) -> Self {
         Self {
-            established_connections: HashMap::with_capacity(*config.peering_degree.end()),
+            negotiated_peers: HashMap::with_capacity(*config.peering_degree.end()),
             events: VecDeque::new(),
             waker: None,
             exchanged_message_identifiers: HashMap::with_capacity(
@@ -172,13 +171,13 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
         let message_id = message.id();
 
         let serialized_message = serialize_encapsulated_message(message);
-        let mut num_peers = 0;
-        self.established_connections
+        let mut num_peers = 0u32;
+        self.negotiated_peers
             .iter()
             // Exclude the connection the message was received from.
             .filter(|(connection, _)| (excluded_connection != Some(**connection)))
             // Exclude from the list of candidate peers any peer that is not in a healthy state.
-            .filter(|(_, peer_state)| **peer_state == Some(NegotiatedPeerState::Healthy))
+            .filter(|(_, peer_state)| **peer_state == NegotiatedPeerState::Healthy)
             .for_each(|((peer_id, connection_id), _)| {
                 if self
                     .exchanged_message_identifiers
@@ -224,9 +223,9 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
 
     #[must_use]
     pub fn num_healthy_peers(&self) -> usize {
-        self.established_connections
+        self.negotiated_peers
             .values()
-            .filter(|state| **state == Some(NegotiatedPeerState::Healthy))
+            .filter(|state| **state == NegotiatedPeerState::Healthy)
             .count()
     }
 
@@ -238,7 +237,7 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
     pub fn available_connection_slots(&self) -> usize {
         self.peering_degree
             .end()
-            .saturating_sub(self.established_connections.len())
+            .saturating_sub(self.negotiated_peers.len())
     }
 
     fn try_wake(&mut self) {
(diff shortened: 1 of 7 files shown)


nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -250,30 +497,78 @@ where
...
+                // comes, if we have a slot available for it.
+                (Endpoint::Dialer, _) => {}
+                // The disconnected peer was never upgraded to the Blend
+                // protocol (i.e., its negotiated state is `None`), either because they do not
+                // support it or because we prevented it due to connection limits.
+                // So, we do not notify the swarm at all about this, as this is an
+                // internal detail.
+                // We bubble this up to the swarm only during tests, to test that connections beyond
+                // our maximum capacity are effectively ignored.
+                (peer_role, None) => {
```
youngjoon-lee: > The core->core behaviour removes the old map, and introduces two maps, one that tracks outgoing connections and one that tracks incoming connections

I'm trying to clearly understand the reason why we need to maintain inbound and outbound connections separately. 
I have several questions for better understanding:
1. If we had 4 inbound connections and 0 outbound connection, and if all the 4 inbound connections have been disconnected, shouldn't we try to establish 4 new outbound connections to meet the minimum peering degree? If so, shouldn't we bubble up an event to the swarm in case of `Endpoing::Dialer` as well?
2. Do we limit the number of **established** connections (not negotiated connections). I think it was the same before because we relied on only the `connection_limits` behaviour. Just asking to check if my understanding is correct.
  2.a. If so, if we have 6 established connections, but if all of them are not negotiated, do we do nothing? Again, I know this is not something wrong made by this PR, but asking to check if we're on the same page.

These questions are related to another question why we name `min_peering_degree` as `min_outgoing`.
https://github.com/logos-co/nomos/blob/ba35ec1a80e295de4ba0fe3eca6984a8b0190154/nomos-blend/network/src/core/with_core/behaviour/mod.rs#L67-L71

Please let me know if I'm missing something. I always feel that it's not so intuitive to understand the connection lifecycle and limits because of the rust-libp2p structure (not because of the implementation in this PR).

ntn-x2: > If we had 4 inbound connections and 0 outbound connection, and if all the 4 inbound connections have been disconnected, shouldn't we try to establish 4 new outbound connections to meet the minimum peering degree? If so, shouldn't we bubble up an event to the swarm in case of Endpoing::Dialer as well?

No, the spec distinguishes between inbound and outbound connections. The swarm, when it's created, will try to dial the minimum number of peers. Then, getting to 0 outbound connections means there have been 4 disconnect events, and each of which has been notified to the swarm which will try to establish new ones as soon as we drop below the minimum number outgoing peering degree. Incoming connections are entirely managed and processed by the behaviour, since the swarm is not really concerned about those, but it is concerned about the outgoing peering degree.

> Do we limit the number of established connections (not negotiated connections). I think it was the same before because we relied on only the connection_limits behaviour. Just asking to check if my understanding is correct.

Yeah I think I can change the configuration of the limits behaviour to allow for 302 max incoming connections (300 from edge nodes + 2 from core nodes), 6 outgoing connections (4 are minimum + 2 in case of unhealthy peers, if there is no incoming connections established), and 306 total connections. To answer your question, yes, we limit the number of connections, be them upgraded or not. If a peer is not meant to talk to us, the use of the dummy connection handler will notify the swarm to close the connection immediately, freeing the slot for another peer. We do not have any sophisticated DoS protection mechanism, neither did we before.

> 2.a. If so, if we have 6 established connections, but if all of them are not negotiated, do we do nothing? Again, I know this is not something wrong made by this PR, but asking to check if we're on the same page.

Yes, this is correct. Ideally, the outgoing connections are initiated by the swarm which tries to connect to the right peers. Once we implement connection retries as part of https://github.com/logos-co/nomos/issues/1460, we can bubble up one more event to the swarm which says "Hey, we failed to connect to peer X", and the swarm will likely choose another peer to try to connect to. This behaviour is currently not implemented.

> These questions are related to another question why we name min_peering_degree as min_outgoing.

If you mean `min_outgoing` and `max_total`, is because the spec defines a minimum outgoing degree and a maximum of incoming degree. The sum of the two gives us `max_total`. But, outgoing degree can go above the minimum, taking "slots" from the incoming degree, up to the maximum. So we can have from 4 to 6 outbound connections, and from 0 to 2 incoming connections eventually. These are explained in the [Core network bootstrapping and maintenance section](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81e1b624ea608fc9e39f) of the spec, as well as in this [Discord thread](https://discord.com/channels/1111286067413405788/1399369352775467170).

Please let me know if some bits are still unclear.

ntn-x2: > I'm trying to clearly understand the reason why we need to maintain inbound and outbound connections separately.

As for this, it's simply for easier lookup to know the current number of incoming and outgoing connections, to decide what to do with a new incoming or outgoing connection request. Else, we would use a counter and a single map, but I thought this approach would be cleaner.

ntn-x2:
@youngjoon-lee I pushed a commit that changes the enforced connection limits with the connection limits behaviour: https://github.com/logos-co/nomos/pull/1511/commits/c860fb1c46cb4a5ecfdaf335253f558798ff3b1e. We can enforce an upper bound on both incoming and outgoing connections, but for the actual logic differentiating between the two and between edge and core nodes, I believe the approach in this PR is still needed. Please let me know your opinion about it, thank you!

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: Thanks for detailed explanations.

Now I realized that all of my questions came from different understandings about peering degrees ($\Phi_{CC}^{Min}$, $\Phi_{CC}^{Max}$) defined in the [spec](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81828f2bf8fee56fc410).
image

I thought the spec doesn't differentiate inbound vs. outbound.

Also, I think the [step 6 and 7](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81b8bc75cde60ce4e4d3) in the spec also doesn't differentiate inbound vs. outbound.

I think you know better which way is easier to implement. I'm okay to go with the way you think it's better, but I want to be on the same page with you first before approving this PR.

I can't open the discord link for some reason. I'll contact you via discord.

ntn-x2:
Based on the internal Discord discussion, I will move this back to draft and update it with the latest decision.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -47,7 +53,13 @@ const LOG_TARGET: &str = "blend::network::behaviour";
...
 /// this node as malicious due to an invalid Blend message.
 pub struct Behaviour<ObservationWindowClockProvider> {
-    negotiated_peers: HashMap<PeerId, NegotiatedPeerState>,
+    /// Tracks ALL established connections between this node and other core
+    /// nodes.
+    ///
+    /// Only connections with other core nodes that are established before the
+    /// specified connection limit is reached will be upgraded and the state of
+    /// the peer negotiated, monitored, and reported to the swarm.
+    established_connections: HashMap<(PeerId, ConnectionId), Option<NegotiatedPeerState>>,
```
youngjoon-lee: I'd like to ask about more details of the background why we change this from `negotiated_peers` to `established_connections`. 

I understand how the logic works well with the `established_connections`. 
But, I'm just personally curious if there was any special reason to change this from `negotiated_peers`. For example, is there any case that cannot be covered if we use the `negotiated_peers`?

PR #1512, opened 2025-07-31
feat(circuits): Implement binary wrappers

strinnityk:
Implement wrappers for `pol` circuit, prover and verifier. 

Yes.

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.


circuits/pol/src/lib.rs
```diff
@@ -1 +1,67 @@
...
+///
+/// * `inputs_file` - The path to the file containing the public and private
+///   inputs.
+/// * `witness_file` - The path to the file where the witness will be written.
+///
+/// # Returns
+///
+/// An `io::Result<PathBuf>` which contains the path to the witness file if
+/// successful, or an `io::Error` if the command fails.
+pub fn pol(inputs_file: &PathBuf, witness_file: &PathBuf) -> io::Result<PathBuf> {
```
danielSanchezQ: I'm wondering if we really want this. Unless wrapped, what we would want is to provide the inputs directly from code, not really a file (cc @thomaslavaur). This implementation should handle dumping into a file those inputs with the proper format.

circuits/pol/src/lib.rs
```diff
@@ -1 +1,67 @@
...
+    let mut inputs_file = NamedTempFile::new()?;
+    let witness_file = NamedTempFile::new()?;
+    inputs_file.write_all(inputs.as_bytes())?;
+
+    pol(
+        &inputs_file.path().to_path_buf(),
+        &witness_file.path().to_path_buf(),
+    )?;
+    read_to_string(witness_file.path())
+}
```
danielSanchezQ: Yes, this, but the inputs should probably not be `&str` but our proper types.

strinnityk: That's taken care of, I hadn't updated them to `&[u8]` yet.

circuits/pol/src/lib.rs
strinnityk: Because the binaries work with file paths, I implemented a simple wrapper on top of them that uses `PathBuf` (this one), and another on top of it that handles the file management (`pol_from_content`).
The same thing happens with the `prover` and `verifier`.
I'm not sure we can make the binary work with content directly without modifying the source code.

circuits/prover/src/traits.rs
```diff
@@ -0,0 +1,21 @@
...
+    ///
+    /// # Returns
+    ///
+    /// An [`io::Result<(String, String)>`] which contains the proof and public
+    /// inputs as strings if successful, or an [`io::Error`] if the command
+    /// fails.
+    fn generate_proof(
+        circuit_contents: &[u8],
+        witness_contents: &[u8],
+    ) -> io::Result<(String, String)>;
```
danielSanchezQ: Why is the output a string? Shouldnt it be a `[u8]`?

circuits/prover/src/lib.rs
```diff
@@ -1 +1,198 @@
+pub mod rapidsnark;
+pub mod traits;
```
danielSanchezQ: This should probably be private, and then we just re-export the actual trait inside. `traits` is too a generic kind of naming.

circuits/prover/src/lib.rs
```diff
@@ -1 +1,198 @@
...
+        &circuit_file.path().to_path_buf(),
+        &witness_file.path().to_path_buf(),
+        &proof_file.path().to_path_buf(),
+        &public_file.path().to_path_buf(),
+    )?;
+
+    let proof = read_to_string(proof_file)?;
+    let public = read_to_string(public_file)?;
+    Ok((proof, public))
+}
```
danielSanchezQ: This should not be in the lib. Lib should contain the public trait and a type that implements that trait. This should be under a binary based module or something like that.

circuits/verifier/src/lib.rs
```diff
@@ -1 +1,156 @@
...
+    verification_key_file.write_all(verification_key_contents)?;
+    public_file.write_all(public_contents)?;
+    proof_file.write_all(proof_contents)?;
+
+    verifier(
+        &verification_key_file.path().to_path_buf(),
+        &public_file.path().to_path_buf(),
+        &proof_file.path().to_path_buf(),
+    )
+}
```
danielSanchezQ: Same same. This should be part of an underlaying implementation. We should just export as a library the trait and the type that implements the functionality. We could still make this public, but under a module.

circuits/verifier/src/rapidsnark.rs
```diff
@@ -0,0 +1,13 @@
+use crate::{traits::Verifier, verifier_from_contents};
+
+pub struct Rapidsnark;
+
+impl Verifier for Rapidsnark {
+    fn verify(
+        verification_key_contents: &[u8],
+        public_contents: &[u8],
+        proof_contents: &[u8],
+    ) -> std::io::Result<bool> {
```
danielSanchezQ: This should not be io, it should be a custom result. This is a library, the underlaying implementation may use io, or may not.

circuits/verifier/src/traits.rs
```diff
@@ -0,0 +1,22 @@
...
+    ///
+    /// # Returns
+    ///
+    /// An [`io::Result<bool>`] which indicates whether the verification was
+    /// successful or not, or an [`io::Error`] if the command fails.
+    fn verify(
+        verification_key_contents: &[u8],
+        public_contents: &[u8],
+        proof_contents: &[u8],
+    ) -> io::Result<bool>;
```
danielSanchezQ: ditto

circuits/mantle/pol/src/traits.rs
```diff
@@ -0,0 +1,15 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `inputs` - A string containing the public and private inputs.
+    ///
+    /// # Returns
+    ///
+    /// An [`io::Result<Vec<u8>>`] which contains the witness if successful, or
+    /// an [`io::Error`] if the command fails.
+    fn generate_witness(inputs: &str) -> io::Result<Vec<u8>>;
```
danielSanchezQ: no io here, same as other comments

circuits/mantle/pol/src/pol.rs
```diff
@@ -0,0 +1,11 @@
+use std::io;
+
+use crate::{WitnessGenerator, pol_from_content};
+
+pub struct Pol;
+
+impl WitnessGenerator for Pol {
+    fn generate_witness(inputs: &str) -> io::Result<Vec<u8>> {
```
danielSanchezQ: gone `io`

circuits/prover/src/lib.rs
strinnityk: So something like:
```
- circuits
  - Mantle
    - pol
      - bin (crate that adds the wrapper functions) 
      - lib (trait + type)
```
Why a bin crate, though?

circuits/prover/src/traits.rs
strinnityk: Makes sense, I'll check.

circuits/mantle/pol/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+mod pol;
+mod witness_generator;
+mod wrappers;
+
+pub use pol::Pol;
+pub use witness_generator::WitnessGenerator;
+
+pub type Result<T> = std::io::Result<T>;
```
danielSanchezQ: This is just renaming the issue from before. I think it is better to just have an associated type in the traits with the error type (which will be io::Error for the current implementations).

strinnityk: Makes sense. Will do so.

circuits/mantle/pol/src/pol.rs
```diff
@@ -0,0 +1,11 @@
+use crate::{WitnessGenerator, wrappers::pol_from_content};
+
+pub struct Pol;
+
+impl WitnessGenerator for Pol {
+    type Error = std::io::Error;
+
+    fn generate_witness(inputs: &str) -> Result<Vec<u8>, Self::Error> {
```
danielSanchezQ, line 8: is inputs the public inputs? This is wrong API. It should definitely be something else. `str` would be just for the (now) underlying binaries that we are gonna use 🤔 . We may have to rethink this either now or later.

circuits/verifier/src/rapidsnark.rs
```diff
@@ -0,0 +1,15 @@
...
+
+pub struct Rapidsnark;
+
+impl Verifier for Rapidsnark {
+    type Error = std::io::Error;
+
+    fn verify(
+        verification_key_contents: &[u8],
+        public_contents: &[u8],
+        proof_contents: &[u8],
```
danielSanchezQ, line 11: Yeah, now that I see this. Is the same. We should have types for this. Concrete types which later we can bridge to whatever needed. 🤔

danielSanchezQ (approved):
The APIs and the structure will be reworked in a following PR

PR #1514, opened 2025-08-01
feat(blend): proxy service skeleton

youngjoon-lee:
### Background

This PR introduces the **skeleton for the Blend Proxy Service**, which I will now refer to simply as `BlendService`.

Previously, I proposed the PR #1480. However, it had a notable downside: the user of `BlendService` (i.e. `ChainService`) had to implement `Adapter`s for the underlying services (`BlendCoreService` and `BlendEdgeService`). I agree that this is not a good abstraction. 

### What's new

This PR presents a completely new approach -- one we haven't used before:
- `ChainService` no longer defines a concrete `BlendAdapter`. 
  - Instead, it holds a generic `BlendService` type in its definition.
  - It no longer needs to know whether libp2p or any other backend is used underneath.
  - All it needs to do is open a relay to the `BlendService` and send messages.

You can find this key change in the code: https://github.com/logos-co/nomos/pull/1514#discussion_r2248402133.

This structure improves the abstraction between the services and offers better reusability. If other services besides `ChainService` want to use `BlendService` in the future, they won't need to reimplement their own `BlendAdapter`.

If this structure works well, we may want to adopt it for other adapter-heavy services too.

### A note on `ServiceExt`

You'll also notice a trait called `nomos_blend_service::ServiceExt`. I left a self-comment in the PR for further explanations: https://github.com/logos-co/nomos/pull/1514#discussion_r2248395275.

Yes

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nodes/nomos-executor/config.yaml
+++ b/nodes/nomos-executor/config.yaml
@@ -11,7 +11,6 @@ network:
     node_key: ea1e1dcc31612bd20987f017f0ca435cd2a59a4bd9fd6e14883b4803ae3b803d
     initial_peers: [ ]
 blend:
-blend_core:
   backend:
     listening_address: /ip4/127.0.0.1/udp/3001/quic-v1
     node_key: ea1e1dcc31612bd20987f017f0ca435cd2a59a4bd9fd6e14883b4803ae3b803d
@@ -38,26 +37,6 @@ blend_core:
     message_frequency_per_round: 1.0
     redundancy_parameter: 0
     maximum_release_delay_in_rounds: 3
-  membership:
-    - id: 12D3KooWKD62Q1nAAMHeV2Cx3rNNMzjgHkTT2edFMzHZ5ur7sGnh
-      address: /ip4/127.0.0.1/udp/8720/quic-v1
-      public_key: cd81cab2112a1cf8e6d986d53c0ab6c740e3a82dc04e0a6497f62727969fa5fd
-    - id: 12D3KooWA5LkQcuefjUBTktxCXUQ4rx4PcqRY1tgbSpV9eZDHWir
-      address: /ip4/127.0.0.1/udp/8721/quic-v1
-      public_key: 92f041e26306548444400e88eefa622dc125a45f3e3119a04ba1ffe4946dd4e6
-blend_edge:
-  backend:
-    node_key: ea1e1dcc31612bd20987f017f0ca435cd2a59a4bd9fd6e14883b4803ae3b803d
-  crypto:
-    signing_private_key: 32e4b52b78e0e8273f31e2ae0826d09b0348d4c66824af4f51ec4ef338082211
-    num_blend_layers: 1
-  time:
-    # 21,600 blocks * 30s per block * 1s per round = 648,000 rounds
-    rounds_per_session: 648000
-    rounds_per_interval: 30
-    round_duration: [ 1, 0 ]
-    # Defined in the spec
-    rounds_per_observation_window: 30
   membership:
     - id: 12D3KooWKD62Q1nAAMHeV2Cx3rNNMzjgHkTT2edFMzHZ5ur7sGnh
       address: /ip4/127.0.0.1/udp/8720/quic-v1
--- a/nodes/nomos-executor/executor/src/config/mod.rs
+++ b/nodes/nomos-executor/executor/src/config/mod.rs
@@ -1,8 +1,8 @@
 use color_eyre::eyre::Result;
 use nomos_node::{
     config::{
-        mempool::MempoolConfig, update_blend, update_cryptarchia_consensus, update_network,
-        BlendArgs,
+        blend::BlendConfig, mempool::MempoolConfig, update_blend, update_cryptarchia_consensus,
+        update_network, BlendArgs,
     },
     generic_services::{MembershipService, SdpService},
     CryptarchiaArgs, HttpArgs, LogArgs, NetworkArgs,
@@ -11,9 +11,9 @@ use overwatch::services::ServiceData;
 use serde::{Deserialize, Serialize};
 
 use crate::{
-    ApiService, BlendCoreService, BlendEdgeService, BlendService, CryptarchiaService,
-    DaDispersalService, DaIndexerService, DaNetworkService, DaSamplingService, DaVerifierService,
-    NetworkService, RuntimeServiceId, StorageService, TimeService,
+    ApiService, CryptarchiaService, DaDispersalService, DaIndexerService, DaNetworkService,
+    DaSamplingService, DaVerifierService, NetworkService, RuntimeServiceId, StorageService,
+    TimeService,
 };
 
 #[derive(Deserialize, Debug, Clone, Serialize)]
@@ -21,9 +21,7 @@ pub struct Config {
     #[cfg(feature = "tracing")]
     pub tracing: <nomos_node::Tracing<RuntimeServiceId> as ServiceData>::Settings,
     pub network: <NetworkService as ServiceData>::Settings,
-    pub blend: <BlendService as ServiceData>::Settings,
-    pub blend_core: <BlendCoreService as ServiceData>::Settings,
-    pub blend_edge: <BlendEdgeService as ServiceData>::Settings,
+    pub blend: BlendConfig,
     pub da_dispersal: <DaDispersalService as ServiceData>::Settings,
     pub da_network: <DaNetworkService as ServiceData>::Settings,
     pub da_indexer: <DaIndexerService as ServiceData>::Settings,
@@ -60,7 +58,7 @@ impl Config {
         #[cfg(feature = "tracing")]
         nomos_node::config::update_tracing(&mut self.tracing, log_args)?;
         update_network::<RuntimeServiceId>(&mut self.network, network_args)?;
-        update_blend(&mut self.blend_core, &mut self.blend_edge, blend_args)?;
+        update_blend(&mut self.blend, blend_args)?;
         update_http(&mut self.http, http_args)?;
         update_cryptarchia_consensus(&mut self.cryptarchia, cryptarchia_args)?;
         Ok(self)
--- a/nodes/nomos-executor/executor/src/main.rs
+++ b/nodes/nomos-executor/executor/src/main.rs
@@ -65,12 +65,14 @@ async fn main() -> Result<()> {
         return Ok(());
     }
 
+    let (blend_config, blend_core_config, blend_edge_config) = config.blend.into();
+
     let app = OverwatchRunner::<NomosExecutor>::run(
         NomosExecutorServiceSettings {
             network: config.network,
-            blend: config.blend,
-            blend_core: config.blend_core,
-            blend_edge: config.blend_edge,
+            blend: blend_config,
+            blend_core: blend_core_config,
+            blend_edge: blend_edge_config,
             #[cfg(feature = "tracing")]
             tracing: config.tracing,
             http: config.http,
--- a/nodes/nomos-node/config.yaml
+++ b/nodes/nomos-node/config.yaml
@@ -11,7 +11,6 @@ network:
     node_key: ea1e1dcc31612bd20987f017f0ca435cd2a59a4bd9fd6e14883b4803ae3b803d
     initial_peers: [ ]
 blend:
-blend_core:
   backend:
     listening_address: /ip4/127.0.0.1/udp/3001/quic-v1
     node_key: ea1e1dcc31612bd20987f017f0ca435cd2a59a4bd9fd6e14883b4803ae3b803d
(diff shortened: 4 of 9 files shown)


nomos-services/blend/src/lib.rs
```diff
@@ -1,4 +1,122 @@
...
+        edge::BlendService<edge::backends::libp2p::Libp2pBlendBackend, PeerId,
+            <core::network::libp2p::Libp2pAdapter<RuntimeServiceId> as core::network::NetworkAdapter<RuntimeServiceId>>::BroadcastSettings, RuntimeServiceId>,
+        RuntimeServiceId,
+    >
+{
+    type BroadcastSettings =
+        <core::network::libp2p::Libp2pAdapter<RuntimeServiceId> as core::network::NetworkAdapter<
+            RuntimeServiceId,
+        >>::BroadcastSettings;
+}
```
youngjoon-lee: This defines **additional** types required to interact with `BlendService`. Basic types are already provided by `overwatch::ServiceData`, but they're sometimes not sufficient.

Currently, `BroadcastSettings` is the only one that `ChainService` needs to know when relaying messages to the `BlendService`. But `BroadcastSettings` (e.g. `topic` if libp2p) is determined by the `NetworkAdapter` implemented by the inner `BlendCoreService`. 

However, as noted earlier, we want to allow `ChainService` to interact with `BlendService` _**without knowing the details**_ of its internal subcomponents. To achieve that, I've implemented `ServiceExt` for a specific concrete `BlendService` type.

youngjoon-lee: P.S.
I recall a discussion around `BroadcastSettings`, specifically the idea that `ChainService` shouldn't have to specify it when `BlendService` is in edge mode. However, we've abstracted `ChainService` away from knowing the internal mode of `BlendService`. So, `ChainService` always specifies `BroadcastSettigns` when relaying messages to the `BlendService`.
For more details, let me summarize the reason why we introduced `BroadcastSettings`:
- Broadcasting is part of the Blend protocol.
- Multiple services (not just `ChainService`) should be able to use `BlendService`, and each might need to configure `BroadcastSettings` independently. 
Thus, regardless of the current mode of `BlendService`, `BroadcastSettings` should be always specified.

nomos-services/chain-service/src/lib.rs
```diff
@@ -422,12 +420,15 @@ where
         + Sync
         + 'static,
     NetAdapter::Settings: Send + Sync + 'static,
-    BlendAdapter: blend::BlendAdapter<RuntimeServiceId, Tx = ClPool::Item, BlobCertificate = DaPool::Item>
-        + Clone
+    BlendService: ServiceData<
+            Message = nomos_blend_service::message::ServiceMessage<
+                <BlendService as BlendServiceExt>::BroadcastSettings,
+            >,
+        > + BlendServiceExt
```
youngjoon-lee, line 439: This is a KEY CHANGE. We no longer need an adapter. Instead, we have the `BlendService` as the generic type. Using this, we can open a relay and send messages to it.

nomos-services/blend/src/lib.rs
```diff
@@ -1,4 +1,122 @@
...
+                error!(target: LOG_TARGET, "Failed to relay message to core service: {e:?}");
+            }
+        }
+
+        Ok(())
+    }
+}
+
+/// Defines additional types required for communicating with [`BlendService`].
+pub trait ServiceExt {
```
ntn-x2, line 106: This is good, and it goes in the direction I was initially suggesting, when I joined, that services should not really depend on each other explicitly, but rather be abstracted away. Now, I think it would still be fair to have this trait declared in the service itself, e.g., in the chain service, but its implementation can be provided either by the blend service, if it has all the information available to implement it, or by the node, if the implementation requires types not available to the blend service or interactions between different services that the blend service is not aware of.

Defining the trait in the chain service, allows it to add any functionality it needs without affecting any other service that might depend on the blend service via a different trait definition.

This is the approach taken by the Substrate framework, which has the upside of having smaller service definitions, and quite a large "runtime" definition. But I would argue that is a better approach since everything is defined in a single place and has a hierarchical top-down approach.

TL;DR: I am in support of this new approach, but I would move the "helper" trait definition into the crate that needs it, and only provide the impl to that trait either in the consumed service (as it could be in this case), or in the node if impl needs some higher-level logic.

ntn-x2: I would also suggest to add the `ServiceData` trait requirement a supertrait of this trait, if this trait is moved to the chain service crate, so everything the chain service needs from the blend service would be included in this trait and not split between the trait and the chain service definition itself. What do you think?

nomos-services/chain-service/src/lib.rs
```diff
@@ -422,12 +420,15 @@ where
         + Sync
         + 'static,
     NetAdapter::Settings: Send + Sync + 'static,
-    BlendAdapter: blend::BlendAdapter<RuntimeServiceId, Tx = ClPool::Item, BlobCertificate = DaPool::Item>
-        + Clone
+    BlendService: ServiceData<
```
ntn-x2, line 435: See my other comment for this `ServiceData` requirement: https://github.com/logos-co/nomos/pull/1514#discussion_r2250534454

ntn-x2 (commented):
I like this approach more than having services depend on each other. I would personally only move the trait definition into the service that needs that trait, so changes can be as localized as possible without affecting other services that might need other features from the blend service, so I think having many trait helpers instead of a single trait definition as part of the blend service would be a better alternative.

nomos-services/blend/src/lib.rs
youngjoon-lee: Thanks.

Initially, I thought about defining the `ServiceExt` trait in the chain service (not in the blend service), as you suggested. But, I felt that it is not natural to implement the trait in the chain service, like below, because it means that the chain service implementer should know the details of the blend service to implement `ServiceExt`.
```rust
impl<...> ServiceExt
    for BlendService<
        core::BlendService<
            ..,
            core::network::libp2p::Libp2pAdapter<RuntimeServiceId>,
            ..,
        >,
        ..,
    >
{
    type BroadcastSettings =
        <core::network::libp2p::Libp2pAdapter<..> as core::network::NetworkAdapter<..>>::BroadcastSettings;
}
```

It would make sense to define this trait in the `nomos-node` crate, but then we should devise a way to implement the trait for the `BlendService`, which is defined outside of the `nomos-node` crate.
Let me try, although I think we can keep the current structure as it is, IMO.

ntn-x2: I was not suggesting to implement the trait in the chain service, but only to define it. It would then be implemented in the blend service itself (as it's done now in this PR), for the blend service type.

ntn-x2: In case the trait could only be implemented by the node because it requires some higher-level logic, then a simple tuple struct that wraps the blend service (with all concrete types since we are in the node binary) would be sufficient to satisfy the orphan rule requirement.

youngjoon-lee: > I was not suggesting to implement the trait in the chain service, but only to define it. It would then be implemented in the blend service itself (as it's done now in this PR), for the blend service type.

I see your point. But then, we will encounter a circular dependency: blend <-> chain. 

I think, the `BroadcastSettings` type is a essential thing for all users who want to send messages to the blend service. Thus, I don't think we need to move the `ServiceExt` definition to the user side.

ntn-x2: Mhh, well if the abstraction is correct, chain should not depend on blend anymore but only the other way around, that is the whole point of the trait and the abstraction.

But in the end it's your call. Perhaps let's just add a comment on the trait that explains what it does and how it's meant to be used.

youngjoon-lee: > Mhh, well if the abstraction is correct, chain should not depend on blend anymore but only the other way around, that is the whole point of the trait and the abstraction.

Sorry, I didn't clearly understand this point. Why the other way around? I think the chain service should have the blend service as a dependency, because it is the user of the blend service.

youngjoon-lee: > But in the end it's your call. Perhaps let's just add a comment on the trait that explains what it does and how it's meant to be used.

6ff189d1 Yeah. In the new comments, I emphasized that the `ServiceExt` defines types that are not covered by `ServiceData`. I hope this delivers my intention clearly. 

IMO, I think we can go with this structure. But, if you have any idea for improvement that I don't understand now, could you please propose it as a next PR? I think it's an easier way if you don't mind.

ntn-x2: Sure, I can give it a try when I have some time to do it. But in short my suggestion was something along the lines of the old adapter approach, but with the implementation provided by the blend crate itself. So the chain service would define the trait with the functionalities it needs (it could also be just the `BroadcastSettings` type as in this PR), it removes any direct dependency on blend since it would only depend on this trait impl wherever it comes from, and the trait impl would be provided in the blend service. This is dependency injection and it also makes unit testing of services more manageable, which we lack right now because of how complicated it is to set up a mock testing environment. Anyway, this is getting out of scope for this PR, so let's go with this approach, and perhaps gather more opinions from other team members as to whether this is a better way to go than we currently have.

danielSanchezQ: I think this is leading to something. We can make POC of the idea at some point @ntn-x2. But It would have to be in the future. We need to keep the focus on the targets at hand at the moment.

ntn-x2: Sure, I agree. I also think having a good testing infrastructure in place before going live is a must, of course, and adapting the architecture for that might help in that regard as well. But yes, once we develop all features we can think about improvements to increase our test coverage.

youngjoon-lee: I now understand the approach you suggested. Thank you for detailed explanation. I agree with the approach though I still have something unclear that I cannot be sure until I try it.

If you're okay, could we first merge this PR and try the approach soon? By doing that, we can proceed the further developments (such as the session transition) and trying the approach in parallel. 
I feel that trying the approach will involve many discussions.
I think this PR is meaningful (though it's not perfect) because this is the first step which can be improved/extended.

ntn-x2: Sorry, yes, forgot to approve! I just wanted to make sure my point comes across clear so we can evaluate it without any misunderstandings. But I agree this PR is good to go, so I will approve it just now!

ntn-x2 (approved):

nomos-services/chain-service/src/lib.rs
```diff
@@ -1464,6 +1460,24 @@ where
...
+                nomos_blend_service::message::NetworkMessage {
+                    message: wire::serialize(&messages::NetworkMessage::Block(block)).unwrap(),
+                    broadcast_settings: blend_broadcast_settings,
+                },
+            ))
+            .await
+        {
+            error!(target: LOG_TARGET, "Failed to relay block to blend service: {e:?}");
+        }
+    }
```
danielSanchezQ: Althouhg I see the point. I think it it could get better. This is gonna be no trouble for blend as its sole responsibility is to send the message. But we will end up abusing relays for other services APIs. I would suggest for future updates of the system to keep the adapters, but keep them internal to the services. Each service will hold an adapter for each of the other services it contains. Initialize them then use them later on. Letting this as:

```rust
pub struct BlendAdapter<BlendService> {
    blend_relay : OutboundRelay<<BlendService as ServiceData>::Message>,
    blend_broadcast_settings: <BlendService as BlendServiceExt>::BroadcastSettings,
    _blend: PhantomData<BlendService>
}

impl BlendAdapter {
    pub fn new(realay: OutboundRelay<<BlendService as ServiceData>::Message>) -> Self { .. }
    pub fn publish_block<ClPoolItem, DaPoolItem>(&self, block: Block<ClPoolItem, DaPoolItem>) {
        if let Err((e, _)) = self.blend_relay
            .send(nomos_blend_service::message::ServiceMessage::Blend(
                nomos_blend_service::message::NetworkMessage {
                    message: wire::serialize(&messages::NetworkMessage::Block(block)).unwrap(),
                    broadcast_settings: self.blend_broadcast_settings,
                },
            ))
            .await
        {
            error!(target: LOG_TARGET, "Failed to relay block to blend service: {e:?}");
        }
    }
}
```
The idea is to keep the modularization of adapters but removing the complexity of generics for the outer layers.

youngjoon-lee: Yess. Keeping the adapter but simplifying it by using the `BlendServiceExt` is a good idea, which is one step ahead of my implementation. Although you guys have approved this PR, I just added this adapter to this PR [a4d35d0](https://github.com/logos-co/nomos/pull/1514/commits/a4d35d04d587dab44728c7d70d4a741156498af6) since it's a simple change that just substitutes the `ChainService::publish_message` method. I believe this is a better foundation to try/discuss about the approach that @ntn-x2 is suggesting.

PR #1515, opened 2025-08-04
feat(sync): reject sync protocol requests cleanly

### LINKED ISSUE #1451: Reject chainsync events in bootstrap mode
https://github.com/logos-co/nomos/pull/1428 implements how the chain service in the bootstrap mode handles events. It doesn't handle chainsync events. However, as suggested in https://github.com/logos-co/nomos/pull/1428#discussion_r2194388725, it's better to reject chainsync events instead of silently ignoring them. For that, we may need to update the response type of the chainsync events.

andrussal:

When Cryptarchia is running in Bootstrapping mode, a node doesn't process chain sync requests. Previously we just dropped these. This PR introduces failure response types for cleaner reply.

It also includes some tests refactor to simplify adding new tests.

Closes #1451 

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


nomos-services/chain-service/src/sync/block_provider.rs
```diff
@@ -488,16 +490,15 @@ where
...
-        let stream = stream::once(async move { Err(DynError::from(msg)) });
         if let Err(e) = reply_sender
-            .send(Box::pin(stream))
+            .send(ProviderResponse::Unavailable { reason })
             .await
-            .map_err(|_| GetBlocksError::SendError("Failed to send error stream".to_owned()))
+            .map_err(|_| GetBlocksError::SendError("Failed to send error response".to_owned()))
         {
-            error!("Failed to send error stream: {e}");
+            error!("Failed to send error response: {e}");
```
youngjoon-lee, line 536: Shouldn't we keep the previous implementation because the provider is available but we need to send an error (i.e. a stream that contains a single error)?

andrussal: I think both versions work here though, `cryptarchia_sync` logic handles the case when we fail to produce stream at all and also when stream fails in the middle . It seemed a bit cleaner to change it to this here because we are not creating block stream.

youngjoon-lee: I understand that the system works regardless of this change, as you said. 
Nevertheless, I think it’s not natural to return the Unavailable response in this case, even though the provider is actually available, if I understand correctly. Please let me know if I am missing something or if I misunderstood the intention of this method

youngjoon-lee: Let me rephrase my question. Is the Unavailable response meant to be used only when the chain service is in Bootstrapping mode? Or, is it for broader cases?

andrussal: I am using it as response type when requesting block stream didn't succeed. Unavailable is probably not good naming though. But it kind of made sense because it's used when block stream is not available. But in this PR context it seems to indicate that we rejected request. Maybe Failed is better naming.

youngjoon-lee: Okay, then it means that the `Unavailable` response has a broader meaning than I understood.

I have a follow-up question.

https://github.com/logos-co/nomos/blob/ee18c8ba300746411ecc5cfe59580f897e9aea2c/consensus/cryptarchia-sync/src/libp2p/provider.rs#L57-L58

Here, we're converting the`Unavailable` response (received from the remote) to the `GetTipResponse::Failure`. 
I have a question: "Why do we need to define the `Unavailable` response if it is not actually meaningful to the requester?"
Can't we just let the remote respond with the `GetTipResponse::Failure`?

andrussal: Here we receive `ProviderResponse` from the chain service and send `GetTipResponse` to peer. It would be simpler in case of GetTipResponse to return `GetTipResponse::Failure` from the service, true. But it seemed cleaner for block download request to return `ProviderResponse::Unavailable` instead of creating stream that returns error. But I am up to use this approach if you think it's better.

youngjoon-lee: Ah, From the chain service. Yeah, I was confused. Thanks.

> But it seemed cleaner for block download request to return `ProviderResponse::Unavailable` instead of creating stream that returns error. 

I don't understand this point. Why does the "request** return something? Shouldn't it receive something? I still want to understand why this approach is cleaner. Let me also dive into details again for better understanding.

youngjoon-lee: I just took another round of review, with keeping in mind your intention.

Now I understand that you wanted to make the chain service return either a success result or an error (i.e. `Unavailable`). I translated your implementation like below just to check if my understanding is correct. Could you double check if I understood it correctly? (I'm not suggesting this change right now. It doesn't compile, and I'm not 100% sure if this change is better.)

```diff
diff --git a/consensus/cryptarchia-sync/src/libp2p/behaviour.rs b/consensus/cryptarchia-sync/src/libp2p/behaviour.rs
index b985acf0..6eefb434 100644
--- a/consensus/cryptarchia-sync/src/libp2p/behaviour.rs
+++ b/consensus/cryptarchia-sync/src/libp2p/behaviour.rs
@@ -55,7 +55,8 @@ pub type BoxedStream<T> = Box<dyn futures::Stream<Item = T> + Send + Unpin>;
 
 
 type ToSwarmEvent = ToSwarm<
     <Behaviour as NetworkBehaviour>::ToSwarm,
@@ -753,8 +754,8 @@ mod tests {
     }
 
     trait ProviderBehavior: Send + 'static {
-        fn handle_tip_request(&self) -> TipResponse;
-        fn handle_blocks_request(&self, blocks_count: usize) -> BlocksResponse;
+        fn handle_tip_request(&self) -> Result<Tip, Error>;
+        fn handle_blocks_request(&self, blocks_count: usize) -> Result<BoxedStream, Error>;
     }
 
     struct StandardProvider {
@@ -762,38 +763,36 @@ mod tests {
     }
 
     impl ProviderBehavior for StandardProvider {
-        fn handle_tip_request(&self) -> TipResponse {
-            ProviderResponse::Available(Tip {
+        fn handle_tip_request(&self) -> Result<Tip, Error> {
+            Ok(Tip {
                 tip: HeaderId::from([0; 32]),
                 slot: Slot::from(0),
             })
         }
 
-        fn handle_blocks_request(&self, _requested: usize) -> BlocksResponse {
-            ProviderResponse::Available(
-                futures::stream::iter(
-                    iter::repeat_with(|| Bytes::from_static(&[0; 32]))
-                        .take(self.blocks_count)
-                        .map(Ok),
-                )
-                .boxed(),
+        fn handle_blocks_request(&self, _requested: usize) -> Result<BoxedStream, Error> {
+            Ok(futures::stream::iter(
+                iter::repeat_with(|| Bytes::from_static(&[0; 32]))
+                    .take(self.blocks_count)
+                    .map(Ok),
             )
+            .boxed())
         }
     }
 
     struct RejectingProvider;
 
     impl ProviderBehavior for RejectingProvider {
-        fn handle_tip_request(&self) -> TipResponse {
-            ProviderResponse::Unavailable {
+        fn handle_tip_request(&self) -> Result<Tip, Error> {
+            Err(Error::ChainSyncUnavailable {
                 reason: "Node is not in online mode".to_string(),
-            }
+            })
         }
 
-        fn handle_blocks_request(&self, _requested: usize) -> BlocksResponse {
-            ProviderResponse::Unavailable {
+        fn handle_blocks_request(&self, _requested: usize) -> Result<BoxedStream, Error> {
+            Err(Error::ChainSyncUnavailable {
                 reason: "Node is not in online mode".to_string(),
-            }
+            })
         }
     }
 
diff --git a/consensus/cryptarchia-sync/src/libp2p/provider.rs b/consensus/cryptarchia-sync/src/libp2p/provider.rs
index c09f261a..39e49f0a 100644
--- a/consensus/cryptarchia-sync/src/libp2p/provider.rs
+++ b/consensus/cryptarchia-sync/src/libp2p/provider.rs
@@ -51,11 +51,11 @@ impl Provider {
         })?;
 
         match response {
-            ProviderResponse::Available(tip) => {
+            Ok(tip) => {
                 send_message(peer_id, &mut libp2p_stream, &tip).await?;
             }
-            ProviderResponse::Unavailable { reason } => {
-                let response = GetTipResponse::Failure(reason);
+            Err(e) => {
+                let response = GetTipResponse::Failure(e);
                 send_message(peer_id, &mut libp2p_stream, &response).await?;
             }
         };
@@ -75,11 +75,9 @@ impl Provider {
         })?;
 
         match response {
-            ProviderResponse::Available(stream) => {
-                Self::send_blocks(peer_id, stream, libp2p_stream).await
-            }
-            ProviderResponse::Unavailable { reason } => {
-                let response = DownloadBlocksResponse::Failure(reason);
+            Ok(stream) => Self::send_blocks(peer_id, stream, libp2p_stream).await,
+            Err(e) => {
+                let response = DownloadBlocksResponse::Failure(e);
                 send_message(peer_id, &mut libp2p_stream, &response).await?;
                 Ok(())
             }

```

andrussal: "I don't understand this point. Why does the "request** return something? Shouldn't it receive something?" lol, sorry about my English.

I guess it's personal preference if we create a (`futures`) stream for error response or return a separate type (`ProviderResponse:: Unavailable`) that indicates error. Maybe "cleaner" is not best wording. For me it seemed "cleaner" because if we have no data to send, creating a (`futures`) stream to say that creating stream failed didn't feel right.

youngjoon-lee: Yes. I 100% agree that we don't need to create a stream to return only a single error. 

I think I have better understanding on this PR. Could you please check my last comment https://github.com/logos-co/nomos/pull/1515#discussion_r2254119201?

andrussal: I was considering this approach as well but wasn't sure if instantiating an Error type inside chain service that is defined in `cryptarchia_sync` module  would be an idiomatic approach... But maybe we could just use `DynError`? Using `Result` instead of custom response is better for sure.

youngjoon-lee: > but wasn't sure if instantiating an Error type inside chain service that is defined in cryptarchia_sync module would be an idiomatic approach

I see your point. I think it's not an issue because the chain service in this PR is already returning the `ProviderResponse` type defined in the `cryptarchia_sync` module. I think it's okay even if we replace the `ProviderResponse` with `Result` by defining an `Error` in the `cryptarchia-sync` module.

But, I don't want to block this PR by suggesting the way that I didn't test yet. I think we can go with the current approach since you've came up with this after much consideration.

youngjoon-lee: BTW, in near future, we may need to refactor the `cryptarchia_sync` to replace all `DynError`s with a concrete type. It's fine to use `DynError` for now because the requester doesn't check the error type yet. But, I feel we may need to branch based on error types in the future.

andrussal: > I think it's not an issue because the chain service in this PR is already returning the ProviderResponse type defined in the cryptarchia_sync module. 

My question really was if Errors defined in a module are supposed to be instantiated externally. Some things are part of public API for sure and meant to be instantiated externally as well.

youngjoon-lee: > My question really was if Errors defined in a module are supposed to be instantiated externally. Some things are part of public API for sure and meant to be instantiated externally as well.

Yes. I think it depends how you define the Error.

andrussal: > BTW, in near future, we may need to refactor the `cryptarchia_sync` to replace all `DynError`s with a concrete type. It's fine to use `DynError` for now because the requester doesn't check the error type yet. But, I feel we may need to branch based on error types in the future.

I will use DynError in this PR. I wasn't comfortable with `ProviderResponse` from the beginning. Glad that it's now clear about the whole error situation:).

youngjoon-lee: I just had another round of review before approving this PR, but I came up with another question.

Why is `BlocksResponse` not symmetric with `TipResponse`?
https://github.com/logos-co/nomos/blob/ee18c8ba300746411ecc5cfe59580f897e9aea2c/consensus/cryptarchia-sync/src/libp2p/behaviour.rs#L56-L58

The `TipResponse` contains the `GetTipResponse` is an enum that is either success or failure. However, the chain service never initiate it with a failure.
https://github.com/logos-co/nomos/blob/ee18c8ba300746411ecc5cfe59580f897e9aea2c/consensus/cryptarchia-sync/src/messages.rs#L69-L74

In contrast, `BlocksResponse` contains only the `BoxedStream`, which is only the success case.

I know that the `GetTipResponse` is for defining the output that should be written to the libp2p stream. So, it makes sense to define it as an enum which contains either a success or a failure.

But then, I think it's clearer if we define the `TipReponse` like below:
```rust
 pub type TipResponse = ProviderResponse<(HeaderId, Slot)>; 
```
(or by defining another concrete type).

When I read this code at the beginning, I focused on the tip response (not the blocks response), and I was confused why there are two ways for the chain service to specify an failure case to the `TipResponse` (by using `Unavailable` or by using `Available(GetTipResponse::Failure(..))`. But I realized that the `GetTipResponse::Failure` is never used by the chain service.
Now I think this is the beginning of all of my questions above.

If you intended to not let the chain service to initiate the `GetTipResponse::Failure`, I think we don't need to (shouldn't) let the chain service to deal with the `GetTipResponse` enum. 
What do you think?

andrussal: I think this asymmetry was here from the beginning. I remember initially we tried to avoid duplicating types and it was easy to use `GetTipResponse` in both chain service and network protocol. When I introduced `ProviderResponse`, I kept `GetTipResponse` as successful response type for Tip request but used `ProviderResponse::Unavailable` for both Tip and Block failure. 

> But then, I think it's clearer if we define the TipReponse like below:

I actually like idea of using Result as you suggested. Given that we are fine of letting "library" users to instantiate Errors types defined in the library. I just wasn't sure what is the idiomatic way here. But I am not sure if we should keep using `GetTipResponse` in chain service or not...

youngjoon-lee: Yeah, I understand the background. 

> Given that we are fine of letting "library" users to instantiate Errors types defined in the library.
 
My opinion is Yes. We can define an Error in the `cryptarchia-sync`. This Error is sent to the requester. This Error is initiated by both the chain service and the `cryptarchia-sync`. I think it's not a bad practice to let the chain service initiate the Error because that Error is part of the chain sync protocol that we want for all participants (chain service and requesters) to comply with.

But, I will approve this PR now. I understand the background and I also have some simplification idea which may affect this error propagations. I will propose them as PRs when I have time, so that we can extend discussions there. I think it will be easier. Thanks.

Issue #1461, opened 2025-07-16
Closing mutual duplicate connections


youngjoon-lee:
If two nodes open two connections with each other, so that both have incoming and outgoing connections to the same neighbor, each node should close those extra connections, as defined in the [spec](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81bc8166dcafcf534c33
)

PR #1522, opened 2025-08-06
feat(blend): ensure unique connections per peer

ntn-x2:

Built on top of https://github.com/logos-co/nomos/pull/1511 and fixes https://github.com/logos-co/nomos/issues/1461.

Main changes are in the core->core network behaviour, where we now need to keep track of the "role" of a remote peer, whether they are dialing us or we are dialing them. Then, we start rejecting duplicate connections in the same direction (i.e. we won't accept a second incoming connection from a peer from which we already have an incoming connection), and if we get a new connection in the opposite direction, we will evaluate which one to keep and which one to discard based on peer ID comparison, as per the spec.

Removing the `max_established_per_peer(1)` from the aggregated behaviour also removed the "bug" of not being able to receive more than one message at the same time from the same edge peer, which is now not a problem anymore, and it should not be indeed, since the spec does not mention it and there is no reason to limit ourselves in that regard.

Yes.

@ntn-x2 @youngjoon-lee, spec: @madxor 

There was a change in how nodes evaluate which connection to maintain open and which one to close. The spec now refers to lexicographically compare base58 representations of peer IDs, which is what we do in code as well.

Yes, see point above.

## Checklist

* [x] 2. Description added.


nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -166,207 +218,382 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
+        else {
+            // We expect not to find a connection to upgrade only if we have already
+            // processed it (e.g., we are processing a `FullyNegotiatedInbound` after we
+            // have already processed a `FullyNegotiatedOutbound` or viceversa).
+            // Otherwise, this is an inconsistency and we should panic.
+            assert!(
+                self.negotiated_peers.contains_key(&peer_id),
+                "Negotiated connection {connection_id:?} with peer {peer_id:?} not found in storage of pending connections."
+            );
+            return;
```
youngjoon-lee, line 310: Question:
I'm trying to understand the `e.g.` case you provided.
When we process a `FullyNegotiatedOutbound`, do we remove all the connections, which have the same `PeerId`, from the `connections_waiting_upgrade`? If so, I understand that there is a case where this `assert!` is true. 
But, if we don't clean up the `connections_waiting_upgrade` proactively, I think this `assert!` will be always false, and we can replace it with a `panic!`.

It seems that the following code is the only the other point that removes elements from `connections_waiting_upgrade`:
https://github.com/logos-co/nomos/blob/49bacfa2612f81021979c8a087fb5e0740385df3/nomos-blend/network/src/core/with_core/behaviour/mod.rs#L720-L729

If so, does the `assert!` is for the case where we had already requested closing a certain connection (e.g. as a result of deduplication) but a `FullyNegotiated...` event has been received **before** the Swarm closes the connection? If my explanation is correct, it makes sense to have this `assert!`.

Could you confirm this?

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -166,207 +218,382 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
+        &mut self,
+        (peer_id, connection_id): (PeerId, ConnectionId),
+        role: Endpoint,
+    ) {
         // We need to check if we still have available connection slots, as it is
         // possible, especially upon session transition, that more than the maximum
         // allowed number of peers are trying to connect to us. So once the stream is
         // actually upgraded, we downgrade it again if we do not have space left for it.
         // By not adding the new connection to the map of negotiated peers, the swarm
         // will not be notified about this dropped connection, which is what we want.
```
youngjoon-lee, line 344: Question: Is this comment still valid? It seems that we don't notify the Swarm even when we add the new connection to the `self.negotiated_peers`. I guess we can just remove this sentence, if I understand the logic correctly?

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -166,207 +218,382 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
+    /// Handle a newly upgraded connection for a peer that this peer is already
+    /// connected to.
+    ///
+    /// Depending on the outcome of comparing the two peers' IDs, either the
+    /// existing connection is replaced with the new one, or the new one is
+    /// discarded in favor of the existing one.
+    ///
+    /// # Panics
+    ///
+    /// If there is no established connection for the given peer in the relative
```
youngjoon-lee:
```suggestion
    /// If there is no negotiated connection for the given peer in the relative
```
I think this change helps us avoiding confusions between `self.connections_waiting_upgrade` and `self.negotiated_peers`. Wdyt?

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -166,207 +218,382 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
+
+        if should_close_established {
+            tracing::trace!(target: LOG_TARGET, "Replacing established connection {:?} with peer {peer_id:?} with upgraded connection {new_connection_id:?}.", existing_connection_details.connection_id);
+            let existing_connection = (peer_id, existing_connection_details.connection_id);
+            // Modify the `negotiated_peers` storage directly so
+            // that when the old connection is dropped, the swarm is
+            // not notified.
+            update_connection_id_and_direction(existing_connection_details, new_connection_id);
+            // Notify the current connection handler to drop the substreams.
+            self.close_connection(existing_connection);
```
youngjoon-lee, line 447: Question (related to https://github.com/logos-co/nomos/pull/1522/files#r2261975826):

I'd like to ask if the comment `the swarm is not notified` is valid. Am I missing any logic that notifies the Swarm (automatically) when an entry is inserted or removed from the `self.negotiated_peers`?

The actual code looks good. My question is only about the comment.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: Ah, I'm revisiting this after reading the `on_swarm_event` method below. 

So, I guess, the comment is mentioning about not notifying the Swarm back with the `Event::PeerDisconnected` after this ignored connection will be actually closed by the Swarm? If so, I think this comment is valid.

https://github.com/logos-co/nomos/blob/49bacfa2612f81021979c8a087fb5e0740385df3/nomos-blend/network/src/core/with_core/behaviour/mod.rs#L720-L746

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: I'm revisiting this after reading the `on_swarm_event` method below. Please see the https://github.com/logos-co/nomos/pull/1522/files#r2261998166. I left details that I understood. If my understanding is correct, I think the comment `the swarm is not notified` is also valid.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
ntn-x2: No, the point here is that the connection handler generates two events for the same connection: `FullyNegotiatedInbound` and `FullyNegotiatedOutbound`. When either of them is generated, we consider the connection upgraded, hence when we receive the second one, we expect not to find the connection in the set of pending connections. The `on_swarm_event`, instead, is for cases where the connection is closed before even having a chance to upgrade any of its substreams, e.g., if there's a problem with the remote peer. In that case, since the connection has never been negotiated, we simply "forget" about it and we don't notify the swarm about it. In an upcoming PR about dial retires, we do actually notify the swarm in this case and only for outbound connections, so that the swarm can try a re-dial or, in case of maximum attempts, switch to a different peer: https://github.com/logos-co/nomos/pull/1524/files#diff-2a6abfe9a1ac517407672a6214af0382d69d53489549213c5a41f927b07c5dbdR750.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
ntn-x2: Yes, your understanding is correct. This connection attempt is invisible to the swarm since we never managed to know whether the remote peer even speaks Blend at all.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
ntn-x2: Yes in this case the same logic applies. We are replacing one connection with the other, but semantically it's the same connection, just the reverse direction. Indeed, we only change direction and connection ID, we do not update the negotiated peer state, which we maintain. So the old connection that gets dropped is not really something the swarm should be concerned since we are still connected to the same peer that maintains the same status as before.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
ntn-x2: Agreed! https://github.com/logos-co/nomos/pull/1522/commits/02a70fafcc1fc4a83e4d1252249697fd0de1dfd4

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: > No, the point here is that the connection handler generates two events for the same connection: FullyNegotiatedInbound and FullyNegotiatedOutbound. When either of them is generated, we consider the connection upgraded, hence when we receive the second one, we expect not to find the connection in the set of pending connections. 

Oh, yeah. I now understand this `assert!`. Yes, the connection handler generate **two** events for **the same** connection. Thus, this `assert!` is not related to the `on_swarm_event` method that I referred to. Thank you for your explanation.

PR #1525, opened 2025-08-07
feat(da-network): request historic sampling

pradovic:

> [!NOTE]  
> Review the 

> [!IMPORTANT]  
> explained.  

> [!IMPORTANT]  
> advancing  
> the review.

> [!IMPORTANT]

> [!IMPORTANT]
    modification/solutions.

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/nomos-services/data-availability/network/src/backends/mod.rs
+++ b/nomos-services/data-availability/network/src/backends/mod.rs
@@ -1,10 +1,13 @@
 pub mod libp2p;
 pub mod mock;
 
-use std::pin::Pin;
+use std::{collections::HashMap, pin::Pin};
 
+use ::libp2p::PeerId;
 use futures::Stream;
-use nomos_da_network_core::addressbook::AddressBookHandler;
+use multiaddr::Multiaddr;
+use nomos_core::{block::BlockNumber, da::BlobId};
+use nomos_da_network_core::{addressbook::AddressBookHandler, protocols::sampling::SubnetsConfig};
 use overwatch::{overwatch::handle::OverwatchHandle, services::state::ServiceState};
 use subnetworks_assignations::MembershipHandler;
 
@@ -25,6 +28,7 @@ pub trait NetworkBackend<RuntimeServiceId> {
         overwatch_handle: OverwatchHandle<RuntimeServiceId>,
         membership: Self::Membership,
         addressbook: Self::Addressbook,
+        subnets_settings: SubnetsConfig,
     ) -> Self;
     fn shutdown(&mut self);
     async fn process(&self, msg: Self::Message);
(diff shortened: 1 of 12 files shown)


nomos-services/data-availability/network/src/lib.rs
```diff
@@ -462,6 +496,103 @@ where
...
+            );
+
+            return None;
+        }
+
+        let mut selected_peers = HashMap::new();
+        let mut rng = rand::thread_rng();
+
+        // Select `subnets_to_sample` random subnetworks from `all_subnet_ids`
+        // this selection allows repeated subnetworks to be selected
```
pradovic: If i understood correctly from the spec, we need to select 20 random subnetworks. The code in spec seem to allow repetition of choices, so I have implemented it here as well.  Alternative would be to just use choices() method which returns unique choices.
 
@bacv @danielSanchezQ what do you guys think?

bacv: Yes, this is valid assumption and the same is implemented inside the sampling behaviour. One thing missing here is filtering out local peer id, we should be able to get it from the backend.

pradovic: Good catch. This should be the fix: https://github.com/logos-co/nomos/pull/1525/commits/b69ac8163e1d96897f81fe1fd1a67c8222c33fe0.

bacv (approved):

nomos-services/data-availability/network/src/lib.rs
pradovic: I have also realized just now that providing hashmap is not good as peerids might repeat, so I have switched to a vector instead: https://github.com/logos-co/nomos/pull/1525/commits/11affa16e043bf6bbf7691aa8829f06d1088c963.

nomos-services/data-availability/network/src/backends/mod.rs
```diff
@@ -25,11 +28,19 @@ pub trait NetworkBackend<RuntimeServiceId> {
...
         &mut self,
         event: Self::EventKind,
     ) -> Pin<Box<dyn Stream<Item = Self::NetworkEvent> + Send>>;
+    async fn start_historic_sampling(
+        &self,
+        block_number: BlockNumber,
+        blob_id: BlobId,
+        membership: Vec<(PeerId, Multiaddr)>,
+    );
+    fn local_peer_id(&self) -> PeerId;
```
danielSanchezQ: This is not the API this should have. Ideally we should pass here the set of blob_ids for a block and the membership. Nothing else. This is not ready yet in the underlaying behaviours/swarm. So not sure if we should stall this meanwhile.

nomos-services/data-availability/network/src/lib.rs
```diff
@@ -462,6 +496,108 @@ where
...
+            } else {
+                tracing::error!(
+                "No addresses found for peer {selected_peer:?} in addressbook at block {block_number}"
+            );
+                return None;
+            }
+        }
+
+        Some(selected_peers)
+    }
```
danielSanchezQ: This will be handled by the historic behaviour as the other behaviours too.

danielSanchezQ:
@bacv This is not what we discussed about. We agreed on having a behaviour that would deal with the connections. Let's rediscuss in discord.

PR #1529, opened 2025-08-08
feat(da-network): cumulative in-memory addressbook 

pradovic:
This PR keeps the list of all known latest addresses for all peers in memory. It is a precondition for the https://github.com/logos-co/nomos/pull/1525.

### DIFF
--- a/nomos-services/data-availability/network/src/addressbook.rs
+++ b/nomos-services/data-availability/network/src/addressbook.rs
@@ -19,14 +19,12 @@ impl AddressBookHandler for AddressBook {
     }
 }
 
-#[async_trait::async_trait]
 pub trait AddressBookMut: AddressBookHandler {
-    async fn update(&self, new_peers: AddressBookSnapshot<Self::Id>);
+    fn update(&self, new_peers: AddressBookSnapshot<Self::Id>);
 }
 
-#[async_trait::async_trait]
 impl AddressBookMut for AddressBook {
-    async fn update(&self, new_peers: AddressBookSnapshot<Self::Id>) {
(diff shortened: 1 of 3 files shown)


nomos-services/data-availability/network/src/addressbook.rs
```diff
@@ -6,44 +6,29 @@ use nomos_da_network_core::addressbook::AddressBookHandler;
...
     }
 }
 
-impl AddressBookMut for AddressBook {
-    fn update(&self, new_peers: AddressBookSnapshot<Self::Id>) {
-        self.peers.store(Arc::new(new_peers));
-    }
+#[async_trait::async_trait]
+pub trait AddressBookMut: AddressBookHandler {
+    async fn update(&self, new_peers: AddressBookSnapshot<Self::Id>);
```
danielSanchezQ: Why is it async now?

danielSanchezQ (approved):

nomos-services/data-availability/network/src/addressbook.rs
pradovic: It does not have to be, but it is called from the async context so I thought it can be async as well. I can make it sync again if you think it is beneficial?

danielSanchezQ: it was just curiosity. But if it is not needed for now it doesn't need to be async.

pradovic: It does not have to be, I will switch it back to sync then

pradovic: Fixed: https://github.com/logos-co/nomos/pull/1529/commits/9c02e3a89a711033c89bbe178e3be4cb41de8ebf.

PR #1541, opened 2025-08-13
chore: Remove unnecessary heap allocation

c410-f3r:

It is not necessary to call `ToString::to_string` when passing a parameter to one of the formatting macros.


nomos-services/utils/src/overwatch/status/service_status_entry/error.rs
```diff
@@ -12,13 +12,15 @@ pub struct ServiceStatusEntriesError<RuntimeServiceId: Display> {
...
-            .collect();
-        let entries = entries.join(", ");
-        write!(f, "ServiceStatuses: {entries}")
+        let mut iter = self.service_status_entries.iter();
+        write!(f, "ServiceStatuses: [")?;
+        if let Some(elem) = iter.next() {
+            write!(f, "{elem}")?;
+        }
+        for elem in iter {
+            write!(f, ",{elem}")?;
```
danielSanchezQ:
```suggestion
            write!(f, ", {elem}")?;
```

danielSanchezQ (commented):
Just one comment. But nice catch. Thanks.

PR #1544, opened 2025-08-14
feat(nat

andrussal:

This PR adds support for `NAT-PMP` mapping protocol. 

---
We also need to add mapping lease renewal, I will do this another PR.

Please see [spec](https://www.notion.so/nomos-tech/Nomos-P2P-Network-NAT-Solution-206261aa09df81c2a8c2ecc0f9365664) and base PR: https://github.com/logos-co/nomos/pull/1391

@andrussal 

Yes

Yes

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-libp2p/src/behaviour/nat/address_mapper/nat_pmp.rs
+++ b/nomos-libp2p/src/behaviour/nat/address_mapper/nat_pmp.rs
@@ -9,6 +9,8 @@ use crate::{
     config::NatMappingSettings,
 };
 
+type PortNumber = u16;
+
 pub struct NatPmp {
     settings: NatMappingSettings,
     nat_pmp: NatpmpAsync<UdpSocket>,
@@ -18,7 +20,7 @@ impl NatPmp {
     async fn send_map_request(
         &self,
         protocol: Protocol,
-        port: u16,
+        port: PortNumber,
     ) -> Result<(), AddressMapperError> {
         self.nat_pmp
             .send_port_mapping_request(protocol, port, port, self.settings.lease_duration)
@@ -26,7 +28,7 @@ impl NatPmp {
             .map_err(|e| AddressMapperError::PortMappingFailed(e.to_string()))
     }
 
-    async fn recv_map_response_public_port(&self) -> Result<u16, AddressMapperError> {
+    async fn recv_map_response_public_port(&self) -> Result<PortNumber, AddressMapperError> {
         match self
             .nat_pmp
             .read_response_or_retry()
@@ -101,7 +103,9 @@ impl MappingProtocol for NatPmp {
     }
 }
 
-fn extract_port_and_protocol(addr: &Multiaddr) -> Result<(u16, Protocol), AddressMapperError> {
+fn extract_port_and_protocol(
+    addr: &Multiaddr,
+) -> Result<(PortNumber, Protocol), AddressMapperError> {
     addr.iter()
         .find_map(|p| match p {
             MaProto::Tcp(p) => Some((p, Protocol::TCP)),
@@ -118,7 +122,7 @@ fn extract_port_and_protocol(addr: &Multiaddr) -> Result<(u16, Protocol), Addres
 fn build_public_address(
     internal: &Multiaddr,
     public_ip: Ipv4Addr,
-    public_port: u16,
+    public_port: PortNumber,
 ) -> Result<Multiaddr, AddressMapperError> {
     let with_ip = internal
         .replace(0, |_| Some(multiaddr::Protocol::Ip4(public_ip)))


nomos-libp2p/src/behaviour/nat/address_mapper/protocol.rs
```diff
@@ -1,12 +1,17 @@
...
+    },
+    config::NatMappingSettings,
+};
 
 /// Trait for NAT mapping protocols
 #[async_trait::async_trait]
 pub trait MappingProtocol: Send + Sync + 'static {
     /// Initialize the protocol
-    async fn initialize() -> Result<Box<Self>, AddressMapperError>
+    async fn initialize(settings: NatMappingSettings) -> Result<Box<Self>, AddressMapperError>
```
ntn-x2, line 14: Mhh should we not follow the approach overwatch takes? How do we know what settings the mapping protocol needs? We need to change the `NatMappingSettings` type whenever we add a new protocol?

ntn-x2: I'm also ok leaving it as it is for time reasons, but wanted to make sure this is a calculated thing.

nomos-libp2p/src/behaviour/nat/address_mapper/protocol.rs
```diff
@@ -15,21 +20,32 @@ pub trait MappingProtocol: Send + Sync + 'static {
...
     upnp: Box<dyn MappingProtocol>,
 }
 
 impl ProtocolManager {
-    pub async fn initialize() -> Result<Self, AddressMapperError> {
-        let upnp = UpnpProtocol::initialize().await?;
-        Ok(Self { upnp })
+    pub async fn initialize(settings: NatMappingSettings) -> Result<Self, AddressMapperError> {
+        let upnp = UpnpProtocol::initialize(settings).await?;
+        let nat_pmp = NatPmp::initialize(settings).await?;
```
ntn-x2: Do we need to initialize all the mapping protocols? Would it not make more sense to initialize each protocol when it's needed?

nomos-libp2p/src/behaviour/nat/address_mapper/protocol.rs
```diff
@@ -15,21 +20,32 @@ pub trait MappingProtocol: Send + Sync + 'static {
...
     }
 
     pub async fn try_map_address(
         &mut self,
         address: &Multiaddr,
     ) -> Result<Multiaddr, AddressMapperError> {
-        let external_address = self.upnp.map_address(address).await?;
+        if let Ok(external_address) = self.nat_pmp.map_address(address).await {
+            tracing::info!(
+                "Successfully mapped {} to {} using NAT-PMP",
```
ntn-x2:
```suggestion
                "Successfully mapped {address} to {external_address} using NAT-PMP",
```

nomos-libp2p/src/behaviour/nat/address_mapper/protocol.rs
```diff
@@ -15,21 +20,32 @@ pub trait MappingProtocol: Send + Sync + 'static {
...
+                address,
+                external_address
+            );
+
+            return Ok(external_address);
+        }
 
+        let external_address = self.upnp.map_address(address).await?;
         tracing::info!(
             "Successfully mapped {} to {} using {}",
```
ntn-x2:
```suggestion
            "Successfully mapped {address} to {external_address} using UPnP",
```

nomos-libp2p/src/behaviour/nat/address_mapper/protocol.rs
```diff
@@ -40,3 +56,36 @@ impl ProtocolManager {
...
+#[cfg(test)]
+mod real_gateway_tests {
+    use tokio::runtime::Runtime;
+
+    use super::*;
+
+    #[test]
+    #[ignore = "Runs against a real gateway"]
+    fn map_address_via_protocol_manager_real_gateway() {
+        let rt = Runtime::new().unwrap();
```
ntn-x2: What about `spawn_blocking`?

nomos-libp2p/src/behaviour/nat/address_mapper/protocol.rs
```diff
@@ -40,3 +56,36 @@ impl ProtocolManager {
...
+
+            let settings = NatMappingSettings::default();
+
+            let mut mgr = ProtocolManager::initialize(settings)
+                .await
+                .expect("initialize ProtocolManager");
+
+            let external = mgr.try_map_address(&internal_addr).await.unwrap();
+
+            let expected_external: Multiaddr =
```
ntn-x2: I don't think it makes sense to leave this test as it is. If anything, we should probably just output the result of the mapping process, without asserting anything since it will never be replicable.

nomos-libp2p/src/behaviour/nat/address_mapper/nat_pmp.rs
```diff
@@ -0,0 +1,158 @@
...
+            .map_err(|e| AddressMapperError::PortMappingFailed(e.to_string()))?
+        {
+            Response::UDP(r) | Response::TCP(r) => Ok(r.public_port()),
+            Response::Gateway(_) => Err(AddressMapperError::PortMappingFailed(
+                "Expected UDP/TCP mapping response; got Gateway response".to_owned(),
+            )),
+        }
+    }
+
+    async fn query_public_ip(&mut self) -> Result<Ipv4Addr, AddressMapperError> {
```
ntn-x2, line 45: As far as I can see this only works with IPv4? Should we mention that?

nomos-libp2p/src/behaviour/nat/address_mapper/nat_pmp.rs
```diff
@@ -0,0 +1,158 @@
...
+            _ => None,
+        })
+        .ok_or_else(|| {
+            AddressMapperError::MultiaddrParseError(
+                "No TCP or UDP port found in multiaddr".to_owned(),
+            )
+        })
+}
+
+pub fn public_address(
```
ntn-x2: Why `pub`?

nomos-libp2p/src/behaviour/nat/address_mapper/nat_pmp.rs
```diff
@@ -0,0 +1,158 @@
...
+    internal: &Multiaddr,
```
ntn-x2, line 123: Should we take a `Multiaddr` instead of a reference to give a hint the old one is not meant to be used anymore? Although `Multiaddr` implements `Copy`, I think it would still be a better design.

ntn-x2: Or maybe call this function `to_public`.

nomos-libp2p/src/behaviour/nat/address_mapper/nat_pmp.rs
```diff
@@ -0,0 +1,158 @@
...
+    #[ignore = "Needs real NAT-PMP gateway to test"]
+    async fn test_nat_pmp() {
+        let mut natpmp = NatPmp::initialize(NatMappingSettings {
+            lease_duration: 3600,
+            timeout: std::time::Duration::from_secs(5),
+        })
+        .await
+        .unwrap();
+
+        let addr: Multiaddr = "/ip4/127.0.0.1/tcp/54321".parse().unwrap();
```
ntn-x2: Same as my other comment. This test has very little meaning if we leave it like this. Either we mock it, if possible, or we simply leave it to be run and capture its output.

ntn-x2 (commented):
Looks good! Left some comments.

nomos-libp2p/src/behaviour/nat/address_mapper/protocol.rs
andrussal: > I'm also ok leaving it as it is for time reasons, but wanted to make sure this is a calculated thing.

Here yes I was assuming that these settings are common nat mapping settings, maybe in future if we see that protocols diverge, we could split or make generic.

nomos-libp2p/src/behaviour/nat/address_mapper/protocol.rs
andrussal: Good point, not reason to force initialize if not used, I will update.

nomos-libp2p/src/behaviour/nat/address_mapper/protocol.rs
andrussal: > I don't think it makes sense to leave this test as it is. If anything, we should probably just output the result of the mapping process, without asserting anything since it will never be replicable.

Makes sense, I will make it more usable generally.

nomos-libp2p/src/behaviour/nat/address_mapper/nat_pmp.rs
andrussal: > As far as I can see this only works with IPv4? Should we mention that?

nat-pmp as mapping protocol only supports ipv4

nomos-libp2p/src/behaviour/nat/address_mapper/nat_pmp.rs
```diff
@@ -0,0 +1,158 @@
...
+        protocol: Protocol,
+        port: u16,
+    ) -> Result<(), AddressMapperError> {
+        self.nat_pmp
+            .send_port_mapping_request(protocol, port, port, self.settings.lease_duration)
+            .await
+            .map_err(|e| AddressMapperError::PortMappingFailed(e.to_string()))
+    }
+
+    async fn recv_map_response_public_port(&self) -> Result<u16, AddressMapperError> {
```
danielSanchezQ: Ca we have a typealias for port number?
`type PortNumber = u16;`

nomos-libp2p/src/behaviour/nat/address_mapper/nat_pmp.rs
danielSanchezQ: It doesnt matter, Multiaddr is copy iirc. Although it would be nice to consume it.

nomos-libp2p/src/behaviour/nat/address_mapper/protocol.rs
```diff
@@ -40,3 +56,36 @@ impl ProtocolManager {
...
+
+#[cfg(test)]
+mod real_gateway_tests {
+    use tokio::runtime::Runtime;
+
+    use super::*;
+
+    #[test]
+    #[ignore = "Runs against a real gateway"]
+    fn map_address_via_protocol_manager_real_gateway() {
```
danielSanchezQ: This could be a `tokio::test` ? There is no need for manual runtime creation I think

PR #1545, opened 2025-08-14
feat(da-network): historic sampling behaviour

pradovic:

> [!NOTE]  
> Review the 

> [!IMPORTANT]  
> explained.  

> [!IMPORTANT]  
> advancing  
> the review.

> [!IMPORTANT]

> [!IMPORTANT]
    modification/solutions.

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/nomos-da/network/core/src/protocols/sampling/historic/mod.rs
+++ b/nomos-da/network/core/src/protocols/sampling/historic/mod.rs
@@ -1,0 +1,30 @@
+use kzgrs_backend::common::share::{DaLightShare, DaSharesCommitments};
+use nomos_core::{block::BlockNumber, da::BlobId, header::HeaderId};
+use subnetworks_assignations::SubnetworkId;
+
+use crate::protocols::sampling::errors::SamplingError;
+
+pub mod request_behaviour;
+
+// Enhanced events that include block context
+#[derive(Debug)]
+pub enum HistoricSamplingEvent {
+    SamplingSuccess {
+        block_number: BlockNumber,
+        block_id: HeaderId,
+        blob_id: BlobId,
+        subnetwork_id: SubnetworkId,
+        light_share: Box<DaLightShare>,
+    },
+    CommitmentsSuccess {
+        block_number: BlockNumber,
+        block_id: HeaderId,
+        blob_id: BlobId,
+        commitments: Box<DaSharesCommitments>,
+    },
+    SamplingError {
+        block_number: BlockNumber,
+        block_id: HeaderId,
+        error: SamplingError,
+    },
+}
--- a/nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
+++ b/nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
@@ -1,0 +1,620 @@
+use std::{
+    collections::{HashMap, VecDeque},
+    task::{Context, Poll, Waker},
+};
+
+use either::Either;
+use futures::{
+    future::BoxFuture,
+    stream::{BoxStream, FuturesUnordered},
+    AsyncWriteExt as _, FutureExt as _, StreamExt as _,
+};
+use libp2p::{
+    core::{transport::PortUse, Endpoint},
+    swarm::{
+        dial_opts::DialOpts, ConnectionDenied, ConnectionId, DialFailure, FromSwarm,
+        NetworkBehaviour, THandler, THandlerInEvent, THandlerOutEvent, ToSwarm,
+    },
+    Multiaddr, PeerId,
+};
+use libp2p_stream::Control;
+use nomos_core::{block::BlockNumber, da::BlobId, header::HeaderId};
+use nomos_da_messages::{sampling, sampling::SampleResponse};
+use rand::{rngs::ThreadRng, seq::IteratorRandom as _};
+use subnetworks_assignations::MembershipHandler;
+use tokio::sync::mpsc::{self, UnboundedSender};
+use tokio_stream::wrappers::UnboundedReceiverStream;
+use tracing::warn;
+
+use crate::{
+    addressbook::AddressBookHandler,
+    protocol::SAMPLING_PROTOCOL,
+    protocols::sampling::{
+        connections::Connections,
+        errors::SamplingError,
+        historic::HistoricSamplingEvent,
+        streams::{self, SampleStream},
+        SampleFutureError, SampleResponseFutureSuccess, SubnetsConfig,
+    },
+    swarm::validator::SampleArgs,
+    SubnetworkId,
+};
+
+// Context for tracking sampling requests with block information
+#[derive(Debug, Clone)]
+struct SamplingContext<Membership> {
+    block_number: BlockNumber,
+    block_id: HeaderId,
+    peer_id: PeerId,
+    subnetwork_id: SubnetworkId,
+    blob_id: BlobId,
+    membership: Membership,
+}
+
+// Future that includes context with the original sampling future result
+type HistoricSamplingResponseFuture<Membership> = BoxFuture<
+    'static,
+    (
+        SamplingContext<Membership>,
+        Result<SampleResponseFutureSuccess, SampleFutureError>,
+    ),
+>;
+
+/// Historic sampling protocol that uses membership snapshots
+/// Takes care of sending sampling requests using provided historic membership
+pub struct HistoricRequestSamplingBehaviour<Membership, Addressbook>
+where
+    Membership: MembershipHandler,
+    Addressbook: AddressBookHandler,
+{
+    /// Self peer id
+    local_peer_id: PeerId,
+    /// Underlying stream behaviour
+    stream_behaviour: libp2p_stream::Behaviour,
+    /// Underlying stream control
+    control: Control,
+    /// Pending sampling stream tasks with their context
+    stream_tasks: FuturesUnordered<HistoricSamplingResponseFuture<Membership>>,
+    /// Pending blobs that need to be sampled from `PeerId` with context
+    to_sample: HashMap<PeerId, VecDeque<(SamplingContext<Membership>, sampling::SampleRequest)>>,
+    /// Queue of sampling contexts that still need a peer for sampling
+    to_retry: VecDeque<SamplingContext<Membership>>,
+    /// Sample streams that have no tasks and should be closed
+    to_close: VecDeque<SampleStream>,
+    /// Addressbook used for getting addresses of peers
+    addressbook: Addressbook,
+    /// Pending samples for historic requests sender
+    historic_request_sender: UnboundedSender<SampleArgs<Membership>>,
+    /// Pending samples for historic requests stream
+    historic_request_stream: BoxStream<'static, SampleArgs<Membership>>,
+    /// Subnets sampling config that is used when picking new subnetwork peers
+    subnets_config: SubnetsConfig,
+    /// Pending subnetwork connections for randomly selected peers
+    connections: Connections,
+    /// Waker for sampling polling
+    waker: Option<Waker>,
+}
+
+impl<Membership, Addressbook> HistoricRequestSamplingBehaviour<Membership, Addressbook>
+where
+    Membership: MembershipHandler + Clone + Send + Sync + 'static,
+    Membership::NetworkId: Send,
+    Addressbook: AddressBookHandler + 'static,
+{
+    pub fn new(
+        local_peer_id: PeerId,
+        addressbook: Addressbook,
+        subnets_config: SubnetsConfig,
+    ) -> Self {
+        let stream_behaviour = libp2p_stream::Behaviour::new();
+        let control = stream_behaviour.new_control();
+
+        let stream_tasks = FuturesUnordered::new();
+        let to_sample = HashMap::new();
+        let to_retry = VecDeque::new();
+        let to_close = VecDeque::new();
+
+        let (historic_request_sender, receiver) = mpsc::unbounded_channel();
+        let historic_request_stream = UnboundedReceiverStream::new(receiver).boxed();
+
+        let connections = Connections::new(subnets_config.shares_retry_limit);
+
+        Self {
+            local_peer_id,
+            stream_behaviour,
+            control,
+            stream_tasks,
+            to_sample,
+            to_retry,
+            to_close,
+            addressbook,
+            historic_request_sender,
+            historic_request_stream,
+            subnets_config,
+            connections,
+            waker: None,
+        }
+    }
+
+    /// Open a new stream from the underlying control to the provided peer
+    async fn open_stream(
+        peer_id: PeerId,
+        mut control: Control,
+    ) -> Result<SampleStream, SamplingError> {
+        let stream = control
+            .open_stream(peer_id, SAMPLING_PROTOCOL)
+            .await
+            .map_err(|error| SamplingError::OpenStream { peer_id, error })?;
+        Ok(SampleStream { stream, peer_id })
+    }
+
+    /// Get a hook to the sender channel for historic sampling requests
+    pub fn historic_request_channel(&self) -> UnboundedSender<SampleArgs<Membership>> {
+        self.historic_request_sender.clone()
+    }
+
+    pub fn try_wake(&mut self) {
+        if let Some(waker) = self.waker.take() {
+            waker.wake();
+        }
+    }
+}
+
+impl<Membership, Addressbook> HistoricRequestSamplingBehaviour<Membership, Addressbook>
+where
+    Membership:
+        MembershipHandler<Id = PeerId, NetworkId = SubnetworkId> + Clone + Send + Sync + 'static,
+    Addressbook: AddressBookHandler<Id = PeerId> + 'static,
+{
+    /// Schedule sampling tasks for blobs using the provided historic membership
+    fn sample_historic(&mut self, sample_args: SampleArgs<Membership>) {
+        let (blob_ids, block_number, block_id, membership) = sample_args;
+
+        // Get random subset of subnetworks for sampling
+        let mut rng = rand::thread_rng();
+        let subnets: Vec<SubnetworkId> = (0..membership.last_subnetwork_id())
+            .choose_multiple(&mut rng, self.subnets_config.num_of_subnets);
+
+        for blob_id in blob_ids {
+            // Sample shares from multiple subnetworks
+            for subnetwork_id in &subnets {
+                let membership = membership.clone();
+                if let Some(peer_id) =
+                    self.pick_subnetwork_peer(*subnetwork_id, &membership, &mut rng)
+                {
+                    let context = SamplingContext {
+                        block_number,
+                        block_id,
+                        peer_id,
+                        subnetwork_id: *subnetwork_id,
+                        blob_id,
+                        membership: membership.clone(),
+                    };
+
+                    let control = self.control.clone();
+                    let sample_request =
+                        sampling::SampleRequest::new_share(blob_id, *subnetwork_id);
+
+                    let with_dial_task: HistoricSamplingResponseFuture<Membership> = async move {
+                        // If we don't have an existing connection to the peer, this will
+                        // immediately return `ConnectionReset` io error. To
+                        // handle that, we need to queue `sample_request`
+                        // for a peer and try again when the connection is
+                        // established.
+                        match Self::open_stream(peer_id, control).await {
+                            Ok(stream) => {
+                                let result = streams::stream_sample(stream, sample_request).await;
+                                (context, result)
+                            }
+                            Err(err) => (context, Err((err, None))),
+                            // todo: try to connect when the connection fails
+                            // if the connection fails try up to 5 peers
+                        }
+                    }
+                    .boxed();
+
+                    self.stream_tasks.push(with_dial_task);
+                    self.connections
+                        .register_pending_peer(peer_id, *subnetwork_id);
+                } else {
+                    warn!("Subnetwork {subnetwork_id} has no peers for block {block_number}");
+                }
+            }
+        }
+    }
+
+    // todo: sample commitments
+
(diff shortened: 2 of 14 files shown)


nomos-da/network/core/src/protocols/sampling/historic/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+// Enhanced events that include block context
+#[derive(Debug)]
+pub enum HistoricSamplingEvent {
+    SamplingSuccess {
+        block_number: BlockNumber,
+        block_id: HeaderId,
+        blob_id: BlobId,
+        subnetwork_id: SubnetworkId,
+        light_share: Box<DaLightShare>,
+    },
```
danielSanchezQ: This is wrong. You query a pack you get returned a pack. A single blob id is not a sample success in this case. We are sampling a block with all the shares inside. Also there is no need for the blob_id, I think it comes with the share.

nomos-da/network/core/src/protocols/sampling/historic/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+    CommitmentsSuccess {
+        block_number: BlockNumber,
+        block_id: HeaderId,
+        blob_id: BlobId,
+        commitments: Box<DaSharesCommitments>,
+    },
```
danielSanchezQ: Likewise.

Ideally we have a single event:

```rust
SamplingSucess {
    blob_id: HeaderId,
    commitments: Box<DaSharesCommitments>,
    shares: Vec<DaLightShare>,
}
```

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
```diff
@@ -0,0 +1,620 @@
...
+                    .boxed();
+
+                    self.stream_tasks.push(with_dial_task);
+                    self.connections
+                        .register_pending_peer(peer_id, *subnetwork_id);
+                } else {
+                    warn!("Subnetwork {subnetwork_id} has no peers for block {block_number}");
+                }
+            }
+        }
```
danielSanchezQ: I think this is not the way to go. All blob_ids need to be sampled. You are trying to schedule per blob_id instead of per peer. At the end, you will select N peers which translate in N streams, for each stream you query all pertienent shares for each blob id. It will be much simpler if you do a single future handling all, instead possible hundreds of futures for this.

pradovic: I have changed implementation to spawn a future per subnet: https://github.com/logos-co/nomos/pull/1545/commits/12cfe01019624dcf61f2ab511c62e2c9cf5c71e7.

If this approach makes sense what do you say that I wrap this up without connection, commitments and retries. And then open smaller PRs for the rest:
1. Commitments
2. Error handling
3. Dial connections 
4. Retry with reconnecting to up to 5 peers

nomos-da/network/core/src/protocols/sampling/historic/mod.rs
pradovic: I agree. But if we go by the events, I am not sure if we should add something like `request_id` (random uuid) that is supplied by the caller, so the caller can distinguish the results.
Another approach would be to supply return channel so we can send results to that instead of event. We could also support both

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
pradovic: You are right: 💯  I have changed implementation to spawn a future per subnet: https://github.com/logos-co/nomos/pull/1545/commits/12cfe01019624dcf61f2ab511c62e2c9cf5c71e7. I hope this is better.

If this approach makes sense what do you say that I wrap this up without connection, commitments and retries. And then open smaller PRs (does not need to be PR per each) for the rest:
1. ~~Commitments~~
2. ~~Error handling~~ 
3. ~~Retry with reconnecting to up to 5 peers~~
4. Tests

pradovic: Added commitments as well, and refactored a little bit: https://github.com/logos-co/nomos/pull/1545/commits/6997008071b7634a71bb5b2d3a739c4399357436.

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
```diff
@@ -0,0 +1,432 @@
...
+        self.historic_request_tasks.push(request_future);
+    }
+
+    async fn sample_all_shares(
+        subnets: Vec<SubnetworkId>,
+        membership: &Membership,
+        local_peer_id: &PeerId,
+        blob_ids: &HashSet<BlobId>,
+        control: &Control,
+    ) -> Result<Vec<DaLightShare>, Box<dyn std::error::Error>> {
```
bacv: Could an Error type for historic sampling be introduced?

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
```diff
@@ -0,0 +1,432 @@
...
+            .map_err(|error| SamplingError::OpenStream { peer_id, error })?;
+        Ok(SampleStream { stream, peer_id })
+    }
+
+    /// Get a hook to the sender channel for historic sampling requests
+    pub fn historic_request_channel(&self) -> UnboundedSender<SampleArgs<Membership>> {
+        self.historic_request_sender.clone()
+    }
+
+    pub fn try_wake(&mut self) {
```
bacv: This seems not used anywhere?

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
```diff
@@ -0,0 +1,439 @@
...
+        local_peer_id: &PeerId,
+        blob_ids: &HashSet<BlobId>,
+        control: &Control,
+    ) -> Result<Vec<DaLightShare>, Box<dyn std::error::Error>> {
+        let mut subnetwork_tasks = Vec::new();
+
+        for subnetwork_id in subnets {
+            let peer_id = {
+                let mut rng = rand::thread_rng();
+                Self::pick_subnetwork_peer(*subnetwork_id, membership, local_peer_id, &mut rng)
```
bacv: This could be an error instead of an unwrap.

nomos-da/network/core/src/swarm/validator.rs
```diff
@@ -41,7 +41,7 @@ use crate::{
     SubnetworkId,
 };
 
-pub type SampleArgs<Membership> = (HashSet<BlobId>, BlockNumber, Membership);
+pub type SampleArgs<Membership> = (HashSet<BlobId>, BlockNumber, HeaderId, Membership);
```
bacv, line 44: At this point we should have a struct instead of type, I've seen the use of `response.2`, explicit names would be more readable.

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
```diff
@@ -0,0 +1,439 @@
...
+        let local_peer_id = self.local_peer_id;
+
+        let request_future = async move {
+            // todo: handle errors and retries
+            let shares =
+                Self::sample_all_shares(&subnets, &membership, &local_peer_id, &blob_ids, &control)
+                    .await
+                    .unwrap();
+
+            // todo: maybe reuse the connection for commitments
```
bacv: This one is not explicit in the specification, but we are sampling shares, and have a special request for commitments when we as a node don't have them during the dispersal because of timing. So in historic case we should be indeed reusing the same connection and consider sampling failed, if remote node doesn't have either of them.

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
```diff
@@ -0,0 +1,439 @@
...
+                }
+                Err((block_id, sampling_error)) => {
+                    return Some(Self::handle_historic_error(block_id, sampling_error));
+                }
+            }
+        }
+        None
+    }
+
+    const fn handle_historic_response(
```
bacv: nit: `handle_historic_success`

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
```diff
@@ -0,0 +1,439 @@
...
+    /// Get a hook to the sender channel for historic sampling requests
+    pub fn historic_request_channel(&self) -> UnboundedSender<SampleArgs<Membership>> {
+        self.historic_request_sender.clone()
+    }
+
+    pub fn try_wake(&mut self) {
+        if let Some(waker) = self.waker.take() {
+            waker.wake();
+        }
+    }
```
danielSanchezQ: this should probably not be public

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
```diff
@@ -0,0 +1,439 @@
...
+                &subnets,
+                &local_peer_id,
+                &blob_ids,
+                &control,
+            )
+            .await
+            .unwrap();
+
+            Ok((block_id, shares, commitments))
+        }
```
danielSanchezQ, line 221: maybe a set would be better? as they have to be different.

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
```diff
@@ -0,0 +1,439 @@
...
+        .boxed();
+
+        self.historic_request_tasks.push(request_future);
+    }
+
+    async fn sample_all_shares(
```
danielSanchezQ, line 227: `FuturesUnordered` would be better, so it can be resolved asynchronously.

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
```diff
@@ -0,0 +1,439 @@
...
+                    .unwrap()
+            };
+
+            let task = Self::sample_shares_for_subnetwork(
+                peer_id,
+                control.clone(),
+                blob_ids,
+                *subnetwork_id,
+            );
+            subnetwork_tasks.push(task);
```
danielSanchezQ: this is completely sync...better async...

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
```diff
@@ -0,0 +1,439 @@
...
+        }
+    }
+
+    async fn handle_commitment_request(
+        stream: SampleStream,
+        request: sampling::SampleRequest,
+    ) -> Result<(DaSharesCommitments, SampleStream), Box<dyn std::error::Error>> {
+        // todo: handle errors and retries
+        let response = streams::stream_sample(stream, request).await.unwrap();
+        let new_stream = response.2;
```
danielSanchezQ:
```suggestion
        let (response_result, new_stream) = streams::stream_sample(stream, request).await.unwrap();

        match response_result {
            SampleResponse::Commitments(comm) => Ok((comm, new_stream)),
            SampleResponse::Error(err) => Err(format!("Commitment request failed: {err:?}").into()),
            SampleResponse::Share(_) => Err("Expected commitment response".into()),
        }
    }
```

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
```diff
@@ -0,0 +1,439 @@
...
+
+        Ok(commitments)
+    }
+
+    async fn handle_share_request(
+        stream: SampleStream,
+        request: sampling::SampleRequest,
+    ) -> Result<(DaLightShare, SampleStream), Box<dyn std::error::Error>> {
+        // todo: handle error
+        let response = streams::stream_sample(stream, request).await.unwrap();
```
danielSanchezQ:
```suggestion
        let (response_result, new_stream) = streams::stream_sample(stream, request).await.unwrap();

        match response_result {
            SampleResponse::Share(share) => Ok((share.data, new_stream)),
            SampleResponse::Error(err) => Err(format!("Share request failed: {err:?}").into()),
            SampleResponse::Commitments(_) => Err("Expected share response".into()),
        }
```

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
```diff
@@ -0,0 +1,439 @@
...
+            SampleResponse::Commitments(_) => Err("Expected share response".into()),
+        }
+    }
+
+    async fn handle_commitment_request(
+        stream: SampleStream,
+        request: sampling::SampleRequest,
+    ) -> Result<(DaSharesCommitments, SampleStream), Box<dyn std::error::Error>> {
+        // todo: handle errors and retries
+        let response = streams::stream_sample(stream, request).await.unwrap();
```
danielSanchezQ: Use a let match instead.

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
pradovic: 💯 Yes, this draft did not cover errors yet. I left todo commets, I will cover those in the end once I have structure finished

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
pradovic: So, do we need to try and fetch commitments for each subnetwork? 
I guess we only need to do it once, in 1 of choosen 20 networks on a random peer?
I have added some retry logic now, let me know if it makes sense. But, it is treating commitments separately after sample all shares. The peer is chosen randomly again. Maybe I could try to reuse some of existing connection, but it is tricky to pick which one to choose.

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
pradovic: Adding these in the next step, I left todos and unwraps for errors

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
danielSanchezQ: Commitments should be the first thing to try. If after trying the same number of retries (5 peers) we cannot get them, the sampling is rendered failed inmidiately. We do not even need to request the shares, because you cannot validate them.

bacv: For now lets go with a stricter path where we request commitment per peer for blob_id.

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
```diff
@@ -205,22 +202,56 @@ where
...
+                        match Self::handle_share_request(stream, request).await {
+                            Ok((share_data, new_stream)) => {
+                                all_shares.push(share_data);
+                                blob_ids.remove(&blob_id);
+                                stream = new_stream;
+                            }
+                            Err(err) => {
+                                log::error!("Failed to handle share request: {err}");
+                                break; // Break inner loop, try next peer with
+                                       // remaining blobs
```
danielSanchezQ: instead tag the for loop `'blobs_loop: for....` & `break 'blobs_loop`

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
pradovic: I did this at the "global" level:
```
let commitments = Self::sample_all_commitments(
                &membership,
                &subnets,
                &local_peer_id,
                &blob_ids,
                &control,
            )
            .await
            .unwrap();

            // todo: handle errors and retries
            let shares =
                Self::sample_all_shares(&subnets, &membership, &local_peer_id, &blob_ids, &control)
                    .await
                    .unwrap();

            Ok((block_id, shares, commitments))
```

The question is if we should do it inside sample_all_shares and ask it many times and discard duplicate results, or is it enough for the caller to get commitments for each block. The first would check each peer for each blob for commitments, and this implementation will try it on a random peer in some of the choosen subnet for up to 5 peers and thats it.

bacv: > Commitments should be the first thing to try. If after trying the same number of retries (5 peers) we cannot get them, the sampling is rendered failed inmidiately. We do not even need to request the shares, because you cannot validate them.

This is how we do things in new block sampling and I think it's correct, but we might want to clarify the historic sampling case, at least in our implementation. When sampling historic data do we want to:
- Ensure that nodes store LightShares, and only some of them have commitments (they all should in theory, but how do we know?).
- Ensure that nodes store DaShares (LightShares + Commitments), even though for verification we need only to obtain one set of commitments from any peer for that blob_id

I suggest in our implementation to expect every peer to have all data, and query it, even though it might be not optimal, because we only need to get it once. Then later optimise it if needed.
This suggestion is not because it's simpler to implement, I think it's important for us now to have a very strict rules for our own testing. If this works, then we can for sure just request one set of commitments later on, our protocol allows it and other implementations will certainly do this optimization.

danielSanchezQ: No, this was discussed as part of sampling. We added one extra step for sampling, so instead of 20 samples we have 21. Starting from commitments. If any of those samples fails, then its rendered failed. In this case we can even short-circuit the whole block, as just on blob makes sampling fail.

pradovic: Ok, then I will sample first commitments for all blobs, and then shares for 20 samples. I guess that should cover it.

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
pradovic: Oups, missed to fix this, good catch, thanks

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
pradovic: This should actually work as is (I think) as break will break the inner blob_loop and it will continue to try with new peer if there is more blobs to be done, correct?

danielSanchezQ: yes it will, but its less explicit, so when reading its easier to see what tag loop is breaking.

pradovic: oh ok, makes sense, deal

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
pradovic: Would it actually make sense to just return success or failure? As this has retry logic up to 5 peers, so in the end the failure will be after retries. Maybe we can distinguish between could not find commitments and or samples? Maybe some that would map to internal server error would be useful as well.

bacv: I was referring to all the places where `Box<dyn Error>` is used. This specific method might actually have io errors, protocol errors, serialization errors. If possible, we could reuse types defined in `nomos-da/network/core/src/protocols/sampling/errors.rs`

pradovic: Makes sense. My point was that on most of the errors we encounter, we will try a new peer and continue until 5 peers limit is reached. At that point we return an error, and we could return sampling failed and some peer non related error, like if membership does not have candidate peers and stuff. But for peer (stream) related errors we would continue retrying (we can add debug logs). I was wondering do we actually need to return the sampling errors from errors.rs. Would we return the last one we tried?

pradovic: Maybe I can explain better with the example: https://github.com/logos-co/nomos/pull/1545/commits/12e232170bd2cb1a4ed468ffd860c62097f95a06.
It looks to me that this might be enough + logging the individual errors. And if some specific errors from sampling appear that would break the retry, we can add and map them.

bacv: Okay, now that error handling is visible I understand what the proposal is. Unfortunately, at least for the open stream errors, we can't just log them. Similar problem was encountered in the `sampling/requests/request_behaviour.rs` where if we don't have an existing connection, open stream will return error immediately, but not because of something wrong - just to indicate that we first need to wait for a connection and then retry opening the stream again. Let me come back to this tomorrow, I need to some time to think about it, but since both sampling behaviours are very similar, we might want to borrow solution from there.

bacv: By the way, we'll remove `log` crate from here, I think it was introduced for debugging, lets change to `tracing` for now, other behaviours will need to be updated in other PR. I think we might be able to completely remove logging (unless for debugging) from this crate.

pradovic: Ah yes, I see what you mean now. I was just talking about what do we want to return as the parent response to the caller.

I will take a look about handling stream error as well, it is a different and much more important problem, actually. I still did not handle call cases for dialing

pradovic: A bit different approach, but similar. Maybe something like this could work: https://github.com/logos-co/nomos/pull/1555.

nomos-da/network/core/src/swarm/validator.rs
pradovic: I would leave this for a different "clean-up" PR where I want to  clean-up the types in da-network a bit. Is that ok?

bacv (approved):
Thanks for fixing the comments! And sorry for blocking with late review.

danielSanchezQ (approved):
Took time, but here it is. Thanks!

PR #1542, opened 2025-08-13
test(blend): extensive unit tests for core->core logic

ntn-x2:
## 1. What does this PR implement?

PR 1/3. This setups up the testing infrastructure and introduces unit test cases for core->core behaviour. Core->edge as well as the swarm logic (e.g., peer re-dials) will be introduced in subsequent PRs. We cover all edge cases and potential concurrency issues arising from how rust-lip2p works.

## 2. Does the code have enough context to be clearly understood?

Yes.

## 3. Who are the specification authors and who is accountable for this PR?

@ntn-x2 

## 4. Is the specification accurate and complete?

Yes.

## 5. Does the implementation introduce changes in the specification?

No.

## Checklist

* [x] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [x] 2. Description added.
* [x] 3. Context and links to Specification document(s) added.
* [x] 4. Main contact(s) (developers and specification authors) added
* [x] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 6. Link PR to a specific milestone.

PR #1547, opened 2025-08-14
test(blend): add core->edge behavior tests

ntn-x2:

Blend testing PR 2/3 and based on top of the core->core testing one from which it shares most of the test setup infrastructure: https://github.com/logos-co/nomos/pull/1542.

This PR introduces test cases for all the edge cases I could think of, and helped me find a bug that happens when a number of edge peers connect to a core node which would bring the total number of connected peers above the maximum threshold. In some cases, we would still receive and bubble up to the swarm the messages we receive from both of them, although some of the connections should have been rejected. So we have a fix now with a two-step phase: we first upgrade the connection, and then when we are ready to receive we check again, and since libp2p is not multithreaded, we are sure that at any given time we are able to block unwanted connections from sending us messages. So I would say the testing effort is paying back the effort.

Yes.

@ntn-x2, spec: @madxor.

Yes.

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/network/src/core/with_edge/behaviour/tests/utils.rs
+++ b/nomos-blend/network/src/core/with_edge/behaviour/tests/utils.rs
@@ -12,14 +12,14 @@ use crate::{core::with_edge::behaviour::Behaviour, PROTOCOL_NAME};
 
 #[derive(Default)]
 pub struct BehaviourBuilder {
-    edge_peer_ids: Vec<PeerId>,
+    core_peer_ids: Vec<PeerId>,
     max_incoming_connections: Option<usize>,
     timeout: Option<Duration>,
 }
 
 impl BehaviourBuilder {
-    pub fn with_edge_peer_membership(mut self, edge_peer_id: PeerId) -> Self {
-        self.edge_peer_ids.push(edge_peer_id);
+    pub fn with_core_peer_membership(mut self, edge_peer_id: PeerId) -> Self {
+        self.core_peer_ids.push(edge_peer_id);
         self
     }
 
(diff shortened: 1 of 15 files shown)


nomos-blend/network/src/core/with_edge/behaviour/tests/utils.rs
```diff
@@ -0,0 +1,81 @@
...
+use nomos_blend_scheduling::membership::{Membership, Node};
+
+use crate::{core::with_edge::behaviour::Behaviour, PROTOCOL_NAME};
+
+#[derive(Default)]
+pub struct BehaviourBuilder {
+    edge_peer_ids: Vec<PeerId>,
+    max_incoming_connections: Option<usize>,
+    timeout: Option<Duration>,
+}
```
youngjoon-lee: I guess we should rename this to `core_peer_ids`? 
```suggestion
pub struct BehaviourBuilder {
    core_peer_ids: Vec<PeerId>,
    max_incoming_connections: Option<usize>,
    timeout: Option<Duration>,
}
```
because the membership should be built from the list of core nodes, as the following code does?
https://github.com/logos-co/nomos/blob/9325cc6888ed91e2a035c0d1d67aad394caa8fb1/nomos-blend/network/src/core/with_edge/behaviour/tests/bootstrapping.rs#L62

Please let me know if I'm missing anything.

ntn-x2: You are right. I changed that https://github.com/logos-co/nomos/pull/1547/commits/f2490b623728e015b723288f51d8d9cd28cf7e1d.

PR #1548, opened 2025-08-15
feat(blend): session transition: core node network behaviour

youngjoon-lee:

This is the 1st PR of the session transition ([spec](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df8156a1d2dc9e7f81d289)).

- Modified two `NetworkBeahviour`s for the core service (`nomos-blend-network::core::with_core::Behaviour` and `nomos-blend-network::core::with_edge::Behaviour`).
  - Maintain both current and previous sessions.
    - Close substreams of the previous session when we receive a signal that the transition period has passed.
  - Choose the correct session for each operation. 
    - **Publish** messages to the peers in the **current** session.
    - **Forward** messages to the peers in the **session to which the receiving connection belongs** (Of course, except the receiving connection).

I didn't modify the CoreService yet. I will do it at a next PR. There, I will implement establishing connections for a new session and notifying the behaviours of the session transition.

You can find some TODO comments in this PR. I will resolve them at a next PR because we have to refactor `Membership`.

Also, I will add tests at a next PR by merging this with https://github.com/logos-co/nomos/pull/1542. For now, please focus on the approach to choose the correct session for each operation.

Yes

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/network/src/core/with_core/behaviour/mod.rs
+++ b/nomos-blend/network/src/core/with_core/behaviour/mod.rs
@@ -354,7 +354,11 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
             .then_some(current_session_behaviour!(self, session))
     }
 
-    /// Creates [`SessionBehaviour`] for the session that has the given negotiated connection.
+    /// Creates [`SessionBehaviour`] for the session that has the given
+    /// pending upgrade.
+    ///
+    /// The pending upgrade will be removed from the set of pending upgrades
+    /// in the session.
     fn session_behaviour_for_pending_upgrade(
         &mut self,
         conn: &(PeerId, ConnectionId),


nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -161,23 +180,26 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
     /// assumed to have been properly formed.
     pub fn publish_validated_message(
         &mut self,
         message: &EncapsulatedMessageWithValidatedPublicHeader,
     ) -> Result<(), Error> {
-        self.forward_validated_message_and_maybe_exclude(message, None)?;
-        Ok(())
+        self.current_session_behaviour()
+            .forward_validated_message_and_maybe_exclude(message, None)
     }
```
youngjoon-lee, line 203: This is a **key change**. We use the **current** session to publish messages. For that, I introduced the `SessionBehaviour` struct, which is a temporary struct for session-based operations. You can find it below.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -187,17 +209,22 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
     ) -> Result<(), Error> {
-        self.forward_validated_message_and_maybe_exclude(message, Some(excluded_peer))?;
-        Ok(())
+        self.session_behaviour(&(excluded_peer_id, excluded_conn_id))
+            .ok_or(Error::NoSessionForPeer {
+                peer_id: excluded_peer_id,
+                connection_id: excluded_conn_id,
+            })?
+            .forward_validated_message_and_maybe_exclude(message, Some(excluded_peer_id))
     }
```
youngjoon-lee: This is another **key change**. We forward messages to the peers in the **session to which the excluded peer belongs**.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -129,28 +116,60 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
+    /// Closes all negotiated connections in the session.
+    fn close_session(&mut self, session: &Session) {
+        session.negotiated_peers().for_each(
+            |(&peer_id, RemotePeerConnectionDetails { connection_id, .. })| {
+                self.events.push_back(ToSwarm::NotifyHandler {
+                    peer_id,
+                    handler: NotifyHandler::One(*connection_id),
+                    event: Either::Left(FromBehaviour::CloseSubstreams),
+                });
+                self.try_wake();
```
ntn-x2, line 166: We should call this once at the end, if we dealt with at least one open connection.

nomos-blend/network/src/core/with_core/behaviour/session.rs
```diff
@@ -0,0 +1,238 @@
...
+        self.previous = Some(mem::replace(
+            &mut self.current,
+            Session::new(new_membership),
+        ));
+        previous
+    }
+
+    /// Takes the previous session out of the [`Sessions`],
+    /// leaving a [`None`] in its place.
+    pub const fn close_previous_session(&mut self) -> Option<Session> {
```
ntn-x2, line 49: Would it not make more sense to have a `close` method on `Session`, which we can only get by calling `transition` on `Sessions`? It would be a more type-safe approach, but I think it would require `Session` to hold a mutable reference to `Sessions`, since it needs to call `take` on it. Ok maybe it is, maybe it is not worth the complexity.

nomos-blend/network/src/core/with_core/behaviour/session.rs
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    /// Returns a mutable reference to the current session.
+    pub const fn current_session_mut(&mut self) -> &mut Session {
+        &mut self.current
+    }
+
+    /// Returns a mutable reference to the session
+    /// that has the given negotiated peer.
+    pub fn session_mut(
```
ntn-x2:
```suggestion
    pub fn session_mut_for_connection(
```

nomos-blend/network/src/core/with_core/behaviour/session.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+impl Session {
+    fn new(membership: Membership<PeerId>) -> Self {
+        let membership_size = membership.size();
+        Self {
+            membership,
+            negotiated_peers: HashMap::new(),
+            connections_waiting_upgrade: HashMap::new(),
+            exchanged_message_identifiers: HashMap::with_capacity(membership_size),
```
ntn-x2: I know this is what was done before, but I think it does not really make too much sense. We will definitely have way more messages than number of peers.

nomos-blend/network/src/core/with_core/behaviour/session.rs
```diff
@@ -0,0 +1,238 @@
...
+
+    pub fn is_core_node(&self, peer_id: &PeerId) -> bool {
+        self.membership.contains_remote(peer_id)
+    }
+
+    pub fn num_negotiated_peers(&self) -> usize {
+        self.negotiated_peers.len()
+    }
+
+    pub fn available_connection_slots(&self, peering_degree: &RangeInclusive<usize>) -> usize {
```
ntn-x2: Why not passing directly the maximum value instead of the whole range? The `start` of the range is irrelevant here. I think I would even prefer this value to be stored inside the struct and passed to it at construction time rather, otherwise this function feels quite "leaky" to me.

nomos-blend/network/src/core/with_core/behaviour/session.rs
```diff
@@ -0,0 +1,238 @@
...
+            }
+        }
+
+        None
+    }
+
+    /// Checks if the given peer has been negotiated in either the current or
+    /// previous session.
+    pub fn is_negotiated(&self, peer_id: PeerId) -> bool {
+        self.current.negotiated_peer(&peer_id).is_some()
```
ntn-x2: I would add another method that returns a `bool` and that uses `contains_key` internally, so that we don't need to read the whole value from storage but can simply decide based on the presence of the key in the map.

nomos-blend/network/src/core/with_core/behaviour/session.rs
ntn-x2: Or we could even implement `Drop` for a session.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -207,11 +234,138 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
+    /// set of pending connections since it has now been processed.
+    ///
+    /// # Panics
+    ///
+    /// If the specified connection is not present in the map of connections
+    /// waiting to be upgraded and this connection has not already been upgraded
+    /// before, since we need to peer role (i.e., dialer or listener) before
+    /// moving the connection into a different storage map.
+    fn handle_negotiated_connection(&mut self, (peer_id, connection_id): (PeerId, ConnectionId)) {
+        let Some((mut behaviour, role)) =
```
ntn-x2: I think we need to check that the session is the latest one. We should not be accepting new connections from peers from the old session, right? At least the spec does not mention that and I think we should not, indeed.

ntn-x2: I'm saying this because while I was testing, I discovered a few concurrency cases which need to be handled properly. In this case, we close connections with all negotiated peers when we close a session, but any of the pending one would be able to be upgraded once the connection is closed, which is not what we want.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -207,11 +234,138 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
+                    self.sessions.is_negotiated(peer_id),
+                    "Negotiated connection {connection_id:?} with peer {peer_id:?} not found in storage of pending connections."
+                );
+            return;
+        };
+
+        behaviour.handle_negotiated_connection((peer_id, connection_id), role);
+    }
+
+    fn handle_spammy_connection(&mut self, (peer_id, connection_id): (PeerId, ConnectionId)) {
```
ntn-x2, line 283: Also for all these events, we want to report them to the swarm only if they belong to the new connection. The old connection should mostly be invisible to the swarm, since we do not want to trigger things like re-dialing or connection to a new peer, in those cases.

ntn-x2: But maybe it's worth clarifying with Marci what the node should do in case some connections are dropped during the transition period.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -207,11 +234,138 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
+            events: &mut self.events,
+            waker: &mut self.waker,
+            peering_degree: self.peering_degree.clone(),
+            local_peer_id: self.local_peer_id,
+        }
+    }
+
+    /// Creates [`SessionBehaviour`] for the session
+    /// that has the given negotiated peer.
+    fn session_behaviour(
```
ntn-x2:
```suggestion
    fn session_behaviour_for_connection(
```

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -207,11 +234,138 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
+            events: &mut self.events,
+            waker: &mut self.waker,
+            peering_degree: self.peering_degree.clone(),
+            local_peer_id: self.local_peer_id,
+        })
+    }
+
+    /// Creates [`SessionBehaviour`] for the session
+    /// that has the given connection waiting to be upgraded.
+    fn session_behaviour_for_conn_waiting_upgrade(
```
ntn-x2: I think we might want to specify that the pending connection is removed from the map.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -207,11 +234,138 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
+                local_peer_id: self.local_peer_id,
+            },
+            role,
+        ))
+    }
+}
+
+/// A temporary structure that is created from [`Behaviour`]
+/// in order to perform the behaviour only for the specified session.
+struct SessionBehaviour<'a> {
```
ntn-x2:
```suggestion
struct SessionBehaviour<'behaviour> {
```

ntn-x2: Or at least

```suggestion
struct SessionBehaviour<'b> {
```

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -589,17 +704,39 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
+    fn handle_closed_negotiated_connection(&mut self, peer_id: PeerId) {
+        let details = self
+            .session
+            .remove_negotiated_peer(&peer_id)
+            .expect("Peer must exist");
+        self.events
+            .push_back(ToSwarm::GenerateEvent(Event::PeerDisconnected(
+                peer_id,
+                details.negotiated_state,
+            )));
```
ntn-x2: As I was saying elsewhere, I don't think we want to notify the swarm about it if this is the session we have passed, right?

nomos-blend/network/src/core/with_edge/behaviour/mod.rs
```diff
@@ -68,6 +68,18 @@ impl Behaviour {
...
 
+    pub fn start_new_session(&mut self, new_membership: Membership<PeerId>) {
+        tracing::debug!(target: LOG_TARGET, "Starting new session with membership {new_membership:?}.");
+        self.current_membership = new_membership;
+        // Close all the upgraded connections without waiting for the transition period,
+        // so that edge nodes can retry with the new membership.
+        // No need to maintain the transition period for messages that have not yet
+        // been fired.
+        for connection in mem::take(&mut self.upgraded_edge_peers) {
+            self.close_substream(connection);
```
ntn-x2, line 79: `close_substream` calls `try_awake` internally. Maybe a boolean flag in the function signature indicating whether the waker should be awoken would help avoiding unnecessary awakes. We can then call awake once at the end of this loop if there was at least one component.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: I think it's okay to call this multiple times because (inside the `try_wake`) we call `wake` only when `waker` is not `None` (by `take`-ing the `waker`).

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: I thought about it. The issue is that we don't know which session the dialer belongs to. So, in this PR, I'm accepting all the negotiated connections, assuming that dialers belong to the current session.
If a dialer is one of core nodes in the current session (fortunately), we will create a proper connection handler. If not, we will create a dummy connection handler. 
Even if a dialer established a connection in the previous session, and if the upgrade was done after a new session is started, I think it's not a problem to accept the connection. The dialer will anyway build messages using the new membership and send them through the connection. 
Please let me know what you think about this perspective.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: TBH, I don't think these details need to specifically defined in the spec. We can choose the way that we think is reasonable because it's kind of internal policy (e.g. re-dialing) that each node can define.

When implementing this PR, I tried to minimize differentiating the current and previous sessions by introducing a common struct `SessionBehaviour`. And, I thought that the swarm can decide what to do with the event it receives, because the behaviour doesn't know what the swarm will do. 

But I agree that it's more convenient to not bubble up events generated for the previous session. It will make the swarm logic simpler. We can later generate events if necessary.

I will address this after reading all other comments.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: I agree https://github.com/logos-co/nomos/pull/1548/files/9ad102e8ac1a90e73ca777410a3025bdcce9f375#r2283896103

nomos-blend/network/src/core/with_core/behaviour/session.rs
youngjoon-lee: This `close_previous_session` is meant to be called after the session transition period $T$ has passed. In other words, this method needs to be called **separately** with the `transition` method (which is called when a new session is started). I will improve the comment.

It might be convenient if we can implement `Drop` for `Session`. But, like the following code (and as you mentioned), we need to access some data structures (e.g. `events`) to clean up a `Session`. 
https://github.com/logos-co/nomos/blob/9ad102e8ac1a90e73ca777410a3025bdcce9f375/nomos-blend/network/src/core/with_core/behaviour/mod.rs#L143-L149

So, I'd like to go with the current implementation if it is not so hard to read.

nomos-blend/network/src/core/with_edge/behaviour/mod.rs
youngjoon-lee: As I commented in https://github.com/logos-co/nomos/pull/1548/files/9ad102e8ac1a90e73ca777410a3025bdcce9f375#r2283883552, I don't think we need to worry about calling `try_wake` multiple times. Please let me know if I'm missing anything.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
ntn-x2: Yes ok, fair enough.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
ntn-x2: Ok, since we are anyway closing all connections at the end of the transition period, I guess indeed it does not matter if we accept the connection, so we can probably leave it.

youngjoon-lee: Actually, this topic is highly related to how we define the purpose of the transition period. If the purpose is processing scheduled messages as many as possible instead of giving them up right away, I think it's meaningful to accept new conn upgrades from the old session because the peer may be dialing to us since he has 0 healthy connection right now.
But, yes. That's an edge case. Either approach makes sense. The complexity of implementation is almost the same in both approaches. 

Okay, I will follow the approach you suggested. It also aligns well with your another suggestion: https://github.com/logos-co/nomos/pull/1548#discussion_r2281649131, which I said I agree with.
I will modify the code and re-request your review.

nomos-blend/network/src/core/with_core/behaviour/session.rs
youngjoon-lee: [done](https://github.com/logos-co/nomos/pull/1548/commits/f6f92cc38014c7bef0d73b626e36f331fb0bb984#diff-d32b143b043108b6e80bba259ba756bf204e78447767afa44377830e20f6af7aR65)

nomos-blend/network/src/core/with_core/behaviour/session.rs
youngjoon-lee: [done](https://github.com/logos-co/nomos/pull/1548/commits/f6f92cc38014c7bef0d73b626e36f331fb0bb984#diff-d32b143b043108b6e80bba259ba756bf204e78447767afa44377830e20f6af7aR138)

nomos-blend/network/src/core/with_core/behaviour/session.rs
youngjoon-lee: [done](https://github.com/logos-co/nomos/pull/1548/commits/f6f92cc38014c7bef0d73b626e36f331fb0bb984#diff-d32b143b043108b6e80bba259ba756bf204e78447767afa44377830e20f6af7aR150)

nomos-blend/network/src/core/with_core/behaviour/session.rs
youngjoon-lee: [done](https://github.com/logos-co/nomos/pull/1548/commits/f6f92cc38014c7bef0d73b626e36f331fb0bb984#diff-d32b143b043108b6e80bba259ba756bf204e78447767afa44377830e20f6af7aR174)

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: [done](https://github.com/logos-co/nomos/pull/1548/commits/f6f92cc38014c7bef0d73b626e36f331fb0bb984#diff-2a6abfe9a1ac517407672a6214af0382d69d53489549213c5a41f927b07c5dbdR331)

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: done: 4442bb3b

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: [done](https://github.com/logos-co/nomos/pull/1548/commits/f6f92cc38014c7bef0d73b626e36f331fb0bb984#diff-2a6abfe9a1ac517407672a6214af0382d69d53489549213c5a41f927b07c5dbdR382)

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: f6f92cc38014c7bef0d73b626e36f331fb0bb984 I applied your suggestions. 

No test yet, but please check first if the architecture looks good. I will add tests as a next PR by merging this with https://github.com/logos-co/nomos/pull/1542. I won't merge this PR to the master until we have the test PR approved.

In summary, I introduced 3 structs.
- `CommonSessionBehaviour`
  - defines common methods that can be used for either the current or previous session. 
- `PreviousSessionBehaviour`
  - extends `CommonSessionBehaviour` by defining extra methods only for the previous session.
- `CurrentSessionBehaviour`
  - extends `CommonSessionBehaviour` by defining extra methods only for the current session.

We construct a proper struct when necessary.

In principle,
- We don't accept connection upgrades for the previous session.
- We ignore any conn mgmt (spammy, unhealthy, ..) events for the previous session.
  - But we don't block the swarm passing those event to the behaviour, because those events are originally generated from the connection handler and passed to the swarm. The swarm doesn't know whether the event is related for the current session or not.

PR #1550, opened 2025-08-16
feat(lib): Add MMR tree

zeegomo:
Add a simple persistent MMR tree as a base crate. A first use will be leader vouchers, but we can envision more applications in the future. 
Compared to the already existing dynamic merkle tree implementation, this uses much less memory since it does not store every single element but only the frontier (O(log(n)), at the cost of not supporting deletions and not saving each leaf.

Yes

@zeegomo 
N/A

N/A
## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/mmr/Cargo.toml
+++ b/mmr/Cargo.toml
@@ -19,3 +19,7 @@ workspace = true
 
 [features]
 serde = ["dep:serde", "rpds/serde"]
+
+[package.metadata.cargo-machete]
+# False positives.
+ignored = ["proptest"]


mmr/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+    static PRECOMPUTED_EMPTY_ROOTS: OnceLock<[[u8; 32]; 32]> = OnceLock::new();
+    assert!(height < 32, "Height must be less than 32: {height}");
+    PRECOMPUTED_EMPTY_ROOTS.get_or_init(|| {
+        let mut hashes = [EMPTY_VALUE; 32];
+        for i in 1..32 {
+            hashes[i] = hash::<[u8; 32], Hash>(hashes[i - 1], hashes[i - 1]);
+        }
+        hashes
+    })[height]
+}
```
danielSanchezQ, line 150: Why not `LazyLock`? Because of the Generic hash usage?

zeegomo: mmm, can't remember, let me try

mmr/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+use rpds::StackSync;
+
+const EMPTY_VALUE: [u8; 32] = [0; 32];
+
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MMR<T, Hash, const MAX_HEIGHT: u8 = 32> {
+    roots: StackSync<Root>,
+    _hash: std::marker::PhantomData<(T, Hash)>,
+}
```
davidrusu: The naming guidelines state that acronyms should only capitalize the first letter: https://rust-lang.github.io/api-guidelines/naming.html
e.g `MMR` should be `Mmr`. 

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn. In snake_case, acronyms and contractions are lower-cased: is_xid_start.


Spelling out the whole name is also not the worst thing: `MerkleMountainRange`

mmr/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+        let mut roots = self.roots.clone();
+
+        while let Some(root) = roots.peek().cloned() {
+            if last_root.height == root.height {
+                roots.pop_mut();
+                last_root = Root {
+                    root: hash::<_, Hash>(root.root, last_root.root),
+                    height: last_root.height + 1,
+                };
+                assert!(last_root.height < MAX_HEIGHT, "Height must be less than 32");
```
davidrusu: Should use the `MAX_HEIGHT` value in the assert error message instead of `32`.

mmr/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+    height: u8,
+}
+
+impl<const MAX_HEIGHT: u8, T, Hash> MMR<T, Hash, MAX_HEIGHT>
+where
+    T: AsRef<[u8]>,
+    Hash: Digest<OutputSize = digest::typenum::U32>,
+{
+    pub fn new() -> Self {
+        Self {
```
davidrusu: Wdyt of an early assert here on initialization for `assert_le!(MAX_HEIGHT, 32)`?

mmr/src/lib.rs
davidrusu: Maybe a comment on why strict inequality is important here? i.e. we want the frontier root's tree to not exceed MAX_HEIGHT.

mmr/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+            root = hash::<_, Hash>(last.root, root);
+            depth += 1;
+        }
+        assert!(depth <= MAX_HEIGHT);
+        // ensure a fixed depth
+        while depth < MAX_HEIGHT {
+            root = hash::<_, Hash>(root, empty_subtree_root::<Hash>(depth as usize));
+            depth += 1;
+        }
+
```
davidrusu, line 123: do you mind adding one more assert here?
```
assert_eq!(depth, MAX_HEIGHT);
```

The frontier tree should always be at MAX_HEIGHT

mmr/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+        Self {
+            roots,
+            _hash: std::marker::PhantomData,
+        }
+    }
+
+    pub fn frontier_root(&self) -> [u8; 32] {
+        let mut root = empty_subtree_root::<Hash>(0);
+        let mut depth = 0;
```
davidrusu: consider renaming `depth` to `height` here, the rest of the comparisons in this function will make a bit more sense

mmr/src/lib.rs
zeegomo: I had to go back to `OnceLock` because otherwise the `Hash` generic would leak into the static (through the initialization function) var which is not yet supported

PR #1551, opened 2025-08-18
feat(nat): mapping lease renewal

andrussal:

This PR adds mapping lease renewal to the nat mapping behaviour. It will request new mapping before active lease expires.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-libp2p/src/behaviour/nat/address_mapper/nat_pmp.rs
+++ b/nomos-libp2p/src/behaviour/nat/address_mapper/nat_pmp.rs
@@ -5,8 +5,7 @@ use natpmp::{new_tokio_natpmp, NatpmpAsync, Protocol, Response};
 use tokio::{net::UdpSocket, time::timeout};
 
 use crate::{
-    behaviour::nat::address_mapper::{errors::AddressMapperError, protocol::NatMapper},
-    config::NatMappingSettings,
+    behaviour::nat::address_mapper::errors::AddressMapperError, config::NatMappingSettings,
 };
 
 type PortNumber = u16;
@@ -62,9 +61,8 @@ impl NatPmp {
     }
 }
 
-#[async_trait::async_trait]
-impl NatMapper for NatPmp {
-    async fn map_address(
+impl NatPmp {
+    pub async fn map_address(
         local_address: &Multiaddr,
         settings: NatMappingSettings,
     ) -> Result<Multiaddr, AddressMapperError> {
--- a/nomos-libp2p/src/behaviour/nat/address_mapper/upnp.rs
+++ b/nomos-libp2p/src/behaviour/nat/address_mapper/upnp.rs
@@ -9,8 +9,7 @@ use multiaddr::Protocol;
 use tracing::info;
 
 use crate::{
-    behaviour::nat::address_mapper::{errors::AddressMapperError, protocol::NatMapper},
-    config::NatMappingSettings,
+    behaviour::nat::address_mapper::errors::AddressMapperError, config::NatMappingSettings,
 };
 
 type AddressWithProtocol = (SocketAddr, PortMappingProtocol);
@@ -28,9 +27,8 @@ impl UpnpProtocol {
     }
 }
 
-#[async_trait::async_trait]
-impl NatMapper for UpnpProtocol {
-    async fn map_address(
+impl UpnpProtocol {
+    pub async fn map_address(
         address_to_map: &Multiaddr,
         settings: NatMappingSettings,
     ) -> Result<Multiaddr, AddressMapperError> {


nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
```diff
@@ -1,77 +1,276 @@
...
+use tracing::{debug, info, warn};
 
 use crate::{
-    behaviour::nat::address_mapper::{errors::AddressMapperError, protocol::ProtocolManager},
+    behaviour::nat::address_mapper::{errors::AddressMapperError, protocol::NatMapper},
     config::NatMappingSettings,
 };
 
+/// Renewal delay as a fraction of the lease duration
+const RENEWAL_DELAY_FRACTION: f64 = 0.8;
```
ntn-x2: We have a `NonNegativeF64` type which might be better here, since we don't want to deal with negative values, I guess. Even `0` might be problematic, so we might have a `NonZeroF64` type maybe.

ntn-x2: https://github.com/logos-co/nomos/blob/72dac5883bf597a71d16241fc2d5f8729fe8d8a7/nomos-utils/src/math.rs#L9

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
```diff
@@ -1,77 +1,276 @@
...
+type MappingFuture = BoxFuture<'static, Result<Multiaddr, AddressMapperError>>;
+
+/// Events emitted by the NAT address mapper
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum Event {
+    /// Address mapping failed for the given internal address
     AddressMappingFailed(Multiaddr),
+    /// The default gateway has changed
     DefaultGatewayChanged,
+    /// The local address has changed
```
ntn-x2, line 50: Is local address the same as internal address?

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
```diff
@@ -1,77 +1,276 @@
...
     LocalAddressChanged(Multiaddr),
+    /// A new external address mapping has been successfully established
     NewExternalMappedAddress(Multiaddr),
```
ntn-x2: Should we return both the internal and the external address? Or rather, the unmapped and the mapped one?

nomos-libp2p/src/behaviour/nat/address_mapper/protocol.rs
```diff
@@ -19,21 +18,28 @@ pub trait MappingProtocol: Send + Sync + 'static {
     async fn map_address(&mut self, address: &Multiaddr) -> Result<Multiaddr, AddressMapperError>;
 }
 
-pub struct ProtocolManager;
+pub struct ProtocolManager {
+    settings: NatMappingSettings,
+}
+
+#[async_trait::async_trait]
+impl NatMapper for ProtocolManager {
```
ntn-x2, line 22: I am wondering if we need this trait anymore? We have the `ProtocolManager` type that takes care of implementing a step-by-step mapping logic. It internally calls `nat_pmp` and `upnp` and calls `map_address`, which does not require them to implement the `NatMapper` trait at all since the types are known to `ProtocolManager`. Then we can make the `new` function for the `ProtocolManager` fit what we need, meaning a `non-async` function that does not return a `Box` and does not return a `Result`, but it simply returns `Self`. What do you think? I could now find any place where each mapping logic implementation would be required to implement the `NatMapper` trait.

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
```diff
@@ -1,77 +1,276 @@
...
+        Self::Retry {
+            internal,
+            retry_timer: Box::pin(time::sleep(RETRY_INTERVAL)),
+            retry_count,
+        }
+    }
+}
+
+/// Network behaviour for managing NAT address mapping
+pub struct AddressMapperBehaviour<P> {
```
ntn-x2: We can store the mapper inside here, if we remove any unnecessary requirements and we make the `initialize` function return simply a `Self`.

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
```diff
@@ -1,77 +1,276 @@
...
             settings,
+            _phantom: std::marker::PhantomData,
         }
     }
 
+    /// Attempts to map the given internal address to an external address
+    ///
+    /// Returns an error if mapping is already in progress for a different
+    /// address. If the same address is already mapped, this is a no-op.
     pub fn try_map_address(&mut self, address: Multiaddr) -> Result<(), AddressMapperError> {
```
ntn-x2, line 329: Just sharing a personal opinion: I would have done it relying more on the compiler, similar to what we have done for the core->edge Blend connection handler: https://github.com/logos-co/nomos/blob/master/nomos-blend/network/src/core/with_edge/behaviour/handler/mod.rs. Where also each state of the state machine would be able to handle Swarm events and implement polling logic. I just wanted to share, but this solution works fine.

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
```diff
@@ -1,77 +1,276 @@
...
+    /// NAT mapping is being established or renewed.
+    Mapping {
+        /// Internal address being mapped.
+        address: Multiaddr,
+        /// Future for the mapping operation.
+        future: MappingFuture,
+        /// Whether this is the initial mapping (vs renewal).
+        is_initial: bool,
+        /// Number of retry attempts made.
+        retry_count: u32,
```
ntn-x2: Why do we need this for `Mapping`? Once we fail once we move into `Retry` state, right?

nomos-libp2p/src/behaviour/nat/address_mapper/protocol.rs
ntn-x2: Ok I saw in the tests it might be good to keep the trait for unit tests, but we can make the `initialize` function non-failing and non-async and make it return `Self`. No need to box I think.

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
ntn-x2: Also, why is this not returning a future? We can poll it from the behaviour `poll` instead, and perhaps it would also be easier to test since we just need to `await` its result.

andrussal: > Just sharing a personal opinion: I would have done it relying more on the compiler, similar to what we have done for the core->edge Blend connection handler: https://github.com/logos-co/nomos/blob/master/nomos-blend/network/src/core/with_edge/behaviour/handler/mod.rs. Where also each state of the state machine would be able to handle Swarm events and implement polling logic. I just wanted to share, but this solution works fine.

This is nice! I refactored mapper a bit based on that.

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
andrussal: Great idea, I created `PositiveF64` for this use case.

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
andrussal: In existing state machine it wasn't strictly required because current state machines assume that only one mapping is in progress. But I added it nevertheless, so state machine can do additional check on it.

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
andrussal: I removed instance creation altogether with other refactoring.

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
andrussal: > Also, why is this not returning a future?

It is called from the wrapper behaviour in poll. It's intent is just to submit mapping request and later wait events.

In general I am not sure if `AddressMapperBehaviour` should be a behaviour at all. All swarm interaction could be done by the wrapping behaviour, so it doesn't need swarm callback functions.  But maybe we can think about in future.

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
andrussal: Mapper moves between `Mapping` and `WaitRetry` states(I renamed it to `WaitRetry`). Here it is stored when `WaitRetry` timer completes. This way it was easy to share `Mapping` for both regular case and retries.

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
andrussal: > Is local address the same as internal address?

I renamed all _internal_ to _local_ to have consistent naming.

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
danielSanchezQ: Nice solution guys, this looks neat. Thanks a lot.

nomos-utils/src/math.rs
```diff
@@ -39,11 +39,54 @@ impl TryFrom<f64> for NonNegativeF64 {
     }
 }
 
+#[derive(Clone, Copy, PartialEq, PartialOrd, Debug)]
+#[cfg_attr(feature = "serde", derive(::serde::Serialize, ::serde::Deserialize))]
+pub struct PositiveF64(
```
ntn-x2, line 44: I think we can probably just make this a tuple struct around a `NonNegativeF64` and in its `TryFrom` implementation call `NonNegativeF64::try_from` and then check if it's `zero`, in which case we also fail. So we avoid duplicating serialization and other logic.

nomos-libp2p/src/behaviour/nat/address_mapper/nat_pmp.rs
```diff
@@ -63,35 +63,31 @@ impl NatPmp {
...
-            .await
-            .map_err(|e| AddressMapperError::PortMappingFailed(e.to_string()))?;
-
-        Ok(Box::new(Self { settings, nat_pmp }))
-    }
-
+impl NatMapper for NatPmp {
     async fn map_address(
-        &mut self,
         internal_address: &Multiaddr,
```
ntn-x2: I still see a bunch of `internal` mentions, did you say all these are meant to be local?

nomos-libp2p/src/behaviour/nat/address_mapper/nat_pmp.rs
```diff
@@ -63,35 +63,31 @@ impl NatPmp {
...
-        Self: Sized,
-    {
-        let nat_pmp = new_tokio_natpmp()
-            .await
-            .map_err(|e| AddressMapperError::PortMappingFailed(e.to_string()))?;
-
-        Ok(Box::new(Self { settings, nat_pmp }))
-    }
-
+impl NatMapper for NatPmp {
```
ntn-x2: Do we need these impls anymore? Aren't we assuming the top-level mapper implements all the logic itself? We probably don't need each step of the NAT machine to implement the trait anymore, but just expose the method the upper-level machine needs.

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
```diff
@@ -1,77 +1,358 @@
...
+        info!(%self.local_address, %external_address, "NAT mapping established");
+
+        let local_address = self.local_address.clone();
+        let new_state = State::active(
+            local_address.clone(),
+            settings.lease_duration,
+            settings.renewal_delay_fraction,
+        );
+
+        let result = if self.is_initial {
```
ntn-x2, line 151: Don'we we need to set `is_initial` to `false` here to avoid generating a new event on the next poll?

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
ntn-x2: I think we could do even better by using typestate and return a different type for a mapper that is already in progress, so that we can't call `try_map_address` again until the previous attempt is completed. But I do not want to drag this for much longer, just leaving my 2 cents to maybe tackle in the future.

nomos-libp2p/src/behaviour/nat/address_mapper/mod.rs
andrussal: > Don'we we need to set `is_initial` to `false` here to avoid generating a new event on the next poll?

Here I think no because we transition to another state immediately and this `Mapper` is not called anymore.

nomos-libp2p/src/behaviour/nat/address_mapper/nat_pmp.rs
andrussal: > Do we need these impls anymore? Aren't we assuming the top-level mapper implements all the logic itself? We probably don't need each step of the NAT machine to implement the trait anymore, but just expose the method the upper-level machine needs.

I removed trait from individual protocols but kept the trait for now for manager. I think it can be useful to mock mapping.

PR #1555, opened 2025-08-18
feat(da-network): historic sampling behavior - wait for connection on dial

pradovic:

Idea for solving the open stream returning connection reset error on dial. Context:https://github.com/logos-co/nomos/pull/1545#discussion_r2283152905

### DIFF
--- a/nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
+++ b/nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
@@ -511,7 +511,7 @@ where
         local_addr: &Multiaddr,
         remote_addr: &Multiaddr,
     ) -> Result<THandler<Self>, ConnectionDenied> {
-        self.connection_broadcast_sender.send(peer).unwrap();
+        let _ = self.connection_broadcast_sender.send(peer);
 
         self.stream_behaviour
             .handle_established_inbound_connection(connection_id, peer, local_addr, remote_addr)
@@ -526,7 +526,7 @@ where
         role_override: Endpoint,
         port_use: PortUse,
     ) -> Result<THandler<Self>, ConnectionDenied> {
-        self.connection_broadcast_sender.send(peer).unwrap();
+        let _ = self.connection_broadcast_sender.send(peer);
 
         self.stream_behaviour
             .handle_established_outbound_connection(
--- a/nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
+++ b/nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
@@ -48,7 +48,7 @@ use crate::{
 };
 
 const MAX_PEER_RETRIES: usize = 5;
-const CONNECTION_WAIT_TIMEOUT: u64 = 3;
+const CONNECTION_WAIT_TIMEOUT: Duration = Duration::from_secs(3);
 
 #[derive(Debug, Error)]
 enum StreamSamplingError {
@@ -141,7 +141,7 @@ where
         match stream_result {
             Ok(stream) => Ok(SampleStream { stream, peer_id }),
             Err(OpenStreamError::Io(io_err)) if io_err.kind() == ErrorKind::ConnectionReset => {
-                let timer = sleep(Duration::from_secs(CONNECTION_WAIT_TIMEOUT));
+                let timer = sleep(CONNECTION_WAIT_TIMEOUT);
                 tokio::pin!(timer);
 
                 loop {
--- a/nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
+++ b/nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::HashSet,
+    future::ready,
     io::ErrorKind,
     task::{Context, Poll, Waker},
 };
@@ -32,7 +33,7 @@ use tokio::{
     },
     time::{sleep, Duration},
 };
-use tokio_stream::wrappers::UnboundedReceiverStream;
+use tokio_stream::wrappers::{BroadcastStream, UnboundedReceiverStream};
 
 use crate::{
     addressbook::AddressBookHandler,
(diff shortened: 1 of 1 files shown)


nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
```diff
@@ -104,20 +125,52 @@ where
...
+    ) -> Result<SampleStream, StreamSamplingError> {
+        let stream_result = control.open_stream(peer_id, SAMPLING_PROTOCOL).await;
+
+        match stream_result {
+            Ok(stream) => Ok(SampleStream { stream, peer_id }),
+            Err(OpenStreamError::Io(io_err)) if io_err.kind() == ErrorKind::ConnectionReset => {
+                let timer = sleep(CONNECTION_WAIT_TIMEOUT);
+                tokio::pin!(timer);
+
+                loop {
```
bacv: This loop won't return Pending and block until one of select branches is executed. Since swarm run multiple behaviours, it would better be to avoid this.

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
```diff
@@ -104,20 +125,52 @@ where
...
+                    tokio::select! {
+                        res = connection_receiver.recv() => match res {
+                            Ok(pid) if pid == peer_id => break,
```
bacv: We might get connections unrelated to the task being executed, or out of order events because of async nature, one future would get pid, but another wouldn't, because we already skipped some peer_ids that didn't match.

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
pradovic: This is inside a futureunordered. Are you sure it will not return pending until one of the branches is executed? I thought that tokio select is yielding

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
pradovic: This is broadcast, and subscribe should happen before attempting to open a stream the first time. The receiver needs to filter peer_ids it is not interested in, but I think it should get all the peers since the subscribe.
The race condition might be between getting inbound or outbound connection event and next open_stream actually passing.

danielSanchezQ:
> > Does opening a stream fails or waits for dials? If it doesn't wait for dials then we will have enqueue tasks and process them from the queue (?)
> > It does actually try to connect: https://docs.rs/libp2p-stream/0.4.0-alpha/libp2p_stream/struct.Control.html#method.open_stream
> 
> It does, but there is a comment in request_behavior that in that case it might return `ConnectionReset` right away (because the connection takes some time), and then we need to schedule the task to try again. In this PR I tried a simple idea with waiting on a broadcast channel for connected peer event, before trying again, but I am not sure if it is too naive. Would need to test it.

I think its easier to await for a bit and retry the connection again in-place, if after a retry it hasnt connected we jump to next peer.

pradovic:
> > > Does opening a stream fails or waits for dials? If it doesn't wait for dials then we will have enqueue tasks and process them from the queue (?)
> > > It does actually try to connect: https://docs.rs/libp2p-stream/0.4.0-alpha/libp2p_stream/struct.Control.html#method.open_stream
> > 
> > 
> > It does, but there is a comment in request_behavior that in that case it might return `ConnectionReset` right away (because the connection takes some time), and then we need to schedule the task to try again. In this PR I tried a simple idea with waiting on a broadcast channel for connected peer event, before trying again, but I am not sure if it is too naive. Would need to test it.
> 
> I think its easier to await for a bit and retry the connection again in-place, if after a retry it hasn't connected we jump to next peer.

Would that be enough? How long would we wait?

danielSanchezQ:
> @bacv, @danielSanchezQ We went to a similar but a bit different approach in the main PR, where there is a single parent future for historic sampling request, and each of those have futureunordered and spawn new task inside per peer id for shares, and a task for commitments. The flow to cover waiting for connection:
> 
>     1. There is a broadcast sender in the behavior
> 
>     2. On incoming and outgoing connection we broadcast peers that we see to all listeners
> 
>     3. Before behavior.open_stream is called receiver is created by sender.subscribe() and provided to open_stream
> 
>     4. Open stream first tries open_stream and if it gets ConnectionReset it will use this receiver to wait for the events. Events should be there subscribe happened before calling open_stream. We wait on the receiver until we see the peer which signals that connection was established.
> 
>     5. We can (should) also add timeouts too all waits
> 
> 
> Does this makes sense?

Uhm, it does makes sense. But i think in this case we would need a broadcast per peer_id or something like that. Maybe some kind of barrier. Also we probably just want it to be  single time thing, not awaiting in a loop.

pradovic:
> > @bacv, @danielSanchezQ We went to a similar but a bit different approach in the main PR, where there is a single parent future for historic sampling request, and each of those have futureunordered and spawn new task inside per peer id for shares, and a task for commitments. The flow to cover waiting for connection:
> > ```
> > 1. There is a broadcast sender in the behavior
> > 
> > 2. On incoming and outgoing connection we broadcast peers that we see to all listeners
> > 
> > 3. Before behavior.open_stream is called receiver is created by sender.subscribe() and provided to open_stream
> > 
> > 4. Open stream first tries open_stream and if it gets ConnectionReset it will use this receiver to wait for the events. Events should be there subscribe happened before calling open_stream. We wait on the receiver until we see the peer which signals that connection was established.
> > 
> > 5. We can (should) also add timeouts too all waits
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Does this makes sense?
> 
> Uhm, it does makes sense. But i think in this case we would need a broadcast per peer_id or something like that. Maybe some kind of barrier. Also we probably just want it to be single time thing, not awaiting in a loop.

We broadcast all peers, so the caller needs to discard non interested ones, hence the small loop, but tokio::select appears to yield when all arms are pending, so it should not be too problematic, not a busy wait

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
bacv: You are correct, this will not block, tokio::select returns pending on each branch, sorry for this brain fart.

nomos-da/network/core/src/protocols/sampling/historic/request_behaviour.rs
bacv: I now see that it subscribes before dial is called, and since the peer_id will be broadcasted to all tasks, no one should miss the new connection.

PR #1560, opened 2025-08-19
feat(da): Publish blob Tx if dispersal conditions are met

bacv:

Validator publishes blob related transactions only if a certain threshold of shares are met:
- If all shares received - publish immediately
- If threshold is reached - publish after the timeout
- If threshold is not reached - do not publish

Because of dynamic assignations, validator might expect different number of shares depending on the session that the DA network was during the dissemination.  Because of possible lag, we might want to add the session information in the transaction message that executor forms.

For now da network service of validator appends required information for internal services to know how many shares the needs to be expected.

[Spec of publishing constraints](https://www.notion.so/nomos-tech/NomosDA-Dispersal-1fd261aa09df815288c9caf45ed72c95?source=copy_link#22d261aa09df80f1aed0e4bece7cf3cb)

@bacv @danielSanchezQ 

Yes

No

## Checklist

* [x] 2. Description added.


nomos-services/data-availability/verifier/src/backend/trigger.rs
```diff
@@ -0,0 +1,280 @@
...
+
+        if entry.expired.load(Ordering::Acquire) {
+            return ShareState::Expired;
+        }
+
+        let new_count = entry.count.fetch_add(1, Ordering::AcqRel) + 1;
+
+        if new_count >= entry.assignations {
+            ShareState::Complete
+        } else {
```
pradovic, line 88: nitpick: no need for else

nomos-services/storage/src/api/backend/rocksdb/da.rs
```diff
@@ -135,27 +135,48 @@ impl<SerdeOp: StorageSerde + Send + Sync + 'static> StorageDaApi for RocksBacken
...
+
+        let Some(mut assignations) = storage_bytes else {
+            return Ok(None);
+        };
+
+        let tx_bytes = assignations.split_off(2);
+        let assignations = u16::from_be_bytes(
+            assignations[..2]
+                .try_into()
+                .expect("assignations slice was splitoff at 2 bytes"),
```
pradovic: Maybe this expect should behave the same as failure to deserialize below

nomos-services/data-availability/verifier/src/backend/trigger.rs
danielSanchezQ: its needed if you do not want to use `return` in the if statement but rather use it as an expression

nomos-services/storage/src/api/da/mod.rs
```diff
@@ -104,7 +104,15 @@ pub trait StorageDaApi {
         shared_commitments: Self::Commitments,
     ) -> Result<(), Self::Error>;
 
-    async fn store_tx(&mut self, blob_id: Self::BlobId, tx: Self::Tx) -> Result<(), Self::Error>;
+    async fn store_tx(
+        &mut self,
+        blob_id: Self::BlobId,
+        assignations: u16,
+        tx: Self::Tx,
+    ) -> Result<(), Self::Error>;
```
danielSanchezQ, line 112: We store the assignations just in case we need to recover the node? This will probably messed up other places if we need such tx alone 🤔

nomos-services/data-availability/verifier/src/backend/trigger.rs
```diff
@@ -0,0 +1,280 @@
...
+struct ShareEntry {
+    count: AtomicU16,
+    created_at: Instant,
+    assignations: u16,
+    expired: AtomicBool,
+}
+
+pub struct MempoolPublishTrigger<Id> {
+    config: MempoolPublishTriggerConfig,
+    received: Arc<RwLock<HashMap<Id, Arc<ShareEntry>>>>,
```
danielSanchezQ, line 53: Can we somehow make this without locking? Its fine as this for now anyway, we can think on improving it if we have a bottleneck.

danielSanchezQ (approved):
Leaving a couple concerns, otherwise is probably good to go. Thanks!

nomos-services/storage/src/api/da/mod.rs
bacv: This is only for DA related pending transactions storage. We won't use this `store_tx` anywhere else apart from DA, and in that case we'll just use the `(_, tx)`. If node is down for less than a period of "share_duration", then it would be able to receive and process the ~~TX~~ shares.

nomos-services/data-availability/verifier/src/backend/trigger.rs
bacv: Yeah, this locks only on inserting new item, it's not locking when updating an existing one (the hot path). We can optimize later but not sure what that might be.

bacv:
Merging this, the appended assignations upon TX reception is temporal and will be removed after this TX dissemination feature is merged into the master.

nomos-services/storage/src/api/da/mod.rs
bacv: This is temporal and will be removed, assignations will be reconstructed from the session number in the tx.

PR #1568, opened 2025-08-20
feat(sync): improve IBD test

andrussal:

Improve IBD test accuracy and adjust timeout for slow CI env.

@youngjoon-lee I remember we decided that it's accurate to check max height of initial nodes. But could there a valid reason when behind node is able to add only chain from the node of smaller height?

```
Ok(Response { url: "http://127.0.0.1:8507/cryptarchia/info", status: 200, headers: {"content-type": "application/json", "content-length": "112", "date": "Wed, 20 Aug 2025 14:00:38 GMT"} })
   Initial validators: [9/Online, 12/Online]
Ok(Response { url: "http://127.0.0.1:8506/cryptarchia/info", status: 200, headers: {"content-type": "application/json", "content-length": "112", "date": "Wed, 20 Aug 2025 14:00:39 GMT"} })
Ok(Response { url: "http://127.0.0.1:8507/cryptarchia/info", status: 200, headers: {"content-type": "application/json", "content-length": "112", "date": "Wed, 20 Aug 2025 14:00:39 GMT"} })
   Initial validators: [10/Online, 12/Online]
   All validators reached are in mode Online
Starting behind node with IBD peers...
Behind node started, waiting for it to sync...
Ok(Response { url: "http://127.0.0.1:8506/cryptarchia/info", status: 200, headers: {"content-type": "application/json", "content-length": "112", "date": "Wed, 20 Aug 2025 14:00:42 GMT"} })
Ok(Response { url: "http://127.0.0.1:8507/cryptarchia/info", status: 200, headers: {"content-type": "application/json", "content-length": "112", "date": "Wed, 20 Aug 2025 14:00:42 GMT"} })
Ok(Response { url: "http://127.0.0.1:8509/cryptarchia/info", status: 200, headers: {"content-type": "application/json", "content-length": "119", "date": "Wed, 20 Aug 2025 14:00:42 GMT"} })
behind node info: CryptarchiaInfo { tip: HeaderId([227, 205, 58, 85, 77, 106, 72, 68, 118, 3, 220, 96, 52, 8, 125, 88, 9, 111, 29, 76, 102, 203, 8, 236, 134, 20, 170, 108, 85, 176, 194, 86]), slot: Slot(21), height: 10, mode: Bootstrapping }
slot_duration:5s, active_slot_coeff:0.9
Estimated block time: 5.555555556s
Estimated height margin: 1
Checking if the behind node has caught up to the highest initial validator

thread 'test_ibd_behind_nodes' panicked at tests/src/tests/cryptarchia/bootstrap.rs:94:5:
assertion failed: behind_node_info.height.abs_diff(*max_initial_validator_height) <=
    height_margin
```    
    

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/tests/src/tests/cryptarchia/bootstrap.rs
+++ b/tests/src/tests/cryptarchia/bootstrap.rs
@@ -59,8 +59,9 @@ async fn test_ibd_behind_nodes() {
     config.cryptarchia.bootstrap.ibd.peers = initial_peer_ids.clone();
     // Shorten the delay to quickly catching up with peers that grow during IBD.
     // e.g. We start a download only for peer1 because two peers have the same tip
-    //      at the moment. But, the peer2 may grow faster than peer1 before IBD is done.
-    //      So, we want to check peer1's progress frequently with a very short delay.
+    //      at the moment. But, the peer2 may grow faster than peer1 before IBD is
+    // done.      So, we want to check peer1's progress frequently with a very
+    // short delay.
     config.cryptarchia.bootstrap.ibd.delay_before_new_download = Duration::from_millis(10);
     // Disable the prolonged bootstrap period for the behind node
     // because we want to check the height of the behind node
--- a/tests/src/topology/configs/network.rs
+++ b/tests/src/topology/configs/network.rs
@@ -38,7 +38,7 @@ pub fn create_network_configs(
                 node_key,
                 port: get_available_port(),
                 chain_sync_config: cryptarchia_sync::Config {
-                    peer_response_timeout: Duration::from_secs(5),
+                    peer_response_timeout: Duration::from_secs(60),
                 },
                 ..Default::default()
             }


tests/src/tests/cryptarchia/bootstrap.rs
```diff
@@ -91,12 +91,7 @@ async fn test_ibd_behind_nodes() {
 
     println!("Checking if the behind node has caught up to the highest initial validator");
 
-    assert!(
-        behind_node_info
-            .height
-            .abs_diff(*max_initial_validator_height)
-            <= height_margin
-    );
+    assert!(behind_node_info.height >= *min_initial_validator_height);
```
youngjoon-lee: I think this change is still not safe because initial validators can grow before we perform this assert, though the probability is very low.

Also, we want to test whether the behind node tries its best to catch up **all the peers**.

So, I'd like to suggest a new approach. I opened a branch for it: https://github.com/logos-co/nomos/commit/950eb2535be4a7d6c59eae188a8acf8ea8a092ed. Please have a look.

That branch has two key changes:

1. Instead of sleeping 1s to wait until the behind node finishes IBD, now we wait until the behind node switches to the Online mode. It happens only after IBD is done. This approach is more accurate.
To make the test faster, I disabled the prolonged bootstrap period only for the behind node. By doing so, the behind node switches to the Online mode immediately after IBD is done.
After that, we check whether the behind node is as high as the **highest** peers with an acceptable margin.
We need to use the margin because **any** node can grow right before we compare the heights.

2. I set a very short `delay_before_new_download` for the behind node.
There was a subtle timing issue in the test. For example, if two initial peers have the same height, the behind node starts a download only for peer1 and schedules a delay (10s) for peer2. But, peer2 can grow immediately. Our test didn't cover this case. Our test slept just for 1s and check if the behind node catches up the highest peer (i.e. peer2). But, it's false in this example.
So, along with the change `1` above, I shorten the delay, so that the behind node frequently checks the progress of peer2 and catches up with it (without finishing IBD).

andrussal: Thank you @youngjoon-lee. I agree that these changes make the test more accurate. 1)We want to make sure that IBD is actually finished before doing checks 2)We want to make sure that IBD had the chance to catch up with most recent state. Do you think we should integrate your changes here or you wanna create another branch?

I think there might be something else that causes this specific failure because before IBD starts, heights are 10/12 and behind node reaches only 10. But I haven't been able to reproduce it anymore, although at one point in CI both linux and mac failed at the same time with this error. This sounds like some env related issue. sync timeout is 5s which is long but we have seen that sometimes 30s is not enough a node to receive/create blocks. So potentially there could be some timeout issue...

andrussal: I think lets add your changes and also make timeout much longer in CI.

tests/src/common/sync.rs
```diff
@@ -14,23 +15,42 @@ pub async fn wait_for_validators_mode_and_height(
...
+        print_validators_info(&infos);
+
+        if infos.iter().all(|info| info.mode == mode) {
+            println!("   All validators reached are in mode {mode:?}");
+            break;
+        }
+
+        tokio::time::sleep(Duration::from_secs(1)).await;
+    }
+}
```
danielSanchezQ, line 46: Nice.

PR #1569, opened 2025-08-20
feat(nat): gateway monitor

andrussal:

This PR adds gateway monitor to NAT subsystem. We need periodically check if default gateway is changed and request new mapping if this is the case.

Yes. Also please see NAT [spec](https://www.notion.so/Nomos-P2P-Network-NAT-Solution-1d28f96fb65c80198b03f777be835967)

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-libp2p/src/behaviour/nat/inner.rs
+++ b/nomos-libp2p/src/behaviour/nat/inner.rs
@@ -38,8 +38,6 @@ type Task = BoxFuture<'static, Multiaddr>;
 pub struct InnerNatBehaviour<R, Mapper, Detector>
 where
     R: RngCore + 'static,
-    Mapper: NatMapper,
-    Detector: GatewayDetector,
 {
     /// `AutoNAT` client behaviour which is used to confirm if addresses of our
     /// node are indeed publicly reachable.
@@ -85,8 +83,6 @@ impl<R: RngCore + 'static> NatBehaviour<R> {
 impl<R, Mapper, Detector> InnerNatBehaviour<R, Mapper, Detector>
 where
     R: RngCore + 'static,
-    Mapper: NatMapper,
-    Detector: GatewayDetector,
 {
     fn create(
         rng: R,
--- a/nomos-libp2p/src/behaviour/nat/state_machine/states.rs
+++ b/nomos-libp2p/src/behaviour/nat/state_machine/states.rs
@@ -185,7 +185,6 @@ impl MappedPublic {
         Private { local_address }
     }
 
-
     pub fn into_try_map_address(self) -> TryMapAddress {
         let Self { local_address, .. } = self;
         TryMapAddress {


nomos-libp2p/src/behaviour/nat/gateway_monitor.rs
```diff
@@ -0,0 +1,178 @@
...
+        new_gateway: IpAddr,
+    },
+}
+
+pub trait GatewayDetector: Send + Sync {
+    /// Detect the current gateway address
+    fn detect() -> Result<IpAddr, String>;
+}
+
+pub struct SystemGatewayDetector;
```
ntn-x2, line 35: Maybe a small comment describing how this type detects the gateway would help in the future.

nomos-libp2p/src/behaviour/nat/gateway_monitor.rs
```diff
@@ -0,0 +1,178 @@
...
+impl GatewayDetector for SystemGatewayDetector {
+    fn detect() -> Result<IpAddr, String> {
+        default_net::get_default_gateway()
+            .map(|gateway| gateway.ip_addr)
+            .map_err(|e| format!("Failed to get default gateway: {e}"))
+    }
+}
+
+/// Gateway monitoring behavior that periodically checks for gateway changes
+pub struct GatewayMonitor<Detector: GatewayDetector> {
```
ntn-x2: No need for the trait bound already on the struct declaration.

nomos-libp2p/src/behaviour/nat/gateway_monitor.rs
```diff
@@ -0,0 +1,178 @@
...
+            }
+        }
+    }
+
+    fn check_gateway(&mut self) -> Option<GatewayMonitorEvent> {
+        match Detector::detect() {
+            Ok(new_gateway) => match self.current_gateway {
+                Some(old_gateway) if old_gateway != new_gateway => {
+                    info!("Gateway address changed from {old_gateway} to {new_gateway}");
+                    self.current_gateway = Some(new_gateway);
```
ntn-x2: I think we can reduce this and `perform_initial_check` to a single function that does what `perform_initial_check` does, basically, but that returns the new address instead. Then, we can use the result to update the struct and generate the event if needed. It feels like duplicated logic to me.

nomos-libp2p/src/behaviour/nat/inner.rs
```diff
@@ -19,31 +19,38 @@ use libp2p::{
...
 };
 
 type Task = BoxFuture<'static, Multiaddr>;
 
-/// Provides dynamic NAT-status detection, improvement (via address mapping on
-/// the NAT-box), and periodic maintenance capabilities.
-pub struct InnerNatBehaviour<R: RngCore + 'static> {
+pub struct InnerNatBehaviour<R, Mapper, Detector>
+where
+    R: RngCore + 'static,
```
ntn-x2, line 40: Again, I think we can reduce the number of trait bounds for struct declarations, which will get less in the way when we actually don't need them.

andrussal: I think these can come handy later when we want to add tests? I also don't like too many bounds but I am not sure it will be possible to test it easily otherwise.

ntn-x2: Mhh but in testing we always know the concrete types, so the compiler is able to figure out whether a given method is available on the struct based on the type used.

andrussal: ah I see what you mean. I removed unnecessary bounds.

nomos-libp2p/src/behaviour/nat/gateway_monitor.rs
```diff
@@ -0,0 +1,187 @@
...
+            info!("Initial gateway detected: {gateway}");
+            monitor.current_gateway = Some(gateway);
+        }
+
+        monitor
+    }
+
+    pub fn poll(&mut self, cx: &mut Context<'_>) -> Poll<Option<GatewayMonitorEvent>> {
+        if self.check_timer.as_mut().poll_unpin(cx).is_ready() {
+            self.check_timer = Box::pin(tokio::time::sleep(self.settings.check_interval));
```
ntn-x2, line 82: Just a nit: since we instantiate the timer in two places, might be good to factor it out into a function with a `check_interval` input so we avoid duplication.

ntn-x2 (approved):
Looks good!

PR #1571, opened 2025-08-21
feat(da): Validate before replicating data

bacv:

DA Dispersal specification requires data to be validated before replicating it to subnetwork peers. This change introduces validation response channel to validator dispersal events.

[Spec section](https://www.notion.so/nomos-tech/NomosDA-Subnetwork-Replication-1fd261aa09df811d93f8c6280136bfbb?source=copy_link#1fd261aa09df8141b387d9bfd5c3749f)

@danielSanchezQ @bacv 

Yes

No

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nodes/nomos-executor/config.yaml
+++ b/nodes/nomos-executor/config.yaml
@@ -75,9 +75,6 @@ da_network:
         num_of_subnets: 20
         shares_retry_limit: 5
         commitments_retry_limit: 5
-      refresh_interval:
-        secs: 30
-        nanos: 0
     num_subnets: 2
   membership:
     subnetwork_size: 2
@@ -88,6 +85,9 @@ da_network:
   api_adapter_settings:
     api_port: 8722
     is_secure: false
+  subnet_refresh_interval:
+    secs: 30
+    nanos: 0
 da_dispersal:
   backend:
     encoder_settings:
--- a/nodes/nomos-executor/config.yaml
+++ b/nodes/nomos-executor/config.yaml
@@ -332,6 +332,9 @@ storage:
 mempool:
   cl_pool_recovery_path: ./recovery/cl_mempool.json
   da_pool_recovery_path: ./recovery/da_mempool.json
+  trigger_sampling_delay:
+    secs: 5
+    nanos: 0
 membership:
   backend:
     settings_per_service: {}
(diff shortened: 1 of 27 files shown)


nodes/nomos-executor/config.yaml
```diff
@@ -332,6 +332,9 @@ storage:
 mempool:
   cl_pool_recovery_path: ./recovery/cl_mempool.json
   da_pool_recovery_path: ./recovery/da_mempool.json
+  trigger_sampling_delay:
+    secs: 5
```
danielSanchezQ, line 336: Isn't this too much? Maybe for local tests is fine tho, as everything goes too quick there.

nomos-services/data-availability/network/src/lib.rs
```diff
@@ -274,18 +285,28 @@ where
...
         );
 
+        // Sampling subnetwork peers need to be updatedd periodically.
+        // They also need to be updated when the assignations change.
+        let (subnet_refresh_sender, refresh_rx) = mpsc::channel(1);
+        let interval = tokio::time::interval(settings.subnet_refresh_interval);
+        let refresh_ticker = IntervalStream::new(interval).map(|_| ());
+        let refresh_signal = ReceiverStream::new(refresh_rx);
+        let subnet_refresh_signal = select(refresh_ticker, refresh_signal);
+
```
danielSanchezQ, line 295: This should be based on the period of membership service iirc. Or at least from the time service. It definetely should not be a configurable time (as it need to be in sync).

danielSanchezQ (commented):
Nice stuff. We need to discuss about the assignations refresh interval. If this was a temporary solution we can go ahead opening an issue to be resolved next. Otherwise we need to change it.

nodes/nomos-executor/config.yaml
bacv: Yes, it's a bit much, and we probably want to have a dedicated service message `AddSamplingDelay` for a mempool to not delay blob items from network, just when publishing from within ourselves.

nomos-services/data-availability/network/src/lib.rs
bacv: Added a task https://github.com/logos-co/nomos/issues/1576

PR #1573, opened 2025-08-22
feat(nat): pcp core protocol

andrussal:

This PR implements PCP nat mapping protocol according to PCP specification.

**PCP protocol features**

**Implemented**

  - Server Discovery: Detect and connect to PCP-enabled router
  - Port Mappings: Create and remove port mapping

 **Not implemented(low importance)**

  - Mapping Renewal: Automatically renew mappings before they expire (manual renewal sufficient for current use cases)
  - Peer Mappings: Direct peer-to-peer connection setup (rarely used in practice)
  - IPv6 Support: Full IPv6 router communication (IPv4 covers majority of deployments)
  - Advanced Options: Extended filtering and preference controls (basic mapping meets most requirements)
  - Multiple Routers: Redundancy across multiple gateway devices (single router setup is standard)
  - Version Support: Compatibility with older PCP implementations (PCP v2 is widely adopted)

Yes. It is not expected to read the spec, implementation is tested against gateway.  Spec: https://tools.ietf.org/html/rfc6887

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


nomos-libp2p/src/swarm.rs
```diff
@@ -14,39 +15,30 @@ use std::{
...
-    #[error("no known peers")]
-    NoKnownPeers,
-}
-
 /// How long to keep a connection alive once it is idling.
 const IDLE_CONN_TIMEOUT: Duration = Duration::from_secs(300);
 
 /// Wraps [`libp2p::Swarm`], and config it for use within Nomos.
-pub struct Swarm {
+pub struct Swarm<R: Clone + Send + RngCore + 'static> {
```
ntn-x2:
```suggestion
pub struct Swarm<Rng: Clone + Send + RngCore + 'static> {
```

nomos-libp2p/src/lib.rs
```diff
@@ -3,7 +3,11 @@ mod config;
 pub mod protocol_name;
 mod swarm;
 
-pub use config::{secret_key_serde, IdentifySettings, KademliaSettings, SwarmConfig};
+pub use config::{
+    secret_key_serde, AutonatClientSettings, IdentifySettings, KademliaSettings, NatSettings,
+    SwarmConfig,
+};
+pub use cryptarchia_sync::{self, Event};
```
ntn-x2: Why this?

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/wire.rs
```diff
@@ -0,0 +1,397 @@
...
+    /// PCP protocol version (must be 2)
+    pub version: u8,
+    /// Operation code (0 for ANNOUNCE)
+    pub opcode: u8,
+    /// Reserved field (must be 0)
+    pub reserved: [u8; 2],
+    /// Requested lifetime in seconds (must be 0 for ANNOUNCE)
+    pub lifetime: U32NE,
+    /// Client IPv4 address as IPv4-mapped IPv6
+    pub client_ip: [u8; 16],
```
ntn-x2: I think at least this deserves to be its own type. What do you think?

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/wire.rs
```diff
@@ -0,0 +1,397 @@
...
+        request.version = PCP_VERSION;
+        request.opcode = OPCODE_ANNOUNCE;
+        request.lifetime.set(0);
+        request.client_ip = ipv4_to_ipv6_mapped(client_addr);
+        request
+    }
+}
+
+impl PcpAnnounceResponse {
+    pub fn validate(data: &[u8]) -> Result<(), PcpError> {
```
ntn-x2: Why not implementing `TryFrom<Vec<u8>>` for this type?

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/wire.rs
```diff
@@ -0,0 +1,397 @@
...
+    ExcessiveRemotePeers = 13,
+}
+
+/// ANNOUNCE request packet (24 bytes total)
+///
+/// Used for server discovery and capability negotiation.
+/// Lifetime field must be 0 for requests.
+#[derive(Debug, FromBytes, IntoBytes, Unaligned, Immutable)]
+#[repr(C)]
+pub struct PcpAnnounceRequest {
```
ntn-x2: I think it would be better to declare two submodules, one for announce and one for map operations.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/wire.rs
```diff
@@ -0,0 +1,397 @@
...
+    /// Operation code (1 for MAP)
+    pub opcode: u8,
+    /// Reserved field (must be 0)
+    pub reserved1: [u8; 2],
+    /// Requested lifetime in seconds (0=delete)
+    pub lifetime: U32NE,
+    /// Client IPv4 address as IPv4-mapped IPv6
+    pub client_ip: [u8; 16],
+    /// Random 96-bit nonce for request/response matching
+    pub nonce: [u8; 12],
```
ntn-x2: As far as I understand from the high-level read I gave to the spec, from this point on this is opcode-specific information. Would it make sense to have a generic container for requests that is generic of the specific request type, and then specify only announce-specific and map-specific structures? The first would be an empty tuple, the second would have all these details here. We can then expose methods on the generic container to return instances of a specified type. I think it would be better encapsulation and we would reduce code duplication.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/wire.rs
```diff
@@ -0,0 +1,397 @@
...
+    /// Transport protocol (6=TCP, 17=UDP)
+    pub protocol: u8,
```
ntn-x2: Why not the `Protocol` enum defined above?

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/wire.rs
```diff
@@ -0,0 +1,397 @@
...
+            .map_err(|e| PcpError::ParseError(format!("Failed to parse ANNOUNCE response: {e}")))?;
+
+        if response.version != PCP_VERSION {
+            return Err(PcpError::InvalidResponse(format!(
+                "Version mismatch: expected {}, got {}",
+                PCP_VERSION, response.version
+            )));
+        }
+
+        if response.opcode & OPCODE_RESPONSE_MASK != OPCODE_ANNOUNCE {
```
ntn-x2: Could this `response.opcode & OPCODE_RESPONSE_MASK` be a method on the `response` type instead?

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/wire.rs
ntn-x2: Also, probably a type alias for all these types would help better understanding.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/client.rs
```diff
@@ -0,0 +1,339 @@
...
+    /// Maximum number of retry attempts
+    pub max_retries: usize,
+    /// Request timeout duration
+    pub request_timeout: Duration,
+}
+
+impl Default for PcpConfig {
+    fn default() -> Self {
+        Self {
+            mapping_lifetime: DEFAULT_MAPPING_LIFETIME,
```
ntn-x2: Why is only this a constant? I think we don't need to make it a constant at all.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/client.rs
```diff
@@ -0,0 +1,339 @@
...
+
+pub const PCP_PORT: u16 = 5351;
+pub const DEFAULT_MAPPING_LIFETIME: u32 = 7200; // 2 hours
+
+#[derive(Debug, Clone)]
+pub struct PcpConfig {
+    /// Default mapping lifetime in seconds
+    pub mapping_lifetime: u32,
+    /// Initial retry delay in milliseconds
+    pub initial_retry_ms: u64,
```
ntn-x2: Why not make this and `max_retry_delay` `Duration`s?

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/client.rs
```diff
@@ -0,0 +1,339 @@
...
+            max_retries: 5,
+            request_timeout: Duration::from_secs(1),
+        }
+    }
+}
+
+/// PCP (RFC 6887) client for NAT port mapping.
+///
+/// Protocol flow:
+/// 1. ANNOUNCE: Discovers PCP server and learns external IP (0-byte lifetime)
```
ntn-x2: I think that's part of the spec, we probably do not need to specify it here.

```suggestion
/// 1. ANNOUNCE: Discovers PCP server and learns external IP
```

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/connection.rs
```diff
@@ -0,0 +1,99 @@
...
+/// Implements exponential backoff as specified in RFC 6887 Section 8.1.1:
+/// "Clients should use exponential backoff with jitter to avoid thundering
+/// herd"
+fn create_backoff(
+    initial_retry_ms: u64,
+    max_delay_secs: u64,
+    max_retries: usize,
+) -> ExponentialBuilder {
+    ExponentialBuilder::default()
+        .with_min_delay(Duration::from_millis(initial_retry_ms))
```
ntn-x2: We should directly accept `Duration` as arguments, I would say.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/connection.rs
```diff
@@ -0,0 +1,99 @@
...
+        config: PcpConfig,
+    ) -> Result<Self, PcpError> {
+        let socket = UdpSocket::bind((client_addr, ANY_PORT)).await?;
+
+        socket.connect(server_addr).await?;
+
+        Ok(Self { socket, config })
+    }
+
+    pub async fn send_with_retry(&self, request: &[u8]) -> Result<Vec<u8>, PcpError> {
```
ntn-x2: This is a `pub` function, we should aim to have strong types. We can accept a request that implements the required trait, and then call `.as_bytes(0)` internally. Else, I could send `0x010101010`, which the compiler should at least try to prevent me from doing.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/connection.rs
```diff
@@ -0,0 +1,99 @@
...
+                }
+                Ok(Err(e)) => Err(PcpError::from(e)),
+                Err(elapsed) => Err(PcpError::Timeout(elapsed)),
+            }
+        })
+        .retry(&backoff)
+        .await
+    }
+
+    pub async fn send_announce_request(
```
ntn-x2: As far as I got from the spec, announce requests are sent by the PCP server to the client's port `5050`, while map responses are sent to the port they originated from. If my understanding is correct, we need two different sockets, or at least we would need to keep one socket open for announce responses and open a socket on demand to send a map request. Am I missing something?

ntn-x2: Or in this case we open sockets on demand regardless, so we should probably hard code the UDP port number fo announce requests and let the OS choose a port for map requests.

ntn-x2: But that should probably be done in the `new` method, by replacing it with `new_announce_connection` and `new_map_connection`.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/connection.rs
```diff
@@ -0,0 +1,99 @@
...
+        &self,
+        request: &PcpAnnounceRequest,
+    ) -> Result<Vec<u8>, PcpError> {
+        self.send_with_retry(request.as_bytes()).await
+    }
+
+    pub async fn send_map_request(&self, request: &PcpMapRequest) -> Result<Vec<u8>, PcpError> {
```
ntn-x2: Would it make sense to make this function generic over the return type? So we can call it with the type we expect to get back from the request, and fail otherwise? In that case, we could merge `send_map_request` and `send_announce_request` into a single function that takes two generics: an input type and an expected response type. If we want to keep the two methods separated for better naming, then we could return the expected response type instead of `Vec<u8>`.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/client.rs
```diff
@@ -0,0 +1,339 @@
...
+            mapping.nonce,
+        );
+
+        let response = connection.send_with_retry(request.as_bytes()).await?;
+
+        Self::validate_delete_response(&response, &request.nonce)
+    }
+
+    #[cfg(test)]
+    fn validate_delete_response(
```
ntn-x2: This could also be a test-gated `TryFrom` implementation for a test-only struct called `DeleteResponse`.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/mapping.rs
```diff
@@ -0,0 +1,464 @@
...
+    pub lifetime_seconds: u32,
+    #[cfg_attr(not(test), expect(dead_code, reason = "Part of protocol"))]
+    pub nonce: [u8; 12],
+    #[cfg_attr(not(test), expect(dead_code, reason = "Part of protocol"))]
+    pub epoch_time: u32,
+}
+
+impl Mapping {
+    pub fn from_map_response(
+        data: &[u8],
```
ntn-x2: Why not pass the `PcpMapResponse` type directly here?

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/connection.rs
danielSanchezQ: yes, it is easy to miss and send 10 expecting it to be ms and internally its instead used as seconds.

danielSanchezQ (commented):
Neat use of zerocopy crate here. I agree with @ntn-x2 points to be solved. Overall it looks quite good anyway. I wonder if it would be ok to publish this as a public crate.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+//!   adequate)
+//! - **Version Negotiation**: Automatic fallback to older versions (PCP v2
+//!   universally supported)
+//!
+//! ## Error Handling
+//!
+//! PCP defines standard result codes (0=success, 1-13=various errors).
+//! Unknown codes are logged and treated as `InvalidResponse` errors.
+//! Network errors trigger retry with increasing delays (250ms → 1s, max 5
+//! attempts).
```
ntn-x2: Isn't this configurable? Maybe we can just remove this line.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/client.rs
```diff
@@ -0,0 +1,281 @@
...
+}
+
+/// PCP (RFC 6887) client for NAT port mapping.
+///
+/// Protocol flow:
+/// 1. ANNOUNCE: Discovers PCP server and learns external IP
+/// 2. MAP: Creates port mappings with nonce-based request/response matching
+/// 3. DELETE: Removes mappings via MAP with 0-byte lifetime
+///
+/// All requests use UDP port 5351 with exponential backoff retry.
```
ntn-x2:
```suggestion
/// All requests use server UDP port 5351 with exponential backoff retry.
```

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/connection.rs
```diff
@@ -0,0 +1,93 @@
...
+
+        (|| async {
+            self.socket.send(request.as_bytes()).await?;
+
+            let mut buffer = vec![0u8; BUFFER_SIZE];
+
+            match timeout(self.config.request_timeout, self.socket.recv(&mut buffer)).await {
+                Ok(Ok(size)) => {
+                    buffer.truncate(size);
+                    Response::try_from(buffer.as_slice())
```
ntn-x2: Just a nit, I think we could also not require the response to implement `TryFrom` with a `PcpError`, but could map it ourselves internally into an `InvalidResponse` variant. But this is also ok, although somewhat slightly less flexible.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/connection.rs
```diff
@@ -0,0 +1,93 @@
...
+///
+/// Handles connection to PCP server with exponential backoff retry.
+/// Each connection is bound to client IP and connected to server.
+pub struct PcpConnection {
+    socket: UdpSocket,
+    config: PcpConfig,
+}
+
+impl PcpConnection {
+    pub async fn connect(
```
ntn-x2: `PcpConnection::connect` sounds redundant since we are still creating a connection object. What about `PcpConnection::establish`? Or `PcpConnection::open`

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/client.rs
```diff
@@ -0,0 +1,281 @@
...
+    pub async fn probe_available(self) -> Result<ActivePcpClient, PcpError> {
+        let gateway_ip = Self::get_default_gateway()?;
+
+        tracing::debug!("Probing PCP support at gateway {gateway_ip}");
+
+        let connection =
+            PcpConnection::connect(self.client_addr, gateway_ip, self.config.clone()).await?;
+
+        let request = PcpAnnounceRequest::new(self.client_addr);
+        let _: PcpAnnounceResponse = connection.send_request(&request).await?;
```
ntn-x2: Maybe add a comment specifying that since we got a valid response back from the gateway, it means it's reachable and speaks the PCP protocol. Else it might not be immediate to understand.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/wire/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+
+#[derive(Debug, FromBytes, IntoBytes, Unaligned, Immutable, KnownLayout)]
+#[repr(C)]
+pub struct PcpResponse<Payload> {
+    pub header: PcpResponseHeader,
+    pub payload: Payload,
+}
+
+impl<Payload> PcpRequest<Payload> {
+    pub fn new_with_header(client_addr: Ipv4Addr, opcode: u8, lifetime: u32) -> Self
```
ntn-x2: Question: is this somehow required by the `zerocopy` crate? Why can't we have a function that takes the payload and returns a fully composed request instead of a request with a zeroed payload that needs filling afterwards? The way I see it, this is error prone since the compiler will not complain at you if you try to then send over the wire a half-created request.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/mapping.rs
```diff
@@ -0,0 +1,338 @@
...
+use zerocopy::FromBytes as _;
+
+use crate::behaviour::nat::address_mapper::protocols::pcp::{
+    client::PcpError,
+    wire::{Ipv4MappedIpv6, PcpMapResponse, Protocol, ResultCode, PCP_MAP_SIZE},
+};
+
+#[derive(Debug)]
+pub struct Mapping {
+    #[cfg_attr(not(test), expect(dead_code, reason = "Part of protocol"))]
```
ntn-x2: There is a lint that we have not enabled (yet), that will discourage using `#[cfg(not(test))]` for the reasons explained in the Clippy lint: https://rust-lang.github.io/rust-clippy/master/index.html#cfg_not_test. I believe this would trigger that lint as well. Could this be worked around in a different way than marking fields as non-test only (or silence lints only for non-test builds)? For instance, if the fields are only accessed in tests, then they could all be marked as private, and we could expose public getters that would be called in tests to check the relevant conditions.

ntn-x2: Also I think in general it's good not to make these fields public, I would argue, but rather expose the relevant methods, since I doubt consumers of this mapping are interested in knowing all the details about it.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/connection.rs
ntn-x2: I re-read the spec, and port `5050` is used for multicast messages from the PCP server to clients, and since we don't support those, this is fine to ignore.

ntn-x2 (commented):
I like the new structure a lot, thank you for that! I left some more comments. Most of them are nits, which you can accept or ignore, and there's a couple about suggested improvements.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/wire/mod.rs
andrussal: > Question: is this somehow required by the `zerocopy` crate? Why can't we have a function that takes the payload and returns a fully composed request instead of a request with a zeroed payload that needs filling afterwards? The way I see it, this is error prone since the compiler will not complain at you if you try to then send over the wire a half-created request.

I think we can restructure this, let me try.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/connection.rs
andrussal: > `PcpConnection::connect` sounds redundant since we are still creating a connection object. What about `PcpConnection::establish`? Or `PcpConnection::open`

Agree, probably  `establish` or `open` are a bit nicer here as a `PcpConnection` method. `connect` is something that would be external to `PcpConnection`.

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+//! - **Multiple Servers**: Multi-server redundancy and failover (single server
+//!   adequate)
+//! - **Version Negotiation**: Automatic fallback to older versions (PCP v2
+//!   universally supported)
+//!
+//! ## Error Handling
+//!
+//! PCP defines standard result codes (0=success, 1-13=various errors).
+//! Unknown codes are logged and treated as `InvalidResponse` errors.
+//! Network errors trigger retry with increasing delays
```
ntn-x2:
```suggestion
//! Network errors trigger retry with increasing delays.
```

nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp/connection.rs
```diff
@@ -0,0 +1,98 @@
...
+        Ok(Self { socket, config })
+    }
+
+    pub(super) async fn send_request<Request, Response>(
+        &self,
+        request: &Request,
+    ) -> Result<Response, PcpError>
+    where
+        Request: IntoBytes + Immutable + Sync,
+        Response: for<'a> TryFrom<&'a [u8]>,
```
ntn-x2: With this you can remove the line after this to specify a bound on the `Error` associated type

```suggestion
        Response: for<'a> TryFrom<&'a [u8], Error: Debug>,
```

ntn-x2 (approved):
Looks  very good to me! Thanks for this effort, it will be extremely useful!

PR #1574, opened 2025-08-22
feat(nat): integrate pcp mapping

andrussal:

This PR integrates PCP mapping protocol to nat mapping flow.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


nomos-libp2p/src/behaviour/nat/address_mapper/protocols/pcp.rs
```diff
@@ -0,0 +1,134 @@
...
+
+        let config = PcpConfig {
+            request_timeout: settings.timeout,
+            ..PcpConfig::default()
+        };
+
+        let client = PcpClient::new(local_ip, config)
+            .map_err(|e| AddressMapperError::GatewayDiscoveryFailed(e.to_string()))?;
+
+        let is_available = timeout(settings.timeout, client.probe_available())
```
ntn-x2: I understand the `announce` function on the client is then perhaps meant to be used internally only? In that case it should probably not be made `pub`.

ntn-x2: I am commenting here stuff that is probably related to the PCP PR, but since I am now seeing how things are meant to be used. Would it make more sense to have an infallible `new`, that does not try to reach the gateway, but then have the `probe_available` return something which includes the gateway address and on which `request_mapping` can be called? So the flow would be `PcpClient::new(...).probe_avilable().await.request_mapping`. So that we can't try to map if there is no gateway that speaks PCP at all. What do you think?

ntn-x2: Cause as far as I can see, it would be fine to call `PcpClient::new(...).await.unwrap().request_mapping` even if we don't know whether the gateway supports PCP at all?

PR #1575, opened 2025-08-22
feat(blend): session transition: core behaviours

youngjoon-lee:

This is the 1st PR of the session transition ([spec](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df8156a1d2dc9e7f81d289)).

Previously, I opened a PR https://github.com/logos-co/nomos/pull/1548, but closed it and opened this, which is an improved version.

This PR adds the session transition functionality to the `with_core` and `with_edge` network behaviour. This PR doesn't integrate this functionality to the blend core service yet. It will be done by a next PR.

## Functionalities

- `with_core` behaviour
  - Before the transition period passes, messages from the connections in the old session are forwarded to the other connections in the old session. 
  - However, we don't accept any new connection upgrades for the old session.
  - New connection establishments are always considered as for the new (current) session.
  - Also, any events related to the connection maintenance are ignored if they are for the connections in the old session.
    - Some of those events are bubbled up from the connection handler, which doesn't differentiate sessions. Those events are received by the swarm, and forwarded to the behaviour. Then, the behaviour ignore the events if they're not for the current session.
- `with_edge` behaviour
  - As soon as a new session is started, we close all the connections in the old session. They are just one-time connections. It's better to let the peer rebuild a message and re-establish a connection with the new membership. No need to wait for the transition period.

## Architecture Overview

The `with_edge` behaviour is super simple. So, I want to just explain little bit about the `with_core` behaviour.

Most of the logic in the `Behaviour` remains as is. They work only for the current session.
Additionally, I defined the `OldSession` struct, which is responsible for receiving/forwarding messages **within** the old session. The logic in the `OldSession` is very similar as the `Behaviour` but much simpler.

```rust
pub struct Behaviour<...> {
  negotiated_peers: ...,
  exchanged_message_identifiers: ...,
  old_session: OldSession,
}

struct OldSession {
  negotiated_peers: ...,
  exchanged_message_identifiers: ...,
}
```
I believe this PR is more readable and maintainable than the previous one. Also, you can find tests.
 

Yes

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.


nomos-services/blend/src/core/backends/libp2p/swarm.rs
```diff
@@ -317,14 +317,14 @@ impl<SessionStream, Rng> BlendSwarm<SessionStream, Rng> {
     fn forward_validated_swarm_message(
         &mut self,
         msg: &EncapsulatedMessageWithValidatedPublicHeader,
-        except: PeerId,
+        from: (PeerId, ConnectionId),
```
ntn-x2: Should we keep it `except`?

```suggestion
        except: (PeerId, ConnectionId),
```

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -175,9 +183,26 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
             peering_degree: config.peering_degree.clone(),
             connections_waiting_upgrade: HashMap::new(),
             local_peer_id,
+            old_session: OldSession::new(),
         }
     }
 
+    pub fn start_new_session(&mut self, new_membership: Membership<PeerId>) {
```
ntn-x2, line 207: What is the reason to expose this function to the swarm instead of using a stream and perform everything internally? I think this is abstraction leak which could be problematic going forward, don't you think? At most we could have this test-gated so we can manually rotate sessions, although a test implementation of a stream object should be sufficient.

ntn-x2: We already have a TODO in this behaviour struct definition to introduce a session stream. Ideally this would be implemented in the `poll` of the `OldSession` so the implementation would be encapsulated in the old connection, and would also be neat and clean.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -955,6 +1000,10 @@ where
         &mut self,
         cx: &mut Context<'_>,
     ) -> Poll<ToSwarm<Self::ToSwarm, THandlerInEvent<Self>>> {
+        if let Poll::Ready(event) = self.old_session.poll(cx) {
```
ntn-x2: Can we chain all the different conditions instead of mixing returns and not?

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -175,9 +183,26 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
             peering_degree: config.peering_degree.clone(),
             connections_waiting_upgrade: HashMap::new(),
             local_peer_id,
+            old_session: OldSession::new(),
         }
     }
 
+    pub fn start_new_session(&mut self, new_membership: Membership<PeerId>) {
+        self.connections_waiting_upgrade.clear();
```
ntn-x2, line 208: Just a nit, I think we should do with this the same we do for negotiated peers, which is pass it to the old session, who then implements its own logic, i.e., clearing the map.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -215,10 +245,15 @@ impl<ObservationWindowClockProvider> Behaviour<ObservationWindowClockProvider> {
...
         &mut self,
         message: &EncapsulatedMessageWithValidatedPublicHeader,
-        excluded_peer: PeerId,
+        from: (PeerId, ConnectionId),
     ) -> Result<(), Error> {
-        self.forward_validated_message_and_maybe_exclude(message, Some(excluded_peer))?;
-        Ok(())
+        if self.old_session.is_negotiated(&from) {
+            tracing::debug!(target: LOG_TARGET, "Forwarding message to the conns in the old session, except {from:?}.");
+            self.old_session.forward_validated_message(message, from.0)
```
ntn-x2: I guess this would work, but might be an even better idea to have these method be implemented inside the `OldSession` directly, which would return `true` if it was handled there. So the `if self.old_session.is_negotiated` would actually be internal logic of the old session. Similar considerations for other methods.

nomos-blend/network/src/core/with_core/behaviour/old_session.rs
```diff
@@ -0,0 +1,212 @@
...
+            waker: None,
+        }
+    }
+
+    /// Starts an old session with the given states.
+    ///
+    /// If it still has the old session that is not stopped yet, it stops it
+    /// first. This can happen if this method was called before
+    /// [`Self::stop`] is called.
+    pub fn start(
```
ntn-x2: Why do we need to `start` an old connection? Ideally this would happen in the `new` function, right? And old session is, by definition, still ongoing. A `stop` event makes sense tho, and ideally would also return a different type that can only be polled and on which no other methods can be called, so we can enforce all these constraints in the type system directly.

ntn-x2 (commented):
I like the new, much simpler approach, which is also easier to follow. I have few suggestions aimed at relying even more on the type system, and for a better encapsulation of the logic related to the old session.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: Yeah, I thought about it at the beginning, but I came up with this approach since it looks simpler to me. I think it's a common practice for NetworkBehaviour to expose functions to accept inputs. I don't think it is abstraction leak, at least for now.
Of course, we could use streams but I couldn't see clear reasons why it's safer.  

> so the implementation would be encapsulated in the old connection, and would also be neat and clean.

I don't clearly get what this means. We need a new membership for the new session, not the old one. If I'm missing anything from your perspective, could you please help me understand it with a simple code snippet?

p.s.
I updated the TODO comments: cbb27cf8

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: I'm not sure if I understand your comment. What does it mean to chain all the different condition?

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: The map is cleared by https://github.com/logos-co/nomos/blob/d511ded66ba59514f95b65c539e8d83352b63135/nomos-blend/network/src/core/with_core/behaviour/mod.rs#L194

nomos-services/blend/src/core/backends/libp2p/swarm.rs
youngjoon-lee: IMO, `from` is better. I think `old_session.is_negotiated(&from)` is more natural than `old_session.is_negotiated(&except)`. 
https://github.com/logos-co/nomos/blob/d511ded66ba59514f95b65c539e8d83352b63135/nomos-blend/network/src/core/with_core/behaviour/mod.rs#L250
But, I'm okay with `except` as well. Reverted 7982d968.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: Fixed e4e4711c

nomos-blend/network/src/core/with_core/behaviour/old_session.rs
youngjoon-lee: I agree. f1853b5f I took an approach simpler than the type system you suggested. 
```diff
     /// States for processing messages from the old session
     /// before the transition period has passed.
-    old_session: OldSession,
+    old_session: Option<OldSession>,
```

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: I guess you're saying that we could merge two `if` clauses. Yes, we could, but I pushed a commit f1853b5f, which makes this function little bit longer. So, I think it's cleaner to have two separate `if` clauses.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
ntn-x2: Ok, I don't think it's so much a matter of safety as is a matter of logic encapsulation. Now we have one more dependency between the swarm and the behaviour which I think is not necessary, as is, for instance, dialing a new peer when one disconnects since the swarm is the only one to know what peers are currently blacklisted. Anyway, we can go with this approach.

> I don't clearly get what this means. We need a new membership for the new session, not the old one. If I'm missing anything from your perspective, could you please help me understand it with a simple code snippet?

What I mean here is that the `OldSession` type could expose its own `poll` method that the behaviour queries. If this returns `Pending`, the behaviour polls the rest of the components to return whatever result. This would then mean that the `OldSession` would implement all the polling logic internally instead of the behaviour having to deal with values the old session returns (e.g, `is_negotiated`), and act accordingly.

But I said, let's go with this approach. We have tests to catch any inconsistencies, so those will probably be enough.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
ntn-x2: Yeah what I mean is that we have an early `return` in the first condition, and then an `if-else` later on, so we would either always early `return` or chain everything into an `if-elseif-else` chain. Sorry if I was not clear enough on this.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
ntn-x2: Yes exactly. What I was suggesting is that we pass `self.connections_waiting_upgrade` to `OldSession`, which would then call `.clear()` on it, so that we have all "clean-up" logic in `OldSession`, instead of doing some of it in the behaviour and some of it in the `OldSession`. Reason is, what if, for any reason, we might want to keep some connections waiting upgrade upon session rotations? This would then imply a change in `OldSession`'s `start()` method instead of the network behaviour. Again, this is not so much a safety concern as a clear distinction of responsibilities between the behaviour and the `OldSession`, but if you think this way it's better, let's go with this.

youngjoon-lee: The `Behaviour` manages/clears data structures necessary for the current session. Because `self.connections_waiting_upgrade` is only for the current session, it is cleared by the `Behaviour`, not by the `OldSession`. 

In my previous PR, we'd discussed to not accept connection upgraded from the old session. So, in this PR, I clear `self.connections_waiting_upgrade` without passing it the `OldSession`.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: 1df839a5 maybe, is this what you suggested?

ntn-x2: Yes, it's more consistent now, thanks.

ntn-x2 (approved):
There's nothing left unaddressed. I like the new simplified structure!

PR #1579, opened 2025-08-22
feat(ledger): Mantle SDP ledger definition

bacv:

This PR removes service based SDP ledger and implements a new one in `nomos-ledger` crate.
Most important changes are in:
 - New SDP ledger in [ledger/nomos-ledger/src/mantle/sdp.rs](https://github.com/logos-co/nomos/compare/mantle-sdp-ledger?expand=1#diff-b8739c207808f04332d06811f8abc5268da5d4cdb46bb1cd5f2b064f5d0e4445)
- TransientDeclarationState mutates DeclarationState in place [nomos-core/chain-defs/src/sdp/state.rs](https://github.com/logos-co/nomos/compare/mantle-sdp-ledger?expand=1#diff-39a53f9b4a40f89a003b1b95aa945e986c7030b7cbc817796b7508967119a02d)

SDP service was kept around as placeholder, as we'll use it in membership service, and also use it for DeclarationInfo storage.

New SdpLedger will be integrated into Nomos Ledger in future PRs, this contains only the SdpLedger definition.

[SDP spec](https://www.notion.so/nomos-tech/Service-Declaration-Protocol-Specification-1fd261aa09df819ca9f8eb2bdfd4ec1d)
[SDP mantle op spec](https://www.notion.so/nomos-tech/Mantle-Specification-21c261aa09df810c8820fab1d78b53d9?source=copy_link#21c261aa09df81518ae0e607d2b0f468)

@zeegomo @bacv 

Yes

No

## Checklist

* [x] 2. Description added.


danielSanchezQ (approved):
Missing actual integration (?). This is clean anyway.

ledger/nomos-ledger/src/mantle/mod.rs
```diff
@@ -25,6 +30,20 @@ pub enum Error {
     UnsupportedOp,
     #[error("Invalid signature")]
     InvalidSignature,
+    #[error("Sdp ledger error: {0:?}")]
+    Sdp(SdpLedgerError),
+}
+
+impl From<SdpLedgerError> for Error {
```
zeegomo: nit: why not `#[from]`?

ledger/nomos-ledger/src/mantle/sdp.rs
```diff
@@ -0,0 +1,112 @@
...
+    SdpStateError(DeclarationStateError),
+    #[error("Sdp declaration id not found: {0:?}")]
+    SdpDeclarationNotFound(DeclarationId),
+    #[error("Invalid sdp message nonce: {0:?}")]
+    SdpInvalidNonce(Nonce),
+    #[error("Duplicate sdp declaration id: {0:?}")]
+    SdpDuplicateDeclaration(DeclarationId),
+}
+
+impl From<DeclarationStateError> for SdpLedgerError {
```
zeegomo: same

nomos-core/chain-defs/src/sdp/mod.rs
```diff
@@ -109,32 +109,39 @@ pub struct DeclarationInfo {
...
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
-pub enum DeclarationState {
-    Active,
-    Inactive,
-    Withdrawn,
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct DeclarationState {
+    pub created: BlockNumber,
+    pub active: Option<BlockNumber>,
```
zeegomo: the spec says for `DeclarationInfo`:

> active refers to the latest block number for which the active message was sent (it is set to created by default);

It seems this would serve a similar purpose, so why an optional for `active`?

ledger/nomos-ledger/src/mantle/sdp.rs
```diff
@@ -0,0 +1,112 @@
...
+        let Some(current_state) = self.declarations.get_mut(id) else {
+            return Err(SdpLedgerError::SdpDeclarationNotFound(*id).into());
+        };
+
+        if nonce != current_state.nonce + 1 {
+            return Err(SdpLedgerError::SdpInvalidNonce(nonce).into());
+        }
+
+        TransientDeclarationState::try_from_state(block_number, current_state, service_params)?
+            .try_into_withdrawn(block_number, EventType::Withdrawal, service_params)?;
```
zeegomo: what's up with this api? why is the even type passed twice? (once as a parameter and once in the function signature)

nomos-core/chain-defs/src/sdp/state.rs
```diff
@@ -127,101 +127,82 @@ pub enum ProviderStateError {
...
-            TransientDeclarationState::Withdrawn(withdrawn_state) => withdrawn_state.0,
-        }
-    }
-}
-
-impl TransientDeclarationState {
-    pub fn try_from_info(
+impl<'a> TransientDeclarationState<'a> {
+    pub fn try_from_state(
         block_number: BlockNumber,
```
zeegomo: what's this block_number supposed to represent?

nomos-core/chain-defs/src/sdp/state.rs
```diff
@@ -127,101 +127,82 @@ pub enum ProviderStateError {
...
             > block_number
         {
-            return Ok(ActiveState(declaration_info).into());
+            return Ok(ActiveState(declaration_state).into());
         }
 
         // Check if provider has ever got the activity recorded first and then see if
         // the activity record was recent.
-        if let Some(activity) = declaration_info.active {
+        if let Some(activity) = declaration_state.active {
```
zeegomo: you can probably merge this with the check above, either by setting activity at creation as in the spec or by doing the check on `max(activity, created)`

nomos-core/chain-defs/src/sdp/mod.rs
```diff
@@ -213,6 +220,13 @@ impl<Metadata> SdpMessage<Metadata> {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum FinalizedDeclarationState {
```
zeegomo, line 224: could you elaborate on what this is for?

ledger/nomos-ledger/src/mantle/sdp.rs
bacv: The statement can be expressed like this:
```
let current_state = TransientDeclarationState::try_from_state(block_number, current_state, service_params)?;
let new_state = current_state.try_into_withdrawn(block_number, EventType::Withdrawal, service_params)?;
```
It's passed because state recieved from TransientDeclarationState doesn't hold service parameters. I guess it could hold them, if you think such api would be clearer.

nomos-core/chain-defs/src/sdp/mod.rs
bacv: Good catch, will remove optional and set it on creation.

ledger/nomos-ledger/src/mantle/sdp.rs
zeegomo: I mean in :
```
current_state.try_into_withdrawn(block_number, EventType::Withdrawal, service_params)?;
                         ^                                    ^ 
```
                              
the fact this is a withdrawal is passed twice. Why do we need the `EventType` arg for? don't we already know that it's a withdrawal from the function signature?

nomos-core/chain-defs/src/sdp/state.rs
bacv: > what's this block_number supposed to represent?

The current block number, during which the Declaration State is being checked - depending of current Block Number, the state can be `Active` or `Inactive`.

nomos-core/chain-defs/src/sdp/state.rs
bacv: > you can probably merge this with the check above

Will do.

nomos-core/chain-defs/src/sdp/mod.rs
bacv:
```
pub enum FinalizedDeclarationState {
```

> could you elaborate on what this is for?

At the moment this definition is used in Membership service, where we are supposed to receive the state changes of declarations that are not withdrawn or expired yet.

ledger/nomos-ledger/src/mantle/sdp.rs
bacv: > don't we already know that it's a withdrawal from the function signature?

Good point, this was added for previous implementation, will cleanup all unused state transitions and simplify the api.

nomos-core/chain-defs/src/sdp/mod.rs
bacv: Fixed in https://github.com/logos-co/nomos/pull/1579/commits/7d5260ced46365f8e5fed8082f5a8a166b07f520, the same change has `max(active, created)` suggestion added.

ledger/nomos-ledger/src/mantle/sdp.rs
bacv: > don't we already know that it's a withdrawal from the function signature?

Simplified in https://github.com/logos-co/nomos/pull/1579/commits/fd5857b137e72eea08b6be239ed624104147961a

ledger/nomos-ledger/src/mantle/mod.rs
bacv: Fixed in https://github.com/logos-co/nomos/pull/1579/commits/e4ae58ff0f3373f6d29816d524c28f14681bb62c

nomos-core/chain-defs/src/sdp/state.rs
zeegomo: can you rename it to `current_block_number`?

nomos-core/chain-defs/src/sdp/state.rs
```diff
@@ -160,17 +162,16 @@ impl<'a> TransientDeclarationState<'a> {
 
         // Check if provider has ever got the activity recorded first and then see if
         // the activity record was recent.
-        if let Some(activity) = declaration_state.active {
-            if block_number.wrapping_sub(activity) <= service_params.inactivity_period {
-                return Ok(ActiveState(declaration_state).into());
-            }
+        let activity = max(declaration_state.active, declaration_state.created);
```
zeegomo: this is redundant now, as active can't be older than created

nomos-core/chain-defs/src/sdp/state.rs
```diff
@@ -179,10 +180,7 @@ impl<'a> TransientDeclarationState<'a> {
         if let Some(withdrawn_timestamp) = declaration_state.withdrawn {
             return withdrawn_timestamp;
         }
-        if let Some(activity_timestamp) = declaration_state.active {
-            return activity_timestamp;
-        }
-        declaration_state.created
+        max(declaration_state.active, declaration_state.created)
```
zeegomo: same

zeegomo (changes requested):

nomos-core/chain-defs/src/sdp/state.rs
```diff
@@ -102,23 +100,24 @@ impl<'a> TransientDeclarationState<'a> {
...
+            > current_block_number
         {
             return Ok(ActiveState(declaration_state).into());
         }
 
         // Check if provider has ever got the activity recorded first and then see if
         // the activity record was recent.
-        let activity = max(declaration_state.active, declaration_state.created);
-        if block_number.wrapping_sub(activity) <= service_params.inactivity_period {
+        if current_block_number.wrapping_sub(declaration_state.active)
```
zeegomo: why wrapping? it should not happen that `declaration_state > current_block_number `, but if it does I don't think wrapping is the right approach. What do you think? we could use checked + panic in case of a overflow

I also see `wrapping` used in other places, maybe I'm missing something?

ledger/nomos-ledger/src/mantle/sdp.rs
```diff
@@ -0,0 +1,106 @@
...
+        let Some(current_state) = self.declarations.get_mut(id) else {
+            return Err(SdpLedgerError::SdpDeclarationNotFound(*id).into());
+        };
+
+        if nonce != current_state.nonce + 1 {
+            return Err(SdpLedgerError::SdpInvalidNonce(nonce).into());
+        }
+
+        TransientDeclarationState::try_from_state(block_number, current_state, service_params)?
+            .try_into_withdrawn(block_number, service_params)?;
```
zeegomo, line 100: There's nothing obviously wrong and I see what you're trying to do, but the api seems a bit weird to me, as you create an ephemeral type `TransientDeclarationState` and pass twice the same arguments (`block_number`, `service_params`) in short succession.
I'd just ask if this is debt from the past implementation or intended. If the latter I'm fine with it, I trust your judgment here

nomos-core/chain-defs/src/sdp/state.rs
```diff
@@ -127,149 +74,122 @@ pub enum ProviderStateError {
...
                 .map(Into::into)
-                .map_err(ProviderStateError::from),
+                .map_err(DeclarationStateError::from),
             Self::Inactive(inactive_state) => inactive_state
-                .try_into_withdrawn(block_number, event_type, service_params)
+                .try_into_withdrawn(current_block_number, service_params)
                 .map(Into::into)
-                .map_err(ProviderStateError::from),
-            Self::Withdrawn(_) => Err(ProviderStateError::WithdrawnToOtherState),
+                .map_err(DeclarationStateError::from),
```
zeegomo: could use `?`?

zeegomo (commented):
Last round of comments, good to go after these are resolved

ledger/nomos-ledger/src/mantle/sdp.rs
bacv: > I'd just ask if this is debt from the past implementation or intended

I suppose it's both. Previous implementation operated on DeclarationInfo and created a new copies of it after modification. But the api where we pass `block_number` and `service_params` is intended, you can see it's usage in the state related unit tests.

This api is probably weird as it's more suited for tests than actual usage, but in my opinion since the usage is so little but sdp has alot of tests cases , and we don't sacrifice any performance by passing those params twice, it's acceptable compromise.

nomos-core/chain-defs/src/sdp/state.rs
bacv: Updated all places here https://github.com/logos-co/nomos/pull/1579/commits/431ea73d66f8759002ff150da2affce37569224c

nomos-core/chain-defs/src/sdp/state.rs
bacv: Changed to this https://github.com/logos-co/nomos/pull/1579/commits/473d075f3c2f116be33244f86c4286da95dfdf2f

PR #1582, opened 2025-08-25
fix(blend): post-rebase issues

ntn-x2:
`master` CI is broken since I merged two PRs before first rebasing the second one. I learnt my lesson 😊

### DIFF
--- a/nomos-services/blend/src/edge/backends/libp2p/mod.rs
+++ b/nomos-services/blend/src/edge/backends/libp2p/mod.rs
@@ -50,7 +50,7 @@ impl<RuntimeServiceId> BlendBackend<PeerId, RuntimeServiceId> for Libp2pBlendBac
             current_membership,
             rng,
             swarm_command_receiver,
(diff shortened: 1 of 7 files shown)


nomos-services/blend/src/edge/backends/libp2p/mod.rs
```diff
@@ -50,6 +50,7 @@ impl<RuntimeServiceId> BlendBackend<PeerId, RuntimeServiceId> for Libp2pBlendBac
             current_membership,
             rng,
             swarm_command_receiver,
+            settings.protocol_name.clone().into_inner() ,
```
youngjoon-lee:
```suggestion
            settings.protocol_name.clone().into_inner(),
```

ntn-x2: Yes I already have a commit locally to address the fmt issue. Will push it once everything else in the CI turns green. Thanks!

PR #1585, opened 2025-08-25
feat(ledger): Add support for leader rewards in ledger

zeegomo:

Adds support for:
* tracking amount to distribute in ledger
* redeeming a share through a leaderclaim op

Does not deal with:
* cryptography, in particular no verification of leader claim proof
* calculating how much to reward for each epoch
* make use of this functionality in the node

Yes

https://www.notion.so/nomos-tech/Mantle-Specification-21c261aa09df810c8820fab1d78b53d9, https://www.notion.so/nomos-tech/Anonymous-Leaders-Reward-Protocol-206261aa09df8120a49ffa49c71ba70d

@thomaslavaur @davidrusu authors, @zeegomo implementing

Yes

No
## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/ledger/nomos-ledger/src/mantle/leader.rs
+++ b/ledger/nomos-ledger/src/mantle/leader.rs
@@ -1,0 +1,122 @@
+use crate::Balance;
+use cryptarchia_engine::Epoch;
+use mmr::MerkleMountainRange;
+use nomos_core::mantle::{
+    ops::leader_claim::{LeaderClaimOp, RewardsRoot, VoucherNullifier, VoucherCm},
+    Value,
+};
+use std::cmp::Ordering;
+
+
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct LeaderState {
+    // current epoch
+    epoch: Epoch,
+    // vouchers that can be claimed in this epoch
+    // this is updated once at the start of each epoch from the root of
+    // the vouchers merkle tree
+    claimable_vouchers_root: RewardsRoot,
+    n_claimable_vouchers: u64,
+    // nullifiers of vouchers that have been claimed since genesis
+    nfs: rpds::HashTrieSetSync<VoucherNullifier>,
+    // rewards to be distributed
+    // at the start of each epoch this is increased by the amount of rewards
+    // that have been collected in the previous epoch.
+    // unclaimed rewards are carried over to the next epoch.
+    rewards: Value,
+    // Merkle tree of vouchers, vouchers can only be claimed with a delay
+    // of one epoch.
+    vouchers: MerkleMountainRange<VoucherCm, nomos_core::crypto::Hasher>,
+}
+
+#[derive(Debug, thiserror::Error, Clone, PartialEq, Eq)]
+pub enum Error {
+    #[error("voucher nullifier already used")]
+    DuplicatedVoucherNullifier,
+    #[error("voucher not found")]
+    VoucherNotFound,
+    #[error("Cannot time travel to the past")]
+    InvalidEpoch { current: Epoch, incoming: Epoch },
+}
+
+impl LeaderState {
+    #[must_use]
+    pub fn new() -> Self {
+        Self {
+            epoch: 0.into(),
+            claimable_vouchers_root: [0; 32],
+            n_claimable_vouchers: 0,
+            nfs: rpds::HashTrieSetSync::new_sync(),
+            rewards: 0,
+            vouchers: MerkleMountainRange::new(),
+        }
+    }
+
+    #[must_use]
+    pub fn try_apply_header(self, epoch: Epoch, voucher_cm: VoucherCm) -> Result<Self, Error> {
+        Ok(self.update_epoch_state(epoch)?
+            .add_voucher(voucher_cm))
+    }
+
+    fn update_epoch_state(mut self, epoch: Epoch) -> Result<Self, Error> {
+        match epoch.cmp(&self.epoch) {
+            Ordering::Equal => Ok(self),
+            Ordering::Less => Err(Error::InvalidEpoch {
+                current: self.epoch,
+                incoming: epoch,
+            }),
+            Ordering::Greater => {
+                self.epoch = epoch;
+                self.claimable_vouchers_root = self.vouchers.frontier_root();
+                self.n_claimable_vouchers = self.vouchers.len() as u64;
+                // TODO: increase rewards, what about epoch jumps?
+                Ok(self)
+            }
+        }
+    }
+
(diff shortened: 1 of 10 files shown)


ledger/nomos-ledger/src/mantle/leader.rs
```diff
@@ -0,0 +1,122 @@
...
+    // the vouchers merkle tree
+    claimable_vouchers_root: RewardsRoot,
+    n_claimable_vouchers: u64,
+    // nullifiers of vouchers that have been claimed since genesis
+    nfs: rpds::HashTrieSetSync<VoucherNullifier>,
+    // rewards to be distributed
+    // at the start of each epoch this is increased by the amount of rewards
+    // that have been collected in the previous epoch.
+    // unclaimed rewards are carried over to the next epoch.
+    rewards: Value,
```
youngjoon-lee:
```suggestion
    // rewards to be distributed
    // during the current epoch this is increased by the amount of rewards
    // that have been collected in the previous epoch.
    // unclaimed rewards are carried over to the next epoch.
    rewards: Value,
```
What do you think about this change? I thought so, because rewards are redeemed only when leaders claim them, if I understand it correctly.

mmr/src/lib.rs
```diff
@@ -21,6 +21,7 @@ const EMPTY_VALUE: [u8; 32] = [0; 32];
 #[derive(Debug, Clone)]
 pub struct MerkleMountainRange<T, Hash, const MAX_HEIGHT: u8 = 32> {
     roots: StackSync<Root>,
+    len: usize,
```
davidrusu: Can you not derive len from roots?
e.g.

    roots: [Root{height:3}, Root{height:1}]

implies `2^(3-1) + 2^(1-1) = 4 + 1 = 5` elements in the MMR

mmr/src/lib.rs
```diff
@@ -223,6 +230,7 @@ mod test {
         assert_eq!(mmr.roots.peek().unwrap().root, leaf(b"hello"));
 
         mmr = mmr.push(b"world".as_ref());
+        assert_eq!(mmr.roots.size(), 1);
```
davidrusu: do you mind to add some asserts for `mmr.len()` as well?

ledger/nomos-ledger/src/mantle/leader.rs
```diff
@@ -0,0 +1,122 @@
...
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct LeaderState {
+    // current epoch
+    epoch: Epoch,
+    // vouchers that can be claimed in this epoch
+    // this is updated once at the start of each epoch from the root of
+    // the vouchers merkle tree
+    claimable_vouchers_root: RewardsRoot,
+    n_claimable_vouchers: u64,
```
davidrusu, line 21: What do you think of storing the snapshot of the MMR here rather than these derived fields? The MMR is pretty lightweight, and then you can compute these values when you need them rather than storing the derived field.

```suggestion
    claimable_vouchers: MerkleMountainRange<VoucherCm, nomos_core::crypto::Hasher>,
```

ledger/nomos-ledger/src/mantle/leader.rs
davidrusu: Do we need the same, double-buffered structure for rewards as well?
```
claimable_rewards: Value,
rewards: Value
```

ledger/nomos-ledger/src/mantle/leader.rs
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+    }
+
+    fn add_voucher(self, voucher_cm: VoucherCm) -> Self {
+        Self {
+            vouchers: self.vouchers.push(voucher_cm),
+            ..self
+        }
+    }
```
davidrusu, line 89: For testing purposes, can we increase `reward` by 1 for each voucher added? That way we can test the claim flow

ledger/nomos-ledger/src/mantle/leader.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    /// Claim the reward associated with a voucher.
+    /// Any cryptographic proof of correct derivation of the voucher nullifier and
+    /// membership proof in the merkle tree is expected to happen outside of this function.
```
davidrusu:
```suggestion
    /// membership proof in the merkle tree is expected to be verified outside of this function.
```

ledger/nomos-ledger/src/mantle/leader.rs
```diff
@@ -0,0 +1,122 @@
...
+                claimable_vouchers_root: self.claimable_vouchers_root,
+                n_claimable_vouchers: self.n_claimable_vouchers,
+                nfs,
+                rewards: new_rewards,
+                vouchers: self.vouchers.clone(),
+            },
+            Balance::from(reward_amount),
+        ))
+    }
+}
```
davidrusu, line 128: This looks right, but some tests would be nice to make sure that the epoch transitions and claim amounts are correct.

mmr/src/lib.rs
zeegomo: true

ledger/nomos-ledger/src/mantle/leader.rs
zeegomo: Yeah that was my initial idea, not sure what happened

ledger/nomos-ledger/src/mantle/leader.rs
zeegomo: aren't all rewards claimable? or are you referring to live computing the new rewards to add for next epoch? In that case I haven't had a look at that yet (is it finished?) so it's missing because of that

ledger/nomos-ledger/src/mantle/leader.rs
davidrusu: This way is fine as well

ledger/nomos-ledger/src/mantle/leader.rs
davidrusu: Yeah, because there is that period where new vouchers are not getting added to the pool but we are still accumulating rewards.

So if all rewards are claimable, then the reward share will gradually increase over the course of the epoch until the next batch of vouchers are added.

ledger/nomos-ledger/src/mantle/leader.rs
zeegomo: Just leaving the rationale here for the future:

you would need to compute the frontier root each block instead of maintaining one more field. I'd say both are fine although I have a slight preference for the current method since it's still quite easy to understand.

davidrusu (approved):

PR #1586, opened 2025-08-25
feat(zk): Proof of Leadership lib

danielSanchezQ:

Proof of Leadership as per the specification. Uses rapidsnarkjs as backend for prover and witness generator and custom groth16 for verification.

This includes a small restructure of circuits crates @strinnityk 

It has a few massaging types but overall API is quite simple.

@danielSanchezQ @thomaslavaur 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.


danielSanchezQ:
@strinnityk This includes some refactor, mostly removing the traits we intended for the rapidsnarkjs stuff. Finally they are of no use.

zk/proofs/pol/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+///
+/// # Errors
+/// - Returns a `ProveError::Io` if an I/O error occurs while generating the
+///   witness or proving from contents.
+/// - Returns a `ProveError::Json` if there is an error during JSON
+///   serialization or deserialization.
+pub fn prove(inputs: &PolWitnessInputs) -> Result<(PoLProof, PolVerifierInput), ProveError> {
+    let witness = witness::generate_witness(inputs).map_err(ProveError::Io)?;
+    let (proof, verifier_inputs) =
+        circuits_prover::prover_from_contents(POL_PROVING_KEY_PATH.as_path(), witness.as_ref())
```
thomaslavaur, line 86: So each time we want to prove something we will need to charge the key for a file?

thomaslavaur: It's ok if it's only for proving and that it's not often but for blend it could be a problem if core nodes need to generate keys often (in PoQ)

danielSanchezQ: There is the server mode. But I would like to avoid it as it involves extra complexity that we do not want at the moment. This should be fast enough for now. Imo the best approach is to transform the rapidsnark into a library instead of an external binary, which would solve all this troubles. But that is not a fight I'm willing to take right now.

zk/proofs/pol/src/chain_inputs.rs
```diff
@@ -0,0 +1,233 @@
...
+}
+
+/// Public inputs of the POL cirmcom circuit to be provided by the chain.
+pub struct PolChainInputsData {
+    pub slot_number: u64,
+    pub epoch_nonce: u64,
+    pub total_stake: u64,
+    pub aged_root: Fr,
+    pub latest_root: Fr,
+    pub leader_pk: (Fr, Fr),
```
zeegomo, line 33: I guess this is the ed25519 key, maybe we can do the conversion inside this lib?

zk/proofs/pol/src/chain_inputs.rs
```diff
@@ -0,0 +1,233 @@
...
+    BigUint::from_str("28794005923809446652337194229268641024881242442862297438215833784455126237")
+        .expect("Constant should parse")
+});
+
+#[derive(Debug, Error)]
+pub enum PolInputsFromDataError {
+    #[error("Slot number is greater than P")]
+    SlotGreaterThanP,
+    #[error("Epoch nonce is greater than P")]
+    EpochGreaterThanP,
```
zeegomo, line 161: both of this are impossible in practice since we use 64bit numbers for those but let's leave this here for now

zk/proofs/pol/src/chain_inputs.rs
```diff
@@ -0,0 +1,233 @@
...
+        let epoch_nonce = BigUint::from(epoch_nonce);
+        if epoch_nonce > *P {
+            return Err(PolInputsFromDataError::EpochGreaterThanP);
+        }
+        let total_stake = BigUint::from(total_stake);
+
+        let lottery_0 = T0_CONSTANT.deref().div(total_stake.clone());
+        let lottery_1 = P
+            .checked_sub(&T1_CONSTANT.deref().div(total_stake.pow(2)))
+            .expect("(T1 / (S^2)) must be less than P");
```
zeegomo, line 190: the spec says these are 256bit numbers, but AFAIK `BigUInt` variable length. Not that it matters in precision as those are integers, but this could behave differently for edge cases which are not even possible with current sizes of total stake. Anyway, could be something to note

zk/proofs/pol/src/wallet_inputs.rs
```diff
@@ -0,0 +1,118 @@
...
+    starting_slot: Groth16Input,
+}
+
+/// Private inputs of the POL cirmcom circuit to be provided by the wallet.
+pub struct PolWalletInputsData {
+    pub note_value: u64,
+    pub transaction_hash: Fr,
+    pub output_number: u64,
+    pub aged_path: Vec<Fr>,
+    pub aged_selector: Vec<bool>,
```
zeegomo, line 26: IIRC we had a merkle path type to fuse these two together, something like 
```
enum MerkleNode {
  Left(value),
  RIght(value)
}
```

zeegomo (approved):
Looks good, left a few comments for ergonomics

zk/proofs/pol/src/chain_inputs.rs
danielSanchezQ: cc @thomaslavaur , is it? From my convos with Thomas the idea is to have everything as Fr's already.

zk/proofs/pol/src/chain_inputs.rs
danielSanchezQ: True, I can remove it. I added this as it was in the poc scripts for the t0 and t1 computations.

zk/proofs/pol/src/chain_inputs.rs
danielSanchezQ: Yes. Initially I used `U256`, but I had to pass through BigUint (U256 -> BigUint -> Fr) so I changed everything to BigUints. I think it should be fine as you said. But if it feels safer I can bring it back.

zk/proofs/pol/src/wallet_inputs.rs
danielSanchezQ: Ah, nice. I didn't realize. I'll change this type during the integration phase for convenience then.

Issue #1500, opened 2025-07-28
Use broadcasting if network is too small for edge nodes


ntn-x2:
As per spec: https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#232261aa09df8036847ccaf549c38ea7.

Issue #1483, opened 2025-07-22
Use broadcasting if network is too small for core nodes


ntn-x2:
From the spec: https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#232261aa09df80a2a6e2fcd731bf0eac.

PR #1587, opened 2025-08-26
feat(blend): fallback to gossipsub if Blend network is too small

ntn-x2:

This PR fixes https://github.com/logos-co/nomos/issues/1500 and fixes https://github.com/logos-co/nomos/issues/1483 by introducing logic mainly in the proxy service to NOT forward a message to the core service (and once SDP is integrated also the edge service) if the current session membership is not large enough, but instead to broadcast it via gossipsub directly.

To increase reliability and reduce assumptions about inter-service interactions, also core and edge services check the network size, and refuse to accept incoming messages to blend or to start a libp2p swarm if the network is too small. The only missing piece is to change the state of each service based on new membership, which is currently being tackled by @youngjoon-lee.

Spec paragraph describing this mechanism: https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#232261aa09df80b9ba20ec70636e0db6.

Yes.

Yes.

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/core/backends/libp2p/tests/utils.rs
+++ b/nomos-services/blend/src/core/backends/libp2p/tests/utils.rs
@@ -58,7 +58,14 @@ impl SwarmBuilder {
 
     pub fn with_empty_membership(mut self) -> Self {
         assert!(self.membership.is_none());
-        self.membership = Some(Membership::new(&[], None));
+        self.membership = Some(Membership::new(
+            &[Node {
+                address: Multiaddr::empty(),
+                id: PeerId::random(),
+                public_key: Ed25519PrivateKey::generate().public_key(),
+            }],
+            None,
+        ));
         self
     }
 
@@ -115,7 +122,14 @@ impl BlendBehaviourBuilder {
 
     pub fn with_empty_membership(mut self) -> Self {
         assert!(self.membership.is_none());
-        self.membership = Some(Membership::new(&[], None));
+        self.membership = Some(Membership::new(
+            &[Node {
+                address: Multiaddr::empty(),
+                id: PeerId::random(),
+                public_key: Ed25519PrivateKey::generate().public_key(),
+            }],
+            None,
+        ));
         self
     }
 
--- a/nomos-services/blend/src/edge/backends/libp2p/tests/redials.rs
+++ b/nomos-services/blend/src/edge/backends/libp2p/tests/redials.rs
@@ -117,7 +117,6 @@ async fn edge_redial_same_peer() {
 #[test(tokio::test)]
 async fn edge_redial_different_peer_after_redial_limit() {
     let random_peer_id = PeerId::random();
-    println!("Random peer ID: {random_peer_id:?}");
     let empty_multiaddr: Multiaddr = Protocol::Memory(0).into();
 
     let CoreTestSwarm {
@@ -133,7 +132,6 @@ async fn edge_redial_different_peer_after_redial_limit() {
         });
     let (core_swarm_membership_entry, _) =
         core_swarm.listen_and_return_membership_entry(None).await;
-    println!("Core swarm peer ID: {:?}", core_swarm_membership_entry.id);
 
     // We include both the core and the unreachable swarm in the membership.
     let edge_membership = Membership::new(


nodes/nomos-executor/config.yaml
```diff
@@ -48,6 +48,7 @@ blend:
     - id: 12D3KooWA5LkQcuefjUBTktxCXUQ4rx4PcqRY1tgbSpV9eZDHWir
       address: /ip4/127.0.0.1/udp/8721/quic-v1
       public_key: 92f041e26306548444400e88eefa622dc125a45f3e3119a04ba1ffe4946dd4e6
+    minimum_network_size: 1
```
danielSanchezQ: Should this be a config? 🤔 Or is it part of genesis config?

nomos-services/blend/src/core/mod.rs
```diff
@@ -92,92 +91,120 @@ where
...
+            // interact with this service by sending messages to it, and it indeed should
+            // not happen, as all interactions should happen via the proxy service.
+            service_resources_handle.status_updater.notify_ready();
+            tracing::info!(
+                target: LOG_TARGET,
+                "Service '{}' is ready.",
+                <RuntimeServiceId as AsServiceId<Self>>::SERVICE_ID
+            );
+        } else {
+            let mut rng = ChaCha12Rng::from_entropy();
```
danielSanchezQ: Chacha12 should not be used anymore. Any rng should use the blake based one as per spec iirc

nomos-services/blend/src/edge/mod.rs
```diff
@@ -59,64 +57,83 @@ where
...
+            .get_updated_settings();
+        let membership = settings.membership();
+        let current_membership = Some(membership.clone());
+        let minimum_network_size = settings.minimum_network_size;
 
-        let mut messages_to_blend = inbound_relay.map(|ServiceMessage::Blend(message)| {
-            wire::serialize(&message)
-                .expect("Message from internal services should not fail to serialize")
-        });
+        // TODO: Add logic to try process new sessions. I.e:
```
danielSanchezQ, line 78: We should have an issue with this. Although I know you will tackle it soon.

danielSanchezQ (commented):
Just did a quick overview. I'm concern only about the chacha usage. Why are we still using chacha in blend?

nodes/nomos-executor/config.yaml
ntn-x2: What do you mean exactly? If we don't make it a config, at least for the time being, all our integration tests would break since we mostly use networks of 2 nodes. Or rather, they would not break, but they would only use gossipsub, so we would not test Blend logic at all.

nomos-services/blend/src/core/mod.rs
ntn-x2: Yes good point. I will change that.

nomos-services/blend/src/edge/mod.rs
ntn-x2: Yes, this is what @youngjoon-lee is working on, and I hope he can review this PR before doing major changes in the service logic, so that we work in sync on this.

ntn-x2: Also, ideally the edge service would not be started at all if the network size is too small. Here I wanted to be extra cautious. But I don't think it's too complicated to add logic that re-evaluate the conditions every time a new session starts. But at the moment neither the core nor the edge service has a concept of session transition, so I did not want to introduce it in this PR.

nomos-services/blend/src/core/mod.rs
ntn-x2: https://github.com/logos-co/nomos/pull/1587/commits/5f30324ad11702b4545091ab75e0f9b172d1a77c

nomos-services/blend/src/edge/mod.rs
youngjoon-lee: Although we'd discussed in [Discord](https://discord.com/channels/1111286067413405788/1399328124268974173), while reading this PR, I came up with a question: **What about just creating all the backends and let them work regardless of the network size?**
That way, we only rely on the proxy service. And, that might cause the inter-service assumptions that you mentioned in the PR description. But, now I feel (not strongly) that we might not need to make our architecture more complicated because of this fallback mechanism. 

From the perspective of core nodes, it doesn't matter even if they exchange cover messages regardless of the network size, in my opinion.
Even if someone wants to blend his message regardless of the network size, I don't see any clear reason to block him doing so.
He just has to understand that the small blend network might be less secure unless core nodes increase their cover message rate (they won't).

I'm aware that the spec says that the network bootstrapping should be **stopped** if the network is small. And, I think that's the reason why you came up with this approach in this PR. 
It's true that we don't need to bootstrap a network that we may not use. But I feel our code architecture is getting bigger because of a feature that I think isn't really that significant.

Maybe I'm wrong. I'm probably feeling that the architecture is complicated since I've given this less thought than you have. So, I'd like to hear your opinion. If you also think it may be worth trying to simplify the architecture by relaxing the spec, we could open (or continue) a discussion in Discord. Or, if you think it's better to keep this approach as is, I will approve this PR and continue implementing the session transition according to this approach.

youngjoon-lee: Or, what do you think about just building an empty membership (or `None`) if the network is small? :) It just suddenly came to my mind, and it might be a strange idea, but it could simplify the architecture.

ntn-x2: I realized that entirely stopping the core swarm might not be the best idea when a new session comes in where the node is not a core node anymore, but we will want to keep connections with the core peers of the previous session. So I was already thinking to start the core swarm regardless, and make the swarm itself aware that if the network is too small, no bootstrapping will be performed and no message belonging to the new session will be sent/received (this is just a two-line checks, although we would lose some compile-time guarantees in case we inadvertently remove those checks in the future - which could be solved with a different swarm type, but I understand code complexity is becoming a topic).

The way I see it, since our implementation is some sort of reference implementation, we should **not** allow messages to be sent via Blend if the network is too small. People might fork our client or come up with an alternative client that bypasses this.

Also, the reason I put the logic in the services is that I wanted to avoid adding it to the swarms, but as I said above, while we can afford to not make the edge swarm aware of the minimum network size (we simply drop the swarm if the new network is too small - also because the edge swarm does not try to bootstrap a network upon starting up, so starting up is very cheap), we need to keep the core swarm around for the session transition.

As for the approach, as you might have seen by now, I am a big fan of relying on the compiler as much as possible, which is why I did not want to instantiate a swarm regardless.

Anyway, given my points above, I would suggest to keep the edge swarm logic as it is already defined in this PR, since as of now it does not run any additional logic on session rotation, but I would update the core swarm logic to 1. instantiate the swarm regardless of the network size, and 2. make the core swarm aware of the minimum network size, so that it does not perform any beginning-of-session bootstrapping (nor it will accept any incoming connection requests) when requested to. This should simplify session transition code to simply callling `rotate_session` on the swarm from the service.

TL;DR We should enforce sensible defaults where possible. Since this client is a reference implementation of the spec, I am arguing for not allowing any messages to go through Blend at all, nor to establish a network of core nodes at all to save resources where possible. I am of course open to counter-arguments. I am ccing @danielSanchezQ as well for another input.

youngjoon-lee: In my opinion, the spec is defining things in too much detail. I'm not fully convinced why we need the minimal network size. That was why I didn't want to have extra complexity for this feature. But anyway, that's what the spec currently defines. So, I won't go into that any further here.

I didn't mean to discuss about the design pattern. I also like your approach and agree it's required by the current spec.

ntn-x2: @youngjoon-lee @danielSanchezQ I pushed some changes that work as a proposal for how I envision things. I think it's a waste of resources if we won't blend any messages to maintain connections. So I am proposing that, as Youngjoon suggested, we start the core swarm regardless, because of session transition logic, but we tell the behaviour how to "behave" depending on the membership size. Meaning: we won't accept connections nor rely any messages. This behaviour changes automatically if a new session with enough peers starts. This means that the core service logic is reverted to what it was before, while I would maintain the new one for the edge swarm, since for edge nodes we don't need to run a swarm at all if we're not gonna use Blend.

nomos-services/blend/src/edge/mod.rs
```diff
@@ -59,64 +57,83 @@ where
...
+        // * If the old session membership was too small and the new one is large
+        //   enough, create a new backend and start receiving messages to blend on the
+        //   inbound channel.
+        // * If the old and new session membership are both too small, do nothing.
+        // * If the old session membership was large and the new one too small, perform
+        //   the session rotation on the backend but stop listening on the inbound
+        //   channel, since we don't want to blend new messages.
```
youngjoon-lee: Just to understand your intention clearly, do you think it's better to **deconstruct** the backend if the new session is too small? Or, do you think it's okay to let it be?

Anyway, I think we need to update the comment as below because the edge swarm doesn't accept inbound connections.
```suggestion
        // * If the old session membership was large and the new one too small, perform
        //   the session rotation on the backend.
```

ntn-x2: The backend trait has a `shutdown` method, and since creating an edge swarm does not involve anything besides creating an object, to get help from the compiler I was thinking dropping it completely would be a good idea. Actually, we can change the trait definition to make sure we can't use the swarm anymore after it is dropped. Let me try that.

ntn-x2: Something like this? https://github.com/logos-co/nomos/pull/1587/commits/064ccd9e327c523ddddb2023ac4d6868c16c640c

Then we avoid interacting with the swarm when we are not meant to, and we are forced to instantiate a new one.

ntn-x2:
Moving back to draft until we settle on which way to go.

Issue #1572, opened 2025-08-22
Merge header with public header


youngjoon-lee:
In the [spec](https://www.notion.so/nomos-tech/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#23b261aa09df8034a122d6c71eff87aa), we removed `Header` and move its contents to the `PublicHeader`.

PR #1588, opened 2025-08-26
chore(blend): merge header into public header

ntn-x2:

Fixes https://github.com/logos-co/nomos/issues/1572.

Yes.

Yes. I prompted @madxor to also update the pseudo-code to include the header in the encapsulation and decapsulation logic.

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/message/src/message.rs
+++ b/nomos-blend/message/src/message.rs
@@ -10,44 +10,50 @@ use crate::{
     error::Error,
 };
 
-pub const VERSION: u8 = 1;
-
-// A message header that is revealed to all nodes.
-#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
-pub struct Header {
-    version: u8,
-}
-
-impl Header {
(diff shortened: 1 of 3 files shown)


nomos-blend/message/src/message.rs
```diff
@@ -10,45 +10,51 @@ use crate::{
...
-        Self { version: VERSION }
-    }
-}
-
-impl Default for Header {
-    fn default() -> Self {
-        Self::new()
-    }
-}
+const LATEST_BLEND_MESSAGE_VERSION: u8 = 1;
```
youngjoon-lee, line 13: I think we don't need the `LATEST` prefix because we're always going to define only the latest version in this code, probably. But anyway, both are fine by me.

ntn-x2: I think for the sake of clarity I would keep the `LATEST` prefix. In the future we might be dealing with old clients that speak an older version, and here we want to clarify that we are always using the latest version when producing new messages.

PR #1589, opened 2025-08-26
feat(membership): introduce session forming in mock backend membershi…

pradovic:
…p implementation

This PR implements session forming in membership backend mock implementation. Mock implementation is not storing the current state so it can resume in the case of recovery, but should cover the whole business logic.

In order for downwards code to compile (da network assignations) that depends on block number, the session id is passed downstream as blocknumber, and session size is 1 block in tests for them to pass. This enables us to open several PR for this without conflicts and test the whole chain in the end.

So the following PRs should promote session id downstream (might be trivial just block number -> session ID) and add tests for session size > 1 block.

Yes.

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/nomos-services/membership/src/backends/mod.rs
+++ b/nomos-services/membership/src/backends/mod.rs
@@ -1,26 +1,13 @@
 use std::collections::HashMap;
 
 use async_trait::async_trait;
-use nomos_core::{
-    block::BlockNumber,
-    sdp::{FinalizedBlockEvent, ServiceType},
-};
+use nomos_core::sdp::{FinalizedBlockEvent, ServiceType};
 use overwatch::DynError;
-use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
 use crate::MembershipProviders;
 
 pub mod mock;
-
-pub struct MembershipBackendSettings {
-    pub settings_per_service: HashMap<ServiceType, MembershipBackendServiceSettings>,
-}
-
-#[derive(Clone, Debug, Copy, Serialize, Deserialize)]
-pub struct MembershipBackendServiceSettings {
-    pub historical_block_delta: u64,
-}
 
 #[derive(Debug, Error)]
 pub enum MembershipBackendError {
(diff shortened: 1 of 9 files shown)


nomos-services/membership/src/backends/mod.rs
```diff
@@ -31,16 +18,13 @@ pub enum MembershipBackendError {
...
 #[async_trait]
 pub trait MembershipBackend {
     type Settings: Send + Sync;
 
     fn init(settings: Self::Settings) -> Self;
-    async fn get_providers_at(
-        &self,
-        service_type: ServiceType,
-        block_number: BlockNumber,
-    ) -> Result<MembershipProviders, MembershipBackendError>;
```
danielSanchezQ, line 43: Why are we removing this? The only thing to change here would be to replace Blocknumber by session. (?)

nomos-services/membership/src/lib.rs
```diff
@@ -172,19 +167,6 @@ where
...
-                service_type,
-            } => {
-                let result = self
-                    .backend
-                    .get_providers_at(service_type, block_number)
-                    .await;
-                if let Err(e) = reply_channel.send(result) {
-                    tracing::error!("Failed to send response: {:?}", e);
-                }
-            }
```
danielSanchezQ, line 187: How are we gonna get snapshots for different sessions?

nomos-services/membership/src/backends/mod.rs
pradovic: It had in memory snapshot per block. I think now that all we need is a current session, in memory, that is forming. No needed to support get snapshot at, as services only use streaming.

nomos-services/membership/src/lib.rs
pradovic: Do we need this all we only need streaming? We have assignation snapshots in da

pradovic: SDP (block) -> Membership Service (builds sessions per service type) ->  DA.

Da needs snapshots for historic query and will have them saved. Membership service might need snapshot only for recover  i think, and we can maybe avoid it completely. It needs to save current forming session in memory in order to continue on restart. This message was not used anywhere, only subscribe is used.

Does this makes sense?

danielSanchezQ: I recall that for historical sampling we have to send the the membership with the query? Where do we get that?

pradovic: This is fetched from da network, as we need snapshot of assignations for session.

danielSanchezQ (approved):
Ok, lets go with this. We will just need to change where to get config if we plan to use genesis at some point. Thanks!

bacv (approved):

PR #1590, opened 2025-08-26
feat(zk): Use poseidon2 for hashes

danielSanchezQ:

This PR adds a base feature which is to use Poseidon2 as the hash system for zk related stuff. In order to use it a few elements changed mostly from `[u8; 32]` (bytes) to `Fr` (base elements poseidon2 works with). This includes public and private keys. As well as other ids such of notes, txs, etc.

As it touches many places I'm adding a guide on changes:
* ZK
  * Groth16: Exposed a few types for convenience. Added a serde utility module to serialize `Fr` elements.
  * Poseidon2: Added a custom Digest trait that has replace the common digest one in some crates.
* Utxotree: Now it works with poseidon underneath. Its abstracted using the Digest trait, but every `[u8; 32]` is now a `Fr`.
* Nomos-core
  * chain-defs: Replace hashes to use Poseidon and changed a the types accordingly to hold `Fr`'s. A few convenience methods were added too.
  * proof-statements: This types are to be replaced later with the proper proofs types from the libs (pol wip at #1586). But it is gonna make the integration much smoother after this.

For the other crates is just adapting the types mostly.

Yes, it should be straight forwards seeing the changes. Tests were adapted but not changed so it assures everything fits in place.

@danielSanchezQ @zeegomo @thomaslavaur 

It was a bit confusing on when to use one hash (blake) or the other (poseidon). But it was fixed.

No, just clarifications.

## Checklist

> [!WARNING]  

* [x] 2. Description added.


Cargo.toml
```diff
@@ -61,13 +61,14 @@ resolver = "2"
 bundler                  = { path = "./nomos-bundler" }
 cfgsync                  = { path = "./testnet/cfgsync" }
 chain-service            = { path = "./nomos-services/chain-service" }
-circuits-prover          = { path = "zk/circuits/prover" }
-circuits-utils           = { path = "zk/circuits/utils" }
-circuits-verifier        = { path = "zk/circuits/verifier" }
+circuits-prover          = { path = "./zk/circuits/prover" }
```
zeegomo, line 63: what's up with this?

nomos-core/chain-defs/src/mantle/ledger.rs
```diff
@@ -67,52 +72,84 @@ pub struct Utxo {
...
 
+static NOMOS_NOTE_ID_V1: LazyLock<Fr> = LazyLock::new(|| {
+    let mut hasher = blake2::Blake2bVar::new(31).expect("blake2 initialization error");
+    hasher.update(b"NOMOS_NOTE_ID_V1");
+    let mut buff = [0u8; 31];
+    hasher
+        .finalize_variable(&mut buff)
+        .expect("blake2 variable hashing error");
+    BigUint::from_bytes_le(buff.as_slice()).into()
+});
```
zeegomo: this is weird, if we have to convert bytes at the end into a Fr anyway why not convert the domain separator directly? What's stopping us from doing `NOMOS_NOTE_ID_V1: Fr = BigUint::from_bytes_le(b"NOMOS_NOTE_ID_V1").into()`

nomos-core/chain-defs/src/mantle/ledger.rs
```diff
@@ -67,52 +72,84 @@ pub struct Utxo {
...
-        const NOMOS_NOTE_ID_V1: &[u8] = b"NOMOS_NOTE_ID_V1";
-
-        let mut hasher = crate::crypto::Hasher::new();
 
-        hasher.update(NOMOS_NOTE_ID_V1);
-        hasher.update(self.tx_hash.0);
-        hasher.update(self.output_index.to_le_bytes());
-        hasher.update(self.note.value.to_le_bytes());
-        hasher.update(self.note.pk.as_bytes());
+        let mut hasher = Poseidon2Bn254Hasher::default();
```
zeegomo: maybe we can still keep the crate level hasher and add zkhasher

nomos-core/chain-defs/src/mantle/ledger.rs
```diff
@@ -67,52 +72,84 @@ pub struct Utxo {
...
+
 impl Tx {
     #[must_use]
     pub const fn new(inputs: Vec<NoteId>, outputs: Vec<Note>) -> Self {
         Self { inputs, outputs }
     }
 
     #[must_use]
-    pub fn as_sign_bytes(&self) -> Bytes {
+    pub fn as_signing_fr(&self) -> Fr {
```
zeegomo: the idea of this method was that hashing does not happen here and we only build the input data (so returning a `Vec<Fr>`), any reason to change that?

nomos-core/chain-defs/src/mantle/mod.rs
```diff
@@ -33,7 +33,7 @@ pub trait Transaction {
     /// The resulting bytes are then used by the `HASHER`
     /// to produce the transaction's unique hash, which is what is typically
     /// signed by the transaction originator.
-    fn as_sign_bytes(&self) -> Bytes;
+    fn as_signing_fr(&self) -> Fr;
```
zeegomo: see other comment, the doc would also need updating

nomos-core/chain-defs/src/mantle/ops/channel/blob.rs
```diff
@@ -22,18 +23,25 @@ impl BlobOp {
...
+        buff.extend_from_slice(self.blob.as_ref());
+        buff.extend_from_slice(self.blob_size.to_le_bytes().as_ref());
+        buff.extend_from_slice(self.da_storage_gas_price.to_le_bytes().as_ref());
+        buff.extend_from_slice(self.parent.as_ref());
+        buff.extend_from_slice(self.signer.as_ref());
+        buff.freeze()
+    }
+
+    #[must_use]
+    pub fn as_sign_bytes(&self) -> Bytes {
```
zeegomo: I guess this should return `payload_bytes` as well? or maybe it's a leftover

nomos-core/chain-defs/src/mantle/ledger.rs
```diff
@@ -67,52 +72,84 @@ pub struct Utxo {
     pub note: Note,
 }
 
+static NOMOS_NOTE_ID_V1: LazyLock<Fr> = LazyLock::new(|| {
+    let mut hasher = blake2::Blake2bVar::new(31).expect("blake2 initialization error");
+    hasher.update(b"NOMOS_NOTE_ID_V1");
+    let mut buff = [0u8; 31];
```
zeegomo: why 31?

zk/groth16/src/serde.rs
```diff
@@ -0,0 +1,23 @@
...
+    use ark_bn254::Fr;
+    use ark_ff::biginteger::BigInteger as _;
+    use num_bigint::BigUint;
+    use serde::{Deserialize as _, Deserializer, Serializer};
+
+    pub fn serialize<S>(item: &Fr, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let hex = hex::encode(item.0.to_bytes_le()); // Convert `Fr` to hex representation
```
zeegomo: we should use `into_bigInt` as in the docs

zeegomo (changes requested):

Cargo.toml
danielSanchezQ: I think in previous prs were added without the relative path. I changed so it fits with the common ones.

nomos-core/chain-defs/src/mantle/ledger.rs
danielSanchezQ: Yep, I'll change. Also its much simpler.

nomos-core/chain-defs/src/mantle/ledger.rs
danielSanchezQ: Indeed. I added the crate type later during the refactor. I'll this on other places as well. Good catch, thanks!

nomos-core/chain-defs/src/mantle/ledger.rs
danielSanchezQ: Uhm, I see. I saw sign and returning a single thing to be signed. So I thought we should sign a single Fr as well. But you are right. Better to extract it.

nomos-core/chain-defs/src/mantle/ops/channel/blob.rs
danielSanchezQ: I dont recall. But I think we were using this before when it was only implemented for the blob. So maybe now its just a leftover.

nomos-core/chain-defs/src/mantle/ledger.rs
danielSanchezQ: Not all 32bytes vales fits in a bn254 Fr.

zk/groth16/src/serde.rs
danielSanchezQ: Not sure its available 🤔 , will double check. Thanks.

nomos-core/chain-defs/src/mantle/ledger.rs
danielSanchezQ: After coming to it. I was completely code drunk before. It should be fine now https://github.com/logos-co/nomos/pull/1590/commits/4753a08d2ce99403f208278f5ff8024ca4cebe2a

nomos-core/chain-defs/src/mantle/ops/channel/blob.rs
bacv: This was used [here (link to a code change in this PR)](https://github.com/logos-co/nomos/pull/1590/files/b8cf13c92a63c9ebf758792c8ca198bf80ee03cd#diff-62d1efd2a33e5dec2d05514334ea45fcf6fd62cc45349b858737693974dbdd4cL116) and renamed to `payload_bytes`.

`as_sign_bytes` can be removed from blob op now.

zk/groth16/src/serde.rs
danielSanchezQ: Where is that in the docs? You mean it should be a `str` with the bignum representation?

nomos-core/chain-defs/src/mantle/ledger.rs
zeegomo: yeah but a not all `Fr` fit into 31 bytes as well. Tbf the `Fr::from(BigInt)` does modular reduction so it's always possible, but we should keep an eye on that as that skews the probability so not to be used for cryptographic stuff cc @megonen @thomaslavaur

zk/groth16/src/serde.rs
zeegomo: where do you see the internal BigInt in the public docs? If I look at the source code it's `#[doc(hidden)]` with a comment to use the `into_bigint` https://docs.rs/ark-ff/0.5.0/src/ark_ff/fields/models/fp/mod.rs.html#106-112

nomos-core/chain-defs/src/mantle/ledger.rs
danielSanchezQ: But that would be Fr to Bytes not bytes to Fr. I get what you mean. I think we are not going Fr -> 31bytes anywhere. But lets keep the eyes open.

zk/groth16/src/serde.rs
danielSanchezQ: Ah, ok, i can try that. Thanks!

nomos-core/chain-defs/src/mantle/ledger.rs
zeegomo: My point is that if you want to sample a `Fr` at random, it's not correct to sample 31 bytes and then convert those to `Fr`, as you're basically ignoring ~6 bits (254 - 248). It's fine now since we removed it from here but just to be sure it's not used somewhere else

danielSanchezQ: Uhm, Are you sure of that? I'm missing the point. Anything 31bytes is fit into the Fr as if the remaining bytes were padded to 0 iirc. The problem would be to *dump* the Fr into 31bytes were we would have data loss.

bacv (approved):
Reviewed DA related parts, looks good.

nomos-core/chain-defs/src/mantle/ledger.rs
megonen: In the case of DST usage, the goal is simply to ensure uniqueness, and in this context using 31 bytes to fit within the prime field is not problematic. However, in situations where randomness is required, restricting to 31 bytes would compromise uniformity when sampling over prime modulus. Since DSTs are not about randomness, I don’t think that issue applies here

nodes/nomos-node/node/src/config/mod.rs
```diff
@@ -403,14 +404,16 @@ pub fn update_cryptarchia_consensus(
         return Ok(());
     };
 
-    let sk = nomos_core::mantle::keys::SecretKey::from(<[u8; 16]>::from_hex(secret_key)?);
+    let sk = nomos_core::mantle::keys::SecretKey::from(BigUint::from_bytes_le(
+        &<[u8; 16]>::from_hex(secret_key)?,
+    ));
```
danielSanchezQ, line 409: @thomaslavaur This is probably not good. We would have to actually change it when we have the wallet tho.

nomos-core/chain-defs/src/mantle/ops/mod.rs
```diff
@@ -110,14 +113,28 @@ impl Op {
...
+    #[must_use]
+    pub fn as_signing_fr(&self) -> Vec<Fr> {
+        let mut buff = Vec::new();
+        buff.push(BigUint::from(self.opcode()).into());
+
+        for chunk in self.payload_bytes().chunks(31) {
+            buff.push(BigUint::from_bytes_le(chunk).into());
         }
-        buff.freeze()
+        buff
```
danielSanchezQ, line 137: @thomaslavaur , this is what we talked for the payload. But maybe we should specify how to get those Frs from the payload in the specification.

nomos-core/chain-defs/src/mantle/ledger.rs
```diff
@@ -1,35 +1,43 @@
...
     #[must_use]
-    pub const fn as_bytes(&self) -> &[u8; 32] {
+    pub const fn as_fr(&self) -> &Fr {
         &self.0
     }
+
+    #[must_use]
+    pub fn as_bytes(&self) -> Bytes {
+        self.0 .0 .0.iter().flat_map(|b| b.to_le_bytes()).collect()
+    }
```
danielSanchezQ, line 29: This is the concatenation of the le bytes of the u64 that represents the inner BigInt @thomaslavaur

nodes/nomos-node/node/src/config/mod.rs
thomaslavaur: Indeed it's not good. The specification to derive the secret keys can be found:
- here https://www.notion.so/nomos-tech/Wallet-Technical-Standard-215261aa09df80e9884ad7cf039e2c57?source=copy_link#253261aa09df804d884beec790e96826
and here for the one participating in PoS: https://www.notion.so/nomos-tech/Proof-of-Leadership-Specification-21c261aa09df819ba5b6d95d0fe3066d?source=copy_link#21c261aa09df8162a035efaf3581702a

nomos-core/chain-defs/src/mantle/ops/mod.rs
thomaslavaur: I added in the Mantle specification that the payload is interpreted as a slice of 31-byte elements. If we add elsewhere how to convert from bytes to Fr I guess it's enough?

nomos-core/chain-defs/src/mantle/ledger.rs
thomaslavaur: so from Fr elements to bytes?

nodes/nomos-node/node/src/config/mod.rs
danielSanchezQ: I'll add an issue for this. Need to be fixed after we have the wallet.

nomos-core/chain-defs/src/mantle/ops/mod.rs
danielSanchezQ: > I added in the Mantle specification that the payload is interpreted as a slice of 31-byte elements. If we add elsewhere how to convert from bytes to Fr I guess it's enough?

Yeah, is enough if we added already. Thanks.

nomos-core/chain-defs/src/mantle/ledger.rs
danielSanchezQ: > so from Fr elements to bytes?

Yes. We should probably have something in the specs for this as well. What is the bytes representation of the `Fr`?

ledger/nomos-ledger/src/cryptarchia/mod.rs
```diff
@@ -337,12 +340,11 @@ pub mod tests {
...
     #[must_use]
     pub fn utxo() -> Utxo {
-        let mut tx_hash = [0; 32];
-        thread_rng().fill_bytes(&mut tx_hash);
+        let tx_hash: Fr = BigUint::from(thread_rng().next_u64()).into();
         Utxo {
             tx_hash: tx_hash.into(),
             output_index: 0,
-            note: Note::new(10000, [0; 32].into()),
+            note: Note::new(10000, Fr::from(BigUint::from(0u8)).into()),
```
thomaslavaur, line 347: we should use Fr::ZERO I think here

ledger/nomos-ledger/src/cryptarchia/mod.rs
```diff
@@ -697,15 +699,15 @@ pub mod tests {
 
     #[test]
     fn test_tx_processing_valid_transaction() {
-        let input_note = Note::new(10000, [1; 32].into());
+        let input_note = Note::new(11000, Fr::from(BigUint::from(1u8)).into());
```
thomaslavaur, line 702: same here Fr::ONE

ledger/nomos-ledger/src/cryptarchia/mod.rs
```diff
@@ -697,15 +699,15 @@ pub mod tests {
...
         let input_utxo = Utxo {
-            tx_hash: [1; 32].into(),
+            tx_hash: Fr::from(BigUint::from(1u8)).into(),
```
thomaslavaur, line 704: ditto and same for Fr::TWO later

nodes/nomos-node/node/src/config/mod.rs
danielSanchezQ: https://github.com/logos-co/nomos/issues/1605

ledger/nomos-ledger/src/lib.rs
```diff
@@ -261,8 +263,8 @@ mod tests {
     fn test_ledger_try_update_with_transaction() {
         let (ledger, genesis_id, utxo) = create_test_ledger();
 
-        let output_note = Note::new(1, [1; 32].into());
-        let pk = [0; 32];
+        let output_note = Note::new(1, PublicKey::new(BigUint::from(1u8).into()));
+        let pk = BigUint::from(0u8).into();
```
thomaslavaur, line 267: shouldn't the public key be a Fr element?

nodes/nomos-node/node/src/config/mod.rs
```diff
@@ -403,14 +404,16 @@ pub fn update_cryptarchia_consensus(
...
     cryptarchia.leader_config.sk = sk;
 
     let pk = sk.to_public_key();
 
-    let tx_hash = <[u8; 32]>::from_hex(tx_hash)?;
+    let tx_hash: TxHash = BigUint::from_bytes_le(&<[u8; 32]>::from_hex(tx_hash)?).into();
```
thomaslavaur, line 414: I fill like the stored transaction hash should be a Fr and the conversion done only for signature verification

nomos-core/chain-defs/src/mantle/ledger.rs
```diff
@@ -144,51 +150,52 @@ impl Tx {
...
 
 #[cfg(test)]
 mod test {
-
     use super::*;
-
     #[test]
     fn test_utxo_by_index() {
+        let pk0 = PublicKey::from(Fr::from(BigUint::from(0u8)));
+        let pk1 = PublicKey::from(Fr::from(BigUint::from(0u8)));
```
thomaslavaur: shouldn't pk0 = 0, pk1 = 1 ect? because here they are all equal to 0

nomos-core/chain-defs/src/mantle/merkle.rs
```diff
@@ -1,38 +1,49 @@
...
 use serde::{Deserialize, Serialize};
 
-use crate::crypto::{Digest as _, Hasher};
+use crate::mantle::NoteId;
 
 #[must_use]
-pub fn padded_leaves<const N: usize>(elements: &[Vec<u8>]) -> [[u8; 32]; N] {
-    let mut leaves = [[0u8; 32]; N];
+pub fn padded_leaves<const N: usize>(elements: &[NoteId]) -> [Fr; N] {
+    let mut leaves = vec![BigUint::from(0u8).into(); N];
```
thomaslavaur: for performance we should use Fr::ZERO everywhere possible

nomos-core/chain-defs/src/mantle/merkle.rs
```diff
@@ -1,38 +1,49 @@
...
-        leaves[i] = leaf(element);
+        leaves[i] = leaf(element.as_fr());
     }
 
-    leaves
+    leaves.try_into().expect("Size is asserted per loop")
 }
 
+static NOMOS_MERKLE_LEAF: LazyLock<Fr> =
+    LazyLock::new(|| BigUint::from_bytes_be(b"NOMOS_MERKLE_LEAF").into());
```
thomaslavaur, line 23: Is this DST in the specs? If not and necessary, the issue should be rased so we add it

thomaslavaur: But we shouldn't add a DST here IMO as it would drastically increase the complexity of the zk circuit

nomos-core/chain-defs/src/mantle/merkle.rs
```diff
@@ -1,38 +1,49 @@
...
-    hasher.finalize().into()
+pub fn leaf(data: &Fr) -> Fr {
+    let mut hasher = Poseidon2Bn254Hasher::default();
+    <Poseidon2Bn254Hasher as Digest>::update(&mut hasher, &NOMOS_MERKLE_LEAF);
+    <Poseidon2Bn254Hasher as Digest>::update(&mut hasher, data);
+    hasher.finalize()
 }
 
+static NOMOS_MERKLE_NODE: LazyLock<Fr> =
+    LazyLock::new(|| BigUint::from_bytes_be(b"NOMOS_MERKLE_NODE").into());
```
thomaslavaur, line 34: ditto

nomos-core/chain-defs/src/sdp/mod.rs
```diff
@@ -51,6 +52,16 @@ pub enum ServiceType {
     ExecutorNetwork,
 }
 
+impl AsRef<str> for ServiceType {
+    fn as_ref(&self) -> &str {
+        match self {
+            Self::BlendNetwork => "BN",
+            Self::DataAvailability => "DA",
+            Self::ExecutorNetwork => "EX",
```
thomaslavaur, line 60: We don't have Executor network anymore in the specs so we can remove it

nomos-core/proof_statements/src/leadership.rs
```diff
@@ -1,23 +1,27 @@
...
+    pub latest_root: Fr,
     // TODO: missing rewards
 }
 
 impl LeaderPublic {
     #[must_use]
     pub fn new(
-        aged_root: [u8; 32],
-        latest_root: [u8; 32],
+        aged_root: Fr,
```
thomaslavaur, line 23: I don't know here but at the very end when giving the input to the circuit, the name must match the one of the circuit that are the same as in the specs of PoL (ledger_aged and ledger_latest)

nomos-core/proof_statements/src/leadership.rs
```diff
@@ -58,10 +62,16 @@ impl LeaderPublic {
     }
 
     #[must_use]
-    pub fn check_winning(&self, value: u64, note_id: [u8; 32], sk: [u8; 16]) -> bool {
+    pub fn check_winning(&self, value: u64, note_id: Fr, sk: Fr) -> bool {
         let threshold = phi_approx(U256::from_u64(value), self.scaled_phi_approx);
```
thomaslavaur, line 66: The threshold is also a Fr element computed from the value as a Fr element and phi isn't the same as in the original paper (we use the second degree of the Taylor extension)

nomos-core/proof_statements/src/leadership.rs
```diff
@@ -58,10 +62,16 @@ impl LeaderPublic {
...
-        let ticket = ticket(note_id, sk, self.epoch_nonce, self.slot);
-        ticket < threshold
+        let threshold = BigUint::from_bytes_le(&threshold.to_le_bytes());
+        let ticket = ticket(
+            note_id,
+            sk,
+            BigUint::from_bytes_le(&self.epoch_nonce).into(),
+            BigUint::from(self.slot).into(),
+        );
+        ticket < threshold.into()
```
thomaslavaur, line 74: and so the threshold and the ticket are two Fr element that can be compared

nomos-core/proof_statements/src/leadership.rs
```diff
@@ -77,23 +87,16 @@ fn phi_approx(stake: U256, approx: (U256, U256)) -> U256 {
...
-    hasher.update(epoch_nonce);
-    hasher.update(slot.to_be_bytes());
-    hasher.update(note_id);
-    hasher.update(sk);
-
-    let ticket_bytes: [u8; 32] = hasher.finalize().into();
-
-    U256::from_be_bytes(ticket_bytes)
+fn ticket(note_id: Fr, sk: Fr, epoch_nonce: Fr, slot: Fr) -> Fr {
+    Poseidon2Bn254Hasher::digest(&[note_id, sk, epoch_nonce, slot])
```
thomaslavaur, line 91: the ticket should use a DST in the hash

nomos-services/chain-service/src/states.rs
```diff
@@ -191,7 +193,7 @@ mod tests {
         );
 
         // Empty leader utxos.
-        let leader = Leader::new(vec![], [0; 16].into(), ledger_config);
+        let leader = Leader::new(vec![], Fr::from(BigUint::from(1u8)).into(), ledger_config);
```
thomaslavaur, line 196: Fr::ONE

ledger/nomos-ledger/src/cryptarchia/mod.rs
danielSanchezQ: This is just for testing. IT doesnt reallly matter. Fr::ZERO or ONE is not used because it needs an extra trait which would be used just for this probably.

ledger/nomos-ledger/src/cryptarchia/mod.rs
danielSanchezQ: Same on this ones

ledger/nomos-ledger/src/lib.rs
danielSanchezQ: It is, but we have some wrapper types in some stages.

nodes/nomos-node/node/src/config/mod.rs
danielSanchezQ: Yeah, ill open an issue for this. This is node config.

danielSanchezQ: https://github.com/logos-co/nomos/issues/1606

nomos-core/proof_statements/src/leadership.rs
danielSanchezQ: Yes, there is an upcoming PR that makes sure of it. It also will be replacing this type entirely. #1586

nomos-core/proof_statements/src/leadership.rs
danielSanchezQ: This will be updated in the #1586 PR after merging this one as base. But thanks, I wasn't aware of the changes.

nomos-core/proof_statements/src/leadership.rs
danielSanchezQ: I transformed the threshold to a Fr before yes.

nomos-core/proof_statements/src/leadership.rs
danielSanchezQ: Could you share a link to the documentation? Anyway, this will be adapted with the changes on the Leadership proof (#1586) or in a parallel PR.

nomos-services/chain-service/src/states.rs
danielSanchezQ: Just for testing. Same issue as before. But thanks.

nomos-core/chain-defs/src/sdp/mod.rs
danielSanchezQ: Uhm, that should be done in a different PR. Ill open an issue too.

danielSanchezQ: https://github.com/logos-co/nomos/issues/1607

nomos-core/chain-defs/src/mantle/merkle.rs
danielSanchezQ: @zeegomo This was here before this PR. Should we add to spec or remove it?

nomos-core/chain-defs/src/mantle/merkle.rs
danielSanchezQ: Here yes!

nomos-core/chain-defs/src/mantle/merkle.rs
thomaslavaur: We should definitively removed. It would affect way too much the performance of PoL otherwise

nomos-core/proof_statements/src/leadership.rs
thomaslavaur: https://www.notion.so/nomos-tech/Proof-of-Leadership-Specification-21c261aa09df819ba5b6d95d0fe3066d?source=copy_link#21c261aa09df818aa5b4dd04c5a3aec1

nomos-core/chain-defs/src/mantle/ledger.rs
danielSanchezQ: Let me double check this. It may be wrong.

nomos-core/chain-defs/src/mantle/merkle.rs
danielSanchezQ: https://github.com/logos-co/nomos/issues/1609

danielSanchezQ:
Merging this on spite the flaky test in macos

PR #1591, opened 2025-08-26
feat(da): DA flow with indexer

bacv:

DA Indexer was removed and integration tests updated to follow a flow where shares are requested by the blob_id instead of the deprecated app_id and range.

[Spec](https://www.notion.so/nomos-tech/NomosDA-Reconstruction-1fd261aa09df815ea043d8c48e4917f1?source=copy_link#1fd261aa09df81a6801dc38d01e044b0)

@danielSanchezQ @bacv 

Yes

No

## Checklist

* [x] 2. Description added.


tests/src/tests/cryptarchia/bootstrap.rs
```diff
@@ -45,10 +45,10 @@ async fn test_ibd_behind_nodes() {
 
     let failing_behind_node = Validator::spawn(config).await;
 
-    tokio::time::sleep(Duration::from_secs(2)).await;
-
     // IBD failed and node stopped
     assert!(failing_behind_node.is_err());
+    // Kill the node process.
+    drop(failing_behind_node);
```
bacv, line 51: @andrussal I've noticed an occurrence where `failing_behind_node` correctly panics in `chain_service`, but the whole node doesn't shut down, only the thread of chian service is killed. In that case `Validator::spawn` doesn't return error, and `assert!(failing_behind_node.is_err());` fails.

While we don't have handling of such state in the node, I've updated the `wait_online` in our tests to poll `consensus_info` instead of mempool stats, which then fails if chain service has paniced.

I've also removed the timout after spawning the validator node here as we're `awaiting` for it to start or fail already, maybe I've missed the point of that timeout.

andrussal: > @andrussal I've noticed an occurrence where `failing_behind_node` correctly panics in `chain_service`, but the whole node doesn't shut down, only the thread of chian service is killed. In that case `Validator::spawn` doesn't return error, and `assert!(failing_behind_node.is_err());` fails.
> 
> While we don't have handling of such state in the node, I've updated the `wait_online` in our tests to poll `consensus_info` instead of mempool stats, which then fails if chain service has paniced.
> 
> I've also removed the timout after spawning the validator node here as we're `awaiting` for it to start or fail already, maybe I've missed the point of that timeout.

Thanks for letting me know! I will update the test and remove this altogether I think because this check is not really at the integration test level importance.

Issue #1502, opened 2025-07-29
Message redundancy


youngjoon-lee:
## Core nodes

The [spec](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df8119be73fabe77a870e4) defines:
- $$R_C$$: A redundancy for cover messages
- $$R_D$$: A redundancy for data messages

We've already added this parameter to the config in our code base, but we're not using it yet when sending messages to the Blend network.

However, this is not a high priority. It's okay to implement this after all other features.

## Edge nodes

Similarly, the [spec](https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81feb71ecb944ffd82f5) defines $$\Phi_{EC}^{Min}$$, which is the minimum number of connections. But, from the [discussion](https://www.notion.so/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?d=23b261aa09df805d9628001c77c3941f&source=copy_link#215261aa09df81feb71ecb944ffd82f5), it seems that the parameter means a redundancy, rather than the number of connections that edge nodes should maintain (because we defined that edge nodes establish connections on demand). I proposed to change the symbol to $$R_E$$.
Once the spec becomes clear, we can schedule implementing this, but not a high priority.

PR #1593, opened 2025-08-27
feat(blend): introduce redundancy factor for edge nodes

ntn-x2:

Fixes https://github.com/logos-co/nomos/issues/1502 (see issue comments for why this PR alone is sufficient to fix the issue). For more context about the meaning of the redundancy parameter, refer to the [Discord thread](https://discord.com/channels/1111286067413405788/1410163003999125524).

The only thing I did not implement in here, because I do not think the gains justify the additional overhead, is to keep track of what peers are being sent each message, so that when a dialing with a peer for a given message fails, we avoid picking up a peer we have already sent the same message to. Nevertheless, since edge nodes are not banned and core nodes simply discard duplicated messages, I do not think it's a big issue. Also, since Blend is not meant to run for sets smaller than 32 elements, the chances of this happening for a message are small. For all the copies of the same message much smaller.
Tackling this would involve having one more storage map mapping from a message to a set of peers currently receiving it, that we can clear once all peers have successfully received a message. We could argue that is worth implementing, but otherwise I would only do it if we see that this becomes a problem in the future, but I don't think that will be the case.

Yes.

Yes.

It introduced minor changes to improve clarity around what an edge node is supposed to do. Refer to the Discord thread linked above for more context.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/edge/backends/libp2p/tests/message_handling.rs
+++ b/nomos-services/blend/src/edge/backends/libp2p/tests/message_handling.rs
@@ -2,7 +2,7 @@ use core::{slice::from_ref, time::Duration};
 
 use nomos_blend_scheduling::membership::Membership;
 use test_log::test;
-use tokio::{spawn, time::sleep};
+use tokio::{select, spawn, time::sleep};
 
 use crate::{
     core::backends::libp2p::core_swarm_test_utils::{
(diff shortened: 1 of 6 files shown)


nomos-services/blend/src/edge/backends/libp2p/tests/message_handling.rs
```diff
@@ -58,6 +58,9 @@ async fn edge_message_propagation() {
         swarm: edge_swarm,
         command_sender: edge_swarm_command_sender,
     } = EdgeSwarmBuilder::default()
+        // We test that we can pick the `min` between the propagation factor and the available
+        // peers.
+        .with_propagation_factor(usize::MAX)
```
youngjoon-lee: What do you think about renaming this to `with_replication_factor`? I guess this is a leftover? If you think the `replication_factor` is better, other parts in the test need to be renamed as well.

ntn-x2: Good point! I changed that here, thanks! https://github.com/logos-co/nomos/pull/1593/commits/443c8fa8296f459e1b9c251c7ebe2f7d40ffda7e

PR #1595, opened 2025-08-27
feat(blend): include local node in membership

youngjoon-lee:

Previously, the `Membership` contained only `remote_nodes`. But now we need to check whether the local node is a core node for implementing the proxy service.

This PR updates the `Membership` to contain all the core nodes, including the local node (if it's a core).

Yes

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/scheduling/src/membership.rs
+++ b/nomos-blend/scheduling/src/membership.rs
@@ -13,9 +13,14 @@ use serde::{Deserialize, Serialize};
 
 use crate::serde::ed25519_pubkey_hex;
 
+/// A set of core nodes in a session.
 #[derive(Clone, Debug)]
 pub struct Membership<NodeId> {
+    /// All nodes, including local and remote.
     nodes: HashMap<NodeId, Node<NodeId>>,
+    /// IDs of remote nodes only, excluding the local node if present.
+    /// Kept as a separate [`Vec`] to enable efficient random sampling,
+    /// which is faster than sampling from the keys in [`HashMap`].
     remote_nodes: Vec<NodeId>,
 }
 


nomos-blend/scheduling/src/membership.rs
```diff
@@ -30,39 +34,44 @@ pub struct Node<Id> {
...
-    NodeId: Clone,
+    NodeId: Clone + Hash + Eq,
 {
     #[must_use]
     pub fn new(nodes: &[Node<NodeId>], local_public_key: Option<&Ed25519PublicKey>) -> Self {
+        let mut core_nodes = HashMap::with_capacity(nodes.len());
+        let mut remote_core_nodes = Vec::with_capacity(nodes.len());
+        for node in nodes {
+            core_nodes.insert(node.id.clone(), node.clone());
+            if !matches!(local_public_key, Some(key) if node.public_key == *key) {
```
ntn-x2, line 50: Mhh, can't use store the local node in a simple `Node` instance instead? This way we are duplicating stuff, but I don't see any reason to do that?

ntn-x2 (approved):
Looks good, nice catch! I think we could do better in terms of how we store nodes in the memberhsip, but otherwise all good!

nomos-blend/scheduling/src/membership.rs
youngjoon-lee: That was my initial plan :) 
```rust
struct Membership {
  local_node: Node,
  remote_nodes: HashMap<NodeId, Node>,
}
```
but, I realized that sampling random items from HashMap involves iterating keys. It's slower than the random sampling from Vec. 
So, I came up with using both HashMap and Vec, and not having `local_node`.

We may be able to use indexmap instead for faster sampling, but I thought this duplication is acceptable.

ntn-x2: It is probably acceptable, but it was not clear what was the decision behind it. If using an indexmap is not too much work, perhaps we could use that. Otherwise we can keep this, but perhaps explain with a comment why we do what we do :)

youngjoon-lee: I just had a look at indexmap, but the memory size is almost similar as my approach (hashmap + vec). Also, the API of indexmap is not better than this PR. So, I'd like to keep the current approach. I added comments to explain this choice. Thanks. [1ea49cd](https://github.com/logos-co/nomos/pull/1595/commits/1ea49cdaed97e0daa577c14c2efea076ffdce68f)

PR #1597, opened 2025-08-27
feat(nat): test basic full flow

andrussal:

This PR tests 
- NAT happy flow: **Private listen address**->**Autonat test**->**Mapping**->**Autonat test**->**Public**.
- NAT mapping failure: **Private listen address**->**Autonat test**->**Mapping**->**Mapping Failure**->**Private**.

Because we can't easily block listen address in unit test, it introduces simple transport wrapper that blocks dials.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-libp2p/Cargo.toml
+++ b/nomos-libp2p/Cargo.toml
@@ -42,7 +42,6 @@ tracing = { workspace = true }
 zerocopy = { features = ["derive"], version = "0.8.26" }
 
 [dev-dependencies]
-libp2p             = { workspace = true, default-features = false, features = ["plaintext", "tcp", "yamux"] }
-libp2p-swarm-test  = { features = ["tokio"], version = "0.5.0" }
-serde_json         = { default-features = false, version = "1.0.99" }
-tracing-subscriber = { features = ["env-filter"], version = "0.3" }
+libp2p            = { workspace = true, default-features = false, features = ["plaintext", "tcp", "yamux"] }
+libp2p-swarm-test = { features = ["tokio"], version = "0.5.0" }
+serde_json        = { default-features = false, version = "1.0.99" }


nomos-libp2p/Cargo.toml
```diff
@@ -43,6 +43,7 @@ tracing     = { workspace = true }
 zerocopy    = { version = "0.8.26", features = ["derive"] }
 
 [dev-dependencies]
+libp2p             = { workspace = true, features = ["plaintext", "yamux"] }
```
ntn-x2: Since we will merge soon a PR that removes default features, can we import this with `default-features` off and only enable the features we need? Probably turning them off should be enough.

nomos-libp2p/src/behaviour/nat/inner.rs
```diff
@@ -269,3 +269,207 @@ where
...
+    fn create_server_with_blocking_transport() -> (libp2p::Swarm<Server>, Multiaddr) {
+        let mut swarm = create_memory_swarm(true, |keypair| Server {
+            autonat: server::Behaviour::new(OsRng),
+            identify: identify::Behaviour::new(identify::Config::new(
+                "/test/1.0.0".to_owned(),
+                keypair.public(),
+            )),
+        });
+
+        let server_addr: Multiaddr = "/memory/3000".parse().unwrap();
```
ntn-x2: There's a `listen` function on the `libp2p_swarm_test` crate which automatically picks a random port to start listening on, maybe we can do that here? In case we were to use the same port elsewhere.

ntn-x2: For example: https://github.com/logos-co/nomos/blob/580c6de658eb0c9dc5c2660fba72571019d678a1/nomos-services/blend/src/core/backends/libp2p/tests/message_handling.rs#L31

nomos-libp2p/src/behaviour/nat/inner.rs
```diff
@@ -269,3 +269,207 @@ where
...
+    struct TestMapper;
+
+    #[async_trait::async_trait]
+    impl NatMapper for TestMapper {
+        async fn map_address(
+            _address_to_map: &Multiaddr,
+            _settings: NatMappingSettings,
+        ) -> Result<Multiaddr, AddressMapperError> {
+            MAPPER_CALLED.with(|called| *called.borrow_mut() = true);
+            Ok("/memory/2000".parse().unwrap())
```
ntn-x2: Could we choose a random port to avoid concurrency or re-use issues in the future?

nomos-libp2p/src/behaviour/nat/inner.rs
```diff
@@ -269,3 +269,207 @@ where
...
+    use crate::{
+        behaviour::nat::address_mapper::errors::AddressMapperError, config::NatMappingSettings,
+    };
+
+    thread_local! {
+        static MAPPER_CALLED: RefCell<bool> = const { RefCell::new(false) };
+    }
+
+    #[tokio::test]
+    async fn test_nat_mapping_happy_flow() {
```
ntn-x2: No need to prefix test cases with `test`. There's actually a lint for it which we will enable, at some point :)

nomos-libp2p/src/behaviour/nat/inner.rs
```diff
@@ -269,3 +269,207 @@ where
...
+            _address_to_map: &Multiaddr,
+            _settings: NatMappingSettings,
+        ) -> Result<Multiaddr, AddressMapperError> {
+            MAPPER_CALLED.with(|called| *called.borrow_mut() = true);
+            Ok("/memory/2000".parse().unwrap())
+        }
+    }
+
+    #[derive(Default)]
+    struct FirstDialBlockingTransport {
```
ntn-x2, line 380: What is the need for this transport? What does it help us test?

nomos-libp2p/src/behaviour/nat/inner.rs
```diff
@@ -269,3 +269,207 @@ where
...
+
+        let mut swarm = create_memory_swarm(false, move |keypair| Client {
+            nat: InnerNatBehaviour::create(
+                StdRng::from_entropy(),
+                nat_settings,
+                AddressMapperBehaviour::<TestMapper>::new(nat_settings.mapping),
+                GatewayMonitor::<SystemGatewayDetector>::new(nat_settings.gateway_monitor),
+            ),
+            identify: identify::Behaviour::new(identify::Config::new(
+                "/test/1.0.0".to_owned(),
```
ntn-x2: Maybe this can be a `const`?

nomos-libp2p/src/behaviour/nat/inner.rs
ntn-x2: At this point I think we could add a bunch of other tests? Like what happens if the server does not support identify? Or if the server does not support Autonat? I don't think it takes too much time to add those tests but it would give us an idea of exactly what errors to expect on our nodes?

nomos-libp2p/src/behaviour/nat/inner.rs
andrussal: > What is the need for this transport? What does it help us test?

I think Autonat always succeeds otherwise because it will be called on client listen address. This helps artificially to block server dials.

nomos-libp2p/src/behaviour/nat/inner.rs
andrussal: > At this point I think we could add a bunch of other tests? Like what happens if the server does not support identify? Or if the server does not support Autonat? I don't think it takes too much time to add those tests but it would give us an idea of exactly what errors to expect on our nodes?

Good idea, let me try add some others as well.

nomos-libp2p/src/behaviour/nat/inner.rs
ntn-x2: Mhh ok, makes sense. Can you maybe add a comment explaining that?

nomos-libp2p/src/behaviour/nat/inner.rs
andrussal: > There's a `listen` function on the `libp2p_swarm_test` crate which automatically picks a random port to start listening on, maybe we can do that here? In case we were to use the same port elsewhere.

This seems to work only with TCP, not with memory transport.

nomos-libp2p/src/behaviour/nat/inner.rs
andrussal: > At this point I think we could add a bunch of other tests? Like what happens if the server does not support identify? Or if the server does not support Autonat? I don't think it takes too much time to add those tests but it would give us an idea of exactly what errors to expect on our nodes?

I added another test for mapping failure which I think is second main flow. 

I played around a bit with edge cases when server doesn't support identify or no autonat server is available. They didn't seem to provide very useful insight to current nat behaviour though. In both cases nat client doesn't report anything and we would need to rely on this missing piece. But I think until we have it, these tests aren't very useful.

```rust
if let Poll::Ready(Some(_addr)) = self.next_autonat_client_tick.poll_unpin(cx) {
            // TODO: This is a placeholder for the missing API of the
            // autonat client
            // self.autonat_client_behaviour.retest_address(addr);
        }
```

nomos-libp2p/src/behaviour/nat/inner.rs
ntn-x2: You'd need to add a `with_memory_address_external` , e.g., https://github.com/logos-co/nomos/blob/a5b817c8c61519b5f974334138b75639a5f0fc86/nomos-blend/network/src/core/with_core/behaviour/tests/bootstrapping.rs#L27.

nomos-libp2p/src/behaviour/nat/inner.rs
ntn-x2: You mean that we don't get any useful event to consume?

andrussal: > You mean that we don't get any useful event to consume?

It seems that AutoNat client doesn't emit events if it doesn't have a server to connect. It reports failure if address if private but not if it can't test.

nomos-libp2p/src/behaviour/nat/inner.rs
andrussal: > You'd need to add a `with_memory_address_external` , e.g.,
> 
> https://github.com/logos-co/nomos/blob/a5b817c8c61519b5f974334138b75639a5f0fc86/nomos-blend/network/src/core/with_core/behaviour/tests/bootstrapping.rs#L27
> 
> .

Thanks, will use this!

PR #1602, opened 2025-08-28
feat(blend): membership service adapter

youngjoon-lee:

Implemented an adapter for the membership service.

This PR doesn't integrate the membership stream (returned from the membership service) to the system yet. I will work on it as a next PR since it requires some other work.

This is an incremental PR based on #1598.

Yes

@youngjoon-lee @ntn-x2 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/membership/mod.rs
+++ b/nomos-services/blend/src/membership/mod.rs
@@ -1,3 +1,4 @@
+pub mod node_id;
 pub mod service;
 
 use std::pin::Pin;
--- a/nomos-services/blend/src/membership/service.rs
+++ b/nomos-services/blend/src/membership/service.rs
@@ -1,11 +1,9 @@
-use std::collections::BTreeSet;
+use std::{collections::BTreeSet, hash::Hash, marker::PhantomData};
 
 use futures::StreamExt as _;
-use libp2p::{core::signed_envelope::DecodingError, PeerId};
 use nomos_blend_message::crypto::Ed25519PublicKey;
 use nomos_blend_scheduling::membership::{Membership, Node};
 use nomos_core::sdp::{Locator, ProviderId, ServiceType};
-use nomos_libp2p::ed25519;
 use nomos_membership::{
     backends::MembershipBackendError, MembershipMessage, MembershipSnapshotStream,
 };
@@ -16,9 +14,9 @@ use overwatch::{
 use tokio::sync::oneshot;
 use tracing::warn;
 
-use crate::membership::{MembershipStream, ServiceMessage};
+use crate::membership::{node_id, MembershipStream, ServiceMessage};
 
-pub struct Adapter<Service>
+pub struct Adapter<Service, NodeId>
 where
     Service: ServiceData,
 {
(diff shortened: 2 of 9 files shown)


nomos-services/blend/src/membership/service.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    /// Subscribe to membership updates.
+    ///
+    /// It returns a stream of [`Membership`] instances,
+    async fn subscribe(&self) -> Result<MembershipStream<Self::NodeId>, Self::Error> {
+        let signing_public_key = self.signing_public_key;
+        Ok(Box::pin(
+            self.subscribe_stream(ServiceType::BlendNetwork)
+                .await?
+                .map(|(_block_number, providers_map)| {
```
ntn-x2:
```suggestion
                .map(|(_, providers_map)| {
```

nomos-services/blend/src/membership/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+    Pin<Box<dyn Stream<Item = Membership<NodeId>> + Send + Sync + 'static>>;
+
+/// An adapter for the membership service.
+#[async_trait::async_trait]
+pub trait Adapter {
+    type Service: ServiceData;
+    type NodeId;
+    type Error: std::error::Error;
+
+    fn new(
```
ntn-x2, line 24: I find the fact that the creation is infallible but the `subscribe` is fallible not ideal, in my opinion. Why not making the creation fallible, but the `subscribe` infallible instead? Genuine question :)

nomos-services/blend/src/core/mod.rs
```diff
@@ -74,13 +77,18 @@ where
...
-impl<Backend, NodeId, Network, RuntimeServiceId> ServiceCore<RuntimeServiceId>
-    for BlendService<Backend, NodeId, Network, RuntimeServiceId>
+impl<Backend, NodeId, Network, MembershipAdapter, RuntimeServiceId> ServiceCore<RuntimeServiceId>
+    for BlendService<Backend, NodeId, Network, MembershipAdapter, RuntimeServiceId>
 where
     Backend: BlendBackend<NodeId, ChaCha12Rng, RuntimeServiceId> + Send + Sync,
     NodeId: Clone + Eq + Hash + Send + Sync + 'static,
     Network: NetworkAdapter<RuntimeServiceId, BroadcastSettings: Unpin> + Send + Sync,
+    MembershipAdapter: membership::Adapter + Send,
+    <<MembershipAdapter as membership::Adapter>::Service as ServiceData>::Message:
```
ntn-x2: I think this would be a perfect type alias defined in the `memberhsip` module.

ntn-x2 (commented):
Looks good! Just a couple of questions about the design of the adapter trait.

nomos-services/blend/src/core/mod.rs
youngjoon-lee: True. Thanks. Defined alias 3f529d3a68958cd44b895b2abd589242a33b919d

nomos-services/blend/src/membership/mod.rs
youngjoon-lee: I agree with your point, but I made the `subscribe` fallible because the membership service API is defined to return a `Result`.

https://github.com/logos-co/nomos/blob/f069ebadd033c615f06c7f31506d6abca42bef97/nomos-services/membership/src/lib.rs#L43-L47

ntn-x2: Yes ok, I was suggesting something like storing the result of `subscribe_stream` in the `new` method, so that if anything goes wrong, we would already know at construction step. Once we build the object, we know we can call `subscribe` on it, following Rust's RAII principle, somehow. But this works fine as well, won't make it too much of a deal :)

ntn-x2 (approved):
Great stuff!

nomos-services/blend/src/membership/mod.rs
youngjoon-lee: You're right because we're going to anyway call the `subscribe` method, which is the only one method that we have currently. 
I'd suggest to refactor all the adapters which have the same shape at once soon. Most of our adapters look like this. So, I'd like to refactor them at once.

PR #1608, opened 2025-08-29
fix(blend): relax the libp2p dependency for membership adapter

youngjoon-lee:
I realized that the membership adapter (added by the PR #1602) is tightly coupled with libp2p, though we've abstracted the node ID as `NodeId`. The CI is complaining about it: https://github.com/logos-co/nomos/actions/runs/17316057868/job/49159031139?pr=1602.

This PR resolves it by introducing the `membership::node_id::TryFrom` trait, which is used for converting a provider ID (returned by the membership service) into a `NodeId` (e.g. `libp2p::PeerId`).
Plus, I turned all libp2p-related dependencies to optional in the `Cargo.toml`.

If this looks good, we can use this practice for the membership adapter of the DA network service, if necessary. I referred to that when implementing #1602, but that's tightly coupled with libp2p.
https://github.com/logos-co/nomos/blob/5db0ae9ec9c7b00802e09dead91e7f4f55cbd613/nomos-services/data-availability/network/src/membership/adapters/service.rs#L66 

Yes

@youngjoon-lee @ntn-x2 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/membership/node_id/libp2p.rs
+++ b/nomos-services/blend/src/membership/node_id/libp2p.rs
@@ -13,9 +13,8 @@ impl super::TryFrom for PeerId {
 
 #[cfg(test)]
 mod tests {
-    use crate::membership::node_id::TryFrom as _;
-
     use super::*;
+    use crate::membership::node_id::TryFrom as _;
 
     #[test]
     fn test_try_from_provider_id() {


nomos-services/blend/src/membership/node_id/mod.rs
```diff
@@ -0,0 +1,8 @@
+#[cfg(feature = "libp2p")]
+pub mod libp2p;
+
+pub trait TryFrom: Sized {
+    type Error: std::error::Error;
+
+    fn try_from_provider_id(provider_id: &[u8]) -> Result<Self, Self::Error>;
```
ntn-x2, line 7: Mhh, why not using the `TryFrom` trait provided by Rust?

youngjoon-lee: because `TryFrom` is a foreign trait, and `PeerId` is also a foreign.

ntn-x2: Ok good point. Missed that. This works, although best practice would be to have a new type that wraps around `PeerId`, and implements `Deref` and `DerefMut` for it: https://doc.rust-lang.org/book/ch20-02-advanced-traits.html?highlight=orphan#using-the-newtype-pattern-to-implement-external-traits-on-external-types.

youngjoon-lee: Actually I tried it at the beginning but wasn't satisfied with the result. It made other code little more complicated. Let me check it again.

youngjoon-lee: Yeah, I tried it again, but couldn't make it simple to handling both the wrapper type and the exact type required for `Membership`. I'd like to go with the current approach for now and will try your suggestion again after finishing implementing the session transition.

```diff
 #[async_trait::async_trait]
-impl<Service, NodeId> super::Adapter for Adapter<Service, NodeId>
+impl<Service, NodeId, MembershipNodeId> super::Adapter
+    for Adapter<Service, NodeId, MembershipNodeId>
 where
     Service: ServiceData<Message = MembershipMessage>,
-    NodeId: node_id::TryFrom + Clone + Hash + Eq + Sync,
+    NodeId: for<'a> TryFrom<&'a [u8]> + Deref<MembershipNodeId> + Clone + Has>
 {
     type Service = Service;
-    type NodeId = NodeId;
+    type NodeId = MembershipNodeId;
     type Error = Error;
```

ntn-x2: Sounds good! It's a nit anyway.

PR #1611, opened 2025-08-29
feat(blend): session transition in proxy service

youngjoon-lee:

I'm opening a PR that integrates the session transition only to the proxy service, before moving forward. I want to check if we're on the same page. 

Since Overwatch services (core and edge) cannot be stopped/started on demand (afaik), the proxy service prepares all the relays at start-up and decide how to handle inbound messages according to the current membership.

~~If we're on the same page, I will add tests.~~ I added tests.

Yes

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/instance.rs
+++ b/nomos-services/blend/src/instance.rs
@@ -176,18 +176,8 @@ where
     ) -> Result<Self, modes::Error> {
         match self {
             Self::Core(mode) => Ok(Self::Core(mode)),
-            Self::Edge(mode) => {
+            mode => {
                 mode.shutdown().await;
-                Ok(Self::Core(Self::new_core_mode(overwatch_handle).await?))
-            }
-            Self::EdgeAfterCore { mode, prev } => {
-                prev.shutdown().await;
-                mode.shutdown().await;
-                Ok(Self::Core(Self::new_core_mode(overwatch_handle).await?))
-            }
-            Self::Broadcast(_) => Ok(Self::Core(Self::new_core_mode(overwatch_handle).await?)),
-            Self::BroadcastAfterCore { prev, .. } => {
-                prev.shutdown().await;
                 Ok(Self::Core(Self::new_core_mode(overwatch_handle).await?))
             }
         }
@@ -211,9 +201,8 @@ where
                 prev.shutdown().await;
                 Ok(Self::Edge(mode))
             }
-            Self::Broadcast(_) => Ok(Self::Edge(Self::new_edge_mode(overwatch_handle).await?)),
-            Self::BroadcastAfterCore { prev, .. } => {
-                prev.shutdown().await;
+            mode => {
+                mode.shutdown().await;
                 Ok(Self::Edge(Self::new_edge_mode(overwatch_handle).await?))
             }
         }
@@ -233,23 +222,16 @@ where
                 mode: Self::new_broadcast_mode(overwatch_handle).await?,
                 prev: mode,
             }),
-            Self::Edge(mode) => {
-                mode.shutdown().await;
-                Ok(Self::Broadcast(
-                    Self::new_broadcast_mode(overwatch_handle).await?,
-                ))
-            }
-            Self::EdgeAfterCore { mode, prev } => {
-                prev.shutdown().await;
-                mode.shutdown().await;
-                Ok(Self::Broadcast(
-                    Self::new_broadcast_mode(overwatch_handle).await?,
-                ))
-            }
             Self::Broadcast(mode) => Ok(Self::Broadcast(mode)),
             Self::BroadcastAfterCore { mode, prev } => {
                 prev.shutdown().await;
                 Ok(Self::Broadcast(mode))
+            }
+            mode => {
+                mode.shutdown().await;
+                Ok(Self::Broadcast(
+                    Self::new_broadcast_mode(overwatch_handle).await?,
+                ))
             }
         }
     }
@@ -271,7 +253,6 @@ where
     }
 
     /// Shuts down the instance by shutting down underlying modes.
-    #[cfg(test)]
     async fn shutdown(self) {
         match self {
             Self::Core(mode) => mode.shutdown().await,


nomos-services/blend/src/lib.rs
```diff
@@ -117,89 +126,306 @@ where
...
-                Some(Duration::from_secs(60)),
-                NetworkService<NetworkBackendOfService<CoreService, RuntimeServiceId>, _>,
-                MembershipService<EdgeService>
-            )
-            .await?;
-            let core_network_relay =
-                overwatch_handle
+
+        // Relays to the core and edge services.
+        let core_relay = overwatch_handle.relay::<CoreService>().await?;
```
ntn-x2: Why are we creating relays with core and edge services? These should exist only if the network is large enough, else it should not be possible to interact with them and the compiler should prevent that, which for me means they should be optional and instantiated if a new session is large enough, or deconstructed if the new session is too small.

nomos-services/blend/src/lib.rs
```diff
@@ -117,89 +126,306 @@ where
...
+        let edge_relay = overwatch_handle.relay::<EdgeService>().await?;
+
+        // A network adapter for broadcasting messages directly via gossipsub
+        // when the Blend network is too small.
+        let network_adapter = <CoreService::NetworkAdapter as NetworkAdapterTrait<
```
ntn-x2: This network adapter should also only ever exist if the network is too small. I think we can enforce these constraints at compile time so we don't misuse them. Or what was the reason to have them from the get-go instead? Anything I am not seeing?

ntn-x2 (commented):
I don't fully understand the reasoning behind the current design choices, so I'd love to follow your thought process. We know Overwatch allows us to start-stop services dynamically, and when we devised the existence of the proxy service, we said it would be responsible for starting-stopping services dynamically. Now, we have a nice session stream that not only ticks on new sessions (which we can use to start only the core or only the edge service, or stop a previously running edge service), but that also ticks on every session transition (which we can use to stop a previously running core service). Starting/stopping services is a dynamic operation so won't give us any compilation-time benefits, but we can still wrap those operations into special types, so that we can't do anything wrong if we don't follow the right order of operations. We are introducing a little bit more complicated inter-service dependencies, but I don't think the complexity is unmanageable if we design the right types to handle them. What do you think?

nomos-services/blend/src/lib.rs
youngjoon-lee: I agree with your point. Although my approach works, I agree that it's safer to start/stop underlying services on demand to avoid any potential misuses.

[1bec3f9](https://github.com/logos-co/nomos/pull/1611/commits/1bec3f9921a181fd71fee2d8dddc0fc814334eec) I applied it. The changes look big, but I tried to make them clean and correct. I recommend you to review the entire PR, not the specific commit.

nomos-services/blend/src/lib.rs
youngjoon-lee: This thread can be resolved by https://github.com/logos-co/nomos/pull/1611#discussion_r2317822125

nomos-services/blend/src/instance.rs
```diff
@@ -0,0 +1,715 @@
...
+        // Keep the previous core mode for the session transition period.
+        prev: CoreMode<CoreService, RuntimeServiceId>,
+    },
+    Broadcast(BroadcastMode<CoreService::NetworkAdapter, CoreService::Message, RuntimeServiceId>),
+    BroadcastAfterCore {
+        mode: BroadcastMode<CoreService::NetworkAdapter, CoreService::Message, RuntimeServiceId>,
+        // Keep the previous core mode for the session transition period.
+        prev: CoreMode<CoreService, RuntimeServiceId>,
+    },
+}
```
youngjoon-lee: I'm not 100% satisfied with this approach using enum. But, this is the optimal I've found so far. Even if you guys might have better ideas, this version works, so I'd like to move forward with it for now. I think this is not so bad and provides a reasonable level of the safety we intend. We can finish all the development first and then revisit this, if necessary.

nomos-services/blend/src/modes/ondemand.rs
```diff
@@ -0,0 +1,227 @@
...
+            }
+        };
+
+        Ok(Self {
+            relay,
+            overwatch_handle,
+        })
+    }
+
+    async fn stop_service(overwatch_handle: &OverwatchHandle<RuntimeServiceId>) {
```
ntn-x2: This does not need to be an associated function but can be a regular function, so we avoid monomorphizing it and save a little bit on compile time!

nomos-services/blend/src/modes/mod.rs
```diff
@@ -0,0 +1,28 @@
...
+use overwatch::services::relay::RelayError;
+
+#[cfg(test)]
+pub use crate::modes::broadcast::tests as broadcast_tests;
+pub use crate::modes::{broadcast::BroadcastMode, core::CoreMode, edge::EdgeMode};
+
+const LOG_TARGET: &str = "blend::service::modes";
+
+#[async_trait::async_trait]
+pub trait Mode<Message> {
```
ntn-x2: I honestly don't think this trait is required. All types used are concrete in this case, so we can call methods on them (also allowing for differences between them) instead of forcing them to comply with a trait for, as I understand, no benefits? If this were to be consumed in the service behind the trait and used for testing, it might make sense, but as things are now I don't think we need it and it introduced complexity which we don't need. What do you think?

nomos-services/blend/src/modes/broadcast.rs
```diff
@@ -0,0 +1,297 @@
...
+pub struct BroadcastMode<Adapter, Message, RuntimeServiceId> {
+    adapter: Adapter,
+    _phantom: PhantomData<(Message, RuntimeServiceId)>,
+}
+
+impl<Adapter, Message, RuntimeServiceId> BroadcastMode<Adapter, Message, RuntimeServiceId>
+where
+    Adapter: NetworkAdapter<RuntimeServiceId> + Send + Sync,
+{
+    pub async fn new<Service>(
```
ntn-x2:
```suggestion
    pub async fn new<NetworkService>(
```

nomos-services/blend/src/modes/broadcast.rs
```diff
@@ -0,0 +1,297 @@
...
+    services::{AsServiceId, ServiceData},
+};
+use services_utils::wait_until_services_are_ready;
+
+use crate::{
+    core::{network::NetworkAdapter, service_components::MessageComponents},
+    modes::{Error, Mode},
+};
+
+pub struct BroadcastMode<Adapter, Message, RuntimeServiceId> {
```
ntn-x2: If we keep the `Mode` trait, we don't need the `Message` generic since it's not used in the type impl and it's provided by the `Mode` trait. If we remove the `Mode` trait as I suggested in a different comment, we might not even need `RuntimeServiceId` if we simply define it in the `new` function, since we don't need it for the `handle_inbound_message` method either. So the definition of `BroadcastMode` would become much simpler.

ntn-x2 (approved):
I like it a lot! I am approving since this could be merged as is, but I provided a couple of suggestiong on how to simplify the codebase a bit, by removing the unnecessary trait and consequently some unnecessary generics. But this looks good, thank you!

nomos-services/blend/src/modes/mod.rs
danielSanchezQ: Yeah, probably a static dispatch (enum) them. Which you are kinda doing already iirc (?)

nomos-services/blend/src/instance.rs
danielSanchezQ: Not gonna lie, its a bit convoluted at first sight. But think it actually shows the cases really well. And will save a few if/else cases in places where we would not want them. So I think is fine.

ntn-x2: Yeah I think I would have gone with a state machine-like approach, which is a perfect fit for these cases, but this also works and it still gives us compile-time guarantees, which is the important bit here, so I have nothing against this solution.

nomos-services/blend/src/modes/ondemand.rs
youngjoon-lee: Good point. Thanks. cff269f28043306292e80991d21d4c84fa966960

nomos-services/blend/src/modes/mod.rs
youngjoon-lee: True. It's actually a leftover. I forgot to remove. I defined it when I tried the state machine pattern, but we don't need it now in the current approach. 1f37f1f8449aace5c3daa023f4f3c4f21c7ed573

nomos-services/blend/src/modes/broadcast.rs
youngjoon-lee: https://github.com/logos-co/nomos/pull/1611#discussion_r2321417076 Removed the `Mode` trait and simplified the definiotn of `BroadcastMode`.

nomos-services/blend/src/modes/broadcast.rs
youngjoon-lee: done 1f37f1f8449aace5c3daa023f4f3c4f21c7ed573

PR #1620, opened 2025-09-01
Wallet 

davidrusu:

> [!NOTE]  
> Review the 

> [!IMPORTANT]  
> explained.  

> [!IMPORTANT]  
> advancing  
> the review.

> [!IMPORTANT]

> [!IMPORTANT]
    modification/solutions.

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/nomos-services/wallet/src/lib.rs
+++ b/nomos-services/wallet/src/lib.rs
@@ -1,13 +1,17 @@
 use std::collections::HashSet;
 
 use async_trait::async_trait;
-use chain_service::{ConsensusMsg, CryptarchiaInfo};
+use chain_service::{
+    storage::{adapters::storage::StorageAdapter, StorageAdapter as StorageAdapterTrait},
+    ConsensusMsg, CryptarchiaInfo,
+};
 use kzgrs_backend::dispersal::BlobInfo;
 use nomos_core::{
+    block::Block,
     header::HeaderId,
-    mantle::{keys::PublicKey, Utxo, Value},
+    mantle::{keys::PublicKey, SignedMantleTx, Utxo, Value},
 };
-use wallet::{Wallet, WalletBlock, WalletError};
+use nomos_storage::backends::StorageBackend;
 use overwatch::{
     services::{
         state::{NoOperator, NoState},
@@ -16,6 +20,7 @@ use overwatch::{
     DynError, OpaqueServiceResourcesHandle,
 };
 use tokio::sync::oneshot;
+use wallet::{Wallet, WalletBlock, WalletError};
 
 #[derive(Debug)]
 pub enum WalletMsg {
@@ -37,17 +42,20 @@ pub struct WalletServiceSettings {
     pub known_keys: HashSet<PublicKey>,
 }
 
-pub struct WalletService<CryptarchiaService, RuntimeServiceId>
+pub struct WalletService<CryptarchiaService, Storage, RuntimeServiceId>
 where
     CryptarchiaService: ServiceData,
+    Storage: StorageBackend + Send + Sync + 'static,
 {
     service_resources_handle: OpaqueServiceResourcesHandle<Self, RuntimeServiceId>,
+    _storage: std::marker::PhantomData<Storage>,
 }
 
-impl<CryptarchiaService, RuntimeServiceId> ServiceData
-    for WalletService<CryptarchiaService, RuntimeServiceId>
+impl<CryptarchiaService, Storage, RuntimeServiceId> ServiceData
+    for WalletService<CryptarchiaService, Storage, RuntimeServiceId>
 where
     CryptarchiaService: ServiceData,
+    Storage: StorageBackend + Send + Sync + 'static,
 {
     type Settings = WalletServiceSettings;
     type State = NoState<Self::Settings>;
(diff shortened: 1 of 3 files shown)


nodes/nomos-node/node/src/main.rs
```diff
@@ -69,6 +69,16 @@ async fn main() -> Result<()> {
...
+            wallet: wallet::WalletServiceSettings {
+                known_keys: {
+                    let mut keys = std::collections::HashSet::new();
+                    let sample_key = nomos_core::mantle::keys::PublicKey::from(
+                        num_bigint::BigUint::from(12345u64),
+                    );
+                    keys.insert(sample_key);
+                    keys
+                },
+            },
```
danielSanchezQ: We should load this from config.

nomos-services/chain-service/src/leadership.rs
```diff
@@ -44,9 +44,11 @@ impl Leader {
     ) -> Option<Risc0LeaderProof> {
         for utxo in &self.utxos {
             let Some(_aged_witness) = aged_tree.witness(&utxo.id()) else {
+                println!("leader: skipping utxo since it's not aged: {}");
                 continue;
             };
             let Some(_latest_witness) = latest_tree.witness(&utxo.id()) else {
+                println!("skipping utxo since it's not in latest");
```
danielSanchezQ: This should be logs if we want to keep them

nomos-services/chain-service/src/lib.rs
```diff
@@ -1037,6 +1039,11 @@ where
                 tx.send(res)
                     .unwrap_or_else(|_| error!("could not send blocks through channel"));
             }
+            ConsensusMsg::GetLedgerState { block_id, tx } => {
+                let ledger_state = cryptarchia.ledger.state(&block_id).cloned();
+                tx.send(ledger_state)
+                    .unwrap_or_else(|_| error!("could not send ledger state through channel"));
```
danielSanchezQ, line 1045: this is better as a let Err(e) match instead.

nomos-services/wallet/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+        chain_relay
+            .send(
+                ConsensusMsg::GetLedgerState {
+                    block_id: lib,
+                    tx: lib_state_tx,
+                }
+                .into(),
+            )
+            .await
+            .map_err(|e| format!("Failed to request LIB ledger state: {:?}", e.0))?;
```
danielSanchezQ, line 183: Ideally we have a proper set of errors. Does String implement `DynError`? 🤔

nomos-services/wallet/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+                    Self::handle_wallet_message(msg, &mut wallet).await;
+                }
+                Ok(block) = block_receiver.recv() => {
+                    let wallet_block = wallet::WalletBlock::from(block);
+                    match wallet.apply_block(&wallet_block) {
+                        Ok(()) => {
+                            eprintln!("Applied block {:?} to wallet", wallet_block.id);
+                        }
+                        Err(e) => {
+                            eprintln!("Failed to apply block to wallet: {:?}", e);
```
danielSanchezQ: This should be logs I think.

nomos-services/wallet/src/wallet.rs
danielSanchezQ, line 1: This and the error.rs files, should live in their own wallet crate. But we can do that as a separated PR.

nomos-services/wallet/src/wallet.rs
```diff
@@ -0,0 +1,410 @@
...
+impl WalletState {
+    pub fn from_ledger(known_keys: &HashSet<PublicKey>, ledger: LedgerState) -> Self {
+        let utxos = BTreeMap::from_iter(
+            ledger
+                .latest_commitments()
+                .utxos()
+                .iter()
+                .map(|(_, (utxo, _))| *utxo)
+                .filter(|utxo| known_keys.contains(&utxo.note.pk))
+                .map(|utxo| (utxo.id(), utxo)),
```
danielSanchezQ: This could be a single `filter_map` instruction instead

nomos-services/wallet/src/wallet.rs
```diff
@@ -0,0 +1,410 @@
...
+
+    pub fn utxos_for_amount(
+        &self,
+        amount: Value,
+        pks: impl IntoIterator<Item = PublicKey>,
+    ) -> Option<Vec<Utxo>> {
+        let mut utxos: Vec<Utxo> = pks
+            .into_iter()
+            .flat_map(|pk| self.pk_index.get(&pk))
+            .flatten()
```
danielSanchezQ: This flatten is redundant? Anyway the `flat_map` + `flatten` + `map` could be a single `flat_map` I think.

nomos-services/wallet/src/wallet.rs
```diff
@@ -0,0 +1,410 @@
...
+        known_keys: impl IntoIterator<Item = PublicKey>,
+        lib: HeaderId,
+        ledger: LedgerState,
+    ) -> Self {
+        let known_keys: HashSet<PublicKey> = known_keys.into_iter().collect();
+
+        let wallet_state = WalletState::from_ledger(&known_keys, ledger);
+
+        Self {
+            known_keys: known_keys.into_iter().collect(),
```
danielSanchezQ: this into_iter + collect is redundant too.

nomos-services/chain-service/src/lib.rs
davidrusu: I was following the other consensus message cases above, they all use this same pattern

nomos-services/wallet/src/wallet.rs
davidrusu: It would lead to a some nested iteration, not sure it's clearer to read.
```
        let mut utxos: Vec<Utxo> = pks
            .into_iter()
            .filter_map(|pk| self.pk_index.get(&pk))
            .flat_map(|note_ids| note_ids.iter().map(|id| self.utxos[id]))
            .collect();
```

for now, I changed the first flat_map to filter_map to make it clearer what's going on (they are semantically identical since Option<T> implements IntoIterator as Some(x) => iter::once(x) and None => iter::never())

nomos-services/wallet/src/wallet.rs
davidrusu: Then I would look like this:

```rust
        let utxos = BTreeMap::from_iter(ledger.latest_commitments().utxos().iter().filter_map(
            |(id, (utxo, _))| {
                if known_keys.contains(&utxo.note.pk) {
                    Some((id, *utxo))
                } else {
                    None
                }
            },
        ));
```

Do you prefer it like this? I like seeing each transformation as a different step but I am flexible!

nomos-services/chain-service/src/lib.rs
danielSanchezQ: True. ok then, we will change all of them in due time. Thanks!

nomos-services/wallet/src/wallet.rs
danielSanchezQ: How about

```rust
        let utxos = BTreeMap::from_iter(ledger.latest_commitments().utxos().iter().filter_map(
            |(id, (utxo, _))| known_keys.contains(&utxo.note.pk).then((id, *utxo))
        ));
```
?

nomos-services/wallet/src/wallet.rs
danielSanchezQ: Ah yes I see. Ok then, original version was fine, although I didn't got it well at the beginning. Sorry for the troubles.

PR #1638, opened 2025-09-04
feat(blend): session transition for edge service

youngjoon-lee:

This is an incremental PR based on the PR #1611.

This PR updates the edge service to properly handle new sessions.

I left a comment below to highlight key changes.

Just FYI, I thought about recreating the edge service even when we are transitioning from Edge to Edge. In that way, I guessed we could pass a single membership to the edge service, instead of passing the session stream to it. If it's possible, we could make the edge service structure much simpler. But, I didn't choose that way. Here, I'm not describing all details why I didn't choose it, but we can discuss separately if you think it's worth trying it. Anyway, I believe the approach in this PR is good enough.

Yes

yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-da/network/core/src/protocols/dispersal/executor/behaviour.rs
+++ b/nomos-da/network/core/src/protocols/dispersal/executor/behaviour.rs
@@ -418,7 +418,6 @@ where
         let members = self.membership.members_of(&subnetwork_id);
         let peers: Vec<_> = members
             .iter()
-            .filter(|peer_id| self.connected_peers.contains_key(peer_id))
             .filter(|peer_id| !self.pending_peer_open_stream_requests.contains(peer_id))
             .collect();
 


nomos-blend/scheduling/src/membership.rs
```diff
@@ -44,13 +44,13 @@ where
     NodeId: Clone + Hash + Eq,
 {
     #[must_use]
-    pub fn new(nodes: &[Node<NodeId>], local_public_key: Option<&Ed25519PublicKey>) -> Self {
+    pub fn new(nodes: &[Node<NodeId>], local_public_key: &Ed25519PublicKey) -> Self {
```
youngjoon-lee, line 47: KEY CHANGE 3/3:
The `local_public_key` doesn't need to be optional. Better to make it mandatory to avoid potential misuses.

nomos-services/blend/src/edge/backends/mod.rs
```diff
@@ -19,8 +16,7 @@ where
     fn new<Rng>(
         settings: Self::Settings,
         overwatch_handle: OverwatchHandle<RuntimeServiceId>,
-        session_stream: Pin<Box<dyn Stream<Item = Membership<NodeId>> + Send>>,
-        current_membership: Option<Membership<NodeId>>,
+        membership: Membership<NodeId>,
         rng: Rng,
     ) -> Self
```
youngjoon-lee, line 21: KEY CHANGE 1/3:
Now we don't pass a session stream to the backend. Instead, we recreate a backend with a new membership whenever a new session is started. It's simpler. We don't need to have any special logic for session transition inside the backend.

nomos-services/blend/src/edge/mod.rs
```diff
@@ -87,118 +91,557 @@ where
...
+                    "Service '{}' is ready.",
+                    <RuntimeServiceId as AsServiceId<Self>>::SERVICE_ID
+                );
+            },
+        )
+        .await
+        .map_err(|e| {
+            error!(target: LOG_TARGET, "Edge blend service is being terminated with error: {e}");
+            e.into()
+        })
```
youngjoon-lee: KEY CHANGE 2/3:
I wrapped all the service logic into the `run` function for easy testing. This function returns an error if the initial or any new membership does not satisfy the edge node condition.

We discussed to make the service panic on those errors. But, in this PR, we only prints a log and returns the error, so that the service can be terminated. I'm open to change this to panic, but I found that the panic is not propagated to the user who started the service. In other words, regardless of whether we panic or return an error, the user can detect it by checking the result from `wait_until_services_are_ready!`. So, instead of panicking, I followed the common way (returning an error) that we're using for other services. But again, we can easily change this to panic if we think it's better.

nomos-services/blend/src/edge/mod.rs
```diff
@@ -87,118 +91,557 @@ where
...
+/// Run the event loop of the service.
+///
+/// It returns an [`Error`] immediately if the initial membership does not
+/// satisfy the edge node condition defined in [`check_edge_condition`].
+/// Otherwise, it notifies readiness and starts the event loop.
+///
+/// In the event loop, it listens for new sessions and messages to blend,
+/// It recreates the [`MessageHandler`] on each new session to handle messages
+/// with the new membership.
+/// It returns an [`Error`] if the new membership does not satisfy the edge node
```
ntn-x2, line 171: Repetition of few lines above?

nomos-services/blend/src/edge/mod.rs
```diff
@@ -87,118 +91,557 @@ where
...
+    settings: &Settings<Backend, NodeId, RuntimeServiceId>,
+    overwatch_handle: &OverwatchHandle<RuntimeServiceId>,
+    notify_ready: impl Fn(),
+) -> Result<(), Error>
+where
+    Backend: BlendBackend<NodeId, RuntimeServiceId> + Sync,
+    NodeId: Clone + Eq + Hash + Send + Sync + 'static,
+    RuntimeServiceId: Clone,
+{
+    // Read the initial membership, expecting it to be yielded immediately.
```
ntn-x2, line 190: Should we have a very small timeout and fail in case we get no value back? This is something we need to clarify with the others. What happens if a new node is started up mid session? Does it get a tick immediately? Probably yes. Still, if that does not happen, this service should be resilient against that. So I'd suggest having a timeout of 5 seconds within which the latest membership info should be provided.

nomos-services/blend/src/edge/mod.rs
```diff
@@ -87,118 +91,557 @@ where
...
+    let SessionEvent::NewSession(membership) = session_stream
+        .next()
+        .await
+        .ok_or(Error::SessionStreamClosed)?
+    else {
+        return Err(Error::UnexpectedSessionEvent);
+    };
+
+    check_edge_condition(&membership, settings.minimum_network_size.get() as usize)?;
```
ntn-x2: This is good, but is there any way we can trigger a compilation failure in case we accidentally remove this line? Maybe we could return a message handler?

nomos-services/blend/src/edge/mod.rs
```diff
@@ -87,118 +91,557 @@ where
...
+    membership: Membership<NodeId>,
+    settings: &Settings<Backend, NodeId, RuntimeServiceId>,
+    overwatch_handle: &OverwatchHandle<RuntimeServiceId>,
+) -> Result<MessageHandler<Backend, NodeId, RuntimeServiceId>, Error>
+where
+    Backend: BlendBackend<NodeId, RuntimeServiceId>,
+    NodeId: Clone + Eq + Hash + Send + 'static,
+    RuntimeServiceId: Clone,
+{
+    check_edge_condition(&membership, settings.minimum_network_size.get() as usize)?;
```
ntn-x2: Yes same here, I think it would be much better if `check_edge_condition` would be called `new_message_handler` and it would fail if the required condition is not met. Then, if we accidentally delete this line, we would get a compilation issue since we would not know how to create a `MessageHandler`, assuming its constructor is private and can only be called from inside the checking function, so moving the type in a different module and perhaps expose a method on the type itself that would check for the condition. What do you think?

nomos-services/blend/src/edge/mod.rs
```diff
@@ -87,118 +91,557 @@ where
...
+                tracing::error!(target: LOG_TARGET, "Failed to encapsulate message: {e:?}");
+            })
+        else {
+            return;
+        };
+        self.backend.send(message).await;
+    }
+}
+
+#[cfg(test)]
```
ntn-x2: Can we move the tests into its own module?

nomos-services/blend/src/edge/mod.rs
```diff
@@ -87,118 +91,557 @@ where
...
+        msg_sender.send(vec![0]).await.expect("channel opened");
+        assert_eq!(
+            node_id_receiver.recv().await.expect("channel opened"),
+            core_node
+        );
+    }
+
+    /// [`run`] ignores [`SessionEvent::TransitionPeriodExpired`].
+    #[test_log::test(tokio::test)]
+    async fn run_ignores_transtition_period_expired() {
```
ntn-x2:
```suggestion
    async fn run_ignores_transition_period_expired() {
```

nomos-blend/network/src/core/with_edge/behaviour/tests/utils.rs
```diff
@@ -53,7 +53,7 @@ impl BehaviourBuilder {
                     })
                     .collect::<Vec<_>>()
                     .as_ref(),
-                None,
+                &Ed25519PrivateKey::generate().public_key(),
```
ntn-x2: Might also make sense to have a function that only takes a vec of nodes and automatically generate a random public key locally, and that function could be feature-gated behind an `unsafe-test-functions` feature. But this is a nit so that we don't have to repeat this everywhere.

ntn-x2 (commented):
I agree removing the session stream from the edge swarm was a very nice change. Besides some minor suggestions, my question would still remain the same: I think also with the current design, we want the edge service to panic. In the current approach, if the `run` function fails, we log the error but we never report to overwatch about the failure. So, I thinkin panicking would actually be the way to signal to overwatch "Hey, I'm out", and overwatch COULD technically report that to services that depend on this one, erroneously. Also, not sure what happens with the inbound relay of the edge service, but we probably want an error generated when the message is sent from another service, after the edge service failed to start. Maybe panicking does that because we free up resources, not sure if simply getting out of the `run` loop would do the same.

As for starting/stopping the edge service on edge->edge transitions, I agree with you it's not necessary, nor desirable, I think. We just need to handle the case in which the node transitions from edge -> core, maybe the edge service could shut itself down automatically? So we don't run into issues where the new membership is received by the edge service before it is shut down.

nomos-services/blend/src/edge/mod.rs
youngjoon-lee: > I agree removing the session stream from the edge swarm was a very nice change. Besides some minor suggestions, my question would still remain the same: I think also with the current design, we want the edge service to panic. In the current approach, if the run function fails, we log the error but we never report to overwatch about the failure. So, I thinkin panicking would actually be the way to signal to overwatch "Hey, I'm out", and overwatch COULD technically report that to services that depend on this one, erroneously. Also, not sure what happens with the inbound relay of the edge service, but we probably want an error generated when the message is sent from another service, after the edge service failed to start. Maybe panicking does that because we free up resources, not sure if simply getting out of the run loop would do the same.

@ntn-x2 I'm here replying to your suggestion. In short, I agree with using panic, as the same reason you mentioned.

> We just need to handle the case in which the node transitions from edge -> core, maybe the edge service could shut itself down automatically? So we don't run into issues where the new membership is received by the edge service before it is shut down.

In this PR, the service exits from the `ServiceCore::run` function if the node is not an edge node in a **new** membership. 

-----------

When we put these two points together, we arrive at one decision to make.

1. Do we want to panic only if the **initial** membership doesn't satisfy the condition?
2. Or, do we also want to panic if any new membership doesn't satisfy the condition?

I think the 1 is enough. I will first modify this PR to address it.

nomos-services/blend/src/edge/mod.rs
youngjoon-lee: I thought these unit tests are better to be in the same file. Is it the better practice to have the tests in a separate module?

ntn-x2: In this case we are testing the `mod.rs` file, so having the tests in a separate file called `tests` is equivalent to have them in `mod.rs` within a submodule. I personally prefer to always have a `mod.rs` file and a `tests.rs` file so that all the logic is in the same place. Ideally, you could add more logic at the end of a file, after the `tests` submodule, and that would be ok. So it's more as a way to keep stuff self-contained and easy to follow. But this is a nit.

nomos-services/blend/src/edge/mod.rs
ntn-x2: Yes. I think it's enough to panic only in the first case, I agree. As for the second case, I was suggesting to use overwatch so that we would have an overwatch log line in there, and also so that overwatch might react to that event by, for instance, notifying other services. If we simply let the `run` function exit, I think it does not offer too many opportunities to notices it nor to react to it. But in general I agree, a new session should not lead to a panic.

nomos-services/blend/src/edge/mod.rs
youngjoon-lee: Updated the comment according to the updated logic: e4f84867b4d6820e7c7da3dd3c4d1af1a47ea6df

nomos-services/blend/src/edge/mod.rs
youngjoon-lee: e21689a7f600b3b999a147a6cdf456d74bd6859a I improved this code to panic if the stream doesn't yield immediately. This stream should be anyway immediate. If not, the proxy service should wait until the next session is started, which is not the behavior we want. Fortunately, the current implementation of the membership service yields the current membership immediately. I will suggest the team to enforce this in the membership service level.

nomos-services/blend/src/edge/mod.rs
youngjoon-lee: 81d95c6d7f8905e574cd67dbbc0479013590f903 I applied the approach. Thanks. Very helpful.

nomos-services/blend/src/edge/mod.rs
youngjoon-lee: can be resolved by https://github.com/logos-co/nomos/pull/1638#discussion_r2322199352

nomos-services/blend/src/edge/mod.rs
youngjoon-lee: I agree. thanks. f712f1c2176d15dcc1c4546acac8fede5bd646da

nomos-blend/network/src/core/with_edge/behaviour/tests/utils.rs
youngjoon-lee: good point. ecabd0807efc979921945d721cebaa63c89479d6

nomos-services/blend/src/edge/mod.rs
youngjoon-lee: @ntn-x2 I realized that the [`now_or_never`](https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.now_or_never) that I added in https://github.com/logos-co/nomos/commit/e21689a7f600b3b999a147a6cdf456d74bd6859a doesn't work for `tokio::time::interval` (which we currently use as a mock), even though its [doc](https://docs.rs/tokio/latest/tokio/time/fn.interval.html) says that it yields the first item immediately. It's because it calls [`sleep_until(Instant::now())`](https://docs.rs/tokio/latest/tokio/time/fn.sleep_until.html) at the beginning, which is not strictly immediate. So, I replaced it with [`timeout`](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html) (1s) to tolerate small delays: 419fa2d0.
I will merge this once CI passes. Thank you so much!

PR #1642, opened 2025-09-04
feat(core): Update block format

zeegomo:
Initial step in updating the block format to follow https://www.notion.so/nomos-tech/Block-Construction-Validation-and-Execution-Specification-215261aa09df8164b5a9f0783c0e99b9#215261aa09df8162a641f53e06ed3236.
This is mostly a removal of old leftovers in the block format + initial sketch of new structures.
In addition, bounds on txs have been modified from `Tx` into `AuthenticatedMantleTx` as:
* that's what the ledger will accept
* we need to peek into the txs to determine blob status

You can expect the following changes in future PRs:
* PoL updates
* Conversion of proposal / references into a block
* Validation
* Integration into the node
Yes

@madxor , @thomaslavaur authors, @zeegomo implementing

Yes

No, but it's not a complete implementation yet

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/chain-service/Cargo.toml
+++ b/nomos-services/chain-service/Cargo.toml
@@ -39,7 +39,6 @@ tracing-futures        = { default-features = false, version = "0.2" }
 utoipa                 = { optional = true, workspace = true }
 
 [dev-dependencies]
-kzgrs-backend = { workspace = true }
 nomos-network = { features = ["mock"], workspace = true }
 nomos-storage = { features = ["rocksdb-backend"], workspace = true }
 tempfile      = { default-features = false, version = "3" }
--- a/tests/src/tests/da/api.rs
+++ b/tests/src/tests/da/api.rs
@@ -18,6 +18,7 @@ use tests::{
 };
 
 #[tokio::test]
+#[ignore = "Reenable after transaction mempool is used"]
 #[serial]
 async fn test_get_share_data() {
     let topology = Topology::spawn(TopologyConfig::validator_and_executor()).await;
@@ -157,6 +158,7 @@ async fn test_block_peer() {
 
 #[tokio::test]
 #[serial]
+#[ignore = "Reenable after transaction mempool is used"]
 async fn test_get_shares() {
     let topology = Topology::spawn(TopologyConfig::validator_and_executor()).await;
     let executor = &topology.executors()[0];
--- a/tests/src/tests/da/disperse.rs
+++ b/tests/src/tests/da/disperse.rs
@@ -195,6 +195,7 @@ async fn four_subnets_disseminate_retrieve_reconstruct() {
 
 #[tokio::test]
 #[serial]
+#[ignore = "Reenable after transaction mempool is used"]
 async fn disseminate_same_data() {
     const ITERATIONS: usize = 10;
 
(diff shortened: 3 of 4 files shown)


nomos-services/chain-service/src/lib.rs
```diff
@@ -1209,88 +1175,72 @@ where
...
-                .with_transactions(cl_txs)
-                .with_blobs_info(
-                    da_blobs_info.filter(move |info| blobs_ids.contains(&info.blob_id())),
-                )
-                .build()
-                .expect("Proposal block should always succeed to be built");
+        let txs = get_mempool_contents(relays.cl_mempool_relay().clone()).await;
+        match txs {
+            Ok(txs) => {
+                let txs = tx_selector.select_tx_from(txs).collect::<Vec<_>>();
```
bacv: Is the idea here to move the sampling result checking into the tx_selector, or should we continue appending to txs in this block by directly requesting a list of sampled blob IDs, like we did before?

bacv (approved):
NIce! I think the integration tests that need blob_id should get the `#[ignore]` label for now, those won't pass unlit validators publish blob related transactions to correct mempool.

nomos-core/chain-defs/src/header/mod.rs
```diff
@@ -19,31 +21,28 @@ pub struct Nonce([u8; 32]);
...
-        // TODO: header hash missing in specs
+        h.update(b"BLOCK_ID_V1");
+        h.update([BEDROCK_VERSION]);
+        h.update(self.parent_block.0);
+        h.update(self.slot.to_le_bytes());
+        h.update(self.block_root.0);
+        // TODO: add leader voucher
+        h.update(self.proof_of_leadership.entropy()); // TODO: blake2b to Fr
+                                                      // TODO: serialize proof
+                                                      // TODO: add leader key
```
danielSanchezQ, line 45: Is this hash or cyptohash in the specs? why Blake2b to Fr?

nomos-core/chain-defs/src/header/mod.rs
```diff
@@ -55,82 +54,26 @@ impl Header {
...
-            leader_proof,
-            orphaned_leader_proofs: vec![],
-        }
-    }
-
-    #[must_use]
-    pub fn with_orphaned_proofs(mut self, orphaned_leader_proofs: Vec<Self>) -> Self {
-        self.orphaned_leader_proofs = orphaned_leader_proofs;
-        self
-    }
```
danielSanchezQ, line 98: 😢 they are gone... 😂

danielSanchezQ (approved):
Nice to see thing getting cleaned up. Looking neat!

nomos-core/chain-defs/src/header/mod.rs
zeegomo: leader voucher is a zkhash, so a `Fr`, but the current implementation uses blake2b so it's just 32bytes, we will need to update this to a `Fr` and to the appropriate byte conversion

nomos-core/chain-defs/src/header/mod.rs
zeegomo: yeeeeees

nomos-services/chain-service/src/lib.rs
zeegomo: Ah, good catch, I see what you mean. Let's keep the second not to add additional stuff in here, we can think about ways to improve it later

PR #1644, opened 2025-09-05
refactor(test): random ports

andrussal:

Use random ports in tests to avoid port already in use. 

```
ailures:

---- backends::libp2p::swarm::tests::test_kademlia_bootstrap stdout ----

thread 'backends::libp2p::swarm::tests::test_kademlia_bootstrap' panicked at nomos-services/network/src/backends/libp2p/swarm/mod.rs:67:48:
called `Result::unwrap()` on an `Err` value: Other(Custom { kind: Other, error: Other(Transport(Right(Io(Os { code: 48, kind: AddrInUse, message: "Address already in use" })))) })
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-da/network/core/Cargo.toml
+++ b/nomos-da/network/core/Cargo.toml
@@ -36,6 +36,7 @@ kzgrs              = { workspace = true }
 kzgrs-backend      = { features = ["testutils"], workspace = true }
 libp2p             = { features = ["plaintext", "yamux"], workspace = true }
 libp2p-swarm-test  = { default-features = false, features = ["tokio"], version = "0.5.0" }
+nomos-utils        = { workspace = true }
 rstest             = { default-features = false, version = "0.25" }
 tokio              = { default-features = false, version = "1" }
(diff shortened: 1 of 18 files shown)


danielSanchezQ:
> > How about a dispatcher of ports from a static list? That way we would never have collisions (?)
> 
> I think this would help to avoid bunch of potential collisions. Only thing that comes to mind is that in some tests we use `swarm.listen()` that internally finds a free port(I think lets OS to choose). What do you think of using something like this instead of `rng` in places where we need to get port for config? (don't call listen in test directly)
> 
> ```rust
> pub fn get_available_port() -> u16 {
>     loop {
>         let port = NET_PORT.fetch_add(1, Ordering::SeqCst);
>         if TcpListener::bind(("127.0.0.1", port)).is_ok() {
>             return port;
>         }
>     }
> }
> ```
> 
> There's also few libs that dot this kind of stuff.

Sounds good. If we can have a function that returns a free from os would be superb.

Issue #1533, opened 2025-08-11
Core: Handle new sessions


youngjoon-lee:
https://github.com/logos-co/nomos/blob/22f40f9d7e368f02ea37c439b1f55819f9dfec29/nomos-services/blend/src/core/backends/libp2p/swarm.rs#L318-L321

As defined in the [Session Transition](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df8156a1d2dc9e7f81d289) spec,
- Open new connections to core nodes in the new membership, if the node is one of the core nodes in the new membership.
- Maintain all the connections with the core nodes in the old membership, and process all old messages for the STP (Session Transition Period). Close old connections after STP has passed.
- Clear the message cache after STP has passed.
  https://github.com/logos-co/nomos/blob/22f40f9d7e368f02ea37c439b1f55819f9dfec29/nomos-blend/network/src/core/with_core/behaviour/mod.rs#L87-L89

PR #1647, opened 2025-09-05
feat(blend): build SessionInfo stream from the real SessionEvent stream

youngjoon-lee:

This is the `1/N`-th PR for #1533.

We have `MessageScheduler` which reads a `Stream<SessionInfo>`. Until now, we've been passing it a mock stream that repeatedly yielded the same value.

This PR updates it to provide the real session stream to the `MessageScheduler`.

Next PRs will be coming soon.

Yes

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/test_utils/membership.rs
+++ b/nomos-services/blend/src/test_utils/membership.rs
@@ -1,24 +1,31 @@
+use std::hash::Hash;
+
 use libp2p::Multiaddr;
 use nomos_blend_message::crypto::{Ed25519PrivateKey, Ed25519PublicKey};
 use nomos_blend_scheduling::membership::{Membership, Node};
 
-pub type NodeId = u8;
-
-pub fn membership(ids: &[NodeId], local_id: NodeId) -> Membership<NodeId> {
+pub fn membership<NodeId>(ids: &[NodeId], local_id: NodeId) -> Membership<NodeId>
+where
+    NodeId: Clone + Eq + Hash,
+    [u8; 32]: From<NodeId>,
+{
     Membership::new(
         &ids.iter()
             .map(|id| Node {
-                id: *id,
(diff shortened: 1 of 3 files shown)


nomos-services/blend/src/test_utils/membership.rs
```diff
@@ -0,0 +1,24 @@
+use libp2p::Multiaddr;
+use nomos_blend_message::crypto::{Ed25519PrivateKey, Ed25519PublicKey};
+use nomos_blend_scheduling::membership::{Membership, Node};
+
+pub type NodeId = u8;
+
+pub fn membership(ids: &[NodeId], local_id: NodeId) -> Membership<NodeId> {
```
ntn-x2: Mhh I think `NodeId` should be a generic of this function. This is a test util, so the concrete type should be defined by the test case.

nomos-services/blend/src/test_utils/membership.rs
```diff
@@ -0,0 +1,24 @@
...
+                id: *id,
+                address: Multiaddr::empty(),
+                public_key: key(*id).1,
+            })
+            .collect::<Vec<_>>(),
+        &key(local_id).1,
+    )
+}
+
+pub fn key(id: NodeId) -> (Ed25519PrivateKey, Ed25519PublicKey) {
```
ntn-x2: Same here. We probably want `NodeId` to be generic.

nomos-services/blend/src/test_utils/membership.rs
youngjoon-lee: makes sense. done [becbee3](https://github.com/logos-co/nomos/pull/1647/commits/becbee363b53dca6154eb05af82236367a045e82)

PR #1648, opened 2025-09-05
feat(membership): sessions per service type

pradovic:

This PR implements support for different session intervals per service type, as data availability and blend don't have the same session sizes in blocks.

Yes

@pradovic 
Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/nomos-services/membership/src/backends/mock.rs
+++ b/nomos-services/membership/src/backends/mock.rs
@@ -12,53 +12,63 @@ use serde::{Deserialize, Serialize};
 use super::{MembershipBackend, MembershipBackendError};
 use crate::{backends::NewSesssion, MembershipProviders};
 
-type MockMembershipEntry = HashMap<ServiceType, HashSet<ProviderId>>;
+#[derive(Debug, Clone)]
+struct SessionState {
+    session_number: SessionNumber,
+    membership: HashSet<ProviderId>,
+    locators: HashMap<ProviderId, BTreeSet<Locator>>,
+}
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct MockMembershipBackendSettings {
-    pub session_size_blocks: u32,
-    pub session_zero_membership: MockMembershipEntry,
-    pub session_zero_locators_mapping: HashMap<ServiceType, HashMap<ProviderId, BTreeSet<Locator>>>,
+    pub session_sizes: HashMap<ServiceType, u32>,
+    // Initial state for each service type
+    pub session_zero_memberships: HashMap<ServiceType, HashSet<ProviderId>>,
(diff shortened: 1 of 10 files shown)


nomos-services/membership/src/backends/mock.rs
```diff
@@ -12,55 +12,71 @@ use serde::{Deserialize, Serialize};
 use super::{MembershipBackend, MembershipBackendError};
 use crate::{backends::NewSesssion, MembershipProviders};
 
-type MockMembershipEntry = HashMap<ServiceType, HashSet<ProviderId>>;
+#[derive(Debug, Clone)]
+struct SessionState {
+    session_number: SessionNumber,
+    membership: HashSet<ProviderId>,
+    locators: HashMap<ProviderId, BTreeSet<Locator>>,
```
youngjoon-lee: nit: could we probably merge these two into a single hashmap?

nomos-services/membership/src/backends/mock.rs
```diff
@@ -78,101 +94,84 @@ impl MembershipBackend for MockMembershipBackend {
...
+                    // Get snapshot before promoting
+                    completed_sessions.insert(*service_type, promoted_session.to_snapshot());
+
+                    // Promote forming to active
+                    self.active_sessions
+                        .insert(*service_type, promoted_session.clone());
+
+                    // Start new forming session with incremented session number
+                    let mut new_forming = promoted_session;
+                    new_forming.session_number = next_session + 1;
```
youngjoon-lee, line 118: Question: It seems that this was previously `self.forming_sessions_id += 1`. Now is it correct to use `next_session + 1`?

pradovic: In the old code, there was a single global `forming_session_id` that incremented. Now each service type has its own session progression. The logic is:

`next_session` = the session we're transitioning TO (calculated as `(block_number + 1) / session_size`)
`next_session + 1` = the NEW forming session (one ahead of what just became active)

This is should be correct (if I am not mistaken) because:

When we promote:
-  `next_session` becomes active
- the forming session should always be one ahead, so next_session + 1

Does this makes sense?

nomos-services/membership/src/backends/mock.rs
pradovic: It would look like something like:
```
struct SessionState {
    session_number: SessionNumber,
    providers: HashMap<ProviderId, BTreeSet<Locator>>,
}
```

You are correct, I will apply it to configs as well. Thanks!

nomos-services/membership/src/backends/mock.rs
youngjoon-lee: Yeah, now I understand it. Thanks!

bacv (approved):

danielSanchezQ (approved):

PR #1649, opened 2025-09-05
Tl/poq

thomaslavaur:

Implements PoQ types as well as prove and verify methods. Proving is a wrapper on top of rapidsnark. Verification uses our own groth16 wrapper too.

Yes

@thomaslavaur 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


zk/proofs/poq/src/chain_inputs.rs
```diff
@@ -0,0 +1,202 @@
...
+/// where computed in python which uses a signed representation of the integer.
+/// `BigInt` type (signed), parsed the hex values correctly. But `BigUint` does
+/// not for the motives just described. Using the `str` representation allows
+/// use to choose the unsigned type.
+static T1_CONSTANT: LazyLock<BigUint> = LazyLock::new(|| {
+    // t1 constant:
+    // -0x104bfd09ebdd0a57772289d0973489b62662a4dc6f09da8b4af3c5cfb1dcdd
+    BigUint::from_str("28794005923809446652337194229268641024881242442862297438215833784455126237")
+        .expect("Constant should parse")
+});
```
danielSanchezQ: This constants are from the pol. If the poq needs of pol it would be better to export them in pol and use them here.

zk/proofs/poq/src/chain_inputs.rs
```diff
@@ -0,0 +1,202 @@
...
+        let session = BigUint::from(session);
+        if session > *P {
+            return Err(PoQInputsFromDataError::SessionGreaterThanP);
+        }
+        let total_stake = BigUint::from(total_stake);
+
+        let lottery_0 = T0_CONSTANT.deref().div(total_stake.clone());
+        let lottery_1 = P
+            .checked_sub(&T1_CONSTANT.deref().div(total_stake.pow(2)))
+            .expect("(T1 / (S^2)) must be less than P");
```
danielSanchezQ: We should have a method in the pol for this and export it. Then use it here. We do not want to deviate in just a single crate and break all the process.

zk/proofs/poq/src/chain_inputs.rs
```diff
@@ -0,0 +1,202 @@
...
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn constants() {
```
danielSanchezQ: this can be removed after refactoring.

zk/proofs/poq/src/inputs.rs
```diff
@@ -0,0 +1,154 @@
...
+    pub blend: PoQBlendInputs,
+    pub wallet: PoQWalletInputs,
+}
+
+impl PoQWitnessInputs {
+
+    pub fn from_leader_data(
+        chain: PoQChainInputs,
+        common: PoQCommonInputs,
+        wallet: PoQWalletInputs,
```
danielSanchezQ: Its better to use the `Data` inputs instead, then call into/try_into inside this function to the proper needed types. Its better ux. 
I changed it in pol after you branched probably, so you couldn't relate before.

thomaslavaur: Changed

zk/proofs/poq/src/chain_inputs.rs
thomaslavaur: changed

zk/proofs/poq/src/chain_inputs.rs
thomaslavaur: changed

zk/proofs/poq/src/chain_inputs.rs
thomaslavaur: removed

zk/proofs/poq/dependencies_graph.dot
danielSanchezQ, line 1: Sorry, I didn't explain it well. This should be run from the workspace, as it is for the whole project. This file should be deleted.

zk/proofs/poq/src/lib.rs
```diff
@@ -0,0 +1,342 @@
...
+}
+
+///
+/// This function verifies a proof against a set of public inputs.
+///
+/// # Arguments
+///
+/// - `proof`: A reference to the proof (`PoLProof`) that needs verification.
+/// - `public_inputs`: A reference to `PolVerifierInput`, which contains the
+///   public inputs against which the proof is verified.
```
danielSanchezQ: There are mentions to pol in the documentation that should be poq

zk/proofs/poq/src/lib.rs
```diff
@@ -0,0 +1,342 @@
...
+/// This function generates a proof for the given set of inputs.
+///
+/// # Arguments
+/// - `inputs`: A reference to `PolWitnessInputs`, which contains the necessary
+///   data to generate the witness and construct the proof.
+///
+/// # Returns
+/// - `Ok((PoLProof, PolVerifierInput))`: On success, returns a tuple containing
+///   the generated proof (`PoLProof`) and the corresponding public inputs
+///   (`PolVerifierInput`).
```
danielSanchezQ: Check this pol->poq too

zk/proofs/pol/src/lib.rs
```diff
@@ -121,6 +125,63 @@ pub fn verify(proof: &PoLProof, public_inputs: &PolVerifierInput) -> Result<bool
...
+/// where computed in python which uses a signed representation of the integer.
+/// `BigInt` type (signed), parsed the hex values correctly. But `BigUint` does
+/// not for the motives just described. Using the `str` representation allows
+/// use to choose the unsigned type.
+pub static P: LazyLock<BigUint> = LazyLock::new(|| {
+    BigUint::from_str(
+        "21888242871839275222246405745257275088548364400416034343698204186575808495617",
+    )
+        .expect("P constant should parse")
+});
```
danielSanchezQ: Uhm, why are this changes as green? This should not have changed from the code in master 🤔

danielSanchezQ: Ah...I see, you should re-export this from chain-inputs instead. Or maybe move them to some lottey.rs module instead, then re-export here.

zk/proofs/poq/dependencies_graph.dot
thomaslavaur: changed

zk/proofs/poq/src/lib.rs
thomaslavaur: replaced

zk/proofs/poq/src/lib.rs
thomaslavaur: replaced

zk/proofs/pol/src/lib.rs
thomaslavaur: But I need to export P and the lottery function for use in Poq (P is used to check errors)

thomaslavaur: can I move them in a lottery.rs file and just export the function and constant like a .h in C?

danielSanchezQ: Yes, that sounds good. To export you just to use `pub use lottery::{method, CONST};` in the lib rs file.

danielSanchezQ (commented):
Its missing adding the zkey path so tests can be green. Otherwise is good to go.

danielSanchezQ (approved):
Once all tests and lints are 🟢 this is good to be merged.

zk/proofs/pol/src/chain_inputs.rs
```diff
@@ -201,33 +145,3 @@ impl TryFrom<PolChainInputsData> for PolChainInputs {
...
-        );
-        let lottery_1 = P
-            .checked_sub(&T1_CONSTANT.deref().div(total_stake.pow(2)))
-            .expect("(T1 / (S^2)) must be less than P");
-        assert_eq!(
-            lottery_1.to_string(),
-            "21888242870687514985294027879163787319377618759420784645983712289047175144239"
-        );
-    }
-}
```
danielSanchezQ, line 233: This tests are missing now @thomaslavaur

danielSanchezQ (changes requested):

danielSanchezQ (approved):

Cargo.toml
```diff
@@ -103,6 +104,7 @@ nomos-tracing            = { default-features = false, path = "./nomos-tracing"
 nomos-tracing-service    = { default-features = false, path = "./nomos-services/tracing" }
 nomos-utils              = { default-features = false, path = "./nomos-utils" }
 nomos_proof_statements   = { default-features = false, path = "./nomos-core/proof_statements" }
+pol                      = { default-features = false, path = "zk/proofs/pol" }
```
ntn-x2, line 107: There should also be a `poq` dependency here, not just a `pol` (which was probably forgotten in a previous PR)

zk/proofs/poq/src/lib.rs
```diff
@@ -0,0 +1,327 @@
...
+/// - `Ok(true)`: If the proof is successfully verified against the public
+///   inputs.
+/// - `Ok(false)`: If the proof is invalid when compared with the public inputs.
+/// - `Err`: If an error occurs during the verification process.
+///
+/// # Errors
+///
+/// - Returns an error if there is an issue with the verification key or the
+///   underlying verification process fails.
+pub fn verify(proof: &PoQProof, public_inputs: &PoQVerifierInput) -> Result<bool, impl Error> {
```
ntn-x2, line 95: If I am the receiver of a Blend message, I want to create my own set of inputs from my context, and then I need to add to that the key nullifier that is part of the proof. How would that fit here? How can I create my own `PoQVerifierInput` instance to verify a PoQ embedded in a Blend message?

ntn-x2: The way I was thinking of it, is to define the PoQ as the concatenation of (key nullifier, groth16 proof), which is then embedded in a Blend private header. The receiver will generate its own verifier inputs, add the key nullifier, and verify the proof like that. For this to work, the `PoQVerifierInput` type must expose a constructor or declare its fields as public (preferred).

ntn-x2: If this is a lower level library, which would also be ok, it needs to at least make it possible to create a `PoQVerifierInput` instance.

danielSanchezQ: > The way I was thinking of it, is to define the PoQ as the concatenation of (key nullifier, groth16 proof), which is then embedded in a Blend private header. The receiver will generate its own verifier inputs, add the key nullifier, and verify the proof like that. For this to work, the `PoQVerifierInput` type must expose a constructor or declare its fields as public (preferred).

Yes, they should be public. But you can do that as part of the integration too. This is not set in stone so we can add/modify whats needed as long as it keeps consistent (Like what you are commenting).

ntn-x2: If by "you can do that as part of the integration too" you mean we can modify this code to fit our use case, yes we can do that. Else we need to make the fields public now (which I think it's ok since they are just verifier inputs), and then we can build this object in our own way by fetching info from the verifier context.

danielSanchezQ: > If by "you can do that as part of the integration too" you mean we can modify this code to fit our use case, yes we can do that. Else we need to make the fields public now (which I think it's ok since they are just verifier inputs), and then we can build this object in our own way by fetching info from the verifier context.

Yeah, I mean we can modify this ofc.

thomaslavaur: Should I change something?

Cargo.toml
thomaslavaur: done

zk/proofs/poq/src/lib.rs
ntn-x2: Yeah I think marking all fields of the `PoQVerifierInput` struct as `pub` should be enough.

thomaslavaur: Done

PR #1650, opened 2025-09-05
feat(chain): block (re)construction

andrussal:

Reconstructs block from Proposal. When receiving Proposal from network(which contains only transaction references), we need to retrieve transactions from mempool and construct full block. [Spec](https://www.notion.so/nomos-tech/Block-Construction-Validation-and-Execution-Specification-215261aa09df8164b5a9f0783c0e99b9#215261aa09df8162a641f53e06ed3236.)

It also includes basic bock construction and validation changes.

---
**DON'T MEGRE TO MASTER UNTIL FULL FLOW WITH PROPOSALS IS WORKING**

Yes

@andrussal 

Yes

Yes

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-core/chain-defs/src/utils/merkle.rs
+++ b/nomos-core/chain-defs/src/utils/merkle.rs
@@ -32,6 +32,33 @@ pub fn node(left: impl AsRef<[u8]>, right: impl AsRef<[u8]>) -> [u8; 32] {
     hasher.finalize().into()
 }
 
+/// Calculates a 32-byte Merkle root for the given elements, with deterministic
+/// padding.
+///
+/// Padding rules:
+/// - Target size = `max(elements.len()`, `pad_to.unwrap_or(elements.len())`, 1)
+/// - Then rounded up to the next power of two
+/// - Missing leaves are filled using the default value of `T`
+/// - Special case: single leaf (`target_size` = 1) returns leaf hash directly
+///
+/// Examples:
+/// - 0 elements: 1 leaf → `leaf(default_element)`
+/// - 1 element: 1 leaf → leaf(element)
+/// - 2 elements: 2 leaves → node(leaf1, leaf2)
+/// - 3 elements: 4 leaves (padded) → full binary tree structure
+///
+/// Parameters:
+/// - `elements`: Items included in the tree; each converts to a 32-byte leaf.
+/// - `pad_to`: Optional minimum leaf count before rounding up to a power of
+///   two.
+///
+/// Returns:
+/// - The Merkle root as a 32-byte array.
+///
+/// Notes:
+/// - Using a fixed `pad_to` keeps the tree height consistent across varying
+///   input sizes.
+/// - Powers of two that are already satisfied don't trigger padding.
 pub fn calculate_merkle_root<T>(elements: &[T], pad_to: Option<usize>) -> [u8; 32]
 where
     T: Into<[u8; 32]> + Default + Clone,
@@ -47,10 +74,8 @@ where
         .max(1)
         .next_power_of_two();
 
-    if leaves.len() < target_size {
-        let zero_leaf = leaf(&T::default().into());
-        leaves.resize(target_size, zero_leaf);
-    }
+    let zero_leaf = leaf(&T::default().into());
+    leaves.resize(target_size, zero_leaf);
 
     while leaves.len() > 1 {
         leaves = leaves
@@ -115,6 +140,29 @@ mod tests {
         let leaf1 = leaf(&bytes1);
         let leaf2 = leaf(&bytes2);
         let expected = node(leaf1, leaf2);
+
+        assert_eq!(result, expected);
+    }
+
+    #[test]
+    fn test_root_single_element() {
+        let elements = vec![TxHash::from(Fr::from(42u64))];
+        let result = calculate_merkle_root(&elements, None);
+
+        let bytes: [u8; 32] = elements[0].into();
+        let expected = leaf(&bytes);
+
+        assert_eq!(result, expected);
+    }
+
+    #[test]
+    fn test_root_empty_elements() {
+        let elements: Vec<TxHash> = vec![];
+        let result = calculate_merkle_root(&elements, None);
+
+        let zero_hash = TxHash::default();
+        let zero_bytes: [u8; 32] = zero_hash.into();
+        let expected = leaf(&zero_bytes);
 
         assert_eq!(result, expected);
     }


nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -22,6 +25,10 @@ pub struct References {
 }
 
 /// A block
+#[expect(
+    clippy::unsafe_derive_deserialize,
+    reason = "Temporary dummy signature"
```
zeegomo: is the signature supposed to be added to the block?

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -68,21 +75,48 @@ impl<Tx> Block<Tx> {
...
+            .map(|tx| tx.hash().into())
+            .collect();
+
+        let references = References {
+            service_reward: None,
+            mempool_transactions,
+        };
+
+        // SAFETY: Temporary dummy signature
+        let signature = unsafe { std::mem::zeroed() };
```
zeegomo: we could ask for a key in input and sign here

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -68,21 +75,48 @@ impl<Tx> Block<Tx> {
...
+        Tx::Hash: Into<Fr>,
+    {
+        let mempool_transactions: Vec<Fr> = self
+            .transactions
+            .iter()
+            .map(|tx| tx.hash().into())
+            .collect();
+
+        let references = References {
+            service_reward: None,
```
zeegomo: what about service reward?

nomos-core/chain-defs/src/block/wire.rs
```diff
@@ -0,0 +1,102 @@
...
+
+use crate::block::References;
+
+#[derive(serde::Serialize, serde::Deserialize)]
+struct WireData {
+    service_reward: Option<Vec<u8>>,
+    mempool_transactions: Vec<Vec<u8>>,
+}
+
+impl serde::Serialize for References {
```
zeegomo: you could have also used the `serde_fr` module from groth16 and newtype the `Fr` in `WireData`

nomos-services/chain-service/src/lib.rs
```diff
@@ -812,6 +823,7 @@ where
                         }
                     }
 
+                    // TODO: we maybe can keep existing orphan downloading logic but convert Block to Proposal
```
zeegomo: yes

nomos-services/chain-service/src/lib.rs
```diff
@@ -1644,3 +1656,197 @@ where
...
     receiver.await.map_err(|error| Box::new(error) as DynError)
 }
+
+/// Reconstruct a Block from a Proposal by looking up transactions from mempool
+async fn reconstruct_block_from_proposal<Payload, Item, Key>(
+    proposal: Proposal,
+    mempool: OutboundRelay<MempoolMsg<HeaderId, Payload, Item, Key>>,
+) -> Result<Block<Item>, Error>
+where
+    Key: Send + Eq + Hash + From<ZkHash> + Debug,
```
zeegomo: I think we can remove this generic, we already know we're dealing with `TxHash`

nomos-services/chain-service/src/lib.rs
```diff
@@ -1644,3 +1656,197 @@ where
...
+
+    if !missing_transactions.is_empty() {
+        return Err(Error::BlockReconstruction(format!(
+            "Failed to reconstruct block: {missing_transactions:?} transactions not found in mempool",
+        )));
+    }
+
+    let header = proposal.header().clone();
+
+    Ok(Block::new(header, reconstructed_transactions))
```
zeegomo: we need a lot more validation here, but this was not the scope for this PR :), I'd say just put a comment here that we miss validation

tests/src/tests/cryptarchia/bootstrap.rs
```diff
@@ -20,6 +20,7 @@ use tests::{
 
 #[tokio::test]
 #[serial]
+#[ignore = "Re enable after new block format is fully integrated"]
```
zeegomo, line 22: what's the issue with this?

nomos-services/chain-service/src/lib.rs
```diff
@@ -1644,3 +1656,197 @@ where
...
+async fn reconstruct_block_from_proposal<Payload, Item, Key>(
+    proposal: Proposal,
+    mempool: OutboundRelay<MempoolMsg<HeaderId, Payload, Item, Key>>,
+) -> Result<Block<Item>, Error>
+where
+    Key: Send + Eq + Hash + From<ZkHash> + Debug,
+    Payload: Send,
+    Item: Clone + Transaction<Hash = Key>,
+{
+    let tx_map: HashMap<Key, Item> = get_mempool_contents(mempool)
```
zeegomo: do we dump the full mempool here to get a few items? seems a bit inefficient. One example could be to add a custom op to review txs by hash

zeegomo (changes requested):
I think we should keep this in a separate branch until we can re-enable tests since those are critical and we need to be sure everything is working

nomos-core/chain-defs/src/block/mod.rs
andrussal: > we could ask for a key in input and sign here

I added signing but currently just using dummy key. If I understand correctly, we should derive the signing key? Is this something I could add?

nomos-core/chain-defs/src/block/mod.rs
andrussal: Added reward too, I was reluctant to change Block structure but it seemed neccessesary now.

nomos-services/chain-service/src/lib.rs
andrussal: It seems that we can actually use Block directly? I didn't do conversion here.

nomos-services/chain-service/src/lib.rs
andrussal: I was thinking that most validation is part of `process_block_and_update_state'. I added some basic validation now including txts count as you suggested.

tests/src/tests/cryptarchia/bootstrap.rs
andrussal: This also expects some blocks to be created before IBD runs.

nomos-services/chain-service/src/lib.rs
andrussal: Great point, this was way too basic. I added more efficient lookup.

nomos-core/chain-defs/src/block/mod.rs
andrussal: Now I added signature to block so we can construct Proposal and also store it. Please let me know if this is correct.

nomos-services/chain-service/src/lib.rs
andrussal: Do we need to handle also situation where there are multiple proposals for slot? Current system removes transactions from mempool after successfully adding block to local tree. Now if another proposal/block arrives, they may require the same transactions. It seems like we should be able to handle this? Perhaps having some mempool snapshot that is later removed? 

---
Followup - I think I need but I'll handle this in a separate PR not to bloat this PR.

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -18,13 +34,15 @@ pub struct Proposal {
...
+    pub mempool_transactions: Vec<Fr>, // MAX_TRANSACTIONS - len(service_reward)
 }
 
 /// A block
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug)]
 pub struct Block<Tx> {
     header: Header,
+    signature: ed25519_dalek::Signature,
+    service_reward: Option<Fr>,
```
zeegomo: we should probably resolve this to a Tx anyway

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -68,6 +93,65 @@ impl<Tx> Block<Tx> {
...
+
+        let leader_public_key = self.header.leader_proof().leader_key();
+        let header_bytes = <Header as SerdeOp>::serialize(&self.header)?;
+
+        leader_public_key
+            .verify(&header_bytes, &self.signature)
+            .map_err(|e| Error::Signing(format!("Invalid signature: {e}")))?;
+
+        Ok(())
+    }
```
zeegomo: I think we can make it so it's impossible to build an invalid block instead of requiring validation

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -68,6 +93,65 @@ impl<Tx> Block<Tx> {
...
+
+    pub fn to_proposal(&self) -> Result<Proposal, Error>
+    where
+        Tx: Transaction,
+        Tx::Hash: Into<Fr>,
+    {
+        let mempool_transactions: Vec<Fr> = self
+            .transactions
+            .iter()
+            .map(|tx| tx.hash().into())
```
zeegomo: if reference is not a list of `TxHash` we should probably update it

nomos-core/chain-defs/src/header/mod.rs
```diff
@@ -20,23 +22,64 @@ pub struct ContentId([u8; 32]);
 #[derive(Clone, Debug, Eq, PartialEq, Copy)]
 pub struct Nonce([u8; 32]);
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Version {
+    pub proposal_version: u8,
+    pub cryptarchia_version: u8,
```
zeegomo: where are these coming from?

nomos-services/chain/chain-service/src/sync/block_provider.rs
```diff
@@ -815,11 +818,16 @@ mod tests {
...
-                vec![],
-            )
+        ) -> Option<Block<SignedMantleTx>> {
+            let dummy_signing_key = ed25519_dalek::SigningKey::from_bytes(&[1u8; 32]);
+            let header = Header::new(prev_header, [0; 32].into(), slot, self.proof.clone());
+
+            let Ok(signature) = header.sign(&dummy_signing_key) else {
+                error!("Failed to sign header in block provider");
+                return None;
+            };
```
zeegomo: any reason this should fail in tests?

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -844,32 +864,49 @@ where
...
-                            blob_validation.as_ref(),
-                            &storage_blocks_to_remove,
-                            &relays,
-                            &self.new_block_subscription_sender,
-                            &self.lib_subscription_sender,
-                            &self.service_resources_handle.state_updater
-                        ).await {
-                            Ok((new_cryptarchia, new_storage_blocks_to_remove)) => {
-                                cryptarchia = new_cryptarchia;
-                                storage_blocks_to_remove = new_storage_blocks_to_remove;
```
zeegomo, line 824: I know it's pre-existing, but can't we merge these two branches?

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -1656,3 +1725,77 @@ async fn broadcast_finalized_block(
...
+    if reconstructed_transactions.len() != needed_hashes.len() {
+        let missing_transactions: Vec<TxHash> = needed_hashes
+            .into_iter()
+            .filter(|hash| {
+                !reconstructed_transactions
+                    .iter()
+                    .any(|tx| tx.hash() == *hash)
+            })
+            .collect();
+
```
zeegomo: maybe we should return the unmatched ones in the error variant?

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -1656,3 +1725,77 @@ async fn broadcast_finalized_block(
...
+        return Err(Error::InvalidBlock(format!(
+            "Failed to reconstruct block: {missing_transactions:?} transactions not found in mempool",
+        )));
+    }
+
+    let header = proposal.header().clone();
+    let service_reward = proposal.references().service_reward;
+    let signature = *proposal.signature();
```
zeegomo, line 1627: this is where we should probably recover the reward tx imo, but ofc not something to add now because we need some additional work on sdp

nomos-core/chain-defs/src/block/mod.rs
andrussal: I am not sure if this is what you meant but I added a Builder to construct blocks which verifies some stuff.

nomos-core/chain-defs/src/block/mod.rs
andrussal: Changed references from `Fr` to `TxHash`

nomos-core/chain-defs/src/header/mod.rs
andrussal: > where are these coming from?

It is this stuff in spec - https://www.notion.so/Block-Construction-Validation-and-Execution-Specification-MIGRATED-1a48f96fb65c8078b625dbd3a101a7bc?source=copy_link#2088f96fb65c80caa105d6d47b8c586c

nomos-services/chain/chain-service/src/lib.rs
andrussal: > this is where we should probably recover the reward tx imo, but ofc not something to add now because we need some additional work on sdp

Just added a TODO comment for now.

nomos-services/chain/chain-service/src/lib.rs
andrussal: > maybe we should return the unmatched ones in the error variant?

I made mempool to return a type that has two distinct sets of hashes: found and not found. So mempool API treats this as normal if it can't find some hashes but chain service ofc treats this as an error.

nomos-core/chain-defs/src/header/mod.rs
zeegomo: as the title says, that is migrated and out of date, please see https://www.notion.so/nomos-tech/Block-Construction-Validation-and-Execution-Specification-215261aa09df8164b5a9f0783c0e99b9 for the correct version

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -23,38 +27,188 @@ pub enum Error {
...
-    pub references: References,
+    pub transactions: Transactions<Tx>,
     pub signature: ed25519_dalek::Signature,
 }
 
 #[derive(Clone, Debug)]
-pub struct References {
-    pub service_reward: Option<Fr>,
-    pub mempool_transactions: Vec<Fr>, // MAX_TRANSACTIONS - len(service_reward)
+pub struct Transactions<Tx>
```
zeegomo: why the rename? I'd like to keep naming as in the spec

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -23,38 +27,188 @@ pub enum Error {
...
 }
 
 /// A block
 #[derive(Clone, Debug)]
 pub struct Block<Tx> {
     header: Header,
     signature: ed25519_dalek::Signature,
-    service_reward: Option<Fr>,
+    service_reward: Option<Tx>,
+    transactions: Vec<Tx>,
```
zeegomo, line 55: any reason to keep these two separated?

nomos-services/mempool/src/lib.rs
```diff
@@ -10,6 +10,40 @@ use backend::{MempoolError, Status};
 use tokio::sync::oneshot::Sender;
 pub use tx::{service::TxMempoolService, settings::TxMempoolSettings};
 
+/// Response for `GetTransactionsByHashes` request
+#[derive(Debug, Clone)]
+pub struct TransactionsByHashesResponse<Item, Key> {
+    /// Transactions that were found in the mempool
+    pub found_transactions: Vec<Item>,
```
zeegomo: nit: I think we can remove `_transaction` and `_hashes`, it's clear from the signature what they represent

nomos-services/tx-service/src/lib.rs
```diff
@@ -10,6 +10,40 @@ use backend::{MempoolError, Status};
...
+
+    #[must_use]
+    pub const fn found_count(&self) -> usize {
+        self.found_transactions.len()
+    }
+
+    #[must_use]
+    pub const fn not_found_count(&self) -> usize {
+        self.not_found_hashes.len()
+    }
```
zeegomo, line 49: are these used somewhere except testing?

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -23,38 +27,188 @@ pub enum Error {
...
+    signing_key: ed25519_dalek::SigningKey,
+}
+
+#[derive(Debug)]
+pub struct BuilderWithSignature<Tx> {
+    header: Header,
     transactions: Vec<Tx>,
+    service_reward: Option<Tx>,
+    signature: ed25519_dalek::Signature,
+}
```
zeegomo: what's the rationale behind this API? Unless there's a real use-case I'd like to keep things simple and on-point, we can always add later if needed

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -1,42 +1,214 @@
...
+        if transactions.len() > MAX_TRANSACTIONS {
+            return Err(Error::Signing(format!(
+                "Too many transactions (max {MAX_TRANSACTIONS})"
+            )));
+        }
+
+        self.transactions = transactions;
+
+        Ok(self)
+    }
```
zeegomo: header will depends on the included txs at least for the tx hash, how do you plan this API to be used? Or is it temporary?

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -1,42 +1,214 @@
...
+            .map_err(|e| Error::Signing(format!("Signature verification failed: {e}")))?;
+
+        Ok(Block::new_unchecked(
+            self.header,
+            self.transactions,
+            self.service_reward,
+            signature,
+        ))
+    }
+}
```
zeegomo: why is this separated from `signing_key`, what cases do you envision where somebody would supply the signing key but not build the block yet, and why?

nomos-core/chain-defs/src/block/mod.rs
zeegomo: Is this ever called by the builder?

nomos-core/chain-defs/src/block/wire.rs
```diff
@@ -0,0 +1,235 @@
...
+
+use crate::{
+    block::{Block, Transactions},
+    header::Header,
+    mantle::{AuthenticatedMantleTx, TxHash},
+};
+
+#[derive(serde::Serialize, serde::Deserialize)]
+#[serde(transparent)]
+struct SerializableFr(#[serde(with = "serde_fr")] Fr);
```
zeegomo: is this actually used?

nomos-core/chain-defs/src/block/mod.rs
andrussal: > Is this ever called by the builder?

This is yes not needed anymore, removed.

nomos-core/chain-defs/src/block/mod.rs
andrussal: > any reason to keep these two separated?

I kept them separate so it is straightforward to create Proposal from Block.

nomos-core/chain-defs/src/block/mod.rs
andrussal: > what's the rationale behind this API? Unless there's a real use-case I'd like to keep things simple and on-point, we can always add later if needed

This API is really not needed. I only added it to integrate construction with validation but basic construction functions("`new`" + some validation) are actually much better than builder. Removed this design altogether.

nomos-core/chain-defs/src/block/mod.rs
andrussal: This is now removed with the rest of this API.

nomos-core/chain-defs/src/block/mod.rs
andrussal: I agree, this builder design was confusing and over-engineered. Removed all this.

nomos-core/chain-defs/src/block/wire.rs
andrussal: It wasn't used, removed now.

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -1,30 +1,53 @@
...
+    Serialisation(#[from] crate::codec::Error),
+    #[error("Signing error: {0}")]
+    Signing(String),
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Proposal {
     pub header: Header,
-    pub references: References,
+    pub transactions: References,
```
zeegomo: nit: probably keep this as references

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -46,12 +74,114 @@ impl Proposal {
...
+        service_reward: Option<&Tx>,
+    ) -> ContentId
+    where
+        Tx: Transaction<Hash = TxHash>,
+    {
+        let mut all_transactions = Vec::new();
+        if let Some(reward) = service_reward {
+            all_transactions.push(reward);
+        }
+        all_transactions.extend(transactions.iter());
```
zeegomo: `service_reward.iter().chain(transactions.iter().collect()`?

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -46,12 +74,114 @@ impl Proposal {
...
+            .iter()
+            .enumerate()
+            .map(|(i, tx)| {
+                if i % 2 == 0 {
+                    merkle::MerkleNode::Left(tx.hash())
+                } else {
+                    merkle::MerkleNode::Right(tx.hash())
+                }
+            })
+            .collect();
```
zeegomo: mmm no? Not sure whether we need to pad the tree (added a comment on the spec), but you can take the  necessary building blocks from https://github.com/logos-co/nomos-pocs/blob/master/emmarin/cl/cl/src/ds/merkle.rs

nomos-core/chain-defs/src/block/wire.rs
```diff
@@ -0,0 +1,195 @@
...
+    block::{Block, References},
+    header::Header,
+    mantle::TxHash,
+};
+
+#[derive(serde::Serialize, serde::Deserialize)]
+struct WireReferences {
+    service_reward: Option<TxHash>,
+    mempool_transactions: Vec<TxHash>,
+}
```
zeegomo: isn't this the same format as `References`?

nomos-core/chain-defs/src/block/wire.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+#[derive(serde::Serialize, serde::Deserialize)]
+struct WireBlock<Tx> {
+    header: Header,
+    signature: ed25519_dalek::Signature,
+    service_reward: Option<Tx>,
+    transactions: Vec<Tx>,
+}
```
zeegomo: same here, can we remove these duplicated structures?

nomos-core/chain-defs/src/header/mod.rs
```diff
@@ -20,23 +22,36 @@ pub struct ContentId([u8; 32]);
 #[derive(Clone, Debug, Eq, PartialEq, Copy)]
 pub struct Nonce([u8; 32]);
 
+pub type Version = u8;
+
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Header {
+    version: Version,
```
zeegomo, line 40: let's use an enum with `repr(u8)` so that we get automatic verification during deserialization

nomos-core/chain-defs/src/header/mod.rs
```diff
@@ -20,23 +22,36 @@ pub struct ContentId([u8; 32]);
...
 impl Header {
+    #[must_use]
+    pub const fn version(&self) -> &Version {
+        &self.version
+    }
+
+    #[must_use]
+    pub const fn is_valid_bedrock_version(&self) -> bool {
+        self.version == BEDROCK_VERSION
+    }
```
zeegomo: this can be removed if we have an enum with a single variant

nomos-core/chain-defs/src/utils/merkle.rs
```diff
@@ -13,3 +18,112 @@ impl<T> MerkleNode<T> {
 }
 
 pub type MerklePath<T> = Vec<MerkleNode<T>>;
+
+pub fn calculate_merkle_root<T>(leaves: &[MerkleNode<T>]) -> Option<Fr>
```
zeegomo: a `MerkleNode` is meant to be for internal nodes, not leaves, we can just ask for `T`s here

nomos-core/chain-defs/src/utils/merkle.rs
```diff
@@ -13,3 +18,112 @@ impl<T> MerkleNode<T> {
...
+where
+    T: Into<Fr> + Copy,
+{
+    if leaves.is_empty() {
+        return None;
+    }
```
zeegomo: what about padding?

nomos-core/chain-defs/src/utils/merkle.rs
```diff
@@ -13,3 +18,112 @@ impl<T> MerkleNode<T> {
...
+            let combined = if chunk.len() == 2 {
+                ZkHasher::digest(&[chunk[0], chunk[1]])
+            } else {
+                chunk[0]
+            };
+            next_level.push(combined);
+        }
+
+        current_level = next_level;
+    }
```
zeegomo: i think we can simplify and assume / check for full binary trees

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -773,10 +796,17 @@ where
...
                             }
+                            Err(Error::Ledger(nomos_ledger::LedgerError::ParentNotFound(parent))
+                                | Error::Consensus(cryptarchia_engine::Error::ParentMissing(parent))) => {
+
+                                orphan_downloader.enqueue_orphan(header_id, cryptarchia.tip(), cryptarchia.lib());
+                                error!(target: LOG_TARGET, "Orphan block with parent {:?} not in ledger state", parent);
+                            }
                             Err(e) => {
-                                error!(target: LOG_TARGET, "Error processing orphan downloader block: {e:?}");
+                                error!(target: LOG_TARGET, "Error processing orphan block: {e:?}");
```
zeegomo: duplicated reporting?

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -773,10 +796,17 @@ where
...
 
+                                orphan_downloader.remove_orphan(&header_id);
                                 info!(counter.consensus_processed_blocks = 1);
                             }
+                            Err(Error::Ledger(nomos_ledger::LedgerError::ParentNotFound(parent))
+                                | Error::Consensus(cryptarchia_engine::Error::ParentMissing(parent))) => {
+
+                                orphan_downloader.enqueue_orphan(header_id, cryptarchia.tip(), cryptarchia.lib());
+                                error!(target: LOG_TARGET, "Orphan block with parent {:?} not in ledger state", parent);
+                            }
```
zeegomo: is this an artifact of the rebase?

nomos-services/mempool/src/tx/service.rs
```diff
@@ -116,6 +116,27 @@ where
...
+        pool: &Pool,
+        hashes: Vec<Pool::Key>,
+    ) -> (Vec<Pool::Item>, Vec<Pool::Key>)
+    where
+        Pool::Item: Clone,
+    {
+        let mut found_transactions = Vec::new();
+        let mut not_found_hashes = Vec::new();
+
+        for hash in hashes {
```
zeegomo: nit: have a look at https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition

nomos-services/tx-service/src/lib.rs
zeegomo: ^

nomos-core/chain-defs/src/block/mod.rs
andrussal: > mmm no? Not sure whether we need to pad the tree (added a comment on the spec), but you can take the necessary building blocks from https://github.com/logos-co/nomos-pocs/blob/master/emmarin/cl/cl/src/ds/merkle.rs

Now I followed this but used `blake2` for hashing because it was used elsewhere and for now added fixed padding.  Added TODO comment for padding as well.

nomos-services/mempool/src/tx/service.rs
andrussal: > nit: have a look at https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition

Thanks for the tip! In this case it seemed better approach to use fold because result type is over `Vec<Item>` and `Vec<Key>`. Please let me know if this is fine.

nomos-services/chain/chain-service/src/lib.rs
andrussal: Actually this was "downloader block" because we have this component called `OrphanBlocksDownloader`.

nomos-core/chain-defs/src/block/wire.rs
andrussal: Now I was able to remove the whole file, thank you for pointing this out!

nomos-core/chain-defs/src/header/mod.rs
andrussal: This is removed, not sure why not showing up here.

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -115,20 +112,16 @@ impl<Tx> Block<Tx> {
     where
         Tx: Transaction<Hash = TxHash>,
     {
-        if !header.is_valid_bedrock_version() {
-            return Err(Error::Signing("Invalid header version".to_owned()));
-        }
-
         if transactions.len() > MAX_TRANSACTIONS {
-            return Err(Error::Signing(format!(
+            return Err(Error::Validation(format!(
```
zeegomo: could also be its own variant `TOO_MANY_TXS`

nomos-core/chain-defs/src/header/mod.rs
```diff
@@ -22,7 +22,18 @@ pub struct ContentId([u8; 32]);
...
-pub type Version = u8;
+#[derive(Clone, Debug, Eq, PartialEq, Copy, Serialize, Deserialize)]
+#[repr(u8)]
+pub enum Version {
+    Bedrock = BEDROCK_VERSION,
+}
+
+impl Version {
+    #[must_use]
+    pub const fn to_le_bytes(self) -> [u8; 1] {
```
zeegomo: nit: not sure endianess makes a lot of sense with just one byte

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -1,30 +1,51 @@
...
 pub type SessionNumber = u64;
 
-/// A block proposal
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Failed to serialize: {0}")]
+    Serialisation(#[from] crate::codec::Error),
+    #[error("Signature error: {0}")]
+    Signature(String),
```
zeegomo: better to bubble up the underlying error

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -39,19 +60,100 @@ impl Proposal {
...
+        if transactions.len() > MAX_TRANSACTIONS {
+            return Err(Error::Validation(format!(
+                "Too many transactions (max {MAX_TRANSACTIONS})"
+            )));
+        }
+
+        let calculated_content_id =
+            Self::calculate_content_id(&transactions, service_reward.as_ref());
+        if header.block_root() != &calculated_content_id {
+            return Err(Error::Validation(
```
zeegomo: same

nomos-core/chain-defs/src/utils/merkle.rs
```diff
@@ -13,3 +15,98 @@ impl<T> MerkleNode<T> {
...
+        let zero_leaf = leaf(&zero_element.into());
+        leaves.resize(pad_to, zero_leaf);
+    }
+
+    let leaves_count = leaves.len();
+
+    assert!(
+        leaves_count >= 2 && leaves_count.is_power_of_two(),
+        "Input must be full binary tree"
+    );
```
zeegomo: nit: we could also automatically pad to the next_po2

zeegomo (commented):
Looking good, just make sure we solved all open comments

nomos-core/chain-defs/src/block/mod.rs
zeegomo: For this, better to always sync with spec. If you find something that is not specified, ask in the relevant spec doc

nomos-services/tx-service/src/lib.rs
andrussal: > are these used somewhere except testing?

I am not sure why here it doesn't show outdated, if I check file directly, these are not there: 
https://github.com/logos-co/nomos/blob/0e927d5965ef4b4f3e0e6053940e13dbd92fe8a3/nomos-services/mempool/src/lib.rs

nomos-core/chain-defs/src/block/mod.rs
andrussal: > For this, better to always sync with spec. If you find something that is not specified, ask in the relevant spec doc

Makes sense! I trusted that existing code uses expected hashing functions but it could have been wrong. (spec also says `blake2`)

nomos-core/chain-defs/src/block/mod.rs
andrussal: > could also be its own variant `TOO_MANY_TXS`

This was also fixed but not sure why not showing up here. But I am gonna add `TOO_MANY_TXS` as you suggested.

nomos-core/chain-defs/src/utils/merkle.rs
andrussal: > nit: we could also automatically pad to the next_po2

Good point. In that case caller just has to make sure that it uses correct padding if mandatory but this function handles it permissively.

nomos-core/chain-defs/src/block/mod.rs
zeegomo: > spec also says blake2

FYI, that's because I asked in a comment there after this thread :)

andrussal: > > spec also says blake2
> 
> FYI, that's because I asked in a comment there after this thread :)

Ah ok, thank you!

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -23,7 +24,11 @@ pub enum Error {
     #[error("Failed to serialize: {0}")]
     Serialisation(#[from] crate::codec::Error),
     #[error("Signature error: {0}")]
-    Signature(String),
+    Signature(Arc<ed25519_dalek::SignatureError>),
```
zeegomo: I think a `Box` should be good enough?

nomos-core/chain-defs/src/header/mod.rs
```diff
@@ -30,7 +30,7 @@ pub enum Version {
 
 impl Version {
     #[must_use]
-    pub const fn to_le_bytes(self) -> [u8; 1] {
+    pub const fn to_bytes(self) -> [u8; 1] {
```
zeegomo: `as_byte/u8`?

nomos-core/chain-defs/src/utils/merkle.rs
```diff
@@ -42,22 +42,21 @@ where
...
 
-    if leaves.len() < pad_to {
-        let zero_element = T::default();
-        let zero_leaf = leaf(&zero_element.into());
-        leaves.resize(pad_to, zero_leaf);
-    }
+    let target_size = leaves.len().max(pad_to).max(1).next_power_of_two();
 
-    let leaves_count = leaves.len();
+    if leaves.len() < target_size {
```
zeegomo: can this ever be less if `target_size` is `leaves.len().next_power_of_two()`?

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -39,19 +65,99 @@ impl Proposal {
...
-    #[must_use]
-    pub const fn new(header: Header, transactions: Vec<Tx>) -> Self {
-        Self {
+    pub fn create(
+        parent_block: HeaderId,
+        slot: Slot,
+        proof_of_leadership: Groth16LeaderProof,
+        transactions: Vec<Tx>,
+        service_reward: Option<Tx>,
+        signing_key: &ed25519_dalek::SigningKey,
```
zeegomo, line 87: nit: could check this is the one used in proof of leadership

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -39,19 +65,99 @@ impl Proposal {
...
+    where
+        Tx: Transaction<Hash = TxHash>,
+    {
+        let tx_hashes: Vec<TxHash> = service_reward
+            .into_iter()
+            .chain(transactions.iter())
+            .map(Transaction::hash)
+            .collect();
+
+        let root_hash = merkle::calculate_merkle_root(&tx_hashes, MAX_TRANSACTIONS);
```
zeegomo: we don't need to pad to `MAX_TRANSACTIONS`

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -85,3 +222,137 @@ impl<Tx: Clone + Eq + Serialize + DeserializeOwned> TryFrom<Block<Tx>> for Bytes
...
+            create_transactions(MAX_TRANSACTIONS + 1),
+            service_reward,
+            &signing_key,
+        );
+
+        assert!(invalid_block_result.is_err());
+        let error_msg = invalid_block_result.unwrap_err().to_string();
+
+        assert!(error_msg.contains("Too many transactions"));
+        assert!(error_msg.contains(&MAX_TRANSACTIONS.to_string()));
```
zeegomo: nit: you can check on the err variant now

nomos-core/chain-defs/src/utils/merkle.rs
```diff
@@ -13,3 +15,90 @@ impl<T> MerkleNode<T> {
...
+
+pub fn node(left: impl AsRef<[u8]>, right: impl AsRef<[u8]>) -> [u8; 32] {
+    let mut hasher = Hasher::new();
+    hasher.update(b"NOMOS_MERKLE_NODE");
+    hasher.update(left.as_ref());
+    hasher.update(right.as_ref());
+    hasher.finalize().into()
+}
+
+pub fn calculate_merkle_root<T>(elements: &[T], pad_to: usize) -> [u8; 32]
```
zeegomo: padding should be optional

nomos-core/chain-defs/src/utils/merkle.rs
andrussal: I think it can if for example `pad_to` is larger? But maybe you mean something else, please let me know.

nomos-core/chain-defs/src/utils/merkle.rs
```diff
@@ -32,27 +32,7 @@ pub fn node(left: impl AsRef<[u8]>, right: impl AsRef<[u8]>) -> [u8; 32] {
...
-///
-/// Returns:
-/// - The Merkle root as a 32-byte array.
-///
-/// Notes:
-/// - Using a fixed `pad_to` keeps the tree height consistent across varying
-///   input sizes.
-/// - For an empty `elements`, the root corresponds to a single default leaf.
-pub fn calculate_merkle_root<T>(elements: &[T], pad_to: usize) -> [u8; 32]
+pub fn calculate_merkle_root<T>(elements: &[T], pad_to: Option<usize>) -> [u8; 32]
```
zeegomo, line 62: where did the comment go?

nomos-core/chain-defs/src/utils/merkle.rs
zeegomo: I mean `target_size` is always at least `leaves.len()` (+ maybe a pad and rounded up)

nomos-core/chain-defs/src/utils/merkle.rs
```diff
@@ -121,4 +104,18 @@ mod tests {
 
         assert_eq!(result, expected);
     }
+
+    #[test]
+    fn test_root_without_padding() {
+        let elements = vec![TxHash::from(Fr::from(1u64)), TxHash::from(Fr::from(2u64))];
+        let result = calculate_merkle_root(&elements, None);
```
zeegomo, line 136: try that with 1 element it should still pad to the next po2, also try with empty

nomos-core/chain-defs/src/utils/merkle.rs
andrussal: > where did the comment go?

I removed it because it seemed simpler just to read the function, and keeping it up to date was was a bit tricky:) But I'll add it back.

PR #1651, opened 2025-09-05
chore(mempool): Remove da mempool

bacv:

Outdated DA mempool is removed as all blob related operations should be sent as transactions containing blob operations.
Delayed triggering for sampling after the transaction with blob op was observed was transferred to transactions mempool and tests re enabled.

Yes

@bacv @danielSanchezQ @zeegomo 

Yes

No

## Checklist

* [x] 2. Description added.


danielSanchezQ (approved):
It served us well for a long time. But its time to let let go...

pradovic (approved):

nomos-services/mempool/src/processor/tx.rs
```diff
@@ -47,41 +47,27 @@ where
...
-                } else {
-                    None
-                }
-            })
-            .collect();
+    async fn process(
+        &self,
+        trigger_sampling_tasks: &mut FuturesUnordered<SamplingFutureResult<Self::Error>>,
+        trigger_sampling_delay: Duration,
+        payload: &Self::Payload,
```
zeegomo: not sure I get the intended use for this. Later the code just pushes futures in here, could we return those futures or even process them inside this function? Why pushing to a different task?

zeegomo (commented):
Good job 🚀 , just left one comment

nomos-services/mempool/src/processor/tx.rs
bacv: The futures have a delay (`tokio::time::sleep`), we can't execute them inside the this function if we don't want to block the `run` method. We could return futures from this method, I'll make a change.

bacv: Updated in https://github.com/logos-co/nomos/pull/1651/commits/e68bed02040ff812905ec81dcbd376058e23b01d

nomos-services/data-availability/network/src/backends/libp2p/executor.rs
```diff
@@ -358,3 +360,11 @@ async fn handle_executor_dispersal_events_stream(
         }
     }
 }
+
+async fn handle_stream_event<Event>(
+    event: Result<Event, BroadcastStreamRecvError>,
+) -> Option<Event> {
+    event
+        .inspect_err(|err| tracing::error!("Stream event broadcast error: {err:?}"))
```
bacv, line 368: @pradovic your suggestion from https://github.com/logos-co/nomos/pull/1450/files#r2212349438 applied here.

PR #1652, opened 2025-09-06
feat(blend): pass SessionEvent stream to core backend

youngjoon-lee:

This is the `2/N`-th PR for https://github.com/logos-co/nomos/issues/1533.

Previously, when initializing the core service, we created the backend with a stream built from the static membership specified in the settings.

This PR:
- Changes the flow so that the backend is created only after subscribing to a stream of `SessionEvent`s provided by the membership adapter (though we're yet using the static membership instead of the stream from the membership adapter).
- Modifies the backend to notify new sessions to the underlying swarm.

Next PRs are coming soon. (Checking new sessions in the service level)

yes

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/core/mod.rs
+++ b/nomos-services/blend/src/core/mod.rs
@@ -12,14 +12,16 @@ use std::{
 
 use async_trait::async_trait;
 use backends::BlendBackend;
+use fork_stream::StreamExt as _;
 use futures::{future::join_all, StreamExt as _};
 use network::NetworkAdapter;
 use nomos_blend_message::{crypto::random_sized_bytes, encap::DecapsulationOutput, PayloadType};
 use nomos_blend_network::EncapsulatedMessageWithValidatedPublicHeader;
 use nomos_blend_scheduling::{
+    membership::Membership,
     message_blend::crypto::CryptographicProcessor,
     message_scheduler::{round_info::RoundInfo, MessageScheduler},
-    session::SessionEventStream,
+    session::{SessionEvent, SessionEventStream},
     UninitializedMessageScheduler,
 };
 use nomos_core::wire;
@@ -35,8 +37,7 @@ use overwatch::{
 use rand::{seq::SliceRandom as _, RngCore, SeedableRng as _};
 use serde::{Deserialize, Serialize};
 use services_utils::wait_until_services_are_ready;
-use tokio::time::interval;
-use tokio_stream::wrappers::IntervalStream;
+use tokio::time::timeout;
 
 use crate::{
     core::settings::BlendConfig,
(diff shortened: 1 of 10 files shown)


nomos-services/blend/src/core/mod.rs
```diff
@@ -157,27 +144,48 @@ where
...
-            blend_config.session_info_stream(session_stream),
+            blend_config.session_info_stream(session_stream.clone().boxed()),
             blend_config.scheduler_settings(),
             rng.clone(),
         )
         .wait_next_session_start()
         .await;
 
+        // Read the initial membership, expecting it to be yielded immediately.
+        // We use 1s timeout to tolerate small delays.
```
ntn-x2, line 170: We have a `wait_next_session_start()` called in the previous line. I think that would be stuck for a long time, and anyway it would consume the first stream element. What I would suggest (I could also work on it if you prefer) is to not wait for the next session to start anymore, but rather fetch the membership info immediately, and call the message scheduler with an initial membership info + the membership stream for future updates. What do you think?

nomos-services/blend/src/core/mod.rs
```diff
@@ -157,27 +144,48 @@ where
...
+        // TODO: Refactor this to a separate struct.
+        let SessionEvent::NewSession(membership) =
+            timeout(Duration::from_secs(1), session_stream.next())
```
ntn-x2, line 173: I remember we use the same timeout also for the edge service, right? Maybe it would make sense to factor this out into its own utility function/type that we can use across services.

nomos-services/blend/src/core/mod.rs
```diff
@@ -350,3 +358,14 @@ async fn handle_release_round<NodeId, Rng, Backend, NetAdapter, RuntimeServiceId
     join_all(processed_messages_relay_futures).await;
     tracing::debug!(target: LOG_TARGET, "Sent out {total_message_count} processed and/or cover messages at this release window.");
 }
+
+// TODO: Remove this and use the membership service stream instead: https://github.com/logos-co/nomos/issues/1532
+fn initial_membership<BackendSettings, NodeId>(
```
ntn-x2, line 363: Why is this not a method on the `Settings` object instead?

nomos-services/blend/src/core/mod.rs
ntn-x2: Nevermind, this is done in a subsequent PR.

nomos-services/blend/src/core/mod.rs
ntn-x2: This is also done in a subsequent PR.

ntn-x2 (approved):
I'all approve with a minor outstanding nit since everything else has been tackled in a stacked PR.

PR #1653, opened 2025-09-06
refactor(blend): embed the first session handling into the stream impl

youngjoon-lee:

This is the `3/N`-th PR for https://github.com/logos-co/nomos/issues/1533.

All the proxy, core, and edge services are expecting the first session to be yielded (almost) immediately from the membership stream (provided by the membership adapter).

Previously, I've copied that logic to all the services. But, this PR embeds it into the session stream implementation, so that it can be reused anywhere. It's also safer.

Yes

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/scheduling/src/session.rs
+++ b/nomos-blend/scheduling/src/session.rs
@@ -8,6 +8,50 @@ use std::{
 use futures::StreamExt as _;
 use tokio::time::{sleep, timeout, Sleep};
 
+/// A staging type that initializes a [`SessionEventStream`] by consuming
+/// the first [`Session`] from the underlying stream, expected to be yielded
+/// within a short timeout.
+pub struct UninitializedSessionEventStream<Stream> {
+    stream: FirstReadyStream<Stream>,
+    transition_period: Duration,
+}
+
+impl<Stream> UninitializedSessionEventStream<Stream> {
+    #[must_use]
+    pub const fn new(
+        session_stream: Stream,
+        first_ready_timeout: Duration,
+        transition_period: Duration,
+    ) -> Self {
+        Self {
+            stream: FirstReadyStream::new(session_stream, first_ready_timeout),
+            transition_period,
+        }
+    }
+}
+
+impl<Stream, Session> UninitializedSessionEventStream<Stream>
+where
+    Stream: futures::Stream<Item = Session> + Unpin,
+{
+    /// Initializes a [`SessionEventStream`] by consuming the first [`Session`]
+    /// from the underlying stream.
+    ///
+    /// It returns the first [`Session`] and the initialized
+    /// [`SessionEventStream`].
+    /// It returns an error if the first session is not yielded within the
+    /// configured timeout.
+    pub async fn await_first_ready(
+        self,
+    ) -> Result<(Session, SessionEventStream<Stream>), FirstReadyStreamError> {
+        let (first_session, remaining_stream) = self.stream.first().await?;
+        Ok((
+            first_session,
+            SessionEventStream::new(remaining_stream, self.transition_period),
+        ))
+    }
+}
+
 #[derive(Clone)]
 pub enum SessionEvent<Session> {
     NewSession(Session),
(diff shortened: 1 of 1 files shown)


nomos-blend/scheduling/src/session.rs
```diff
@@ -78,10 +78,92 @@ impl<Session> Stream for SessionEventStream<Session> {
     }
 }
 
+/// A staging type that initializes a [`SessionEventStream`] by consuming
+/// the first [`Session`] from the underlying stream, expected to be yielded
+/// within a short timeout.
+pub struct UninitializedSessionEventStream<Stream> {
```
ntn-x2: Should we move the definition of this before the definition of the stream type it yields when ready?

nomos-blend/scheduling/src/session.rs
```diff
@@ -78,10 +78,92 @@ impl<Session> Stream for SessionEventStream<Session> {
...
+where
+    Stream: futures::Stream<Item = Session> + Unpin,
+{
+    /// Initializes a [`SessionEventStream`] by consuming the first [`Session`]
+    /// from the underlying stream.
+    ///
+    /// It returns the first [`Session`] and the initialized
+    /// [`SessionEventStream`].
+    ///
+    /// It returns an error if the first session is not yieled within a short
```
ntn-x2:
```suggestion
    /// It returns an error if the first session is not yielded within a short
```

nomos-blend/scheduling/src/session.rs
```diff
@@ -78,10 +78,92 @@ impl<Session> Stream for SessionEventStream<Session> {
...
+pub struct UninitializedSessionEventStream<Stream> {
+    stream: FirstReadyStream<Stream>,
+    transition_period: Duration,
+}
+
+impl<Stream> UninitializedSessionEventStream<Stream> {
+    #[must_use]
+    pub const fn new(session_stream: Stream, transition_period: Duration) -> Self {
+        Self {
+            stream: FirstReadyStream::new(session_stream, Duration::from_secs(1)),
```
ntn-x2: Why not making also the timeout duration configurable? And then perhaps define a crate-wide constant for the duration all Blend services should use?

nomos-blend/scheduling/Cargo.toml
```diff
@@ -17,6 +17,7 @@ nomos-blend-message = { workspace = true }
 nomos-core          = { workspace = true }
 rand                = { features = ["alloc"], workspace = true }
 serde               = { default-features = false, version = "1.0" }
+thiserror           = "1"
```
ntn-x2: Let's disable default features by default.

nomos-services/blend/src/core/mod.rs
```diff
@@ -204,10 +188,12 @@ where
             <RuntimeServiceId as AsServiceId<Self>>::SERVICE_ID
         );
 
+        let mut rng = BlakeRng::from_entropy();
```
ntn-x2: Just for clarity, we should initialize this before the service reports it is ready. What do you think?

nomos-blend/scheduling/src/session.rs
```diff
@@ -78,10 +78,92 @@ impl<Session> Stream for SessionEventStream<Session> {
...
+{
+    /// Initializes a [`SessionEventStream`] by consuming the first [`Session`]
+    /// from the underlying stream.
+    ///
+    /// It returns the first [`Session`] and the initialized
+    /// [`SessionEventStream`].
+    ///
+    /// It returns an error if the first session is not yieled within a short
+    /// time.
+    pub async fn initialize(
```
ntn-x2: I think we could use a better naming here that gives away better what it does. I am thinking something like `expect_ready_before_timeout`, or something like that. `initialize` I feel is not the best match for its logic.

ntn-x2: Or maybe just `expect_ready`.

nomos-blend/scheduling/src/session.rs
youngjoon-lee: Makes sense. Moved to the top: f326b9cd93956afe6f5e9045e8f04c73c7fdea2a.

I initially thought so, but didn't put it before `SessionEventStream` because `SessionEventStream` the main struct of this module. But, I also agree with your point.

nomos-blend/scheduling/src/session.rs
youngjoon-lee: I agree. Done with 7c520f1c129f478429803340b7f2053d38a9ee35

nomos-blend/scheduling/Cargo.toml
youngjoon-lee: Done 5a7a5037e4dfa60ed4e3f72012dff6d0a7a94d17

nomos-services/blend/src/core/mod.rs
youngjoon-lee: Moved it to before signaling ready: 824af744ff13fb380923f4c7970e229c962da7c2

I personally prefer declaring variables right before their usage. But, I agree with your point since all other variables are initialized/declared at the top.

nomos-blend/scheduling/src/session.rs
youngjoon-lee: I agree that it might sound unclear. I renamed it to `await_first_ready`, which I think is better. 834c23c44f9ae90d47e035949fe56ff5ba5ba4c1

ntn-x2: I still think it does not give the idea that we are **expecting** it to yield an element within the given timeout, not just awaiting that the first item is ready. Maybe `expect_first_ready` is actually a better fit, in my opinion. Nevertheless, address this comment in any way you like. I'll go ahead and approve since it looks good!

youngjoon-lee: In my opinion, `await_first_ready` makes sense because we explicitly set `first_ready_timeout`. I considered naming it `expect_first_ready` at first, but didn't choose it because it could imply that it internally calls `expect(..)` (or `panic!`).

Currently, the code looks like below, which I think reasonable.
```rust
let (first_session, session_stream) = UninitializedSessionEventStream::new(
  session_stream,
  first_ready_timeout,
  transition_period,
)
.await_first_ready()
.await
.expect("The first session must be ready in time");
```

youngjoon-lee: But, I'm open to change :) I honestly don't have strong preference. Thank you again. I just merged this PR but we can change this name anytime later.

PR #1668, opened 2025-09-09
chore(kzgrs): Optimize h roots calculation

### LINKED ISSUE #1667: Optimize h-powers generation in combined polynomial
nomos-da/kzgrs/src/bdfg_proving.rs

In the `compute_h_roots`, powers of the challenge scalar `h` are generated as `map(|i| h.pow([i]))` separately. With the `parallel` feature enabled, exponentiations are distributed across threads, so each `h^i` is recomputed independently. This adds redundant work. Each `h.pow([i])` call is an O(log i) exponentiation. For size rows, this yields O(size x log p) complexity instead of linear. 
A simple multiplication chain `(h^0 = 1; h^{i+1} = h^i * h)` can be faster in this case.

danielSanchezQ:

Fixes #1667 

Simple

@danielSanchezQ @megonen 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.


megonen (commented):
Since the following lines is no longer used, I think it can be removed.
#[cfg(feature = "parallel")]
use rayon::iter::{IntoParallelIterator as _, ParallelIterator as _};

nomos-da/kzgrs/src/bdfg_proving.rs
```diff
@@ -233,16 +232,7 @@ pub fn verify_column(
 }
 
 fn compute_h_roots(h: Fr, size: usize) -> Vec<Fr> {
```
thomaslavaur, line 234: This can also be changed on the verification procedure

thomaslavaur: My bad I was thinking we was using it somewhere else :/

nomos-da/kzgrs/src/bdfg_proving.rs
```diff
@@ -233,16 +232,7 @@ pub fn verify_column(
...
-    {
-        #[cfg(feature = "parallel")]
```
bacv, line 237: Was this the only place where feature flag is used? Then we could remove from cargo.toml

danielSanchezQ: No it isn't the only place.

PR #1671, opened 2025-09-10
fix(zk): Fix proof sizes in proof libraries

thomaslavaur:

Compression of the Groth16 proofs in order to be 128 B

Yes

@thomaslavaur 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/zk/groth16/Cargo.toml
+++ b/zk/groth16/Cargo.toml
@@ -9,6 +9,7 @@ ark-bn254   = { default-features = false, features = ["curve"], version = "0.4" 
 ark-ec      = "0.4"
 ark-ff      = { version = "0.4" }
 ark-groth16 = { default-features = false, features = ["std"], version = "0.4" }
+ark-serialize = "0.4.2"
 hex         = { version = "0.4", default-features = false, features = ["alloc"], optional = true }
 num-bigint  = { version = "0.4", default-features = false, optional = true }
 serde       = { default-features = false, optional = true, features = ["derive"], version = "1.0" }
--- a/zk/groth16/Cargo.toml
+++ b/zk/groth16/Cargo.toml
@@ -9,7 +9,7 @@ ark-bn254     = { default-features = false, features = ["curve"], version = "0.4
 ark-ec        = "0.4"
 ark-ff        = { version = "0.4" }
 ark-groth16   = { default-features = false, features = ["std"], version = "0.4" }
-ark-serialize = "0.4.2"
+ark-serialize = { default-features = false, version = "0.4.2" }
 hex           = { default-features = false, features = ["alloc"], optional = true, version = "0.4" }
 num-bigint    = { default-features = false, optional = true, version = "0.4" }
 serde         = { default-features = false, features = ["derive"], optional = true, version = "1.0" }
--- a/zk/groth16/Cargo.toml
+++ b/zk/groth16/Cargo.toml
@@ -5,16 +5,16 @@ name              = "groth16"
 version           = "0.1.0"
 
 [dependencies]
-ark-bn254   = { default-features = false, features = ["curve"], version = "0.4" }
-ark-ec      = "0.4"
-ark-ff      = { version = "0.4" }
-ark-groth16 = { default-features = false, features = ["std"], version = "0.4" }
+ark-bn254     = { default-features = false, features = ["curve"], version = "0.4" }
+ark-ec        = "0.4"
+ark-ff        = { version = "0.4" }
+ark-groth16   = { default-features = false, features = ["std"], version = "0.4" }
 ark-serialize = "0.4.2"
-hex         = { version = "0.4", default-features = false, features = ["alloc"], optional = true }
-num-bigint  = { version = "0.4", default-features = false, optional = true }
-serde       = { default-features = false, optional = true, features = ["derive"], version = "1.0" }
-serde_json  = { default-features = false, optional = true, version = "1.0", features = ["alloc"] }
-thiserror   = { default-features = false, optional = true, version = "2.0" }
+hex           = { default-features = false, features = ["alloc"], optional = true, version = "0.4" }
+num-bigint    = { default-features = false, optional = true, version = "0.4" }
+serde         = { default-features = false, features = ["derive"], optional = true, version = "1.0" }
+serde_json    = { default-features = false, features = ["alloc"], optional = true, version = "1.0" }
+thiserror     = { default-features = false, optional = true, version = "2.0" }
 
 [dev-dependencies]
 serde_json = { default-features = false, features = ["alloc"], version = "1.0" }
--- a/zk/groth16/src/proof/mod.rs
+++ b/zk/groth16/src/proof/mod.rs
@@ -4,6 +4,7 @@ pub mod deserialize;
 #[cfg(feature = "deser")]
 use ark_bn254::Bn254;
 use ark_ec::pairing::Pairing;
+use ark_serialize::{CanonicalDeserialize, CanonicalSerialize, SerializationError};
 
 #[cfg(feature = "deser")]
 use crate::from_json_error::FromJsonError;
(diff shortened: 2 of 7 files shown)


zk/groth16/src/proof/mod.rs
```diff
@@ -20,6 +21,12 @@ pub struct Proof<E: Pairing> {
     pi_c: E::G1Affine,
 }
 
+pub struct CompressedProof {
+    pi_a: Vec<u8>,
```
ntn-x2, line 25: We don't know the exact length of each components? Meaning, can't we use arrays with a fixed length for each of the fields instead of `Vec`s?

zk/groth16/Cargo.toml
```diff
@@ -5,15 +5,16 @@ name              = "groth16"
...
-hex         = { version = "0.4", default-features = false, features = ["alloc"], optional = true }
-num-bigint  = { version = "0.4", default-features = false, optional = true }
-serde       = { default-features = false, optional = true, features = ["derive"], version = "1.0" }
-serde_json  = { default-features = false, optional = true, version = "1.0", features = ["alloc"] }
-thiserror   = { default-features = false, optional = true, version = "2.0" }
+ark-bn254     = { default-features = false, features = ["curve"], version = "0.4" }
+ark-ec        = "0.4"
+ark-ff        = { version = "0.4" }
+ark-groth16   = { default-features = false, features = ["std"], version = "0.4" }
+ark-serialize = "0.4.2"
```
ntn-x2: Can you disable default features with `default-features = false`?

zk/proofs/poc/src/lib.rs
```diff
@@ -118,7 +121,11 @@ pub fn prove(inputs: &PoCWitnessInputs) -> Result<(PoCProof, PoCVerifierInput),
 ///   underlying verification process fails.
 pub fn verify(proof: &PoCProof, public_inputs: &PoCVerifierInput) -> Result<bool, impl Error> {
     let inputs = public_inputs.to_inputs();
-    groth16::groth16_verify(verification_key::POC_VK.as_ref(), proof, &inputs)
+    groth16::groth16_verify(
+        verification_key::POC_VK.as_ref(),
+        &Groth16Proof::try_from(proof).unwrap(),
```
ntn-x2: Mhh here we don't want to `unwrap` tho. We are expanding the proof from the compressed one, and we don't know if it's valid. So if `groth16_verify` can only verify uncompressed proofs, then we need to capture the error of `try_from` and return it. Calling `unwrap` on untrusted input is not a thing we should do.

zk/proofs/poc/src/lib.rs
```diff
@@ -88,8 +90,9 @@ pub fn prove(inputs: &PoCWitnessInputs) -> Result<(PoCProof, PoCVerifierInput),
     let proof: Groth16ProofJsonDeser = serde_json::from_slice(&proof).map_err(ProveError::Json)?;
     let verifier_inputs: PoCVerifierInputJson =
         serde_json::from_slice(&verifier_inputs).map_err(ProveError::Json)?;
+    let proof: Groth16Proof = proof.try_into().map_err(ProveError::Groth16JsonProof)?;
     Ok((
-        proof.try_into().map_err(ProveError::Groth16JsonProof)?,
+        CompressedGroth16Proof::try_from(&proof).unwrap(),
```
ntn-x2, line 96: Here it's fine to uwnrap since we have generated the proof locally.

zk/proofs/pol/src/lib.rs
```diff
@@ -120,7 +123,11 @@ pub fn prove(inputs: &PolWitnessInputs) -> Result<(PoLProof, PolVerifierInput),
 ///   underlying verification process fails.
 pub fn verify(proof: &PoLProof, public_inputs: &PolVerifierInput) -> Result<bool, impl Error> {
     let inputs = public_inputs.to_inputs();
-    groth16::groth16_verify(verification_key::POL_VK.as_ref(), proof, &inputs)
+    groth16::groth16_verify(
+        verification_key::POL_VK.as_ref(),
+        &Groth16Proof::try_from(proof).unwrap(),
```
ntn-x2: Same as for the `poc`. We should not `unwrap()` and unverified proof.

zk/proofs/poq/src/lib.rs
```diff
@@ -94,7 +97,11 @@ pub fn prove(inputs: &PoQWitnessInputs) -> Result<(PoQProof, PoQVerifierInput),
 ///   underlying verification process fails.
 pub fn verify(proof: &PoQProof, public_inputs: &PoQVerifierInput) -> Result<bool, impl Error> {
     let inputs = public_inputs.to_inputs();
-    groth16::groth16_verify(verification_key::POQ_VK.as_ref(), proof, &inputs)
+    groth16::groth16_verify(
+        verification_key::POQ_VK.as_ref(),
+        &Groth16Proof::try_from(proof).unwrap(),
```
ntn-x2: Same.

zk/proofs/zksign/src/lib.rs
```diff
@@ -96,7 +99,7 @@ pub fn verify(
 ) -> Result<bool, impl Error> {
     groth16::groth16_verify(
         verification_key::ZKSIGN_VK.as_ref(),
-        proof,
+        &Groth16Proof::try_from(proof).unwrap(),
```
ntn-x2: Same.

ntn-x2 (changes requested):
Besides not unwrapping an unverified proof, I was wondering if we know the exact size of the compressed points, so that we could use fixed-size arrays instead of vectors?

zk/groth16/src/proof/mod.rs
thomaslavaur: The size depends of the elliptic curve so I guess no?

zk/proofs/poc/src/lib.rs
thomaslavaur: How can I catch it? I don't know how to deal with it because there is two types of possible errors we can return

zk/groth16/src/proof/mod.rs
danielSanchezQ: We should do this differently. We can specialize it for our need and implement it just for Proof<BN254>.

ntn-x2: I just realized I posted my answer under a different comment.

> There is probably a way to know the size of the compressed proof a compile time, maybe we would need a different type per proof. But anyway, let's keep them Vecs for now, and we will just panic at runtime if it has a size larger than what we expect in each place where the proof is consumed.

ntn-x2: I agree we should have a different proof size (and type) per proof, so that we can enforce a compile-time length on the underlying byte sequence.

danielSanchezQ (changes requested):
We should have an implementation with fixed sized for BN254 instead of a generic one probably. I will make a proper review in a few hours...I'm doing this from the phone 😅

zk/groth16/src/proof/mod.rs
ntn-x2: @danielSanchezQ I would suggest to merge this PR, so I can rebase mine on top of this, and we can open a new one that has the different proof types. What do you think? I need this PR merged in order to open my first PR that integrates these proofs.

thomaslavaur: Every Groth16 proofs will have the same size. The size just change with the elliptic curve used. So we can have a CompressedGroth16ProofBn254 for example and the size would then be fixed

danielSanchezQ: Ok, but will imply a change later on. I'll make it green.

danielSanchezQ (approved):
Will change this implementation later. But for the shake of unblocking @ntn-x2 we can go ahead.

PR #1674, opened 2025-09-10
chore(groth16): Implement compressed proof as constant sizes

danielSanchezQ:
## 1. What does this PR implement?

Tie up types and sizes for compressed proofs in groth16 crate

## 2. Does the code have enough context to be clearly understood?

Yes, just a few changes.

## 3. Who are the specification authors and who is accountable for this PR?

@thomaslavaur @danielSanchezQ @ntn-x2 

## 4. Is the specification accurate and complete?

N/A

## 5. Does the implementation introduce changes in the specification?

N/A

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [x] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [x] 2. Description added.
* [x] 3. Context and links to Specification document(s) added.
* [x] 4. Main contact(s) (developers and specification authors) added
* [x] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 6. Link PR to a specific milestone.

PR #1678, opened 2025-09-10
feat(zk): Improve API ergonomics

zeegomo:

* Add a way to construct the PoL verifier input from outside
* Change epoch_nonce from `u64` to `Fr` (it's a hash)
* Derive `Debug`  for public types
* Easier conversion for hex to `BigInt` and `u64` to `Fr`

Yes

N/A

N/A
N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.


zk/groth16/src/proof/mod.rs
```diff
@@ -30,6 +31,18 @@ impl<E: Pairing> From<&Proof<E>> for ark_groth16::Proof<E> {
         }
     }
 }
+
+impl<E: Pairing> From<ark_groth16::Proof<E>> for Proof<E> {
```
danielSanchezQ: Why this? Will we leak the ark one or planning to use it? Is it related to de/serialization?

zk/proofs/pol/src/lottery.rs
```diff
@@ -1,56 +1,36 @@
...
-/// `BigInt` type (signed), parsed the hex values correctly. But `BigUint` does
-/// not for the motives just described. Using the `str` representation allows
-/// use to choose the unsigned type.
+/// From [Proof of Leadership spec](https://www.notion.so/nomos-tech/Proof-of-Leadership-Specification-21c261aa09df819ba5b6d95d0fe3066d?source=copy_link#256261aa09df800fbc88e5aae5ea7e06)
 pub static P: LazyLock<BigUint> = LazyLock::new(|| {
-    BigUint::from_str(
-        "21888242871839275222246405745257275088548364400416034343698204186575808495617",
+    BigUint::from_str_radix(
+        "30644e72e131a029b85045b68181585d2833e84879b9709143e1f593f0000001",
+        16,
```
danielSanchezQ, line 13: Does this work?

zk/groth16/src/proof/mod.rs
zeegomo: Yes this is to serialize a proof (also we already have the `From` in the other direction so we're not leaking more stuff). We could also derive serialization / deserialization internally if you prefer

danielSanchezQ: Check on the lately added Compressed form of the proof, which is what we need to use for serialization/deserialization:
https://github.com/logos-co/nomos/pull/1674

danielSanchezQ: Serde is not derived for it tho. I would prefer to have an external type for serializing it where needed. Although I understand it may be easier to derive here. If we chose to derive, lets feature gate it (?)

zk/proofs/pol/src/lottery.rs
zeegomo: [Yes?](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=8afb6dc653e4541c8811f3c3447dd443)
I could add tests as well if necessary but it's a bit of a strange to put there, I don't see why this shouldn't work 🤔

zk/groth16/src/proof/mod.rs
zeegomo: uuh, thanks, taking a look.

zk/proofs/pol/src/lottery.rs
danielSanchezQ: I had it like that. But it didn't work. My understanding was the warning I posted in the comments. But It may have been due to disparity of constants when we were updating the circuits and the binaries. If tests do not pass it may be due to this. If they are green better in this form ofc.

danielSanchezQ: Well, it looks like it worked. So even better. Thanks for double checking!

PR #1680, opened 2025-09-10
chore(groth16): Match sizes and pairings at type level

danielSanchezQ:

It just adds extra type level information for sizes related to compressed forms of proofs related to pairing.
I had to use GenericArray, but I think its a good solution. Thanks @zeegomo for the discussion.

Its simple.

@danielSanchezQ 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/zk/groth16/src/proof/mod.rs
+++ b/zk/groth16/src/proof/mod.rs
@@ -25,17 +25,32 @@ pub struct Proof<E: Pairing> {
     pub pi_c: E::G1Affine,
 }
 
-#[derive(Clone, Debug)]
 pub trait CompressSize: Pairing {
     type G1CompressedSize: ArrayLength;
     type G2CompressedSize: ArrayLength;
 }
 
+#[derive(Clone, Debug)]
 pub struct CompressedProof<E: CompressSize> {
     pub pi_a: GenericArray<u8, E::G1CompressedSize>,
     pub pi_b: GenericArray<u8, E::G2CompressedSize>,
     pub pi_c: GenericArray<u8, E::G1CompressedSize>,
     _pairing: PhantomData<E>,
+}
+
+impl<E: CompressSize> CompressedProof<E> {
+    pub const fn new(
+        pi_a: GenericArray<u8, E::G1CompressedSize>,
+        pi_b: GenericArray<u8, E::G2CompressedSize>,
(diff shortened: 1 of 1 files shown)


zeegomo (approved):

zk/groth16/src/lib.rs
```diff
@@ -16,11 +16,14 @@ mod verifier;
 pub use ark_bn254::{Bn254, Fr};
 pub use ark_ff::Field;
 pub use verifier::groth16_verify;
-const BN254_G1_COMPRESSED_SIZE: usize = 32;
-const BN254_G2_COMPRESSED_SIZE: usize = 64;
+
+impl proof::CompressSize for Bn254 {
+    type G1CompressedSize = generic_array::typenum::U32;
```
ntn-x2, line 25: I mean, if this is anyway not pulled from some other definition but we are simply providing this value, why do we need to use `GenericArray` and `ArrayLength`? Would it not be sufficient to define a `const COMPRESSED_SIZE: usize` in the trait and implement it for the compressed points directly? I can't see what the advantage of this approach is, or if it makes possible something that would not be possible otherwise? Would you help me understand better?

danielSanchezQ: Yes, so you are on the good track. Ill explain what happend here:

* @zeegomo Raised a question on why did we have the `E: Paring` in the compressed form as well as the constant sizes (in the previous implementation). His concern was that it wasnt usuful becuase you couldn't match a pairing with specific sizes. Someone could still build a tipe with same pairing and different sizes, the compiler would still not catch the mismatch on sizing. TLDR: We needed a way of matching a `Pairing` with some sizes, just a single match. So the solution idea was to wrap that match in a trait that need to be implemented by each `Pairing`.
* I tried the approach of using `const G1_COMRESSED_SIZE: usize` in the trait. But Rust doesn't allow them to be used in constant generics...(yes it doesn't make sense to me neither).
* The solution was to use types instead of constants. `GenericArray` solves it.

I hope this gives more context on the process followed here.

ntn-x2: Yes, thank you! The potential mismatch between the compressed size and the pairing was also a concern for me that I raised in the other PR, so glad to see this was tackled here.

PR #1681, opened 2025-09-11
feat(blend): use membership stream provided by membership service

### LINKED ISSUE #1532: Use memberships provided by the membership service
Read the new membership from SDP and propagate it to all relevant services.

youngjoon-lee:

Closes #1532 

Finally, the blend service uses the membership stream provided by the membership service. 

It involves the following additional changes:
- Removed the `blend.membership` config from YAML files.
- Updated the test framework to include Blend service providers.

Yes

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/tests/src/topology/configs/membership.rs
+++ b/tests/src/topology/configs/membership.rs
@@ -69,6 +69,9 @@ pub fn create_membership_configs(
     ids.iter().map(|_| config.clone()).collect()
 }
 
+/// Create membership configs without DA providers.
+/// Blend providers should be included because the Blend service expects to have
+/// the initial membership at startup.
 #[must_use]
 pub fn create_empty_da_membership_configs(
     ids: &[[u8; 32]],


tests/src/topology/configs/membership.rs
```diff
@@ -49,20 +69,26 @@ pub fn create_membership_configs(ids: &[[u8; 32]], ports: &[u16]) -> Vec<General
...
-                },
-            },
-        });
+    // Remove DA config
+    for config in &mut configs {
+        config
+            .service_settings
+            .backend
+            .session_zero_providers
+            .remove(&ServiceType::DataAvailability);
```
pradovic, line 90: I think we might need DataAvailability service type set and  initialized to empty, as service type needs to be set in configs even if they are empty, because of the session size. I might be wrong, maybe I am missing the whole context

youngjoon-lee: Yeah, I also had the same thought, but removed the service type from the `session_zero_providers`, because the `session_zero_providers` was previously just an empty hashmap. It seems that any missing item in the `session_zero_providers` is initialized by the following logic: https://github.com/logos-co/nomos/blob/4d64201ade79b58029ca116e307b86b4fa34d599/nomos-services/membership/src/backends/mock.rs#L43-L48.

So, I think, either way is fine.

pradovic: Ah, yes, sorry! :)

youngjoon-lee: No worries, thank you!

PR #1683, opened 2025-09-11
feat(groth16): Groth16 batch verification

thomaslavaur:

Implement the Groth16 batch verification for v1.1 as defined in [Block verification v1.1](https://www.notion.so/nomos-tech/v1-1-Block-Construction-Validation-and-Execution-Specification-269261aa09df807185a9e0764acffe22?source=copy_link#269261aa09df80669556fcac40ea7f8b)

Yes

@thomaslavaur 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


zk/proofs/zksign/src/lib.rs
```diff
@@ -112,6 +113,28 @@ pub fn verify(
     .map_err(|e| VerifyError::ProofVerify(Box::new(e)))
 }
 
+pub fn batch_verify(
+    proofs: &[ZkSignProof],
+    public_inputs: &[ZkSignVerifierInputs],
```
megonen: since each proof is verified against its own public-input vector, should we add proofs.len() == public_inputs.len() check here ?

thomaslavaur: because of the Groth16 verification part after I don't think it's necessary. If some public inputs are missing or if there is too much input, the check won't pass. Moreover, the public inputs are supposed to be constructed by the verifier outside this library

davidrusu:
Can you update the PR description (see the checklist in the template)

nomos-blend/message/Cargo.toml
```diff
@@ -15,10 +15,10 @@ groth16         = { workspace = true }
 itertools       = { default-features = false, version = "0.14" }
 nomos-core      = { workspace = true }
 nomos-utils     = { features = ["rng"], workspace = true }
-num-bigint      = { version = "0.4", default-features = false }
+num-bigint      = { default-features = false, version = "0.4" }
 poq             = { workspace = true }
 serde           = { default-features = false, version = "1" }
-serde-big-array = { version = "0.5", default-features = false }
+serde-big-array = { default-features = false, version = "0.5" }
```
davidrusu: are these cargo toml changes from a lint? curious why it's showing up only on your PR?

thomaslavaur: I think no. But maybe I removed it to check something and when I reinserted it I used the same architecture as before.

thomaslavaur:
> 

What did you see missing? I've added the annex specification link but I didn't catch something else?

nomos-da/kzgrs/src/kzg.rs
```diff
@@ -33,22 +30,16 @@ pub fn commit_polynomial(
...
     }
-
-    // Instead of evaluating over the polynomial, we can reuse the evaluation points
-    // from the rs encoding let v = polynomial.evaluate(&u);
-    let v = evaluations.evals[element_index];
-    let f_x_v = polynomial + &DensePolynomial::<Fr>::from_coefficients_vec(vec![-v]);
     let x_u = DensePolynomial::<Fr>::from_coefficients_vec(vec![-u, Fr::one()]);
-    let witness_polynomial: DensePolynomial<_> = &f_x_v / &x_u;
-    let proof = commit_polynomial(&witness_polynomial, global_parameters)?;
+    let witness_polynomial: DensePolynomial<_> = polynomial / &x_u;
```
megonen: I’m not sure if f/x-u and f−v/x-u will give the same result in every case

megonen: It seems we give the proof for f(u)=0 instead of f(u)=v for arbitrary v

thomaslavaur: Here it's the euclidean division and we get the quotient so it doesn't change anything. Arkworks use the exact same thing in there poly-commit repository

thomaslavaur: Because the remainder is exactly this f(u) value.

davidrusu:
@thomaslavaur spec link was the main thing, just make sure to checkoff the checkboxes in the checklist that you've completed

nomos-blend/message/Cargo.toml
davidrusu: ok, then lets just revert these if it's not meaningful for the PR.

You can use this command to do the revert on each file
```
git checkout master -- nomos-blend/messages/Cargo.toml
```

nomos-da/kzgrs/src/kzg.rs
megonen: Ahh, I remember now, we checked this one 👍

nomos-blend/message/Cargo.toml
davidrusu: There are a bunch more toml files edited in the same way, make sure to revert those as well

thomaslavaur: Weird. When I see how much I reverted I guess my taplo did the change not me

zk/groth16/src/verifier.rs
```diff
@@ -14,6 +21,60 @@ pub fn groth16_verify<E: Pairing>(
...
+    g1_terms.push(vk.vk.vk.alpha_g1.mul(r_sum).neg().into());
+    g2_terms.push(vk.vk.vk.beta_g2.into());
+    g1_terms.push(batched_ic);
+    g2_terms.push(vk.vk.gamma_g2_neg_pc.clone());
+    g1_terms.push(batched_pi_c);
+    g2_terms.push(vk.vk.delta_g2_neg_pc.clone());
+
+    let qap = Bn254::multi_miller_loop(g1_terms, g2_terms);
+
+    let test = Bn254::final_exponentiation(qap).ok_or(SynthesisError::UnexpectedIdentity)?;
```
davidrusu: Can we use the pairing API here?
```suggestion
    let test = Bn254::multi_pairing(g1_terms, g2_terms);
```

zk/groth16/src/verifier.rs
```diff
@@ -14,6 +21,60 @@ pub fn groth16_verify<E: Pairing>(
...
+    Ok(test.0.is_one())
```
davidrusu: Is this correct?

I see a different implementation of testing the pairing result here:

```
impl<P: Pairing> Valid for PairingOutput<P> {
    fn check(&self) -> Result<(), SerializationError> {
        if self.0.pow(P::ScalarField::characteristic()).is_one() {
            Ok(())
        } else {
            Err(SerializationError::InvalidData)
        }
    }
}
```

namely, it checks that `test.0.pow(Fr::characteristic()).is_one()` rather than `test.0.is_one()`

consider replacing with

```suggestion
    test.check()
```

And have the function return `R1CSResult<()>` instead.

zk/groth16/src/verifier.rs
thomaslavaur: Yes I didn't know it was existing

zk/groth16/src/verifier.rs
thomaslavaur: It's ok and faster. Where did you find this one?
I think this check here just verify that the point is on the curve because any point on the curve would validate this verification (because the characteristic of the field is equal to 0 and it's like doing x^0 so it should be 1 which correspond to the 0 element of the curve)

davidrusu: https://github.com/arkworks-rs/algebra/blob/master/ec/src/pairing.rs#L155-L163

I understand, but the reason I dug into this is because you are going into the internals of the test result by accessing `test.0`. That was a red flag to me since surely the library authors had wrapped the pairing result in a struct for a reason, they could have just returned the pairing directly if it was not important.

I'd rather do the slower safe things here unless you can justify why the library does it differently than us. i.e. under what scenarios could the pairing produce a point outside the curve? are you 100% sure that the verifier can not be tricked into producing such a point? If there's any doubt, then lets just call `test.check()`.

thomaslavaur: No this is wrong! test.check() is not veryfing that the pairing is equal to 0 but that the pairing result is on the curve which is always true even for false proofs.

thomaslavaur: using this test.0 is precisely what arkwork did for groth16:
https://github.com/arkworks-rs/groth16/blob/b3b4a152b2f379930011ad0699b710d3746a552e/src/verifier.rs#L61

thomaslavaur: I can use Ok(test.is_zero()) as it's exactly the same

davidrusu: > No this is wrong! test.check() is not veryfing that the pairing is equal to 0 but that the pairing result is on the curve which is always true even for false proofs.

ah damn, you're right. That's a very confusing api

zk/groth16/src/verification_key/mod.rs
```diff
@@ -65,7 +65,7 @@ impl TryFrom<VerificationKeyJsonDeser> for VerificationKey<ark_bn254::Bn254> {
     }
 }
 pub struct PreparedVerificationKey<E: Pairing> {
-    vk: ark_groth16::PreparedVerifyingKey<E>,
+    pub vk: ark_groth16::PreparedVerifyingKey<E>,
```
danielSanchezQ: This probably should not be public, you can access using `into` anyway.

zk/groth16/src/verifier.rs
davidrusu: > I can use Ok(test.is_zero()) as it's exactly the same

yes, that's better

zk/groth16/src/verifier.rs
```diff
@@ -14,6 +21,58 @@ pub fn groth16_verify<E: Pairing>(
...
+
+    let test = Bn254::multi_pairing(g1_terms, g2_terms);
+    Ok(test.is_zero())
+}
+
+fn compute_r_powers(r: Fr, size: usize) -> Vec<Fr> {
+    std::iter::successors(Some(Fr::from(1)), |x| Some(r * x))
+        .take(size)
+        .collect()
+}
```
danielSanchezQ: If we use this a lot maybe is worth to cache it?

zk/groth16/tests/zk_signature_cpu_cycles.rs
danielSanchezQ, line 1: This should probably be a bench (?)

zk/groth16/tests/proof_of_claim_cpy_cycles.rs
danielSanchezQ, line 1: Can we check if we can move this to benches instead?

zk/groth16/src/verifier.rs
```diff
@@ -14,6 +21,58 @@ pub fn groth16_verify<E: Pairing>(
...
+    for (i, proof) in proofs.iter().enumerate() {
+        g1_terms.push(proof.pi_a.mul(r_roots[i]).into_affine());
+        g2_terms.push(proof.pi_b.into());
+    }
+    g1_terms.push(vk.vk.vk.alpha_g1.mul(r_sum).neg().into());
+    g2_terms.push(vk.vk.vk.beta_g2.into());
+    g1_terms.push(batched_ic);
+    g2_terms.push(vk.vk.gamma_g2_neg_pc.clone());
+    g1_terms.push(batched_pi_c);
+    g2_terms.push(vk.vk.delta_g2_neg_pc.clone());
```
danielSanchezQ: Can we add methods to our vk to access inner fields instead of nesting `vk` calls?

zk/groth16/src/verifier.rs
thomaslavaur: I'm not sure what you mean but we call this function several times but for different values of r so we cannot really keep it in memory

danielSanchezQ: > I'm not sure what you mean but we call this function several times but for different values of r so we cannot really keep it in memory

Brainfart. Didn't realize about `r`. Forget about it :sweat_smile:

zk/groth16/tests/zk_signature_cpu_cycles.rs
thomaslavaur: This was written like this by Alex. If we decide to move it I think it would be cleaner to do it in a different PR

zk/groth16/src/verifier.rs
thomaslavaur: vk is a structure from arkworks. How can you do it?

danielSanchezQ: > vk is a structure from arkworks. How can you do it?

No, its not. We have our own prepared verification key wrapper type on top of arkworks one (that is why there are so  many nested vk). We can add methods to it (?)

thomaslavaur: done

zk/groth16/src/verifier.rs
```diff
@@ -14,6 +21,58 @@ pub fn groth16_verify<E: Pairing>(
...
+    let r = Fr::rand(&mut rng);
+    let r_roots = compute_r_powers(r, proofs.len());
+    let r_sum: Fr = r_roots.iter().sum();
+
+    let pis_c: Vec<G1Affine> = proofs.iter().map(|proof| proof.pi_c).collect();
+
+    let batched_pi_c = G1Projective::msm(&pis_c, &r_roots).unwrap().into_affine();
+
+    let batched_public_inputs: Vec<Fr> = std::iter::once(r_sum)
+        .chain((0..public_inputs[0].len()).map(|i| {
```
davidrusu: This use of `public_inputs[0].len()` here seems unwise, idk if it can be exploited, but generally we shouldn't be conditioning verification logic on untrusted inputs if we can derive these values from trusted sources. Can we get this length from the verification key (something like `vk.ic.len() - 1`)?

```rust
        .chain((0..public_inputs[0].len()).map(|i| {
```

thomaslavaur: Done

thomaslavaur:
I've change the randomness generation from powers or r to different random number as now specified in the spec (to exactly match zcash specs)

PR #1690, opened 2025-09-11
feat(wallet): Core wallet functionality

davidrusu:

This PR introduces the wallet crate. The wallet will track the spendable notes corresponding to some known keys, it supports bootstrapping from LIB, and remaining in sync with Cryptarchia as new blocks are provided.

yes

@davidrusu 

NA

NA

## Checklist

* [x] 2. Description added.


wallet/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+                    temp_amount -= utxo.note.value;
+                    skip_count += 1;
+                } else {
+                    break;
+                }
+            }
+
+            selected_utxos.drain(..skip_count);
+
+            Some(selected_utxos)
```
danielSanchezQ: I think this should be its own method. The comment should be part of the method documentation.

danielSanchezQ (changes requested):
Looks quite clean thanks @davidrusu . I have just a little request.

wallet/src/lib.rs
davidrusu: yes, that'd be better, done!

danielSanchezQ (approved):
Good to go, thanks!

wallet/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+            }
+        }
+
+        Self { utxos, pk_index }
+    }
+
+    pub fn utxos_for_amount(
+        &self,
+        amount: Value,
+        pks: impl IntoIterator<Item = PublicKey>,
```
zeegomo, line 63: is this collecting only utxos with these pks? What's the intended usage for a wallet (like, aren't all utxos in here already under user control) ?

wallet/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+    ///
+    /// To resolve this, we remove as many of the smallest notes as we can while
+    /// still keep us above the requested amount.
+    fn remove_redundant_utxos(amount: Value, mut sorted_utxos: Vec<Utxo>) -> Vec<Utxo> {
+        debug_assert!(sorted_utxos.is_sorted_by_key(|utxo| utxo.note.value));
+
+        let mut skip_count = 0;
+        let mut temp_amount: Value = sorted_utxos.iter().map(|u| u.note.value).sum();
+
+        for utxo in &sorted_utxos {
```
zeegomo, line 164: nit: you could also iterate in reverse and stop when it's more than the requested value but this works too

wallet/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+        let known_keys: HashSet<PublicKey> = known_keys.into_iter().collect();
+        let wallet_state = WalletState::from_ledger(&known_keys, ledger);
+
+        Self {
+            known_keys,
+            wallet_states: [(lib, wallet_state)].into(),
+        }
+    }
+
+    pub fn apply_block(&mut self, block: &WalletBlock) -> Result<(), WalletError> {
```
zeegomo, line 200: this grows indefinitely?

davidrusu: In #1697, we introduce purging of wallet states as LIB advances

zeegomo: Ok yes, just saw that, thanks!

wallet/src/lib.rs
davidrusu: > is this collecting only utxos with these pks? 

Yep

> What's the intended usage for a wallet

I want eventually provide a CLI that will connect to this wallet service. Then all the normal account management ideas apply. e.g. pk for "business", pk for "personal" or whatever accounting system you want to use.

But all this is subject to change. We'll see how it's actually used

zeegomo: could you use different wallet instances instead?

davidrusu: yes, you could. I guess we'll see how we end up using it. I'm not married to this API

wallet/src/lib.rs
davidrusu: true, I wrote both out and it's basically the same logic, I didn't see a complexity reduction there.

I'll keep this one as is

Issue #1317, opened 2025-06-01
Proof of Selection


youngjoon-lee:
Include [proof of selection](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81d6bb3febd62b598138) when creating Blend messages.

Also, verify proofs when processing Blend messages.

Issue #1316, opened 2025-06-01
Proof of Quota


youngjoon-lee:
In the current code base, we're using a dummy value for Proof of Quota. 
It should be implemented as per [spec](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81edb561ef75a31f65a4).

PR #1691, opened 2025-09-11
feat(blend): introduce PoQ

ntn-x2:

First PR of a few, that will gradually introduce actual Proof of Quota and Proof of Selection everywhere in the codebase.

This PR is not meant to be merged as-is, since it is not compiling, but I decided to split up the functionality into a few stacked PRs for ease of reviewing. That is why I will keep this PR in a draft state until the whole branch is ready to be merged onto `master`. When everything is ready, this PR fixes https://github.com/logos-co/nomos/issues/1316 and fixes https://github.com/logos-co/nomos/issues/1317.

This PR does two things:

1. Re-organizes the `nomos-blend-message` crate by splitting large modules into smaller ones, and
2. Refactor the Proof of Quota type and its API, and integrate it into the `nomos-blend-message` crate, which is the only crate that, at the moment, successfully compiles and has all unit tests passing

In order to not having to generate actual proofs for tests, I evaluated two options:

1. Have a `ProofOfQuota` trait which would expose an API we need, and that would be implemented by the actual PoQ and by a mock one, or
2. Leave the ProofOfQuota as a concrete type and introduce a `Verifier` trait which deals with proof verification, that is injected, for now, in the method that verifies the public header of an encapsulated message.

Option 1 required to update all structs and methods dealing with ProofOfQuotas to be generic over the proof type, and I also foresaw we would need to do the same for ProofOfSelections. The second approach is much less invasive and it will be specular to a `ProofGenerator` trait (coming soon), which will be in charge of generic actual or mock proofs depending on the context.

A second PR will soon land on top of this one that introduces Proof of Selection to the `nomos-blend-message`. Then, few other PRs will integrate these features into the rest of blend core, and then into the blend service. Tests will also follow suit, once we are sure the current design fits well with the rest of the system.

Yes.

@ntn-x2 @madxor @youngjoon-lee 

Yes. There were a few inconsistencies as I went on with the implementation, but I believe they have all been addressed by now.

Yes, but all changes were minor.

## Checklist

* [x] 2. Description added.


nomos-blend/message/src/crypto/proofs/quota/inputs/prove/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+            core_path,
+            core_path_selectors,
+            core_sk,
+        }) => (
+            PoQBlendInputsData {
+                core_path,
+                core_path_selectors,
+                core_sk,
+            },
+            PoQWalletInputsData {
```
thomaslavaur: I think we should reused the function of PoQ that populate values for wallet data (you could modify it if needed)

nomos-blend/message/src/crypto/proofs/quota/inputs/prove/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+            aged_selector,
+            note_value,
+            output_number,
+            slot,
+            slot_secret,
+            slot_secret_path,
+            starting_slot,
+            transaction_hash,
+        }) => (
+            PoQBlendInputsData {
```
thomaslavaur: Ditto for the function for leaders

nomos-blend/message/src/crypto/proofs/quota/inputs/prove/private.rs
```diff
@@ -0,0 +1,74 @@
...
+
+#[derive(Debug, Clone)]
+pub enum ProofType {
+    CoreQuota(ProofOfCoreQuotaInputs),
+    LeadershipQuota(ProofOfLeadershipQuotaInputs),
+}
+
+#[derive(Debug, Clone)]
+pub struct ProofOfCoreQuotaInputs {
+    pub core_sk: ZkHash,
```
thomaslavaur, line 57: what is ZkHash here? Why not Fr?

nomos-blend/message/src/crypto/proofs/quota/inputs/prove/private.rs
```diff
@@ -0,0 +1,74 @@
...
+    fn from(value: ProofOfCoreQuotaInputs) -> Self {
+        Self::CoreQuota(value)
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct ProofOfLeadershipQuotaInputs {
+    pub slot: u64,
+    pub note_value: u64,
+    pub transaction_hash: ZkHash,
```
thomaslavaur, line 72: ditto

nomos-blend/message/src/crypto/proofs/quota/inputs/prove/public.rs
```diff
@@ -0,0 +1,36 @@
...
+        use crate::crypto::keys::Ed25519PrivateKey;
+
+        Self {
+            core_quota: u64::default(),
+            core_root: ZkHash::ZERO,
+            leader_quota: u64::default(),
+            pol_epoch_nonce: ZkHash::ZERO,
+            pol_ledger_aged: ZkHash::ZERO,
+            session: u64::default(),
+            signing_key: Ed25519PrivateKey::generate().public_key(),
```
thomaslavaur: Why not zero?

nomos-blend/message/src/crypto/proofs/quota/inputs/prove/public.rs
```diff
@@ -0,0 +1,36 @@
...
+            total_stake: 1,
```
thomaslavaur, line 31: ditto

nomos-blend/message/src/crypto/proofs/quota/inputs/prove/public.rs
```diff
@@ -0,0 +1,36 @@
+use nomos_core::crypto::ZkHash;
+
+use crate::crypto::keys::Ed25519PublicKey;
+
+/// Public inputs for all types of Proof of Quota. Spec: <https://www.notion.so/nomos-tech/Proof-of-Quota-Specification-215261aa09df81d88118ee22205cbafe?source=copy_link#25a261aa09df80ce943dce35dd5403ac>.
+#[derive(Debug, Clone, Copy)]
+pub struct Inputs {
+    pub session: u64,
+    pub core_root: ZkHash,
+    pub pol_ledger_aged: ZkHash,
```
thomaslavaur, line 10: can use Fr directly?

nomos-blend/message/src/crypto/proofs/quota/inputs/verify.rs
```diff
@@ -0,0 +1,49 @@
...
+
+/// Set of inputs required to verify a Proof of Quota.
+///
+/// It includes the inputs used to generate the proof (which must be fetched
+/// from the verifier's context), and the proof key nullifier, which is part of
+/// the Proof of Quota that is included in a Blend header.
+#[derive(Debug, Clone, Copy)]
+pub struct Inputs {
+    pub prove_inputs: PublicInputs,
+    pub key_nullifier: ZkHash,
```
thomaslavaur, line 15: I think there is confusion because of the python notation in specs. ZkHash is a hash function, we used zkhash for type in the specs like for hash in python that are bytes. for Poseidon2, ZkHash are Fr elements

nomos-blend/message/src/crypto/proofs/quota/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+        Ok(Self {
+            key_nullifier: key_nullifier.into_inner(),
+            proof,
+        })
+    }
+
+    #[must_use]
+    pub fn from_bytes_unchecked(bytes: [u8; PROOF_OF_QUOTA_SIZE]) -> Self {
+        let (key_nullifier_bytes, proof_circuit_bytes) = bytes.split_at(KEY_NULLIFIER_SIZE);
+        let key_nullifier = fr_from_bytes_unchecked(key_nullifier_bytes.try_into().unwrap());
```
thomaslavaur: I think it's better to check it's on Fr if it will be use for verification

nomos-blend/message/src/crypto/proofs/quota/serde.rs
```diff
@@ -0,0 +1,79 @@
...
+        fr_to_bytes(*input).serialize(serializer)
+    }
+
+    pub fn deserialize<'de, D>(deserializer: D) -> Result<ZkHash, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        fr_from_bytes(&Deserialize::deserialize(deserializer)?).map_err(Error::custom)
+    }
+}
```
danielSanchezQ: You can probably use `serde_fr` now as it does this for non human readable schemes.

nomos-blend/message/src/crypto/proofs/quota/serde.rs
```diff
@@ -0,0 +1,79 @@
...
+        S: Serializer,
+    {
+        fr_to_bytes(*input).serialize(serializer)
+    }
+
+    pub fn deserialize<'de, D>(deserializer: D) -> Result<ZkHash, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        fr_from_bytes(&Deserialize::deserialize(deserializer)?).map_err(Error::custom)
```
thomaslavaur: Do we check it's less than p before serializing into Fr?

zk/proofs/poq/src/inputs.rs
```diff
@@ -111,6 +112,19 @@ pub struct PoQVerifierInput {
     pub pol_ledger_aged: Groth16Input,
 }
 
+pub struct PoQVerifierInputData {
+    pub key_nullifier: Fr,
+    pub session: u64,
+    pub core_quota: u64,
```
thomaslavaur, line 118: we supposed in the circuit that both core_quota and leader_quota is less than 20 bits. I may have forget it so we need to ensure it somewhere

nomos-blend/message/src/crypto/proofs/quota/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+        .checked_add(BN254_G2_COMPRESSED_SIZE)
+        .expect("Second index overflow");
+    const THIRD_POINT_END_INDEX: usize = SECOND_POINT_END_INDEX
+        .checked_add(BN254_G1_COMPRESSED_SIZE)
+        .expect("Third index overflow");
+
+    (
+        bytes
+            .get(..FIRST_POINT_END_INDEX)
+            .unwrap()
```
danielSanchezQ:
```suggestion
        bytes[..FIRST_POINT_END_INDEX]
```
If you are unwrapping anyway...

nomos-blend/message/src/crypto/proofs/quota/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+        let key_nullifier = fr_from_bytes(key_nullifier_bytes.try_into().unwrap())
+            .map_err(|e| Error::InvalidInput(Box::new(e)))?;
+        let (pi_a, pi_b, pi_c) = split_proof_components(proof_circuit_bytes.try_into().unwrap());
+
+        Ok(Self {
+            key_nullifier,
+            proof: PoQProof::from_components(pi_a, pi_b, pi_c),
+        })
+    }
+}
```
danielSanchezQ: Why do we have this if `ProofOfQuota` implements `Deserialize`? Wouldn't it be better to implement deserialize manually then? 🤔

nomos-blend/message/src/crypto/proofs/quota/serde.rs
```diff
@@ -0,0 +1,79 @@
...
+    use groth16::{BN254_G1_COMPRESSED_SIZE, BN254_G2_COMPRESSED_SIZE};
+    use poq::PoQProof;
+    use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+    #[derive(Serialize, Deserialize)]
+    struct SerializationHelper(
+        [u8; BN254_G1_COMPRESSED_SIZE],
+        #[serde(with = "serde_big_array::BigArray")] [u8; BN254_G2_COMPRESSED_SIZE],
+        [u8; BN254_G1_COMPRESSED_SIZE],
+    );
```
danielSanchezQ: This can probably go away now that we have `GenericArray` (it supports implements Deserialize directly), you should be able to create a derived proxy type.

nomos-blend/message/src/crypto/proofs/mod.rs
```diff
@@ -0,0 +1,25 @@
+use nomos_core::crypto::ZkHash;
+
+use crate::crypto::proofs::quota::{inputs::prove::PublicInputs, ProofOfQuota};
+
+pub mod quota;
+pub mod selection;
+
+pub enum Error {
+    ProofOfQuota(quota::Error),
+}
```
danielSanchezQ: It does not implement error (?)

nomos-blend/message/src/crypto/proofs/quota/inputs/prove/mod.rs
ntn-x2: What do you mean exactly?

nomos-blend/message/src/crypto/proofs/quota/inputs/prove/private.rs
ntn-x2: It's a type alias, in our case. The types are equivalent.

nomos-blend/message/src/crypto/proofs/quota/inputs/prove/public.rs
ntn-x2: Yes I can change this.

nomos-blend/message/src/crypto/proofs/quota/inputs/verify.rs
ntn-x2: In Rust code we use `ZkHasher` for the hashing function, and `ZkHash` for the output of the hash function. I used `ZkHash` for convenience, but if you have a strong argument against it I can change it.

nomos-blend/message/src/crypto/proofs/quota/mod.rs
ntn-x2: This is used only when a message is initialized with pseudo-random data. When the message is deserialized, the bound checks are enforced, and deserialization will fail if the value is outside the allowed range.

nomos-blend/message/src/crypto/proofs/quota/serde.rs
ntn-x2: Yes, that's why the `fr_from_bytes` is a fallible function, because it fails in precisely that case.

zk/proofs/poq/src/inputs.rs
ntn-x2: If anything, we would need to enforce it at the compiler level, so we need to define a type that is always max 20 bits. If that is introduced in one of the zk crates, I have no problems changing it here as well.

nomos-blend/message/src/crypto/proofs/quota/mod.rs
ntn-x2: Yeah I am just not a fan of implicit panicking. I always prefer to use `unwrap` or `expect` whenever possible.

nomos-blend/message/src/crypto/proofs/quota/mod.rs
ntn-x2: We do implement `deserialize`, but that is generic over the deserializer. I don't understand how having a `Deserialize` implementation would make this redundant? Can you help me understand?

nomos-blend/message/src/crypto/proofs/quota/serde.rs
ntn-x2: Will try it out, thanks!

nomos-blend/message/src/crypto/proofs/mod.rs
ntn-x2: Yeah it should, you are right.

nomos-blend/message/src/encap/encapsulated.rs
```diff
@@ -71,108 +80,58 @@ impl<const ENCAPSULATION_COUNT: usize> EncapsulatedMessage<ENCAPSULATION_COUNT>
...
     }
 
-    #[must_use]
-    pub const fn public_header(&self) -> &PublicHeader {
-        &self.public_header
+    pub fn verify_and_unwrap_public_header<Verifier>(
+        self,
+        poq_verification_input: &PublicInputs,
+        verifier: &Verifier,
+    ) -> Result<UnwrappedEncapsulatedMessage<ENCAPSULATION_COUNT>, Error>
```
youngjoon-lee:
```suggestion
    ) -> Result<VerifiedEncapsulatedMessage<ENCAPSULATION_COUNT>, Error>
```
What do you think about calling it `VerifiedEncapsulatedMessage`. I don't have strong opinion but just asking which one is more appropriate.

nomos-blend/message/src/input.rs
```diff
@@ -36,16 +39,16 @@ impl<const ENCAPSULATION_COUNT: usize> EncapsulationInputs<ENCAPSULATION_COUNT>
...
 pub struct EncapsulationInput {
-    /// An ephemeral signing key generated by the encapsulating node.
-    pub(crate) signing_key: Ed25519PrivateKey,
+    /// An ephemeral signing key (ESK) generated by the encapsulating node.
+    ephemeral_signing_key: Ed25519PrivateKey,
     /// A shared key derived from the ephemeral encryption key and
     /// the selected blend node's non-ephemeral encryption key.
     /// Encryption keys are derived from the signing keys.
-    pub(crate) shared_key: SharedKey,
+    ephemeral_encryption_key: SharedKey,
```
youngjoon-lee: Although I approved this PR, I think we should use the previous name `shared_key` (or just changing it to `shared_secret`) because EEK (Ephemeral Encryption Key) means an ingredient of `SharedKey`, as described in the code comment and the [spec](https://www.notion.so/nomos-tech/Key-Types-and-Generation-Specification-215261aa09df81088b8fd7c3089162e8?source=copy_link#215261aa09df81a498deee1765724532).

nomos-blend/message/src/crypto/proofs/quota/inputs/prove/mod.rs
thomaslavaur: Reusing these functions:
https://github.com/logos-co/nomos/blob/68972a25a69320fa48e1b6b33160b27b1b4662f8/zk/proofs/poq/src/inputs.rs#L21

nomos-blend/message/src/crypto/proofs/quota/inputs/prove/public.rs
thomaslavaur: It would make more sense IMO

zk/proofs/poq/src/inputs.rs
thomaslavaur: It wasn't introduced. It needs to be added somewhere
For now, it's only in the circuit

danielSanchezQ: > It wasn't introduced. It needs to be added somewhere

this should be part of the poq crate ofc. Where is this data in the spec? If its not there it should be included 100%.

nomos-blend/message/src/crypto/proofs/quota/mod.rs
danielSanchezQ: I would prefer to use `expect` then.

nomos-blend/message/src/crypto/proofs/quota/mod.rs
danielSanchezQ: Maybe I'm missing something. What I mean is that we should not have bytes outside of the wire packing (?), so getting from bytes should be in deserialization?

nomos-blend/message/src/crypto/proofs/quota/serde.rs
ntn-x2: Thanks. Please check the latest commit (https://github.com/logos-co/nomos/pull/1691/commits/e55cdad561400d19fc8ab1feca1fe748a4a10fda) that addresses all your PR comments. Specifically, I removed the unnecessary `PhantomData` to allow for `serde(remote)` to work, I implemented `Copy` for `ProofOfQuota`, and I hope I got the serialization/deserialization correct according to how I understood `GenericArray`s are supposed to be used. Please let me know if there is a more efficient way of going from bytes to array, thanks!

nomos-blend/message/src/encap/encapsulated.rs
ntn-x2: Hey, this type is actually gone with the latest PR (https://github.com/logos-co/nomos/pull/1696). We don't need to unwrap anymore since we can't forward the same received message to other peers but sender. So in the latest PR I changed it to be a `Verified*` type which keeps all the components to be able to send it out to peers. So I will mark this comment as resolved, but please let me know in that PR if there is anything that is not yet convincing you. Thanks.

nomos-blend/message/src/input.rs
ntn-x2: You are absolutely right, thank you. I reverted the change in https://github.com/logos-co/nomos/pull/1691/commits/07f7b0a05ec83e7f93d608e19e6f1222b482740c.

zk/proofs/poq/src/inputs.rs
ntn-x2: I will leave this comment open until the PR that introduces this feature is merged, and we rebase this on top of that.

nomos-blend/message/src/crypto/proofs/quota/inputs/prove/mod.rs
ntn-x2: I did it here: https://github.com/logos-co/nomos/pull/1691/commits/7611e4215cbfa68aaf15535d9154a640bc7fe337. I did not like the repetition, so I'm glad we switched to using the defaults provided by `poq`. I also unified the inputs for the two different paths, since one was taking lower-level values while the other higher-level ones, I made them both higher-level.

PR #1692, opened 2025-09-11
feat(blend): introduce PoSel

ntn-x2:

PR built on top of https://github.com/logos-co/nomos/pull/1691. It mirrors the same API as for PoQ, integrating the necessary inputs in the message decapsulation process.

Next PR will try to make the code compile for the remaining of the Blend core crates.

Yes.

@ntn-x2 @madxor @youngjoon-lee 

Yes. There is a missing detail explaining how to convert `Fr` to bytes for hashing in the PoSel, but I opened a comment which will most likely be addressed soon: https://www.notion.so/Blend-Protocol-215261aa09df81ae8857d71066a80084?d=26b261aa09df80198a78001c9e617d18&source=copy_link#215261aa09df819991e6f9455ff7ec92.

See https://www.notion.so/Blend-Protocol-215261aa09df81ae8857d71066a80084?d=26b261aa09df80198a78001c9e617d18&source=copy_link#215261aa09df819991e6f9455ff7ec92.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/message/src/crypto/proofs/selection/mod.rs
+++ b/nomos-blend/message/src/crypto/proofs/selection/mod.rs
@@ -1,5 +1,5 @@
 use ::serde::{Deserialize, Serialize};
-use groth16::fr_from_bytes;
+use groth16::{fr_from_bytes, fr_from_bytes_unchecked, fr_to_bytes};
 use nomos_core::crypto::{ZkHash, ZkHasher};
 use num_bigint::BigUint;
 
@@ -13,7 +13,10 @@ mod serde;
 pub const PROOF_OF_SELECTION_SIZE: usize = size_of::<ProofOfSelection>();
 const DOMAIN_SEPARATION_TAG: [u8; 9] = *b"BlendNode";
 
-#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
+/// A Proof of Selection as described in the Blend v1 spec: <https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81d6bb3febd62b598138>.
+// TODO: To avoid proofs being misused, remove the `Clone` and `Copy` derives,
+// so once a proof is verified it cannot be (mis)used anymore.
+#[derive(Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
 pub struct ProofOfSelection {
     #[serde(with = "self::serde::selection_randomness")]
     selection_randomness: ZkHash,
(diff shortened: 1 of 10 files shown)


nomos-blend/message/src/crypto/proofs/selection/mod.rs
```diff
@@ -1,26 +1,107 @@
...
+        let selection_randomness_bytes = fr_to_bytes(&self.selection_randomness);
+        let selection_randomness_blake_hash =
+            blake2b512(&[&DOMAIN_SEPARATION_TAG[..], &selection_randomness_bytes[..]].concat());
+        let pseudo_random_output: u64 = {
+            let pseudo_random_output_bytes =
+                pseudo_random_sized_bytes::<8>(&selection_randomness_blake_hash);
+            let pseudo_random_biguint = BigUint::from_bytes_be(&pseudo_random_output_bytes[..]);
+            pseudo_random_biguint
+                .try_into()
+                .map_err(|_| Error::Overflow)?
```
youngjoon-lee, line 71: I guess the term `BigInt` in the spec should be changed to `unsigned 64bit integer`, since the definition of `BigInt` is not clear and may depend on platforms. Then, I think we can just use `u64` here instead of `num_bigint`. What do you think?

youngjoon-lee: Never mind, I just saw that `num_bigint` is also used in `groth16::fr_to_bytes`. I'm fine with it.
https://github.com/logos-co/nomos/blob/8f7f4ab894401576f4eef2fb552c8ed126da628c/zk/groth16/src/lib.rs#L42-L49

nomos-blend/message/src/encap/unwrapped.rs
```diff
@@ -1,27 +1,47 @@
...
 }
 
+#[derive(Debug, Clone, Copy)]
+#[cfg_attr(test, derive(Default))]
+/// Required inputs to verify a `PoSel` proof, minus the key nullifier that is
+/// retrieved from the verified `PoQ` of the outer Blend layer.
+pub struct MissingProofOfSelectionVerificationInputs {
+    pub expected_node_index: u64,
+    pub total_membership_size: u64,
+}
```
youngjoon-lee: I'd like to ask what `Missing` prefix means. Does it mean that this struct doesn't contain `key_nullifier`?

ntn-x2: Yeah I find naming particularly hard in these context, where we have one type, and then another one where the difference with the previous type is either one field less or one field more. Anyway, I renamed `Missing` to `Required`. Hopefully that clarifies it a bit more. https://github.com/logos-co/nomos/pull/1692/commits/4f85bd26d838f36de6034a6f069df10231de6cf0

I will go ahead and merge this on top of the PoQ one, please let me know if there's anything else.

PR #1694, opened 2025-09-12
feat(chain): Block broadcast service

bacv:

Block broadcast service that SDP or other services could subscribe to avoid circular dependency on chain service.

Proof of concept

@bacv @youngjoon-lee 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-services/chain/chain-service/src/lib.rs
+++ b/nomos-services/chain/chain-service/src/lib.rs
@@ -1095,8 +1095,8 @@ where
             let height = cryptarchia
                 .consensus
                 .branches()
-                .get(&cryptarchia.tip())
-                .expect("tip branch not available")
+                .get(&cryptarchia.lib())
+                .expect("LIB branch not available")
                 .length();
             let block_info = BlockInfo {
                 height,


nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -1085,8 +1091,20 @@ where
...
+            let height = cryptarchia
+                .consensus
+                .branches()
+                .get(&cryptarchia.tip())
+                .expect("tip branch not available")
+                .length();
+            let block_info = BlockInfo {
+                height,
+                header_id: new_lib,
+            };
```
youngjoon-lee:
```suggestion
        if prev_lib != new_lib {
            let height = cryptarchia
                .consensus
                .branches()
                .get(&cryptarchia.lib())
                .expect("LIB branch not available")
                .length();
            let block_info = BlockInfo {
                height,
                header_id: new_lib,
            };
```
 I think we need to get the height of the LIB, not the tip.

Other than that, this change looks correct. We may extend this logic to stream "potentially" finalized blocks during bootstrapping, but I can work on it soon. For now, this change is correct since LIB is never reverted and increase by 1, by its definition.

bacv: Thanks for spotting it, applied your suggestion.

PR #1697, opened 2025-09-13
feat(wallet): Wallet Service

davidrusu:

This PR adds the Wallet Service.
1. it integrates with cryptarchia, watching for changes to the chain and reflects them in the wallet
2. it exposes API's for fetching available notes for spending and leadership

yes

@davidrusu 

NA

NA

## Checklist

* [x] 2. Description added.


nomos-services/chain/chain-service-defs/src/api.rs
danielSanchezQ, line 1: Is not a bad idea. But we usually do it reversed. The consumer service implements some adapter for the functionalities he needs. We can consider this too for all services tho, adapters may use of this api structures directly.

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -740,7 +748,7 @@ where
                     }
 
                     Some(msg) = self.service_resources_handle.inbound_relay.next() => {
-                        Self::process_message(&cryptarchia, &self.block_subscription_sender, msg);
+                        Self::process_message(&cryptarchia, &self.block_subscription_sender, &self.lib_subscription_sender, msg);
```
danielSanchezQ: cc @bacv, I think he was working on lib subscriptions.

nomos-services/wallet/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+        // TODO: there may be a race condition here where the caller knows a more recent
+        // tip than the wallet. In that case, we will have received a
+        // LedgerState for the tip from Cryptarchia, but we would be missing the
+        // WalletState for that tip.
+        //
+        // Currently the way we deal with that is to just return an error but that's
+        // not ideal.
+        //
+        // To resolve, we could trigger an immediate sync here to ensure that the
+        // wallet is in sync with the caller and Cryptarchia.
```
danielSanchezQ: Maybe we can hold the execution awaiting till we have the proper tip (?)
We should open an issue if it is not gonna be resolved in this PR

nomos-services/chain/chain-service-defs/src/lib.rs
```diff
@@ -12,19 +16,51 @@ pub enum ConsensusMsg<Block> {
...
+        sender: oneshot::Sender<broadcast::Receiver<LibUpdate>>,
+    },
     GetHeaders {
         from: Option<HeaderId>,
         to: Option<HeaderId>,
         tx: oneshot::Sender<Vec<HeaderId>>,
     },
+    GetLedgerState {
+        block_id: HeaderId,
+        tx: oneshot::Sender<Option<nomos_ledger::LedgerState>>,
```
zeegomo: nit: should this be an error instead?

nomos-services/wallet/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+                return Err(WalletServiceError::BlockNotFoundInStorage(header_id));
+            };
+
+            if let Err(e) = wallet.apply_block(&block.into()) {
+                error!(
+                    block_id = ?header_id,
+                    err = %e,
+                    "Failed to apply backfill block to wallet"
+                );
+                return Err(WalletServiceError::BackfillFailedToApplyBlock(header_id));
```
zeegomo, line 337: can this actually happen? isn't parent missing the only possible error case?

nomos-services/wallet/src/lib.rs
zeegomo: could this happen for the other methods as well?

nomos-services/wallet/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+                Ok(block) = block_receiver.recv() => {
+                    let wallet_block = WalletBlock::from(block);
+                    match wallet.apply_block(&wallet_block) {
+                        Ok(()) => {
+                            trace!(block_id = ?wallet_block.id, "Applied block to wallet");
+                        }
+                        Err(WalletError::UnknownBlock) => {
+
+                            info!(block_id = ?wallet_block.id, "Missing block in wallet, backfilling");
+                            Self::backfill_missing_blocks(wallet_block.id, &mut wallet, &storage_adapter, &cryptarchia_api).await?;
```
zeegomo, line 191: under what circumstances would this happen?

nomos-services/chain/chain-service-defs/src/lib.rs
davidrusu: Not necessarily. Just like the StorageAdapter doesn't return error if you ask for a block that doesn't exist. 

I think it's an error only if the caller expected it to be there, but the caller could use this query to see if a LedgerState exists. The caller turns it into an error if they expected a ledger state that was not there.

nomos-services/wallet/src/lib.rs
davidrusu: Currently it's just parent missing, but not necessarily the case in the future if the wallet gets more error variants and we start tracking other state in the wallet as well.

nomos-services/wallet/src/lib.rs
davidrusu: No, it's only an issue for `GetLeaderAgedNotes` because it is fetching both from CryptarchiaService and from the Wallet. The other messages only query the Wallet so we don't risk a state mismatch.

nomos-services/wallet/src/lib.rs
davidrusu: When we first startup, we initialize from LIB and then we backfill from tip to LIB.

That's just a single chain from tip to LIB, we don't backfill all the other branches. If a new block comes in that extends a non-canonical branch, then it would be detected here and we'd do a JIT backfill for that branch.

nomos-services/wallet/src/lib.rs
davidrusu: True, lets do that, we already have the backfill function implemented, it's just a matter of calling it here to do a JIT sync.

davidrusu: made the change

nomos-services/chain/chain-service-defs/src/api.rs
davidrusu: yes, I first tried doing the relays in the naive way where we manage channels at the call site, but that was just too complicated. I found this a bit easier to use in the Wallet.

I would like to provide a similar API for the wallet as well

nomos-services/wallet/src/lib.rs
zeegomo: The api for wallet operations require a `HeaderId` that the caller fetches from somewhere else, I guess it could happen that that source it more up to date than the wallet?

davidrusu: yes, that's true. I guess we could try to backfill those requests as well.

The difference with this case is that we have proof that the wallet is behind Cryptarchia because we were able to fetch the ledger state corresponding to the tip. So backfilling *should* resolve the descripency.

I'll keep this in mind as we do integration with PoL

PR #1702, opened 2025-09-16
DRAFT: feat(chain): tx storage for proposals

andrussal:

Most of this if not all will be replaced by #1756 

---

This PR introduces transaction storage to keep transactions available for proposal reconstruction. As previously, processing a Proposal(Block) removes transactions from mempool. But now we also store them for later Proposals that may reference them. 

These transactions are removed from storage when a block including them becomes immutable. Any fork that comes after this immutable block can’t include them. And any fork that forked before this immutable block is pruned already. 

#### Storage

The new transaction storage is a **temporary store** that keeps transactions on disk only until we are certain they are no longer needed for future blocks.  Blocks, along with their transactions, are still stored **long-term** as before.

This design is ideal when we need a block together with its transactions (e.g., for IBD) — **O(1)** access. Without this approach, it would require **O(N + 1)**, where *N* is the number of transactions.  

If, in the future, we need to support queries for individual transactions, we can add an index mapping:

`TxHash → HeaderId`

This would allow us to easily locate the block containing a given transaction.

---

#### Mempool

The current mempool API remains almost **unchanged** because it doesn't need to know how transactions are consumed.  
It keeps the same simple interface:

- `View`
- `MarkInBlock`
- *(new)* `GetTransactionsByHashes`

**Alternative design:**  
We could expand the mempool API to include:

- `View`
- `MarkInBlock` – pushes transactions to temporary storage
- `Remove` – deletes them from temporary storage
- *(new)* `GetTransactionsByHashes`

However, this would require adding storage access to mempool service(which might be fine though).

---

#### Chain Service

When a block is received from the network or proposed locally, the chain service requests its transactions from the mempool.  Once the block is successfully added to the chain:

1. **Remove transactions** from the mempool.  
2. **Store transactions** in temporary storage.  
3. **Delete transactions** from temporary storage once the block becomes immutable.

---

#### Transactions Manager

The `TransactionsManager` is a new component within the chain service that **encapsulates this workflow**.  
It provides the following API:

- **`reconstruct_block_from_proposal`** – Fills a proposal with transactions.  
- **`store_block_transactions`** – Stores transactions in temporary storage after they are removed from the mempool.  
- **`cleanup_transactions_from_immutable_blocks`** – Deletes transactions from temporary storage once their block is finalized.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-core/chain-defs/src/header/mod.rs
+++ b/nomos-core/chain-defs/src/header/mod.rs
@@ -6,11 +6,11 @@ use serde::{Deserialize, Serialize};
 pub const BEDROCK_VERSION: u8 = 1;
 
 use crate::{
+    codec::SerdeOp,
     crypto::Hasher,
     proofs::leader_proof::{LeaderProof, Risc0LeaderProof},
     utils::{display_hex_bytes_newtype, serde_bytes_newtype},
 };
-use crate::codec::SerdeOp;
 
 #[derive(Clone, Debug, Eq, PartialEq, Copy, Hash, PartialOrd, Ord)]
 pub struct HeaderId([u8; 32]);
@@ -199,8 +199,8 @@ pub enum Error {
 #[test]
 fn test_serde() {
     assert_eq!(
-        <HeaderId as crate::codec::SerdeOp>::deserialize::<HeaderId>(
-            &<HeaderId as crate::codec::SerdeOp>::serialize(&HeaderId([0; 32]))
+        <HeaderId as SerdeOp>::deserialize::<HeaderId>(
+            &<HeaderId as SerdeOp>::serialize(&HeaderId([0; 32]))
                 .expect("HeaderId should be able to be serialized")
         )
         .unwrap(),
--- a/nomos-core/chain-defs/src/mantle/tx.rs
+++ b/nomos-core/chain-defs/src/mantle/tx.rs
@@ -1,12 +1,13 @@
 use std::sync::LazyLock;
 
 use bytes::Bytes;
-use groth16::{serde::serde_fr, Fr};
+use groth16::{fr_to_bytes, serde::serde_fr, Fr};
 use num_bigint::BigUint;
 use poseidon2::{Digest, ZkHash};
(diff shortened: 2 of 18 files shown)


zeegomo:
Thanks for the nice description!
Regarding the flow:
> When a block is received from the network or proposed locally, the chain service requests its transactions from the mempool. Once the block is successfully added to the chain:
> 1. Remove transactions from the mempool.
> 2. Store transactions in temporary storage.
> 3. Delete transactions from temporary storage once the block becomes immutable.

It seems we move from one temporary storage to another (mempool -> tx store). Have you though of an alternative that avoids this? Maybe the mempool does not contain the full txs but only hashes so that incoming txs are only saved in the temporary storage and we avoid that additional data movement.

zeegomo:
> Maybe I misunderstand it but don't we need to keep used transactions around in case some other proposals references them? If there are multiple proposals per slot? Or the same transaction is included in different forks? If this is not the case then maybe we don't need to change anything at all. Please let me know!

Sure you need to keep them around, but they can be in the tx store not in the mempool

danielSanchezQ:
> The role of the mempool does not change, it's still something that stores txs that can be added to new blocks. It would just not store the content itself but a reference to the content, just like a proposal contain a reference to the content instead of the content itself. For instance, I don't think we need to change the mempool api (not sure why you would need to store used txs).
> 
> As for `View`, it's not strictly necessary to access db, it could just return tx hashes and the consumer will access the db themselves, and `TransactionByHashes` might be best implemented in the tx store.
> 
> Chain service does not necessarily need to know the contents of the transactions, it just need to be sure they're valid. Ofc at some point in the process of building or validating a block you need to access the tx store (and mempool)
> 
> cc @danielSanchezQ

Yeah, I like the idea of making the mempool lightweight. Mempool can still do the exact same job but just store the tx id.
Prunning should be fine, we just prune on lib updates (iirc), so it is guaranteed that no older branch will need the tx anyway.

PR #1706, opened 2025-09-16
feat(blend): integrate PoQ and PoSel into rest of codebase

ntn-x2:

Last PR before the whole branch can be merged. It fixes the rest of the codebase (mostly the `nomos-blend-service` crate and the `nomos-node` and `nomos-executor` binaries). There is a lot of noise in the binaries, but unfortunately at this moment it is what it is. I tried to contain the changes into a new `blend` module that is then re-used by both the validator and the executor binaries, but still cryptarchia stuff needed to know about the Blend proof generation and verification logic, which I think is not ideal for what cryptarchia needs from Blend, but that discussion is out of scope here.

I use a mock stream within the Blend services, and a mock generator/verifier in the binaries. I also prepared the testing framework for the Blend service crate to be able to test more failure cases that come from invalid proofs. Next steps are:

1. Expand the existing tests to include more edge cases since we can now mock success/failure for both generation and verification of both PoQ and PoSel.
2. Integrate the actual stream coming from other services, most likely the wallet and the chain ones, to provide Blend with the session info it needs to build and verify proofs, after which we can swap the current mock implementations (which keep the logic as it is currently on `master`, i.e., does not verify proofs) with the actual one.

**EDIT: I found out that while the Blend spec does not prevent a node from sending messages to itself, this possibility breaks down in our integration test setup, because we only have two nodes, and if node A sends message to node B, node B will not send it back to node A, hence the message will never be decapsulated and it will get lost. For this reason I updated the API of the proof generator trait and the logic of the mock proof generator used in the binaries to try to generate a new proof whenever a generated one would be addressed to itself. In this way, we can ensure that all outgoing messages are addressed to the other peer.**

Yes.

Yes.

No. Necessary changes were introduced when working on previous PRs already.

## Checklist

* [x] 2. Description added.


ntn-x2:
CI failure for the feature combinations is unrelated to this PR as it happens also on `master`.

nomos-blend/scheduling/src/message_blend/crypto/send_and_receive.rs
```diff
@@ -0,0 +1,112 @@
...
+    #[must_use]
+    pub const fn from_sender_and_verifier(
+        sender_processor: SenderSessionCryptographicProcessor<NodeId, ProofsGenerator>,
+        proofs_verifier: ProofsVerifier,
+    ) -> Self {
+        Self {
+            sender_processor,
+            proofs_verifier,
+        }
+    }
```
youngjoon-lee: It seems that this method is not used anywhere except `Self::new`. Then, wdyt about merging this with `Self::new`? If we want to keep this method, I guess it should be renamed to `from_sender`?

ntn-x2: Good catch. It's a leftover from a late refactor I think. I am removing this.

PR #1708, opened 2025-09-16
feat(kms): Generalise keys and encodings

strinnityk:
Generalise `KMSBackend`, specifically the encodings and keys modules, to:
- Enforce safety at compile time through traits.
- Allow broader flexibility for key and encoding structures, accepting any type that satisfies the required traits (with some auto-implemented).
- Support multiple usage styles: relying only on the crate’s defaults, fully custom types, or mixing both approaches.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/nomos-services/key-management-system/src/backend/preload.rs
+++ b/nomos-services/key-management-system/src/backend/preload.rs
@@ -289,7 +289,7 @@ mod tests {
             Error = <<Backend as KMSBackend>::Key as SecuredKey>::Error,
         > + 'a;
     fn noop_operator<Backend: KMSBackend>() -> KMSOperatorBackend<Backend> {
-        Box::new(move |_: &mut BackendKey<Backend>| Box::pin(async move { Ok(()) }))
+        Box::new(move |_: &BackendKey<Backend>| Box::pin(async move { Ok(()) }))
     }
 
     #[tokio::test]
--- a/nomos-services/key-management-system/src/lib.rs
+++ b/nomos-services/key-management-system/src/lib.rs
@@ -22,7 +22,7 @@ use crate::{backend::KMSBackend, keys::SecuredKey};
 // TODO: Use [`AsyncFnMut`](https://doc.rust-lang.org/stable/std/ops/trait.AsyncFnMut.html#tymethod.async_call_mut) once it is stabilized.
 pub type KMSOperator<Payload, Signature, PublicKey, KeyError, OperatorError> = Box<
     dyn FnMut(
-            &mut dyn SecuredKey<
+            &dyn SecuredKey<
                 Payload = Payload,
                 Signature = Signature,
                 PublicKey = PublicKey,


nomos-services/key-management-system/src/keys/secured_key.rs
```diff
@@ -0,0 +1,75 @@
...
+/// # Example
+///
+/// ```ignore
+/// let signature = key.sign_adapted(adapted_data)?;
+/// ```
+///
+/// # Errors
+///
+/// Returns a [`KeyError`] if the conversion or signing fails.
+pub trait SecuredKeyAdapter<TargetEncodingFormat>: SecuredKey
```
ntn-x2: If we want a key to support multiple formats, why not having `EncodingFormat` be a generic for `SecuredKey` instead of an associated type? Then we probably would not need this trait. If we want to have a single representation of the public key, the base trait could define the public key format as an associated type, and the subtrait could be the generic that we use to pass the data to sign.

nomos-services/key-management-system/src/keys/ed25519.rs
```diff
@@ -1,20 +1,27 @@
...
 
 impl SecuredKey for Ed25519Key {
-    fn sign(&self, data: Bytes) -> Result<Bytes, DynError> {
-        Ok(Bytes::copy_from_slice(&self.0.sign(&data).to_bytes()))
+    type EncodingFormat = crate::encodings::Bytes;
+    type Error = KeyError;
+
+    fn sign(&self, data: Self::EncodingFormat) -> Result<Self::EncodingFormat, Self::Error> {
+        let data_bytes = data.as_bytes();
+        let signature_slice = self.0.sign(data_bytes).to_bytes();
```
ntn-x2: I think it would be better for the `sign` function to take a reference instead of owning the value, since what should happen is that a new value is produced, which is the signature. In case the chosen `EncodingFormat` does not implement `Copy`, or in case it is very large in size, we avoid copying large amounts of data to then just sign a reference to them.

nomos-services/key-management-system/src/keys/ed25519.rs
```diff
@@ -1,20 +1,27 @@
...
+        let signature_bytes = RawBytes::copy_from_slice(&signature_slice);
+        Ok(Self::EncodingFormat::from(signature_bytes))
     }
 
-    fn as_pk(&self) -> Bytes {
-        Bytes::copy_from_slice(self.0.verifying_key().as_bytes())
+    fn as_pk(&self) -> Self::EncodingFormat {
```
ntn-x2: Should this also return a reference to the public key? The caller could then decide if they want to clone/copy in case they need to store it somewhere, but if they need, for instance, it's byte representation or its size, then having a reference should be enough.

ntn-x2 (commented):
I want to see this updated KMS being used before understanding whether this complexity is indeed required. I left a couple of comments, and I would say in this case perhaps some `examples` would help, not only in having enough context to review the PR, but also for people that want to interact with or extend the KMS service.

nomos-services/key-management-system/src/keys/ed25519.rs
danielSanchezQ: yes, sign should get a reference of everything, no need to get ownership of stuff here.

nomos-services/key-management-system/src/keys/ed25519.rs
danielSanchezQ: Probably not (?) or we would need `Cow`, usually pk is not stored but derived from `sk`.

nomos-services/key-management-system/src/keys/secured_key.rs
danielSanchezQ: The problem of this is how to compose later with the api. His solution is to static dispatch. Otherwise we would have to support generic API messages later in the kms service (?)

nomos-services/key-management-system/src/keys/ed25519.rs
ntn-x2: But why return `Self::EncodingFormat` instead of returning `Self::PublicKey`? Or `&Self::PublicKey`?

nomos-services/key-management-system/src/keys/secured_key.rs
ntn-x2: This is it the bit that I am not sure how would work, that's why I mentioned maybe some `examples` would help, for keys other than the one already implemented, for instance, or for some non-trivial use cases. Then we would know whether this interface is all required, we need more, or we can do with less.

danielSanchezQ: Yeah, it can actually be part of the documentation as it may be clear in @strinnityk head but not for us, so :heavy_plus_sign: 1

nomos-services/key-management-system/src/keys/ed25519.rs
danielSanchezQ: Ah, I understood something else. You are right indeed. Its much clearer to have a pk type. I think this comes from the old bytes for everything old style.

nomos-services/key-management-system/src/keys/secured_key.rs
strinnityk: I feel like what you're proposing is the same as what's here. Kind of?

We don’t want a key to support multiple formats. Each `SecuredKey` has a single `EncodingFormat` for both input and output. That’s why the base trait fixes it as an associated type.

The reason we still have an adapter is on the backend side: a `KMSBackend` typically handles multiple key types, which means multiple formats overall (often enums). The adapter lets callers provide a different representation and convert it into the key’s encoding format. This way, each key stays strict with one format, while the KMS can flexibly accept multiple encodings across different keys.

nomos-services/key-management-system/src/keys/ed25519.rs
strinnityk: The entrypoint, `KMSBackend::sign`, receives `data` as value, and I have my doubts `data` will be used (in a different manner) outside of passing it to `SecuredKey::sign`. 

This being said, if we assume for all (or most) operations in `SecuredKey::sign` a ref is enough, we might as well make it that. Wdyt?

nomos-services/key-management-system/src/keys/ed25519.rs
strinnityk: Going by the specific example of `Ed25519`: When calling `self.0.verifying_key()` it's returning a value, likely a `Copy`. We're copying the bytes and wrapping them and returning them. Do you still think it makes sense to return a ref?

strinnityk: The reason we're returning `Self::EncodingFormat` is because that's what it is. The pk will follow that specific format.

It also facilitates the backend returning it.

nomos-services/key-management-system/src/keys/secured_key.rs
ntn-x2: Yes, I was probably proposing something similar, but in a simpler structure, in my opinion. A base trait which does not allow signing but just "bind" to a specific return type, and a polimorphic trait that allows to sign different payload types, if that makes sense.

strinnityk: I kinda feel it's exactly the same as what exists already. If you don't mind, can you give me a very basic pseudo-code example?

ntn-x2: You are right, it is. What I was suggesting was simply taking the `sign` method out of the "base" trait, and put it into a polimorphic trait, like below.

```rust
pub trait SecuredKey: ZeroizeOnDrop {
    type EncodingFormat: EncodingFormat;
    type Error;
    fn as_pk(&self) -> Self::EncodingFormat;
}

pub trait SigningKey<InputFormat>: SecuredKey {
    fn sign(&self, data: &InputFormat) -> Result<Self::EncodingFormat, Self::Error>;
}
```

In this way, a key can only implement `SecuredKey` once, but it can support multiple different input formats for signing operations.

danielSanchezQ: Uhm, I see, Yes we can do that. Then the constrain of tryfroms would be in the Signingkey implementation instead right? So we just decouple it if I'm right @ntn-x2 ?

ntn-x2: Yes, as far as I understood this PR, that was my suggestion. Pinning down the one-for-all-impls requirements into a base trait, and let the subtrait be a case-by-case implementation for the different formats supported.

danielSanchezQ: Yeah, I think its a good idea. Thanks.

danielSanchezQ: > We can provide the autotrait implementation for SigningKey if we want (and can) too. Imo.

Brainfart, I don't know why I edited it...I missclicked the reply and hit edit instead...

strinnityk: Right, but doing it this way means letting go of the idea of auto-impls. Doesn't it? Since we don't have a base `sign` we can use.
Or are we considering maintaining a `SigningKeyAdapter`?

nomos-services/key-management-system/src/backend/mod.rs
```diff
@@ -1,24 +1,38 @@
...
-
-use crate::KMSOperator;
+use crate::{keys::SecuredKey, KMSOperatorBackend};
 
 #[async_trait::async_trait]
 pub trait KMSBackend {
-    type SupportedKeyTypes;
     type KeyId;
+    type DataEncoding;
+    type SupportedKey: SecuredKey<Self::DataEncoding>;
```
danielSanchezQ: Probably you do not need `DataEncoding`, you can instead use `<Self::SuportedKey as SecuredKey>::DataEncoding` instead.

nomos-services/key-management-system/src/keys/ed25519.rs
```diff
@@ -1,20 +1,49 @@
...
 use serde::{Deserialize, Serialize};
 use zeroize::ZeroizeOnDrop;
 
-use crate::secure_key::SecuredKey;
+use crate::{
+    encodings::Encoding,
+    keys::{secured_key::SecuredKey, KeyError},
+};
+
+type InnerEncoding = crate::encodings::Bytes;
```
danielSanchezQ: Im not so sure about all this encoding naming to be honest. I would just use `Bytes` explicitly.

nomos-services/key-management-system/src/keys/ed25519.rs
```diff
@@ -1,20 +1,49 @@
...
+impl SecuredKey<Encoding> for Ed25519Key {
+    type Signature = Encoding;
+    type PublicKey = Encoding;
+    type Error = KeyError;
+
+    fn sign(&self, data: &Encoding) -> Result<Self::Signature, Self::Error> {
+        match data {
+            Encoding::Bytes(data) => self.sign(data).map(Encoding::from),
+        }
     }
```
danielSanchezQ: Is there 2 impl blocks for Ed25519Key? This second one is for the wrapper enum right?

danielSanchezQ (approved):
Looks good to me. We can improve ergonomics later on if needed. I will be expecting another PR with the zk keys support.

nomos-services/key-management-system/src/keys/ed25519.rs
strinnityk: Correct.

nomos-services/key-management-system/src/backend/mod.rs
strinnityk: We can't because `SecuredKey::DataEncoding` is a generic param, not an associated type.

nomos-services/key-management-system/src/keys/ed25519.rs
strinnityk: You mean:
```rust
enum Encoding {
    Bytes(bytes::Bytes)
}
```
?

nomos-services/key-management-system/src/backend/mod.rs
danielSanchezQ: Ah, true. Thanks.

nomos-services/key-management-system/src/keys/ed25519.rs
```diff
@@ -1,22 +1,49 @@
...
 #[derive(Serialize, Deserialize, ZeroizeOnDrop)]
 pub struct Ed25519Key(pub(crate) ed25519_dalek::SigningKey);
 
-impl SecuredKey for Ed25519Key {
-    fn sign(&self, data: Bytes) -> Result<Bytes, DynError> {
-        Ok(Bytes::copy_from_slice(&self.0.sign(&data).to_bytes()))
+impl SecuredKey<InnerEncoding> for Ed25519Key {
+    type Signature = InnerEncoding;
+    type PublicKey = InnerEncoding;
+    type Error = KeyError;
```
danielSanchezQ: @strinnityk [I meant something like](https://github.com/logos-co/nomos/pull/1708/files#r2376090300):

```rust
impl SecuredKey<InnerEncoding> for Ed25519Key {
    type Signature = [u8; 32];
    type PublicKey = [u8; 32];
    type Error = KeyError;
...
```
Or even the actual types. What does `Ed25519` use for pk's? Probably it has an specific type right?

strinnityk: Hmm, you mean the naming or the wrapper types?
So, `Ed25519` uses `bytes::Bytes`. Do you mean using `bytes::Bytes` instead of `encodings::Bytes`?

danielSanchezQ: Yeah, using specific types. `Signature` is probably specified in the Ed25519 create, same for `PublicKey`. Or should we always go through bridged types?

nomos-services/key-management-system/src/lib.rs
```diff
@@ -165,11 +199,12 @@ impl<Backend, RuntimeServiceId> KMSService<Backend, RuntimeServiceId>
...
     Backend::KeyId: Debug,
-    Backend::SupportedKeyTypes: Debug,
+    Backend::Key: Debug,
     Backend::Settings: Clone,
+    Backend::Error: Debug,
 {
-    async fn handle_kms_message(msg: KMSMessage<Backend>, backend: &mut Backend) {
-        match msg {
+    async fn handle_kms_message(message: KMSMessageBackend<Backend>, backend: &mut Backend) {
+        match message {
```
ntn-x2, line 206: Should we send back an error to the sender of the message instead of panicking if an operation fails in the backend?

nomos-services/key-management-system/src/lib.rs
```diff
@@ -14,50 +17,77 @@ use overwatch::{
...
+
+pub enum KMSMessage<Backend, Data, Signature, PublicKey, KeyError, OperatorError>
 where
     Backend: KMSBackend,
 {
     Register {
         key_id: Backend::KeyId,
-        key_type: Backend::SupportedKeyTypes,
+        key_type: Backend::Key,
         reply_channel: oneshot::Sender<Backend::KeyId>,
```
ntn-x2: Just a nit: why don't we return the public key together with the key ID, instead of having to have a second call to get the public key of the key we just generated?

nomos-services/key-management-system/src/keys/ed25519.rs
```diff
@@ -1,22 +1,49 @@
...
+
+    fn sign(&self, data: &InnerEncoding) -> Result<Self::Signature, Self::Error> {
+        let data_bytes = data.as_bytes();
+        let signature_slice = self.0.sign(data_bytes).to_bytes();
+        let signature_bytes = Bytes::copy_from_slice(&signature_slice);
```
ntn-x2: `.to_bytes()` returns an array and not a slice. Do we need one more step of converting it to a slice and then back into `Bytes`?

nomos-services/key-management-system/src/encodings/encoding_bytes.rs
```diff
@@ -0,0 +1,28 @@
...
+};
+
+type Inner = bytes::Bytes;
+
+/// An encoding of arbitrary bytes.
+#[derive(Clone)]
+pub struct Bytes(Inner);
+
+impl Bytes {
+    pub fn as_bytes(&self) -> &[u8] {
```
ntn-x2: Why not implementing `AsRef<[u8]>` directly? So we can use Rust language capabilities and traits?

nomos-services/key-management-system/src/backend/preload.rs
```diff
@@ -1,35 +1,214 @@
+//! This module contains a simple test implementation of [`KMSBackend`] that
+//! uses an alternate set of keys and encodings, different from the one provided
+//! in this crate.
+//!
+//! It serves as a reference [`KMSBackend`] and demonstrates how encodings, keys
+//! and adapters interact.
+
 #[cfg(test)]
```
ntn-x2, line 8: I think it would be easier to follow if we were to split these tests into separate files.

nomos-services/key-management-system/src/backend/preload.rs
ntn-x2, line 1: Also, why not having the whole module test-gated instead? And then have sub-modules into separate files?

ntn-x2 (commented):
Nothing to add about the features. I left some more nit-picking comments. I guess we will see once we integrate it if something needs to be done differently. But thanks a lot for the effort so far.

Actually, one thing we should consider: generating ZK proofs requires knowledge of the private key. Do you (and @danielSanchezQ) think it would be a good idea to let KMS handle proof generation? Of course it would not be in the scope of this PR, but I think it could do more than simple signing.

nomos-services/key-management-system/src/keys/ed25519.rs
strinnityk: Since we removed encoding wrappers, this has also been fixed.

nomos-services/key-management-system/src/encodings/encoding_bytes.rs
strinnityk: Since we removed encoding wrappers, this has been fixed.

nomos-services/key-management-system/src/backend/preload.rs
strinnityk: They certainly would.
Since I updated old code, I didn't want to do a file refactor at the same time to make it easier to review. Might as well go for it.

nomos-services/key-management-system/src/backend/preload.rs
strinnityk: Tbh I'm not sure what's the intention behind this module. It's declared like this:
```
#[cfg(feature = "preload")]
pub mod preload;
```
It seems it could be used outside testing, but everything inside it points to testing-only.

strinnityk: Update: We'll handle this in a different PR.

danielSanchezQ:
> Actually, one thing we should consider: generating ZK proofs requires knowledge of the private key. Do you (and @danielSanchezQ) think it would be a good idea to let KMS handle proof generation? Of course it would not be in the scope of this PR, but I think it could do more than simple signing.

Yes (partially), the idea is that you send the computation to the kms so sk is only access from it.

nomos-services/key-management-system/src/lib.rs
danielSanchezQ: This operations should not fail. If a panic happens in the kms its because of a human coding error imo.

nomos-services/key-management-system/src/lib.rs
danielSanchezQ: This is a good idea. Should be implemented.

nomos-services/key-management-system/src/backend/preload.rs
danielSanchezQ: This is legacy implementation from the poc. But It is actually the first backend we are gonna use, so just remove the feature gate.

nomos-services/key-management-system/src/backend/preload.rs
danielSanchezQ: Just remove the testing feature from the previous comment. This preload feature can stay.

nomos-services/key-management-system/src/lib.rs
ntn-x2: How? In the case of `KMSMessage::PublicKey`, we panic if the requested public key does not exist, and we panic. That should not be a panic, should it?

nomos-services/key-management-system/src/backend/preload.rs
ntn-x2: Ok, makes sense you don't want to disrupt too much. Although having different test modules in a file which contains nothing else (as far as I can see), is not ideal.

ntn-x2:
> Yes (partially), the idea is that you send the computation to the kms so sk is only access from it.

Is that the reason for `KMSOperator` to exist? Or is that something else? Cause in the case of `KMSOperator` I don't see how that could be achieved, since all we do is take a reference (why mutable) to the secret key, and return an empty tuple.

nomos-services/key-management-system/src/lib.rs
danielSanchezQ: You are right, I was actually thinking on mistmatching key types from the review and mixed it. We should probably return a result from the operation indeed.

danielSanchezQ:
> > Yes (partially), the idea is that you send the computation to the kms so sk is only access from it.
> 
> Is that the reason for `KMSOperator` to exist? Or is that something else? Cause in the case of `KMSOperator` I don't see how that could be achieved, since all we do is take a reference (why mutable) to the secret key, and return an empty tuple.

Yes it is. I would have to check specifics (we can also modify it later on if we need). But basically you can send a future that will be resolved in the kms. So you can attach a return channel to it if you need something back.

nomos-services/key-management-system/src/backend/preload.rs
danielSanchezQ: No, all of them (`cfg(test)`) should be removed. Then we can refactor later into their own modules (separated PR if possible).

danielSanchezQ (approved):
After taking care of the other comments I think this is good to go. Thanks!

danielSanchezQ:
> > Yes it is. I would have to check specifics (we can also modify it later on if we need). But basically you can send a future that will be resolved in the kms. So you can attach a return channel to it if you need something back.
> 
> Ok, then I will wait to see how we could use this, but I am still wondering why do we need a mutable reference to the secret key? That is just not safe, since I can mess up the stored key, did I get that right?

Yeah, it should be just a `&` not `&mut`. I think this is a leftover from an old usecase in where we needed a `&mut`, maybe @youngjoon-lee remembers why. Thats why @strinnityk left it like that.

PR #1710, opened 2025-09-17
feat(da): emmit sampling opinions

pradovic:

This PR is the first PR in the implementation of DA rewards. The scope of this PR is to define sampling behavior opinion events and dummy gather them in the da-sampling service. If this approach looks viable, then we would proceed with emitting events from historical behavior as well, and then implement opinion tracking and aggregation in da-sampling service.

Yes

@pradovic @bacv 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/data-availability/network/src/lib.rs
+++ b/nomos-services/data-availability/network/src/lib.rs
@@ -416,8 +416,7 @@ where
                 Some(_) = opinion_stream.next() => {
                     // todo: aggregate opinions
                     // opinion_tracker.reportOpinion()
-                    // Opinion tracker will also need to subscribe to da-network or
-                    // membership service to track sessions
+                    // opinion tracker also needs to track session change and create results
                 }
             }
         }


nomos-da/network/core/src/protocols/sampling/requests/request_behaviour.rs
```diff
@@ -359,6 +364,24 @@ where
         sample_response: SampleResponse,
         stream: SampleStream,
     ) -> Poll<ToSwarm<<Self as NetworkBehaviour>::ToSwarm, THandlerInEvent<Self>>> {
+        match &sample_response {
+            SampleResponse::Share(_) | SampleResponse::Commitments(_) => {
+                self.opinion_events.push_back(OpinionEvent::Positive {
+                    peer_id,
+                    session_id: None, // None for current session
```
bacv: Could this be SessionNumber instead of an Option? We could get the current session number from the membership, right?

bacv (approved):
This looks great and clean. In pr description you mention that sampling service should do the aggregation, I think it could be done directly in the network service, it knows when to send opinions to the sdp

nomos-da/network/core/src/protocols/sampling/opinions.rs
```diff
@@ -0,0 +1,18 @@
+use libp2p::PeerId;
+use nomos_core::block::SessionNumber;
+
+#[derive(Debug, Clone)]
+pub enum OpinionEvent {
+    Positive {
+        peer_id: PeerId,
+        session_id: Option<SessionNumber>, // None = current session
```
danielSanchezQ: I would rather have a custom enum:
```rust
pub enum Session {
    Current
    Id(SessionNumber)
}
```

Or just force it. always a session number, it should be easy to fill and its much better api than having an option to something we already know.

nomos-da/network/core/src/protocols/sampling/requests/request_behaviour.rs
pradovic: I think at the moment we don't have the session number in regular behavior (we are working on "latest"), but for historic we do have session number.

danielSanchezQ: > I think at the moment we don't have the session number in regular behavior (we are working on "latest"), but for historic we do have session number.

Thats just a matter of enumerating the session stream? or wrap the call of update maybe.

pradovic: The behaviors are emitting events, so the behavior itself needs to know the session number.
We could expand request sampling behavior to be aware of the current session number it is working on (via refresh subnet), but it might be enough to know that it is "Latest" as @danielSanchezQ suggested in the proposed enum:
```
#[derive(Debug, Clone)]
pub enum OpinionEvent {
    Positive { peer_id: PeerId, session: Session },
    Negative { peer_id: PeerId, session: Session },
    Blacklist { peer_id: PeerId, session: Session },
}

#[derive(Debug, Clone)]
pub enum Session {
    Current,
    Id(SessionNumber),
}

```

Does that makes sense? Sorry, maybe I missed a point of enumerating session stream?

pradovic: I think it is important for historic to see which session it is as we should probably discard older session then S-1. But the regular behavior works on the latest session anyway, right? and the da-network will know which session it is when it receives events as it will know current session (it is subscribed to membership).

So all the events it receives for latest will form new_opinions, while session current_session-1 events will be used to form old_opinions. 

On session change, da-network will publish to SDP and move the window to S+1 and S.

In any case, if we end up needing the session number from the regular sampling behavior as well, we can expand it

nomos-da/network/core/src/protocols/sampling/opinions.rs
pradovic: I have added this enum. I can change it if we end up needing sesion number for latest as well.

nomos-da/network/core/src/protocols/sampling/requests/request_behaviour.rs
bacv: > da-network will know which session it is when it receives events

If we make membership handler know the session number, then da network wouldn't need to care about for which session opinions came from, behaviour could use the session number from membership handler and set it instead of None, wdyt?

pradovic: Yes, it would have session number as well but I would argue that it (opinion tracker) would still need to know current session number. Per spec, we need to map latest session to new_opinions, and session -1 to old_opinions vector, right?

So if we return session number instead of latest, we would need to know latest_session_number and then map:
1. latest_session_number to new_opinions
2. latest_session_number - 1 to old_opinions
3. if session < lastest_session_number -1: discard

It can be done both ways I think, as I think opinion tracker will need to know latest session number in both cases in order to discard too old session from historic. It should not be a problem for regular sampling as it should go to new_opinions

pradovic: We can expand MembershipHandler to hold a session number as well, but it directly maps to assignations so it would need to be supported there as well. It should not be a problem, just maybe a separate PR?

danielSanchezQ (approved):

Issue #1699, opened 2025-09-16
Send blocks to the broadcast service during bootstrapping


youngjoon-lee:
The chain service has to send "potentially" finalized blocks to the broadcast service, which will forward them to the SDP service. This is necessary to make the membership service and the DA sampling service to be ready for the blob sampling/validation requests that the chain service will request later during bootstrapping.

We call them "potentially" finalized blocks because no blocks become finalized during bootstrapping if Cryptarchia is not in the Online mode. In other words, LIB never change. 

So, if the chain is reorganized during bootstrapping, the chain service must send all the blocks of the new honest chain starting from the divergence point with the old honest chain.

Issue #1479, opened 2025-07-21
Blob Validation during Chain Bootstrapping


youngjoon-lee:
During [IBD](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81bd899ff05e97d66375) and [ProlongedBootstrapPeriod](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df8162be49e5aa02199378), we should **enable** the blob validation once we enter the DA window. Until then, we should **disable** blob validation.

PR #1711, opened 2025-09-17
feat(chain): stream blocks even during bootstrapping

youngjoon-lee:

Closes https://github.com/logos-co/nomos/issues/1699 (part of https://github.com/logos-co/nomos/issues/1479).

This is a stacked PR based on #1657.

We added the `BroadcastService` in the PR https://github.com/logos-co/nomos/pull/1694, but the chain service was streaming blocks to it only after bootstrapping is done.

But, as explained in issue #1479, the chain service must also stream blocks during bootstrapping. This allows the SDP and membership services to continuously build membership information, which is necessary for the chain service to request DA historic sampling in the final phase of bootstrapping.

You can find the detailed strategy from the code comment.

In later PRs, I will implement requesting historic sampling during bootstrapping.

yes

@youngjoon-lee @zeegomo 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/chain/chain-service/src/blob.rs
+++ b/nomos-services/chain/chain-service/src/blob.rs
@@ -37,7 +37,11 @@ where
     Tx: AuthenticatedMantleTx + Sync,
 {
     async fn validate(&self, block: &Block<Tx>) -> Result<(), Error> {
-        debug!(target = LOG_TARGET, "Validating recent blobs");
+        debug!(
+            target = LOG_TARGET,
+            "Validating blobs in block {:?}",
+            block.header().id()
+        );
         let sampled_blobs = get_sampled_blobs(&self.sampling_relay)
             .await
             .map_err(|_| Error::RelayError)?;
@@ -65,8 +69,12 @@ pub struct SkipBlobValidation;
 
 #[async_trait::async_trait]
 impl<BlobId, Tx> BlobValidation<BlobId, Tx> for SkipBlobValidation {
-    async fn validate(&self, _: &Block<Tx>) -> Result<(), Error> {
-        debug!(target = LOG_TARGET, "Skipping blob validation");
+    async fn validate(&self, block: &Block<Tx>) -> Result<(), Error> {
+        debug!(
+            target = LOG_TARGET,
+            "Skipping blob validation for block {:?}",
+            block.header().id()
+        );
         Ok(())
     }
 }


bacv (approved):
Great api, nicely done.

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -518,6 +517,43 @@ where
...
+    pub fn block_at_security_param(&self) -> Branch<Id> {
+        self.branches()
+            .nth_ancestor(&self.local_chain, self.config.security_param.get().into())
+    }
+
+    pub fn blocks_in_inclusive_range(
+        &self,
+        start: Id,
+        target: Id,
+        limit: NonZeroUsize,
```
zeegomo, line 533: the behavior of this limit param should be documented

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -518,6 +517,43 @@ where
...
+
+            if path.len() > limit.get() {
+                path.pop_back();
+            }
+
+            if current.id == start {
+                return Ok(path.into_iter());
+            }
+
+            if current.parent == current.id {
```
zeegomo, line 550: unlikely we will ever reach the genesis in a real execution, but it's ok, most likely we will hit the error below this

nomos-services/chain/chain-service/src/blob.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+}
+
+#[async_trait::async_trait]
+impl<Tx> BlobValidation<da::BlobId, Tx> for RecentBlobValidation<da::BlobId>
+where
+    Tx: AuthenticatedMantleTx + Sync,
+{
+    async fn validate(&self, block: &Block<Tx>) -> Result<(), Error> {
+        debug!(target = LOG_TARGET, "Validating recent blobs");
```
zeegomo: nit: why not "validating blobs from block {block.id()}" ?

nomos-services/chain/chain-service/src/blob.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+}
+
+/// Skips blob validation.
+pub struct SkipBlobValidation;
+
+#[async_trait::async_trait]
+impl<BlobId, Tx> BlobValidation<BlobId, Tx> for SkipBlobValidation {
+    async fn validate(&self, _: &Block<Tx>) -> Result<(), Error> {
+        debug!(target = LOG_TARGET, "Skipping blob validation");
```
zeegomo: same

nomos-services/chain/chain-service/src/blob.rs
```diff
@@ -0,0 +1,97 @@
...
+#[derive(Debug, Clone, thiserror::Error)]
+pub enum Error {
+    #[error("Block contains invalid blobs")]
+    InvalidBlobs,
+    #[error("Relay error")]
+    RelayError,
+}
+
+/// Retrieves all the blobs that have been sampled by the sampling service.
+pub async fn get_sampled_blobs<BlobId>(
```
zeegomo, line 91: isn't it a bit expensive to return *all* blobs when we're only interested in the ones from this block?

nomos-services/chain/chain-service/src/broadcaster.rs
```diff
@@ -0,0 +1,508 @@
...
+        relay: &BroadcastRelay,
+        lib_broadcaster: &broadcast::Sender<LibUpdate>,
+        pruned_blocks: &PrunedBlocks<HeaderId>,
+    ) -> Result<(), Error> {
+        if !cryptarchia.state().is_online() {
+            return Err(Error::UnexpectedCryptarchiaState(*cryptarchia.state()));
+        }
+
+        let lib = cryptarchia.lib_branch();
+        if self.prev_lib != lib.id() {
```
zeegomo, line 72: what if this skips blocks? I.e, what's guaranteeing `lib` i one of the `prev_lib` direct children? Pruned blocks makes sense in that context

nomos-services/chain/chain-service/src/broadcaster.rs
```diff
@@ -0,0 +1,508 @@
...
+        cryptarchia: &Cryptarchia<HeaderId>,
+        relay: &BroadcastRelay,
+        // Since LIB never changes in Bootstrapping state,
+        // we don't need to send any LIB update through this channel.
+        _lib_broadcaster: &broadcast::Sender<LibUpdate>,
+        _pruned_blocks: &PrunedBlocks<HeaderId>,
+    ) -> Result<(), Error> {
+        if !cryptarchia.state().is_bootstrapping() {
+            return Err(Error::UnexpectedCryptarchiaState(*cryptarchia.state()));
+        }
```
zeegomo, line 135: what happens when we transition from bootstrapping to online?

nomos-services/chain/chain-service/src/broadcaster.rs
```diff
@@ -0,0 +1,508 @@
...
+}
+
+#[async_trait::async_trait]
+impl FinalizedBlockBroadcaster for BootstrappingFinalizedBlockBroadcaster {
+    /// Broadcasts all blocks between the previous security block (exclusive)
+    /// and the security block (inclusive).
+    ///
+    /// If the chain has been reorganized (i.e. the current security block is
+    /// not a descedant of the previous one), all blocks between the common
+    /// ancestor (exclusive) and the current security block (inclusive).
```
zeegomo, line 123: the whole point of bootstrapping is to avoid finalization, maybe we can put a comment here on why we're doing this?

nomos-services/chain/chain-service/src/blob.rs
youngjoon-lee: Yes, it is expensive. But, I didn't want to change the DA sampling API in this PR. I copied this `get_sampled_blobs` function from `chain-service/src/lib.rs` to here without changing anything, just to make this PR easy to review. But, yes, I will suggest a change for the DA API in a new PR.

nomos-services/chain/chain-service/src/broadcaster.rs
youngjoon-lee: Actually, it shouldn't happen. This trait should be used as below:
```rust
let broadcaster = new_finalized_block_broadcaster(&cryptarchia.consensus);
let (cryptarchia, _) = cryptarchia.try_apply_header(header)?;
broadcaster.broadcast(&cryptarchia.consensus, ...);
```
In this flow, the cryptarchia is never switched to online.

It is switched only when we manually call `cryptarchia.online()` (which is called only after ProlongedBootstrapPeriod has passed).

So, returning the error here is a safeguard.

nomos-services/chain/chain-service/src/broadcaster.rs
youngjoon-lee: If this trait is used as I explained below https://github.com/logos-co/nomos/pull/1711#discussion_r2371765968, `lib` will be always a direct child of `prev_lib`. But, yes, I agree that the API of this trait is not so safe and doesn't handle misuses. Let me think how to improve.

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: Added a doc comment 4bb8aa6099b006e05793f938d4e8fa1376b1d92e

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: Yes, it's unlikely that we will reach the genesis, but I added this safety check just in case.

nomos-services/chain/chain-service/src/blob.rs
youngjoon-lee: Yeah, I updated the comment 1d95440dca3dbdbc076ac5676a613d58c435c3b8

nomos-services/chain/chain-service/src/blob.rs
youngjoon-lee: Fixed in 1d95440dca3dbdbc076ac5676a613d58c435c3b8

PR #1715, opened 2025-09-18
refactor(cryptarchia): define Length type

youngjoon-lee:
## 1. What does this PR implement?

We've been using `u64` to represent chain length, height, and depth.
This PR defines the `Length(u64)` type for them.

That's useful since now we deal with it in many places outside of the cryptarchia engine (such as in `chain/broadcast_service` and, in the near future, chain sync).

## 2. Does the code have enough context to be clearly understood?

Yes

## 3. Who are the specification authors and who is accountable for this PR?
@zeegomo @youngjoon-lee 

## 4. Is the specification accurate and complete?
n/a

## 5. Does the implementation introduce changes in the specification?
n/a

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [x] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [x] 2. Description added.
* [x] 3. Context and links to Specification document(s) added.
* [x] 4. Main contact(s) (developers and specification authors) added
* [x] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 6. Link PR to a specific milestone.

PR #1716, opened 2025-09-18
feat(chainsync): add height to the TipResponse

youngjoon-lee:

(stacked on PR #1715 )

Added `height` to the `TipResponse`.
It's necessary for IBD in the near future, since the node must enable blob validation during IBD once it enters the DA window (e.g. 8640 blocks). In other words, it should enable the blob validation once it reaches the 8640-deep block from the peer's tip. (See details in https://github.com/logos-co/nomos/issues/1479). This PR is a preparation for it.

yes

@youngjoon-lee @zeegomo 

no

we should add `height` to the spec.

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/consensus/cryptarchia-sync/src/libp2p/behaviour.rs
+++ b/consensus/cryptarchia-sync/src/libp2p/behaviour.rs
@@ -607,12 +607,13 @@ mod tests {
 
         tokio::spawn(async move { downloader_swarm.loop_on_next().await });
 
-        let Tip { tip, slot } = receiver.await.unwrap().unwrap() else {
+        let Tip { tip, slot, height } = receiver.await.unwrap().unwrap() else {
             panic!("Expected a tip response");
         };
 
         assert_eq!(tip, HeaderId::from([0; 32]));
         assert_eq!(slot, Slot::from(0));
+        assert_eq!(height, 0);
     }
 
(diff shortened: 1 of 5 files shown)


consensus/cryptarchia-sync/src/libp2p/behaviour.rs
```diff
@@ -511,7 +511,7 @@ impl NetworkBehaviour for Behaviour {
 mod tests {
     use std::{collections::HashSet, iter, time::Duration};
 
-    use cryptarchia_engine::Slot;
+    use cryptarchia_engine::{Length, Slot};
```
andrussal: I see that `Length` comes from engine. I wonder if we should later rename it to `Height`(or the other way)?

youngjoon-lee: Yeah, that's an on-going discussion https://github.com/logos-co/nomos/pull/1715#issuecomment-3305695728. I also think that `Height` sounds better.

danielSanchezQ: `Height` ➕ 1

danielSanchezQ (approved):

Issue #1675, opened 2025-09-10
Enable blob validation once entering the DA window


youngjoon-lee:
1. Update `ChainService::process_block`. This is used for both bootstrapping and online scenario. (A reference discussion: [link](https://discord.com/channels/1111286067413405788/1427978612266500156/1428005927654658100))
  ```rust
  fn process_block(cryptarchia: Cryptarchia, block: Block) -> Result<Cryptarchia, Error> {
    let cryptarchia = cryptarchia.try_apply_header(block.header)?;

    if should_validate_blob_availability(block.header.slot, current_slot) {
      validate_blobs_in_block(block)?;
    }
  
    ...
  
    Ok(cryptarchia)
  }
  
  fn should_validate_blob_availability(block_slot, current_slot) -> bool {
    (current_slot - block_slot) < (DA_VALIDATION_WINDOW_IN_SLOTS)  // DA_VALIDATION_WINDOW_IN_SLOTS should include a safety margin
  }

  fn validate_blobs_in_block(block) -> Result<(), Error> {
    // Sampling blobs for checking if they are available in DA nodes
  }
  ```

2. Define `DA_VALIDATION_WINDOW_IN_SLOTS`.
  - If this value is too large, blob validation will frequently fail because it will attempt to sample blobs that have already been pruned by DA nodes. Therefore, it’s important to keep this value small yet reasonable.
  - This should be shorter than `finality_window` (= `k/f` slots in Cryptarchia).
    - The reason is explained in [this discussion](https://discord.com/channels/1111286067413405788/1427978612266500156/1428009401410785330).
    - The reason why we don't use `3k/f` is explained in [this discussion](https://discord.com/channels/1111286067413405788/1427978612266500156/1429755254034923551).
  - This should be shorter than the period that is incentivized by DA rewarding (i.e. 2 DA sessions). 
  - For now, I will calculate `DA_VALIDATION_WINDOW_IN_SLOTS = finality_window / 2`. We can modify this once we agree on the reasonable value.

3. Special cases
  - Still we don't need to validate blobs when recovering the chain by loading blocks from disk. We may have to handle this case differently, since the 1 above cannot handle this case.
  - We should use historic sampling during IBD because the sampling service hasn't sampled any blobs yet.

PR #1721, opened 2025-09-19
DRAFT: feat(chain): check remaining blocks in IBD to enable blob validation

youngjoon-lee:

Partially resolves https://github.com/logos-co/nomos/issues/1675.

We starts IBD with disabling blob validation.
At some point during IBD, we should enable the blob validation.
For now, we agreed on enabling it once entering the DA window (which is not defined yet). 
So, this PR defines a parameter `ibd.last_blocks_with_blob_validation`.  If there are remaining blocks to the target less than the parameter, IBD enables blob validation.

This PR doesn't yet use the historical blob sampling API after enabling blob validation. Still it skips blob validation. I will change it once all other components (services) are ready. This PR just prepares the foundation.

Also, I promise I will open a refactoring PR soon to prevent making the chain service more verbose. I have an idea, but didn't want to make this PR hard to review.

yes

@youngjoon-lee 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nodes/nomos-executor/config.yaml
+++ b/nodes/nomos-executor/config.yaml
@@ -252,7 +252,7 @@ cryptarchia:
       state_recording_interval: [ 60, 0 ]
     ibd:
       peers: [ ]
-      final_blocks_with_blob_validation: 8640
+      last_blocks_with_blob_validation: 8640
   sync:
     orphan:
       max_orphan_cache_size: 5
--- a/nomos-services/chain/chain-service/src/bootstrap/config.rs
+++ b/nomos-services/chain/chain-service/src/bootstrap/config.rs
@@ -46,7 +46,7 @@ where
     pub delay_before_new_download: Duration,
     /// Threshold in number of remaining blocks to the target height
     /// at which blob validation becomes enabled during IBD.
-    pub final_blocks_with_blob_validation: Length,
+    pub last_blocks_with_blob_validation: Length,
 }
 
 const fn default_offline_grace_period() -> Duration {
--- a/nodes/nomos-node/config.yaml
+++ b/nodes/nomos-node/config.yaml
@@ -241,7 +241,7 @@ cryptarchia:
       state_recording_interval: [ 60, 0 ]
     ibd:
       peers: [ ]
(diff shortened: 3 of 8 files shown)


nodes/nomos-executor/config.yaml
```diff
@@ -252,6 +252,7 @@ cryptarchia:
       state_recording_interval: [ 60, 0 ]
     ibd:
       peers: [ ]
+      last_blocks_with_blob_validation: 8640
```
zeegomo, line 255: how should we interpret this value and how would one select it?

nomos-services/chain/chain-service/src/bootstrap/config.rs
```diff
@@ -43,6 +44,9 @@ where
     /// when no download is needed at the moment from a peer.
     #[serde(default = "default_delay_before_new_download")]
     pub delay_before_new_download: Duration,
+    /// Threshold in number of remaining blocks to the target height
+    /// at which blob validation becomes enabled during IBD.
+    pub last_blocks_with_blob_validation: Length,
```
zeegomo, line 49: shouldn't this be *first* block with validation?

nodes/nomos-executor/config.yaml
youngjoon-lee: This value must be defined in the DA spec, because this means the period how long DA nodes will keep blobs. But, it is not determined yet. I added this parameter here because we anyway need to for IBD.

nomos-services/chain/chain-service/src/bootstrap/config.rs
youngjoon-lee: I intended this parameter to represent the number of "remaining" blocks. So, I think that "last" makes more sense. 

If there are N blocks in total that we have to catch up during IBD, and if there are `last_blocks = 8000`,
we're going to disable blob validation while we process `N - 8000` blocks. After that, we will enable blob validation for next 8000 blocks. 

Please let me know if this naming doesn't sound good. I will try to think about other names (or updating comments).

PR #1722, opened 2025-09-19
chore: update Rust to 1.90

ntn-x2:
Fixes https://github.com/logos-co/nomos/issues/1720.


danielSanchezQ (approved):

ntn-x2:
@andrussal there were some message types in Cryptarchia that were only used when the `libp2p` feature was enabled, so I moved those inside the `libp2p` module to keep `rustc` happy. Can you double check this is ok? It does not trigger any test issues nor any compilation errors, so I assume it's ok, but a quick review from you would be appreciated, thanks!

PR #1723, opened 2025-09-19
feat(logos-core): Nomos C bindings

danielSanchezQ:

The intention is to provide an interface to the nomos node for the logos-core project (or any other C based ffi project).
As for now it just provides a way of initializing the node, run it and stopping it. We will decide on the offered API later on and add the functionality.  There are things to solve so this works on logos-core, for example how to distribute the needed binaries, but I guess is a different problem than just exposing the node.

I hope really hope so. Although there is a lot of ffi context needed to fully understand it.

N/A

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -135,8 +135,10 @@ reqwest           = { default-features = false, version = "0.12" }
 risc0-zkvm        = { default-features = false, version = "2.3.2" }
 serde_json        = { default-features = false, version = "1.0" }
 serde_with        = { default-features = false, version = "3.14.0" }
+serde_yaml        = { version = "0.9.33" }
 tempfile          = { default-features = false, version = "3" }
 thiserror         = { default-features = false, version = "2.0" }
+tokio             = { default-features = false, version = "1" }
 tracing           = { default-features = false, version = "0.1" }
 utoipa            = { default-features = false, version = "4.0" }
 utoipa-swagger-ui = { default-features = false, version = "7.0" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ members = [
   "nomos-blend/network",
   "nomos-blend/scheduling",
   "nomos-bundler",
+  "nomos-c",
   "nomos-cli",
   "nomos-core/chain-defs",
   "nomos-da/kzgrs",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -135,7 +135,7 @@ reqwest           = { default-features = false, version = "0.12" }
 risc0-zkvm        = { default-features = false, version = "2.3.2" }
 serde_json        = { default-features = false, version = "1.0" }
 serde_with        = { default-features = false, version = "3.14.0" }
-serde_yaml        = { version = "0.9.33" }
+serde_yaml        = { default-features = false, version = "0.9.33" }
 tempfile          = { default-features = false, version = "3" }
 thiserror         = { default-features = false, version = "2.0" }
 tokio             = { default-features = false, version = "1" }
--- a/nomos-c/Cargo.toml
+++ b/nomos-c/Cargo.toml
@@ -5,7 +5,7 @@ name              = "nomos-c"
 version           = "0.1.0"
 
 [dependencies]
-nomos-node = { default-features = false, workspace = true, features = ["tracing"] }
+nomos-node = { default-features = true, workspace = true }
 overwatch  = { workspace = true }
 serde_yaml = { default-features = false, workspace = true }
 tokio      = { default-features = false, workspace = true, features = ["rt-multi-thread"] }
--- a/nomos-c/Cargo.toml
+++ b/nomos-c/Cargo.toml
@@ -1,14 +1,14 @@
 [package]
-edition           = "2024"
+edition = "2024"
 license.workspace = true
-name              = "nomos-c"
-version           = "0.1.0"
+name = "nomos-c"
+version = "0.1.0"
 
 [dependencies]
 nomos-node = { workspace = true, features = ["tracing"] }
-overwatch  = { workspace = true }
+overwatch = { workspace = true }
 serde_yaml = { workspace = true }
-tokio      = { default-features = false, workspace = true, features = ["rt-multi-thread"] }
+tokio = { default-features = false, workspace = true, features = ["rt-multi-thread"] }
 
 [build-dependencies]
 cbindgen = "0.29"
@@ -18,3 +18,4 @@ crate-type = ["staticlib"]
 
 [lints]
 workspace = true
+
--- a/nomos-c/Cargo.toml
+++ b/nomos-c/Cargo.toml
@@ -14,7 +14,7 @@ tokio      = { default-features = false, workspace = true, features = ["rt-multi
 cbindgen = "0.29"
 
 [lib]
-crate-type = ["cdylib"]
+crate-type = ["staticlib"]
 
 [lints]
 workspace = true
--- a/nomos-c/Cargo.toml
+++ b/nomos-c/Cargo.toml
@@ -5,7 +5,7 @@ name              = "nomos-c"
 version           = "0.1.0"
 
 [dependencies]
-nomos-node = { default-features = false, workspace = true, features = ["tracing"] }
+nomos-node = { default-features = true, workspace = true, features = ["tracing"] }
 overwatch  = { workspace = true }
 serde_yaml = { workspace = true }
 tokio      = { default-features = false, workspace = true, features = ["rt-multi-thread"] }
--- a/nomos-c/Cargo.toml
+++ b/nomos-c/Cargo.toml
@@ -19,3 +19,6 @@ name       = "libnomos"
 
 [lints]
 workspace = true
+
+[package.metadata.cargo-machete]
+ignored = ["cbindgen"] # machete does not recognize this for build dep and complains.
--- a/nomos-c/Cargo.toml
+++ b/nomos-c/Cargo.toml
@@ -1,21 +1,21 @@
 [package]
-edition = "2024"
+edition           = "2024"
 license.workspace = true
-name = "nomos-c"
-version = "0.1.0"
+name              = "nomos-c"
+version           = "0.1.0"
 
 [dependencies]
-nomos-node = { workspace = true, features = ["tracing"] }
-overwatch = { workspace = true }
+nomos-node = { default-features = false, workspace = true, features = ["tracing"] }
+overwatch  = { workspace = true }
 serde_yaml = { workspace = true }
-tokio = { default-features = false, workspace = true, features = ["rt-multi-thread"] }
+tokio      = { default-features = false, workspace = true, features = ["rt-multi-thread"] }
 
 [build-dependencies]
 cbindgen = "0.29"
 
 [lib]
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -132,8 +133,10 @@ reqwest           = { default-features = false, version = "0.12" }
 risc0-zkvm        = { default-features = false, version = "2.3.2" }
 serde_json        = { default-features = false, version = "1.0" }
 serde_with        = { default-features = false, version = "3.14.0" }
+serde_yaml        = { version = "0.9.33" }
```
ntn-x2: No default features?

nomos-c/cbindgen.toml
```diff
@@ -0,0 +1,3 @@
+cpp_compat  = false
```
ntn-x2: Why not C++ as well, since it comes for free?

nomos-c/Cargo.toml
ntn-x2, line 1: I am not sure including this in the workspace is a good idea. This will run `build.rs` whenever we check out a different branch. My machine started working a lot after checking out the PR branch. So I think this should be a separate, isolated crate. Have you thought about it? In the end, it only depends on `nomos-node` as an actual workspace dependency.

nomos-c/src/lib.rs
```diff
@@ -0,0 +1,131 @@
+// Safety: Well, this is gonna be a shit show of unsafe calls...
+#![allow(clippy::allow_attributes_without_reason)]
```
ntn-x2: Why do we not want to enforce an explanation for `allow` and `expect` blocks?

nomos-c/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+}
+
+fn initialize_nomos_node(config_path: *const c_char) -> Result<NomosNode, NomosNodeErrorCode> {
+    let must_blend_service_group_start = true;
+    let must_da_service_group_start = true;
+    let config_path = unsafe { std::ffi::CStr::from_ptr(config_path) }
+        .to_str()
+        .map_err(|e| {
+            eprintln!("Could not convert config path to string: {}", e);
```
ntn-x2:
```suggestion
            eprintln!("Could not convert config path to string: {e}");
```

nomos-c/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+            NomosNodeErrorCode::CouldNotInitialize
+        })?;
+    let config =
+        serde_yaml::from_reader::<_, Config>(std::fs::File::open(config_path).map_err(|e| {
+            eprintln!("Could not open config file: {}", e);
```
ntn-x2:
```suggestion
            eprintln!("Could not open config file: {e}");
```

nomos-c/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+    let must_blend_service_group_start = true;
+    let must_da_service_group_start = true;
+    let config_path = unsafe { std::ffi::CStr::from_ptr(config_path) }
+        .to_str()
+        .map_err(|e| {
+            eprintln!("Could not convert config path to string: {}", e);
+            NomosNodeErrorCode::CouldNotInitialize
+        })?;
+    let config =
+        serde_yaml::from_reader::<_, Config>(std::fs::File::open(config_path).map_err(|e| {
```
ntn-x2: I think it would be better to un-nest this error mapping into separate statements for a clearer flow.

nomos-c/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+            eprintln!("Could not open config file: {}", e);
+            NomosNodeErrorCode::CouldNotInitialize
+        })?)
+        .map_err(|e| {
+            eprintln!("Could not parse config file: {}", e);
```
ntn-x2:
```suggestion
            eprintln!("Could not parse config file: {e}");
```

nomos-c/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+edition = "2024"
+license.workspace = true
+name = "nomos-c"
+version = "0.1.0"
+
+[dependencies]
+nomos-node = { workspace = true, features = ["tracing"] }
```
ntn-x2: I think the binary bindings should probably use the default features of the binary? In which case we should probably just use `features = ["default"]`, because we disable them in the workspace definition.
```suggestion
nomos-node = { workspace = true, features = ["default"] }
```

But, if we go for the route of taking this crate out of the workspace, then we can simply depend on the binary with its default features enabled.

nomos-c/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+overwatch = { workspace = true }
+serde_yaml = { workspace = true }
+tokio = { default-features = false, workspace = true, features = ["rt-multi-thread"] }
```
ntn-x2: I think all workspace dependencies have disabled default features, so we do not need to specify it here.

nomos-c/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+            membership: config.membership,
+        },
+        Some(handle.clone()),
+    )
+    .map_err(|e| {
+        eprintln!("Could not initialize overwatch: {}", e);
+        NomosNodeErrorCode::CouldNotInitialize
+    })?;
+
+    let app_handel = app.handle();
```
ntn-x2:
```suggestion
    let app_handle = app.handle();
```

nomos-c/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    rt.block_on(async {
+        let services_to_start = get_services_to_start(
+            &app,
+            must_blend_service_group_start,
+            must_da_service_group_start,
+        )
+        .await
+        .map_err(|e| {
+            eprintln!("Could not get services to start: {}", e);
```
ntn-x2:
```suggestion
            eprintln!("Could not get services to start: {e}");
```

nomos-c/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+            NomosNodeErrorCode::CouldNotInitialize
+        })?;
+        app_handel
+            .start_service_sequence(services_to_start)
+            .await
+            .map_err(|e| {
+                eprintln!("Could not start services: {}", e);
```
ntn-x2:
```suggestion
                eprintln!("Could not start services: {e}");
```

nomos-c/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+            time: config.time,
+            storage: config.storage,
+            system_sig: (),
+            sdp: (),
+            membership: config.membership,
+        },
+        Some(handle.clone()),
+    )
+    .map_err(|e| {
+        eprintln!("Could not initialize overwatch: {}", e);
```
ntn-x2:
```suggestion
        eprintln!("Could not initialize overwatch: {e}");
```

nomos-c/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+        }
+        Err(error_code) => InitializedNomosNodeResult {
+            nomos_node: core::ptr::null_mut(),
+            error_code,
+        },
+    }
+}
+
+fn initialize_nomos_node(config_path: *const c_char) -> Result<NomosNode, NomosNodeErrorCode> {
+    let must_blend_service_group_start = true;
```
ntn-x2, line 47: This would be changed with a CLI flag, ideally, so perhaps they should be inputs to the `initialize_nomos_node` function?

nomos-c/src/api.rs
```diff
@@ -0,0 +1,86 @@
...
+            // Box the complex types and convert to opaque pointers
+            overwatch: Box::into_raw(Box::new(overwatch)).cast::<c_void>(),
+            runtime: Box::into_raw(Box::new(runtime)).cast::<c_void>(),
+        }
+    }
+
+    // Helper methods to safely access the inner types
+    #[must_use]
+    const fn get_overwatch_handle(&self) -> &OverwatchHandle<RuntimeServiceId> {
+        unsafe { self.overwatch.cast::<NomosOverwatch>().as_ref().unwrap() }.handle()
```
ntn-x2: Should we use `expect`s here? So something will be printed to stderr that we can work with?

nomos-c/src/api.rs
```diff
@@ -0,0 +1,86 @@
...
+            let _ = unsafe { Box::from_raw(self.runtime.cast::<Runtime>()) };
+        }
+    }
+}
+
+#[unsafe(no_mangle)]
+/// # Safety
+///
+/// The caller must ensure that:
+/// - `node` is either null or a valid pointer to a `NomosNode` instance
```
ntn-x2: Does the caller need to ensure that the pointer is `null`? I would perhaps remove that and simply say they need to ensure it points to a valid instance. Then if it happens to be `null`, we can deal with that.

ntn-x2 (commented):
I can't really comment on FFI approach without first spending some time studying the depths of FFI, so I provided some feedback on everything that is not strictly FFI-related. Hope this helps.

nomos-c/src/lib.rs
danielSanchezQ: Its a leftover, it was just meant for the allow below. I fixed it.

nomos-c/src/lib.rs
danielSanchezQ: This will have to change shortly when we do it dynamically in the node (?). So probably not worth it to expose it.

nomos-c/Cargo.toml
danielSanchezQ: I think we can keep it but I can change the build.rs so it only triggers if the crate changed. Otherwise I'll remove from workspace. wdyt?

nomos-c/cbindgen.toml
danielSanchezQ: Another leftover. I was trying to change the generated file. But somehow is not working as expected.

nomos-c/Cargo.toml
ntn-x2: That would already be an improvement, yes. I would also prefer to keep it inside the workspace.

nomos-c/src/lib.rs
ntn-x2: Ok makes sense. Then perhaps we should move them waaay below just before they are used. At least that is what I would prefer, but if you prefer to keep them at the top, let's leave it as it is.

danielSanchezQ: I prefer to have them on top just so its obvious that they will be removed at some point. I'll add a TODO to it as well.

nomos-c/src/api.rs
danielSanchezQ: Brain fart, it should be that is just a valid pointer (hence not null).

nomos-c/Cargo.toml
danielSanchezQ: > That would already be an improvement, yes. I would also prefer to keep it inside the workspace.

Should be fine now.

nomos-c/src/lib.rs
ntn-x2: I think this is still a typo.

strinnityk (approved):

PR #1730, opened 2025-09-22
feat(chain): introduce `subscribe` API for winning slot PoL info

ntn-x2:

This PR introduces a new `subscribe` API for other services to get winning PoL slot info from chain service whenever a new winning slot is computed. This API will be used by Blend, to start pre-computing all the leadership quota proofs as soon as the first winning slot for an epoch is made available by the chain service. I also refactored the leader public and private inputs to take the user-friendlier `InputsData`, so that the same is sent over the new API, while they are converted to their lower-level counterparts only upon actual proof generation.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/zk/proofs/pol/src/inputs.rs
+++ b/zk/proofs/pol/src/inputs.rs
@@ -17,14 +17,36 @@ pub struct PolWitnessInputs {
 
 impl PolWitnessInputs {
     #[must_use]
-    pub fn from_chain_and_wallet_data(
+    pub const fn from_chain_and_wallet_inputs(
+        chain: PolChainInputs,
+        wallet: PolWalletInputs,
+    ) -> Self {
+        Self { wallet, chain }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct PolWitnessInputsData {
+    pub wallet: PolWalletInputsData,
+    pub chain: PolChainInputsData,
+}
+
+impl From<PolWitnessInputsData> for PolWitnessInputs {
+    fn from(data: PolWitnessInputsData) -> Self {
+        Self {
+            wallet: data.wallet.into(),
+            chain: data.chain.into(),
+        }
+    }
+}
+
+impl PolWitnessInputsData {
+    #[must_use]
+    pub const fn from_chain_and_wallet_data(
         chain: PolChainInputsData,
         wallet: PolWalletInputsData,
     ) -> Self {
-        Self {
-            wallet: wallet.into(),
-            chain: chain.into(),
-        }
+        Self { wallet, chain }
     }
 }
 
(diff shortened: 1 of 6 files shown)


zk/proofs/pol/src/inputs.rs
```diff
@@ -17,17 +17,39 @@ pub struct PolWitnessInputs {
...
+impl From<PolWitnessInputsData> for PolWitnessInputs {
+    fn from(data: PolWitnessInputsData) -> Self {
         Self {
-            wallet: wallet.into(),
-            chain: chain.into(),
+            wallet: data.wallet.into(),
+            chain: data.chain.into(),
         }
     }
 }
```
danielSanchezQ, line 41: Isn't this too many `into`'s? 😅  
What I mean is that there is no need for the man in the middle wrapper. Although I see the point so we do not use two types in the other layers.

zk/proofs/pol/src/lib.rs
```diff
@@ -303,9 +303,10 @@ mod tests {
             .collect(),
             starting_slot: 16,
         };
-        let witness_inputs = PolWitnessInputs::from_chain_and_wallet_data(chain_data, wallet_data);
+        let witness_inputs =
+            PolWitnessInputsData::from_chain_and_wallet_data(chain_data, wallet_data);
 
-        let (proof, inputs) = prove(&witness_inputs).unwrap();
+        let (proof, inputs) = prove(&witness_inputs.into()).unwrap();
```
danielSanchezQ, line 309: yeah, I think I like it better the old style.

danielSanchezQ (commented):
Seems simple enough.
 Should we add the slot/epoch information for each as well? or we trust is increasing? What happens if we expect an specific one and then we have some offset in the services?

zk/proofs/pol/src/inputs.rs
ntn-x2: Yeah the point is that in Blend we would need the higher-level types, which we can then transform into the lower-level ones. If we stream directly the lower-level ones, we need to then convert in Blend.

ntn-x2: I am sure we can also implement some ref-to-ref conversion, but I thought it's not a big deal to do this for the time being.

ntn-x2: Also, `LeaderPublic` includes higher-level components (i.e., `u64` and `Fr`), while `LeaderPrivate` includes lower-level components (i.e., `Groth16Input`), so I thought about making them consistent.

ntn-x2:
This solution won't work with the current architecture. I will open a new PR that will superseed this one.

PR #1736, opened 2025-09-24
feat(da-network): opinion aggregation

pradovic:

This PR introduces forming DA samping opinions in da-network, on session change. The next step would be to send these opinions to sdp via adapter, so it the activity proof can be generated.

Yes

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/data-availability/network/src/opinion_aggregator.rs
+++ b/nomos-services/data-availability/network/src/opinion_aggregator.rs
@@ -1,0 +1,157 @@
+use std::collections::{HashMap, HashSet};
+
+use libp2p::PeerId;
+use nomos_core::block::SessionNumber;
+use nomos_da_network_core::protocols::sampling::opinions::{Opinion, OpinionEvent};
+use subnetworks_assignations::MembershipHandler;
+
+const OPINION_THRESHOLD: f32 = 0.9;
+
+pub struct OpinionAggregator<Membership>
+where
+    Membership: MembershipHandler<Id = PeerId>,
+{
+    // Opinion counters for current session (new blocks)
+    positive_opinions: HashMap<PeerId, u32>,
+    negative_opinions: HashMap<PeerId, u32>,
+    blacklist: HashSet<PeerId>,
+
+    // Opinion counters for previous session (old blocks)
+    old_positive_opinions: HashMap<PeerId, u32>,
+    old_negative_opinions: HashMap<PeerId, u32>,
+    old_blacklist: HashSet<PeerId>,
+
+    // Memberships contain both the peer lists and session IDs
+    current_membership: Option<Membership>,
+    previous_membership: Option<Membership>,
+}
+
+impl<Membership> OpinionAggregator<Membership>
+where
+    Membership: MembershipHandler<Id = PeerId>,
+{
+    pub fn new() -> Self {
+        Self {
+            positive_opinions: HashMap::new(),
+            negative_opinions: HashMap::new(),
+            blacklist: HashSet::new(),
+            old_positive_opinions: HashMap::new(),
+            old_negative_opinions: HashMap::new(),
+            old_blacklist: HashSet::new(),
+            current_membership: None,
+            previous_membership: None,
+        }
+    }
+
+    pub fn record_opinion(&mut self, event: OpinionEvent) {
+        for opinion in event.opinions {
+            match opinion {
+                Opinion::Positive {
+                    peer_id,
+                    session_id,
+                } => {
+                    self.handle_positive(peer_id, session_id);
+                }
+                Opinion::Negative {
+                    peer_id,
+                    session_id,
+                } => {
+                    self.handle_negative(peer_id, session_id);
+                }
+                Opinion::Blacklist {
+                    peer_id,
+                    session_id,
+                } => {
+                    self.handle_blacklist(peer_id, session_id);
+                }
+            }
+        }
+    }
+
+    fn handle_positive(&mut self, peer_id: PeerId, session_id: SessionNumber) {
+        if let Some(ref current) = self.current_membership
+            && current.session_id() == session_id
+        {
+            if let Some(count) = self.positive_opinions.get_mut(&peer_id) {
+                *count += 1;
+            }
+            return;
+        }
+        if let Some(ref previous) = self.previous_membership
+            && previous.session_id() == session_id
+            && let Some(count) = self.old_positive_opinions.get_mut(&peer_id)
+        {
+            *count += 1;
+        }
+    }
+
+    fn handle_negative(&mut self, peer_id: PeerId, session_id: SessionNumber) {
+        if let Some(ref current) = self.current_membership
+            && current.session_id() == session_id
+        {
+            if let Some(count) = self.negative_opinions.get_mut(&peer_id) {
+                *count += 1;
+            }
+            return;
+        }
+        if let Some(ref previous) = self.previous_membership
+            && previous.session_id() == session_id
+            && let Some(count) = self.old_negative_opinions.get_mut(&peer_id)
+        {
+            *count += 1;
+        }
+    }
+
+    fn handle_blacklist(&mut self, peer_id: PeerId, session_id: SessionNumber) {
+        if let Some(ref current) = self.current_membership
+            && current.session_id() == session_id
+        {
+            if let Some(count) = self.positive_opinions.get_mut(&peer_id) {
+                *count = 0;
+            }
+            self.blacklist.insert(peer_id);
+            return;
+        }
+        if let Some(ref previous) = self.previous_membership
+            && previous.session_id() == session_id
+        {
+            if let Some(count) = self.old_positive_opinions.get_mut(&peer_id) {
+                *count = 0;
+            }
+            self.old_blacklist.insert(peer_id);
+        }
+    }
+
+    pub fn on_session_change(&mut self, new_membership: Membership) {
+        // Generate activity proof here if we have both memberships (before clearing)
+        if self.current_membership.is_some() && self.previous_membership.is_some() {
+            // This would be session s+1, time to generate proof for session s
+            // TODO: generate_activity_proof() and send it
+        }
+
+        self.previous_membership = self.current_membership.take();
+
+        self.positive_opinions.clear();
+        self.negative_opinions.clear();
+        self.blacklist.clear();
+        self.old_positive_opinions.clear();
+        self.old_negative_opinions.clear();
+        self.old_blacklist.clear();
+
+        // Pre-populate opinion maps with zeros for new membership so the vectors fit
+        for peer_id in new_membership.members() {
+            self.positive_opinions.insert(peer_id, 0);
+            self.negative_opinions.insert(peer_id, 0);
+        }
+
+        // Pre-populate for previous membership (if it exists) so the vectors fit
+        if let Some(ref membership) = self.previous_membership {
+            for peer_id in membership.members() {
+                self.old_positive_opinions.insert(peer_id, 0);
+                self.old_negative_opinions.insert(peer_id, 0);
+            }
+        }
+
+        self.current_membership = Some(new_membership);
+    }
+}
--- a/nomos-services/data-availability/network/src/opinion_aggregator.rs
+++ b/nomos-services/data-availability/network/src/opinion_aggregator.rs
@@ -11,27 +11,37 @@ pub struct OpinionAggregator<Membership>
 where
     Membership: MembershipHandler<Id = PeerId>,
 {
-    // Opinion counters for current session (new blocks)
+    local_peer_id: PeerId,
     positive_opinions: HashMap<PeerId, u32>,
     negative_opinions: HashMap<PeerId, u32>,
     blacklist: HashSet<PeerId>,
 
-    // Opinion counters for previous session (old blocks)
     old_positive_opinions: HashMap<PeerId, u32>,
     old_negative_opinions: HashMap<PeerId, u32>,
     old_blacklist: HashSet<PeerId>,
 
-    // Memberships contain both the peer lists and session IDs
     current_membership: Option<Membership>,
     previous_membership: Option<Membership>,
+}
+
+#[derive(Debug)]
+#[expect(
+    dead_code,
+    reason = "Will be used when SDP adapter integration is complete"
+)]
+pub struct Opinions {
+    pub session_id: SessionNumber,
+    pub new_opinions: Vec<bool>,
+    pub old_opinions: Vec<bool>,
 }
 
 impl<Membership> OpinionAggregator<Membership>
 where
     Membership: MembershipHandler<Id = PeerId>,
 {
-    pub fn new() -> Self {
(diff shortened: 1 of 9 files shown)


nomos-services/data-availability/network/src/opinion_aggregator.rs
```diff
@@ -0,0 +1,224 @@
+use std::collections::{HashMap, HashSet};
+
+use libp2p::PeerId;
+use nomos_core::block::SessionNumber;
+use nomos_da_network_core::protocols::sampling::opinions::{Opinion, OpinionEvent};
+use subnetworks_assignations::MembershipHandler;
+
+const OPINION_THRESHOLD: f32 = 0.9;
```
danielSanchezQ: uhm, we should probably not use f32 but a fixed precission type instead.

nomos-services/data-availability/network/src/opinion_aggregator.rs
```diff
@@ -0,0 +1,224 @@
...
+
+pub struct OpinionAggregator<Membership>
+where
+    Membership: MembershipHandler<Id = PeerId>,
+{
+    local_peer_id: PeerId,
+    positive_opinions: HashMap<PeerId, u32>,
+    negative_opinions: HashMap<PeerId, u32>,
```
danielSanchezQ, line 17: This should be a [bitvec](https://docs.rs/bitvec/latest/bitvec/) instead probably, order comes from ordering the peer ids.

nomos-services/data-availability/network/src/opinion_aggregator.rs
```diff
@@ -0,0 +1,224 @@
...
+    blacklist: HashSet<PeerId>,
+
+    old_positive_opinions: HashMap<PeerId, u32>,
+    old_negative_opinions: HashMap<PeerId, u32>,
+    old_blacklist: HashSet<PeerId>,
+
+    current_membership: Option<Membership>,
+    previous_membership: Option<Membership>,
```
danielSanchezQ, line 25: Why do we have membership here?

nomos-services/data-availability/network/src/opinion_aggregator.rs
pradovic: The reason why is that we need to lexicographically sort ProviderIDs not PeerIds, and we agreed that sdp will make this conversion, so it can sort them. But HashMap are prepopulated  and updated in the fixed size manner, so it should be straightforward. Does that makes sense?

nomos-services/data-availability/network/src/opinion_aggregator.rs
pradovic: Membership has session, so it was easy to provide it, as it is not copied (Membership is Arc underneatch), but this can be changed to session id. 
The point is to distinguish between new opinions (S), and old opinions (S-1)

pradovic: Also, option is there so we don't start forming unless we have both S and S -1

nomos-services/data-availability/network/src/opinion_aggregator.rs
```diff
@@ -0,0 +1,224 @@
...
+                let opinion = if include_self && peer_id == self.local_peer_id {
+                    true
+                } else {
+                    let pos = positive.get(&peer_id).copied().unwrap_or(0);
+                    if pos == 0 {
+                        false
+                    } else {
+                        let neg = negative.get(&peer_id).copied().unwrap_or(0);
+                        let ratio = (pos as f32 - neg as f32) / pos as f32;
+                        ratio > OPINION_THRESHOLD
```
danielSanchezQ: I think this is wrong. This would be `pos - neg / total`?

danielSanchezQ:
> > The opinions should be rethink imo. It was stipulated in the spec that they are bitvectors.We need to send them through wire as well so I would rather get them ready here.
> 
> The only reason the vectors are not created here is that we need to sort ProviderIds in lexicographical order. We agreed that PeerID -> ProviderID conversion will be in SDP, so I offloaded this logic to SDP as well. The hashmaps are fixed size, so they fit, it is just that they need to be sorted after conversion.
> 
> So the SDP part would need to convert PeerID -> ProviderID, sort them and form activity proof, nothing else.

Ah true, ok, then we should provide such method in opinions, so we feed the other hashmap and return the bitvec.

nomos-services/data-availability/network/src/opinion_aggregator.rs
danielSanchezQ: Answered to the overall comment below.

nomos-services/data-availability/network/src/opinion_aggregator.rs
pradovic: If I understood correctly in spec:

```
6. When the next session starts, create an opinions binary for every node $i \in \mathcal N$ such that:
    1. `old_opinions[i] = opinion(i, old.positive_opinions, old.negative_opinions, old.opinions_threshold)`.
    2. `new_opinions[i] = opinion(i, new.positive_opinions, new.negative_opinions, new.opinions_threshold)`.
    
    ```python
    def opinion(i, positive_opinions, negative_opinions, opinion_threshold):
      if positive_opinions > 0:
        ratio = (positive_opinions[i] - negative_opinions[i]) / positive_opinions[i]
      else:
    	  ratio = 0
      return (ratio > opinion_threshold)
    ```
    
7. A node sets a positive opinion about itself in the `new_opinions` vector, which is verified during the activity proof verification.
8. A node sets a positive opinion about itself in the `old_opinions` if the node was taking part in the protocol during the previous session, which is not verified.

```
I dont see `pos - neg / total?` 🙈 ?

nomos-services/data-availability/network/src/opinion_aggregator.rs
pradovic: Good catch! Do we have a lib that we use in the code for this, so I don't introduce the new one? Or maybe we could do somethig like:
```
const OPINION_THRESHOLD_NUM: u32 = 9;
const OPINION_THRESHOLD_DENOM: u32 = 10;

// Instead of: ratio > 0.9
// Use: ratio * OPINION_THRESHOLD_DENOM > positive * OPINION_THRESHOLD_NUM
if (pos - neg) * OPINION_THRESHOLD_DENOM > pos * OPINION_THRESHOLD_NUM {
    // positive opinion
}
```

we can increase precision by increasing the integers

nomos-services/data-availability/network/src/opinion_aggregator.rs
danielSanchezQ: Well, lets follow the spec here then :smile:

pradovic: It looks to me that spec's formula specifically evaluates the degradation of positive opinions by negative ones instead of the net positive rate among all interactions

nomos-services/data-availability/network/src/opinion_aggregator.rs
danielSanchezQ: We do use one somewhere at some point, but I dont recall :sweat_smile: 
But probably: https://docs.rs/fixed/latest/fixed/

pradovic: I'll try to find it! (Update) I found it, it looks like we are using `fixed`. I will update the PR

nomos-services/data-availability/network/src/opinion_aggregator.rs
pradovic: Oups, this was supposed to go before your comment, the internet is slow :) Agreed

nomos-services/data-availability/network/src/opinion_aggregator.rs
pradovic: @madxor What do you think about this proposition:
```
const OPINION_THRESHOLD_NUM: u32 = 9;
const OPINION_THRESHOLD_DENOM: u32 = 10;

fn calculate_opinion(&self, peer_id: &PeerId, positive: &HashMap<PeerId, u32>, negative: &HashMap<PeerId, u32>) -> bool {
    let pos = positive.get(peer_id).copied().unwrap_or(0);
    
    if pos == 0 {
        return false;
    }
    
    let neg = negative.get(peer_id).copied().unwrap_or(0);
    
    // Instead of: (pos - neg) / pos > 0.9
    // We do: (pos - neg) * 10 > pos * 9
    (pos - neg) * OPINION_THRESHOLD_DENOM > pos * OPINION_THRESHOLD_NUM
}
```

This way we avoid division at all, so it might be the most deterministic variant. Also avoids additional crate for fixed precision type.

madxor: Looks good but we might simplify it even more: `pos * OPINION_THRESHOLD_NUM > neg`. I can also define the threshold as "multiplication" of negative opinions, then `OPINION_THRESHOLD=9` so we could even get rid of that  .

pradovic: Hmm, I am probably missing the point, would this give different results, for example:
With 100 positive, 5 negative: 900 > 5  (passes)
With 10 positive, 5 negative: 90 > 5  (false positive) (this is 50% degradation, right?)

madxor: My bad. That should be the other way around:
```
(P - N) * 10    > P * 9
10 * P - 10 * N > P * 9
10 * P          > P * 9 + 10 * N
10 * P - P * 9  > 10 * N
P               > 10 * N
```

pradovic: This looks great! Tnx

pradovic: https://github.com/logos-co/nomos/pull/1736/commits/01af097799241e3b563e1e969019ba86421579d4

nomos-services/data-availability/network/src/opinion_aggregator.rs
```diff
@@ -0,0 +1,238 @@
...
+                    let neg = negative.get(&peer_id).copied().unwrap_or(0);
+                    pos > neg * OPINION_THRESHOLD
+                };
+                (peer_id, opinion)
+            })
+            .collect()
+    }
+}
+
+impl Opinions {
```
bacv, line 225: Nit: lets leave it as is, but the Activity message will be formed by the SDP service, the encoding will be defined in nomos-core, this will be changed when working on SDP service.

pradovic: I agree, this was a helper fn with the assumption that SDP will provide mappins for PeerID -> ProviderID. Should be adapter when implementing.

bacv: You are right, so the mappings now will be managed by DA service, it will form Opinions. I see that there are two Opinions defined, one in `nomos-da`, another in `nomos-da-network-service`, we should probably have one definition in `nomos-core`, as these won't be sent over DA protocol, and will be used in chain service, where we don't want dep of da network service. 

Lets leave it as is, I'll move it nomos-core, make enum for Activity message metadata, as we'll only use it for blend and da. Then DA network will be able to send Opinions with session id as SDP service message.

bacv (approved):

nomos-services/data-availability/network/src/opinion_aggregator.rs
pradovic: @danielSanchezQ suggested this helper functions as his idea was for sdp to figure out the mappings based on locators. If this is not possible then we can work it out in da service by either saving the appropriate mappings to storage while processing updates, or by saving the actual assignations per ProviderID and creating the PeerID from ProviderID only when it is needed by behaviors. In either way, it is probably out of scope for this PR

bacv: > sdp to figure out the mappings based on locators.

@danielSanchezQ what do you think if DA network will provide SDP service opinions already sortable (or sorted) by the provider ids.

Since we move the membership management to chain service, SDP will only know about local node participation and won't track other provider ids, nor their locators. It will only be able to form correct SDP messages on behalf of local node. 

> In either way, it is probably out of scope for this PR

Agree, this PR looks good for opinion forming.

danielSanchezQ: > @danielSanchezQ what do you think if DA network will provide SDP service opinions already sortable (or sorted) by the provider ids.
 
 Yes that was the initial idea, but we do not have the provider ids in the network level from my last convo with @pradovic.
 If we can have it, then we should use the bitarrays for opinions and should be sorted.

bacv: > Yes that was the initial idea, but we do not have the provider ids

We actually receive ProviderId from membership update stream, just convert it to PeerId inside DA network service, instead of converting it to PeerId, we'll create a mapping in DA network service.

pradovic: @danielSanchezQ I proposed to have mappings in the da service, but I understood that you were against it 🙈 We would need to keep track of the mappings. Then we could provide them sorted

danielSanchezQ: > > Yes that was the initial idea, but we do not have the provider ids
> 
> We actually receive ProviderId from membership update stream, just convert it to PeerId inside DA network service, instead of converting it to PeerId, we'll create a mapping in DA network service.

Then I think it is less convoluted to do it properly. Also we remove the hashmaps and use the proper bitarrays.

pradovic: @danielSanchezQ good to merge then? Or should we first open a DA related pr to handle the mappings?

danielSanchezQ: > @danielSanchezQ good to merge then? Or should we first open a DA related pr to handle the mappings?

We were talking on doing it network level. So I guess we should do it here if possible. But a further PR is fine.

pradovic: > > Yes that was the initial idea, but we do not have the provider ids
> 
> We actually receive ProviderId from membership update stream, just convert it to PeerId inside DA network service, instead of converting it to PeerId, we'll create a mapping in DA network service.

@danielSanchezQ  This is in da-network, no? We would need to save ProviderID -> PeerID mapping when receiving updates from membership. @bacv Did I understand this correctly? Here is the relevant code: `nomos-services/data-availability/network/src/membership/adapters/service.rs`. Just want to double check that we agree.

bacv: > ProviderID -> PeerID mapping when receiving updates from membership

I would remove conversion code from the adapter and return membership update (Providers+Locators) as is in the main da network service loop, create a mapping, pass converted PeerId further where needed.

pradovic: I agree. We could maybe even save some storage by saving everything per ProviderId instead PeerID, and then convert to PeerID when it is needed downstream. It is a deterministic function ProviderID -> PeerID

PR #1738, opened 2025-09-24
Drusu/chain service decomposition to leader follower

davidrusu:

> [!NOTE]  
> Review the 

> [!IMPORTANT]  
> explained.  

> [!IMPORTANT]  
> advancing  
> the review.

> [!IMPORTANT]

> [!IMPORTANT]
    modification/solutions.

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


nomos-services/chain/chain-leader/src/lib.rs
```diff
@@ -0,0 +1,552 @@
...
+            TimeService<_, _>
+        )
+        .await?;
+
+        let async_loop = async {
+            loop {
+                tokio::select! {
+                    Some(SlotTick { slot, .. }) = slot_timer.next() => {
+                        // TODO: Don't propose blocks until IBD is done and online mode is activated.
+                        //       Until then, mempool, DA, blend service will not be ready: https://github.com/logos-co/nomos/issues/1656
```
danielSanchezQ, line 326: Now that is separated I think we can wait for common chain to be ready, meaning that ibd is done (?)

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -1044,44 +928,42 @@ where
...
+                let epoch_state = cryptarchia.epoch_state_for_slot(slot).cloned();
+                tx.send(epoch_state).unwrap_or_else(|_| {
+                    error!("Could not send epoch state through channel");
+                });
+            }
+            ConsensusMsg::ProcessLeaderBlock { .. } => {
+                // ProcessBlock is handled separately in the run loop where we have async
+                // context This should never be reached since we filter it out
+                // before calling process_message
+                error!("ProcessBlock should be handled in the run loop, not in process_message");
```
danielSanchezQ: this should be a panic! then, as it is a bug.

nomos-services/chain/chain-leader/src/lib.rs
davidrusu: This TODO was here before so I don't want to address it in this PR, but yeah if chain-service does not signal ready until after IBD, then maybe this is would fix this TODO

danielSanchezQ: Yeah, we can address it in a later PR as well too.

nomos-services/chain/chain-service/src/lib.rs
davidrusu: yes, fixed now

nodes/nomos-executor/executor/src/api/backend.rs
```diff
@@ -395,7 +378,7 @@ where
         wait_until_services_are_ready!(
             &overwatch_handle,
             Some(Duration::from_secs(60)),
-            Cryptarchia<_, _, _, _, _, _, _, _, SIZE>,
+            Cryptarchia<_, _, _, _, _, _>,
```
zeegomo, line 381: love to see this

nomos-services/chain/chain-leader/src/relays.rs
```diff
@@ -0,0 +1,183 @@
...
+            .expect("Relay connection with TimeService should succeed");
+
+        Self::new(blend_relay, cl_mempool_relay, sampling_relay, time_relay)
+    }
+
+    pub const fn blend_relay(&self) -> &BlendRelay<BlendService> {
+        &self.blend_relay
+    }
+
+    pub const fn cl_mempool_relay(
```
zeegomo, line 170: not sure whether this is coming from the old thing, but I thought we got rid of `cl_` naming

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -777,64 +699,41 @@ where
...
-                                tx_selector.clone(),
-                                &relays
-                            ).await;
-
-                            if let Some(block) = block {
-                                // apply our own block
-                                match Self::process_block_and_update_state(
+                    Some(msg) = self.service_resources_handle.inbound_relay.next() => {
+                        // Handle ProcessBlock separately since it needs async and access to more state
+                        if let ConsensusMsg::ProcessLeaderBlock { block, tx } = msg {
```
zeegomo, line 707: why do we need to have two paths for blocks from leader and blocks from network?

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -1044,44 +928,44 @@ where
...
+                    "ProcessLeaderBlock should be handled in the run loop, not in process_message"
+                );
+            }
         }
     }
 
-    #[expect(
-        clippy::type_complexity,
-        reason = "CryptarchiaConsensusState and CryptarchiaConsensusRelays amount of generics."
-    )]
```
zeegomo, line 1056: <3

zeegomo (approved):
Love this , left just one comment about unifying paths for leader and network blocks

nomos-services/chain/chain-leader/src/relays.rs
danielSanchezQ: yeah, it can go away now, we only have a mempool

nomos-services/chain/chain-service/src/lib.rs
danielSanchezQ: We can merge the two streams, the one from network and the one from the leader chain service. Then single path (?)

zeegomo: that was my idea indeed

davidrusu: short answer is that we skip DA blob validation for leader proposed blocks (since the leader had already validated blobs before including them).

Merging these two into a single case is hard because we use a trait for the validation type. In the merged case you would need a `if leader block { skip_validation } else { do_normal_validation}`. To get that conditional to type check you need to boxed dyn traits, and then that change gets quite messy.

But, ignoring all that, how do you go about merging the streams? would you create a `block_stream` channel in this run method and then instead of processing blocks as we receive them, we would instead enqueue them in this `block_stream` channel?

Then we have a handler for this `block_stream` channel as another branch of the `select` loop?

zeegomo: > But, ignoring all that, how do you go about merging the streams? would you create a block_stream channel in this run method and then instead of processing blocks as we receive them, we would instead enqueue them in this block_stream channel?
Then we have a handler for this block_stream channel as another branch of the select loop?

Yes, something like this, you create a new stream where you merge the two sources, then we just need one branch in the select loop for both

How bad would it be to check again for a leader block?

davidrusu: I wanted to keep functionality identical with this refactor, otherwise things start getting super complicated.

I can bring this up to discuss after the PR's in to see if it's worth dropping this branch where we don't sample leader blocks?

If we always sample, than combining the two is trivial

nomos-services/chain/chain-leader/src/relays.rs
davidrusu: It was an old thing, we call the mempool CLPool in a bunch of places in chain-service. I can make a cleanup task after this PR goes in to switch everything to some variant of mempool

zeegomo: no worries, no need to do it in this pr, was just surprised

nodes/nomos-executor/executor/src/lib.rs
```diff
@@ -126,16 +124,11 @@ pub(crate) type DaNetworkAdapter = nomos_da_sampling::network::adapters::executo
...
-        DaNetworkApiAdapter,
-        RuntimeServiceId,
-    >,
-    RuntimeServiceId,
->;
+pub(crate) type CryptarchiaService =
+    nomos_node::generic_services::CryptarchiaService<DaNetworkAdapter, RuntimeServiceId>;
+
+pub(crate) type CryptarchiaLeaderService =
+    nomos_node::generic_services::CryptarchiaLeaderService<DaNetworkAdapter, RuntimeServiceId>;
```
youngjoon-lee, line 132: I think we can finally rename these to `ChainService` and `ChainLeaderService`, but not important in this PR. :)

nomos-services/chain/chain-leader/src/lib.rs
youngjoon-lee: Yeah, I will work on this issue (as the way David said above). Thanks!

nomos-services/chain/chain-service/src/lib.rs
youngjoon-lee: I agree with having this discussion separately after merging this PR. Since the blob validation in this case is not heavy, we could consider combining the two.
I introduced the `SkipBlobValidation` here, since, in principle, the chain service doesn't know how heavy the blob validation will be.

IMO, I'm fine with having two paths (or refactoring it in any way), but also okay with not skipping blob validation. Anyways, yeah, it's out of scope for this PR.

PR #1739, opened 2025-09-24
feat(chain): provide subscribe API for winning epoch slot info

ntn-x2:

This is a preliminary PR to discuss about the approach taken and to introduce a shell API that will be implemented later on. It introduces the following components:

* A `subscribe` API for consumers to be notified whenever a new winning slot has been found by the chain service, along with all the related info that would successfully generate a valid PoL proof. The implementation currently returns a stream which never yields any items, as those will be streamed after the PoL changes have been merged.
* A trait, in the Blend service, that allows Blend to consume the stream above without directly depending on chain service, since chain service already depends on Blend
* An implementation of the trait above provided by the node binary, which mediates between the two components

**Because chain service already waits for Blend to be ready, Blend cannot wait that the stream becomes available before notifying about its readiness else we have a deadlock. The solution to this is that Blend polls the stream only after it has advertised its readiness, which means that the stream is not vital for the correct functioning of Blend, since Blend service can still, on its own, generate and blend cover messages.**

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nodes/nomos-node/node/src/generic_services/blend.rs
+++ b/nodes/nomos-node/node/src/generic_services/blend.rs
@@ -194,48 +194,50 @@ where
             .send(LeaderMsg::WinningPolEpochSlotStreamSubscribe { sender })
             .await
             .ok()?;
-        let item = receiver.await.ok()?;
-        Some(Box::new(BroadcastStream::new(item).filter_map(|res| {
-            let Ok(private) = res else {
-                return ready(None);
-            };
-            let PolWitnessInputsData {
-                chain:
-                    PolChainInputsData {
-                        epoch_nonce,
-                        slot_number,
-                        ..
-                    },
-                wallet:
-                    PolWalletInputsData {
-                        note_value,
-                        transaction_hash,
-                        output_number,
+        let pol_winning_slot_receiver = receiver.await.ok()?;
+        Some(Box::new(
+            BroadcastStream::new(pol_winning_slot_receiver).filter_map(|res| {
+                let Ok(private) = res else {
+                    return ready(None);
+                };
+                let PolWitnessInputsData {
+                    chain:
+                        PolChainInputsData {
+                            epoch_nonce,
+                            slot_number,
+                            ..
+                        },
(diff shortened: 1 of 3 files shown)


danielSanchezQ:
We will have to wait for #1738 to be merged first if you dont mind. It should be easier to update this PR than otherwise.

nodes/nomos-node/node/src/generic_services/blend.rs
```diff
@@ -117,25 +135,109 @@ fn loop_until_valid_proof(
...
+        + 'static,
+{
+    type Stream = Box<dyn Stream<Item = PolEpochInfo> + Send + Unpin>;
+
+    async fn subscribe(
+        overwatch_handle: &OverwatchHandle<RuntimeServiceId>,
+    ) -> Option<Self::Stream> {
+        use groth16::Field as _;
+
+        wait_until_services_are_ready!(overwatch_handle, Some(Duration::from_secs(3)), CryptarchiaLeaderService<SamplingAdapter, RuntimeServiceId>).await.ok()?;
```
danielSanchezQ, line 187: This is smart!, a local waiting!!

danielSanchezQ (approved):
This is the way to go.

ntn-x2:
> Chain leader calls BlendAdapter::propose(pol, block). Blend will then proof the PoL variant of PoQ and wrap the block proposal.

I am not so sure this premise is actually valid. We know `pol` info does not change for the whole epoch, so I think regardless of the time it takes to generate leadership proofs, there could still be a subscribe API for chain leader to notify consumers about what slots will be winning ones in this epoch. This is useful for Blend, but might be useful for other services in the future as well (?). And that is because it is misleading to link proof generation (or `pol` inputs) to the act of blending a message, because in my opinion they are not linked. What is linked to new `pol` inputs is the rotation to a new epoch, not the blending of a new message. These are me 2 cents.

davidrusu:
@ntn-x2 @danielSanchezQ sure thing

PR #1745, opened 2025-09-25
feat(da): return assignations instead of panic when nodes < replication factor

pradovic:
Return empty assignations when the number of nodes is less then replication factor. The code used to panic in this case. This is handling it more gracefully, so the behaviors can skip the work in this case.

Yes

@pradovic @danielSanchezQ 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-da/network/subnetworks-assignations/src/versions/history_aware_refill/assignations.rs
+++ b/nomos-da/network/subnetworks-assignations/src/versions/history_aware_refill/assignations.rs
@@ -552,4 +552,30 @@ mod tests {
         .collect();
         assert!(assignations.iter().all(|a| a == &assignations[0]));
     }
+
+    #[test]
+    fn test_insufficient_nodes() {
+        let mut rng = thread_rng();
+
+        let nodes: Vec<TestId> = std::iter::repeat_with(|| {
+            let mut buff = [0u8; 32];
+            rng.fill_bytes(&mut buff);
+            TestId(buff)
+        })
+        .take(3)
+        .collect();
+
+        let previous_nodes: Vec<BTreeSet<TestId>> = std::iter::repeat_with(BTreeSet::new)
+            .take(SUBNETWORK_SIZE)
+            .collect();
+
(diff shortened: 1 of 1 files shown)


nomos-da/network/subnetworks-assignations/src/versions/history_aware_refill/assignations.rs
```diff
@@ -204,10 +204,11 @@ impl HistoryAwareRefill {
         for<'id> Id: Ord + Copy + Hash + 'id,
         Rng: RngCore,
     {
-        assert!(
-            new_nodes_list.len() >= replication_factor,
-            "The network size is smaller than the replication factor"
-        );
+        // Behaviors can skip when assignations are empty
+        if new_nodes_list.len() < replication_factor {
+            return Vec::new();
```
danielSanchezQ, line 209: we should probably log something here on in the upper layer

nomos-da/network/subnetworks-assignations/src/versions/history_aware_refill/assignations.rs
```diff
@@ -204,10 +204,11 @@ impl HistoryAwareRefill {
         for<'id> Id: Ord + Copy + Hash + 'id,
         Rng: RngCore,
     {
-        assert!(
-            new_nodes_list.len() >= replication_factor,
-            "The network size is smaller than the replication factor"
-        );
+        // Behaviors can skip when assignations are empty
+        if new_nodes_list.len() < replication_factor {
```
danielSanchezQ, line 208: actually they are 2 different things. this sould be `if new_nodes_list.len().is_zero()` and after that keep the panic if nodes_len is less than replication factor (which should never happen anyway).

pradovic: I see, if we expect for this condition never happen then assert make sense. I will update the the PR.

pradovic:
Will be handled on da-network level.

PR #1746, opened 2025-09-25
fix(zk): poq input order

thomaslavaur:

fix the input order of PoQ

Yes

@thomaslavaur 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/nomos-blend/message/src/crypto/proofs/quota/mod.rs
+++ b/nomos-blend/message/src/crypto/proofs/quota/mod.rs
@@ -96,7 +96,7 @@ impl ProofOfQuota {
     pub(super) fn verify(self, public_inputs: &PublicInputs) -> Result<ZkHash, Error> {
         let verifier_input =
             VerifyInputs::from_prove_inputs_and_nullifier(*public_inputs, self.key_nullifier);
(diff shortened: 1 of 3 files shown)


ntn-x2 (approved):
FYI @danielSanchezQ, in case some weird verification failures happen for other proof types as well. It only took a day of debugging, to find out what the issue was.

danielSanchezQ:
> FYI @danielSanchezQ, in case some weird verification failures happen for other proof types as well. It only took a day of debugging, to find out what the issue was. This happened because we are not re-using the verification inputs returned by the `prove` function, of course, but the verifier is fetching them from its own context, and those were transformed differently than what the prove function returns.

Yeah, but this should have happen in the tests as well. Otherwise we may want to update the tests to reflect this.

ntn-x2:
@danielSanchezQ yes, probably. Anyway I am also working on the tests inside Blend crates. But I agree it should probably also be added to those tests.

ntn-x2:
@danielSanchezQ this PR broke the tests and the `verify` function of PoQ if the inputs returned by the `prove` function are used. I opened a draft PR that adds verification with reconstructed inputs: https://github.com/logos-co/nomos/pull/1748. This now works, but using the inputs returned by `prove` does not work anymore.

PR #1747, opened 2025-09-25
feat(kms): Integrate ZK key

strinnityk:
Implement and integrate Zk key into KMS crate.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/nomos-core/chain-defs/src/mantle/keys.rs
+++ b/nomos-core/chain-defs/src/mantle/keys.rs
@@ -1,8 +1,9 @@
 use groth16::{Fr, serde::serde_fr};
 use num_bigint::BigUint;
 use serde::{Deserialize, Serialize};
+use zeroize::ZeroizeOnDrop;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, ZeroizeOnDrop)]
 #[serde(transparent)]
 pub struct SecretKey(#[serde(with = "serde_fr")] Fr);
 
@@ -20,6 +21,11 @@ impl SecretKey {
     #[must_use]
     pub fn to_public_key(&self) -> PublicKey {
         unimplemented!("Conversion from SecretKey to PublicKey is not implemented yet")
+    }
+
+    #[must_use]
+    pub fn sign(&self, _data: &Fr) -> Fr {
+        unimplemented!("Signing is not implemented yet")
     }
 }
 
--- a/nomos-services/key-management-system/src/encodings/errors.rs
+++ b/nomos-services/key-management-system/src/encodings/errors.rs
@@ -1,6 +1,8 @@
-use std::any::type_name;
+use std::any::{type_name, type_name_of_val};
 
 use thiserror::Error;
+
+use crate::keys::SecuredKey;
 
 /// Errors that can occur when encoding or decoding.
 #[derive(Error, Debug)]
@@ -13,7 +15,12 @@ impl EncodingError {
     /// Creates a new `EncodingError::Requires` error.
     /// It contains the path to the required type.
     #[expect(dead_code, reason = "Will be used when integrating KMS.")]
-    pub fn requires<T>() -> Self {
-        Self::Requires(type_name::<T>().to_owned())
+    pub fn requires<Key: SecuredKey, Payload>(key: &Key, received_payload: &Payload) -> Self {
+        let key_type_name = type_name_of_val(key);
+        let payload_type_name = type_name::<Key::Payload>().to_owned();
+        let received_payload_type_name = type_name_of_val(received_payload);
+        Self::Requires(format!(
+            "Key of type `{key_type_name}` requires a payload of type `{payload_type_name}`, but got payload of type `{received_payload_type_name}`",
+        ))
     }
 }
--- a/nomos-services/key-management-system/src/encodings/errors.rs
+++ b/nomos-services/key-management-system/src/encodings/errors.rs
@@ -1,4 +1,4 @@
-use std::fmt::{Debug, Display};
+use std::any::type_name;
 
 use thiserror::Error;
 
@@ -10,8 +10,10 @@ pub enum EncodingError {
 }
 
 impl EncodingError {
-    #[expect(dead_code, reason = "Will be used when adding the ZK key.")]
-    pub fn requires<T: Display>(encoding: T) -> Self {
-        Self::Requires(encoding.to_string())
+    /// Creates a new `EncodingError::Requires` error.
+    /// It contains the path to the required type.
+    #[expect(dead_code, reason = "Will be used when integrating KMS.")]
+    pub fn requires<T>() -> Self {
+        Self::Requires(type_name::<T>().to_owned())
     }
 }
--- a/nomos-services/key-management-system/src/keys/mod.rs
+++ b/nomos-services/key-management-system/src/keys/mod.rs
@@ -1,12 +1,13 @@
 mod ed25519;
 mod errors;
 mod secured_key;
+mod zk;
 
 use serde::{Deserialize, Serialize};
 use zeroize::ZeroizeOnDrop;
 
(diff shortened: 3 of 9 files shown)


nomos-core/chain-defs/src/mantle/keys.rs
```diff
@@ -1,8 +1,9 @@
 use groth16::{Fr, serde::serde_fr};
 use num_bigint::BigUint;
 use serde::{Deserialize, Serialize};
+use zeroize::ZeroizeOnDrop;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, ZeroizeOnDrop)]
```
danielSanchezQ, line 6: There is an open pr for this actually. But I can close the other and let it here.

nomos-core/chain-defs/src/mantle/keys.rs
```diff
@@ -21,6 +22,11 @@ impl SecretKey {
     pub fn to_public_key(&self) -> PublicKey {
         unimplemented!("Conversion from SecretKey to PublicKey is not implemented yet")
     }
+
+    #[must_use]
+    pub fn sign(&self, _data: &Fr) -> Fr {
+        unimplemented!("Signing is not implemented yet")
+    }
```
danielSanchezQ, line 29: Well, just implement it!

nomos-services/key-management-system/src/keys/mod.rs
```diff
@@ -24,17 +26,27 @@ impl SecuredKey for Key {
...
+                Err(EncodingError::requires::<<Ed25519Key as SecuredKey>::Payload>().into())
+            }
+            (Self::Zk(key), Self::Payload::Zk(payload)) => {
+                key.sign(payload).map(Self::Signature::Zk)
+            }
+            (Self::Zk(_), _) => {
+                Err(EncodingError::requires::<<ZkKey as SecuredKey>::Payload>().into())
             }
         }
     }
```
danielSanchezQ, line 39: I think this can be simplified(?)

```suggestion
            (Self::Ed25519(key), Self::Payload::Ed25519(payload)) => {
                key.sign(payload).map(Self::Signature::Ed25519)
            }
            (Self::Zk(key), Self::Payload::Zk(payload)) => {
                key.sign(payload).map(Self::Signature::Zk)
            }
            (key, payload) => {
                Err(EncodingError::requested_and_got::(key, got).into()) // not best naming tho
}
}
}
```

nomos-services/key-management-system/src/keys/mod.rs
```diff
@@ -24,17 +26,27 @@ impl SecuredKey for Key {
...
-        match (self, data) {
-            (Self::Ed25519(key), Self::Payload::Ed25519(data)) => {
-                key.sign(data).map(Self::Signature::Ed25519)
+    fn sign(&self, payload: &Self::Payload) -> Result<Self::Signature, Self::Error> {
+        match (self, payload) {
+            (Self::Ed25519(key), Self::Payload::Ed25519(payload)) => {
+                key.sign(payload).map(Self::Signature::Ed25519)
+            }
+            (Self::Ed25519(_), _) => {
+                Err(EncodingError::requires::<<Ed25519Key as SecuredKey>::Payload>().into())
```
ntn-x2: Mhh, this is once again a runtime error vs a compile-time error. I am thinking how we can change that? Ideally we should not be allowed to pass a wrong payload for a given key. What is the need to gather keys under an enum, instead of just implementing the `SecuredKey` for each key? If we want to prevent outside crates to implement the same, we can seal the trait so only we can implement it. But in this case I would need some help to understand why we need this, and whether we can get rid of it?

ntn-x2 (commented):
I would like to understand the need to declare enums vs using structs and implement traits on them.

nomos-services/key-management-system/src/keys/mod.rs
danielSanchezQ: Yeah, the problem is that we are trying to do dispatch based on generic types. The first idea @strinnityk had was to use the `TryInto` approach, but I made him change that for something that is more explicit. Imo this should be a panic as well because it should not happen. Now, this is a decision to make, do we allow the user to panic instead by giving them an error, or do we panic ourselves here and force it. Same case as the overall API when requesting keys. If we decidied to return errors, so errors it is. 
Now, if you have some idea so we can make the dispatch at compile time using the types that would be awesome.

ntn-x2: Yeah no I am not arguing on either panic or errors, I am arguing whether we can avoid runtime problems by introducing compile-time guarantees. I can try to see if I find an interesting avenue as soon as I have some time to spare. But until then we can go ahead with this PR. I would prefer if the service panics for any internal invariant that is not upheld, while it returns an error if it's something the calling service should manage.

danielSanchezQ: The problem is that this invariant may come from calling the service with wrong values. So again, imo that should be a panic rather than an error, but we should do that in all places. I think we can manage to avoid with some macros (so we generate the code specific for the available types), but that would take time.

nomos-core/chain-defs/src/mantle/keys.rs
strinnityk: As you wish. There's an unneeded clone in your PR, but we can merge yours and then I fix it with this one.

nomos-core/chain-defs/src/mantle/keys.rs
strinnityk: If only I knew how to 😂

nomos-services/key-management-system/src/keys/mod.rs
strinnityk: Well, the thing is I was trying to be a bit more helpful with the error:
```rust
pub fn requires<T>() -> Self {
    Self::Requires(type_name::<T>().to_owned())
}
```
That's why I needed the type, not the values.

strinnityk: The thing is I was using `type_name<T>` to print out better debug info. That being said, I just realised of a way to achieve it with values. Thanks!

nomos-core/chain-defs/src/mantle/keys.rs
danielSanchezQ: Ok, ill do it in a further pr and ping you there.

nomos-services/key-management-system/src/encodings/errors.rs
```diff
@@ -10,8 +12,15 @@ pub enum EncodingError {
...
+    /// Creates a new `EncodingError::Requires` error.
+    /// It contains the path to the required type.
+    #[expect(dead_code, reason = "Will be used when integrating KMS.")]
+    pub fn requires<Key: SecuredKey, Payload>(key: &Key, received_payload: &Payload) -> Self {
+        let key_type_name = type_name_of_val(key);
+        let payload_type_name = type_name::<Key::Payload>().to_owned();
+        let received_payload_type_name = type_name_of_val(received_payload);
+        Self::Requires(format!(
+            "Key of type `{key_type_name}` requires a payload of type `{payload_type_name}`, but got payload of type `{received_payload_type_name}`",
+        ))
```
danielSanchezQ, line 24: Sweet black magic, nice.

nomos-services/key-management-system/src/keys/mod.rs
```diff
@@ -24,17 +26,22 @@ impl SecuredKey for Key {
...
-                key.sign(data).map(Self::Signature::Ed25519)
+    fn sign(&self, payload: &Self::Payload) -> Result<Self::Signature, Self::Error> {
+        match (self, payload) {
+            (Self::Ed25519(key), Self::Payload::Ed25519(payload)) => {
+                key.sign(payload).map(Self::Signature::Ed25519)
             }
+            (Self::Zk(key), Self::Payload::Zk(payload)) => {
+                key.sign(payload).map(Self::Signature::Zk)
+            }
+            (key, payload) => Err(EncodingError::requires(key, payload).into()),
```
danielSanchezQ, line 37: Much cleaner ❤️

PR #1753, opened 2025-09-26
fix(encoding): Use be instead of le

danielSanchezQ:

We were missusing little endian instead of big endian which made things unconsistent. This should remove all uses of unintended le.

Simple renaming

@madxor @thomaslavaur @danielSanchezQ 

Some parts where missing. But it was discussed and will be updated.

Some information about encoding will be added.

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-da/kzgrs/src/common.rs
+++ b/nomos-da/kzgrs/src/common.rs
@@ -76,7 +76,7 @@ pub fn bytes_to_evaluations<const CHUNK_SIZE: usize>(
             .map(
                 // use little endian for convenience as shortening 1 byte (<32 supported)
                 // do not matter in this endianness
-                field_element_from_bytes,
+                field_element_from_bytes_le,
             )
             .collect(),
         domain,
@@ -129,9 +129,9 @@ pub fn bytes_to_polynomial_unchecked<const CHUNK_SIZE: usize>(
 /// data fits within the bls modulus.
 /// Data len cannot be higher than `BYTES_PER_FIELD_ELEMENT`
 #[must_use]
-pub fn field_element_from_bytes(b: &[u8]) -> FieldElement {
+pub fn field_element_from_bytes_le(b: &[u8]) -> FieldElement {
     assert!(b.len() <= BYTES_PER_FIELD_ELEMENT);
-    FieldElement::from(BigUint::from_bytes_be(b))
+    FieldElement::from(BigUint::from_bytes_le(b))
 }
 
 #[must_use]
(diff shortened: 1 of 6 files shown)


ntn-x2 (commented):
`fr_from_bytes`, `fr_from_bytes_unchecked` and `fr_to_bytes` still use LE encoding, so that should change as well.

danielSanchezQ:
> `fr_from_bytes`, `fr_from_bytes_unchecked` and `fr_to_bytes` still use LE encoding, so that should change as well.

Yes, I was checking on it. I don't know why it escaped the refactor...

nomos-da/kzgrs/src/rs.rs
```diff
@@ -83,7 +83,7 @@ pub fn lagrange_interpolate(points: &[Fr], roots_of_unity: &[Fr]) -> DensePolyno
 /// size and a set of points
 pub fn points_to_bytes<const CHUNK_SIZE: usize>(points: &[Fr]) -> Vec<u8> {
     fn point_to_buff<const CHUNK_SIZE: usize>(p: &Fr) -> impl Iterator<Item = u8> {
-        p.into_bigint().to_bytes_le().into_iter().take(CHUNK_SIZE)
+        p.into_bigint().to_bytes_be().into_iter().take(CHUNK_SIZE)
```
ntn-x2: Shouldn't this also use `fr_to_bytes`? Ok maybe out of scope, but we should have one pass where all these instances are now converted to the utility functions we expose, else once again we can get out of sync.

nomos-da/kzgrs/src/common.rs
```diff
@@ -129,9 +129,9 @@ pub fn bytes_to_polynomial_unchecked<const CHUNK_SIZE: usize>(
 /// data fits within the bls modulus.
 /// Data len cannot be higher than `BYTES_PER_FIELD_ELEMENT`
 #[must_use]
-pub fn field_element_from_bytes_le(b: &[u8]) -> FieldElement {
+pub fn field_element_from_bytes_be(b: &[u8]) -> FieldElement {
```
ntn-x2: Should we just drop the last part, and call it `field_element_from_bytes` instead?

ntn-x2 (approved):
I left my 2 cents on what we should still do to avoid inconsistencies in the future. Feel free to tackle them in this PR, in a different one, or to make a ticket for it.

nomos-da/kzgrs/src/rs.rs
danielSanchezQ: This are different Fr (bls instead), so we do not have a custom wrapper.

nomos-da/kzgrs/src/common.rs
danielSanchezQ: Yes, will do, thanks!

PR #1755, opened 2025-09-26
feat(kms): Derive KMS Keys

strinnityk:
Replace manual `Encoding` and `SecuredKey` implementations for a macro. This macro, generates required code so the integration of future keys is automatic, as soon as they're added to `Key`.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/nomos-services/key-management-system-macros/src/lib.rs
+++ b/nomos-services/key-management-system-macros/src/lib.rs
@@ -183,30 +183,20 @@ fn build_key_enum_impl_secured_key_method_sign(
 fn build_key_enum_impl_secured_key_method_as_public_key(
     key_enum_variants: &Punctuated<Variant, Comma>,
 ) -> TokenStream {
-    let match_block = if key_enum_variants.is_empty() {
+    let as_public_key_arms = key_enum_variants.iter().map(|variant| {
+        let variant_ident = &variant.ident;
+        let variant_attributes_cfg: Vec<Attribute> = get_cfg_attributes(&variant.attrs).collect();
         quote! {
-            Err(Self::Error::NoKeysEnabled)
+            #(#variant_attributes_cfg)*
+            Self::#variant_ident(key) => Self::PublicKey::#variant_ident(key.as_public_key()),
         }
-    } else {
-        let as_public_key_arms = key_enum_variants.iter().map(|variant| {
-            let variant_ident = &variant.ident;
-            let variant_attributes_cfg: Vec<Attribute> = get_cfg_attributes(&variant.attrs).collect();
-            quote! {
-                #(#variant_attributes_cfg)*
-                Self::#variant_ident(key) => key.as_public_key().map(Self::PublicKey::#variant_ident),
-            }
-        });
+    });
 
-        quote! {
+    quote! {
+        fn as_public_key(&self) -> Self::PublicKey {
             match self {
                 #(#as_public_key_arms)*
             }
-        }
-    };
-
-    quote! {
-        fn as_public_key(&self) -> Result<Self::PublicKey, Self::Error> {
-            #match_block
         }
     }
 }
--- a/nomos-services/key-management-system/src/keys/secured_key.rs
+++ b/nomos-services/key-management-system/src/keys/secured_key.rs
@@ -8,5 +8,5 @@ pub trait SecuredKey: ZeroizeOnDrop {
     type Error;
 
     fn sign(&self, payload: &Self::Payload) -> Result<Self::Signature, Self::Error>;
-    fn as_public_key(&self) -> Result<Self::PublicKey, Self::Error>;
+    fn as_public_key(&self) -> Self::PublicKey;
 }
--- a/nomos-services/key-management-system/src/backend/preload.rs
+++ b/nomos-services/key-management-system/src/backend/preload.rs
@@ -127,12 +127,9 @@ mod keys {
             }
         }
 
-        fn as_public_key(&self) -> Result<Self::PublicKey, Self::Error> {
+        fn as_public_key(&self) -> Self::PublicKey {
             match self {
-                Self::Ed25519(key) => key
-                    .as_public_key()
-                    .map(PreloadEncoding::from)
-                    .map_err(Into::into),
+                Self::Ed25519(key) => PreloadEncoding::from(key.as_public_key()),
             }
         }
     }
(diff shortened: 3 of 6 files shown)


nomos-services/key-management-system/src/keys/mod.rs
```diff
@@ -1,47 +1,28 @@
+#[cfg(any(feature = "key-ed25519", test))]
```
ntn-x2: Just a stylistic comment: do you think it would be more clear to have pub mods first, then newline, then feature-gated mods?

nomos-services/key-management-system/src/keys/secured_key.rs
```diff
@@ -8,5 +8,5 @@ pub trait SecuredKey: ZeroizeOnDrop {
     type Error;
 
     fn sign(&self, payload: &Self::Payload) -> Result<Self::Signature, Self::Error>;
-    fn as_public_key(&self) -> Self::PublicKey;
+    fn as_public_key(&self) -> Result<Self::PublicKey, Self::Error>;
```
ntn-x2: Why has this become an result now?

danielSanchezQ: :heavy_plus_sign: 1

nomos-services/key-management-system-macros/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+
+fn build_key_enum_impl_secured_key_method_sign(
+    key_enum_variants: &Punctuated<Variant, Comma>,
+) -> TokenStream {
+    let match_block = if key_enum_variants.is_empty() {
+        quote! {
+            Err(Self::Error::NoKeysEnabled)
+        }
```
danielSanchezQ, line 149: Is this so it can be used in an empty enum?

nomos-services/key-management-system-macros/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+                #(#as_public_key_arms)*
+            }
+        }
+    };
+
+    quote! {
+        fn as_public_key(&self) -> Result<Self::PublicKey, Self::Error> {
+            #match_block
+        }
+    }
```
danielSanchezQ: So, we changed the api of the trait to support empty enums. IMO is completely unnecesary. I would rather panic on empty enum so the compilation fails.

nomos-services/key-management-system/src/keys/mod.rs
danielSanchezQ: Maybe this is actually re-arranged by fmt? Not sure.

strinnityk: It's reordered by `fmt`.

ntn-x2: An empty line solves exactly that, to avoid fmt arranges them alphabetically. I think we should distinguish between those import types.

nomos-services/key-management-system-macros/src/lib.rs
strinnityk: Correct. 
I'll give the full answer in the thread above.

nomos-services/key-management-system-macros/src/lib.rs
strinnityk: I'll give the full answer in the thread above.

nomos-services/key-management-system/src/keys/secured_key.rs
strinnityk: The reason is, as Dani mentioned in another comment, to allow empty enums. That is, the case where no keys have been enabled.
At first I treated this situation (no keys enabled) as a compilation error in the `derive`, but `cargo hack` (with `--no-default-features`) complained, and I thought it could make a tiny bit of sense to support backends without keys.
I considered `panic!`-ing, but that didn’t sit right: if you’re explicitly allowing a backend with no keys, then panicking when someone tries to sign without one makes no sense.
Thus the error variant was born and, with it, the need for a `Result`.

danielSanchezQ: I would rather avoid that because:
* It makes no sense to change the API to support for the derive case
* An empty set of keys is completely useless
* If the user really want to have an empty set he can implement it himself

my 2 cnts

ntn-x2: What about feature-gating the whole enum, so that it only exists if any of the features is enabled? This would be completely transparent to consumers, of course, but I think it would be more ergonomic than returning a `Result` when calling `as_public_key`. It's not risky to feature-gate the enum, since if we forget to add a new key type to it, we get compilation error.

danielSanchezQ: > What about feature-gating the whole enum, so that it only exists if any of the features is enabled? This would be completely transparent to consumers, of course, but I think it would be more ergonomic than returning a `Result` when calling `as_public_key`. It's not risky to feature-gate the enum, since if we forget to add a new key type to it, we get compilation error.

Isn't gated code, being inner to the derive skiped already? Otherwise the macro could take care of it skipping implementation for featured gated keys.

ntn-x2: The `Key` enum lives regardless of the macro annotation, and as far as I know a derive macro cannot remove tokens, but only augment them. If it were to be a procedural macro, I think we could do that, but in this case I think the only option is to feature-gate the enum declaration manually.

danielSanchezQ: > The `Key` enum lives regardless of the macro annotation, and as far as I know a derive macro cannot remove tokens, but only augment them. If it were to be a procedural macro, I think we could do that, but in this case I think the only option is to feature-gate the enum declaration manually.

But I mean, if we know a certain enum varian is feature gated, we could simply add the same feature gate for all entries when deriving.
So either the variants are fed depending the features, or we are able to see all variants and feature them ourselves.

ntn-x2: But if I understood correctly, the problem is when there is no feature enabled, hence the enum is empty? If it's just the implementation of the trait, then what you are suggesting would work, although we would have an enum with no fields that is still public, which is kinda ugly, in my opinion.

nomos-services/key-management-system/src/keys/mod.rs
strinnityk: Oh, I thought it reordered them anyway. Nice.

nomos-services/key-management-system/src/keys/secured_key.rs
danielSanchezQ: > But if I understood correctly, the problem is when there is no feature enabled, hence the enum is empty?

~~Ah, ok, I got what you mean now, yes, in that case it shoul be featured gated with `any`.~~ In fact, if it is empty the best solution is not to derive anything and just it.

PR #1756, opened 2025-09-26
feat(mempool): transactions storage

andrussal:

This PR adds transactions storage to mempool.  With this change, mempool only keeps transactions hashes in memory and full transactions in storage.

**Note**: This change comes with significant amount of generics changes which is the reason of large amount of changed files.

### Storage service

Storage service has new functions to manage transactions storage:

- **store_transactions**
- **get_transactions**
- **remove_transactions**

### Mempool service

- Added `StorageAdapter`
- Only keys(transaction hashes) are in memory, full transactions are stored in RocksDb
- Instead of iterator, `View` returns stream

### Chain service

 - Added `MempoolAdapter`
 - Adapted to use stream instead of iterator for `View`
 
 ### Chain leader service
 
 - Added `MempoolAdapter`
 
 ### Nomos core
 
 - `tx_selector` adapted to use stream
 
 ### nomos_validator, nomos_executor, wallet, api
 
  - Generics changes, mostly related to mempool storage adapter

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


nomos-core/chain-defs/src/mantle/tx.rs
```diff
@@ -59,6 +60,11 @@ impl TxHash {
     pub fn as_signing_bytes(&self) -> Bytes {
         self.0.0.0.iter().flat_map(|b| b.to_le_bytes()).collect()
     }
+
+    #[must_use]
+    pub fn as_bytes(&self) -> [u8; 32] {
+        fr_to_bytes(&self.0)
+    }
```
danielSanchezQ: Should we implement hash for txhash instead? so it can be used in maps and stuff?

nomos-services/chain/chain-leader/src/mempool/mod.rs
```diff
@@ -0,0 +1,15 @@
+use std::pin::Pin;
+
+use async_trait::async_trait;
+use futures::Stream;
+use nomos_core::header::HeaderId;
+
+pub mod adapter;
+
+#[async_trait]
+pub trait MempoolAdapterTrait<Tx>: Send + Sync {
```
danielSanchezQ: cant be named MempoolAdapter here, and them imported as `use ..::MempoolAdapter as MempoolAdapterTrait` instead? Its quite reduntand to name a trait ..trait.

nomos-services/chain/chain-service/src/mempool/mod.rs
```diff
@@ -0,0 +1,13 @@
+use async_trait::async_trait;
+use nomos_core::{header::HeaderId, mantle::TxHash};
+
+pub mod adapter;
+
+#[async_trait]
+pub trait MempoolAdapterTrait<Tx>: Send + Sync {
```
danielSanchezQ: ➕ 1

nomos-services/mempool/src/lib.rs
```diff
@@ -39,6 +39,10 @@ pub enum MempoolMsg<BlockId, Payload, Item, Key> {
         items: Vec<Key>,
         reply_channel: Sender<Vec<Status<BlockId>>>,
     },
+    GetTransactionsByHashes {
+        hashes: Vec<Key>,
```
danielSanchezQ: Maybe a hashset instead? Or if we want to keep order a Btreeset

nomos-services/storage/src/backends/mock.rs
```diff
@@ -153,6 +157,24 @@ impl StorageChainApi for MockStorage {
...
+        &mut self,
+        _tx_hashes: &[TxHash],
+    ) -> Result<Pin<Box<dyn Stream<Item = Self::Tx> + Send>>, Self::Error> {
+        unimplemented!()
+    }
+
+    async fn remove_transactions(&mut self, _tx_hashes: &[TxHash]) -> Result<(), Self::Error> {
+        unimplemented!()
+    }
 }
```
danielSanchezQ: Do we need the mockstorage anymore? I think mockstorage could be the old inmemory that we had before(?)

danielSanchezQ (commented):
Nice, a lot of changes tho.

nomos-core/chain-defs/src/mantle/tx.rs
andrussal: It seems to derive hash though. But I removed this stuff and added `From<TxHash> for Bytes` which seems a bit easier to use for storage key.

nomos-services/chain/chain-leader/src/mempool/mod.rs
andrussal: > cant be named MempoolAdapter here, and them imported as `use ..::MempoolAdapter as MempoolAdapterTrait` instead? Its quite reduntand to name a trait ..trait.

Yup, good idea!

nomos-services/mempool/src/lib.rs
andrussal: I went with `BtreeSet`, it makes it simpler to keep it ordered.

nomos-services/storage/src/backends/mock.rs
andrussal: > Do we need the mockstorage anymore? I think mockstorage could be the old inmemory that we had before(?)

This wasn't used anywhere. We probably can just use RocksDb directly if we need some db in tests.

nomos-core/chain-defs/src/mantle/tx.rs
```diff
@@ -310,6 +317,22 @@ impl<'de> Deserialize<'de> for SignedMantleTx {
...
+}
+
+impl TryFrom<Bytes> for SignedMantleTx {
+    type Error = crate::codec::Error;
+
+    fn try_from(bytes: Bytes) -> Result<Self, Self::Error> {
+        <Self as SerdeOp>::deserialize(&bytes)
+    }
+}
+
```
danielSanchezQ: Uhm. This may misslead the code. If SerdeOp is already implemented for `SignedMantleTx` maybe we can make the specific call for bytes in place? I would want to avoid using this bytes conversion for something else. Then changes the serdeop and break the other side of things. Unless we consider SerdeOp the canonical bytes conversion, which I dont think it should.

nomos-services/storage/src/api/backend/rocksdb/chain.rs
```diff
@@ -99,10 +106,75 @@ impl StorageChainApi for RocksBackend {
...
+            }
+        });
+
+        Ok(Box::pin(stream))
+    }
+
+    async fn remove_transactions(&mut self, tx_hashes: &[TxHash]) -> Result<(), Self::Error> {
+        let keys: Vec<Bytes> = tx_hashes.iter().map(|&tx_hash| tx_hash.into()).collect();
+
+        let txn = self.txn(move |db| {
```
danielSanchezQ: what does txn means? Lets use explicit meaning naming please.

nomos-services/tx-service/src/backend/pool.rs
```diff
@@ -0,0 +1,221 @@
...
+use crate::{
+    backend::{MemPool, MempoolError, RecoverableMempool},
+    storage::MempoolStorageAdapter,
+};
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(bound(serialize = "BlockId: Serialize, Key: Serialize"))]
+#[serde(bound(
+    deserialize = "BlockId: serde::de::DeserializeOwned + Hash + Eq, Key: serde::de::DeserializeOwned + Hash + Eq + Ord"
+))]
```
danielSanchezQ: Is this necessary? 🤔 Aren't the bounds required by derive already?

nomos-services/tx-service/src/backend/pool.rs
```diff
@@ -0,0 +1,221 @@
...
+        if let Err(e) = self
+            .storage_adapter
+            .store_item(key.clone(), item.into())
+            .await
+        {
+            tracing::warn!("Failed to store item in storage: {:?}", e);
+        }
+
+        self.pending_items.insert(key);
+        self.last_item_timestamp = timestamp;
```
danielSanchezQ, line 110: I don't get how this is used. Also why is not part of the state (?)

danielSanchezQ:
> @danielSanchezQ Please take a look again if this is ready to go

Sorry, I was just in the middle of reviewing it again. I left some comments.

nomos-core/chain-defs/src/mantle/tx.rs
andrussal: Good point, this is old approach. We should remove also for other types in a separate PR I think.

nomos-services/storage/src/api/backend/rocksdb/chain.rs
andrussal: I gave better naming now.

nomos-services/tx-service/src/backend/pool.rs
andrussal: > Is this necessary? 🤔 Aren't the bounds required by derive already?

This is leftover from old code. I simplified now.

nomos-core/chain-defs/src/mantle/tx.rs
danielSanchezQ: Let's open an issue for it.

nomos-services/tx-service/src/backend/pool.rs
andrussal: This is actually added to `PoolRecoveryState` and also returned as part of `MempoolMetrics`. It doesn't seem currently used by tests but might become handy later.

nomos-core/chain-defs/src/mantle/tx.rs
andrussal: Created issue: https://github.com/logos-co/nomos/issues/1791

PR #1757, opened 2025-09-29
feat(chain): mark the service as ready after bootstrapping is done

### LINKED ISSUE #1656: Disable proposing blocks until bootstrapping is done
Mark the chain service ready after IBD and ProlongedBootstrapPeriod are done. Then, the chain-leader service will be notified and start proposing blocks.

youngjoon-lee:

Closes #1656 

This PR makes the chain service mark itself as ready only after it has fully completed bootstrapping (**IBD + ProlongedBootstrapPeriod**).
In other words, the chain service becomes ready only after **IBD** is done and Cryptarchia switches to **Online**. 

We made this decision so that other services become operational only after the **LIB** is available.

There are currently 3 services that explicitly wait until the chain service becomes ready.
- chain-leader
- wallet
  - It wasn't calling `wait_until_services_are_ready`. This PR adds it.
- api
  - But we may want this service to begin its job earlier even before the chain is ready. I didn’t do it in this PR, as it's not urgent and requires other fixes alongside. If we need it, we can do it nicely by the refactoring that I mentioned below.

In a next PR, I will refactor the chain service into two types (Bootstrapping and Online) to clearly separate their behavior. I didn't do the refactoring in this PR to show the diffs clearly.

Yes

@zeegomo @youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/wallet/src/lib.rs
+++ b/nomos-services/wallet/src/lib.rs
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+use std::{collections::HashSet, time::Duration};
 
 use async_trait::async_trait;
 use chain_service::{
@@ -20,6 +20,7 @@ use overwatch::{
     },
 };
 use serde::{Serialize, de::DeserializeOwned};
+use services_utils::wait_until_services_are_ready;
 use tokio::sync::oneshot;
 use tracing::{debug, error, info, trace};
 use wallet::{Wallet, WalletBlock, WalletError};
@@ -115,6 +116,14 @@ where
             mut service_resources_handle,
             ..
         } = self;
+
+        wait_until_services_are_ready!(
+            &service_resources_handle.overwatch_handle,
+            Some(Duration::from_secs(60)),
+            nomos_storage::StorageService<_, _>,
+            Cryptarchia
+        )
+        .await?;
 
         let settings = service_resources_handle
             .settings_handle
--- a/tests/src/tests/cryptarchia/bootstrap.rs
+++ b/tests/src/tests/cryptarchia/bootstrap.rs
@@ -20,7 +20,7 @@ use tests::{
 #[tokio::test]
 #[serial]
 async fn test_ibd_behind_nodes() {
-    let n_validators = 4;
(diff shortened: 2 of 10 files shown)


tests/src/tests/cryptarchia/bootstrap.rs
```diff
@@ -46,23 +46,10 @@ async fn test_ibd_behind_nodes() {
...
-
-    // Try to spawn the node - it should fail during IBD when peers are
-    // bootstrapping
-    if let Ok(mut node) = Validator::spawn(config).await {
-        assert!(
-            node.wait_for_exit(Duration::from_secs(5)).await,
-            "Expected node to fail during IBD when peers are bootstrapping"
-        );
-    }
-
```
youngjoon-lee, line 60: I disabled this check:
- because this is not the main point of this test.
- because this doesn't work anymore.
  - By the time `Validator::spawn` completes for the two initial validators, they are already in the Online state, since their API service only becomes active at that point and `Validator::spawn` waits for it. Therefore, this check always fails, since it's intended to run the 3rd node before the two initial nodes become Online.

Anyways, this check must be done in the unit tests of the chain service. I will add unit tests after doing refactoring I mentioned in the PR description. (My refactoring will handle this case more safely through the type system).

nomos-services/wallet/src/lib.rs
```diff
@@ -116,6 +117,14 @@ where
             ..
         } = self;
 
+        wait_until_services_are_ready!(
+            &service_resources_handle.overwatch_handle,
+            Some(Duration::from_secs(60)),
+            nomos_storage::StorageService<_, _>,
+            Cryptarchia
+        )
+        .await?;
```
davidrusu, line 126: Thanks, I missed this in the initial implementation.

davidrusu (approved):

nomos-services/wallet/src/lib.rs
danielSanchezQ: Nice catch.

PR #1759, opened 2025-09-29
feat(ledger): Initialize ledger from genesis

zeegomo:
Initialize ledger from genesis block, following https://www.notion.so/nomos-tech/Bedrock-Genesis-Block-21d261aa09df80bb8dc3c768802eb527. 
This also moves the signature check on channel ops out of the ledger since those proofs can be checked at deserialization time and to uniform with other operations. I'll open a new PR to do this soon.
A slight difference from the spec is that the code does not try to validate the contents of the inscription, since 1) it's not strictly needed since the inscription does not carry logic anyway 2) json encoding is not universally specified (e.g. formatting, pretty print, ...) so it might be better to move it a higher level. We can ofc add it back it if we think it's worth it.

Yes

@davidrusu author, @zeegomo implementing
Yes
Does not validate the inscription contents
## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/ledger/nomos-ledger/src/mantle/mod.rs
+++ b/ledger/nomos-ledger/src/mantle/mod.rs
@@ -259,14 +259,16 @@ mod tests {
     }
 
     fn create_multi_signed_tx(ops: Vec<Op>, signing_keys: Vec<&SigningKey>) -> SignedMantleTx {
-        let mut tx = create_test_tx_with_ops(ops);
+        let mut tx = create_test_tx_with_ops(ops.clone());
         let tx_hash = tx.hash();
         tx.ops_proofs = signing_keys
             .into_iter()
-            .map(|key| {
-                Some(OpProof::Ed25519Sig(
+            .zip(ops)
+            .map(|(key, op)| match op {
+                Op::ChannelSetKeys(_) => Some(OpProof::Ed25519Sig(
                     key.sign(tx_hash.as_signing_bytes().as_ref()),
-                ))
+                )),
+                _ => None,
             })
             .collect();
         tx
@@ -358,34 +360,6 @@ mod tests {
     }
 
     #[test]
-    fn test_invalid_signature_error() {
-        let cryptarchia_state = genesis_state(&[utxo()]);
-        let test_config = config();
-        let ledger_state = LedgerState::new();
-        let (_, verifying_key) = create_test_keys_with_seed(1);
-        let (wrong_signing_key, _) = create_test_keys_with_seed(2);
-        let channel_id = ChannelId::from([4; 32]);
-
-        let blob_op = BlobOp {
-            channel: channel_id,
-            blob: [42; 32],
-            blob_size: 1024,
-            da_storage_gas_price: 10,
-            parent: MsgId::root(),
-            signer: verifying_key,
-        };
-
-        let tx = create_signed_tx(Op::ChannelBlob(blob_op), &wrong_signing_key);
-        let result = ledger_state.try_apply_tx::<MainnetGasConstants>(
-            0,
-            &test_config,
-            cryptarchia_state.latest_commitments(),
-            tx,
-        );
-        assert_eq!(result, Err(Error::InvalidSignature));
-    }
-
-    #[test]
     fn test_unsupported_operation() {
         let cryptarchia_state = genesis_state(&[utxo()]);
         let test_config = config();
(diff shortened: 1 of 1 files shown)


ledger/nomos-ledger/src/mantle/mod.rs
```diff
@@ -90,20 +99,18 @@ impl LedgerState {
...
-                (Op::ChannelBlob(op), Some(OpProof::Ed25519Sig(sig))) => {
-                    // these proofs could be verified even before reaching this point
-                    // as you only need the op itself to validate the signature
-                    op.signer
-                        .verify(tx_hash.as_signing_bytes().as_ref(), sig)
-                        .map_err(|_| Error::InvalidSignature)?;
+                (Op::ChannelBlob(op), None) => {
+                    // The signature could be verified even before reaching this point,
+                    // as you only need the signer's public key and tx hash
+                    // Callers are expected to validate the proof before calling this function.
```
danielSanchezQ, line 105: Can we add an assert here for the verification just for debug builds? wdyt?

nomos-core/chain-defs/src/mantle/genesis_tx.rs
```diff
@@ -0,0 +1,390 @@
...
+            Some(Op::ChannelInscribe(op)) => validate_inscription_op(op)?,
+            _ => return Err(Error::MissingInscription),
+        }
+
+        for op in ops {
+            match op {
+                Op::SDPDeclare(_) => {}
+                _ => return Err(Error::UnsupportedGenesisOp(Box::new(op.clone()))),
+            }
+        }
```
danielSanchezQ, line 63:
```rust
if ops.iter().any(|op| !matches!(op, Op::SDPDeclare(_))) {
    return Error()
}
```
Or if we want to report txs, I would check for all the ones that are not allowed instead of just the first one (?)

ledger/nomos-ledger/src/mantle/mod.rs
zeegomo: In the next PR I'll enforce this at the type level for, so I think no need for an additional debug check here

nomos-core/chain-defs/src/mantle/genesis_tx.rs
zeegomo: not sure I get what you mean, could you elaborate?

danielSanchezQ: > not sure I get what you mean, could you elaborate?

So if somehow we messed up more than a single tx only the first one will be reported. I would take all of the ones that should not be there to be reported instead.

zeegomo: That should probably never happen but yes, we can add it

danielSanchezQ (approved):

nomos-core/chain-defs/src/mantle/genesis_tx.rs
```diff
@@ -0,0 +1,391 @@
...
+    }
+
+    if inscription.signer.as_bytes() != &[0; 32] {
+        return Err(Error::InvalidInscription(Box::new(Op::ChannelInscribe(
+            inscription.clone(),
+        ))));
+    }
+
+    Ok(())
+}
```
davidrusu: two things,

1. This inscription is specifically an inscription to initialize cryptarchia. So we can make that concrete by renaming this to `validate_cryptarchia_inscription()`

2. We should validate that the inscription payload `json::decodes` into a struct like:

```rust
struct CryptarchiaGenesisParams {
    chain_id: String,
    genesis_time: DateTime,
    genesis_epoch_nonce: Nonce,
}
```

nomos-core/chain-defs/src/mantle/genesis_tx.rs
```diff
@@ -0,0 +1,391 @@
...
+        // TODO: check inscription with non-zero signer fails
+        let tx = create_signed_tx(vec![Op::ChannelInscribe(inscription_op(
+            ChannelId::from([0; 32]),
+            MsgId::root(),
+            VerifyingKey::from_bytes(&[1; 32]).unwrap(),
+        ))]);
+        assert!(matches!(
+            GenesisTx::from_tx(tx),
+            Err(Error::InvalidInscription(_))
+        ));
```
davidrusu: is the `TODO` done? it looks like this test is checking the non-zero signer case

nomos-core/chain-defs/src/mantle/genesis_tx.rs
```diff
@@ -0,0 +1,391 @@
...
+
+        // TODO: check valid inscription passes
+        let tx = create_signed_tx(vec![Op::ChannelInscribe(inscription_op(
+            ChannelId::from([0; 32]),
+            MsgId::root(),
+            VerifyingKey::from_bytes(&[0; 32]).unwrap(),
+        ))]);
+        assert!(GenesisTx::from_tx(tx).is_ok());
```
davidrusu: same here, `TODO` is done?

nomos-core/chain-defs/src/mantle/genesis_tx.rs
davidrusu: discussed offline, cryptarchia params should probably have a canonical encoding. We'll revisit that in the spec

davidrusu (approved):

PR #1761, opened 2025-09-30
fix(blend): use little-endian

youngjoon-lee:

Replaced all big-endian in blend crates with little-endian.

CI will pass after we revert circuits to the version that uses little-endian.

Yes

spec: @thomaslavaur @madxor 
impl: @ntn-x2 @youngjoon-lee 

Needs to be updated.

- https://www.notion.so/nomos-tech/Common-Cryptographic-Components-1fd261aa09df81ac8ebbe0111e2c2d84?source=copy_link#209261aa09df80b8aec6cc763573ff69
  - Use little-endian 
- https://www.notion.so/nomos-tech/Proof-of-Quota-Specification-215261aa09df81d88118ee22205cbafe?source=copy_link#25e261aa09df802d87edfc54d1d60b80
  - Update Fr values

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/message/src/crypto/proofs/quota/inputs/prove/mod.rs
+++ b/nomos-blend/message/src/crypto/proofs/quota/inputs/prove/mod.rs
@@ -1,4 +1,4 @@
-use num_bigint::BigUint;
+use groth16::fr_from_bytes;
 use poq::{
     PoQBlendInputsData, PoQChainInputsData, PoQCommonInputsData, PoQInputsFromDataError,
     PoQWalletInputsData, PoQWitnessInputs,
@@ -37,8 +37,12 @@ impl TryFrom<Inputs> for PoQWitnessInputs {
             index: value.private.key_index,
             leader_quota: value.public.leader_quota,
             message_key: (
-                BigUint::from_bytes_le(&signing_key_first_half[..]).into(),
-                BigUint::from_bytes_le(&signing_key_second_half[..]).into(),
+                fr_from_bytes(&signing_key_first_half[..]).expect(
+                    "First half of signing public key does not represent a valid `Fr` point.",
+                ),
+                fr_from_bytes(&signing_key_second_half[..]).expect(
+                    "Second half of signing public key does not represent a valid `Fr` point.",
+                ),
             ),
             selector: value.private.selector,
         };
--- a/nomos-blend/message/src/crypto/proofs/quota/mod.rs
+++ b/nomos-blend/message/src/crypto/proofs/quota/mod.rs
@@ -2,9 +2,8 @@ use std::sync::LazyLock;
 
 use ::serde::{Deserialize, Serialize};
 use generic_array::{ArrayLength, GenericArray};
-use groth16::{Bn254, CompressSize, fr_from_bytes_unchecked};
+use groth16::{Bn254, CompressSize, fr_from_bytes, fr_from_bytes_unchecked};
 use nomos_core::crypto::ZkHash;
-use num_bigint::BigUint;
 use poq::{PoQProof, PoQVerifierInput, PoQWitnessInputs, ProveError, prove, verify};
 use thiserror::Error;
 
@@ -124,8 +123,10 @@ impl ProofOfQuota {
 }
 
 const DOMAIN_SEPARATION_TAG: [u8; 23] = *b"SELECTION_RANDOMNESS_V1";
-static DOMAIN_SEPARATION_TAG_FR: LazyLock<ZkHash> =
-    LazyLock::new(|| BigUint::from_bytes_le(&DOMAIN_SEPARATION_TAG[..]).into());
+static DOMAIN_SEPARATION_TAG_FR: LazyLock<ZkHash> = LazyLock::new(|| {
+    fr_from_bytes(&DOMAIN_SEPARATION_TAG[..])
+        .expect("DST for secret selection randomness calculation must be correct.")
+});
 // As per Proof of Quota v1 spec: <https://www.notion.so/nomos-tech/Proof-of-Quota-Specification-215261aa09df81d88118ee22205cbafe?source=copy_link#215261aa09df81adb8ccd1448c9afd68>.
 fn generate_secret_selection_randomness(sk: ZkHash, key_index: u64, session: u64) -> ZkHash {
     [
--- a/nomos-blend/message/src/crypto/proofs/quota/tests.rs
+++ b/nomos-blend/message/src/crypto/proofs/quota/tests.rs
@@ -1,5 +1,5 @@
 use const_hex::FromHex as _;
-use num_bigint::BigUint;
+use groth16::fr_from_bytes_unchecked;
 
 use crate::crypto::proofs::{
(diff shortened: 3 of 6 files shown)


nomos-blend/message/src/crypto/proofs/quota/inputs/prove/mod.rs
```diff
@@ -37,12 +37,8 @@ impl TryFrom<Inputs> for PoQWitnessInputs {
             index: value.private.key_index,
             leader_quota: value.public.leader_quota,
             message_key: (
-                fr_from_bytes(&signing_key_first_half[..]).expect(
-                    "First half of signing public key does not represent a valid `Fr` point.",
-                ),
-                fr_from_bytes(&signing_key_second_half[..]).expect(
-                    "Second half of signing public key does not represent a valid `Fr` point.",
-                ),
+                BigUint::from_bytes_le(&signing_key_first_half[..]).into(),
```
ntn-x2: Why this change? I think it's correct to use `fr_from_bytes`. If we know this is a valid `Fr` value, we can `expect`, but we should use the utility function exposed by the `groth16` crate.

nomos-blend/message/src/crypto/proofs/quota/mod.rs
```diff
@@ -123,10 +124,8 @@ impl ProofOfQuota {
 }
 
 const DOMAIN_SEPARATION_TAG: [u8; 23] = *b"SELECTION_RANDOMNESS_V1";
-static DOMAIN_SEPARATION_TAG_FR: LazyLock<ZkHash> = LazyLock::new(|| {
-    fr_from_bytes(&DOMAIN_SEPARATION_TAG[..])
-        .expect("DST for secret selection randomness calculation must be correct.")
-});
+static DOMAIN_SEPARATION_TAG_FR: LazyLock<ZkHash> =
+    LazyLock::new(|| BigUint::from_bytes_le(&DOMAIN_SEPARATION_TAG[..]).into());
```
ntn-x2: Ditto

nomos-blend/message/src/crypto/proofs/quota/tests.rs
```diff
@@ -14,8 +14,8 @@ fn secret_selection_randomness_dst_encoding() {
     // Blend spec: <https://www.notion.so/nomos-tech/Proof-of-Quota-Specification-215261aa09df81d88118ee22205cbafe?source=copy_link#25e261aa09df802d87edfc54d1d60b80>
     assert_eq!(
         *DOMAIN_SEPARATION_TAG_FR,
-        BigUint::from_bytes_be(
-            &<[u8; 23]>::from_hex("0x31565f5353454e4d4f444e41525f4e4f495443454c4553").unwrap()
+        BigUint::from_bytes_le(
```
ntn-x2: Actually this should also use `fr_from_bytes`, sorry I missed this one.

nomos-blend/message/src/crypto/proofs/selection/mod.rs
```diff
@@ -117,9 +117,7 @@ impl ProofOfSelection {
 
 const KEY_NULLIFIER_DERIVATION_DOMAIN_SEPARATION_TAG: [u8; 16] = *b"KEY_NULLIFIER_V1";
 static KEY_NULLIFIER_DERIVATION_DOMAIN_SEPARATION_TAG_FR: LazyLock<ZkHash> = LazyLock::new(|| {
-    fr_from_bytes(&KEY_NULLIFIER_DERIVATION_DOMAIN_SEPARATION_TAG[..]).expect(
-        "DST for key nullifier derivation from secret selection randomness must be correct.",
-    )
+    BigUint::from_bytes_le(&KEY_NULLIFIER_DERIVATION_DOMAIN_SEPARATION_TAG[..]).into()
```
ntn-x2: Ditto

nomos-blend/message/src/crypto/proofs/selection/tests.rs
```diff
@@ -11,10 +11,11 @@ use crate::crypto::proofs::selection::{
 #[test]
 fn secret_selection_randomness_to_key_nullifier_dst_encoding() {
     // Blend spec: <https://www.notion.so/nomos-tech/Proof-of-Quota-Specification-215261aa09df81d88118ee22205cbafe?source=copy_link#26c261aa09df802f9dbcd0780dc7ac6e>
+
     assert_eq!(
         *KEY_NULLIFIER_DERIVATION_DOMAIN_SEPARATION_TAG_FR,
-        BigUint::from_bytes_be(
-            &<[u8; 16]>::from_hex("0x31565f52454946494c4c554e5f59454b").unwrap()
+        BigUint::from_bytes_le(
```
ntn-x2: This also should be `fr_from_bytes`

nomos-blend/message/src/crypto/proofs/quota/inputs/prove/mod.rs
youngjoon-lee: I adopted the fix made in https://github.com/logos-co/nomos/pull/1753. If you think we don't need this change, I will revert.

nomos-blend/message/src/crypto/proofs/quota/tests.rs
ntn-x2: Ah yeah I missed it because we decided not to export BE and LE from the `groth16` crate, so doing BE stuff would involve doing things manually, which is now not the case anymore.

nomos-blend/message/src/crypto/proofs/quota/inputs/prove/mod.rs
ntn-x2: No I think we can now use `fr_from_bytes` everywhere. Actually, for a public key, there is a question whether we want to use `from_bytes_unchecked`, since there can be values that fall outside of the `Fr` range. I will ask on Discord.

ntn-x2: Nevermind, we are transforming a 16 byte array into an `Fr`, so this operation is guaranteed to succeed. We can use `.expect` in the result since we know it will always succeed.

youngjoon-lee: Got it. I saw that `fr_from_bytes` uses `BigUint::from_bytes_le` internally. Thanks. 
I reverted this change.

nomos-blend/message/src/crypto/proofs/quota/tests.rs
youngjoon-lee: Since this is a test, I chose to use `fr_from_bytes_unchecked` in the new commit. Actually, I think we can use it for other part of the code but didn't touch it since I'm not sure if there's anything I'm missing.

ntn-x2: Mhh the only other part where we could use it is for DSTs and public key split up. Nevertheless, I think it's better to leave `expect`s, and use unchecked only for tests, in my opinion.

PR #1764, opened 2025-09-30
optimize DA operations

thomaslavaur:

Optimize already implemented Fr operations.

Yes

@thomaslavaur 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/nomos-da/kzgrs/src/kzg.rs
+++ b/nomos-da/kzgrs/src/kzg.rs
@@ -4,6 +4,7 @@ use std::{
 };
 
 use ark_bls12_381::{Bls12_381, Fr};
+use ark_ec::CurveGroup;
 use ark_ec::pairing::Pairing as _;
 use ark_poly::{
     DenseUVPolynomial as _, EvaluationDomain as _, GeneralEvaluationDomain,
@@ -12,7 +13,7 @@ use ark_poly::{
 use ark_poly_commit::kzg10::{Commitment, KZG10, Powers, Proof, UniversalParams};
 use num_traits::{One as _, Zero as _};
 
-use crate::{Evaluations, common::KzgRsError};
+use crate::common::KzgRsError;
 
 /// Commit to a polynomial where each of the evaluations are over `w(i)` for the
 /// degree of the polynomial being omega (`w`) the root of unity (2^x).
@@ -33,7 +34,6 @@ pub fn commit_polynomial(
 pub fn generate_element_proof(
     element_index: usize,
     polynomial: &DensePolynomial<Fr>,
-    evaluations: &Evaluations,
     global_parameters: &UniversalParams<Bls12_381>,
     domain: GeneralEvaluationDomain<Fr>,
 ) -> Result<Proof<Bls12_381>, KzgRsError> {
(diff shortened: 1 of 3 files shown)


nomos-da/kzgrs/benches/kzg.rs
```diff
@@ -23,11 +24,15 @@ fn main() {
 // #[global_allocator]
 // static ALLOC: AllocProfiler = AllocProfiler::system();
 
-static GLOBAL_PARAMETERS: LazyLock<UniversalParams<Bls12_381>> = LazyLock::new(|| {
+static PROVING_KEY: LazyLock<UniversalParams<Bls12_381>> = LazyLock::new(|| {
+    println!("WARNING: Proving key is randomly generated. Use for development only.");
```
danielSanchezQ, line 28: Was this a println before as well? We should have a log instead I think.

nomos-da/kzgrs/src/kzg.rs
```diff
@@ -64,15 +55,20 @@ pub fn verify_element_proof(
...
-    lhs == rhs
+    let commitment_check_g1 = verification_key.g.mul(v) - commitment.0 - proof.w.mul(u);
+    let qap = Bls12_381::multi_miller_loop(
+        [commitment_check_g1.into(), proof.w],
+        [
+            verification_key.prepared_h.clone(),
+            verification_key.prepared_beta_h.clone(),
+        ],
+    );
+    let test = Bls12_381::final_exponentiation(qap).unwrap();
```
danielSanchezQ: Use `expect` instead of `unwrap` explaining why it should succeed.

danielSanchezQ (changes requested):

nomos-da/kzgrs/benches/kzg.rs
thomaslavaur: Yes it was a print elsewhere

thomaslavaur: But it should be removed once we have the key loaded from the file

nomos-da/kzgrs/src/kzg.rs
thomaslavaur: Done!

nomos-da/kzgrs/benches/kzg.rs
bacv: I think it's fine to have this println in benches.

nomos-da/kzgrs-backend/src/kzg_keys.rs
```diff
@@ -0,0 +1,42 @@
...
+#[cfg(test)]
+mod tests {
+    use std::fs::File;
+
+    use ark_serialize::{CanonicalSerialize as _, Write as _};
+    use kzgrs::proving_key_from_randomness;
+
+    #[test]
+    #[ignore = "for testing purposes only"]
+    fn write_random_kzgrs_params_to_file() {
```
bacv, line 30: Previously we had this to easily create a file with global params, that were shared between testnet nodes. As I understand, there are no more global params, and each node will have their own proving and verifying key. Are you planning to add an option in node configs to read these keys from a file?

thomaslavaur: No it's exactly as before. The only difference is that you don't need to keep the poving key loaded if you are not a prover. What was previously called global_parameter is the proving key. The verification key is a sub_structure of the proving key and is way smaller. So previously we was loading the proving key and use it everywhere. Now we can still load the proving key (as before) but the verification key is loading a second time the same proving key, derive the verification key and throw away the useless part to only keep in memory the verification key.

So from the config file it's exactly the same as before, except that it's loaded twice, once for the proving key and once to load the proving key and derive the verification key

bacv: Got it! The format didn't change, so the already generated keys work, and tests are actually using them and pass. Thanks, marking 🟢

bacv (approved):

nomos-da/kzgrs/benches/kzg.rs
danielSanchezQ: Ah, this was benches. I missplaced it. Its fine the println in benches yes.

danielSanchezQ (approved):

Issue #1733, opened 2025-09-23
Integrate real public chain PoL data into Blend proofs


ntn-x2:
This will be fetched by the Blend service from the `chain-follower` service, once ready.

PR #1771, opened 2025-10-01
feat(blend): fetch epoch state on each new epoch

ntn-x2:

This is part of a series of PRs aiming at integrating actual PoL info into Blend for proof generation and verification.

In this PR, we follow the approach of the chain leader to fetch epoch state from chain service, with the difference that in this PR the Blend service does not directly depend on the chain service, because of its many generics required, but it rather delegates epoch state fetching logic to a trait that is implemented by the `nomos-node` binary, which waits for the chain service to be ready and then establishes an outbound channel to send those requests.

Fixes https://github.com/logos-co/nomos/issues/1733.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nodes/nomos-node/node/src/generic_services/blend.rs
+++ b/nodes/nomos-node/node/src/generic_services/blend.rs
@@ -20,7 +20,7 @@ use nomos_blend_message::crypto::{
 use nomos_blend_scheduling::message_blend::{BlendLayerProof, SessionInfo};
 use nomos_blend_service::{
     ProofOfLeadershipQuotaInputs, ProofsGenerator, ProofsVerifier,
-    epoch::{PolEpochInfo, PolInfoProvider as PolInfoProviderTrait},
+    epoch_info::{PolEpochInfo, PolInfoProvider as PolInfoProviderTrait},
     membership::service::Adapter,
 };
 use nomos_core::{codec::SerdeOp as _, crypto::ZkHash};
--- a/nomos-services/blend/src/core/mod.rs
+++ b/nomos-services/blend/src/core/mod.rs
@@ -52,7 +52,7 @@ use crate::{
         processor::{CoreCryptographicProcessor, Error},
         settings::BlendConfig,
     },
-    epoch::{EpochHandler, PolInfoProvider as PolInfoProviderTrait},
+    epoch_info::{EpochHandler, PolInfoProvider as PolInfoProviderTrait},
     membership,
     message::{NetworkMessage, ProcessedMessage, ServiceMessage},
     mock_poq_inputs_stream,
--- a/nomos-services/blend/src/edge/mod.rs
+++ b/nomos-services/blend/src/edge/mod.rs
@@ -39,7 +39,7 @@ use tracing::{debug, error, info};
 
 use crate::{
     edge::handlers::{Error, MessageHandler},
-    epoch::{ChainApi, EpochHandler, PolInfoProvider as PolInfoProviderTrait},
+    epoch_info::{ChainApi, EpochHandler, PolInfoProvider as PolInfoProviderTrait},
     membership,
     message::{NetworkMessage, ServiceMessage},
     mock_poq_inputs_stream,
--- a/nomos-services/blend/src/edge/tests/utils.rs
+++ b/nomos-services/blend/src/edge/tests/utils.rs
@@ -23,7 +23,7 @@ use tokio_stream::wrappers::ReceiverStream;
 
 use crate::{
     edge::{backends::BlendBackend, handlers::Error, run, settings::BlendConfig},
-    epoch::{EpochHandler, PolEpochInfo, PolInfoProvider},
+    epoch_info::{EpochHandler, PolEpochInfo, PolInfoProvider},
     mock_poq_inputs_stream,
     session::SessionInfo,
     settings::{FIRST_SESSION_READY_TIMEOUT, TimingSettings},
--- a/nomos-services/blend/src/epoch_info.rs
+++ b/nomos-services/blend/src/epoch_info.rs
@@ -161,7 +161,7 @@ mod tests {
     use test_log::test;
 
     use crate::{
-        epoch::EpochHandler,
+        epoch_info::EpochHandler,
         test_utils::epoch::{NON_EXISTING_EPOCH_STATE_SLOT, TestChainService, default_epoch_state},
     };
 
--- a/nomos-services/blend/src/lib.rs
+++ b/nomos-services/blend/src/lib.rs
@@ -43,7 +43,7 @@ use crate::{
 
 pub mod core;
 pub mod edge;
-pub mod epoch;
+pub mod epoch_info;
 pub mod membership;
 pub mod message;
 pub mod session;
--- a/nomos-services/blend/src/test_utils/epoch.rs
+++ b/nomos-services/blend/src/test_utils/epoch.rs
@@ -3,7 +3,7 @@ use chain_service::Slot;
 use nomos_ledger::EpochState;
 use overwatch::overwatch::OverwatchHandle;
 
-use crate::epoch::ChainApi;
+use crate::epoch_info::ChainApi;
 
 pub const NON_EXISTING_EPOCH_STATE_SLOT: Slot = Slot::new(4);
 


consensus/cryptarchia-engine/src/time.rs
```diff
@@ -20,7 +20,19 @@ pub struct Slot(u64);
...
+        Self(inner)
+    }
+}
+
 impl Slot {
+    #[must_use]
+    pub const fn new(inner: u64) -> Self {
+        Self(inner)
+    }
+
```
danielSanchezQ, line 35: I think we implemented `From` for this (iirc).

nodes/nomos-node/node/src/generic_services/blend.rs
```diff
@@ -115,27 +130,86 @@ fn loop_until_valid_proof(
...
+    PhantomData<CryptarchiaSamplingAdapter>,
+);
+
+impl<CryptarchiaSamplingAdapter, OutboundMsg> Clone
+    for ChainService<CryptarchiaSamplingAdapter, OutboundMsg>
+{
+    fn clone(&self) -> Self {
+        Self(self.0.clone(), self.1)
+    }
+}
```
danielSanchezQ: Why clone for a service?

nomos-services/blend/src/epoch.rs
```diff
@@ -0,0 +1,285 @@
...
+            _phantom: PhantomData,
+        }
+    }
+}
+
+#[expect(
+    clippy::cognitive_complexity,
+    reason = "It's better to keep all the polling logic together."
+)]
+impl<SlotStream, ChainService, RuntimeServiceId> Stream
```
danielSanchezQ: I have the feeling this could be simplified using stream operations instead of manual operations.

danielSanchezQ (commented):
This looks fine overall.

consensus/cryptarchia-engine/src/time.rs
ntn-x2: `From` is not `const`

nodes/nomos-node/node/src/generic_services/blend.rs
ntn-x2: It's not a service, it's a service adapter. And we box the future that is returned by one of its methods, so it required `Clone`.

danielSanchezQ: Lets name it adapter, it is really confusing otherwise :sweat_smile:

nomos-services/blend/src/epoch.rs
ntn-x2: I mean, the logic is a bit too much complicated to be performed entirely with stream operations. I thought about it, but arrived to the conclusion this would be actually easier to read. We poll the first time and on each epoch change, and on the first slot we fetch the epoch state, and retry on the next slot in case the previous attempt failed. I could not find an easy enough way to express this without a custom `Stream` implementation.

danielSanchezQ: Why not transform the slot_stream to an epoch stream? Then it should be simpler?
I trust your decision tho. If you say its better you know better.

ntn-x2: Yes, but we still need to tick on every slot, since unfortunately it's possible that `GetEpochState` returns `None` for a given slot, in which case I decided that we should try on the next slot tick again, instead of panicking. I can't think of a way to implement this logic in a more idiomatic way (?)

nomos-services/blend/src/epoch.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+}
+
+/// A trait that provides the needed functionalities for the epoch stream to
+/// fetch the epoch state for a given slot.
+#[async_trait]
+pub trait ChainApi<RuntimeServiceId> {
+    async fn new(overwatch_handle: &OverwatchHandle<RuntimeServiceId>) -> Self;
+    async fn get_epoch_state_for_slot(&self, slot: Slot) -> Option<EpochState>;
+}
```
davidrusu: Have you taken a look at `chain_service::api::CryptarchiaApi`? Why don't we use that here?

youngjoon-lee: I guess, it's to avoid circular dependency between blend and chain?

youngjoon-lee: I guess it's to not introduce a direct dependency from blend to chain-follower. But, I think it's fine to have it since we factored out the chain-leader. Wdyt, Antonio?

nomos-services/blend/src/epoch.rs
```diff
@@ -0,0 +1,343 @@
...
+                self.epoch_state_future = Some(Box::pin(async move {
+                    chain_service_clone.get_epoch_state_for_slot(slot).await
+                }));
+                tracing::debug!(target: LOG_TARGET, "Found new epoch unseen before. Polling for its state...");
+                // We wake so we poll for the next slot tick.
+                cx.waker().wake_by_ref();
+                Poll::Pending
+            }
+        }
+    }
```
davidrusu: an alternative:

```rust

struct EpochStateTicker<ChainService> {
    current_epoch: Epoch,
    chain_service_api: ChainService,
}

impl EpochStateTicker {
    fn tick(&mut self, slot_tick: SlotTick) -> Option<EpochInfo> {
        if self.current_epoch != slot_tick.epoch {
            let epoch_info = ... // fetch epoch state from chain servcie
            self.current_epoch = slot_tick.epoch
            return Some(epoch_info)
        }
        return None
    }
}
```

And then in edge and core we have

```
let epoch_ticker = EpochStateTicker::new(...);

loop {
    tokio::select! {
        Some(slot_tick) = slot_timer.next() => {
            if let Some(epoch_info) = epoch_ticker.tick(slot_tick) {
                // handle epoch transition
            }
        }
    }
}
```

can we do something like this instead?

ntn-x2: I think it should not be the responsibility of the Blend service as a whole to notify each of its sub-components for whatever they need. Having a `Stream` interface allows us to use all of the stream capabilities, for instance zipping it with other streams (we might want to zip this with the PoL stream coming from chain leader, for instance). Also, operations inside of it are `async`, e.g., fetching the epoch state from the chain service, and here we would need to `await` it, effectively preventing any other event from being processed in the `select` loop, while with a stream approach with have a polling interface that only yields whenever the whole flow is complete, meaning a relevant tick is received, the state is fetched from the chain service, and the state is not `None`. I don't think the Blend service should care for `None` cases, but that should be responsibility of the stream implementation to yield relevant values when necessary.

TL;DR I think having a self-contained stream interface for epoch info plays well with the other streams we have for other pieces, is all encapsulated logic, and removes the complexity from the Blend service, which just needs to create the stream and then poll it for values.

nomos-services/blend/src/epoch.rs
ntn-x2: If you mean [`CryptarchiaServiceApi`](https://github.com/logos-co/nomos/blob/eeb69596d0aeaabf97aecfef257d7abcfdac28c7/nomos-services/chain/chain-service/src/api.rs#L19), then as far as I see it depends on cryptarchia as well as its definition of `Tx`, which we don't need in Blend. It's probably fine for chain leader since they use the same type, but Blend does not know anything about transactions.

nomos-services/blend/src/epoch.rs
danielSanchezQ: But this can be composed as a single stream with @davidrusu suggestion, from the top of my head:

```rust
let epoch_ticker = slot_timer.scan(EpochStateTicker::new(...), |mut state, tick| state.tick(tick));
```

Disclaimer :warning: : I didn't try it.

ntn-x2:
I changed a few things around to test your proposal, @davidrusu, and I noticed these things.

1. Unit testing this has now become much more complicated. `CryptarchiaServiceApi`, which we cannot mock since it's a struct on its own, now needs an actual mocked runtime to open a relay with the cryptarchia service, which we also need to mock now. 
2. `fn tick(&mut self, slot_tick: SlotTick) -> Option<EpochInfo>` is now `async`, which means that, if for any reasons fetching the state takes long, we would block the tokio select loop of Blend. This is avoided by spawning a task, sure, but that already increases complexity, in my opinion. Also, we cannot call `tick` on `EpochStateTicker` until the background task has completed, and that is additional synchronization logic that all of the sudden Blend service needs to take care of.

If you guys think this is the way to go, I can work out the changes to make it work, but I wanted to highlight the downsides of this approach, before we decide we go ahead with this proposal. Also, it might be I am not seeing some other potential shortcomings that would become visible once we integrate this with the Blend proof generator/verifier.

Cc @danielSanchezQ

danielSanchezQ:
> I changed a few things around to test your proposal, @davidrusu, and I noticed these things.
> 
>     1. Unit testing this has now become much more complicated. `CryptarchiaServiceApi`, which we cannot mock since it's a struct on its own, now needs an actual mocked runtime to open a relay with the cryptarchia service, which we also need to mock now.
> 
>     2. `fn tick(&mut self, slot_tick: SlotTick) -> Option<EpochInfo>` is now `async`, which means that, if for any reasons fetching the state takes long, we would block the tokio select loop of Blend. This is avoided by spawning a task, sure, but that already increases complexity, in my opinion. Also, we cannot call `tick` on `EpochStateTicker` until the background task has completed, and that is additional synchronization logic that all of the sudden Blend service needs to take care of.
> 
> 
> If you guys think this is the way to go, I can work out the changes to make it work, but I wanted to highlight the downsides of this approach, before we decide we go ahead with this proposal. Also, it might be I am not seeing some other potential shortcomings that would become visible once we integrate this with the Blend proof generator/verifier.
> 
> Cc @danielSanchezQ

No, if its not an immediate win lets go with what we have. As we have other ideas to be inspected in the future we can come back to it later.

nomos-services/blend/src/epoch.rs
davidrusu: discussed in a call, we can do a transformation where the Tx is moved to an associated type in the `CryptarchiaServiceApi`

ntn-x2: I am honestly not a big fan of this p2p dependencies between services. Now we implement a trait in Blend for something that is defined in chain service. In my opinion, if we are to re-use implementations, which we should aim for, it should come down from the binary. Anyway, that is not the status quo now, so I provided the implementation as you suggested.

nomos-services/blend/src/epoch.rs
ntn-x2: We found some issues with @danielSanchezQ in implementing his proposal, yet I think we can go ahead and merge the simpler, non-stream implementation. Next week I will work on pulling all the stuff together, and if I find it to be more complex without a stream interface, I reserve the possibility to revert to the stream implementation. For now, as you suggested, let's go with the simpler to follow approach, that expects a tick as input, performs some validations to make sure ticks are strictly increasing, and yields `Some` if it receives a new epoch unseen before and if the state for such epoch can be found. Thanks for your input.

davidrusu (approved):
Nice, dropping the Tx type param from CryptarchiaServiceApi is a great to see as well

PR #1777, opened 2025-10-01
feat(kms): Add multisign support.

strinnityk:
Add multisign support to KMS backend trait and services.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/nomos-utils/src/types.rs
+++ b/nomos-utils/src/types.rs
@@ -1,8 +1,42 @@
 pub mod enumerated {
+    /// Checks if all references in a slice point to the same variant of an
+    /// enum.
+    ///
+    /// # Parameters
+    ///
+    /// - `values`: A slice of references to values of type `T`.
+    ///
+    /// # Returns
+    ///
+    /// - `true` if the slice is non-empty and all references in the slice point
+    ///   to the same variant, otherwise returns `false`.
+    ///
+    /// - An empty slice is considered to have all elements the same, thus
+    ///   returns `true`. Following `all`'s behaviour.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use nomos_utils::types::enumerated::is_same_variant;
+    ///
+    /// enum Color {
+    ///     Red,
+    ///     Blue,
+    /// }
+    ///
+    /// let values = vec![&Color::Red, &Color::Red];
+    /// assert!(is_same_variant(&values));
+    ///
+    /// let empty_values: Vec<&Color> = vec![];
+    /// assert!(is_same_variant(&empty_values));
+    ///
+    /// let mixed_values = vec![&Color::Red, &Color::Blue];
+    /// assert!(!is_same_variant(&mixed_values));
+    /// ```
     pub fn is_same_variant<T>(values: &[&T]) -> bool {
         let mut discriminants = values.iter().map(|item| std::mem::discriminant::<T>(item));
         let Some(head) = discriminants.next() else {
-            return false;
+            return true;
         };
         discriminants.all(|tail_item| tail_item == head)
     }
@@ -23,6 +57,9 @@ pub mod enumerated {
             assert!(!enumerated::is_same_variant(&[
                 &values[0], &values[1], &values[2]
             ]));
+
+            let values: [&TestEnum; 0] = [];
+            assert!(enumerated::is_same_variant(&values));
 
             let values = [TestEnum::A, TestEnum::A, TestEnum::A];
             assert!(enumerated::is_same_variant(&[


nomos-services/key-management-system/src/backend/mod.rs
```diff
@@ -26,6 +26,12 @@ pub trait KMSBackend {
         payload: <Self::Key as SecuredKey>::Payload,
     ) -> Result<<Self::Key as SecuredKey>::Signature, Self::Error>;
 
+    fn sign_multiple(
+        &self,
+        key_ids: Vec<Self::KeyId>,
```
ntn-x2, line 31: I would say we either make this a `HashSet`, or we make it an iterator, in my opinion. Since this is a trait, perhaps accepting an iterator is more generic. But if we want to convey that the list should be made of unique elements, then it could be a `HashSet` instead.

nomos-services/key-management-system/src/lib.rs
```diff
@@ -48,6 +48,12 @@ type KeyDescriptor<Backend> = (
     <<Backend as KMSBackend>::Key as SecuredKey>::PublicKey,
 );
 
+#[derive(Debug)]
+pub enum KMSSigningStrategy<KeyId> {
+    Single(KeyId),
+    Multi(Vec<KeyId>),
```
ntn-x2, line 54: Unless we assume it's ok to sign twice with the same key, this should not be a `Vec` in my opinion.

nomos-services/key-management-system-macros/src/lib.rs
thomaslavaur, line 1: I don't understand any of these functions. I guess it's because it's macro? (So I cannot really review it :/ )

nomos-services/key-management-system/src/backend/mod.rs
thomaslavaur: The list isn't necessary made of unique elements... you can have several notes using the same public key and you would sign with the same secret key then (for now we have duplication)

nomos-services/key-management-system/src/lib.rs
thomaslavaur: In the actual spec it's assumed you can sign with the same key. Because it's zk the order is important. a signature with sk1, sk2, sk1 would not be valid if it was signed with sk1, sk1, sk2

danielSanchezQ: Thanks @thomaslavaur thats actually valuable infor here. I was more in the path of @ntn-x2 as well.

nomos-services/key-management-system-macros/src/lib.rs
danielSanchezQ: Yeah, its rust that generates rust code :smile:

nomos-services/key-management-system/src/lib.rs
thomaslavaur: In the Mantle spec we assumed the list to be matching exactly the order of the input note of the ledger transaction. Whatever the repetition or anything else

nomos-services/key-management-system/src/backend/preload.rs
```diff
@@ -127,6 +131,43 @@ mod keys {
...
+            };
+
+            let head_discriminant = std::mem::discriminant::<Self>(head);
+            let mut tail_discriminants =
+                tail.iter().map(|item| std::mem::discriminant::<Self>(item));
+
+            let all_same = tail_discriminants.all(|discriminant| discriminant == head_discriminant);
+            if !all_same {
+                return Err(PreloadKeyError::MultiKeyRequiresSameVariant);
+            }
```
danielSanchezQ, line 145: This could be a generic method so we can reuse it in other implementations.

nomos-services/key-management-system-macros/src/lib.rs
```diff
@@ -143,39 +146,90 @@ fn build_key_enum_impl_secured_key(
...
+    let verify_same_variant = quote! {
+        let head_discriminant = std::mem::discriminant::<Self>(head);
+        let mut tail_discriminants = tail.iter().map(|item| std::mem::discriminant::<Self>(item));
+
+        let all_same = tail_discriminants.all(|discriminant| discriminant == head_discriminant);
+        if !all_same {
+            let error_message = String::from("Multi-key signature requires all keys to have the same variant.");
+            return Err(Self::Error::UnsupportedKey(error_message));
+        }
+    };
```
danielSanchezQ, line 190: Yeah, this should call the helper method.

ntn-x2 (approved):
Looks good to me, but I don't understand the need for the additional `types` feature to use a helper function.

danielSanchezQ (approved):

PR #1778, opened 2025-10-01
feat(blend): pre-compute winning PoL info in chain-leader and share it with Blend

ntn-x2:

Builds on top of the [shell PR](https://github.com/logos-co/nomos/pull/1739) by running a verifier on each new epoch, that checks for winning utxos for all slots in the epoch and notifies consumers of the broadcast channel. Since the process is quite fast, we don't need complicated logic to spawn tasks, but we can change it in the future if we need to, although for now it does not look like something we have to do.

Because there are still a lot of missing pieces in the actual PoL logic, I tried to factor out all the common pieces, so that when replacing the mock impl with the actual implementation, we don't risk inconsistencies between PoL and PoQ logic.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/consensus/cryptarchia-engine/src/time.rs
+++ b/consensus/cryptarchia-engine/src/time.rs
@@ -156,9 +156,8 @@ impl EpochConfig {
     }
 
     #[must_use]
-    pub fn starting_slot(&self, _epoch: &Epoch) -> Slot {
-        // TODO: get actual starting slot
-        0.into()
+    pub fn starting_slot(&self, epoch: &Epoch, base_period_length: NonZero<u64>) -> Slot {
+        Slot::from(u64::from(u32::from(*epoch)) * self.epoch_length(base_period_length))
     }
 }
 
--- a/consensus/cryptarchia-engine/src/time.rs
+++ b/consensus/cryptarchia-engine/src/time.rs
@@ -154,6 +154,12 @@ impl EpochConfig {
             .try_into()
             .expect("Epoch should build from a correct configuration")
     }
+
+    #[must_use]
+    pub fn starting_slot(&self, _epoch: &Epoch) -> Slot {
+        // TODO: get actual starting slot
+        0.into()
+    }
 }
 
 #[cfg_attr(feature = "serde", cfg_eval::cfg_eval, serde_with::serde_as)]
--- a/consensus/cryptarchia-engine/src/time.rs
+++ b/consensus/cryptarchia-engine/src/time.rs
@@ -16,6 +16,13 @@ use tokio::time::{Interval, MissedTickBehavior};
 #[derive(Clone, Debug, Eq, PartialEq, Copy, Hash, PartialOrd, Ord)]
 pub struct Slot(u64);
 
+impl Epoch {
+    #[must_use]
+    pub const fn new(inner: u32) -> Self {
+        Self(inner)
+    }
+}
+
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[derive(Clone, Debug, Eq, PartialEq, Copy, Hash, PartialOrd, Ord)]
 pub struct Epoch(u32);
--- a/consensus/cryptarchia-engine/src/time.rs
+++ b/consensus/cryptarchia-engine/src/time.rs
@@ -16,13 +16,6 @@ use tokio::time::{Interval, MissedTickBehavior};
 #[derive(Clone, Debug, Eq, PartialEq, Copy, Hash, PartialOrd, Ord)]
 pub struct Slot(u64);
 
-impl Epoch {
-    #[must_use]
-    pub const fn new(inner: u32) -> Self {
-        Self(inner)
-    }
-}
-
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[derive(Clone, Debug, Eq, PartialEq, Copy, Hash, PartialOrd, Ord)]
 pub struct Epoch(u32);
--- a/nomos-core/chain-defs/src/mantle/ledger.rs
+++ b/nomos-core/chain-defs/src/mantle/ledger.rs
@@ -57,6 +57,11 @@ impl Note {
     pub fn as_fr_components(&self) -> [Fr; 2] {
         [BigUint::from(self.value).into(), *self.pk.as_fr()]
     }
+
+    #[must_use]
+    pub fn id(&self) -> Fr {
+        BigUint::from(1u8).into() // placeholder for note ID, replace after mantle notes format update
+    }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
--- a/nomos-core/chain-defs/src/mantle/ledger.rs
+++ b/nomos-core/chain-defs/src/mantle/ledger.rs
@@ -57,11 +57,6 @@ impl Note {
     pub fn as_fr_components(&self) -> [Fr; 2] {
         [BigUint::from(self.value).into(), *self.pk.as_fr()]
     }
-
-    #[must_use]
-    pub fn id(&self) -> Fr {
-        BigUint::from(1u8).into() // placeholder for note ID, replace after mantle notes format update
-    }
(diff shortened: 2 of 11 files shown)


nomos-core/chain-defs/src/mantle/ledger.rs
```diff
@@ -57,6 +57,11 @@ impl Note {
     pub fn as_fr_components(&self) -> [Fr; 2] {
         [BigUint::from(self.value).into(), *self.pk.as_fr()]
     }
+
+    #[must_use]
+    pub fn id(&self) -> Fr {
+        BigUint::from(1u8).into() // placeholder for note ID, replace after mantle notes format update
+    }
```
davidrusu: NoteID is defined only for Utxo, not for Note (confusing, I know). NoteID needs output number and txhash which is only available in Utxo

consensus/cryptarchia-engine/src/time.rs
```diff
@@ -142,6 +153,12 @@ impl EpochConfig {
             .try_into()
             .expect("Epoch should build from a correct configuration")
     }
+
+    #[must_use]
+    pub fn starting_slot(&self, _epoch: &Epoch) -> Slot {
+        // TODO: get actual starting slot
+        0.into()
+    }
```
davidrusu: The implementation is:
```suggestion
    pub fn starting_slot(&self, epoch: &Epoch, base_period_length: NonZero<u64>) -> Slot {
        Slot::from((u32::from(epoch) as u64) * self.epoch_length(base_period_length))
    }
```

nomos-services/chain/chain-leader/src/leadership.rs
```diff
@@ -40,37 +35,29 @@ impl Leader {
...
 
-            let note_id: Fr = BigUint::from(1u8).into(); // placeholder for note ID, replace after mantle notes format update
-            let public_inputs = LeaderPublic::new(
-                aged_tree.root(),
-                latest_tree.root(),
-                epoch_state.nonce,
-                slot.into(),
-                epoch_state.total_stake(),
-            );
+            let note_id = utxo.note.id();
```
davidrusu: Note Id comes from the Utxo

```suggestion
            let note_id = utxo.id();
```

nomos-services/chain/chain-leader/src/leadership.rs
```diff
@@ -124,4 +101,139 @@ impl Leader {
...
+}
+
+fn public_inputs_for_utxo(
+    utxo: &Utxo,
+    epoch_state: &EpochState,
+    slot: Slot,
+    latest_tree: &UtxoTree,
+) -> Option<LeaderPublic> {
+    let () = epoch_state.utxos.witness(&utxo.id())?;
+    let () = latest_tree.witness(&utxo.id())?;
```
davidrusu: These would be used in the `private_inputs_..` above, they would fill in the `aged_path` and `latest_path` placeholder values.

nomos-services/chain/chain-leader/src/leadership.rs
```diff
@@ -124,4 +101,139 @@ impl Leader {
...
+
+                if let Err(err) =
+                    self.sender
+                        .send((leader_private, secret_key.clone(), epoch_state.epoch))
+                {
+                    tracing::error!(
+                        "Failed to send pre-calculated PoL winning slots to receivers. Error: {err:?}"
+                    );
+                }
+            }
```
davidrusu, line 234: update the`self.last_processed_epoch` and early return here after we find the first winning slot?

davidrusu (commented):
generally looks good, just a few details to fix

nomos-services/chain/chain-leader/src/leadership.rs
```diff
@@ -124,4 +101,139 @@ impl Leader {
...
     }
+
+    fn private_inputs_for_winning_utxo_and_slot(
+        &self,
+        utxo: &Utxo,
+        epoch: Epoch,
+        public_inputs: LeaderPublic,
+    ) -> LeaderPrivate {
+        // TODO: Get the actual witness paths and leader key
+        let aged_path = Vec::new(); // Placeholder for aged path
```
danielSanchezQ: This stuff should be retrieved from wallet right?

nomos-services/chain/chain-leader/src/leadership.rs
ntn-x2: In the chain leader subscribe API, I was thinking to return ALL winning slots for a new epoch, instead of making it Blend specific and only return the first one. Since this is not expensive, I thought it would not matter. What do you think?

nomos-services/chain/chain-leader/src/leadership.rs
danielSanchezQ: Ah, its already here. So then better to fill. Thanks David.

nomos-services/chain/chain-leader/src/leadership.rs
davidrusu: sure, this works as well, just wanted to make sure it was intentional, we were speaking about only detecting the first win of the epoch earlier

danielSanchezQ (approved):
Nice catches from @davidrusu, I have nothing to add. Let's just fill the gaps if possible and go ahead.

nomos-services/chain/chain-leader/src/leadership.rs
ntn-x2: Yeah I mean this PR was intended to mirror the current behaviour to pre-compute stuff for Blend. I did not want to change too much the current behaviour. I am addressing some of @davidrusu's comments, but some require bigger changes that should be tackled in a different PR. Also, I was not sure David or Giacomo were already working on them.

nomos-services/chain/chain-leader/src/leadership.rs
ntn-x2: Yes but `witness` returns a `()` while the path is expected to be a vector. I did not want to change too much logic in this PR, so I'd feel more comfortable to tackle this change specifically in a separate one, what do you guys think?

nomos-core/chain-defs/src/mantle/ledger.rs
ntn-x2: Good suggestion: https://github.com/logos-co/nomos/pull/1778/commits/619357a95913cade38e67325e859ea04c4ae2f0e.

nomos-services/chain/chain-leader/src/leadership.rs
ntn-x2: https://github.com/logos-co/nomos/pull/1778/commits/619357a95913cade38e67325e859ea04c4ae2f0e

consensus/cryptarchia-engine/src/time.rs
ntn-x2: Good, I added your suggestion here: https://github.com/logos-co/nomos/pull/1778/commits/fe3bf7a9f4ebe362eb587e2df378ea6dfb05ffea, although I think it would have been better to tackle these changes in a separate PR.

nomos-services/chain/chain-leader/src/leadership.rs
ntn-x2: Yes I think we can keep it more generic so any other service can also consume this info if they need to.

nomos-services/chain/chain-leader/src/leadership.rs
davidrusu: well, lets just move these witness calls to the private inputs function, that's where they will be used in the future. (doesn't make sense to have these calls in the public inputs function)

nomos-services/chain/chain-leader/src/leadership.rs
davidrusu: aged_path and latest_path actually comes from the ledger (It's the path in the UtxoTree).

data flow is
1. fetch eligible notes from wallet
2. derive the {latest|aged|_paths from latest ledger utxo tree and epoch_state's utxo tree
3. prove PoL with that data

we'll fill this in soon

Issue #1517, opened 2025-08-05
Devise a way to close a block stream without reading it to the end


youngjoon-lee:
When we run IBD or the sync protocol (e.g. handling orphan blocks), we send a download request to a peer, and we get a block stream as a result. We read blocks from the stream and try to apply them to the Cryptarchia.

If the `Cryptarchia::process_block` fails, it's better to close the stream even without reading it to the end, so that the responder can stop sending blocks. 

The current download logic reads blocks from `Pin<Box<dyn Stream<Item = ...>>>`, which doesn't provide the `close` method. 
Instead, the `libp2p::Stream` (which we use underneath) provides the `close` method. 
We may be able to wrap the `libp2p::Stream` as a concrete Stream type, and implement the `Drop` trait that closes the `libp2p::Stream`.

PR #1780, opened 2025-10-02
fix(chainsync): close libp2p stream when dropped

youngjoon-lee:

Closes https://github.com/logos-co/nomos/issues/1517

As mentioned in the issue #1517, we should close the libp2p stream even when a received block is invalid, because we don't need to receive subsequent blocks. If the stream is not closed, the peer sends blocks until the corresponding connection is closed. The connection is closed only when all the streams are closed.

In the previous implementation, we had closed the stream only when the network error happens. And, it was not possible for the downloader to close the stream when it wants, because the stream is consumed by `stream::try_unfold`. 

So, this PR introduces a struct that wraps `libp2p::Stream` and implements `Drop` for it. By doing so, we don't need to manually call `close_stream` in both sides (provider and downloader). The stream is closed automatically when the wrapper struct is dropped.

We have one limitation. `libp2p::Stream::close` is an async function. That is, it cannot be called in `Drop::drop` which is not async. So, I introduces a new approach but I think it works well. Please see details in the code comments.

yes

@youngjoon-lee @andrussal 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.


consensus/cryptarchia-sync/src/libp2p/stream.rs
```diff
@@ -0,0 +1,131 @@
...
+            .drop_sender
+            .take()
+            .expect("drop_sender must exist")
+            .send(stream)
+            .is_err()
+        {
+            error!("Failed to send a dropped stream to the stream closer");
+        }
+    }
+}
```
danielSanchezQ, line 83: Is this really necessary? Cant closing be resolved in the same stream implementation. Like one extra item that closes the stream and return none instead?

youngjoon-lee: I wanted to avoid introducing `drop_sender`, but couldn't find any better way. What we want is closing the stream even before reading it to the end, if a block that was yielded from the stream is invalid.
So, I ended up introducing `drop_sender` that is useful for both reader and sender sides. This approach also replaces the previous [`incoming_streams_to_close`](https://github.com/logos-co/nomos/pull/1780/files#diff-f91274adb81954c465f4c039d41066421ff8c6fb14da6bef49a7567a7ce8af3aL151). 
Hm, let me think again. I may be able to take an approach similar as what you suggested

PR #1781, opened 2025-10-02
feat(da-network): save peer id -> provider id mappings to da storage

pradovic:

This PR is adding peer id -> provider id mappings to da network storage. This is needed to transfer back from peer ids to provider ids when creating sampling opinion vectors.

Yes

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.


bacv (approved):
Looks good!

nomos-services/data-availability/network/src/membership/adapters/service.rs
```diff
@@ -73,30 +75,37 @@ where
...
-                                provider_id.0,
-                                err
-                            );
-                            None
-                        }
+            let mut peers_map: HashMap<PeerId, Multiaddr> = HashMap::new();
+            let mut provider_mappings: HashMap<PeerId, ProviderId> = HashMap::new();
+
+            for (provider_id, locators) in providers_map {
+                // TODO: Support multiple multiaddrs in the membership.
```
danielSanchezQ, line 83: When is this gonna happen? In this pr or later? If it is latter, lets open an issue.

nomos-services/data-availability/network/src/lib.rs
```diff
@@ -384,17 +384,23 @@ where
...
-                        session_id, providers
+                        update.session_id, update.peers
                     );
-                    match Self::handle_membership_update(session_id, providers, &membership_storage).await {
+                    match Self::handle_membership_update(
+                        update.session_id,
+                        update.peers,
+                        &membership_storage,
+                        update.provider_mappings,
+
```
danielSanchezQ: what is this blank line? 😵

nomos-services/storage/src/api/backend/rocksdb/da.rs
```diff
@@ -199,6 +200,47 @@ impl StorageDaApi for RocksBackend {
...
+    async fn store_providerid_mappings(
+        &mut self,
+        mappings: HashMap<Self::Id, ProviderId>,
+    ) -> Result<(), Self::Error> {
+        let mut key_provider_map = HashMap::new();
+
+        for (peer_id, provider_id) in mappings {
+            let provider_key = key_bytes(DA_PROVIDER_MAPPINGS_PREFIX, peer_id.to_bytes());
+            let serialized_provider_id =
+                <()>::serialize(&provider_id).expect("Serialization of ProviderId should not fail");
```
danielSanchezQ: how can it fail if you are serializing an empty tuple? Does this work through tests?
```suggestion
                <ProviderId>::serialize(&provider_id).expect("Serialization of ProviderId should not fail");
```

nomos-services/storage/src/api/backend/rocksdb/da.rs
```diff
@@ -199,6 +200,47 @@ impl StorageDaApi for RocksBackend {
...
+            key_provider_map.insert(provider_key, serialized_provider_id);
+        }
+
+        self.bulk_store(key_provider_map).await.map_err(|e| {
```
danielSanchezQ, line 217: Probably we should make a version of bulkstore that doesnt own a type per se. But change it to take an `IntoIterator` instead. So we do not need to feed an inmemory hashmap for this kind of stuff.
cc @andrussal

nomos-services/data-availability/network/src/membership/adapters/service.rs
pradovic: This TODO is here since the first time providers mappings was implemented in stream implementation by @bacv. I remember at that point it was agreed to use only one locator per provider. We can open an issue maybe and see how to support it soon?

nomos-services/storage/src/api/backend/rocksdb/da.rs
pradovic: It is this style in other places as well in the file (for assignations), but it works becuase: nomos-core/chain-defs/src/codec/mod.rs - The trait methods are associated functions, not methods on self. They take a generic T, not Self.

I will update it to show the explicit type for clarity

pradovic: https://github.com/logos-co/nomos/pull/1781/commits/47029d32925bb583ccb40c49f44df6edb94a14dc

nomos-services/storage/src/api/backend/rocksdb/da.rs
pradovic: Maybe something like this: https://github.com/logos-co/nomos/pull/1781/commits/908f3f359c81e3e52c4e2a07cf101637863d2cc8/.

nomos-services/data-availability/network/src/membership/adapters/service.rs
pradovic: https://github.com/logos-co/nomos/issues/1790
cc. @bacv just fyi as well.

nomos-services/storage/src/api/backend/rocksdb/da.rs
danielSanchezQ: But why was it implemented for `()`, that makes no sense. In what other places is it like that?

pradovic: It was in a same file for assignations as well, and it is like that in membership as well. I will update those in a separate PR. It works, it is just bad syntax. It works because of blanked impl in `nomos-core/chain-defs/src/codec/mod.rs `:

```
impl<S> SerdeOp for S
where
    S: Serialize + DeserializeOwned,
{
    type Error = Error;

    fn serialize<T: Serialize>(value: &T) -> Result<Bytes> {
        serialize(value) // This will take any value and produce bytes
    }

    fn deserialize<T: DeserializeOwned>(data: &[u8]) -> Result<T> {
        deserialize(data)
    }

    fn serialized_size<T: Serialize>(value: &T) -> Result<u64> {
        serialized_size(value)
    }
}
```

pradovic: It takes generic param T, and not self, which is why it works.

danielSanchezQ: > It was in a same file for assignations as well, and it is like that in membership as well. I will update those in a separate PR. It works, it is just bad syntax. It works because of blanked impl:
> 
> ```
> impl<S> SerdeOp for S
> where
>     S: Serialize + DeserializeOwned,
> {
>     type Error = Error;
> 
>     fn serialize<T: Serialize>(value: &T) -> Result<Bytes> {
>         serialize(value) // This will take any valuie and produce bytes
>     }
> 
>     fn deserialize<T: DeserializeOwned>(data: &[u8]) -> Result<T> {
>         deserialize(data)
>     }
> 
>     fn serialized_size<T: Serialize>(value: &T) -> Result<u64> {
>         serialized_size(value)
>     }
> }
> ```

Well, I see it now. This is wrong. Methods should not be generic. It will just fucked up the implementation. cc @andrussal. It should be `S` everywhere, not `T`. It makes the code weird and difficult to follow.

pradovic: Maybe something like:
```
pub trait SerdeOp: Serialize + DeserializeOwned {
    type Error: StdError;
    
    /// Serialize self to bytes
    fn serialize(&self) -> Result<Bytes>;
    
    /// Deserialize bytes to Self
    fn deserialize(data: &[u8]) -> Result<Self>;
    
    /// Get the serialized size of self
    fn serialized_size(&self) -> Result<u64>;
}

impl<S> SerdeOp for S
where
    S: Serialize + DeserializeOwned,
{
    type Error = Error;
    
    fn serialize(&self) -> Result<Bytes> {
        bincode::serialize(self)
    }
    
    fn deserialize(data: &[u8]) -> Result<Self> {
        bincode::deserialize(data)
    }
    
    fn serialized_size(&self) -> Result<u64> {
        bincode::serialized_size(self)
    }
}
```
And then usage becomes:
```
let serialized_provider_id = provider_id.serialize()
    .expect("Serialization of ProviderId should not fail");
    
```

danielSanchezQ: yes, but to be done in a different pr. The changes here should be enough.

pradovic: I can do it today, if @andrussal is busy with other stuff, should be simple.

andrussal: > I can do it today, if @andrussal is busy with other stuff, should be simple.

Oh, right, this is odd now when I step back and look it, please go ahead!

pradovic: Sounds good, I'm on it

pradovic:
@danielSanchezQ Can this one be merged now? I think all suggestions are applied and the storage stuff is going as a separate PR?

PR #1782, opened 2025-10-02
feat(da-network): opinion aggregator output sorted bitvectors of provider ids

pradovic:

This PR is expanding sampling opinion aggregator to provide sorted bitvecs of provider ids. It also adds a test for opinion aggregator.
It is also expanding the network back end to provide local provider id along peer id.

Yes

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/nomos-services/data-availability/network/src/lib.rs
+++ b/nomos-services/data-availability/network/src/lib.rs
@@ -10,6 +10,7 @@ use std::{
     fmt::{self, Debug, Display},
     marker::PhantomData,
     pin::Pin,
+    sync::Arc,
     time::Duration,
 };
 
@@ -345,13 +346,18 @@ where
             .relay::<StorageAdapter::StorageService>()
             .await?;
 
-        let storage_adapter = StorageAdapter::new(storage_service_relay);
-        let membership_storage =
-            MembershipStorage::new(storage_adapter, membership.clone(), addressbook.clone());
+        let storage_adapter = Arc::new(StorageAdapter::new(storage_service_relay));
+        let membership_storage = MembershipStorage::new(
+            Arc::clone(&storage_adapter),
+            membership.clone(),
+            addressbook.clone(),
+        );
 
         let membership_service_adapter = MembershipServiceAdapter::new(membership_service_relay);
 
-        let mut opinion_aggregator = OpinionAggregator::<Membership>::new(backend.local_peer_id());
+        let (local_peer_id, local_provider_id) = backend.local_peer_id();
+        let mut opinion_aggregator =
+            OpinionAggregator::new(storage_adapter, local_peer_id, local_provider_id);
 
         wait_until_services_are_ready!(
             &self.service_resources_handle.overwatch_handle,
@@ -402,11 +408,7 @@ where
 
                     ).await {
                         Ok(current_membership) => {
-                            let opinions = opinion_aggregator.handle_session_change(current_membership);
-                            if let Some(opinions) = opinions {
-                                tracing::debug!("Processing opinions {opinions:?}");
-                                // todo: sdp_adapter.process_opinions(opinions).await;
-                            }
+                            Self::handle_opinion_session_change(&mut opinion_aggregator, current_membership).await;
                             let _ = subnet_refresh_sender.send(()).await;
                         }
                         Err(e) => {
@@ -481,7 +483,7 @@ where
     async fn handle_network_service_message(
         msg: DaNetworkMsg<Backend, DaSharesCommitments, RuntimeServiceId>,
         backend: &mut Backend,
-        membership_storage: &MembershipStorage<StorageAdapter, Membership, DaAddressbook>,
+        membership_storage: &MembershipStorage<Arc<StorageAdapter>, Membership, DaAddressbook>,
         api_adapter: &ApiAdapter,
         addressbook: &DaAddressbook,
     ) {
@@ -568,7 +570,7 @@ where
     async fn handle_membership_update(
         session_id: SessionNumber,
         update: HashMap<Membership::Id, Multiaddr>,
-        storage: &MembershipStorage<StorageAdapter, Membership, DaAddressbook>,
+        storage: &MembershipStorage<Arc<StorageAdapter>, Membership, DaAddressbook>,
         provider_mappings: HashMap<Membership::Id, ProviderId>,
     ) -> Result<Membership, DynError> {
         let current_membership = storage
@@ -579,7 +581,7 @@ where
 
     async fn handle_historic_sample_request(
         backend: &Backend,
-        membership_storage: &MembershipStorage<StorageAdapter, Membership, AddressBook>,
+        membership_storage: &MembershipStorage<Arc<StorageAdapter>, Membership, AddressBook>,
         session_id: SessionNumber,
         block_id: HeaderId,
         blob_ids: HashSet<[u8; 32]>,
@@ -598,6 +600,32 @@ where
             send.await;
         } else {
             tracing::error!("No membership found for session {session_id}");
+        }
+    }
+
+    async fn handle_opinion_session_change(
+        opinion_aggregator: &mut OpinionAggregator<Membership, Arc<StorageAdapter>>,
+        current_membership: Membership,
+    ) {
+        match opinion_aggregator
+            .handle_session_change(current_membership)
+            .await
+        {
+            Ok(Some(opinions)) => {
+                tracing::debug!(
+                    "Processing opinions - session_id: {}, new_opionions: {}, old_opinions: {}",
+                    opinions.session_id,
+                    opinions.new_opinions,
+                    opinions.old_opinions
+                );
+                // todo: sdp_adapter.process_opinions(opinions).await;
+            }
+            Ok(None) => {
+                tracing::debug!("No opinions generated yet");
+            }
+            Err(e) => {
+                tracing::error!("Failed to generate opinions: {e}");
+            }
         }
     }
 }
--- a/nomos-services/data-availability/network/src/opinion_aggregator.rs
+++ b/nomos-services/data-availability/network/src/opinion_aggregator.rs
@@ -3,16 +3,23 @@ use std::collections::{BTreeMap, HashMap, HashSet};
 use bitvec::prelude::*;
 use libp2p::PeerId;
 use nomos_core::{block::SessionNumber, sdp::ProviderId};
-use nomos_da_network_core::protocols::sampling::opinions::{Opinion, OpinionEvent};
+use nomos_da_network_core::{
+    SubnetworkId,
+    protocols::sampling::opinions::{Opinion, OpinionEvent},
+};
+use overwatch::DynError;
 use subnetworks_assignations::MembershipHandler;
+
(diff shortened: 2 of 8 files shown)


nomos-services/data-availability/network/src/lib.rs
```diff
@@ -600,6 +602,32 @@ where
...
+                tracing::debug!(
+                    "Processing opinions - session_id: {}, new_opinions: {}, old_opinions: {}",
+                    opinions.session_id,
+                    opinions.new_opinions,
+                    opinions.old_opinions
+                );
+                // todo: sdp_adapter.process_opinions(opinions).await;
+            }
+            Ok(None) => {
+                tracing::debug!("No opinions generated yet");
```
bacv: Shouldn't (when sdp.process_opinions is implemented) this send a bitvec of opinions with all zeroes? As a DA service provider we want to inform others that we're active even we didn't performed any sampling for some reason.

nomos-services/data-availability/network/src/opinion_aggregator.rs
```diff
@@ -133,15 +139,20 @@ where
...
-        let opinions = if self.current_membership.is_some() && self.previous_membership.is_some() {
-            self.generate_opinions()
-        } else {
-            None
-        };
-
+    pub async fn handle_session_change(
+        &mut self,
+        new_membership: Membership,
+    ) -> Result<Option<Opinions>, DynError> {
```
bacv: As this will always be called during the session change, I think the return type could be an `enum OpinionsResult`, which would contain all variants: `NoOpinions` `Opinions` `Error(DynError)`,

nomos-services/data-availability/network/src/lib.rs
pradovic: Hmm, now re-reading the spec I think you are right. It is not explicitly stated but based on self inclusion it should be all zeros bitvec. I will update! Tnx

nomos-services/data-availability/network/src/opinion_aggregator.rs
```diff
@@ -3,16 +3,29 @@ use std::collections::{BTreeMap, HashMap, HashSet};
...
+
 const OPINION_THRESHOLD: u32 = 10;
 
-pub struct OpinionAggregator<Membership>
-where
-    Membership: MembershipHandler<Id = PeerId>,
-{
+pub enum OpinionResult {
+    Opinions(Opinions),
+    InsufficientData, // Not enough sessions to form opinions yet
```
bacv: Nice, now it's clear that we shouldn't send Activity message if not enough sessions. By the way, could you point out to the spec where this behaviour is defined?

pradovic: Again, it is not explicit. It just starts with: During the session S + 1. During the session S we gather opinions. Maybe we could ask @madxor this?

bacv: Ah, I think I get it now, is it when the first session switches, and we have no opinions for "old session" because we just started tracking opinions?

pradovic: Yes. We cannot form bitvec for S - 1 if it does not exists, imo. But I might be wrong

bacv: Okay this makes sense, in future we might want to add more explicit check that involves session numbers as this case should never happen after the first session.

bacv (approved):

nomos-services/data-availability/network/src/opinion_aggregator.rs
madxor: I have update the spec a bit to make it clearer what should happen. To sum up, we always send a activity message, but we might not have opinions for previous session as that session might have not exists or the DA service was not operational. https://www.notion.so/nomos-tech/NomosDA-Rewarding-203261aa09df80af8c77dfb3dc593673?source=copy_link#281261aa09df80dcb638f726c2bb5243

pradovic: I will update the code to handle the first activity proof with zero old opinions

pradovic: Actually code already matches the updated spec. Thanks!

pradovic: @madxor we agreed to return an empty bit vec in session 0 for previous session opinions, as we don't know the ordering of the S-1 (old opinions), in this case. This is the code change that reflects that: https://github.com/logos-co/nomos/pull/1782/commits/5feee91641b38464bcdddad18a215786f15636b8.

pradovic: @madxor I was re-reading the spec, and just two nitpicks:

I found a typo in the spec: https://www.notion.so/nomos-tech/NomosDA-Rewarding-203261aa09df80af8c77dfb3dc593673?source=copy_link#281261aa09df80dcb638f726c2bb5243
"pass_session" should be "previous_session"
Also, I noticed some inconsistency regarding when opinion vectors can be empty.

The spec first states:
previous_session_opinions contains opinions gathered from the sampling process done on old blocks. When there are no old blocks (the first session after genesis block or after a non-operational DA period due to insufficient number of declared nodes), the old opinions should neither be collected nor validated.

But then the validation rules define:
The current_sessions_opinions is zero (size of byte) and the DA service was not operational during that session.
The past_sessions_opinions is zero (size of byte) and the DA service was not operational during that session.

Two questions:
1. For `previous_session_opinions`: Based on the first definition, it seems this can only be empty in session 0 (genesis). Is that correct? If so, do we need the non-operational case for previous opinions, or is that redundant? @danielSanchezQ , @bacv Can we have a case, after session 0, that DA is not operational and we have too little members registered via SDP? I think that the consensus was that this should never happen 

2. For current_session_opinions: Even at genesis (session 0), we'll have a current membership, so current opinions should always exist. Do we actually need to define the zero-byte case for current_session_opinions? (this is also related to the previous question).

bacv: > Can we have a case, after session 0, that DA is not operational and we have too little members registered via SDP? I think that the consensus was that this should never happen

SDP can have to little participants for DA network to operate, but we allow provider to send activity message if he choses so. In worst case opinions vector will contain the sole provider. 

If DA network didn't have any nodes in previous session, and current session has, then it should behave as in the genesis case.

madxor: @pradovic thanks for typo, corrected.

Yes, we can have both the `previous` and `current` opinions empty. This was discussed here: https://discord.com/channels/1111286067413405788/1423639218193698899 

A summary:

The opinions cannot be gathered in when the DA service is not operational due to a) start of the chain (the case we discussed - no previous session) or b) not enough nodes declared for service for a session (case discussed with Gusto).

Re 1 and 2: `previous` opinions cannot be gathered for the genesis case AND the case where a service was not operational for a single or more sessions (due to not enough nodes case). Imagine this: ...`ono`... you have operational session (`o`) followed by non-operational (`n`) and then back to operational (`o`). For the session when the service is non-operational `o(n)o` the previous `opinions` can be gathered but not the `current`. However, in the next session `on(o)` a node can gather `current` but not the `previous.

> SDP can have to little participants for DA network to operate, but we allow provider to send activity message if he choses so. In worst case opinions vector will contain the sole provider.

@bacv This is, partially correct as in this case when the session is not operational a node sends an empty byte without indicating its own position in it.

pradovic: Understood, just wanted to double check. Thank you!

cc @bacv , @danielSanchezQ Maybe for todo:
We will need to handle this and provide a way to set an empty membership handler (currently it will panic , we don't have a way to provide empty membership, we could also skip the update if it is not needed on the membership handler level).

pradovic: If we are going to keep membership handler assert < minimum_size, then I will make opinion aggregator work on 
```
// Current session peers (None if DA disabled)
    current_session_peers: Option<HashMap<PeerId, Multiaddr>>,  // instead of MembershipHandler
 // Previous session peers
    previous_session_peers: Option<HashMap<PeerId, Multiaddr>>, // instead of MembershipHandler
 ```
 
 This way it is decoupled from membership handler, and can treat empty sessions per spec.
Another way would be to support this case on membership handler level, either by wrapper type around assignations, or directly in assignations (by returning empty members).

Comment (bacv) [reply nomos-services/data-availability/network/src/opinion_aggregator.rs]: > This way it is decoupled from membership handler, and can treat empty sessions per spec.

I think this is a way to go. This pr can be merged in my opinion.

Comment (pradovic) [reply nomos-services/data-availability/network/src/opinion_aggregator.rs]: > > This way it is decoupled from membership handler, and can treat empty sessions per spec.
> 
> I think this is a way to go. This pr can be merged in my opinion.

I would just update it as suggested, removing membership handler. This way the whole logic is here. It should not be a big commit.

File: nomos-services/data-availability/network/src/opinion_aggregator.rs [hunk 4]
```diff
@@ -124,117 +178,530 @@ where
...
-        new_opinions: &BTreeMap<ProviderId, bool>,
-        old_opinions: &BTreeMap<ProviderId, bool>,
-    ) -> (BitVec, BitVec) {
-        // Values are already in sorted order by ProviderId
-        let new_bits: BitVec = new_opinions.values().copied().collect();
-        let old_bits: BitVec = old_opinions.values().copied().collect();
-
-        (new_bits, old_bits)
+    #[tokio::test]
+    async fn test_regular_flow_after_session_0() {
```
danielSanchezQ, line 431: Nice testing thanks!

nomos-services/data-availability/network/src/opinion_aggregator.rs
```diff
@@ -124,117 +178,530 @@ where
...
+
+        self.reset_opinion_maps();
+
+        result
+    }
+
+    fn generate_opinions(&self) -> Result<Opinions, OpinionError> {
+        // Both memberships are guaranteed to be Some at this point
+        let current = self.current_membership.as_ref().unwrap();
+        let previous = self.previous_membership.as_ref().unwrap();
```
danielSanchezQ, line 257: Use `expect` instead.

nomos-services/data-availability/network/src/opinion_aggregator.rs
```diff
@@ -124,117 +178,530 @@ where
...
+                previous.members().into_iter(),
+                &self.prev_session_positive_opinions,
+                &self.prev_session_negative_opinions,
+                &previous.provider_mappings,
+                include_self_in_old,
+            ) {
+                Ok(bv) => bv,
+                Err(e) => return Err(OpinionError::Error(e)),
+            }
+        };
```
danielSanchezQ: This and the previous (similar) block, could be methods.

pradovic: Ah, this was already handled in peer_opinions_to_provider_bitvec, so I changed it to just call this function, so no even need for additional method.  Good catch, tnx!

PR #1783, opened 2025-10-02
feat(ledger): Membership in nomos ledger

bacv:

Membership tracking implementation in nomos-ledger, membership is reevaluated on every new block, the forming session is promoted to active depending on configured session duration.

Streaming session changes from chain service will be added in next PR.

Yes

@pradovic @bacv 

SDP service parameters were extended in this implementation to contain `session_duration`, specification doesn't have this field, but as parameters come from configuration, not the chain, it can be considered implementation detail.

No

## Checklist

* [x] 2. Description added.

### DIFF
--- a/ledger/nomos-ledger/src/mantle/mod.rs
+++ b/ledger/nomos-ledger/src/mantle/mod.rs
@@ -215,6 +215,8 @@ impl LedgerState {
         Ok((self, balance))
     }
 
+    // This method should be called once per block to ensure that state of previous
+    // declarations is reevaluated.
     pub fn try_update_membership(
         mut self,
         current_block_number: BlockNumber,
--- a/ledger/nomos-ledger/src/mantle/sdp.rs
+++ b/ledger/nomos-ledger/src/mantle/sdp.rs
@@ -220,17 +220,16 @@ impl SdpLedger {
             };
 
             let expected_active_session_num = block_number / params.session_duration;
-            if forming_session.session_number > expected_active_session_num {
-                continue;
+
+            if forming_session.session_number <= expected_active_session_num {
+                let new_active_session = forming_session.clone();
+
+                let mut next_forming_session = new_active_session.clone();
+                next_forming_session.session_number += 1;
+                new_active_sessions = new_active_sessions.insert(service_type, new_active_session);
+                new_forming_sessions =
+                    new_forming_sessions.insert(service_type, next_forming_session);
             }
-
-            let new_active_session = forming_session.clone();
-
-            let mut next_forming_session = new_active_session.clone();
-            next_forming_session.session_number += 1;
-
-            new_active_sessions = new_active_sessions.insert(service_type, new_active_session);
-            new_forming_sessions = new_forming_sessions.insert(service_type, next_forming_session);
         }
 
         self.active_sessions = new_active_sessions;


bacv:
> Also, we are usually only interested in the LIB side of this

The forming session is composed from the contents of new blocks and they might change between the forks, if we don't persist membership, then I would say it shouldn't be in the ledger.

> if we keep it like this it makes the historical sampling much easier

Maybe @pradovic could comment on this, for historical sampling we'd still need to stream LIB membership, and chain service wouldn't use the membership in ledger for historical sampling, chain service would just pass session id and blobs for sampling. This is only for rewards.

danielSanchezQ (approved):

bacv:
> For historical sampling we use saved snapshot of assignations for the desired session, right? 

Yes, idea in this PR is to keep streaming membership changes to subscribed services, which in DA case is DA network service, so subnetwork assignations stays as is, only the source which streams the membership data changes.

ledger/nomos-ledger/src/mantle/mod.rs
```diff
@@ -88,6 +96,10 @@ impl LedgerState {
         Ok(self)
     }
 
+    #[expect(
+        clippy::too_many_lines,
+        reason = "All operations handled in one method"
+    )]
```
zeegomo, line 96: we should probably do something about this

ledger/nomos-ledger/src/config.rs
```diff
@@ -1,14 +1,18 @@
...
 use nomos_core::sdp;
 
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[derive(Copy, Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct Config {
     pub epoch_config: cryptarchia_engine::EpochConfig,
     pub consensus_config: cryptarchia_engine::Config,
-    pub service_params: sdp::ServiceParameters,
+    pub service_params: Arc<HashMap<sdp::ServiceType, sdp::ServiceParameters>>,
```
zeegomo, line 15: why?

ledger/nomos-ledger/src/mantle/membership.rs
zeegomo, line 1: shouldn't this be part of the sdp module? (so that we can avoid e.g. passing declarations)

zeegomo:
> Meaning that we keep membership changes but its a never ending growing list of changes?

One thing that I noticed while reviewing other code is that we store declarations indefinitely. That will bloat up the ledger and is not something we want, we should drop a declaration if it's not relevant to that ledger state anymore (e.g. a declaration far past the inactivity params). 
A similar case in here is `Session`, does that grow indefinitely or is `remove` somewhat guaranteed to be called for each declaration that is not needed anymore?

ledger/nomos-ledger/src/config.rs
bacv: We are cloning the config for every new consensus state, before Config had a "Copy" derived, with the HashMap of service_params, we can't implement "Copy", so it's now cloned instead.

ledger/nomos-ledger/src/mantle/membership.rs
bacv: Yes, we can merge it to one module, would you prefer all functions to be as part of `SdpLedger` struct, or just move the membership.rs to the `sdp` module inside the nomos-ledger?

zeegomo: I'd say it should be part of the `SdpLedger` struct so that we avoid passing information like the set of declarations, but it's up to you on what you think makes most sense

zeegomo:
> If you are referring to the storage of sessions in the rpds,

where does this happen?


> then maybe we shouldn't keep memberhsip sessions in the ledger, this way the storage could be managed separately, only some recent sessions would be persisted, others pruned.

State in the ledger needs to be pruned anyway that's not an issue per se. We need to make sure that 1) only the necessary info is contained in each state (e.g. no inactive declarations are carried over forever) 2) we prune old ledger states

ledger/nomos-ledger/src/config.rs
zeegomo: ah I see, we can actually change that `Ledger::try_update` to accept `&mut self` and modify in place, that one is not optimised for multiple versions anyway

bacv: I'm not sure how this would work, should the `Ledger::try_update` take `&mut self` and instead of returning new state, modify existing in place and return `Result<(), LedgerError>`? Wouldn't that leave ledger in bad state on error? We'd still need to clone the state before modifying it?

ledger/nomos-ledger/src/mantle/membership.rs
bacv: Moved to sdp module https://github.com/logos-co/nomos/pull/1783/commits/520978367ea7c1ccef4d7e5cf2b879c92129b782

ledger/nomos-ledger/src/config.rs
zeegomo: the core logic happens inside `LedgerState` (using the pattern you described), the `Ledger` where this is cloned is just a collection of ledger states, we can only add or not add a new state there, so that's safe

ledger/nomos-ledger/src/mantle/sdp.rs
```diff
@@ -21,12 +25,24 @@ pub enum SdpLedgerError {
...
+    #[error("Service parameters are missing for {0:?}")]
+    ServiceParamsNotFound(ServiceType),
+    #[error("Can't update genesis state during different block number")]
+    NotGenesisBlock,
 }
 
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct SdpLedger {
     pub declarations: rpds::HashTrieMapSync<DeclarationId, DeclarationState>,
```
zeegomo, line 43: btw this is what is scaring me, we keep declarations forever

ledger/nomos-ledger/src/mantle/mod.rs
```diff
@@ -200,6 +214,17 @@ impl LedgerState {
 
         Ok((self, balance))
     }
+
+    pub fn try_update_membership(
```
zeegomo, line 220: should this be public? I think the only public interface should be `apply_tx`

ledger/nomos-ledger/src/config.rs
bacv: Got it, seems that this is potentially a bigger change, as all chain service will need to be updated to use `&cryptarchia` instead of `cryptarchia`, lets do it in another PR?

ledger/nomos-ledger/src/mantle/sdp.rs
```diff
@@ -110,4 +153,271 @@ impl SdpLedger {
...
+            let Some(params) = service_params.get(&service_type) else {
+                return Err(SdpLedgerError::SessionParamsNotFound(service_type));
+            };
+
+            let Some(forming_session) = new_forming_sessions.get(&service_type) else {
+                return Err(SdpLedgerError::FormingSessionNotFound(service_type));
+            };
+
+            let expected_active_session_num = block_number / params.session_duration;
+            if forming_session.session_number > expected_active_session_num {
```
zeegomo: can we reframe the opposite way without the continue and switching when we are at a boundary?

ledger/nomos-ledger/src/mantle/mod.rs
bacv: The membership should be updated once per block in order to reevaluate all declarations that were not present as a transaction. For example declaration was active in previous block, but in this one it's no longer considered active because of the inactivity period. For this reason the name has no `tx` in it, but maybe it could be named `try_apply_block`.

ledger/nomos-ledger/src/mantle/sdp.rs
bacv: Maybe I misunderstood the ledger design, but shouldn't the data that belongs to ledger state be in the ledger, and only if that state is removed, the related data removed just then? How is this solved with utxos?

ledger/nomos-ledger/src/mantle/mod.rs
zeegomo: I see, makes sense, but I think we should add this explanation to the code

ledger/nomos-ledger/src/mantle/sdp.rs
zeegomo: the point is that we never remove declarations that are clearly inactive here, while any utxo might be spent to it has to be kept available. Actually, better question, do we need to keep inactive / very old declarations around (e.g. is it important whether there was a declaration 2y ago that was not updated since)?

bacv: > e.g. is it important whether there was a declaration 2y ago that was not updated since

I think it's not important, but if we remove that info from the ledger, maybe it shouldn't be in the ledger?

bacv: Again, I think I misunderstood the ledger meaning in this context. I'll add a method to remove declaration (maybe a deeper cleaning will be required for `rpds`) and remove on withdrawal. Also remove inactive on garbage collection. Similar cleaning will be needed on stored ledger states.

ledger/nomos-ledger/src/config.rs
zeegomo: but then just remove arc and do a full clone instead

ledger/nomos-ledger/src/mantle/sdp.rs
zeegomo: > I think it's not important, but if we remove that info from the ledger, maybe it shouldn't be in the ledger?

The ledger is not the storage for all historical information of the chain. A ledger state is the (minimal) set of necessary information that describes the state of the system at a particular spot in time and that is needed (and sufficient) to compute future states. It's important to keep this minimal because 1) we have forking so we need to keep track of many states concurrently 2) we have this in memory 

Stored ledger states are e.g. pruned once a block becomes immutable, as we won't ever need that information again. As an analogy of why we don't care about old declarations, we don't keep track of spent utxos

PR #1786, opened 2025-10-06
feat(membership): Membership in chain service

bacv:

Broadcast session changes from the chain service. Broadcast service is used to avoid the `Cryptarchia<DaNetwork<Cryptarchia>>` type. Alternatively chain service could send session changes directly to the Blend and DA services, but that would require expanding the chain service types.

Yes

@bacv

N/A

N/A

## Checklist

* [x] 2. Description added.


danielSanchezQ (approved):
🚀 I'm down for the middle broadcast service.

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -1121,6 +1158,49 @@ where
...
+
+            let previous_session_numbers = match cryptarchia.active_sessions_numbers(&prev_lib) {
+                Ok(session_numbers) => session_numbers,
+                Err(e) => {
+                    warn!("Error getting previous session numbers: {e}");
+                    ServiceType::iter().map(|s| (s, 0)).collect()
+                }
+            };
+
+            if let Ok(sessions) = cryptarchia.active_sessions_numbers(&new_lib) {
```
ntn-x2:
```suggestion
            let Ok(sessions) = cryptarchia.active_sessions_numbers(&new_lib) else {
```

nomos-services/chain/broadcast-service/src/lib.rs
```diff
@@ -80,6 +104,16 @@ where
                         error!("Could not subscribe to new blocks channel: {err:?}");
                     }
                 }
+                BlockBroadcastMsg::SubscribeBlendSession { result_sender } => {
+                    if let Err(err) = result_sender.send(self.blend_session.subscribe()) {
```
ntn-x2, line 108: Are we supposed to get one value immediately after subscribing to this endpoint? Or is that handled elsewhere?

bacv: The retransmission is not implemented, I think we could hold the latest value and send it on subscription, would that be enough in your case?

ntn-x2: Yes, I think that is what we need in case of a node crash/restart.

nomos-services/chain/chain-service/src/lib.rs
bacv: Updated with your suggestion to return an error when active session numbers cant be recieved (should never happen) https://github.com/logos-co/nomos/pull/1786/commits/0c9b3fc1e9d75296e3a12ca5341e335439d5cad3

ntn-x2: My suggestion was simply about replacing the very big if block with an `if-let-else` that would remove one layer of indentation, else `continue`ing. Sorry I should have made this clear. If the error should never happen, then there is no need to throw one. Either we panic or log a `error` message and move on. Again, sorry that I did not provide more context about my suggestion.

bacv: No worries, now the error is propagated, the indentation was reduced, do you think other changes are needed?

nomos-services/chain/broadcast-service/src/lib.rs
bacv: Updated to use watch channels that returns the latest value on subscription https://github.com/logos-co/nomos/pull/1786/commits/47824520fd545486308476ee3e0575a5f8379df8

ntn-x2 (approved):

nomos-services/chain/broadcast-service/src/lib.rs
```diff
@@ -24,14 +24,24 @@ pub struct BlockInfo {
...
+    BroadcastBlendSession(SessionUpdate),
+    BroadcastDASession(SessionUpdate),
     SubscribeToFinalizedBlocks {
         result_sender: oneshot::Sender<broadcast::Receiver<BlockInfo>>,
     },
+    SubscribeBlendSession {
+        result_sender: oneshot::Sender<watch::Receiver<Option<SessionUpdate>>>,
+    },
+    SubscribeDASession {
+        result_sender: oneshot::Sender<watch::Receiver<Option<SessionUpdate>>>,
```
zeegomo, line 36: why the option?

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -94,6 +96,10 @@ pub enum Error {
     Storage(String),
     #[error("Blob validation failed: {0}")]
     BlobValidationFailed(#[from] blob::Error),
+    #[error("Block header not found: {0}")]
```
zeegomo: nit: header/block id not found

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -1123,6 +1160,46 @@ where
             if let Err(e) = lib_broadcaster.send(lib_update) {
                 error!("Could not notify LIB update to services: {e}");
             }
+
+            let previous_session_numbers = match cryptarchia.active_sessions_numbers(&prev_lib) {
```
zeegomo: wouldn't this be pruned when calling `try_apply_header?`

nomos-services/chain/broadcast-service/src/lib.rs
bacv: tokio::sync::watcher takes initial value when initializing. In case somebody subscribes to the service before the first session is provided, it will get None.

nomos-services/chain/chain-service/src/lib.rs
bacv: Fixed in https://github.com/logos-co/nomos/pull/1786/commits/aff06fa06392f49bb592cee34c8791f6f6e9f6cb

nomos-services/chain/chain-service/src/lib.rs
bacv: Good catch, fixed in https://github.com/logos-co/nomos/pull/1786/commits/aff06fa06392f49bb592cee34c8791f6f6e9f6cb

nomos-services/chain/broadcast-service/src/lib.rs
zeegomo: can it not receive anything?

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -1123,6 +1168,38 @@ where
...
                 error!("Could not notify LIB update to services: {e}");
             }
+
+            let sessions = cryptarchia.active_sessions_numbers(&new_lib)?;
+            for (service, session_number) in &sessions {
+                let prev_session_num = previous_session_numbers.get(service).copied().unwrap_or(0);
+
+                if session_number == &prev_session_num {
+                    continue;
+                }
```
zeegomo: nit: it'd prefer avoiding continue unless clearly convenient

zeegomo (approved):
Good work!

nomos-services/chain/broadcast-service/src/lib.rs
bacv: > can it not receive anything? 

Just because we need to hold the latest value, I've chose the `watcher` channel from tokio sync, which requires initial value. In case to not send anything, we could hold the latest value manually and use `tokio::sync::broadcast` channel as before.

PR #1792, opened 2025-10-08
Chore serdeop blanked impl

pradovic:

Context: https://github.com/logos-co/nomos/pull/1781#discussion_r2405262300.

This PR simplifies a SerdeOp blanket implementation to use self instead of generic param.

Yes

@pradovic @danielSanchezQ @andrussal 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/nomos-core/chain-defs/src/codec/mod.rs
+++ b/nomos-core/chain-defs/src/codec/mod.rs
@@ -5,68 +5,68 @@ pub(crate) mod bincode;
 pub mod errors;
 use std::error::Error as StdError;
 
-use bincode::{deserialize, serialize, serialized_size};
 use bytes::Bytes;
 pub use errors::Error;
 use serde::{Serialize, de::DeserializeOwned};
 pub type Result<T> = std::result::Result<T, Error>;
 
 /// Unified serialization trait for all wire and storage operations
-pub trait SerdeOp {
+pub trait SerdeOp: Serialize {
     type Error: StdError;
-    /// Serialize a value to bytes using wire format
-    fn serialize<T: Serialize>(value: &T) -> Result<Bytes>;
-    /// Deserialize bytes to a value using wire format
-    fn deserialize<T: DeserializeOwned>(data: &[u8]) -> Result<T>;
-    /// Get the serialized size of a value without actually serializing it
-    fn serialized_size<T: Serialize>(value: &T) -> Result<u64>;
+
+    fn to_bytes(&self) -> Result<Bytes>;
+
+    fn from_bytes(data: &[u8]) -> Result<Self>
+    where
+        Self: DeserializeOwned;
+
+    fn bytes_size(&self) -> Result<u64>;
 }
 
-impl<S> SerdeOp for S
-where
-    S: Serialize + DeserializeOwned,
-{
+impl<S: Serialize> SerdeOp for S {
     type Error = Error;
 
-    fn serialize<T: Serialize>(value: &T) -> Result<Bytes> {
-        serialize(value)
+    fn to_bytes(&self) -> Result<Bytes> {
+        bincode::serialize(self)
     }
 
-    fn deserialize<T: DeserializeOwned>(data: &[u8]) -> Result<T> {
-        deserialize(data)
+    fn from_bytes(data: &[u8]) -> Result<Self>
+    where
+        Self: DeserializeOwned,
+    {
+        bincode::deserialize(data)
     }
 
-    fn serialized_size<T: Serialize>(value: &T) -> Result<u64> {
-        serialized_size(value)
+    fn bytes_size(&self) -> Result<u64> {
+        bincode::serialized_size(self)
     }
 }
 
 #[cfg(test)]
 mod tests {
-
     use super::*;
 
     #[test]
     fn serialize_deserialize() {
         let tmp = String::from("much wow, very cool");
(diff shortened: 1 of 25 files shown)


nomos-core/chain-defs/src/codec/mod.rs
```diff
@@ -5,68 +5,68 @@ pub(crate) mod bincode;
...
-    /// Serialize a value to bytes using wire format
-    fn serialize<T: Serialize>(value: &T) -> Result<Bytes>;
-    /// Deserialize bytes to a value using wire format
-    fn deserialize<T: DeserializeOwned>(data: &[u8]) -> Result<T>;
-    /// Get the serialized size of a value without actually serializing it
-    fn serialized_size<T: Serialize>(value: &T) -> Result<u64>;
+
+    fn to_bytes(&self) -> Result<Bytes>;
+
+    fn from_bytes(data: &[u8]) -> Result<Self>
```
ntn-x2: I don't think we want to require `Serialize` for `from_bytes` impl? What about splitting the trait in 2, since in many cases we don't want to require both serialization and deserialization capabilities? We can require in the serialization trait for the type to implement `Serialize`, and in the deserialization trait for the type to implement `for<'de> Deserialize<'de>`? What do you think?

ntn-x2 (commented):
If the goal is to provide an opinionated implementation of how serialization and deserialization should happen, I think we can mirror serde traits and define our own that provide our opinionated implementation.

nomos-core/chain-defs/src/codec/mod.rs
danielSanchezQ: We dont need serialization bind here. In fact that bound should only added to the auto implementation below. So we can use it directly to anything that implements `Serialize+Deserialize`

nomos-core/chain-defs/src/codec/mod.rs
```diff
@@ -5,68 +5,68 @@ pub(crate) mod bincode;
...
-        serialize(value)
+    fn to_bytes(&self) -> Result<Bytes> {
+        bincode::serialize(self)
     }
 
-    fn deserialize<T: DeserializeOwned>(data: &[u8]) -> Result<T> {
-        deserialize(data)
+    fn from_bytes(data: &[u8]) -> Result<Self>
+    where
+        Self: DeserializeOwned,
```
danielSanchezQ: Why is the bound in the method instead of in the type S?

nomos-core/chain-defs/src/codec/mod.rs
ntn-x2: Yes, I agree. But still, it might make sense to split the trait in two, for serialization and deserialization separately, in my opinion.

danielSanchezQ: > Yes, I agree. But still, it might make sense to split the trait in two, for serialization and deserialization separately, in my opinion.

I'm not so sure, as this is intended for storage in where you want to load/store items. Otherwise we would have to add both anyway? Or maybe in some places we could add only one, but then the contract may not be that clear because you can just store something and not being able to retrieve (?)

danielSanchezQ: Why changing the method calls names tho?

pradovic: I did not like that it shared the names with the Serialize, Deserialize traits making us always write SerdeOp::variant when we use it together with serde. And it is kind of a different trait. I can return it back to original if it's better?

danielSanchezQ: No its fine, also is good naming actually because its always bytes

pradovic: Maybe something like this, to address all the points?
```
pub trait SerializeOp {
    type Error: StdError;
    
    fn to_bytes(&self) -> Result<Bytes, Self::Error>;
    fn bytes_size(&self) -> Result<u64, Self::Error>;
}

pub trait DeserializeOp: Sized {
    type Error: StdError;
    
    fn from_bytes(data: &[u8]) -> Result<Self, Self::Error>;
}

// Auto-implementation for types that implement Serialize
impl<T: Serialize> SerializeOp for T {
    type Error = Error;
    
    fn to_bytes(&self) -> Result<Bytes, Self::Error> {
        bincode::serialize(self)
    }
    
    fn bytes_size(&self) -> Result<u64, Self::Error> {
        bincode::serialized_size(self)
    }
}

// Auto-implementation for types that implement DeserializeOwned
impl<T: DeserializeOwned> DeserializeOp for T {
    type Error = Error;
    
    fn from_bytes(data: &[u8]) -> Result<Self, Self::Error> {
        bincode::deserialize(data)
    }
}

// Optional: Combined trait for convenience when you need both
pub trait SerdeOp: SerializeOp + DeserializeOp {}

// Auto-implement the combined trait
impl<T> SerdeOp for T 
where 
    T: SerializeOp + DeserializeOp 
{}
```

danielSanchezQ: > Maybe something like this, to address all the points?
Yeah, this looks good to me

pradovic: This is how it looks now: https://github.com/logos-co/nomos/pull/1792/commits/148491a7a76cc59f9106f3639b4b0f7c95a41734.

ntn-x2 (approved):
Nice!

Issue #1763, opened 2025-09-30
Epoch and session transition support for PoQ


ntn-x2:
The same that we do for session transitions, we need to do for epoch transitions when verifying proof of quotas. Spec was updated to include this: https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#27e261aa09df8009b7dfcd9f0c626a73.

PR #1793, opened 2025-10-08
feat(blend): support epoch transitions

ntn-x2:

PR 1/2, that together will fix https://github.com/logos-co/nomos/issues/1763. This PR specifically introduces a comprehensive changeset to the Blend core crates. The follow-up PR will update the Blend service and, if needed, its dependencies. Hence, the codebase is expected NOT to compile at the moment. It will with the work in the follow-up PR.

### Why this PR? 

Because we did not know exactly how to retrieve and interact with epoch information, we postponed this. Now that we have clear what data will come from where and most importantly **WHEN**, which is very important, we can finally fix the API for Blend to interact with PoQ generation and verification.

Specifically, there are now three traits for three different classes of proof generators: one for edge nodes that are only interested in blending messages, one for edge nodes that are interesting in mixing block proposals with cover traffic as well as regular core nodes, and one, currently unused but that could be helpful in some contexts, for nodes that produce only cover traffic without ever generating leadership proofs.

These components have then been integrated into the cryptographic processors, that now support sending only block proposals (suitable for edge nodes), sending block proposals and cover messages (also suitable for edge nodes with added privacy), and sending and receiving both block proposals and cover messages (suitable for core nodes).

The API for proof generation and proof verification have been updated to support epoch transitions, period in which verifiers will accept messages generated in the old epoch. For generators, as soon as a new epoch starts, old proofs are discarded and new ones are generated. For this to happen within the same session boundaries (i.e., when the object is not de-allocated and replaced with a new instance), I had to re-work how we pre-compute proofs in the background. We are still spawning a blocking task, but we have a way for it to detect if it should cease to exist by checking if the receiver of the proofs is still up: if the receiver side of the channel has been dropped, the task aborts. I wrote extensive tests for this behaviour.

Also, leadership quota is applied per message, not per session, meaning that it enforces that each message can be encapsulated a maximum of `N` times, but does not limit how many messages a leader can propagate via the Blend network (obviously): this has also been fixed in this PR, where the background task for leadership proof generation basically never terminates (unless a new epoch or a new session starts) but it keeps yielding proofs at groups of `N`, where `N` is the number of encapsulation layers per block proposal.

TL;DR: All the components of the Blend stack, be them proof generators or proof verifiers, now support epoch rotation + transition as well as full session transition. This is the section in the spec that was recently added and that this PR, along with the next one building on top of this (coming soon), is tackling: https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#27e261aa09df8009b7dfcd9f0c626a73.

Yes.

@ntn-x2 @madxor 

Yes.

No, but it implements some additions introduced in the Blend spec, see link above.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nodes/nomos-node/node/src/generic_services/blend.rs
+++ b/nodes/nomos-node/node/src/generic_services/blend.rs
@@ -321,7 +321,7 @@ where
                 };
 
                 PolEpochInfo {
-                   nonce: *epoch_nonce,
+                    nonce: *epoch_nonce,
                     poq_private_inputs: ProofOfLeadershipQuotaInputs {
                         aged_path_and_selectors: aged_path_and_selectors.try_into().expect("List of aged note paths and selectors does not match the expected size for PoQ inputs, although it has already been pre-processed."),
                         note_value: *note_value,
--- a/nomos-services/blend/src/core/mod.rs
+++ b/nomos-services/blend/src/core/mod.rs
@@ -331,7 +331,7 @@ where
             mut remaining_clock_stream,
         ) = async {
             let (clock_tick, remaining_clock_stream) =
-                UninitializedFirstReadyStream::new(clock_stream, Duration::from_secs(10))
+                UninitializedFirstReadyStream::new(clock_stream, Duration::from_secs(5))
                     .first()
                     .await
                     .expect("The clock system must be available.");
--- a/nomos-services/blend/src/edge/mod.rs
+++ b/nomos-services/blend/src/edge/mod.rs
@@ -248,7 +248,7 @@ where
                 FIRST_STREAM_ITEM_READY_TIMEOUT,
                 settings.time.session_transition_period(),
             ),
-            UninitializedFirstReadyStream::new(clock_stream, Duration::from_secs(3)),
+            UninitializedFirstReadyStream::new(clock_stream, FIRST_STREAM_ITEM_READY_TIMEOUT),
             messages_to_blend_stream,
             epoch_handler,
             &settings,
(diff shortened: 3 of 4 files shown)


nomos-blend/scheduling/src/message_blend/crypto/core_and_leader/send.rs
```diff
@@ -0,0 +1,328 @@
...
+    serialize_encapsulated_message,
+};
+
+/// [`SessionCryptographicProcessor`] is responsible for only wrapping
+/// cover and data messages for the message indistinguishability.
+///
+/// Each instance is meant to be used during a single session.
+///
+/// This processor is suitable for non-core nodes that want to generate noise
+/// (i.e., cover) traffic along with data payloads.
```
youngjoon-lee, line 34: Is this only for edge nodes (non-core nodes)? I'm asking because this is in the `core_and_leader` module. If this is for both core and edge nodes (that want to generate cover traffic as well), I think the comment needs to be slightly updated.

nomos-blend/scheduling/src/message_blend/crypto/core_and_leader/send.rs
```diff
@@ -0,0 +1,328 @@
...
+    ) -> Result<Vec<u8>, Error> {
+        Ok(serialize_encapsulated_message(
+            &self.encapsulate_data_payload(payload).await?,
+        ))
+    }
+
+    // TODO: Think about optimizing this by, e.g., using less encapsulations if
+    // there are less than 3 proofs available, or use a proof from a different pool
+    // if needed (core proof for leadership message or leadership proof for
+    // cover message, since the protocol does not enforce that).
```
youngjoon-lee, line 134: IMO, using less encapsulations sounds simpler than the other way. But I agree it's not urgent.

nomos-blend/scheduling/src/message_blend/crypto/core_and_leader/send.rs
youngjoon-lee: Ah, never mind, I just saw the send_and_receiver submodule below, which makes me recall the previous structure. Now I understand why this code comment is saying so.

nomos-blend/scheduling/src/message_blend/provers/core/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+    // This will kill the previous running task, if any, since we swap the receiver
+    // channel, hence the old task will fail to send new proofs and will abort on
+    // its own.
+    fn spawn_new_proof_generation_task(&mut self, starting_key_index: u64) {
+        if self.remaining_quota == 0 {
+            return;
+        }
+
+        let (proofs_sender, proofs_receiver) = channel(self.remaining_quota as usize);
+        self.proof_generation_task_handle = Some(spawn_core_proof_generation_task(
```
youngjoon-lee, line 110: Don't we need to call `terminate_proof_generation_task` before spawning the new one?

nomos-blend/scheduling/src/message_blend/crypto/core_and_leader/send.rs
ntn-x2: Yes. `core_and_leader` means core and leader proof generator, so generating both cover messages as well as data payloads (block proposals).

nomos-blend/scheduling/src/message_blend/crypto/core_and_leader/send.rs
ntn-x2: Yes this comment was already present before this PR, and it's something we can optimize later on.

nomos-blend/scheduling/src/message_blend/provers/core/mod.rs
ntn-x2: Not really. When generating a new channel, the old receiver is dropped, and the task will be aware of it and terminate automatically, so we don't need to explicitly call `terminate_proof_generation_task` in this case, since the channel will be dropped anyway (which is what `terminate_proof_generation_task` basically does).

ntn-x2: We have some test cases that make sure the task is actually aborted.

youngjoon-lee: Thank you!

PR #1794, opened 2025-10-09
feat: support epoch transition ticks in epoch handler

ntn-x2:

Small PR built on top of https://github.com/logos-co/nomos/pull/1793, and needed to integrate epoch transition into Blend service (this means the codebase is still expected not to compile, until we introduce the necessary changes into Blend service, which will happen in the next PR). I modified the epoch handler and assumed there is always an epoch state associated with a slot of the current new epoch, as discussed with the cryptarchia people. This simplifies things quite a bit since we don't need to retry until the state can be fetched, but we can simply assume it will always succeed.

Also, there were a couple of edge cases that we needed to handle better, that are unlikely in production but that COULD happen in tests: namely that the epoch transition session is set to 1 slot, or that epoch transition has the same length (or greater) than the epoch duration itself. These cases are taken care of now, and make the code a little bit more complicated, but I think protection against those cases is needed to avoid introducing subtle bugs that will be hard(er) to debug. I also extended the test coverage to test for all these new cases.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/epoch_info.rs
+++ b/nomos-services/blend/src/epoch_info.rs
@@ -298,17 +298,16 @@ where
             self.epoch_tracking_state = Some(EpochTrackingState::new(validated_slot_tick));
         }
 
-        // We check if the previous epoch is expired only if two epochs in the past are
-        // not, since the two conditions cannot exist at the same time.
-        let should_notify_about_past_epoch_transition = should_notify_about_two_epochs_back
-            || self.check_and_consume_past_epoch_transition_period(validated_slot_tick);
-
-        let epoch_event =
-            if should_notify_about_two_epochs_back || should_notify_about_past_epoch_transition {
-                EpochEvent::NewEpochAndOldEpochTransitionExpired(epoch_state.into())
-            } else {
-                EpochEvent::NewEpoch(epoch_state.into())
-            };
+        // We need to notify about epoch transitioncheck if the previous epoch is
+        // expired only if two epochs in the past are not, since the two
+        // conditions cannot exist at the same time.
+        let epoch_event = if should_notify_about_two_epochs_back
+            || self.check_and_consume_past_epoch_transition_period(validated_slot_tick)
+        {
+            EpochEvent::NewEpochAndOldEpochTransitionExpired(epoch_state.into())
+        } else {
+            EpochEvent::NewEpoch(epoch_state.into())
+        };
 
         Some(epoch_event)
     }


nomos-services/blend/src/epoch_info.rs
```diff
@@ -70,87 +81,257 @@ where
...
+            current_epoch: epoch,
+        }
+    }
+
+    /// Return the currently tracked epoch.
+    const fn last_processed_epoch(&self) -> Epoch {
+        match self {
+            Self::FirstEpoch { current_epoch } | Self::SecondOrLaterEpoch { current_epoch, .. } => {
+                *current_epoch
+            }
```
youngjoon-lee, line 196: What does `processed_` mean in the function name? Would `current_epoch` be a clearer name here?

nomos-services/blend/src/epoch_info.rs
```diff
@@ -70,87 +81,257 @@ where
...
+    /// epoch information from the chain API.
+    FirstEpoch { current_epoch: Epoch },
+    /// The handler has witnessed an epoch change, and it keeps minimal info
+    /// about the previous epoch, to notify consumer when the configured
+    /// epoch transition period has elapsed.
+    SecondOrLaterEpoch {
+        /// The last slot of the previous epoch. `None` when the consumers have
+        /// been notified of the previous epoch, and `Some` for the time between
+        /// a new epoch and the old epoch transition period.
+        previous_epoch_last_slot: Option<Slot>,
```
youngjoon-lee, line 165: >  `Some` for the time between a new epoch and the old epoch transition period.

Does this comment mean that this value remains `Some` until the old epoch transition period has passed? It seems so, based on `check_and_consume_past_epoch_transition_period` below.

nomos-services/blend/src/epoch_info.rs
```diff
@@ -70,87 +81,257 @@ where
...
+                return Some(EpochEvent::OldEpochTransitionPeriodExpired);
             }
+            return None;
+        }
+
+        tracing::debug!(target: LOG_TARGET, "Found epoch unseen before. Polling for its state...");
+        let epoch_state = self
+            .chain_service
+            .get_epoch_state_for_slot(new_tick.slot)
+            .await;
```
youngjoon-lee:
```suggestion
        tracing::debug!(target: LOG_TARGET, "Found epoch unseen before. Pulling for its state...");
        let epoch_state = self
            .chain_service
            .get_epoch_state_for_slot(new_tick.slot)
            .await;
```
I think we fetch the epoch state without polling.

nomos-services/blend/src/epoch_info.rs
```diff
@@ -70,87 +81,257 @@ where
...
+            if let Some(EpochTrackingState::SecondOrLaterEpoch {
+                previous_epoch_last_slot,
+                ..
+            }) = self.epoch_tracking_state
+                && previous_epoch_last_slot.is_some()
+            {
+                true
+            } else {
+                false
+            };
```
youngjoon-lee, line 290: It seems that `should_notify_about_two_epochs_back` is `true` even when epochs are **not** shorter than the transition period, because the `previous_epoch_last_slot` remains `Some` until the transition period has passed, if I understood it correctly. Could you please correct my understanding?

nomos-services/blend/src/epoch_info.rs
```diff
@@ -70,87 +81,257 @@ where
...
+        // not, since the two conditions cannot exist at the same time.
+        let should_notify_about_past_epoch_transition = should_notify_about_two_epochs_back
+            || self.check_and_consume_past_epoch_transition_period(validated_slot_tick);
+
+        let epoch_event =
+            if should_notify_about_two_epochs_back || should_notify_about_past_epoch_transition {
+                EpochEvent::NewEpochAndOldEpochTransitionExpired(epoch_state.into())
+            } else {
+                EpochEvent::NewEpoch(epoch_state.into())
+            };
```
youngjoon-lee: Is calling `self.check_and_consume_...` for checking the case where the transition period is 1s?

I think we can merge the line 303-304 into the line 307.

nomos-services/blend/src/epoch_info.rs
youngjoon-lee: Also, what do you think about enforcing that the transition period is always shorter than the epoch duration, by checking it when `EpochHandler` is created? It's not type-safe, but we could simplify the logic.

nomos-services/blend/src/epoch_info.rs
ntn-x2: It's the last epoch that we notified to consumers. You are right, the naming could be improved, but `current_epoch` could be misleading for those cases in which we get a new epoch tick and we need to notify consumers for this one. Probably something like `last_notified_epoch` would be better? Or do you have any other suggestions?

nomos-services/blend/src/epoch_info.rs
ntn-x2: Yes, exactly. Is it not clear? Should I improve it? We take the `Some` upon notifying consumers about the end of the transition period.

nomos-services/blend/src/epoch_info.rs
ntn-x2: Yes, you are right. This is a left-over for when this type implemented `Stream`. Good catch, will fix that.

nomos-services/blend/src/epoch_info.rs
ntn-x2: > It seems that should_notify_about_two_epochs_back is true even when epochs are not shorter than the transition period, because the previous_epoch_last_slot remains Some until the transition period has passed, if I understood it correctly. Could you please correct my understanding?

No this happens before we update the epoch info with the epoch that just ended, so the information here refers to the epoch before the one ending now. So if epoch is shorter than transition period, this will be updated anyway on the next new epoch. We check again _after_ processing the new epoch, for cases in which transition period is set to `1`.

> Also, what do you think about enforcing that the transition period is always shorter than the epoch duration, by checking it when EpochHandler is created? It's not type-safe, but we could simplify the logic.

Yeah unfortunately that is not trivial. This function only exposes a `tick` method, which is "propagated" from the time service. So Blend has no idea how long an epoch is (and it should not, in my view). Hence we cannot perform any validation based on epoch duration, but only, if needed, on epoch transition period, which is something strictly relevant to Blend. Hence the solution here to account for those cases, because we cannot verify them before-hand. In a previous implementation I tried to maintain internal trace of how long an epoch lasted, then failed if it was less than the transition period, but we would still need to wait to observe a full epoch (i.e., we cannot draw conclusions if we join mid-epoch), and then we would only fail after the epoch is passed, which is very very bad design. On the base of this information, I decided to go with this solution. What do you think?

nomos-services/blend/src/epoch_info.rs
ntn-x2: Yes this check explicitly checks if the epoch we just rotated from is expired, which can only happen if the transition period is set to 1 slot. What do you mean by merging the lines, exactly?

nomos-services/blend/src/epoch_info.rs
ntn-x2: https://github.com/logos-co/nomos/pull/1794/commits/60d76ef60831a72830acbbc6a0fab61e6060dc28

nomos-services/blend/src/epoch_info.rs
youngjoon-lee: Ah, now I understand why this name makes sense.
https://github.com/logos-co/nomos/blob/0ac5377e72faf5c9c3ed9e62a64d8477702dad87/nomos-services/blend/src/epoch_info.rs#L259

Thank you. I think we can keep this name.

nomos-services/blend/src/epoch_info.rs
youngjoon-lee: I didn't immediately understand what the comment means when I just read it, but it became clear after reading the actual logic. Thank you. I just wanted to confirm that I understood it correctly.

nomos-services/blend/src/epoch_info.rs
youngjoon-lee: > What do you mean by merging the lines, exactly?

I think we can do like: 
```diff
diff --git a/nomos-services/blend/src/epoch_info.rs b/nomos-services/blend/src/epoch_info.rs
index 0a5f00c6..cb5f1364 100644
--- a/nomos-services/blend/src/epoch_info.rs
+++ b/nomos-services/blend/src/epoch_info.rs
@@ -300,15 +300,13 @@ where
 
         // We check if the previous epoch is expired only if two epochs in the past are
         // not, since the two conditions cannot exist at the same time.
-        let should_notify_about_past_epoch_transition = should_notify_about_two_epochs_back
-            || self.check_and_consume_past_epoch_transition_period(validated_slot_tick);
-
-        let epoch_event =
-            if should_notify_about_two_epochs_back || should_notify_about_past_epoch_transition {
-                EpochEvent::NewEpochAndOldEpochTransitionExpired(epoch_state.into())
-            } else {
-                EpochEvent::NewEpoch(epoch_state.into())
-            };
+        let epoch_event = if should_notify_about_two_epochs_back
+            || self.check_and_consume_past_epoch_transition_period(validated_slot_tick)
+        {
+            EpochEvent::NewEpochAndOldEpochTransitionExpired(epoch_state.into())
+        } else {
+            EpochEvent::NewEpoch(epoch_state.into())
+        };
 
         Some(epoch_event)
     }

```

nomos-services/blend/src/epoch_info.rs
youngjoon-lee: > No this happens before we update the epoch info with the epoch that just ended, so the information here refers to the epoch before the one ending now. So if epoch is shorter than transition period, this will be updated anyway on the next new epoch. We check again after processing the new epoch, for cases in which transition period is set to 1.

Ahah, right. It makes sense because the `should_notify_about_two_epochs_back` is evaluated **after** checking the condition below:
https://github.com/logos-co/nomos/blob/0ac5377e72faf5c9c3ed9e62a64d8477702dad87/nomos-services/blend/src/epoch_info.rs#L255-L259
Thank you for your explanation.

> So Blend has no idea how long an epoch is (and it should not, in my view). Hence we cannot perform any validation based on epoch duration, but only, if needed, on epoch transition period, which is something strictly relevant to Blend.

Ah, yeah. I now understand the issue. The epoch duration is a parameter for Blend, but epoch duration is defined outside of Blend. Hm, yeah, I can't find any way to validate them beforehand. I agree with your approach. 

> we would still need to wait to observe a full epoch (i.e., we cannot draw conclusions if we join mid-epoch), and then we would only fail after the epoch is passed, which is very very bad design.

Yeah, I also think that's a bad design.

nomos-services/blend/src/epoch_info.rs
ntn-x2: Yeah, you are right. This is coding while sick, ops... Good catch! https://github.com/logos-co/nomos/pull/1794/commits/ce88dec8ddd49f3557fc7092ce49a0694c9480f2

PR #1795, opened 2025-10-09
feat(chain): Add genesis_tx as chain initialization method

zeegomo:
Add the possibility to launch chain-service from a genesis-tx instead of a ledger state, as specified in https://www.notion.so/nomos-tech/Bedrock-Genesis-Block-21d261aa09df80bb8dc3c768802eb527. 
Difference from the spec:
* we accept a single tx as input instead of the whole block, block construction is deterministic once you know the tx so we skip validation this way since we know we build it correctly. Note this is not a real deviation from the spec, as we still compute the header id as expected by other nodes.
* we still don't read the inscription to extract params, this is pending an update to the format of the inscription

Yes.

@davidrusu @zeegomo 

Yes, but inscription format is pending an update

See above

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nodes/nomos-executor/config.yaml
+++ b/nodes/nomos-executor/config.yaml
@@ -198,7 +198,6 @@ cryptarchia:
             pk: d204000000000000000000000000000000000000000000000000000000000000
       execution_gas_price: 0
       storage_gas_price: 0
-    genesis_nonce: "0000000000000000000000000000000000000000000000000000000000000000"
   network_adapter_settings:
     topic: /cryptarchia/proto
   recovery_file: ./recovery/cryptarchia.json
--- a/nomos-services/chain/chain-service/src/lib.rs
+++ b/nomos-services/chain/chain-service/src/lib.rs
@@ -344,8 +344,6 @@ where
 pub enum StartingState {
     Genesis {
         genesis_tx: GenesisTx,
-        #[serde(with = "groth16::serde::serde_fr")]
-        genesis_nonce: groth16::Fr, // TODO: recover from genesis tx
     },
     Lib {
         lib_id: HeaderId,
--- a/nomos-services/chain/chain-service/src/states.rs
+++ b/nomos-services/chain/chain-service/src/states.rs
@@ -1,5 +1,6 @@
 use std::{collections::HashSet, hash::Hash, marker::PhantomData, time::SystemTime};
 
+use groth16::{Field as _, Fr};
 use nomos_core::header::{Header, HeaderId};
 use nomos_ledger::LedgerState;
 use overwatch::{DynError, services::state::ServiceState};
@@ -71,15 +72,12 @@ where
         settings: &<Self as ServiceState>::Settings,
     ) -> Result<Self, <Self as ServiceState>::Error> {
         let (lib_id, genesis_id, lib_ledger_state) = match &settings.starting_state {
-            StartingState::Genesis {
-                genesis_tx,
-                genesis_nonce,
-            } => {
+            StartingState::Genesis { genesis_tx } => {
                 let lib_id = Header::genesis(genesis_tx).id();
                 let ledger = LedgerState::from_genesis_tx(
                     genesis_tx.clone(),
                     &settings.config,
-                    *genesis_nonce,
+                    Fr::ZERO, // TODO: recover from genesis tx
                 )?;
                 (lib_id, lib_id, ledger)
             }
--- a/nodes/nomos-node/config.yaml
+++ b/nodes/nomos-node/config.yaml
@@ -187,7 +187,6 @@ cryptarchia:
             pk: d204000000000000000000000000000000000000000000000000000000000000
       execution_gas_price: 0
       storage_gas_price: 0
-    genesis_nonce: "0000000000000000000000000000000000000000000000000000000000000000"
   network_adapter_settings:
     topic: /cryptarchia/proto
   recovery_file: ./recovery/cryptarchia.json
--- a/tests/Cargo.toml
+++ b/tests/Cargo.toml
@@ -20,7 +20,6 @@ ed25519-dalek            = { version = "3.0.0-pre.0", features = ["rand_core", "
 executor-http-client     = { workspace = true }
 futures                  = { default-features = false, version = "0.3" }
 futures-util             = { default-features = false, version = "0.3.31" }
-groth16                  = { workspace = true }
 hex                      = { default-features = false, version = "0.4.3" }
 kzgrs-backend            = { workspace = true }
 nomos-api                = { workspace = true }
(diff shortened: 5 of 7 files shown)


nomos-services/chain/chain-service/src/states.rs
```diff
@@ -68,17 +91,29 @@ where
...
+            StartingState::Lib {
+                lib_id,
+                lib_ledger_state,
+                genesis_id,
+            } => Ok(Self::Lib {
+                tip: *lib_id,
+                lib: *lib_id,
+                genesis_id: *genesis_id,
+                lib_ledger_state: lib_ledger_state.clone(),
                 lib_block_length: 0,
```
youngjoon-lee: What do you think about removing the `StartingState::Lib` variant? This `from_settings` is called only when [`StateOperator::try_load`](https://github.com/logos-co/Overwatch/blob/main/overwatch/src/services/state/service_state.rs#L24) doesn't return a state. In that case, I think we can always go with the `StartingState::Genesis` (which is similar as the previous `CryptarchiaSettings`).
Do we have any scenario where a user actually needs to specify the `StartingState::Lib` in his config file?

zeegomo: How would you start from a checkpoint?

nomos-core/chain-defs/src/utils/ed25519_serde.rs
```diff
@@ -0,0 +1,82 @@
...
+pub fn serialize<S>(bytes: &[u8], serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    if serializer.is_human_readable() {
+        serializer.serialize_str(&hex::encode(bytes))
+    } else {
+        serializer.serialize_bytes(bytes)
+    }
+}
```
danielSanchezQ, line 16: Dont we have hex_bytes somewhere already? 🤔

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -325,14 +361,27 @@ where
...
     pub bootstrap: BootstrapConfig<NodeId>,
     pub sync: SyncConfig,
 }
 
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub enum StartingState {
+    Genesis {
+        genesis_tx: GenesisTx,
+        #[serde(with = "groth16::serde::serde_fr")]
+        genesis_nonce: groth16::Fr, // TODO: recover from genesis tx
```
danielSanchezQ: Can it be a method on `GenesisTx`?

nomos-services/chain/chain-service/src/states.rs
```diff
@@ -1,28 +1,40 @@
...
+        last_engine_state: Option<LastEngineState>,
+        // Only neededed for the service state trait
+        _markers: PhantomData<(NodeId, NetworkAdapterSettings)>,
+    },
+    Genesis {
+        genesis_tx: GenesisTx,
+        #[serde(with = "groth16::serde::serde_fr")]
+        genesis_nonce: groth16::Fr, // TODO: recover from genesis tx
+    },
 }
```
danielSanchezQ: Why the distinction? Once loaded the `GenesisTx`, can't it be considered the first `LIB`?

nomos-services/chain/chain-service/src/states.rs
youngjoon-lee: Yeah, you're right. This is necessary for starting from a checkpoint (which will be implemented later). Thanks.

nomos-services/chain/chain-service/src/states.rs
youngjoon-lee: I guess it's because of the requirement of Overwatch. The chain service has to load an initial state from the settings at the very beginning. Then, it can finally process the genesis tx and populate the ledger.

nomos-services/chain/chain-service/src/lib.rs
zeegomo: we can if we're fine with not being able to specify one until we've stabilized the inscription format, I don't think we ever used that anyway

nomos-core/chain-defs/src/utils/ed25519_serde.rs
zeegomo: mmm, let me check if I can use it

nomos-services/chain/chain-service/src/lib.rs
danielSanchezQ: > we can if we're fine with not being able to specify one until we've stabilized the inscription format, I don't think we ever used that anyway

Ok, I get it. Can be like this for now until we transition then.

nomos-services/chain/chain-service/src/states.rs
zeegomo: > Once loaded the GenesisTx, can't it be considered the first LIB?

Yes, but the initialization is different, maybe I could initialize here and just pass the lib state to the service, let me check

nomos-core/chain-defs/src/utils/ed25519_serde.rs
zeegomo: we do have a macro for byte newtypes if that's what you were referring to that

nomos-services/chain/chain-service/src/lib.rs
zeegomo: removed that in the last commit, I don't think we will need that now and in the future it will be recovered from the tx itself

nomos-services/chain/chain-service/src/states.rs
zeegomo: updated

nomos-core/chain-defs/src/utils/ed25519_serde.rs
zeegomo: for additional context, this will probably be overwritten by david's work, but it's needed atm to make tests work since yaml can't deal with raw bytes

nodes/nomos-executor/config.yaml
```diff
@@ -183,64 +182,22 @@ cryptarchia:
...
+      ledger_tx:
+        inputs: []
+        outputs:
+          - value: 1
+            pk: d204000000000000000000000000000000000000000000000000000000000000
+      execution_gas_price: 0
+      storage_gas_price: 0
   network_adapter_settings:
     topic: /cryptarchia/proto
   recovery_file: ./recovery/cryptarchia.json
```
davidrusu, line 203: nice reduction in the config :)

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -325,14 +333,25 @@ where
...
+pub enum StartingState {
+    Genesis {
+        genesis_tx: GenesisTx,
+    },
+    Lib {
+        lib_id: HeaderId,
+        lib_ledger_state: Box<LedgerState>,
+        genesis_id: HeaderId,
+    },
+}
```
davidrusu, line 353: An alternative would be to have a `Lib::from_genesis_tx()` builder for the lib state

nomos-services/chain/chain-service/src/states.rs
youngjoon-lee: Great

PR #1796, opened 2025-10-10
feat(api): Blocks and blocks stream

strinnityk:
Implement the `/blocks` and `/blocks/stream` endpoints under the `block-explorer` feature flag. They're not optimized and only will be used in that context, for now.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/nodes/nomos-node/node/src/api/responses/ndjson.rs
+++ b/nodes/nomos-node/node/src/api/responses/ndjson.rs
@@ -6,7 +6,6 @@ use futures::{Stream, StreamExt as _};
 use http::StatusCode;
 use nomos_api::http::DynError;
 use serde::Serialize;
-// TODO: from stream result<T> to T
 
 pub fn from_stream<T>(stream: impl Stream<Item = T> + Send + 'static) -> Response
 where


nomos-core/chain-defs/src/header/mod.rs
```diff
@@ -55,7 +55,7 @@ impl Header {
     }
 
     #[must_use]
-    pub fn leader_proof(&self) -> &impl LeaderProof {
+    pub const fn leader_proof(&self) -> &Groth16LeaderProof {
```
strinnityk, line 79: Is there a specific reason this is `&impl LeaderProof`? Should I roll it back?

nodes/nomos-node/node/src/api/responses/ndjson.rs
```diff
@@ -0,0 +1,88 @@
+use axum::{
+    body::Body,
+    response::{IntoResponse as _, Response},
+};
+use futures::{Stream, StreamExt as _};
+use http::StatusCode;
+use nomos_api::http::DynError;
+use serde::Serialize;
+// TODO: from stream result<T> to T
```
danielSanchezQ: leftover?

nomos-core/chain-defs/src/header/mod.rs
danielSanchezQ: cc @zeegomo

nomos-services/api/src/http/mantle.rs
```diff
@@ -113,3 +132,204 @@ where
 
     Ok(stream)
 }
+
+#[cfg(feature = "block-explorer")]
```
danielSanchezQ, line 136: Not sure we need to feature gate this 🤔

strinnityk: It's for the lints. This function is only used within another `block-explorer` feature gate.

nodes/nomos-node/node/src/api/responses/ndjson.rs
strinnityk: Indeed. Thanks.

nomos-core/chain-defs/src/header/mod.rs
zeegomo: can't recall right now, but fair to use the specific type here

PR #1804, opened 2025-10-15
feat(chain): change logic for pre-computing winning slots

ntn-x2:

Since Blend sessions are not aligned with epochs, it is very likely the case that Blend services can start/stop at session changes, and need information about the current ongoing epoch. Since there can be potentially a lot of winning slots, I changed the design of the subscribe API to pre-compute the first winning slot of the epoch when the chain leader is started or when a new epoch starts. Then, for subsequent slots we piggy-back on the existing block proposal logic. To support subscribers joining mid-epoch, I changed the channel type from broadcast to watch, so the latest computed slot, if any, is returned automatically to new subscribers.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-services/chain/chain-leader/src/lib.rs
+++ b/nomos-services/chain/chain-leader/src/lib.rs
@@ -29,7 +29,7 @@ use overwatch::{
 use serde::{Deserialize, Serialize, de::DeserializeOwned};
 use services_utils::wait_until_services_are_ready;
 use thiserror::Error;
-use tokio::sync::{broadcast, oneshot};
+use tokio::sync::{oneshot, watch};
 use tracing::{Level, error, info, instrument, span};
 use tracing_futures::Instrument as _;
 use tx_service::{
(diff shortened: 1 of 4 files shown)


nomos-services/chain/chain-leader/src/lib.rs
```diff
@@ -63,15 +63,20 @@ pub enum Error {
...
+    /// * this service is started mid-epoch -> immediately the first winning
+    ///   slot of the ongoing epoch (the slot can also be in the past compared
+    ///   to the current slot as returned by the time service), if any
+    /// * a new winning slot (other than the very first one in the ongoing
+    ///   epoch) is identified when proposing blocks
+    /// * a new consumer subscribes -> the latest value that was sent to all the
+    ///   other consumers, if any
     WinningPolEpochSlotStreamSubscribe {
-        sender: oneshot::Sender<broadcast::Receiver<(LeaderPrivate, SecretKey, Epoch)>>,
+        sender: oneshot::Sender<watch::Receiver<Option<(LeaderPrivate, SecretKey, Epoch)>>>,
```
youngjoon-lee, line 79: If the `watch::Receiver` yields `None`, does it mean that there will be no winning slot in the session? I couldn't find any code which does `sender.send(None)` in the `leadership.rs`. Could you tell me where it is?

ntn-x2: The `Receiver` will receive `None` only at the very beginning, when the chain leader service starts, and before it pre-computes the very first value. Unfortunately, a watch channel needs an initial value, hence we need to create it with `None` and have `Option` as the type. But this is guaranteed to never be `None` after it switches from `None` to `Some`.

As for the `send` call, it happens in two different cases: when a new epoch is processed (including the very first time the chain leader service is started), which then happens here: https://github.com/logos-co/nomos/pull/1804/files#diff-f665f701c4ebd4d16041af403817aa6f9eb6f622857cd08d3209b9940f9e73b1R246.

And then, everytime the leader finds out a new winning slot within the current epoch, which then calls `notify_about_winning_slot` (https://github.com/logos-co/nomos/pull/1804/files#diff-f665f701c4ebd4d16041af403817aa6f9eb6f622857cd08d3209b9940f9e73b1R80), which also sends the new slot over the channel (https://github.com/logos-co/nomos/pull/1804/files#diff-f665f701c4ebd4d16041af403817aa6f9eb6f622857cd08d3209b9940f9e73b1R269). In this case, the notifier will ignore the slot if it's the same one as the pre-computed one, since we don't want to duplicate notifications about it.

Let me know if you have any other outstanding questions.

ntn-x2: In any case, if Blend as a consumer subscribes before the first value is produced, it won't anyway get the `None` value, because we explicitly filter them out here: https://github.com/logos-co/nomos/pull/1804/files#diff-92b5cd2e69659baa6db5bd800656f8a07483b8faab02f09cf2612b7ded04ed7bR238 by using a `filter_map` on the `WatchStream`. So effectively the stream will only yield once `Some` value is available in the channel, which will be the case if either a service subscribes when a value has already been computed by the chain leader, or right when the chain leader pre-computes the very first value.

youngjoon-lee: Aha, I missed the `filter_map`. Thank you. It makes sense to have `Option` as the type for the initial value of the watch channel 👍

youngjoon-lee (approved):
👍

PR #1805, opened 2025-10-15
feat(chain): apply txs to ledger

andrussal:

This PR applies transactions to ledger by removing dummy placeholder. Together with this change, it also updates integration tests to create new channel so that blob channel ops will be valid on ledger.

(This PR is based on [feat/consensus-refactor-base](https://github.com/logos-co/nomos/tree/feat/consensus-refactor-base) which introduced Proposal to leader and chain service. After this PR, it should be fine to merge [feat/consensus-refactor-base](https://github.com/logos-co/nomos/tree/feat/consensus-refactor-base)  to master)

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -215,7 +212,14 @@ impl Cryptarchia {
 
     /// Create a new [`Cryptarchia`] with the updated state.
     #[must_use = "Returns a new instance with the updated state, without modifying the original."]
-    fn try_apply_header(&self, header: &Header) -> Result<(Self, PrunedBlocks<HeaderId>), Error> {
+    fn try_apply_header<Tx>(
+        &self,
+        header: &Header,
+        txs: impl Iterator<Item = Tx>,
+    ) -> Result<(Self, PrunedBlocks<HeaderId>), Error>
```
davidrusu: this is no longer just applying a header, rename to `try_apply_block` and pass it a `block: Block<Mempool::Item>` rather than header and txs?

nomos-services/tx-service/src/tx/service.rs
```diff
@@ -533,24 +531,37 @@ where
...
-            (Vec::new(), Vec::new()),
-            |(mut found_transactions, mut not_found_hashes), hash| {
-                if let Some(transaction) = pool.get_item(&hash).cloned() {
-                    found_transactions.push(transaction);
-                } else {
-                    not_found_hashes.push(hash);
+    ) -> Result<crate::TransactionsByHashesResponse<Pool::Item, Pool::Key>, MempoolError> {
+        let keys_set: BTreeSet<Pool::Key> = hashes.into_iter().collect();
+
+        match pool.get_items_by_keys(&keys_set).await {
```
davidrusu: can we rewrite with `?` to keep indentation low:
```rust
let item_stream = pool.get_items_by_keys(&keys_set).await
    .map_err(|e| MempoolError::StorageError(format!("Failed to get items by keys: {e:?}")))?;

let found_transactions: Vec<...
```

nomos-services/tx-service/src/lib.rs
```diff
@@ -21,25 +22,28 @@ pub struct TransactionsByHashesResponse<Item, Key> {
...
 }
 
-impl<Item, Key> TransactionsByHashesResponse<Item, Key> {
+impl<Item, Key> TransactionsByHashesResponse<Item, Key>
+where
+    Key: Ord,
+{
     #[must_use]
-    pub const fn new(found_transactions: Vec<Item>, not_found_hashes: Vec<Key>) -> Self {
+    pub const fn new(found_transactions: Vec<Item>, not_found_hashes: BTreeSet<Key>) -> Self {
```
davidrusu: nit: rename params to `found`, `not_found` so we can have use the shortcut in initializing `Self { found, not_found }`

tests/src/tests/da/disperse.rs
```diff
@@ -22,13 +23,16 @@ async fn disseminate_and_retrieve() {
     let executor = &topology.executors()[0];
     let validator = &topology.validators()[0];
 
+    tokio::time::sleep(Duration::from_secs(15)).await;
```
davidrusu, line 26: Why 15 seconds? Does setting up the test channel fail if we do it too quickly? Is there a ready check we can do instead of this 15s delay (i.e. maybe we wait until chain height grows to 1?) (feels like it could be a flaky test in the future)

nomos-services/api/src/http/mempool.rs
```diff
@@ -36,7 +36,8 @@ where
         + Clone
         + 'static,
     StorageAdapter::Error: Debug,
-    Item: Clone + Debug + Send + Sync + Serialize + for<'de> Deserialize<'de> + 'static,
+    Item:
+        Transaction + Clone + Debug + Send + Sync + Serialize + for<'de> Deserialize<'de> + 'static,
     Key: Clone + Debug + Ord + Hash + Send + Sync + Serialize + for<'de> Deserialize<'de> + 'static,
```
davidrusu: nit: `for<'de> Deserialize<'de>` is identical to `DeserializedOwned` 

(I know it's not from your change but I think it's less syntax noise)

nomos-services/chain/chain-service/src/mempool/adapter.rs
```diff
@@ -64,6 +64,6 @@ where
             .await
             .map_err(|e| format!("Could not receive response: {e}"))?;
 
-        Ok(response)
+        response.map_err(|e| overwatch::DynError::from(format!("Mempool error: {e}")))
```
davidrusu: nit: just a thought, matches the error style above, feel free to ignore
```suggestion
        Ok(response.map_err(|e| format!("Mempool error: {e}"))?)
```

nomos-services/data-availability/dispersal/src/adapters/storage/mock.rs
```diff
@@ -15,8 +18,23 @@ impl DispersalStorageAdapter for MockDispersalStorageAdapter {
...
-        self.transactions.get(channel_id).cloned()
+    fn last_tx_id(&self, channel_id: &ChannelId) -> MsgId {
+        self.transactions.get(channel_id).map_or_else(
+            || {
+                // Inscription op id used in integration tests
+                [
+                    56, 175, 135, 67, 106, 190, 160, 137, 218, 122, 2, 114, 80, 39, 178, 39, 52,
+                    241, 73, 61, 40, 175, 174, 105, 112, 210, 30, 40, 139, 8, 90, 162,
+                ]
+                .into()
```
zeegomo, line 29: this feels like a hack, why do we need this hardcoded value here?

nomos-services/tx-service/src/backend/mod.rs
```diff
@@ -49,12 +49,9 @@ pub trait MemPool {
     /// Get multiple items by their keys from the mempool via storage lookup
     async fn get_items_by_keys(
         &self,
-        keys: BTreeSet<Self::Key>,
+        keys: &BTreeSet<Self::Key>,
```
zeegomo: could it be an `impl Iterator<Item=Self::Key>`?

tests/src/common/da.rs
```diff
@@ -62,6 +69,81 @@ pub async fn wait_for_blob_onchain(executor: &Executor, blob_id: BlobId) {
     );
 }
 
+/// Sets up a test channel by sending an inscription transaction and waiting for
+/// it to be included in a block Returns the channel ID that was created
```
zeegomo: missin \n?

tests/src/tests/da/api.rs
```diff
@@ -70,12 +75,14 @@ async fn test_get_commitments_from_peers() {
     lone_validator_config.membership = validator.config().membership.clone();
     let lone_validator = Validator::spawn(lone_validator_config).await.unwrap();
 
+    let test_channel_id = setup_test_channel(executor).await;
```
zeegomo: could you share what's the reasoning between always setting up a channel like this? It should work indeed but it's not needed, is this to test some specific functionality?

zeegomo (changes requested):

tests/src/tests/da/api.rs
andrussal: > could you share what's the reasoning between always setting up a channel like this? It should work indeed but it's not needed, is this to test some specific functionality?

Issue was that previously it assumed channel 0 from genesis which have `signer=Ed25519PublicKey_ZERO,`. It didn't seem possible to find corresponding private key to sign blob ops for channel 0 with this signer. Maybe I have some wrong assumptions here about how things can be set up. But it might be still good to do it this way to cover also channel testing.

nomos-services/data-availability/dispersal/src/adapters/storage/mock.rs
andrussal: > this feels like a hack, why do we need this hardcoded value here?

This is part of `mock`. Not sure how this should work outside mock tho. DA tracks channel ops to have previous message id. But not inscriptions which will be parent of first blob op.

bacv: The api will need to change, user will need to provide the parent msg id. Lets leave this as is for now.

tests/src/tests/da/disperse.rs
andrussal: This is just a pattern used throughout the integration tests. Here it probably isn't needed at all. But overall it might be better to have something more deterministic, created this PR: https://github.com/logos-co/nomos/pull/1825

davidrusu (approved):
LGTM

nomos-services/tx-service/src/backend/pool.rs
```diff
@@ -60,7 +61,7 @@ impl<BlockId, Item, Key, Storage, RuntimeServiceId> MemPool
     for Mempool<BlockId, Item, Key, Storage, RuntimeServiceId>
 where
     Key: Hash + Eq + Ord + Clone + Send + Sync + 'static,
-    Item: Clone + Send + Sync + 'static + Serialize + for<'de> Deserialize<'de>,
+    Item: Transaction + Clone + Send + Sync + 'static + Serialize + for<'de> Deserialize<'de>,
```
danielSanchezQ: Why is there a bind to `Transaction` I dont see its used 🤔

nomos-services/tx-service/src/tx/service.rs
```diff
@@ -211,7 +213,7 @@ where
     Pool: MemPoolTrait<Storage = StorageAdapter> + RecoverableMempool + Send + Sync,
     StorageAdapter: MempoolStorageAdapter<RuntimeServiceId> + Clone + Send + Sync,
     <Pool as RecoverableMempool>::RecoveryState: Debug + Send + Sync,
-    Pool::Item: Clone + Send + 'static,
+    Pool::Item: Transaction<Hash = Pool::Key> + Clone + Send + 'static,
```
danielSanchezQ, line 216: At this level we already know that the pool item is `Transaction`, so the bind to transaction should be in this service instead of the pool.

tests/src/tests/da/api.rs
zeegomo: is this just for mock da wallet?

andrussal: > is this just for mock da wallet?

I am not completely sure if we can avoid doing this for real da wallet adapter... My understanding was that we need to set up channel first. 

(And we can't use genesis channel because we don't have its private key, but this point is maybe not relevant at all. I just thought maybe it was an assumption that we can use it).

PR #1802, opened 2025-10-13
feat(mantle_encoding): Mantle tx encoder/decoder

davidrusu:
## 1. What does this PR implement?

> [!NOTE]  
> Review the 
> [Specification to Implementation Process document](https://www.notion.so/Specification-to-Implementation-Process-0cd6547233fa43a1a03d6e360de9eb2b).

## 2. Does the code have enough context to be clearly understood?

> [!IMPORTANT]  
> Add a link to the specification document, pointing at the specific section.  
> Any implementation detail that could be considered idiosyncratic and not direct from the specification, should be  
> explained.  
> Always consider what someone new would need to know for the code to make sense, and add this context to it.

## 3. Who are the specification authors and who is accountable for this PR?

> [!IMPORTANT]  
> Ping the original specification author(s) and add them as reviewers, alongside any other from the engineering team.  
> You are accountable for the PR and its rapid resolution, so make sure anyone involved is aware and actively
> advancing  
> the review.

## 4. Is the specification accurate and complete?

> [!IMPORTANT]
> * If not, engage in a conversation directly with the authors, prompting any updates in the document.
> * You are accountable of this process, and the specification author needs to give priority to unblocking you.

## 5. Does the implementation introduce changes in the specification?

> [!IMPORTANT]
> * Contact the specification author(s) and engage in the necessary conversation to agree on the proposed  
    modification/solutions.
> * Ensure the new changes are on par with the implementation at the end of the process.
> * The PR should not be merged until parity between implementation and specifications has been reached.

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [ ] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [ ] 2. Description added.
* [ ] 3. Context and links to Specification document(s) added.
* [ ] 4. Main contact(s) (developers and specification authors) added
* [ ] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [ ] 6. Link PR to a specific milestone.

PR #1806, opened 2025-10-15
feat(txs): Tx encoding update v1.1 spec

danielSanchezQ:

It adds the encoding update for txs from specification. It doesn't fully implement it as we are still using dummy proofs. Those will be finalize later during integration. This supersedes #1802 from @davidrusu and it is just a cherry-pick of his changes so it doesnt disrupt too much. Just adds the functionality without being integrated.
Next steps are:
* Use of proper proofs
* Modify the wire encode/decode so we use a custom trait
* Integrate this encoding to be used as part of the updated wire format

Its a lots of methods but they are simple.

@davidrusu @zeegomo @danielSanchezQ 

Yes

No, it just not fully implements it (DummySignature is still in place).

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-core/chain-defs/src/mantle/encoding.rs
+++ b/nomos-core/chain-defs/src/mantle/encoding.rs
@@ -194,11 +194,15 @@ fn decode_sdp_declare(input: &[u8]) -> IResult<&[u8], SDPDeclareOp> {
     ))
 }
 
+const LOCATOR_BYTES_SIZE_LIMIT: usize = 329usize;
+
 fn decode_locator(input: &[u8]) -> IResult<&[u8], multiaddr::Multiaddr> {
     // Locator = 2Byte *BYTE
     let (input, len_bytes) = take(2usize).parse(input)?;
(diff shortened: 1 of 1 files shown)


nomos-core/chain-defs/src/mantle/encoding.rs
```diff
@@ -0,0 +1,1201 @@
...
+            provider_id,
+            zk_id,
+            locked_note_id,
+        },
+    ))
+}
+
+fn decode_locator(input: &[u8]) -> IResult<&[u8], multiaddr::Multiaddr> {
+    // Locator = 2Byte *BYTE
+    let (input, len_bytes) = take(2usize).parse(input)?;
```
davidrusu, line 201: Spec says locators should be no longer than 329 bytes

```
Locator      = 2Byte *BYTE ; Max 329 bytes, multiaddr format
```

Shall we reject the parse right here if they are longer tha 329 bytes?

danielSanchezQ: I think so, good catch.

PR #1807, opened 2025-10-15
feat(sdp): declaration, activity, widthdrawn service messages

pradovic:

This PR adds a basic structure of the declaration, withdrawn and activity message posting in sdp service. It also covers a basic usage pattern in DA.

Yes.

@pradovic @bacv 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-core/chain-defs/src/sdp/mod.rs
+++ b/nomos-core/chain-defs/src/sdp/mod.rs
@@ -617,9 +617,8 @@ mod tests {
 
     #[test]
     fn test_activity_metadata_empty_bytes() {
-        let result = ActivityMetadata::from_metadata_bytes(&[]);
-        assert!(result.is_err());
-        assert!(result.unwrap_err().to_string().contains("Empty metadata"));
+        let result = ActivityMetadata::from_metadata_bytes(&[]).expect("should be OK");
+        assert!(result.is_none());
     }
 
     #[test]


youngjoon-lee (approved):
looks good.

nomos-services/data-availability/network/src/sdp/adapters/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+    #[error(transparent)]
+    Other(#[from] DynError),
+}
+
+#[async_trait]
+pub trait SdpAdapter {
+    type SdpService: ServiceData;
+
+    fn new(outbound_relay: OutboundRelay<<Self::SdpService as ServiceData>::Message>) -> Self;
+    async fn post_activity(&self, metadata: Vec<u8>) -> Result<(), SdpAdapterError>;
```
danielSanchezQ: What is the metadata used for? 🤔

nomos-services/data-availability/network/src/opinion_aggregator.rs
```diff
@@ -28,12 +28,27 @@ pub enum OpinionError {
...
+    )]
+    pub fn into_metadata_bytes(self) -> Vec<u8> {
+        let mut bytes = vec![0x01]; // Version byte from spec
+        bytes.extend(&self.current_session.to_le_bytes());
+        let prev_bytes = bitvec_to_bytes(self.previous_session_opinions);
+        bytes.extend(prev_bytes);
+        let curr_bytes = bitvec_to_bytes(self.current_session_opinions);
+        bytes.extend(curr_bytes);
+        bytes
+    }
```
danielSanchezQ: Is this up per spec?

nomos-services/data-availability/network/src/sdp/adapters/mod.rs
danielSanchezQ: Ok, I see it is the activity proof. I think this should be deal inside this adapter. It should take the activity proof and the transformation to bytes should be handled inside.

danielSanchezQ (commented):
Thanks for keeping it short!
I have to check properly how we are encoding the bits to bytes yet.

nomos-services/data-availability/network/src/opinion_aggregator.rs
```diff
@@ -359,11 +374,27 @@ where
...
+    if let Some(last) = bytes.last_mut() {
+        let used_bits = (nbits % 8) as u8;
+        if used_bits != 0 {
+            let mask = (1u8 << used_bits) - 1;
+            *last &= mask;
+        }
+    }
+
+    bytes
+}
```
danielSanchezQ: I think we should use [`as_raw_slice`](https://docs.rs/bitvec/latest/bitvec/vec/struct.BitVec.html#method.as_raw_slice), trim the unused zeros if we want so then later can be loaded again.

pradovic: Great, that is actually much better as there is no clone!

nomos-services/data-availability/network/src/opinion_aggregator.rs
pradovic: I believe it is:
```
The `active_message` `metadata` field must be populated with:

- The `header` that contains a one-byte `version` field which is fixed to `0x01` value.
- The `activity_proof` as defined above in [[Activity Proof](https://www.notion.so/Activity-Proof-203261aa09df806d9ff3f5e3bbe359b1?pvs=21)](https://www.notion.so/Activity-Proof-203261aa09df806d9ff3f5e3bbe359b1?pvs=21).
```

pradovic: And the activity proof size is based on the number of elements, while the empty bitvec is an empty byte. Does that makes sense?

bacv: This might later be changed to an enum in `nomos-core`, I think it'll be more ergonomic for use in the ledger.

nomos-services/data-availability/network/src/opinion_aggregator.rs
```diff
@@ -379,9 +375,11 @@ where
...
 
-fn bitvec_to_bytes(bv: BitVec<u8, Lsb0>) -> Vec<u8> {
+fn bitvec_to_bytes(bv: &BitVec<u8, Lsb0>) -> Vec<u8> {
     let nbits = bv.len();
-    let mut bytes = bv.into_vec();
+    let raw_bytes = bv.as_raw_slice();
+    let needed_bytes = nbits.div_ceil(8);
+    let mut bytes = raw_bytes[..needed_bytes].to_vec();
 
     // Zero out unused padding bits in the last byte (per spec)
```
danielSanchezQ: Doesn't bitvec does this already? unused bits are zero by default :thinking:

nomos-services/data-availability/network/src/opinion_aggregator.rs
danielSanchezQ: Nitpick, you can pre-allocate the vec size so it doesn't resize.

pradovic: I agree! I wanted to keep it simple for this, and I can extract it if needed as well.

nomos-services/data-availability/network/src/opinion_aggregator.rs
pradovic: Hmm, let me test it again

pradovic: You are correct, I have added a few tests for this as well.

nomos-services/data-availability/network/src/opinion_aggregator.rs
```diff
@@ -377,20 +377,8 @@ where
...
-    if let Some(last) = bytes.last_mut() {
-        let used_bits = (nbits % 8) as u8;
-        if used_bits != 0 {
-            let mask = (1u8 << used_bits) - 1;
-            *last &= mask;
-        }
-    }
-
-    bytes
+    bv.as_raw_slice()[..needed_bytes].to_vec()
```
danielSanchezQ: Also no need to use needed_bytes, bitvec should trim to the used bytes only. Can you double check?

danielSanchezQ:
```suggestion
fn bitvec_to_bytes(bv: &BitVec<u8, Lsb0>) -> Vec<u8> {
    bv.as_raw_slice().to_vec()
}
```
Don't commit the suggestion, it is not well place, but that is the idea.

pradovic: Ah actually you are right again, it does everything for us:
```
pub fn as_raw_slice(&self) -> &[T] {
    let (data, len) = (self.bitspan.address().to_const(), self.bitspan.elements());
    unsafe { slice::from_raw_parts(data, len) }
}
```

pradovic: Fixed.

nomos-services/data-availability/network/src/opinion_aggregator.rs
```diff
@@ -28,12 +28,31 @@ pub enum OpinionError {
...
-    pub old_opinions: BitVec,
+pub struct ActivityProof {
+    pub current_session: SessionNumber,
+    pub previous_session_opinions: BitVec<u8, Lsb0>,
+    pub current_session_opinions: BitVec<u8, Lsb0>,
 }
 
+impl ActivityProof {
+    pub fn to_metadata_bytes(&self) -> Vec<u8> {
+        let prev_bytes_len = self.previous_session_opinions.len().div_ceil(8);
```
ntn-x2: Could this be a method on its own?

nomos-services/data-availability/network/src/opinion_aggregator.rs
```diff
@@ -28,12 +28,31 @@ pub enum OpinionError {
...
+        let curr_bytes_len = self.current_session_opinions.len().div_ceil(8);
+        let total_size = 1 // version byte
+            + size_of::<SessionNumber>()
+            + prev_bytes_len
+            + curr_bytes_len;
+
+        let mut bytes = Vec::with_capacity(total_size);
+
+        bytes.push(0x01);
```
ntn-x2: Could this be a constant with a declarative name? Or link to a spec if it is defined in a spec.

nomos-services/data-availability/network/src/opinion_aggregator.rs
```diff
@@ -28,12 +28,31 @@ pub enum OpinionError {
     Error(#[from] DynError),
 }
 
-pub struct Opinions {
-    pub session_id: SessionNumber,
-    pub new_opinions: BitVec,
-    pub old_opinions: BitVec,
+pub struct ActivityProof {
+    pub current_session: SessionNumber,
+    pub previous_session_opinions: BitVec<u8, Lsb0>,
```
ntn-x2: I feel this might be a good case for a type alias?

nomos-services/data-availability/network/src/sdp/adapters/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+pub enum SdpAdapterError {
+    #[error(transparent)]
+    Other(#[from] DynError),
+}
+
+#[async_trait]
+pub trait SdpAdapter {
+    type SdpService: ServiceData;
+
+    fn new(outbound_relay: OutboundRelay<<Self::SdpService as ServiceData>::Message>) -> Self;
```
ntn-x2: Would it make more sense for the adapter to take a `OverwatchHandle` as input instead? Taking an `OutboundRelay` implies that entity interacting with this adapter already knows how to get an outbound relay with the service. I think, personal opinion, it would make a better design to get rid of the associated type altogether, and simply let the `new` function take an `OverwatchHandle`. Then the implementor would need to await for the service to be ready before creating an outbound channel. In this case, DA would need to open a channel with `SdpService` before passing the `outbound` relay to `new`, so I think it kinda mixes two different things. Anyway, this is my personal opinion and in now way blocking this PR.

nomos-services/sdp/src/lib.rs
```diff
@@ -21,57 +22,77 @@ const BROADCAST_CHANNEL_SIZE: usize = 128;
...
         result_sender: oneshot::Sender<FinalizedBlockUpdateStream>,
     },
+
+    PostDeclaration {
+        service_type: ServiceType,
+        locators: Vec<Locator>,
+        reply_channel: oneshot::Sender<Result<DeclarationId, DynError>>,
+    },
+    PostActivity {
+        metadata: Vec<u8>, // DA/Blend specific metadata
```
ntn-x2: Could we use a strong type instead of a `Vec<u8>`?

nomos-services/sdp/src/lib.rs
```diff
@@ -21,57 +22,77 @@ const BROADCAST_CHANNEL_SIZE: usize = 128;
...
     backend: PhantomData<Backend>,
     service_resources_handle: OpaqueServiceResourcesHandle<Self, RuntimeServiceId>,
     finalized_update_tx: broadcast::Sender<FinalizedBlockEvent>,
+    _current_declaration_id: Option<DeclarationId>,
 }
 
-impl<Backend, Metadata, RuntimeServiceId> ServiceData
-    for SdpService<Backend, Metadata, RuntimeServiceId>
+impl<Backend, RuntimeServiceId> ServiceData for SdpService<Backend, RuntimeServiceId>
 where
```
ntn-x2: I think this bound is not necessary at all, and we could then also remove it from hte `SdpService` struct declaration.

nomos-services/data-availability/network/src/sdp/adapters/mod.rs
pradovic: Is this what you had in mind:https://github.com/logos-co/nomos/pull/1807/commits/7d78f1039cf9830434c294d4c58c9b546b33b1b0?

ntn-x2: Yes. In my opinion this is a better approach.

nomos-services/sdp/src/lib.rs
pradovic: We could do something like this:
```
#[derive(Clone, Debug)]
pub struct ActivityMetadata(Vec<u8>);

impl ActivityMetadata {
    pub fn new(bytes: Vec<u8>) -> Self {
        Self(bytes)
    }
    
    pub fn into_bytes(self) -> Vec<u8> {
        self.0
    }
    
    pub fn as_bytes(&self) -> &[u8] {
        &self.0
    }
}

pub enum SdpMessage {
    PostActivity {
        metadata: ActivityMetadata,  
    },
    // ...
}
```

Or go for the trait, but I think this is probably going to end up an enum nomos-core (with Blend and DA version). cc @bacv 

The problem is the Blend and DA will have it's own variant.

nomos-services/data-availability/network/src/sdp/adapters/mod.rs
pradovic: Ok, looks good to me, I will see if I have any problems when I connect it, but looks good

nomos-services/sdp/src/lib.rs
pradovic: Which bound do you mean, the Backend?

pradovic: https://github.com/logos-co/nomos/pull/1807/commits/007f1a2e4c4017a077a4305b0246c8e4e688b504

ntn-x2: Nice!

nomos-services/sdp/src/lib.rs
ntn-x2: Why can it not be an enum as well?

pradovic: So maybe I can move this to nomos-core and do something like:
```

#[derive(Clone, Debug)]
pub enum ActivityMetadata {
    DataAvailability(DAActivityProof),
    Blend(BlendActivityProof),
}

pub struct DAActivityProof {
    pub current_session: SessionNumber,
    pub previous_session_opinions: Vec<u8>,
    pub current_session_opinions: Vec<u8>,
}

pub struct BlendActivityProof {
    // This is where Blend activity proof would go
}
```

And also they would both have to metadata bytes functions. Does that makes sense?

ntn-x2: If `DAActivityProof ` and `BlendActivityProof` have the same structure, then the `ActivityMetadata` variant is sufficient to distinguish between them. If they have a different structure, then yes we need a `struct` for each. Otherwise this would also work:

```
#[derive(Clone, Debug)]
pub enum ActivityMetadata {
    DataAvailability(ActivityProof),
    Blend(ActivityProof),
}

pub struct ActivityProof {
    pub current_session: SessionNumber,
    pub previous_session_opinions: Opinions,
    pub current_session_opinions: Opinions,
}
```

or you can switch things around and have

```
pub struct ActivityProof {
    pub current_session: SessionNumber,
    pub previous_session_opinions: Opinions,
    pub current_session_opinions: Opinions,
    pub service: ServiceType
}

pub enum ServiceType {
    DataAvailability,
    Blend
}
```

I would prefer the first one for a more clear separation between the two, but the choice depends on the usage. Pick your favorite, I guess!

pradovic: I think that spec does not impose a structure other then the header:

```
The `active_message` `metadata` field must be populated with:

The construction of the active message is as follows:

class ActiveMessage:
    declaration_id: DeclarationId
    nonce: Nonce
    metadata: Metadata

where `metadata` is a service-specific node activeness metadata.
```

Blend and DA have different structure and we need to support it, so I would go with the first variant + structs. The only thing I don't like is that it takes service specific type in nomos-core, with implementation (for example DA uses bitvecs underneath), that is why I opted for Vec<u8> initially.

@bacv Should we move this to nomos-core in this PR? Is this the structure you had in mind?

bacv: > Should we move this to nomos-core in this PR? Is this the structure you had in mind?

Yes, we should go with enums, if possible, lets add them in this PR.

pradovic: https://github.com/logos-co/nomos/pull/1814

nomos-core/chain-defs/src/mantle/encoding.rs
```diff
@@ -237,21 +240,30 @@ fn decode_sdp_active(input: &[u8]) -> IResult<&[u8], SDPActiveOp> {
...
     let declaration_id = DeclarationId(declaration_id_bytes);
+
     let (input, nonce) = decode_uint64(input)?;
+
     let (input, metadata_len) = decode_uint32(input)?;
-    let (input, metadata_vec) =
-        map(take(metadata_len as usize), |b: &[u8]| b.to_vec()).parse(input)?;
+    let (input, metadata_bytes) = take(metadata_len as usize)(input)?;
+
+    let metadata = if metadata_bytes.is_empty() {
```
danielSanchezQ: Is this even possible? i think in this case is confirmed that the bytes arrived, or the `take` call would fail.

nomos-core/chain-defs/src/mantle/encoding.rs
```diff
@@ -237,21 +240,30 @@ fn decode_sdp_active(input: &[u8]) -> IResult<&[u8], SDPActiveOp> {
     // Metadata = UINT32 *BYTE
     let (input, declaration_id_bytes) = decode_hash32(input)?;
     let declaration_id = DeclarationId(declaration_id_bytes);
+
     let (input, nonce) = decode_uint64(input)?;
+
     let (input, metadata_len) = decode_uint32(input)?;
-    let (input, metadata_vec) =
-        map(take(metadata_len as usize), |b: &[u8]| b.to_vec()).parse(input)?;
+    let (input, metadata_bytes) = take(metadata_len as usize)(input)?;
```
danielSanchezQ: We updated this to a new version, you should probably rebase.
```suggestion
    let (input, metadata_bytes) = take(metadata_len as usize).parse(input)?;
```

nomos-core/chain-defs/src/mantle/encoding.rs
```diff
@@ -237,21 +240,30 @@ fn decode_sdp_active(input: &[u8]) -> IResult<&[u8], SDPActiveOp> {
...
+
+    let metadata = if metadata_bytes.is_empty() {
+        None
+    } else {
+        // Parse raw bytes into ActivityMetadata
+        match ActivityMetadata::from_metadata_bytes(metadata_bytes) {
+            Ok(meta) => Some(meta),
+            Err(_) => {
+                return Err(nom::Err::Error(Error::new(input, ErrorKind::Fail)));
+            }
```
danielSanchezQ: No need for the if/else, just this should be enough, also better to:
```suggestion
        let metadata = ActivityMetadata::from_metadata_bytes(metadata_bytes).map_err(|_| nom::Err::Error(Error::new(input, ErrorKind::Fail)))?;
```

nomos-core/chain-defs/src/sdp/mod.rs
```diff
@@ -249,22 +249,174 @@ impl WithdrawMessage {
...
+    pub previous_session_opinions: Vec<u8>,
+    pub current_session_opinions: Vec<u8>,
+}
+
+impl DaActivityProof {
+    #[must_use]
+    pub fn to_metadata_bytes(&self) -> Vec<u8> {
+        let total_size = 1 // version byte
+            + size_of::<SessionNumber>()
+            + 4 // previous_session_opinions_length (u32)
```
danielSanchezQ: We should have a type alias for the lenght we use in encodings, then use `size_of` here. This hardcoded values are hard to refactor if needed.

nomos-core/chain-defs/src/sdp/mod.rs
```diff
@@ -249,22 +249,174 @@ impl WithdrawMessage {
...
+        // Encode current opinions with length prefix
+        bytes.extend(&(self.current_session_opinions.len() as u32).to_le_bytes());
+        bytes.extend(&self.current_session_opinions);
+
+        bytes
+    }
+
+    pub fn from_metadata_bytes(bytes: &[u8]) -> Result<Self, Box<dyn std::error::Error>> {
+        // Minimum size: version(1) + session(8) + prev_len(4) + curr_len(4) = 17
+        if bytes.len() < 17 {
```
danielSanchezQ: `17` should be a constant instead.

nomos-core/chain-defs/src/sdp/mod.rs
```diff
@@ -249,22 +249,174 @@ impl WithdrawMessage {
...
+        // Read current session opinions data
+        if offset + curr_len != bytes.len() {
+            return Err(format!(
+                "Current opinions length mismatch: expected {}, got {} remaining bytes",
+                curr_len,
+                bytes.len() - offset
+            )
+            .into());
+        }
+        let current_session_opinions = bytes[offset..offset + curr_len].to_vec();
```
danielSanchezQ: It would be nice to use `nom` instead of raw decoding this. Each of the reads should be a single method. This block is quite hard to read.

danielSanchezQ (changes requested):

nomos-core/chain-defs/src/mantle/encoding.rs
pradovic: Metadata is defined in sdp message as Option<Metadata> so I guess the check is to get the None option. I will check the len instead.
```
fn decode_sdp_active(input: &[u8]) -> IResult<&[u8], SDPActiveOp> {
    // SDPActive = DeclarationId Nonce Metadata
    // Metadata = UINT32 *BYTE
    let (input, declaration_id_bytes) = decode_hash32(input)?;
    let declaration_id = DeclarationId(declaration_id_bytes);

    let (input, nonce) = decode_uint64(input)?;

    let (input, metadata_len) = decode_uint32(input)?;
    let (input, metadata_bytes) = take(metadata_len as usize).parse(input)?;

    let metadata = if metadata_len == 0 {
        None
    } else {
        Some(
            ActivityMetadata::from_metadata_bytes(metadata_bytes)
                .map_err(|_| nom::Err::Error(Error::new(input, ErrorKind::Fail)))?,
        )
    };

    Ok((
        input,
        SDPActiveOp {
            declaration_id,
            nonce,
            metadata,
        },
    ))
}
```

danielSanchezQ: ok, `if metadata_len.is_zero()` better

danielSanchezQ: Actually, should this be abstracted in the `from_metadata_bytes`? It may return `Result<Option<ActivityMetadata>>` and return a None on an empty set of bytes in the first line of the method.

nomos-core/chain-defs/src/sdp/mod.rs
pradovic: Let's see if this is better now: https://github.com/logos-co/nomos/pull/1807/commits/86f2e340c036f2e68e78ab061ba950226e5942ba. Other comments should be addressed as well in this commit + I have added some simple tests

nomos-core/chain-defs/src/mantle/encoding.rs
pradovic: Yes, I think it is better, good catch! https://github.com/logos-co/nomos/pull/1807/commits/86f2e340c036f2e68e78ab061ba950226e5942ba

PR #1809, opened 2025-10-16
chore(blend): use const-size arrays for PoQ inputs

ntn-x2:

The height of all Merkle trees (and hence Merkle paths) is fixed for PoQ stuff. Having the size of these proofs known at compile time allows the compiler to yields more useful warnings, e.g. that the inputs now are quite large, hence it's better to box them. On purpose I only touched PoQ stuff, hence the need in a couple of places for some gimmicks to convert from PoL inputs, which I did not want to touch, for now. This PR is needed to provide better ergonomics when dealing with merkle paths in Blend.

For JSON representation of the inputs, I was able to switch from vecs to arrays, but was not able to "merge" paths and selectors together, since I guess the binary expects them as separate fields.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nodes/nomos-node/node/src/generic_services/blend.rs
+++ b/nodes/nomos-node/node/src/generic_services/blend.rs
@@ -278,7 +278,7 @@ where
                 ready(Some(PolEpochInfo {
                     epoch,
                     poq_private_inputs: ProofOfLeadershipQuotaInputs {
-                        aged_path_and_selectors: aged_path_and_selectors.try_into().expect("List of aged not paths and selectors does not match the expected size for PoQ inputs, although it has already been pre-processed."),
+                        aged_path_and_selectors: aged_path_and_selectors.try_into().expect("List of aged note paths and selectors does not match the expected size for PoQ inputs, although it has already been pre-processed."),
                         note_value: *note_value,
                         output_number: *output_number,
                         pol_secret_key: *secret_key.as_fr(),


zk/proofs/poq/src/wallet_inputs.rs
```diff
@@ -38,40 +41,46 @@ pub struct PoQWalletInputsJson {
...
             pol_sk_starting_slot,
             pol_note_value,
-        }: &PoQWalletInputs,
+        }: PoQWalletInputs,
     ) -> Self {
+        let (aged_path, aged_selector) = {
+            let aged_path = pol_noteid_path_and_selectors.map(|(path, _)| (&path).into());
+            let aged_selector =
+                pol_noteid_path_and_selectors.map(|(_, selector)| (&selector).into());
+            (aged_path, aged_selector)
```
danielSanchezQ, line 71: You can use [`unzip`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.unzip) for this.

ntn-x2: `unzip` does not work in this case, because it's only implemented for `Vec`s and slices. `map` in this case is a const operation, which is actually very nice. It's guaranteed to never fail, because it's a method on array, and not the one on iterators.

danielSanchezQ: Aaah, ok, because now it is an array so it just created another array 👍

nodes/nomos-node/node/src/generic_services/blend.rs
```diff
@@ -251,17 +254,37 @@ where
...
+                    vec_from_inputs.resize(SLOT_SECRET_MERKLE_TREE_HEIGHT, ZkHash::ZERO);
+                    vec_from_inputs
+                };
+
                 ready(Some(PolEpochInfo {
                     epoch,
                     poq_private_inputs: ProofOfLeadershipQuotaInputs {
-                        aged_path: aged_path.clone(),
-                        aged_selector: aged_selector.clone(),
+                        aged_path_and_selectors: aged_path_and_selectors.try_into().expect("List of aged not paths and selectors does not match the expected size for PoQ inputs, although it has already been pre-processed."),
```
thomaslavaur: little typo here

thomaslavaur: of aged notE

thomaslavaur (commented):
lgtm.

PR #1812, opened 2025-10-16
chore(ledger): sdp refactor

zeegomo:
We recently updated the SDP spec to include a section on sessions and how they're updated, which is also used for rewarding. 
Most importantly, session information is not fetched from the LIB anymore but rather from the tip, which contains a snapshot of the stabilized session (similarly to how they carry a snapshot of the epoch state).
I've reflected this change in the sdp ledger while at the same time fixing a few bugs and refactoring to make it fit better in the current ledger architecture (e.g. sessions are updated at the beginning of a block like epochs).

I've also removed most of the logic code from nomos-core/sdp and only left core structs and definitions.
We will need to change services to pull session information from the tip instead of the LIB.

(Don't be too scared by the diffs, 1/3 of it it logic, 600+ lines are just for added tests, and the rest is renames)

Yes

@madxor @zeegomo 
https://www.notion.so/nomos-tech/Service-Declaration-Protocol-1fd261aa09df819ca9f8eb2bdfd4ec1d#1fd261aa09df81c281eae509b3b07a14

Yes

* We don't do garbage collection yet
* We remove a declaration as soon as we receive a withdrawal, unclear whether we will need this information at all

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/ledger/nomos-ledger/src/mantle/sdp/mod.rs
+++ b/ledger/nomos-ledger/src/mantle/sdp/mod.rs
@@ -697,9 +697,9 @@ mod tests {
         }
 
         // Check DataAvailability is promoted to session 1
-        let active_session_b = sdp_ledger.get_active_session(service_a).unwrap();
+        let active_session_b = sdp_ledger.get_active_session(service_b).unwrap();
         assert_eq!(active_session_b.session_n, 1);
-        let forming_session_b = sdp_ledger.get_forming_session(service_a).unwrap();
+        let forming_session_b = sdp_ledger.get_forming_session(service_b).unwrap();
         assert_eq!(forming_session_b.session_n, 2);
 
         // Check BlendNetwork is still in session 0


bacv (approved):
Nice refactor! So for now the ledger initialization where sdp is required should use `StartingState::Lib`, and then we'll migrate to the initialization from the genesis tx?

ledger/nomos-ledger/src/mantle/mod.rs
```diff
@@ -97,46 +80,39 @@ impl LedgerState {
...
+    ) -> Result<Self, Error> {
         self.leaders = self.leaders.try_apply_header(epoch, voucher)?;
+        self.sdp = self.sdp.try_apply_header(&config.sdp_config)?;
         Ok(self)
     }
 
     fn try_apply_ops<'a>(
         mut self,
-        current_block_number: BlockNumber,
+        _current_block_number: BlockNumber,
```
bacv, line 115: Nit: this can be removed.

zeegomo: I'm out of office right now and can't change it, will fix in the next PR

youngjoon-lee (commented):
I'm late for review. but the new logic looks correct and good. Thanks.

PR #1813, opened 2025-10-17
feat(blend): activity proof computation

youngjoon-lee:

This is the first PR for the Blend Rewarding.
This PR implements:
- The types for [BlendingToken](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81b6ab5eea4e715193f1) and [ActivityProof](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df8150a535de45dd8dfb68).
- The data structure to collect blending tokens during a session, and to keep blending tokens from the previous session until the transition period has passed.
- The algorithm to compute an activity proof.
  - For details of the algorithm, please see the following specs:
    - [Hamming distance](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df818fbcb7d9decb5387d7)
    - [Activity Threshold](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df8124bd88cb58d2171ccf)
    - [Activity Message](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81128741c70db11f16e1)

This PR doesn't integrate the implementations to the Blend service. I will work on it as a next PR.

As you can see, we need many math operations with `f64` to implement this spec. So, I added more types and math operations into the `nomos-utils` to handle edge cases safely. 
Also, you may find that computing activity proof is fallible, if `total_core_quote: u64` or `num_core_nodes: u64` is too large, because not all `u64` values can be converted to `f64` without precision loss. To make it explicitly, my implementation returns an error in the case. Later, we can consider changed them to `u32`, which can be safely represented in `f64`, in our entire code base. Anyway, this is not a major issue for now.

I believe so.

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.


nomos-blend/message/Cargo.toml
```diff
@@ -15,16 +15,18 @@ groth16         = { workspace = true }
...
-serde-big-array = { version = "0.5", default-features = false }
+serde-big-array = { default-features = false, version = "0.5" }
 serde_with      = { default-features = false, features = ["alloc"], version = "3" }
 thiserror       = "1"
+tracing         = { workspace = true }
 x25519-dalek    = { default-features = false, features = ["serde", "static_secrets"], version = "2" }
 
 [dev-dependencies]
 const-hex = { default-features = false, version = "1.15.0" }
+test-log  = "0.2"
```
ntn-x2:
```suggestion
test-log  = {default-features = false, features = ["trace"], version = "0.2"}
```

nomos-blend/message/src/reward.rs
```diff
@@ -0,0 +1,521 @@
...
+    digest::{Update as _, VariableOutput as _},
+};
+use nomos_core::{block::SessionNumber, codec::SerializeOp as _};
+use nomos_utils::math::{F64Ge1, NonNegativeF64};
+use serde::Serialize;
+use tracing::{debug, error};
+
+use crate::crypto::proofs::{quota::ProofOfQuota, selection::ProofOfSelection};
+
+/// Blending tokens collected during the current session.
```
ntn-x2:
```suggestion
/// Collection of blending tokens that are relevant for the current session.
```

What do you think of this phrasing?

nomos-utils/src/math.rs
```diff
@@ -2,17 +2,80 @@ use core::ops::{Deref, DerefMut};
...
     #[cfg_attr(feature = "serde", serde(deserialize_with = "serde::deserialize"))] f64,
 );
 
-impl NonNegativeF64 {
+/// A wrapper around [`f64`] that guarantees the value is neither infinite nor
+/// NaN.
+impl FiniteF64 {
+    /// The maximum [`u64`] value that can be represented exactly as an [`f64`]
+    /// without losing precision.
+    pub const MAX_REPRESENTABLE_U64: u64 = (1u64 << f64::MANTISSA_DIGITS) - 1;
```
ntn-x2: I guess no need for this to be `pub`?

nomos-blend/message/src/reward.rs
```diff
@@ -0,0 +1,521 @@
...
+        };
+        prev_session_tokens.activity_proof(self.current_session_randomness)
+    }
+
+    /// Switches to the new session, keeping the current session tokens
+    /// for the session transition.
+    ///
+    /// If there was a previous session that has not yet been consumed for
+    /// an activity proof, it is discarded.
+    pub fn rotate_session(&mut self, new_session_info: &SessionInfo) {
```
ntn-x2: Why not returning the previous session tokens, just in case we want to do something with it?

nomos-blend/message/src/reward.rs
```diff
@@ -0,0 +1,521 @@
...
+        self.current_session_randomness = new_session_info.session_randomness;
+    }
+}
+
+/// Arguments required to build [`BlendingTokens`].
+pub struct SessionInfo {
+    session_number: SessionNumber,
+    session_randomness: SessionRandomness,
+    num_core_nodes: u64,
+    total_core_quota: u64,
```
ntn-x2: No need for `total`, right? Or is this the expected total? Still, I think having a core quota and the number of core nodes is enough to expose a `const` function that calculates the total core quota.

```suggestion
    core_quota: u64,
```

nomos-blend/message/src/reward.rs
```diff
@@ -0,0 +1,521 @@
...
+use tracing::{debug, error};
+
+use crate::crypto::proofs::{quota::ProofOfQuota, selection::ProofOfSelection};
+
+/// Blending tokens collected during the current session.
+///
+/// This also holds the blending tokens collected from the previous session
+/// to produce an activity proof for that session after the session transition
+/// period has passed.
+pub struct BlendingTokens {
```
ntn-x2: Would it make sense to call this struct `UnclaimedBlendingTokens`? Since we have a lot of types with similar names, I feel this might be clearer that we are talking about tokens we have collected but not yet submitted.

nomos-blend/message/src/reward.rs
```diff
@@ -0,0 +1,521 @@
...
+
+/// Sensitivity parameter to control the lottery winning conditions.
+const ACTIVITY_THRESHOLD_SENSITIVITY_PARAM: u64 = 1;
+
+/// Computes the activity threshold, which is the expected maximum Hamming
+/// distance from any blending token in a session to the next session
+/// randomness.
+fn activity_threshold(num_core_nodes: u64, token_repr_bit_len: u64) -> Result<u64, Error> {
+    debug!(
+        "Calculating activity threshold: num_core_nodes={num_core_nodes}, token_repr_bit_len={token_repr_bit_len}"
```
ntn-x2: `LOG_TARGET`?

nomos-blend/message/src/reward.rs
```diff
@@ -0,0 +1,521 @@
...
+    ///
+    /// It returns `Ok(None)` if there was no blending token satisfying the
+    /// activity threshold calculated.
+    /// It returns `Err` if the computation fails.
+    fn activity_proof(
+        self,
+        next_session_randomness: SessionRandomness,
+    ) -> Result<Option<ActivityProof>, Error> {
+        // First, compute the activity threshold.
+        let activity_threshold = activity_threshold(
```
ntn-x2: I think it would make more sense to fail on `new` if the provided input is too large, rather than failing when we try to call `activity_proof`. So if we are able to instantiate `SessionBlendingTokens` with the current session values, we should be able to call `activity_proof`, since it does not depend on `next_session_randomness`.

nomos-blend/message/src/reward.rs
```diff
@@ -0,0 +1,521 @@
...
+            self.num_core_nodes,
+            BlendingToken::repr_bit_len(self.total_core_quota, self.message_frequency_per_round)?,
+        )?;
+        debug!(
+            "Activity threshold for session {}: {}",
```
ntn-x2: LOG_TARGET?

nomos-blend/message/src/reward.rs
```diff
@@ -0,0 +1,521 @@
...
+                        (distance <= activity_threshold).then_some((token, distance))
+                    }
+                    Err(e) => {
+                        error!("Failed to compute Hamming distance: {e}");
+                        None
+                    }
+                }
+            })
+            .min_by_key(|(_, distance)| *distance);
+        let Some((token, _)) = maybe_token else {
```
ntn-x2: If we fail on construction instead, this function could then finally just return an `Option` instead of a `Result<Option>`, which I think it's cleaner.

nomos-blend/message/src/reward.rs
```diff
@@ -0,0 +1,521 @@
...
+        let activity_threshold = activity_threshold(
+            self.num_core_nodes,
+            BlendingToken::repr_bit_len(self.total_core_quota, self.message_frequency_per_round)?,
+        )?;
+        debug!(
+            "Activity threshold for session {}: {}",
+            self.session_number, activity_threshold
+        );
+
+        // Filter tokens whose Hamming distance is below the activity threshold,\
```
ntn-x2: Intentional?

```suggestion
        // Filter tokens whose Hamming distance is below the activity threshold,
```

nomos-blend/message/src/reward.rs
```diff
@@ -0,0 +1,521 @@
...
+            .tokens
+            .into_iter()
+            .filter_map(|token| {
+                match token.hamming_distance(
+                    self.message_frequency_per_round,
+                    self.total_core_quota,
+                    next_session_randomness,
+                ) {
+                    Ok(distance) => {
+                        debug!("Token distance:{distance}, threshold: {activity_threshold}");
```
ntn-x2: LOG_TARGET

nomos-blend/message/src/reward.rs
```diff
@@ -0,0 +1,521 @@
...
+            "Activity threshold for session {}: {}",
+            self.session_number, activity_threshold
+        );
+
+        // Filter tokens whose Hamming distance is below the activity threshold,\
+        // and pick the one with the smallest distance.
+        let maybe_token = self
+            .tokens
+            .into_iter()
+            .filter_map(|token| {
```
ntn-x2: Can't we just take the smallest, and return it if it's less than `activity_threshold`? So basically without `filter_map`, but just calling `min_by_key`, and then return `Some` if it's smaller than `activity_threshold`.

nomos-utils/src/math.rs
youngjoon-lee: Fixed in aa3203c41dca3634bb9b8207ba6018f8f0071528

nomos-blend/message/Cargo.toml
youngjoon-lee: fixed in 1933e615b4ba13233210d3867aca447612cd27aa

nomos-blend/message/src/reward.rs
youngjoon-lee: fixed 0a0e09c13099bb0b2c1795ded6f9062c709644a6

nomos-blend/message/src/reward.rs
youngjoon-lee: fixed 0a0e09c13099bb0b2c1795ded6f9062c709644a6

nomos-blend/message/src/reward.rs
youngjoon-lee: fixed 0a0e09c13099bb0b2c1795ded6f9062c709644a6

nomos-blend/message/src/reward.rs
youngjoon-lee: I did it at the beginning but reverted since I'm not sure it will be useful. But, I agree to have it. 

Applied in 1499fda6150889dfd476d0e26a3852fed4c15f07

nomos-blend/message/src/reward.rs
youngjoon-lee: All blending tokens are always in the unclaimed state. Once we compute an activity proof, all the related blending tokens are dropped from memory. So, I prefer leaving this name shorter. 

Nevertheless, I understand that the names can be confused. In 1499fda6150889dfd476d0e26a3852fed4c15f07, I renamed `SessionBlendingTokens` to `BlendingTokensForSession` to clearly express that it's a collection of tokens for a single session. Here, the `BlendingTokens` means two collections of tokens (current & previous sessions). Maybe this name is too generic, but I couldn't find anything better yet.
Also, in the commit, I updated the doc comments. I believe it's cleaner now.

nomos-blend/message/src/reward.rs
youngjoon-lee: Yes, I agree it's better. In 1499fda6150889dfd476d0e26a3852fed4c15f07, I changed `SessionInfo` to try to compute all the ingredient beforehand. Now, all the subsequent operations are infallible.

nomos-blend/message/src/reward.rs
youngjoon-lee: Yes. Fixed as I commented in https://github.com/logos-co/nomos/pull/1813#discussion_r2443784507

nomos-blend/message/src/reward.rs
youngjoon-lee: Fixed. thanks. 1499fda6150889dfd476d0e26a3852fed4c15f07

nomos-blend/message/src/reward.rs
youngjoon-lee: Improved in 1499fda6150889dfd476d0e26a3852fed4c15f07. Thanks.

nomos-blend/message/src/reward.rs
youngjoon-lee: I updated this comment. Please see details in https://github.com/logos-co/nomos/pull/1813#discussion_r2443783460

nomos-blend/message/src/reward.rs
youngjoon-lee: I think it's fine to have this `total_core_quota` parameter, since this module doesn't need to know how to calculate `total_core_quota` from `core_quota` (though it's trivial in our brain). As I explained in https://github.com/logos-co/nomos/pull/1813#discussion_r2443784507, now, these two parameters (`total_core_quota` and `num_core_nodes`) are used only in `SessionInfo::new` and don't present anywhere in the code. So, I prefer having this parameter as it is. My opinion could be wrong If I’m missing the context of how the core quota is used outside this module.

nomos-blend/message/src/reward.rs
youngjoon-lee: Also, I split them into submodules fba4686627c2dc8459bc49cece064fee66a7cb77. I hope this is now more readable.

youngjoon-lee: I ended up renaming `BlendingTokens` to `BlendingTokenCollector`, which is more natural when it's used in upper layers. 92e4a411

nomos-blend/message/src/reward.rs
ntn-x2: Yes you are right. This component should not really know how to compute total core quota, I agree.

nomos-blend/message/src/reward.rs
ntn-x2: Very nice, thank you!

PR #1815, opened 2025-10-17
feat(chain): validate and remove invalid transactions

andrussal:

This PR fixes the issue where a single invalid transaction could block block production. When block construction encountered an invalid transaction, the whole attempt was aborted and the invalid transactions stayed in the mempool. This PR skips invalid transactions and removes these from mempool.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


danielSanchezQ:
Should we merge the target branch first? I would like to avoid having a heavy branch PR if possible. Are the changes in the target branch independant? Could they be merged first?

zeegomo (changes requested):
Big changes like this should probably pre-discussed before committing to some decision (maybe I just missed the discussion 🥲 ). 
In this particular case, it's not clear to me what benefits there are in splitting validation and execution of a transaction, and it was a design choice of the ledger to avoid this split (e.g. to make sure we never validate on a different state than the one we're applying to). 

Applying a transaction creates a new state anyway (cloning the ledger is reasonably cheap, we do this all the time in forks), so keeping the old state in case an error occurs is perfectly fine and should not incur in noticeable additional cost and we should be able to avoid adding all of this validation context which is a duplicate of the ledger logic.
Have you measured performance to say it's too costly and that this change significantly improves on that?

andrussal:
> Big changes like this should probably pre-discussed before committing to some decision (maybe I just missed the discussion 🥲 ). In this particular case, it's not clear to me what benefits there are in splitting validation and execution of a transaction, and it was a design choice of the ledger to avoid this split (e.g. to make sure we never validate on a different state than the one we're applying to).
> 
> Applying a transaction creates a new state anyway (cloning the ledger is reasonably cheap, we do this all the time in forks), so keeping the old state in case an error occurs is perfectly fine and should not incur in noticeable additional cost and we should be able to avoid adding all of this validation context which is a duplicate of the ledger logic. Have you measured performance to say it's too costly?

I run yes some calculations, numbers seemed large enough to justify avoiding cloning. This is estimate for a year.

Meaning Of Columns

Raw (GB): straightforward count of retained UTXO records × ~96 B each (tx hash, index, value, key) with no container overhead—i.e., the theoretical minimum if you packed entries end-to-end.

Realistic (GB): multiplies the raw byte total by ≈3× to reflect the actual in‑memory footprint once you include the persistent hash‑trie structure, Merkle nodes, allocation overhead, and Rust bookkeeping that UtxoTree keeps for each entry.

Churn | tx/block | Outputs/year (M) | Raw (GB) | Realistic (GB)
-- | -- | -- | -- | --
95% churn | 10 | 0.95 | 0.09 | 0.27
95% churn | 50 | 4.73 | 0.45 | 1.36
95% churn | 100 | 9.46 | 0.91 | 2.72
95% churn | 500 | 47.30 | 4.54 | 13.62
80% churn | 10 | 3.78 | 0.36 | 1.09
80% churn | 50 | 18.92 | 1.82 | 5.45
80% churn | 100 | 37.84 | 3.63 | 10.90
80% churn | 500 | 189.22 | 18.16 | 54.49
50% churn | 10 | 9.46 | 0.91 | 2.72
50% churn | 50 | 47.30 | 4.54 | 13.62
50% churn | 100 | 94.61 | 9.08 | 27.25
50% churn | 500 | 473.04 | 45.41 | 136.24
0% churn | 10 | 18.92 | 1.82 | 5.45
0% churn | 50 | 94.61 | 9.08 | 27.25
0% churn | 100 | 189.22 | 18.16 | 54.49
0% churn | 500 | 946.08 | 90.82 | 272.47

Only way to avoid cloning seems to run validation separately because all ledger functions take ownership of self.

zeegomo:
> > Big changes like this should probably pre-discussed before committing to some decision (maybe I just missed the discussion 🥲 ). In this particular case, it's not clear to me what benefits there are in splitting validation and execution of a transaction, and it was a design choice of the ledger to avoid this split (e.g. to make sure we never validate on a different state than the one we're applying to).
> > Applying a transaction creates a new state anyway (cloning the ledger is reasonably cheap, we do this all the time in forks), so keeping the old state in case an error occurs is perfectly fine and should not incur in noticeable additional cost and we should be able to avoid adding all of this validation context which is a duplicate of the ledger logic. Have you measured performance to say it's too costly?
> 
> I run yes some calculations, numbers seemed large enough to justify avoiding cloning. This is estimate for a year.
> 
> Meaning Of Columns
> 
> Raw (GB): straightforward count of retained UTXO records × ~96 B each (tx hash, index, value, key) with no container overhead—i.e., the theoretical minimum if you packed entries end-to-end.
> 
> Realistic (GB): multiplies the raw byte total by ≈3× to reflect the actual in‑memory footprint once you include the persistent hash‑trie structure, Merkle nodes, allocation overhead, and Rust bookkeeping that UtxoTree keeps for each entry.
> 
> Churn	tx/block	Outputs/year (M)	Raw (GB)	Realistic (GB)
> 95% churn	10	0.95	0.09	0.27
> 95% churn	50	4.73	0.45	1.36
> 95% churn	100	9.46	0.91	2.72
> 95% churn	500	47.30	4.54	13.62
> 80% churn	10	3.78	0.36	1.09
> 80% churn	50	18.92	1.82	5.45
> 80% churn	100	37.84	3.63	10.90
> 80% churn	500	189.22	18.16	54.49
> 50% churn	10	9.46	0.91	2.72
> 50% churn	50	47.30	4.54	13.62
> 50% churn	100	94.61	9.08	27.25
> 50% churn	500	473.04	45.41	136.24
> 0% churn	10	18.92	1.82	5.45
> 0% churn	50	94.61	9.08	27.25
> 0% churn	100	189.22	18.16	54.49
> 0% churn	500	946.08	90.82	272.47
> Only way to avoid cloning seems to run validation separately because all ledger functions take ownership of self.



What's churn in this context? The number of reused utxos? What does the addition of utxos have to do with the fact that we clone or not the ledger (which we do anyway, which is part of my  my point)? 
A clone of the ledger without changes should be close to a pointer clone, and 
```
if validate(ledger) {
  ledger = ledger.apply_unchecked(tx) // cloning here
} else {
  error!()
}
```
Should be pretty much equivalent to:
```
if let Ok(new_ledger) = ledger.apply(tx) {
  ledger = new_ledger
} else {
  error!()
}
```

zeegomo:
> If we are certain that cloning will be always cheap, would it be ok then to do something like this?
if let Ok(new_ledger) = ledger.clone().apply(tx) {
 ledger = new_ledger
} else {
 // handle error
}

yup, this would work, but can be simplified as I've written above, you don't need to reassign in the error case

> Yup, this PR has many changes but it took just couple of days to experiment. I didn't consider it final decision nor something like that:)

well, a couple days that could have been avoided, not too bad eh, just something for the next time

andrussal:
> ledger.clone().apply(tx)

Just one quick question. If we just clone pointers, would ledger.clone().apply(tx) still be atomic? No partial updates can happen?

nomos-services/chain/chain-leader/src/lib.rs
```diff
@@ -548,55 +557,104 @@ where
...
+            }
+        };
+
+        let filtered_stream = txs_stream.filter(move |tx| {
+            let is_valid = tx.mantle_tx().ops.iter().all(|op| match op {
+                Op::ChannelBlob(op) => blobs.contains(&op.blob),
+                _ => true,
+            });
+
+            async move { is_valid }
```
zeegomo: couldn't you put just one `async` block at the beginning like 
```
txs_stream.filter(async move |tx| {
            tx.mantle_tx().ops.iter().all(|op| match op {
                Op::ChannelBlob(op) => blobs.contains(&op.blob),
                _ => true,
})
```

tests/src/tests/cryptarchia/invalid_tx_handling.rs
```diff
@@ -0,0 +1,183 @@
...
+    nodes::validator::Validator,
+    topology::{Topology, TopologyConfig},
+};
+use tokio::time::timeout;
+
+const BATCH_SIZE: usize = 5;
+
+#[tokio::test]
+#[serial]
+async fn invalid_transactions_are_handled() {
```
zeegomo, line 27: I'd appreciate a little explanation for bigger tests on what they do and what is the expectation

Mantle_Specification.pdf
zeegomo, line 1: this probably shouldn't be here

tests/src/tests/cryptarchia/invalid_tx_handling.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    println!(
+        "All {total_valid} transactions posted. Valid: {}, Invalid: {}",
+        first_batch_hashes.len() + second_batch_hashes.len(),
+        invalid_tx_hashes.len()
+    );
+
+    timeout(
+        Duration::from_secs(120),
```
zeegomo: are we this slow?

zeegomo (approved):

PR #1817, opened 2025-10-20
feat(encoding): DummyZkSignature compatible w/ mantle encoding spec

davidrusu:

Our DummyZkSignature is now a hash commitment to the public inputs, the hash is 128 bytes long so that it exactly matches the Groth16 proof once it's implemented.

Verifying the dummy sig is just verifying the hash commitment matches the passed in public values:

`dummy_sig.verify(public_values) ===> hash(public_values)  == dummy_sig`

This enables us to have a correct encoding spec implementation for the signed mantle tx.

yes

@davidrusu 
https://www.notion.so/nomos-tech/Mantle-Transaction-Encoding-287261aa09df80c694b3f2539131b78e?source=copy_link

yes

no

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-core/chain-defs/src/mantle/encoding.rs
+++ b/nomos-core/chain-defs/src/mantle/encoding.rs
@@ -52,9 +52,13 @@ pub fn decode_signed_mantle_tx(input: &[u8]) -> IResult<&[u8], SignedMantleTx> {
     // In release mode without test/debug, we need to verify
     #[cfg(not(any(test, debug_assertions)))]
     {
-        SignedMantleTx::new(mantle_tx, ops_proofs, ledger_tx_proof)
-            .map(|tx| (input, tx))
-            .map_err(|_| nom::Err::Error(Error::new(input, ErrorKind::Verify)))
+        SignedMantleTx::new(
+            mantle_tx,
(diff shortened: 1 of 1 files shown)


nomos-core/chain-defs/src/utils/mod.rs
```diff
@@ -51,5 +51,56 @@ macro_rules! serde_bytes_newtype {
     };
 }
 
+macro_rules! serde_big_bytes_newtype {
```
davidrusu: serde only provides deser implementations for arrays up to `[T; 32]`. This macro provides implementations for any multiple of 32 (i.e. for groth16 proofs which are [u8;128])

implemented by deserializing a tuple of [u8;32]'s. i.e.

`<[[u8;32]; N]>::deserialize()` will work for N up to 32 (up to 1KB).

After deserialization, we flatten that into an `[u8; 32 * N]` array.

We do it this way to avoid having to write a visitor implementation for these larger arrays.

danielSanchezQ: Use generic array instead with the serde feature enabled. (Its what we use in other places).

danielSanchezQ (approved):

Issue #1785, opened 2025-10-03
Integrate secret ZK key into Blend service


ntn-x2:
Once the membership service will return ZK public keys of core nodes, use the private key to compute the Merkle path and selectors used as private inputs in PoQ. The key should be a config of the Blend service, for now. We will switch to the KMS  service later on.

Issue #1734, opened 2025-09-23
Integrate real private Blend data into Blend proofs


ntn-x2:
Calculate the required Merkle paths and selectors when generating proofs of core quota.

PR #1810, opened 2025-10-16
chore(blend): switch from membership service to chain broadcast for session subscription

ntn-x2:
## 1. What does this PR implement?

The changeset of this PR is also part of the overarching [epoch transition PR](https://github.com/logos-co/nomos/pull/1793). Nevertheless, that PR might take a while to be merged, and this is a small change, so I thought to open a separate PR for it that maintains the current behaviour and just changes the source of the session ticks. Proper processing of the new information (i.e., session numbers and zk keys) will still be done in the bigger PR linked above.

## 2. Does the code have enough context to be clearly understood?

Yes.

## 3. Who are the specification authors and who is accountable for this PR?

@ntn-x2 

## 4. Is the specification accurate and complete?

N/A

## 5. Does the implementation introduce changes in the specification?

No.

## Checklist

* [x] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [x] 2. Description added.
* [x] 3. Context and links to Specification document(s) added.
* [x] 4. Main contact(s) (developers and specification authors) added
* [x] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 6. Link PR to a specific milestone.

Issue #1784, opened 2025-10-03
Integrate zk keys from membership when supported


ntn-x2:
Membership at some point will return the list of public ZK keys of core nodes. This should be used to compute the merkle root that is used as a public input in PoQs.

ntn-x2:
Closed via https://github.com/logos-co/nomos/pull/1861.

PR #1821, opened 2025-10-21
feat(blend): integrate zk components into membership info

ntn-x2:

At first, I tried to squeeze all the ZK info inside the `Membership` type. Nevertheless, most of our Blend stack is interested in the membership only for the list of peer IDs, addresses, and in some parts their public key. Hence, I reworked this PR to introduce ZK info only at the service level, which is the only place where we consume it. This means no changes in the underlying Blend stack crates, and I think it's a much cleaner solution.

So, this PR introduces a Blend-specific ZK merkle tree that sorts the list of zk public keys of core nodes for each session, creates a root and, if the node is a core node, generates the path of its own public key to the merkle root. This is created in the membership adapter once per session.

I added tests for the new merkle tree, which uses `rs_merkle_tree` under the hood, but we don't have any reference values to make sure the tree actually returns the same root that the PoQ circuit computes, but we will know whether that is an issue once we integrate the actual proof verifier.

As for integration tests, since we require that a core node's public zk key is included in the membership info, we cannot just mock public keys as they would not match. Hence, I temporarily introduced a way to deterministically compute zk public keys from node's public keys, so we don't break integration tests. Once we switch to the chain broadcast service, this issue will go away since we will get actual zk public keys from it and we won't need to mock anything anymore.

Closes https://github.com/logos-co/nomos/issues/1785 and closes https://github.com/logos-co/nomos/issues/1734.

Yes.

@ntn-x2 - dev, @youngjoon-lee - reviewer, @madxor - Blend spec author, @thomaslavaur ZK and PoQ expert 

Yes.

We introduced some details clarifying how the merkle tree is generated from the input list of zk keys.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/merkle.rs
+++ b/nomos-services/blend/src/merkle.rs
@@ -76,8 +76,9 @@ impl MerkleTree {
     /// Create a new merkle tree with the provided keys.
     ///
     /// Keys are sorted by their numeric value, as described in the [`PoQ` specification](https://www.notion.so/nomos-tech/Proof-of-Quota-Specification-215261aa09df81d88118ee22205cbafe?source=copy_link#215261aa09df81ec850ad7965bf6e76b).
-    /// If the input vector is empty or of it larger than the maximum number of
-    /// leaves supported by this fixed-height Merkle tree, it returns an error.
+    /// If the input vector is empty or if it is larger than the maximum number
+    /// of leaves supported by this fixed-height Merkle tree, it returns an
+    /// error.
     pub fn new(mut keys: Vec<PublicKey>) -> Result<Self, Error> {
         if keys.is_empty() {
             return Err(Error::EmptyKeySet);


nomos-blend/scheduling/src/message_blend/provers/leader/tests.rs
```diff
@@ -51,7 +51,7 @@ async fn proof_generation() {
     // Next proof should still return `Some` since leadership proofs do not have a
     // maximum cap.
     timeout(
-        Duration::from_secs(3),
+        Duration::from_secs(5),
```
ntn-x2, line 54: Slow CI, particularly on Linux runners, make this test fail sometimes, so bumping it by a bit more should be fine,

nomos-services/blend/src/membership/service.rs
```diff
@@ -14,7 +18,17 @@ use overwatch::{
...
+    membership::{MembershipInfo, MembershipStream, ServiceMessage, ZkInfo, node_id},
+    merkle::MerkleTree,
+};
+
+/// Wrapper around [`Node`] that includes its ZK public key.
+#[derive(Debug, Clone)]
+struct ZkNode<NodeId> {
+    pub node: Node<NodeId>,
+    pub zk_key: PublicKey,
+}
```
youngjoon-lee, line 31: Introducing this wrapper is clever.

nomos-services/blend/src/merkle.rs
```diff
@@ -0,0 +1,356 @@
...
+            .finish()
+    }
+}
+
+impl MerkleTree {
+    /// Create a new merkle tree with the provided keys.
+    ///
+    /// Keys are sorted by their numeric value, as described in the [`PoQ` specification](https://www.notion.so/nomos-tech/Proof-of-Quota-Specification-215261aa09df81d88118ee22205cbafe?source=copy_link#215261aa09df81ec850ad7965bf6e76b).
+    /// If the input vector is empty or of it larger than the maximum number of
+    /// leaves supported by this fixed-height Merkle tree, it returns an error.
```
youngjoon-lee:
```suggestion
    /// If the input vector is empty or if it larger than the maximum number of
    /// leaves supported by this fixed-height Merkle tree, it returns an error.
```
A minor typo

nodes/nomos-node/config.yaml
```diff
@@ -45,6 +45,8 @@ blend:
     message_frequency_per_round: 1.0
     redundancy_parameter: 0
     maximum_release_delay_in_rounds: 3
+  zk:
+    secret_key: 32e4b52b78e0e8273f31e2ae0826d09b0348d4c66824af4f51ec4ef338082211
```
danielSanchezQ, line 49: Why this value? Just curiosity if its an easter egg 😉

nomos-services/blend/src/core/settings.rs
```diff
@@ -145,6 +147,19 @@ impl<BackendSettings> BlendConfig<BackendSettings> {
     }
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct ZkSettings {
+    #[serde(rename = "secret_key")]
+    pub sk: SecretKey,
+}
```
danielSanchezQ, line 154: This will migrate to use the kms later on right?

nomos-services/blend/src/membership/service.rs
```diff
@@ -68,12 +83,30 @@ where
...
+                    let membership = Membership::new(&membership_nodes, &signing_public_key);
+                    let zk_info = ZkInfo {
+                        core_and_path_selectors,
+                        root: zk_tree.root(),
+                    };
+                    MembershipInfo {
+                        membership,
+                        zk: zk_info,
+                        // TODO: Replace with actual value.
+                        session_number: 0,
```
danielSanchezQ, line 108: Issue to keep track of this (?)

nodes/nomos-node/config.yaml
ntn-x2: No, it's just an example value for a zk key.

nomos-services/blend/src/core/settings.rs
ntn-x2: Yes, we will migrate keys to KMS later on.

nomos-services/blend/src/membership/service.rs
ntn-x2: We have it: https://github.com/logos-co/nomos/issues/1784. This issue involves switching to the chain broadcast service, which returns both zk keys and session numbers. There's also a PR already open (but blocked): https://github.com/logos-co/nomos/pull/1810.

PR #1822, opened 2025-10-21
feat(blend): collect tokens in core service

youngjoon-lee:

This is the 2nd PR for rewarding. This integrates the activity proof (implemented in #1813) to the blend core service. 

Key changes:
- Blend core service now computes an activity proof once a session transition period has passed.
  - Submitting the activity proof to the SDP service will be implemented in a next PR. I put TODO comments.
- We compute a [session randomness](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df814bbff2fe7e3ea866dc) when a new session starts.

I believe so.

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/message/src/reward/mod.rs
+++ b/nomos-blend/message/src/reward/mod.rs
@@ -8,6 +8,7 @@ pub use activity::ActivityProof;
 use nomos_core::sdp::SessionNumber;
 pub use session::SessionInfo;
 pub use token::BlendingToken;
+use tracing::warn;
 
 use crate::reward::session::SessionRandomness;
 
@@ -60,6 +61,13 @@ impl BlendingTokenCollector {
     /// an activity proof, it is discarded because its activity proof is
     /// no longer acceptable by the protocol.
     pub fn rotate_session(&mut self, new_session_info: &SessionInfo) {
+        if self.previous_session_tokens.is_some() {
+            warn!(
+                target: LOG_TARGET,
+                "Rotating to a new session while previous session tokens are still unconsumed. Those tokens will be discarded."
+            );
+        }
+
         self.previous_session_tokens = Some(std::mem::replace(
             &mut self.current_session_tokens,
             BlendingTokensForSession::new(


nomos-blend/message/src/reward/session.rs
```diff
@@ -16,11 +18,15 @@ pub struct SessionInfo {
...
-        session_randomness: SessionRandomness,
+        pol_epoch_nonce: ZkHash,
         num_core_nodes: u64,
-        total_core_quota: u64,
+        core_quota: u64,
         message_frequency_per_round: NonNegativeF64,
     ) -> Result<Self, Error> {
+        let total_core_quota = core_quota
+            .checked_mul(num_core_nodes)
+            .ok_or(Error::TotalCoreQuotaTooLarge(u64::MAX))?;
```
youngjoon-lee, line 27: Now I understand why you @ntn-x2 suggested https://github.com/logos-co/nomos/pull/1813#discussion_r2439901797. I realized that it's cleaner to compute the `total_core_quota` here inside `SessionInfo` because all other code deals with only `core_quota`. Also, it's okay that this module knows how the `total_core_quota` is computed, because its definition is already implied in the way calculating the activity threshold.

nomos-blend/message/src/encap/decapsulated.rs
```diff
@@ -21,8 +23,14 @@ pub enum DecapsulationOutput<const ENCAPSULATION_COUNT: usize> {
     reason = "Size difference between variants is not too large (small ENCAPSULATION_COUNT)"
 )]
 pub(super) enum PartDecapsulationOutput<const ENCAPSULATION_COUNT: usize> {
-    Incompleted((EncapsulatedPart<ENCAPSULATION_COUNT>, PublicHeader)),
-    Completed(Payload),
+    Incompleted(
```
ntn-x2: I think it's the right time to introduce a struct for this 🙈

ntn-x2: Maybe for all enums. What do you think? Can be a separate PR or we can just not do it, tho.

nomos-blend/message/src/reward/session.rs
ntn-x2: Yes the concept of total core quota is actually only used for rewards, so I think it's fair to let this bit of code know how to compute that.

nomos-blend/message/src/reward/session.rs
```diff
@@ -62,6 +70,21 @@ impl From<[u8; 64]> for SessionRandomness {
...
+const SESSION_RANDOMNESS_TAG: [u8; 27] = *b"BLEND_SESSION_RANDOMNESS_V1";
+
+impl SessionRandomness {
+    /// Derive the session randomness from the given session number and epoch
+    /// nonce.
+    #[must_use]
+    fn new(session_number: SessionNumber, epoch_nonce: ZkHash) -> Self {
+        let mut hasher = Blake2b512::new();
+        hasher.update(SESSION_RANDOMNESS_TAG);
+        hasher.update(BigUint::from(epoch_nonce).to_bytes_le());
```
ntn-x2: We should use `fr_to_bytes`

nomos-blend/message/src/reward/session.rs
```diff
@@ -62,6 +70,21 @@ impl From<[u8; 64]> for SessionRandomness {
...
 
+const SESSION_RANDOMNESS_TAG: [u8; 27] = *b"BLEND_SESSION_RANDOMNESS_V1";
+
+impl SessionRandomness {
+    /// Derive the session randomness from the given session number and epoch
+    /// nonce.
+    #[must_use]
+    fn new(session_number: SessionNumber, epoch_nonce: ZkHash) -> Self {
+        let mut hasher = Blake2b512::new();
+        hasher.update(SESSION_RANDOMNESS_TAG);
```
ntn-x2: Do you think it makes more sense to call the upper `blake2b512` function, and give it a chained slice of the inputs? Then we would avoid repeating logic. As far as I can see, it would be the same logic.

nomos-services/blend/src/core/mod.rs
```diff
@@ -504,6 +516,21 @@ where
...
+                current_public_inputs.leader.pol_epoch_nonce,
+                membership.size() as u64,
+                poq_core_public_inputs.quota,
+                settings.scheduler.cover.message_frequency_per_round,
+            )
+            .expect("Reward session info must be created successfully. Panicking since the service cannot continue with this session");
+            if let Some(activity_proof) =
+                blending_token_collector.rotate_session(&reward_session_info)
+            {
+                info!(target: LOG_TARGET, "Activity proof generated: {activity_proof:?}");
```
ntn-x2: So here what we get is the activity proof of a previous session before the one we are actually transitioning from. Will we still be able to submit activity proofs for it? Just asking for my own understanding.

ntn-x2: Also I think this log line should be different than the one below, since we are effectively doing something different.

nomos-services/blend/src/core/mod.rs
```diff
@@ -519,9 +546,10 @@ where
             Ok((new_processor, new_inputs))
         }
         SessionEvent::TransitionPeriodExpired => {
-            // TODO: Send an activity message to SDP service.
-            // Also, do it before this service is terminated before this event is
-            // detected.
+            if let Some(activity_proof) = blending_token_collector.activity_proof() {
```
ntn-x2: Now that I understand better how this method is used, I think its name could be improved. It does consume the activity proofs from the session whose transition period is now over, and I think the name should somehow reflect that.

nomos-services/blend/src/core/mod.rs
youngjoon-lee: Good point. I just checked the [Service Reward Distribution](https://www.notion.so/nomos-tech/Service-Reward-Distribution-Protocol-211261aa09df816d92c8c6ec00c6e227?source=copy_link#211261aa09df81ac85a9d8595f267851) spec again, and recalled that it is not meaningful to submit proofs in this case, because the SRD spec will anyway ignore them. 

I will modify the `rotate_session` to discard the old tokens without computing an outdated activity proof, as I implemented at the beginning. I think it's reasonable.

nomos-services/blend/src/core/mod.rs
youngjoon-lee: I agree. I renamed it to `compute_activity_proof_for_previous_session`: 9485f67aea4968bdd7c3de0689882404a0a29325. This name doesn't include a word `consume`, but I think this implies that the blending tokens collected during the previous session are used to compute an activity proof. I can't find any better name for now. Instead, I explained details in the doc comment. Wdyt?

nomos-blend/message/src/reward/session.rs
youngjoon-lee: Are you probably referring to the [`crate::reward::token::hash`](https://github.com/logos-co/nomos/blob/9485f67aea4968bdd7c3de0689882404a0a29325/nomos-blend/message/src/reward/token.rs#L52) function? That is for supporting variable size output and returning `Vec<u8>`. Here, I wanted to allocate only a 64-byte array. Wdyt?

nomos-blend/message/src/encap/decapsulated.rs
youngjoon-lee: Yeah, I thought about that too. Applied in 7bbf5fb7acbd0a98d2bded6644bbb3604e5330b1

nomos-blend/message/src/reward/session.rs
youngjoon-lee: Oh, I forgot about it. Thanks. aaa5fb87e28f85585211b027cf189f15ee3076ac

nomos-services/blend/src/core/mod.rs
youngjoon-lee: 3680006826306cba3f5557925762c47b20714663 I modified the `rotate_session`.

nomos-blend/message/src/reward/session.rs
ntn-x2: No we have `crate::crypto::blake2b512`, although it's not visible to the outside, since we are dealing with same-create operations, I think we could make it `pub(crate)`.

nomos-services/blend/src/core/mod.rs
ntn-x2: This looks good. I would have still returned the old one, in case we want to log it as a warning or something (since this is not a scenario that should normally happen), but we just don't submit those proofs. Anyway, this is also fine.

nomos-services/blend/src/core/mod.rs
ntn-x2: Yup, sounds good.

ntn-x2 (approved):
There's two minor open comments: https://github.com/logos-co/nomos/pull/1822#discussion_r2450532709 & https://github.com/logos-co/nomos/pull/1822#discussion_r2450534891, but it anyway looks good, so I'm approving already.

nomos-services/blend/src/core/mod.rs
youngjoon-lee: Yeah, good to print a warning log. I added the warning log in the `rotate_session` b8d548a347e4813c0c2d007381245c90d9e58f54, instead of computing/returning an activity proof (though it's not super expensive). I think this is fine enough. Thanks!

nomos-blend/message/src/reward/session.rs
youngjoon-lee: Ah, thank you. bab848711327b62473f3a0627ac1b7bd966840c5 I refactored the `blake2b512` little bit, so that we don't need to do `concat()` or `chain()`. I think this version is more efficient.

Unfortunately, `Blake2b512::update()` accepts `impl AsRef<[u8]>`. So, we cannot directly pass a chained byte array to it.

PR #1825, opened 2025-10-21
feat(test): wait nodes to be ready

andrussal:

Instead of heuristic sleep in integration tests to estimate when network is ready, use available API endpoints to checks nodes status.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/tests/src/topology/mod.rs
+++ b/tests/src/topology/mod.rs
@@ -568,10 +568,7 @@ fn find_expected_peer_counts(
             }
 
             expected[idx].insert(peer_idx);
-
-            if peer_idx < expected.len() {
-                expected[peer_idx].insert(idx);
-            }
+            expected[peer_idx].insert(idx);
         }
     }
 


tests/src/topology/mod.rs
```diff
@@ -255,4 +261,287 @@ impl Topology {
...
+        let expected_peer_counts = find_expected_peer_counts(&listen_ports, &initial_peer_ports);
+        let labels = self.node_labels();
+
+        let readiness = timeout(adjust_timeout(TIMEOUT), async {
+            loop {
+                let validator_infos =
+                    join_all(self.validators.iter().map(Validator::network_info)).await;
+
+                let executor_infos =
+                    join_all(self.executors.iter().map(Executor::network_info)).await;
```
danielSanchezQ: As this is async, you can just get both futures and await both of them at the same time.

tests/src/topology/mod.rs
```diff
@@ -255,4 +261,287 @@ impl Topology {
...
+
+                sleep(POLL_INTERVAL).await;
+            }
+        });
+
+        if readiness.await.is_err() {
+            let validator_infos =
+                join_all(self.validators.iter().map(Validator::network_info)).await;
+
+            let executor_infos = join_all(self.executors.iter().map(Executor::network_info)).await;
```
danielSanchezQ: Likewise

tests/src/topology/mod.rs
```diff
@@ -255,4 +261,287 @@ impl Topology {
...
+    }
+
+    pub async fn wait_membership_assignations_empty(&self, session: SessionNumber) {
+        self.wait_membership_ready_with(session, "empty assignations", |resp| {
+            resp.assignations.is_empty()
+        })
+        .await;
+    }
+
+    pub async fn wait_membership_ready_with<F>(
```
danielSanchezQ: Can't `wait_network_ready` be specialized version of this since they are (or look so) the same code?

andrussal: They are quite similar by overall setup but `wait_network_ready` asks `network_info` endpoint and `wait_membership_ready_with` checks `da_get_membership` endpoint.

danielSanchezQ: I think it can be done, but if its too complicated no need.

tests/src/topology/mod.rs
andrussal: > As this is async, you can just get both futures and await both of them at the same time.

Right, good point!

tests/src/topology/mod.rs
andrussal: > I think it can be done, but if its too complicated no need.

I went for it and added a trait that encapsulates the flow.

tests/src/topology/mod.rs
```diff
@@ -255,4 +261,318 @@ impl Topology {
...
+            let Some(peer_idx) = listen_ports.iter().position(|p| p == port) else {
+                continue;
+            };
+            if peer_idx == idx {
+                continue;
+            }
+
+            expected[idx].insert(peer_idx);
+
+            if peer_idx < expected.len() {
```
davidrusu: why is this necessary? you check `peer_idx == idx` above, that seems to hint that indices are consistent across the peer ports and listen ports? then that should mean that `expected.len() ==listen_ports.len()`, and `peer_idx` should always be smaller than `expected.len()`?

andrussal: > why is this necessary? you check `peer_idx == idx` above, that seems to hint that indices are consistent across the peer ports and listen ports? then that should mean that `expected.len() ==listen_ports.len()`, and `peer_idx` should always be smaller than `expected.len()`?

Good catch!

tests/src/topology/mod.rs
```diff
@@ -255,4 +261,318 @@ impl Topology {
...
+        format!("timed out waiting for network readiness: {summary}")
+    }
+}
+
+struct MembershipReadiness<'a> {
+    topology: &'a Topology,
+    session: SessionNumber,
+    labels: &'a [String],
+    expect_non_empty: bool,
+}
```
danielSanchezQ, line 446: Superb!

bacv (approved):

PR #1826, opened 2025-10-21
feat(wallet): MantleTxBuilder + Wallet::fund_tx API

davidrusu:

This PR introduce a wallet API for funding MantleTx's with notes from the wallet.

To support that API, we also introduce a MantleTxBuilder which holds on to the Utxo's used in the ledger_tx. This is required because the inputs in ledger tx's are just NoteId's. that's not sufficient to read the net_balance of the transaction.

yes

@davidrusu

NA

NA

## Checklist

* [ ] 2. Description added.

### DIFF
--- a/nomos-services/wallet/src/lib.rs
+++ b/nomos-services/wallet/src/lib.rs
@@ -29,7 +29,7 @@ use serde::{Serialize, de::DeserializeOwned};
 use services_utils::wait_until_services_are_ready;
 use tokio::sync::oneshot;
 use tracing::{debug, error, info, trace};
-use wallet::{Wallet, WalletBlock, WalletError, WalletState};
+use wallet::{Wallet, WalletBlock, WalletError};
 
 #[derive(Debug, thiserror::Error)]
 pub enum WalletServiceError {
@@ -240,9 +240,20 @@ where
         storage: &StorageAdapter<Storage, Tx, RuntimeServiceId>,
         cryptarchia: &CryptarchiaServiceApi<Cryptarchia, RuntimeServiceId>,
     ) {
+        match &msg {
+            WalletMsg::GetBalance { tip, .. }
+            | WalletMsg::FundTx { tip, .. }
+            | WalletMsg::GetLeaderAgedNotes { tip, .. } => {
+                if let Err(err) =
+                    Self::backfill_if_not_in_sync(*tip, wallet, storage, cryptarchia).await
+                {
+                    error!(err=?err, "Failed backfilling wallet to message tip {tip:?}, will attempt to continue processing the message {msg:?}");
+                }
+            }
+        }
+
         match msg {
             WalletMsg::GetBalance { tip, pk, resp_tx } => {
-                Self::try_sync_to_tip(tip, wallet, storage, cryptarchia).await;
                 let balance = wallet.balance(tip, pk);
                 if resp_tx.send(balance).is_err() {
                     error!("Failed to respond to GetBalance");
@@ -256,28 +267,12 @@ where
                 funding_pks,
                 resp_tx,
             } => {
-                Self::try_sync_to_tip(tip, wallet, storage, cryptarchia).await;
                 Self::fund_tx(tip, &tx_builder, change_pk, funding_pks, resp_tx, wallet);
             }
 
             WalletMsg::GetLeaderAgedNotes { tip, resp_tx } => {
-                Self::try_sync_to_tip(tip, wallet, storage, cryptarchia).await;
                 Self::get_leader_aged_notes(tip, resp_tx, wallet, cryptarchia).await;
             }
-        }
-    }
-
-    async fn try_sync_to_tip(
-        tip: HeaderId,
-        wallet: &mut Wallet,
-        storage: &StorageAdapter<Storage, Tx, RuntimeServiceId>,
-        cryptarchia: &CryptarchiaServiceApi<Cryptarchia, RuntimeServiceId>,
-    ) {
-        if let Err(err) =
-            Self::fetch_wallet_state_with_backfill(tip, wallet, storage, cryptarchia).await
-        {
-            // If there is a failure during backfilling, we log the error and continue.
-            error!(err=?err, "Failed while backfilling wallet during responding to fund_tx");
         }
     }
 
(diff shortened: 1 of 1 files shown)


wallet/src/lib.rs
```diff
@@ -57,6 +60,89 @@ impl WalletState {
...
+            let funded_tx_builder = tx_builder
+                .clone()
+                .extend_ledger_inputs(utxos[..i].iter().copied());
+
+            let funded_cost = i128::from(funded_tx_builder.clone().build().gas_cost::<G>());
+            let funded_net_balance = funded_tx_builder.net_balance();
+            if funded_net_balance == funded_cost {
+                // We can exactly pay the tx cost, no change note needed.
+                return Ok(funded_tx_builder);
+            } else if funded_net_balance > funded_cost {
```
danielSanchezQ:
```rust
if {
    return;
}
if funded_net_balance > funded_cost {
   return;
}
...
```

wallet/src/lib.rs
```diff
@@ -57,6 +60,89 @@ impl WalletState {
...
+                    funded_tx_with_change_builder.net_balance(),
+                    i128::from(
+                        funded_tx_with_change_builder
+                            .clone()
+                            .build()
+                            .gas_cost::<G>()
+                    )
+                );
+
+                return Ok(funded_tx_with_change_builder);
```
danielSanchezQ: This can probably be a method in itself? Maybe the continue will make it a bit cumbersome as a method tho.

wallet/src/lib.rs
```diff
@@ -57,6 +60,89 @@ impl WalletState {
...
+        tx_builder: MantleTxBuilder,
+        change_pk: PublicKey,
+        pks: impl IntoIterator<Item = PublicKey>,
+    ) -> Result<MantleTxBuilder, WalletError> {
+        let mut utxos: Vec<Utxo> = pks
+            .into_iter()
+            .filter_map(|pk| self.pk_index.get(&pk))
+            .flatten()
+            .map(|id| self.utxos[id])
+            .collect();
```
danielSanchezQ: Should we consume the PublicKeys? or `&PublicKey` is enough?

wallet/src/lib.rs
```diff
@@ -57,6 +60,89 @@ impl WalletState {
...
+        &self,
+        tx_builder: MantleTxBuilder,
+        change_pk: PublicKey,
+        pks: impl IntoIterator<Item = PublicKey>,
+    ) -> Result<MantleTxBuilder, WalletError> {
+        let mut utxos: Vec<Utxo> = pks
+            .into_iter()
+            .filter_map(|pk| self.pk_index.get(&pk))
+            .flatten()
+            .map(|id| self.utxos[id])
```
danielSanchezQ: Maybe `flat_map` instead?

wallet/src/lib.rs
```diff
@@ -57,6 +60,89 @@ impl WalletState {
...
+
+                // The change note will slightly increase the storage cost of the tx.
+                // We'll first add a dummy change note in order to measure the impact on cost.
+                let funded_with_change_cost = i128::from(
+                    funded_tx_builder
+                        .clone()
+                        .add_ledger_output(Note {
+                            value: 1,
+                            pk: change_pk,
+                        })
```
danielSanchezQ: Can we have a method for building this?
```rust
pub const fn dummy_single_value_note(pk: PublicKey) -> Note {
    Note {
        value: 1,
        pk,
    }
}
```

wallet/src/lib.rs
davidrusu: I'll try to push some of this logic into the MantleTxBuilder as much as possible. I think that will it easier to read

wallet/src/lib.rs
davidrusu: I figured since PublicKey is Copy, it's low overhead to consume here.

wallet/src/lib.rs
davidrusu: what about pushing this into the builder? e.g. something like: 
```rust
tx_builder.gas_cost_with_added_change_note::<G>()
```

danielSanchezQ: > what about pushing this into the builder? e.g. something like:
> 
> ```rust
> tx_builder.gas_cost_with_added_change_note::<G>()
> ```

Sounds better, thanks!

wallet/src/lib.rs
```diff
@@ -57,14 +64,69 @@ impl WalletState {
...
+            .flatten()
+            .map(|id| self.utxos[id])
+            .collect()
+    }
+
+    pub fn fund_tx<G: GasConstants>(
+        &self,
+        tx_builder: &MantleTxBuilder,
+        change_pk: PublicKey,
+        pks: impl IntoIterator<Item = impl Borrow<PublicKey>>,
```
davidrusu, line 82: @danielSanchezQ the `Borrow` trait lets the caller decide if they want to consume or pass by reference. Best of both worlds!

wallet/src/lib.rs
davidrusu: Doesn't work here, the previous iterator step returns something like `Iterator<Item=BTreeSet<Utxo>>`

so, we're flattening the prior step, not this one.
With flat_map we would need something like:
```
.flat_map(|set| set.iter().map(|id| self.utxos[id]))
```

which defeats the point of the flatten which was to avoid doing nested iterations

nomos-services/wallet/src/lib.rs
```diff
@@ -51,17 +54,18 @@ pub enum WalletMsg {
     GetBalance {
         tip: HeaderId,
         pk: PublicKey,
-        tx: oneshot::Sender<Result<Option<Value>, WalletError>>,
+        resp_tx: oneshot::Sender<Result<Option<Value>, WalletError>>,
```
davidrusu, line 57: renamed these fields to avoid naming collision with `tx_builder`. Don't want people confusing this with a mantle tx when reading.

nomos-services/wallet/src/lib.rs
```diff
@@ -228,62 +237,86 @@ where
...
-                    "Failed to fetch wallet state"
+                error!(err = ?err, "Failed to fetch wallet state");
+                Self::send_err(
+                    tx,
+                    WalletServiceError::FailedToFetchWalletStateForBlock(tip),
                 );
-                Self::send_err(tx, err);
                 return;
             }
         };
```
davidrusu, line 317: Change here was to factor out the backfilling step before this handler function was called. This way we have a common backfilling pattern for all message handlers. (all wallet messages need this JIT backfilling behaviour)

nomos-services/wallet/src/lib.rs
```diff
@@ -228,62 +237,81 @@ where
...
+            WalletMsg::GetBalance { tip, .. }
+            | WalletMsg::FundTx { tip, .. }
+            | WalletMsg::GetLeaderAgedNotes { tip, .. } => {
+                if let Err(err) =
+                    Self::backfill_if_not_in_sync(*tip, wallet, storage, cryptarchia).await
+                {
+                    error!(err=?err, "Failed backfilling wallet to message tip {tip:?}, will attempt to continue processing the message {msg:?}");
+                }
+            }
+        }
```
danielSanchezQ, line 253: Nitpick, if all msg constructs include a tip, better to add a `tip()` method to the `WalletMsg`:

```rust 
if let Err(err) = Self::backfill_if_not_in_sync(msg.tip(), wallet, storage, cryptarchia).await {
    error!(err=?err, "Failed backfilling wallet to message tip {tip:?}, will attempt to continue processing the message {msg:?}");
}
```

danielSanchezQ (approved):
Left a nitpick, otherwise really nice stuff. Thanks @davidrusu !

nomos-services/wallet/src/lib.rs
davidrusu: True! do you mind if I do this in the next PR (with the KMS changes) there are other branches forking of this one, I don't want to complicate their history.

danielSanchezQ: > True! do you mind if I do this in the next PR (with the KMS changes) there are other branches forking of this one, I don't want to complicate their history.

Sure, go for it. Thanks!

PR #1829, opened 2025-10-21
chore(membership): Membership in da services

bacv:

Update sdp ledger with the declarations from genesis transaction. Broadcast genesis state (or any loaded state) when cryptarchia is initialized. DA network service subscribes to `broadcast-service` for session updates. Testcases updated to construct genesis transaction according to the node configuration.

Yes, preparation to remove membership service and get all session related information from nomos ledger.

@bacv @zeegomo 

Yes

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-services/chain/chain-service/src/lib.rs
+++ b/nomos-services/chain/chain-service/src/lib.rs
@@ -611,6 +611,7 @@ where
         wait_until_services_are_ready!(
             &self.service_resources_handle.overwatch_handle,
             Some(Duration::from_secs(60)),
+            BlockBroadcastService<_>,
             NetworkService<_, _>,
             TxMempoolService<_, _, _, _, _, _>,
             DaSamplingService<_, _, _, _>,
--- a/nomos-services/chain/broadcast-service/src/lib.rs
+++ b/nomos-services/chain/broadcast-service/src/lib.rs
@@ -13,15 +13,21 @@ use overwatch::{
     },
 };
 use serde::{Deserialize, Serialize};
-use tokio::sync::{broadcast, oneshot, watch};
+use tokio::sync::{broadcast, oneshot};
 use tracing::{error, info};
 
 const BROADCAST_CHANNEL_SIZE: usize = 128;
 
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct SessionUpdate {
     pub session_number: SessionNumber,
     pub providers: HashMap<ProviderId, ProviderInfo>,
+}
+
+#[derive(Debug)]
+pub struct SessionSubscription {
+    pub last_session: Option<SessionUpdate>,
+    pub receiver: broadcast::Receiver<SessionUpdate>,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
@@ -39,18 +45,21 @@ pub enum BlockBroadcastMsg {
         result_sender: oneshot::Sender<broadcast::Receiver<BlockInfo>>,
     },
     SubscribeBlendSession {
-        result_sender: oneshot::Sender<watch::Receiver<Option<SessionUpdate>>>,
+        result_sender: oneshot::Sender<SessionSubscription>,
     },
     SubscribeDASession {
-        result_sender: oneshot::Sender<watch::Receiver<Option<SessionUpdate>>>,
+        result_sender: oneshot::Sender<SessionSubscription>,
     },
 }
 
 pub struct BlockBroadcastService<RuntimeServiceId> {
     service_resources_handle: OpaqueServiceResourcesHandle<Self, RuntimeServiceId>,
     finalized_blocks: broadcast::Sender<BlockInfo>,
-    blend_session: watch::Sender<Option<SessionUpdate>>,
-    da_session: watch::Sender<Option<SessionUpdate>>,
+    blend_session: broadcast::Sender<SessionUpdate>,
+    da_session: broadcast::Sender<SessionUpdate>,
+    // For sending latest session on subscription.
+    last_blend_session: Option<SessionUpdate>,
+    last_da_session: Option<SessionUpdate>,
 }
 
 impl<RuntimeServiceId> ServiceData for BlockBroadcastService<RuntimeServiceId> {
@@ -70,14 +79,16 @@ where
         _initial_state: Self::State,
     ) -> Result<Self, overwatch::DynError> {
         let (finalized_blocks, _) = broadcast::channel(BROADCAST_CHANNEL_SIZE);
-        let (blend_session, _) = watch::channel(None);
-        let (da_session, _) = watch::channel(None);
+        let (blend_session, _) = broadcast::channel(BROADCAST_CHANNEL_SIZE);
+        let (da_session, _) = broadcast::channel(BROADCAST_CHANNEL_SIZE);
 
         Ok(Self {
             service_resources_handle,
             finalized_blocks,
             blend_session,
             da_session,
+            last_blend_session: None,
+            last_da_session: None,
         })
     }
 
@@ -96,12 +107,14 @@ where
                     }
                 }
                 BlockBroadcastMsg::BroadcastBlendSession(session) => {
-                    if let Err(err) = self.blend_session.send(Some(session)) {
+                    self.last_blend_session = Some(session.clone());
+                    if let Err(err) = self.blend_session.send(session) {
                         error!("Could not send to new blocks channel: {err}");
                     }
                 }
                 BlockBroadcastMsg::BroadcastDASession(session) => {
-                    if let Err(err) = self.da_session.send(Some(session)) {
+                    self.last_da_session = Some(session.clone());
+                    if let Err(err) = self.da_session.send(session) {
                         error!("Could not send to new blocks channel: {err}");
                     }
                 }
@@ -114,12 +127,18 @@ where
                     }
                 }
                 BlockBroadcastMsg::SubscribeBlendSession { result_sender } => {
-                    if let Err(err) = result_sender.send(self.blend_session.subscribe()) {
+                    if let Err(err) = result_sender.send(SessionSubscription {
+                        last_session: self.last_blend_session.clone(),
+                        receiver: self.blend_session.subscribe(),
+                    }) {
                         error!("Could not subscribe to blend session channel: {err:?}");
                     }
                 }
                 BlockBroadcastMsg::SubscribeDASession { result_sender } => {
-                    if let Err(err) = result_sender.send(self.da_session.subscribe()) {
+                    if let Err(err) = result_sender.send(SessionSubscription {
+                        last_session: self.last_da_session.clone(),
+                        receiver: self.da_session.subscribe(),
+                    }) {
                         error!("Could not subscribe to DA session channel: {err:?}");
                     }
                 }
--- a/nomos-services/data-availability/network/src/membership/adapters/service.rs
+++ b/nomos-services/data-availability/network/src/membership/adapters/service.rs
@@ -1,13 +1,13 @@
 use std::collections::{BTreeSet, HashMap};
 
-use broadcast_service::BlockBroadcastService;
-use futures::{StreamExt as _, future};
+use broadcast_service::{BlockBroadcastService, SessionSubscription};
+use futures::{StreamExt as _, future, stream::iter};
 use libp2p::{Multiaddr, PeerId, core::signed_envelope::DecodingError};
 use nomos_core::sdp::{Locator, ProviderId, ProviderInfo, SessionNumber};
 use nomos_libp2p::ed25519;
 use overwatch::services::{ServiceData, relay::OutboundRelay};
 use tokio::sync::oneshot;
-use tokio_stream::wrappers::WatchStream;
+use tokio_stream::wrappers::BroadcastStream;
 
 use crate::membership::{
     MembershipAdapter, MembershipAdapterError, PeerMultiaddrStream, SubnetworkPeers,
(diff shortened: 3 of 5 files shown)


ledger/nomos-ledger/src/mantle/sdp/mod.rs
```diff
@@ -131,6 +131,19 @@ impl ServiceState {
...
+    fn declare_genesis(
+        &mut self,
+        id: DeclarationId,
+        declaration: Declaration,
+    ) -> Result<(), Error> {
+        if self.declarations.contains_key(&id) {
+            return Err(Error::DuplicateDeclaration(id));
+        }
+        self.declarations = self.declarations.insert(id, declaration.clone());
+        self.active.declarations = self.active.declarations.insert(id, declaration);
```
zeegomo: aren't we missing the forming session? can we reuse `declare` and clone that set for both forming and active?

ledger/nomos-ledger/src/mantle/sdp/mod.rs
```diff
@@ -302,6 +313,22 @@ impl SdpLedger {
...
+        let declaration_id = op.id();
+        let declaration = Declaration::new(self.block_number, op);
+        if let Some(service_state) = self.services.get_mut(&op.service_type) {
+            service_state.declare_genesis(declaration_id, declaration)?;
+        } else {
+            return Err(Error::ServiceNotFound(op.service_type));
+        }
+
+        Ok(self)
+    }
```
zeegomo: I'd prefer a `from_genesis_tx` method (and remove the empty `new`) like in other parts of the ledger so that it's better typed

ledger/nomos-ledger/src/mantle/mod.rs
```diff
@@ -183,6 +184,38 @@ impl LedgerState {
...
+        for (op, proof) in ops {
+            match (op, proof) {
+                // The signature for channel ops can be verified before reaching this point,
+                // as you only need the signer's public key and tx hash
+                // Callers are expected to validate the proof before calling this function.
+                (Op::ChannelBlob(op), None) => {
+                    self.channels =
+                        self.channels
+                            .apply_msg(op.channel, &op.parent, op.id(), &op.signer)?;
+                }
```
zeegomo: we do not support blob in genesis 🤔

ledger/nomos-ledger/src/mantle/mod.rs
```diff
@@ -183,6 +184,38 @@ impl LedgerState {
 
         Ok((self, balance))
     }
+
+    fn try_apply_genesis_ops<'a>(
```
zeegomo: why do we need this?

pradovic (approved):
Looks good from DA side!

ledger/nomos-ledger/src/mantle/sdp/mod.rs
```diff
@@ -131,6 +131,19 @@ impl ServiceState {
...
 
+    fn declare_genesis(
+        &mut self,
+        id: DeclarationId,
+        declaration: Declaration,
+    ) -> Result<(), Error> {
+        if self.declarations.contains_key(&id) {
+            return Err(Error::DuplicateDeclaration(id));
+        }
+        self.declarations = self.declarations.insert(id, declaration.clone());
```
ntn-x2: I see there's an `insert_mut` that avoids cloning the whole map. Would that not be more efficient?

ledger/nomos-ledger/src/mantle/sdp/mod.rs
ntn-x2: I guess also here we could use `insert_mut`?

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -1539,6 +1524,66 @@ where
...
+                error!("Could not get active session numbers for block {block_id:?}: {e}");
+                return;
+            }
+        };
+
+        for (service, new_session_number) in &new_sessions {
+            // If `previous_sessions` is provided, check if the session number has changed.
+            // Otherwise, always broadcast (for initialization).
+            let should_broadcast = previous_sessions
+                .is_none_or(|prev| prev.get(service).copied().unwrap_or(0) != *new_session_number);
```
ntn-x2: I guess we know this is not `None` because we called `is_none_or`, so an `expect` is probably more explicit here.

```suggestion
                .is_none_or(|prev| prev.get(service).copied().expect("") != *new_session_number);
```

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -1539,6 +1524,66 @@ where
...
+
+            if should_broadcast {
```
ntn-x2: A nit: I think it would be better to call `if !should_broadcast { return;}` and exit immediately, so we avoid one level of indentation.

tests/src/benchmarks/mod.rs
```diff
@@ -1 +1,2 @@
-pub mod eddsa;
+// TODO: Reenable when migrated codebase migrates to ed25519_dalek 3.0.0.pre.0
```
ntn-x2:
```suggestion
// TODO: Reenable when codebase migrates to ed25519_dalek 3.0.0.pre.0
```

ledger/nomos-ledger/src/mantle/sdp/mod.rs
danielSanchezQ: wdym by `map` here?

ledger/nomos-ledger/src/mantle/mod.rs
```diff
@@ -183,6 +184,38 @@ impl LedgerState {
...
+        mut self,
+        config: &Config,
+        _utxo_tree: &UtxoTree,
+        ops: impl Iterator<Item = (&'a Op, Option<&'a OpProof>)> + 'a,
```
danielSanchezQ: Wouldn't it be better to take a GenesisTx instead? So its clear that it only works on genesis ops?

ledger/nomos-ledger/src/mantle/mod.rs
danielSanchezQ: Yeah, this should probably be a panic, as it should not be there (?)

ledger/nomos-ledger/src/mantle/sdp/mod.rs
bacv: Declarations from the genesis tx goes straight into the active declarations, forming session at this point should be empty.

ledger/nomos-ledger/src/mantle/sdp/mod.rs
ntn-x2:
```rust
#[must_use]
    pub fn insert(&self, key: K, value: V) -> HashTrieMap<K, V, P, H> {
        let mut new_map = self.clone();

        new_map.insert_mut(key, value);

        new_map
    }
```

vs

```rust
pub fn insert_mut(&mut self, key: K, value: V) {
        let entry = EntryWithHash::new(key, value, &self.hasher_builder);
        let is_new_key = SharedPointer::make_mut(&mut self.root).insert(entry, 0, self.degree);

        if is_new_key {
            self.size += 1;
        }
    }
```

But I see that what's cloned is a ref-counted pointer, so it's the same.

ledger/nomos-ledger/src/mantle/sdp/mod.rs
bacv: > I'd prefer a from_genesis_tx method (and remove the empty new) like in other parts of the ledger so that it's better typed

I'm not sure I get the suggestion, this method applies individual Op from genesis tx. Would `SdpLedger::from_genesis_tx` iterate over all ops from the genesis tx and only act on sdp declarations?

ledger/nomos-ledger/src/mantle/mod.rs
bacv: Currently genesis tx has no proofs, the `LedgerState::from_ledger_tx` appends `None` to all ops, this special method is only to handle this special case. When we add proofs to the genesis tx, then we could just use `LedgerState::try_apply_ops`.

ledger/nomos-ledger/src/mantle/mod.rs
bacv: We don't really need GenesisTx here, we only need Ops, and there are no difference between genesis ops and ordinary ops, for now we just do not provide Proofs for those ops. Lets first see if we keep special handling for genesis ops, or we should process proofs in this PR.

ledger/nomos-ledger/src/mantle/mod.rs
bacv: Fixed here https://github.com/logos-co/nomos/pull/1829/commits/11df0b7404a3f8527164d3d2e9b4e5cfa4aeaaa9

ledger/nomos-ledger/src/mantle/sdp/mod.rs
bacv: Updated to insert_mut here https://github.com/logos-co/nomos/pull/1829/commits/11df0b7404a3f8527164d3d2e9b4e5cfa4aeaaa9

nomos-services/chain/chain-service/src/lib.rs
bacv: Fixed here https://github.com/logos-co/nomos/pull/1829/commits/1e471dfa03baeb4584e75bd28e17f7d02b298bc1

nomos-services/chain/chain-service/src/lib.rs
bacv: Fixed https://github.com/logos-co/nomos/pull/1829/commits/1e471dfa03baeb4584e75bd28e17f7d02b298bc1

ledger/nomos-ledger/src/mantle/mod.rs
danielSanchezQ: But then rename it to `try_apply_ops`, because including the genesis in the naming is a verbal agreement that may be violated (and its confusing). That was my whole point, I know there is not difference in those ops. But if we really want this method to be run only on Genesis and we want to enforce it, better to actually do it.

bacv: Got it. I'll update this PR to include proofs in GenesisTx, this way we'll be able to use one `try_apply_ops`.

nomos-core/chain-defs/src/mantle/genesis_tx.rs
```diff
@@ -15,8 +16,10 @@ use crate::{
 };
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize)]
-#[serde(transparent)]
-pub struct GenesisTx(MantleTx);
+pub struct GenesisTx {
+    mantle_tx: MantleTx,
+    op_proofs: Vec<Option<OpProof>>,
```
danielSanchezQ: We removed the option for proofs, all inner ops should have their own proofs. (?)

nomos-core/chain-defs/src/mantle/genesis_tx.rs
```diff
@@ -105,14 +128,18 @@ impl GasCost for GenesisTx {
...
     fn genesis_inscription(&self) -> &InscriptionOp {
         // Safe to unwrap because we validated this in from_tx
-        match &self.0.ops[0] {
+        match &self.mantle_tx.ops[0] {
             Op::ChannelInscribe(op) => op,
             _ => unreachable!("GenesisTx always has a valid inscription as first op"),
         }
     }
 
+    fn op_proofs(&self) -> &Vec<Option<OpProof>> {
```
danielSanchezQ: If we remove the Option:
```rust
fn op_proofs(&self) -> &[OpProof] {
```

nomos-core/chain-defs/src/mantle/genesis_tx.rs
bacv: Inscription doesn't have it's proof. We could actually make it without `Option`, as first op is always inscription, and other at least for now can only be sdp declarations.

danielSanchezQ: Ok, if its like that lets keep the option. As in the other layers it accept option as well. But I didn't know the reason. Now its clear :+1:

ledger/nomos-ledger/src/mantle/mod.rs
bacv: Removed `try_apply_genesis_ops` and added `SdpLedger::from_genesis_tx` which filters sdp related ops and uses them to initialize the ledger.

ledger/nomos-ledger/src/mantle/sdp/mod.rs
```diff
@@ -220,10 +226,50 @@ impl SdpLedger {
...
+                }),
+            ) = (op, proof)
+            {
+                let Some((utxo, _)) = utxo_tree.utxos().get(&op.locked_note_id) else {
+                    return Err(Error::NoteNotFound(op.locked_note_id));
+                };
+                sdp =
+                    sdp.apply_declare_msg(op, utxo.note, zk_sig, ed25519_sig, tx.hash(), config)?;
+            }
+        }
```
zeegomo, line 251: ok I see what you meant now, we already have a `from_genesis_tx` in mantle ledger that does most of these operations anyway (e.g checking invalid ops and fetching notes). What aobut we have `from_genesis` with a list of declarations / proofs instead? Otherwise this is good

nomos-core/chain-defs/src/mantle/genesis_tx.rs
zeegomo: Can we have this be a `SignedMantleTx`?

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -1539,6 +1524,72 @@ where
 
         (cryptarchia, storage_blocks_to_remove)
     }
+
+    #[expect(
+        clippy::cognitive_complexity,
+        reason = "TODO: Address this at some point."
```
zeegomo: I'd like this resolved before merging if it does not require big refactors

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -1539,6 +1524,72 @@ where
...
+            let should_broadcast = previous_sessions.is_none_or(|prev| {
+                prev.get(service)
+                    .copied()
+                    .expect("previous session number is set")
+                    != *new_session_number
+            });
+
+            if !should_broadcast {
+                return;
+            }
```
zeegomo: this returns even if other services have updates?

nomos-core/chain-defs/src/mantle/genesis_tx.rs
```diff
@@ -60,7 +63,26 @@ impl GenesisTx {
...
+
+        if num_ops != num_proofs {
+            return Err(Error::UnequalNumberOfProofs {
+                num_ops,
+                num_proofs,
+            });
+        }
+
+        self.0.ops_proofs = ops_proofs;
+        Ok(self)
```
zeegomo: nit: couldn't we get the proofs directly from the `SignedMantleTx` in `from_tx`?

nomos-core/chain-defs/src/mantle/genesis_tx.rs
```diff
@@ -105,14 +127,32 @@ impl GasCost for GenesisTx {
...
     fn genesis_inscription(&self) -> &InscriptionOp {
         // Safe to unwrap because we validated this in from_tx
-        match &self.0.ops[0] {
+        match &self.mantle_tx().ops[0] {
             Op::ChannelInscribe(op) => op,
             _ => unreachable!("GenesisTx always has a valid inscription as first op"),
         }
     }
 
+    fn sdp_ops(&self) -> impl Iterator<Item = (&SDPDeclareOp, &OpProof)> {
```
zeegomo: you could call this declarations

nomos-core/chain-defs/src/mantle/mod.rs
```diff
@@ -56,7 +56,9 @@ pub trait AuthenticatedMantleTx: Transaction<Hash = TxHash> + GasCost {
 //  https://www.notion.so/nomos-tech/Bedrock-Genesis-Block-21d261aa09df80bb8dc3c768802eb527?d=27a261aa09df808e9c66001cf0585dee
 pub trait GenesisTx: Transaction<Hash = TxHash> {
     fn genesis_inscription(&self) -> &InscriptionOp;
+    fn sdp_ops(&self) -> impl Iterator<Item = (&SDPDeclareOp, &OpProof)>;
     fn mantle_tx(&self) -> &MantleTx;
+    fn op_proofs(&self) -> &Vec<OpProof>;
```
zeegomo: are these all needed?

ledger/nomos-ledger/src/mantle/mod.rs
```diff
@@ -61,10 +63,15 @@ impl LedgerState {
         config: &Config,
         utxo_tree: &UtxoTree,
     ) -> Result<Self, Error> {
-        let tx_hash = tx.hash();
-        let ops = tx.mantle_tx().ops.iter().map(|op| (op, None));
-        let (ledger, _) = Self::new().try_apply_ops(0, config, utxo_tree, tx_hash, ops)?;
-        Ok(ledger)
+        let channels = channel::Channels::from_genesis(tx.genesis_inscription())?;
+        let sdp =
+            sdp::SdpLedger::from_genesis(&config.sdp_config, utxo_tree, tx.hash(), tx.sdp_ops())?;
```
zeegomo: nit: we could fetch notes from the utxo in here like we do for declare ops

nomos-core/chain-defs/src/mantle/genesis_tx.rs
bacv: Updated.

nomos-core/chain-defs/src/mantle/genesis_tx.rs
bacv: Fixed.

nomos-core/chain-defs/src/mantle/mod.rs
bacv: Removed unused ops_proofs.

ledger/nomos-ledger/src/mantle/mod.rs
bacv: If we fetch notes here, then we'll still need to iterate over all ops when creating sdp ledger, now the note is fetched during that iteration.

zeegomo (approved):

PR #1833, opened 2025-10-23
feat(blend): delay shutting down core service when session transition ends

youngjoon-lee:

This is the 3rd PR for rewarding.

This is a temporary workaround. I've tried some other ways, but none of them was not perfect without refactoring (which may involve breaking changes against what Antonio's working on). So, I'd like to introduce this workaround for now, considering the dev time we have. 

### Background

As implemented in https://github.com/logos-co/nomos/pull/1822, the core service computes and submits an activity message when the session transition period has passed. It's good.
But, the upper layer (`instance.rs`) shuts down the core service as soon as the session transition period ends (when it's in `EdgeAfterCore` or `BroadcastAfterCore` mode), and switches to other modes (`Edge` or `Broadcast`). Then, the core service may not have time to detect the end of session transition before shutdown. In this case, no activity token will be submitted.

### An workaround

In this PR, we simply sleep 500ms before shutting down the core service, to give it time to detect the end of session transition and to compute/submit an activity proof.
This approach is not perfect, but is maybe the simplest one we can take right now. 
I have more sophisticated ideas but want to discuss about them before making breaking changes. I left my idea in a TODO comment. Let's talk in Discord.

Yes

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/instance.rs
+++ b/nomos-services/blend/src/instance.rs
@@ -239,17 +239,14 @@ where
 
     /// Handles the expiration of the transition period,
     /// by shutting down the previous Core mode if exists.
-    // TODO: Before shutting down the previous Core mode, notify it to submit
-    // activity proof in case it terminates before detecting the end of the
-    // transition period.
     async fn handle_transition_period_expired(self) -> Self {
         match self {
             Self::EdgeAfterCore { mode, prev } => {
-                prev.shutdown().await;
+                prev.graceful_shutdown().await;
                 Self::Edge(mode)
             }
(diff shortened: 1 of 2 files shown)


youngjoon-lee:
Closing this workaround PR.

Discussed with Antonio to go with my 3rd option in the branch [be4260](https://github.com/logos-co/nomos/commit/be42604b4a27a0b1eba6879eac714591a84dc320), which keeps `instance.rs` and let it signal the core service manually before shutting it down. I will polish the branch and open a PR soon.

PR #1837, opened 2025-10-23
feat(sdp): Implement signing and posting sdp messages txs

pradovic:

This PR adds a logic to create signed mantle transactions from sdp messages and post them to mempool.
Both wallet and mempool are currently mocked.

Yes

@pradovic 
@bacv 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/sdp/src/lib.rs
+++ b/nomos-services/sdp/src/lib.rs
@@ -77,6 +77,7 @@ pub enum SdpMessage {
     },
     PostDeclaration {
         declaration: Box<DeclarationMessage>,
+        reply_channel: oneshot::Sender<Result<DeclarationId, DynError>>,
     },
     PostActivity {
         metadata: ActivityMetadata, // DA/Blend specific metadata
@@ -155,9 +156,17 @@ where
                     self.handle_post_activity(metadata, &wallet_adapter, &mempool_adapter)
                         .await;
                 }
-                SdpMessage::PostDeclaration { declaration } => {
-                    self.handle_post_declaration(declaration, &wallet_adapter, &mempool_adapter)
-                        .await;
+                SdpMessage::PostDeclaration {
+                    declaration,
+                    reply_channel,
+                } => {
+                    self.handle_post_declaration(
+                        declaration,
+                        &wallet_adapter,
+                        &mempool_adapter,
+                        reply_channel,
+                    )
+                    .await;
                 }
                 SdpMessage::PostWithdrawal { declaration_id } => {
                     self.handle_post_withdrawal(declaration_id, &wallet_adapter, &mempool_adapter)
@@ -176,17 +185,15 @@ where
     RuntimeServiceId: AsServiceId<Self> + Clone + Display + Send + Sync + 'static,
 {
     async fn handle_post_declaration(
-        &mut self,
+        &self,
         declaration: Box<DeclarationMessage>,
         wallet_adapter: &MockWalletAdapter,
         mempool_adapter: &MockMempoolAdapter,
+        reply_channel: oneshot::Sender<Result<DeclarationId, DynError>>,
     ) {
-        self.update_declaration(declaration.as_ref());
-        self.nonce = 0;
-
         let tx_builder = MantleTxBuilder::new();
 
-        let signed_tx = match wallet_adapter.declare_tx(tx_builder, declaration) {
+        let signed_tx = match wallet_adapter.declare_tx(tx_builder, declaration.clone()) {
             Ok(tx) => tx,
             Err(e) => {
                 tracing::error!("Failed to create declaration transaction: {:?}", e);
@@ -199,7 +206,9 @@ where
             return;
         }
 
-        tracing::info!("Declaration posted successfully");
+        if let Err(e) = reply_channel.send(Ok(declaration.id())) {
+            tracing::error!("Failed to send post declaration response: {:?}", e);
+        }
     }
 
     async fn handle_post_activity(
@@ -298,16 +307,6 @@ where
         }
 
         Ok(())
-    }
-
-    fn update_declaration(&mut self, declaration: &DeclarationMessage) {
-        let new_declaration = Declaration {
-            id: declaration.id(),
-            zk_id: declaration.zk_id,
-            locked_note_id: declaration.locked_note_id,
-        };
-
-        self.current_declaration = Some(new_declaration);
     }
 }
 


nomos-services/sdp/src/adapters/mempool/mock.rs
```diff
@@ -0,0 +1,26 @@
...
+pub struct MockMempoolAdapter;
+
+#[async_trait::async_trait]
+impl SdpMempoolAdapter for MockMempoolAdapter {
+    fn new() -> Self {
+        Self {}
+    }
+
+    async fn post_tx(&self, tx: SignedMantleTx) -> Result<(), DynError> {
+        // TODO: enable when membership service is deleted and we can depend on
```
pradovic, line 15: @bacv Does this makes sense? I was getting a circular dependency when trying to implement a real adapter. I think because of tx_service->da-sampling -> da-network -> membership -> sdp, and now we have da-network -> sdp. 

But it looks like this will be ok when membership service is removed, and then we can add the real adapter implementation.

nomos-core/chain-defs/src/sdp/mod.rs
```diff
@@ -263,7 +265,7 @@ impl WithdrawMessage {
 pub struct ActiveMessage {
     pub declaration_id: DeclarationId,
     pub nonce: Nonce,
-    pub metadata: Option<ActivityMetadata>,
+    pub metadata: ActivityMetadata,
```
davidrusu, line 268: 🥇

davidrusu (approved):
🚤

nomos-services/sdp/src/adapters/mempool/mock.rs
bacv: I've seen this, this is the best approach, once the sdp in genesis is merged our first task will be to remove membership service and update this part.

nomos-services/sdp/src/adapters/wallet/mock.rs
```diff
@@ -0,0 +1,108 @@
...
+    fn new() -> Self {
+        Self {}
+    }
+    fn declare_tx(
+        &self,
+        tx_builder: MantleTxBuilder,
+        declaration: Box<DeclarationMessage>,
+    ) -> Result<SignedMantleTx, Self::Error> {
+        // todo: this is for mock, we need signing key in production
+        let signing_key = SigningKey::from_bytes(&[0u8; 32]);
```
danielSanchezQ, line 26: Where do we hold the pk for this signing key? in #1840 @davidrusu prepared so we can sign stuff.

nomos-services/sdp/src/adapters/wallet/mock.rs
```diff
@@ -0,0 +1,108 @@
...
+    mantle::{NoteId, Op, OpProof, SignedMantleTx, Transaction as _, tx_builder::MantleTxBuilder},
+    proofs::zksig::{DummyZkSignature, ZkSignaturePublic},
+    sdp::{ActiveMessage, DeclarationMessage, WithdrawMessage, ZkPublicKey},
+};
+
+use crate::adapters::wallet::SdpWalletAdapter;
+
+pub struct MockWalletAdapter;
+
+impl SdpWalletAdapter for MockWalletAdapter {
```
danielSanchezQ, line 14: Why is this an adapter if it only creates the txs?

nomos-services/sdp/src/adapters/wallet/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+    mantle::{NoteId, SignedMantleTx, tx_builder::MantleTxBuilder},
+    sdp::{ActiveMessage, DeclarationMessage, WithdrawMessage, ZkPublicKey},
+};
+
+#[async_trait::async_trait]
+pub trait SdpWalletAdapter {
+    type Error;
+
+    // TODO: Pass relay when wallet service is defined.
+    fn new() -> Self;
```
danielSanchezQ, line 13: You can get the overwatch handle already here. Then the implementor decides what to relay to.

nomos-services/sdp/src/lib.rs
```diff
@@ -133,17 +153,163 @@ where
...
+    ) {
+        let tx_builder = MantleTxBuilder::new();
+
+        let signed_tx = match wallet_adapter.declare_tx(tx_builder, declaration.clone()) {
+            Ok(tx) => tx,
+            Err(e) => {
+                tracing::error!("Failed to create declaration transaction: {:?}", e);
+                return;
+            }
+        };
```
danielSanchezQ, line 202:
```suggestion
        let Ok(signed_tx) = wallet_adapter.declare_tx(tx_builder, declaration.clone()) else {
                tracing::error!("Failed to create declaration transaction: {:?}", e);
                return;
        };
```

nomos-services/sdp/src/lib.rs
```diff
@@ -133,17 +153,163 @@ where
...
+        let declaration = self.current_declaration.as_ref().unwrap();
+
+        let signed_tx =
+            match wallet_adapter.active_tx(tx_builder, active_message, declaration.zk_id) {
+                Ok(tx) => tx,
+                Err(e) => {
+                    tracing::error!("Failed to create activity transaction: {:?}", e);
+                    return;
+                }
+            };
```
danielSanchezQ, line 246:
```suggestion
        let Ok(signed_tx) = wallet_adapter.active_tx(tx_builder, active_message, declaration.zk_id) else {
                tracing::error!("Failed to create activity transaction: {:?}", e);
                return;
            };
```

nomos-services/sdp/src/lib.rs
```diff
@@ -133,17 +153,163 @@ where
...
+            withdraw_message,
+            declaration.zk_id,
+            declaration.locked_note_id,
+        ) {
+            Ok(tx) => tx,
+            Err(e) => {
+                tracing::error!("Failed to create withdrawal transaction: {:?}", e);
+                return;
+            }
+        };
```
danielSanchezQ, line 287: let/else too

nomos-services/sdp/src/lib.rs
```diff
@@ -133,17 +153,163 @@ where
...
+
+        if let Err(e) = mempool_adapter.post_tx(signed_tx).await {
+            tracing::error!("Failed to post withdrawal to mempool: {:?}", e);
+            return;
+        }
```
danielSanchezQ, line 292: I'm curious why sometimes you use match (ok retriveal) sometimes let/else patterns (err retrieval) 😅

nomos-services/sdp/src/adapters/wallet/mock.rs
pradovic: Is this going to be from KMS?

nomos-services/sdp/src/adapters/wallet/mock.rs
pradovic: If I understood correctly tx_builder creates transaction, but it reality it will be wallet API when it is implemented. Does this makes sense?

nomos-services/sdp/src/adapters/wallet/mod.rs
pradovic: I copied the same approach from dispersal, where we will provide a handle here when the actual service is ready. Does that makes sense?

nomos-services/sdp/src/lib.rs
pradovic: We would swallow the error in this case then, right?

pradovic: Not super important, but might be useful as part of the log

nomos-services/sdp/src/lib.rs
pradovic: Same as the above, let me know what you think and I can change both if we don't need to print an error

nomos-services/sdp/src/lib.rs
pradovic: I use 3 cases:
1. No error needed : let else pattern
2. To print an error and assign return value I use match
3. and to just log error and without return value I use if let Err(e)


I would use map error, inspect error, but I needed to return in error case.

Does that makes sense?

Issue #1835, opened 2025-10-23
Reject blocks from the future


davidrusu:
Currently chain-service does not reject blocks from the future, which means that if a leader sees he's a winner 4 days from now, he can build the PoL for that slot and send it now and have it accepted by the network.

see block validation logic in the spec
https://www.notion.so/nomos-tech/Cryptarchia-v1-Protocol-Specification-21c261aa09df810cb85eff1c76e5798c?source=copy_link#21c261aa09df81ffaf4ff3168252a097
<img width="591" height="99" alt="Image" src="https://github.com/user-attachments/assets/65cbae40-db41-449b-918d-1bf48664ff36" />

PR #1839, opened 2025-10-24
feat(chain): conditionally enable blob validation by block slot

youngjoon-lee:

Closes https://github.com/logos-co/nomos/issues/1675

As we discussed, this PR implements enabling blob validation during bootstrapping.

Now we decide on whether to perform blob validation or not for each block, by comparing `block.slot` with `current_slot`.
If `(current_slot - block.slot) > blob_validation_window`, we skip blob validation because DA nodes may already prune the blobs.
This logic is applied to the `ChainService::process_block`, so that we can use this logic for all scenarios.

The `current_slot` is retrieved from the time service. 

The `blob_validation_window` is currently `0.5 * k/f`. You can find the rationale from https://github.com/logos-co/nomos/issues/1675.

Also, this PR introduces two strategies for blob validation. One is `Recent` (that we've already had), and the other is `Historic`. You can find details from their doc comments.

I believe so.

@youngjoon-lee @zeegomo 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/consensus/cryptarchia-engine/src/time.rs
+++ b/consensus/cryptarchia-engine/src/time.rs
@@ -78,6 +78,11 @@ impl Slot {
             )
         }
     }
+
+    #[must_use]
+    pub const fn saturating_sub(self, rhs: Self) -> Self {
+        Self(self.0.saturating_sub(rhs.0))
+    }
 }
 
 impl From<u32> for Epoch {
--- a/nomos-services/chain/chain-service/src/blob.rs
+++ b/nomos-services/chain/chain-service/src/blob.rs
@@ -1,46 +1,111 @@
-use std::{collections::BTreeSet, fmt::Debug};
+use std::{collections::BTreeSet, fmt::Debug, marker::PhantomData, num::NonZero};
 
+use cryptarchia_engine::Slot;
 use nomos_core::{
     block::Block,
     da,
     mantle::{AuthenticatedMantleTx, Op},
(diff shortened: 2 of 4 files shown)


danielSanchezQ:
> I think using a tick stream here is not the correct approach since we are not acting on new slot events, we are acting on new blocks events. Can we instead query the time service for the current slot when we receive a block?
> 
> i.e. we can add a new `CurrentSlot` message to the time service:
> 
> ```rust
> pub enum TimeServiceMessage {
>     Subscribe {
>         sender: oneshot::Sender<EpochSlotTickStream>,
>     },
>     CurrentSlot { // add this message to time service
>         sender: oneshot::Sender<SlotTick>,
>     }
>  }
> ```
> 
> That returns the current slot, and then when we receive a new block, we can just request the current slot from the time service.
> 
> We will need a similar flow in the chain-service in order to reject blocks from the future (see #1835). We can centralize time keeping the nomos-time service so that we don't need to keep track of current slot at each caller.

This seems a good idea and easy to do. Then we map the block stream with this current slot call to get the info together.

youngjoon-lee:
@davidrusu @danielSanchezQ Yeah, thank you. I thought about the idea, but I didn't go with it because there are some things to be improved in the time service to support the API. But I agree that having the additional API is a better approach. I will implement it and re-request your review soon.

PR #1842, opened 2025-10-27
feat(time): add CurrentSlot message

youngjoon-lee:

This PR adds a new API `CurrentSlot` to the time service.

It will be useful for the chain service that needs to get the current slot whenever processing a block (to determine whether to validate blobs). I tried another approach for it in the PR #1839 but we discussed that it's better to have the `CurrentSlot` API in the time service. 

This PR is simple, but I left two comments below to help reviewers.

I believe so.

@youngjoon-lee  @danielSanchezQ

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/time/src/backends/ntp/mod.rs
+++ b/nomos-services/time/src/backends/ntp/mod.rs
@@ -60,7 +60,7 @@ impl TimeBackend for NtpTimeBackend {
         Self { settings, client }
     }
 
-    fn tick_stream(self) -> EpochSlotTickStream {
+    fn tick_stream(self) -> (SlotTick, EpochSlotTickStream) {
         let Self { settings, client } = self;
         let mut update_interval = interval(settings.update_interval);
         // if we miss a tick just try next one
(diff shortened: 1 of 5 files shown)


nomos-services/time/src/backends/ntp/mod.rs
```diff
@@ -60,7 +60,7 @@ impl TimeBackend for NtpTimeBackend {
         Self { settings, client }
     }
 
-    fn tick_stream(self) -> EpochSlotTickStream {
+    fn tick_stream(self) -> (SlotTick, EpochSlotTickStream) {
```
youngjoon-lee, line 63: At the beginning, I considered changing the `EpochSlotTickStream` to yield the first (current) slot tick immediately. But, I realized that it's more explicit to return the current tick and a stream of future ticks separately. Also, by doing so, we don't need to change any logic depending on the behavior of `EpochSlotTickStream`.

I wrote a doc comment for this function to the `TimeBackend` trait definition.

You may be curious why we need the current slot tick. You can find the reason in another PR comment I left in the `TimeService::run`: https://github.com/logos-co/nomos/pull/1842#discussion_r2464418753.

nomos-services/time/src/lib.rs
```diff
@@ -119,24 +123,10 @@ where
...
-                            let stream = Pin::new(Box::new(channel_stream));
-                            if let Err(_e) = sender.send(stream) {
-                                error!("Error subscribing to time event: Couldn't send back a response");
-                            }
-                        }
-                    }
+                    handle_service_message(service_message, &broadcast_receiver, &current_slot_tick);
                 }
                 Some(slot_tick) = tick_stream.next() => {
+                    current_slot_tick = slot_tick;
```
youngjoon-lee, line 129: I decided to maintain the `current_slot_tick` in this layer (not inside each backend). I tried the other way but realized that this way is much simpler/cleaner. For example, it's not simple to maintain the `current_slot_tick` in the NTP backend because it calibrates the current time and replace the stream of future ticks periodically.

PR #1846, opened 2025-10-27
feat(blend): notify core service of the end of session transition before shutdown

youngjoon-lee:

This PR handles a corner case in which the core service is shut down by the proxy service when the session transition period has passed, in the following transition cases.
- `EdgeAfterCore` -> `Edge`
- `BroadcastAfterCore` -> `Core`

In these cases, the core service is shut down before it detects a `SessionTransitionPeriodExpired` event. Then, the blending tokens collected so far for the fading-out session will be discarded.

To handle these cases, this PR introduces another message type `FinishSessionTransition` only for CoreService. When the proxy service handles those transition cases, it first notifies the core service with the `FinishSessionTransition` message and waits until it is completely processed. After that, it finally shuts down the core service.

We've discussed other approaches, such as introducing graceful shutdown to Overwatch, or so. But, we decided to have this approach, since this features is currently only for Blend.

I believe so.

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.


nomos-services/blend/src/message.rs
```diff
@@ -39,3 +39,20 @@ impl<BroadcastSettings> From<EncapsulatedMessage> for ProcessedMessage<Broadcast
...
+
+impl<BroadcastSettings> MessageComponents for ServiceMessage<BroadcastSettings> {
+    type Payload = Vec<u8>;
+    type BroadcastSettings = BroadcastSettings;
+
+    fn into_components(self) -> (Self::Payload, Self::BroadcastSettings) {
+        let Self::Blend(network_message) = self;
+        (network_message.message, network_message.broadcast_settings)
+    }
+}
```
youngjoon-lee, line 58: Moved from `nomos-services/blend/src/core/service_components.rs`.

Issue #1847, opened 2025-10-27
Don't start with full core quota if node starts mid-session


ntn-x2:
Upon starting up, check if the last seen session is the same as the current session, and set the core quota respectively, to avoid generating more messages than expected for the session, and be marked as malicious from peers

PR #1851, opened 2025-10-27
feat(blend): recover spent core quota for same-session recovery

ntn-x2:

Fixes https://github.com/logos-co/nomos/issues/1847. This PR also introduces the components needed for Blend core service to implement recovery. More will be added to these components as we add more things to recovery.

For this PR, we only track the used core quota for the ongoing session, so that if the service crashes or if the node goes down, the node won't end up generating more cover messages than the other nodes expect, else it would be flagged as malicious and banned from the Blend p2p network.

**EDIT: Due to some limitations in the Overwatch design for state recovery (specifically that the state must define its own `Settings` type to match the service settings, instead of requiring something like `AsRef` to only get the components it needs), I decided to remove the generic recovery backend from the service, since it would not be possible to declare a generic recovery backend with generic configs without running into a lot of compilation issues. Hence, I have made the core service aware of how service recovery is implemented, for now, and its config reflects that, since it takes a path, which would not make sense if we were to change the recovery backend to something that is not file-based.**

Yes.

@ntn-x2 dev

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/core/mod.rs
+++ b/nomos-services/blend/src/core/mod.rs
@@ -47,7 +47,10 @@ use overwatch::{
 };
 use rand::{RngCore, SeedableRng as _, seq::SliceRandom as _};
 use serde::{Deserialize, Serialize};
-use services_utils::{overwatch::RecoveryOperator, wait_until_services_are_ready};
+use services_utils::{
+    overwatch::{JsonFileBackend, RecoveryOperator},
+    wait_until_services_are_ready,
+};
 use tokio::sync::oneshot;
 use tracing::info;
 
@@ -99,7 +102,6 @@ pub struct BlendService<
     TimeBackend,
     ChainService,
     PolInfoProvider,
-    StateRecoveryBackend,
     RuntimeServiceId,
 > where
     Backend: BlendBackend<NodeId, BlakeRng, ProofsVerifier, RuntimeServiceId>,
@@ -113,7 +115,6 @@ pub struct BlendService<
         ProofsGenerator,
         TimeBackend,
         ChainService,
-        StateRecoveryBackend,
         PolInfoProvider,
     )>,
 }
@@ -128,7 +129,6 @@ impl<
     TimeBackend,
     ChainService,
     PolInfoProvider,
-    StateRecoveryBackend,
     RuntimeServiceId,
 > ServiceData
     for BlendService<
@@ -141,7 +141,6 @@ impl<
         TimeBackend,
         ChainService,
         PolInfoProvider,
-        StateRecoveryBackend,
         RuntimeServiceId,
     >
 where
@@ -150,7 +149,9 @@ where
 {
     type Settings = BlendConfig<Backend::Settings>;
     type State = RecoveryServiceState<Backend::Settings>;
-    type StateOperator = RecoveryOperator<StateRecoveryBackend>;
+    type StateOperator = RecoveryOperator<
+        JsonFileBackend<RecoveryServiceState<Backend::Settings>, BlendConfig<Backend::Settings>>,
+    >;
     type Message = ServiceMessage<Network::BroadcastSettings>;
 }
 
@@ -165,7 +166,6 @@ impl<
     TimeBackend,
     ChainService,
     PolInfoProvider,
-    StateRecoveryBackend,
     RuntimeServiceId,
 > ServiceCore<RuntimeServiceId>
     for BlendService<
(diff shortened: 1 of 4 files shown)


nomos-services/blend/src/core/mod.rs
```diff
@@ -387,10 +398,24 @@ where
             .expect("Reward session info must be created successfully. Panicking since the service cannot continue with this session")
         );
 
+        // Initialize the current session state. If the session matches the stored one,
+        // retrieves the tracked consumed core quota. Else, fallback to `0`.
+        let mut current_recovery_checkpoint = if let Some(saved_state) = last_saved_state
+            && saved_state.last_seen_session() == current_membership_info.public.session
```
youngjoon-lee, line 407: 👍

nomos-services/blend/src/core/scheduler.rs
```diff
@@ -0,0 +1,267 @@
...
+        match self.scheduler.poll_next_unpin(cx) {
+            Poll::Ready(Some(round_info)) => {
+                // Consumed quota for this round equals the number of blended data messages +
+                // the number of cover messages (which at the moment can be either `0` or `1`).
+                let total_consumed_quota = self
+                    .release_round_consumed_quota
+                    .checked_add(u64::from(
+                        round_info.cover_message_generation_flag.is_some(),
+                    ))
+                    .expect("Overflow when computing total consumed quota for release round.");
```
youngjoon-lee, line 107: > + the number of cover messages (which at the moment can be either `0` or `1`).

Do we probably need to remove this assumption in this module? The cover messages are generated by the upper layer, and this module doesn't know how many cover messages the upper layer will generate. What do you think about ~~chaining~~changing this flag to `usize`? Please correct me if my concern is not so meaningful. Maybe, this code is right if this assumption is part of the definition of message scheduler.

youngjoon-lee (commented):
Looks good. I left only one comment before approving.

nomos-services/blend/src/core/scheduler.rs
ntn-x2: Sorry, I am not sure I fully understand your comment. The `RoundInfo` the scheduler returns includes an `Option<()>` to signal whether a new cover message should be generated in this round, so this service knows exactly what to do with this flag. If this assumption were to change, we would change `Option<()>` to a `usize` as you suggested. But at the moment, each release round, as per the spec, includes the processed messages that have been delayed since the last release round and optionally a cover message, if it was scheduled to do so and no data message was blended instead.

youngjoon-lee: Sorry, there was a typo in my comment.
Yeah, I understand it. This scheduler returns `Some`, expecting that a single cover message will be generated. And, I checked, only one cover message is generated by the caller.

I just recalled the `message_frequency_per_round` parameter when I wrote this comment. If the value is > 1.0, and if there's no processed message scheduled, I guess the node should generate more than 1 cover messages in a single round. I just check our code again, and it seems that the parameter is only used to calculate `core_quota`. So, although this may be out of scope of this PR, we probably need to think about using the parameter in the scheduler. What do you think?

ntn-x2: As you say, the `message_frequency_per_round` parameter influences the total core quota. Meaning that a higher value will result in more release rounds containing a cover message. But, for each release round, there can be at most one cover message generated, and that is by design.

ntn-x2: https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df8121980bc14722d77fd8 states that `C = S * F_c`, but it does not mention anything about generating more than a single cover message per message round.

youngjoon-lee: I think `C` implies that more than one message per round **can** be generated. Also by definition of `F_c`:
https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81b49433ddf5ee65e025
> `F_c` denote a frequency at which cover messages are generated per round;

youngjoon-lee: In the current codebase, this assumption (design) is uniformly applied to all modules.  So, we're fine, I think.  I will resolve this thread, but will think about whether we need to use the param in the scheduler.

youngjoon-lee: I was confused. Now I understand that `F_c` is the expected frequency of cover message per round generated by **all the core nodes**.

Issue #1849, opened 2025-10-27
Recovery process for unprocessed messages


ntn-x2:
Two types of messages need to be recovered: messages that are to be blended, and messages that have been decapsulated.

In both cases, if the node is not able to send them because of a crash, the next time it starts it needs to recover them.

For data messages, a fresh leadership `PoQ` should be used. Both message types will be released at the next release round.

PR #1857, opened 2025-10-28
feat(blend): recover processed but unsent messages for same-session recovery

ntn-x2:

Fixes https://github.com/logos-co/nomos/issues/1849. This PR is built on top of https://github.com/logos-co/nomos/pull/1851, which introduces basic recovery support for Blend core service.

This PR does two things:

1. It adds support for recovering processed messages that have been randomly delayed, in case the service or the node shuts down before they are scheduled for release, and
2. Improve the recovery logic using a type-safe approach that allows us to batch multiple changes together before sending them to the recovery backend to store on file. It works by implementing an API that goes from state -> state updater to apply changes, and then state updater -> state to commit those changes to file and return the latest state. This should provide quite a solid ground to add more information to the service recovery.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/core/state.rs
+++ b/nomos-services/blend/src/core/state.rs
@@ -252,8 +252,8 @@ mod state_updater {
             self.inner.spend_quota(amount);
         }
 
-        /// Consumes `self` and return the **changed** state without storing the
-        /// changes via the underlying
+        /// Consumes `self` and return the state with any changes applied to it,
+        /// without storing those changes via the underlying
         /// `overwatch::services::state::StateUpdater`.
         ///
         /// It is important to note that it is not equivalent to calling


nomos-services/blend/src/core/mod.rs
```diff
@@ -723,7 +760,14 @@ fn handle_incoming_blend_message<
...
             blending_token_collector.insert(blending_token);
-            scheduler.schedule_message(remaining_encapsulated_message.into());
+            let processed_message = ProcessedMessage::from(remaining_encapsulated_message);
+            scheduler.schedule_message(processed_message.clone());
+            assert_eq!(
+                state_updater.add_unsent_message(processed_message),
+                Ok(()),
+                "There should not be another copy of the same encapsulated message already processed."
+            );
+            state_updater.commit_changes()
```
youngjoon-lee, line 773: Just asking what you mentioned in the PR description: 
> Improve the recovery logic using a type-safe approach that allows us to batch multiple changes together before sending them to the recovery backend to store on file. 

If I didn't miss anything, we don't batch changes in the current codebase, but we can later when necessary, right? It seems that we currently call `commit_changes` every time we mutate the state.

nomos-services/blend/src/core/state.rs
```diff
@@ -1,69 +1,382 @@
...
+
+    #[derive(Clone)]
+    /// Recovery state for Blend core service.
+    pub struct ServiceState<BackendSettings, BroadcastSettings> {
+        /// The last session that was saved.
+        last_seen_session: u64,
+        /// The last value for the core quota allowance for the session that is
+        /// tracked.
+        spent_core_quota: u64,
+        unsent_processed_messages: HashSet<ProcessedMessage<BroadcastSettings>>,
```
youngjoon-lee, line 79: Could we probably replace these three fields with the `SerializableServiceState`?

nomos-services/blend/src/core/state.rs
```diff
@@ -1,69 +1,382 @@
...
+        pub fn into_inner(self) -> ServiceState<BackendSettings, BroadcastSettings> {
+            self.inner
+        }
+
+        pub const fn consume_core_quota(&mut self, amount: u64) {
+            self.changed = true;
+            self.inner.spend_quota(amount);
+        }
+
+        /// Consumes `self` and return the **changed** state without storing the
```
youngjoon-lee: "maybe changed"? I guess the state can remain the same if the user didn't do anything with it? (as reflected in `self.changed`).

youngjoon-lee (commented):
Good design. Left only few comments not critical before approving.

nomos-services/blend/src/core/mod.rs
ntn-x2: The whole point of recovery is to be safe in case of sudden crashes shutdowns, so we can't postpone too much else we risk losing info. What I mean by batching here is that you take a state, you derive an object to make changes on it, and when you are done with the changes you `commit` them by interacting with the overwatch `StateUpdater`. So you don't store on every single change, but you decide when to do so, and you don't have to interact with the `StateUpdater` component directly, but you instead call a method on this updater object. For instance, when we process previously delayed messages, we remove each message from the "overlay" when we transform them into a future (https://github.com/logos-co/nomos/pull/1857/files#diff-d0b3b76bceb9f74b9327e52d486faa6e49714f4a9da31a0cc5633b5d72673b53R822), but we write the changes only once at the end, after we also account for all the consumed core quota at this round (https://github.com/logos-co/nomos/pull/1857/files#diff-d0b3b76bceb9f74b9327e52d486faa6e49714f4a9da31a0cc5633b5d72673b53R858). This is what I meant by batching updates. Was anything unclear in the current version of this PR?

nomos-services/blend/src/core/state.rs
ntn-x2: Yeah I thought about it, I think we could do it, although the `SerializableServiceState` type might in future hold values that are only needed for (de)serialization, that we don't need in the actual state.

nomos-services/blend/src/core/state.rs
ntn-x2: Yeah I can update the comment, good point.

ntn-x2: https://github.com/logos-co/nomos/pull/1857/commits/89ecf0b80e982303dd151e135b17df980ead9f54

nomos-services/blend/src/core/mod.rs
youngjoon-lee: Aha, yeah, I got it. Thank you. I missed the `state_updater.remove_sent_message`. Yeah, it makes sense and useful.

Issue #1855, opened 2025-10-28
Core service terminates by itself before transition is done


youngjoon-lee:
Even though `instance.rs` maintains the previous `CoreMode` (as a form of `EdgeAfterCore` and `BroadcastAfterCore`), the core service is not maintained during the session transition period in the current codebase. The core service terminates by itself as soon as it receives a new session where it is not core anymore.

This is a bug. The core service shouldn’t be terminated until the last session transition is done. 

~~In the current codebase, this kind of bugs can be produced easily because `instance.rs` cannot check whether a certain service is running or was terminated.~~

~~There may be various ways to resolve this issue, but I think the main point is having transition logic in one place and not producing scenarios where services are terminated (panicked) silently.~~

~~I think the best approach would be redesigned the core and edge module without using Overwatch. Since they are currently Overwatch services, it is limited to manage their lifecycle and transition. We could design a nice state transition model, instead. I am preparing a PoC. I think it is time to try this redesigning, rather than making the core service more complicated.~~

~~In short, I am thinking about having three states: Core, Edge, and Broadcast. A node transitions its state only when its role is changed in a new session.
There is one special case: Transitioning from core. Before switching to the other state, we can spawn-and-forget a task which handles and finishes the last session transition period. We don’t need to define a dedicated state for the task because we don’t need to interact with it. 
This design can be also implemented with keeping the Overwatch services model, but I think it’s easier without Overwatch.~~

UPDATE:
We can fix this in a simple way. I will modify the core service to continue processing the previous session until the next STPExpired event is arrived. We can simply exit from the main event loop once we receive a new session where the node is not core anymore, and start another event loop only for processing the last session transition during only STP. After that, core service must terminate by itself.
This issue is also related to https://github.com/logos-co/nomos/issues/1860, but not directly. So, this can be resolved separately.

PR #1859, opened 2025-10-29
fix(blend): introduce mode system to fix session transition bugs

youngjoon-lee:

This is a draft. Code quality is not production-ready. This is only for progressing discussions with Antonio to fix the session transition and lifecycle bugs in a nice way.

Fortunately, the diff is quite easy to read, even though many codes have been deleted.

Please see the detailed background in https://github.com/logos-co/nomos/issues/1855.

Summary of changes:
- I redesigned core and edge module without using Overwatch services. Now they implement the `Mode` trait, not Overwatch.
- Now we have only one BlendService (= proxy).
- The BlendService runs a certain mode based on parameters.
  - If anything goes wrong in the mode, it returns an error to the BlendService. This is one of what we didn't have in the Overwatch model.
  - When a mode becomes useless in a new session, it terminates itself. Then, the BlendService checks the new membership and starts an appropriate mode. 
  - BlendService never terminate the mode running. It just waits until the `Mode::run` returns a `Result`, and runs a next mode based on the `Result`.
- When the core mode terminates itself (because it's not core in the new session), it spawns-and-forgets a task that handles/finishes the last session transition period. This fixes [#1855](https://github.com/logos-co/nomos/issues/1855).
- Previously, each service initialized streams and structs every time it starts. It was not bad, but required some validation checks and was highly dependent on the responsiveness of other services (e.g. FirstReadyStream). In this new design, all the initialization is done once by the BlendService. The initialization outputs are consolidated into a `Context`, which is passed and handed over to each mode continuously. Each mode can immediately get the current state from the `Context` and continue reading streams.
  - We can probably discuss about this. I'm fine even if we go back to the previous approach where the initialization is done by each mode every time it starts. It's still better than the Overwatch model, because each mode can immediately return an error to the BlendService if anything goes wrong during initialization. 

Again, this is a draft. Please check only how the new design looks, and check if there's any obstacles I missed.

You can find key changes below:
- [Changed `CoreService` to `CoreMode`.](https://github.com/logos-co/nomos/pull/1859/files#diff-d0b3b76bceb9f74b9327e52d486faa6e49714f4a9da31a0cc5633b5d72673b53R69)
- [Changed `EdgeService` to `EdgeMode`](https://github.com/logos-co/nomos/pull/1859/files#diff-9922e730ccc7ffde500676b55cfcbbb5305f409195be34cdc947a9c8245aba89R43)
- [Updated `BlendService` logic](https://github.com/logos-co/nomos/pull/1859/files#diff-3dd38b23c0c6564cbb709d5203965de8a58e28ea39ccb550fb5fdbc1c60782e5R405)
- [Defined `Mode` trait](https://github.com/logos-co/nomos/pull/1859/files#diff-e8a96a15703f7d5d07003bc2458c7159b47f0c5b47e2f1cdea4625ccb19ee296R4): There might be any better design, but this is what I ended up with, and I don't see any critical downsides.
- [Defined `Context`](https://github.com/logos-co/nomos/pull/1859/files#diff-aa0524d242b1bbccedfba3a757b11bdc90249bcd812ca9cdf3f5452d0cdd8267R19) 
- [Spawning `FinalSessoinTransitionTask`](https://github.com/logos-co/nomos/pull/1859/files#diff-d0b3b76bceb9f74b9327e52d486faa6e49714f4a9da31a0cc5633b5d72673b53R323): This could be a separate PR, but I wanted to show you the big picture in this proposal.

I believe this approach is way better than the previous. We could avoid potential bugs more easily and explicitly.

> [!IMPORTANT]  
> explained.  

> [!IMPORTANT]  
> advancing  
> the review.

> [!IMPORTANT]

> [!IMPORTANT]
    modification/solutions.

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/context.rs
+++ b/nomos-services/blend/src/context.rs
@@ -6,7 +6,7 @@ use nomos_time::SlotTick;
 use overwatch::overwatch::OverwatchHandle;
 
 use crate::{
-    epoch_info::{EpochHandler, LeaderInputsMinusQuota, PolEpochInfo},
+    epoch_info::{EpochEvent, EpochHandler, LeaderInputsMinusQuota, PolEpochInfo},
     membership::MembershipInfo,
     message::ServiceMessage,
 };
@@ -44,7 +44,8 @@ impl<NodeId, BroadcastSettings, ChainService, RuntimeServiceId>
     Context<NodeId, BroadcastSettings, ChainService, RuntimeServiceId>
 where
     NodeId: Clone,
-    ChainService: CryptarchiaServiceData,
+    ChainService: CryptarchiaServiceData<Tx: Send + Sync>,
+    RuntimeServiceId: Send + Sync,
 {
     /// Advances the context by waiting for the next item from any of the
     /// streams.
@@ -58,7 +59,16 @@ where
             }
             Some(clock_tick) = self.clock_stream.next() => {
                 self.current_clock = clock_tick;
-                Some(Event::ClockTick(clock_tick))
+
+                let maybe_epoch_event = self.epoch_handler.tick(clock_tick).await;
+                if let Some(EpochEvent::NewEpoch(leader_inputs_minus_quota) | EpochEvent::NewEpochAndOldEpochTransitionExpired(leader_inputs_minus_quota)) = maybe_epoch_event {
+                    self.current_leader_inputs_minus_quota = leader_inputs_minus_quota;
+                }
+
+                Some(Event::ClockTick {
+                    clock_tick,
+                    maybe_epoch_event,
+                })
             }
             Some(pol_info) = self.secret_pol_info_stream.next() => {
                 self.current_private_leader_info = pol_info.clone();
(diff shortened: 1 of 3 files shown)


nomos-services/blend/src/mode.rs
```diff
@@ -0,0 +1,15 @@
...
+#[async_trait]
+pub trait Mode<Context> {
+    async fn run(self: Box<Self>) -> Result<(ModeKind, Context), Error>;
+}
+
+pub enum ModeKind {
+    Core,
+    Edge,
+    Broadcast,
+}
```
youngjoon-lee, line 6:
```suggestion
#[async_trait]
pub trait Mode<Context> {
    async fn run(self: Box<Self>) -> Result<Context, Error>;
}
```
I'll get rid of `ModeKind`. The BlendService can choose a next mode based on the new membership.

youngjoon-lee: done ae2ca09ddbf2874a9f492a0861f14cb074082a54

youngjoon-lee:
I'll reopen this after refactoring, as discussed with Antonio.

PR #1862, opened 2025-10-29
feat(chain): block proposal combined PR

andrussal:

This PR combines already reviewed PRs. Unfortunately I had to create another PR to get all these stacked PRs to master.

Base branch PR that combines:

- https://github.com/logos-co/nomos/pull/1650
- https://github.com/logos-co/nomos/pull/1805
- https://github.com/logos-co/nomos/pull/1815

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-core/chain-defs/src/utils/mod.rs
+++ b/nomos-core/chain-defs/src/utils/mod.rs
@@ -39,7 +39,7 @@ macro_rules! serde_bytes_newtype {
                 D: serde::Deserializer<'de>,
             {
                 if deserializer.is_human_readable() {
-                    let s = <String as serde::Deserialize>::deserialize(deserializer)?;
+                    let s = <String>::deserialize(deserializer)?;
                     const_hex::decode_to_array(s)
                         .map(Self)
                         .map_err(serde::de::Error::custom)
--- a/nomos-services/chain/chain-service/src/lib.rs
+++ b/nomos-services/chain/chain-service/src/lib.rs
@@ -29,8 +29,10 @@ use nomos_core::{
     da::{self},
     header::HeaderId,
     mantle::{
-        AuthenticatedMantleTx, Transaction, TxHash, gas::MainnetGasConstants,
-        genesis_tx::GenesisTx, ops::leader_claim::VoucherCm,
+        AuthenticatedMantleTx, Transaction, TxHash,
+        gas::MainnetGasConstants,
+        genesis_tx::GenesisTx,
+        ops::{Op, leader_claim::VoucherCm},
     },
     sdp::{ProviderId, ProviderInfo, ServiceType},
 };
@@ -1086,6 +1088,7 @@ where
         relays: &CryptarchiaConsensusRelays<
             Mempool,
             MempoolNetAdapter,
+            MempoolDaAdapter,
             NetAdapter,
             SamplingBackend,
             Storage,
@@ -1168,6 +1171,7 @@ where
         relays: &CryptarchiaConsensusRelays<
             Mempool,
             MempoolNetAdapter,
+            MempoolDaAdapter,
             NetAdapter,
             SamplingBackend,
             Storage,
@@ -1287,11 +1291,21 @@ where
             .await
             .unwrap_or_else(|e| error!("Could not mark transactions in block: {e}"));
 
-        mark_blob_in_block(
-            relays.sampling_relay().clone(),
-            vec![], // TODO: pass actual blob ids here
-        )
-        .await;
+        let blob_ids: Vec<da::BlobId> = block
+            .transactions()
+            .flat_map(|tx| tx.mantle_tx().ops.iter())
+            .filter_map(|op| {
+                if let Op::ChannelBlob(blob_op) = op {
+                    Some(blob_op.blob)
+                } else {
+                    None
+                }
+            })
+            .collect();
+
+        if !blob_ids.is_empty() {
+            mark_blob_in_block(relays.sampling_relay().clone(), blob_ids).await;
+        }
 
         relays
             .storage_adapter()
--- a/nomos-services/tx-service/src/tx/service.rs
+++ b/nomos-services/tx-service/src/tx/service.rs
@@ -330,6 +330,7 @@ where
                     network_relay,
                     state_updater,
                     settings,
+                    tx_broadcast,
                 )
                 .await;
             }
@@ -373,6 +374,10 @@ where
         }
     }
 
+    #[expect(
+        clippy::too_many_arguments,
+        reason = "This helper needs both networking and storage handles; splitting it would obscure the call site."
+    )]
     async fn handle_add_message(
         pool: &mut Pool,
         key: Pool::Key,
@@ -381,6 +386,7 @@ where
         network_relay: OutboundRelay<BackendNetworkMsg<NetworkAdapter::Backend, RuntimeServiceId>>,
         state_updater: MempoolStateUpdater<Pool, NetworkAdapter, RuntimeServiceId>,
         settings: NetworkAdapter::Settings,
+        tx_broadcast: &broadcast::Sender<Pool::Item>,
     ) where
         Pool::Settings: Send + Sync,
         NetworkAdapter::Settings: Send + Sync,
@@ -396,11 +402,10 @@ where
                     network_relay,
                     item_for_broadcast,
                     reply_channel,
+                    tx_broadcast,
                 );
             }
-            Err(e) => {
-                Self::handle_add_error(e, reply_channel);
-            }
+            Err(e) => Self::handle_add_error(e, reply_channel),
         }
     }
 
@@ -485,13 +490,17 @@ where
         network_relay: OutboundRelay<BackendNetworkMsg<NetworkAdapter::Backend, RuntimeServiceId>>,
         item_for_broadcast: Pool::Item,
         reply_channel: oneshot::Sender<Result<(), MempoolError>>,
+        tx_broadcast: &broadcast::Sender<Pool::Item>,
     ) {
         state_updater.update(Some(<Pool as RecoverableMempool>::save(pool).into()));
 
+        let broadcast_clone = item_for_broadcast.clone();
         tokio::spawn(async move {
             let adapter = NetworkAdapter::new(settings, network_relay).await;
             adapter.send(item_for_broadcast).await;
         });
+
+        let _ = tx_broadcast.send(broadcast_clone);
 
         if let Err(e) = reply_channel.send(Ok(())) {
             tracing::debug!("Failed to send add reply: {:?}", e);
--- a/nomos-services/data-availability/dispersal/src/adapters/storage/mock.rs
+++ b/nomos-services/data-availability/dispersal/src/adapters/storage/mock.rs
@@ -1,9 +1,6 @@
 use std::collections::HashMap;
 
-use nomos_core::mantle::{
-    AuthenticatedMantleTx as _, Op, SignedMantleTx,
-    ops::channel::{ChannelId, MsgId},
-};
+use nomos_core::mantle::{SignedMantleTx, ops::channel::ChannelId};
 
 use super::DispersalStorageAdapter;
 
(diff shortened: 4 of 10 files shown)


nomos-core/chain-defs/src/utils/merkle.rs
```diff
@@ -13,3 +15,155 @@ impl<T> MerkleNode<T> {
 }
 
 pub type MerklePath<T> = Vec<MerkleNode<T>>;
+
+#[must_use]
+pub fn leaf(data: &[u8]) -> [u8; 32] {
+    let mut hasher = Hasher::new();
+    hasher.update(b"NOMOS_MERKLE_LEAF");
```
zeegomo: cc @thomaslavaur, iirc you said no separators?

thomaslavaur: Yes no separator for poseidon2 hash

thomaslavaur: But here it looks to be for Blake2b (because of [u8;32] ?) I would say no separator if it's not in the spec but it's less critical

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -1,32 +1,56 @@
...
-    pub mempool_transactions: Vec<Fr>, // 1024 - len(service_reward)
+    pub service_reward: Option<TxHash>,
+    pub mempool_transactions: Vec<TxHash>,
 }
 
-/// A block
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Block<Tx> {
     header: Header,
+    signature: ed25519_dalek::Signature,
```
danielSanchezQ, line 52: @andrussal @zeegomo isn't the signature in the header already?

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -41,19 +65,105 @@ impl Proposal {
...
+
+        let signature = header.sign(signing_key)?;
+
+        Ok(Self {
             header,
+            signature,
+            service_reward,
             transactions,
+        })
+    }
```
danielSanchezQ, line 116: This would have been better as a builder maybe?

nomos-core/chain-defs/src/block/mod.rs
```diff
@@ -41,19 +65,105 @@ impl Proposal {
...
+
+    pub fn recover(
```
danielSanchezQ: So this is for when we reconstruct it?

nomos-core/chain-defs/src/utils/mod.rs
```diff
@@ -39,12 +39,12 @@ macro_rules! serde_bytes_newtype {
                 D: serde::Deserializer<'de>,
             {
                 if deserializer.is_human_readable() {
-                    let s = <String>::deserialize(deserializer)?;
+                    let s = <String as serde::Deserialize>::deserialize(deserializer)?;
```
danielSanchezQ: Was this really needed? 🤔

nomos-services/chain/chain-leader/src/lib.rs
```diff
@@ -549,38 +560,106 @@ where
...
-
-                    async move { is_valid }
-                });
+        let (txs_stream, blobs) = match futures::try_join!(txs, blobs_ids) {
+            Ok(result) => result,
+            Err(e) => {
+                error!("Could not fetch block transactions: {e}");
+                return None;
+            }
+        };
```
danielSanchezQ:
```suggestion
        let Ok((txs_stream, blobs)) = futures::try_join!(txs, blobs_ids) else {
                error!("Could not fetch block transactions: {e}");
                return None;
         };
```

nomos-services/chain/chain-leader/src/lib.rs
```diff
@@ -549,38 +560,106 @@ where
...
+                });
 
-                let content_id = [0; 32].into(); // TODO: calculate the actual content id
-                // TODO: this should probably be a proposal or be transformed into a proposal
-                let block = Block::new(Header::new(parent, content_id, slot, proof), txs);
-                info!("proposed block with id {:?}", block.header().id());
-                output = Some(block);
+                future::ready(is_valid)
             }
+        });
```
danielSanchezQ: Why the nested `{}`?
```suggestion
        let filtered_stream = txs_stream.filter(
            move |tx| {
                let is_valid = tx.mantle_tx().ops.iter().all(|op| match op {
                    Op::ChannelBlob(op) => blobs.contains(&op.blob),
                    _ => true,
                });

                future::ready(is_valid)
            });
```

nomos-services/chain/chain-leader/src/lib.rs
```diff
@@ -549,38 +560,106 @@ where
...
-                error!("Could not fetch block transactions: {e}");
+                error!("Header validation/application failed: {e:?}");
+                return None;
+            }
+        };
+
+        let mut valid_txs = Vec::new();
+        let mut invalid_tx_hashes = Vec::new();
+
+        while let Some(tx) = tx_stream.next().await {
```
danielSanchezQ, line 611: I wonder how the order affects the validity of this. @andrussal @zeegomo ?

nomos-services/chain/chain-leader/src/lib.rs
```diff
@@ -549,38 +560,106 @@ where
...
+        // TODO: use PoL signing key
+        let dummy_signing_key = SigningKey::from_bytes(&[0u8; 32]);
+
+        let block = match Block::create(parent, slot, proof, txs, None, &dummy_signing_key) {
+            Ok(block) => block,
+            Err(e) => {
+                error!("Failed to create valid block during proposal: {e}");
+                return None;
+            }
+        };
```
danielSanchezQ:
```suggestion
        let Ok(block) = Block::create(parent, slot, proof, txs, None, &dummy_signing_key) else {
                error!("Failed to create valid block during proposal: {e}");
                return None;
        };
```

nomos-services/chain/chain-service/src/sync/block_provider.rs
```diff
@@ -752,7 +751,10 @@ mod tests {
 
             for i in 0..count {
                 let slot = Slot::from(slot_offset + i as u64);
-                let block = self.build_block_with_parent(prev_header, slot);
+                let Some(block) = self.build_block_with_parent(prev_header, slot) else {
+                    error!("Failed to build block with parent");
+                    break;
+                };
```
danielSanchezQ: Why didn't we used this pattern before 😅

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -752,24 +773,40 @@ where
...
+                                        error!(
+                                            target: LOG_TARGET,
+                                            "Received block with parent {:?} that is not in the ledger state. Ignoring block.",
+                                            parent
+                                        );
+                                    }
+                                    Err(e) => {
+                                        error!(target: LOG_TARGET, "Error processing reconstructed block: {:?}", e);
+                                    }
+                                }
```
danielSanchezQ: This is way too confusing to be honest. Its ok for now, but eventually it would be nice to have some meaningful functions for this.

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -1664,3 +1703,46 @@ async fn broadcast_da_session(
...
+        })?;
+
+    if !mempool_response.all_found() {
+        return Err(Error::InvalidBlock(format!(
+            "Failed to reconstruct block: {:?} mempool transactions not found",
+            mempool_response.not_found()
+        )));
+    }
+
+    // TODO: recover service reward
```
danielSanchezQ, line 1836: Do we have an issue tracking this?

nomos-services/tx-service/src/tx/service.rs
```diff
@@ -526,19 +531,37 @@ where
...
+            .map_err(|e| {
+                MempoolError::StorageError(format!("Failed to get items by keys: {e:?}"))
+            })?;
+
+        let found_transactions: Vec<Pool::Item> = items_stream.collect().await;
 
-        if let Err(_e) = reply_channel.send(transactions) {
-            tracing::debug!("Failed to send transactions reply");
+        if found_transactions.len() == keys_set.len() {
+            return Ok(crate::TransactionsByHashesResponse::new(
```
danielSanchezQ: why the fully qualified path on `TransactionByHashesResponse`? is there any similar named struct?

nomos-services/chain/chain-service/src/lib.rs
zeegomo: I'm working on this, might not even be needed with v1.2 of rewards, which I'm suggesting we adopt straight away

nomos-core/chain-defs/src/block/mod.rs
zeegomo: No, it's separate, you can see it in proposal here https://www.notion.so/nomos-tech/Block-Construction-Validation-and-Execution-Specification-215261aa09df8164b5a9f0783c0e99b9

andrussal: Yep, signature is outside of `Header`. I see this is also written this way in spec for Proposals. (I guess this way is probably easier as we sign the header content...).

danielSanchezQ: Thanks guys

nomos-core/chain-defs/src/block/mod.rs
andrussal: > So this is for when we reconstruct it?

right, probably should call it `reconstruct`.

nomos-services/chain/chain-leader/src/lib.rs
andrussal: I think I go with this instead because then we can log original error

```rust
 let (txs_stream, blobs) = futures::try_join!(txs, blobs_ids)
            .map_err(|e| {
                error!("Could not fetch block transactions: {e}");
                e
            })
            .ok()?;
```

andrussal: Actually I'll make this function to return Result so we can remove mapping errors here(and for the another comment below).

nomos-core/chain-defs/src/block/mod.rs
andrussal: > This would have been better as a builder maybe?

mh, I am not completely sure if it will improve readability. Builder would either need to keep track at each step if fields what current field depends on(for checks). Then this adds complexity. Or do all validation at build step. But then build step is very similar to this function. I am tempted to keep it like this. Do you have strong preference?

nomos-core/chain-defs/src/utils/mod.rs
andrussal: > Was this really needed? 🤔

hah, not sure how this sneaked in, probably experimented something.

nomos-services/chain/chain-leader/src/lib.rs
andrussal: > Why the nested `{}`?

Good catch!

nomos-core/chain-defs/src/block/mod.rs
danielSanchezQ: Lets keep it like this. I was actually mixing it with proposal.

nomos-services/chain/chain-leader/src/lib.rs
andrussal: > I wonder how the order affects the validity of this. @andrussal @zeegomo ?

I thinking that it will affect it for sure for some transactions, depending how users will send them. At some point we should probably add some deps checking between transactions and manage it internally.

nomos-services/chain/chain-service/src/sync/block_provider.rs
andrussal: > Why didn't we used this pattern before 😅

here i think at some point creating block started to return Result and then it affected this here. I am gonna change this just to unwrap in test, will be easier to see if it fails for any reason.

nomos-services/chain/chain-service/src/sync/block_provider.rs
```diff
@@ -751,10 +751,9 @@ mod tests {
 
             for i in 0..count {
                 let slot = Slot::from(slot_offset + i as u64);
-                let Some(block) = self.build_block_with_parent(prev_header, slot) else {
-                    error!("Failed to build block with parent");
-                    break;
-                };
+                let block = self
+                    .build_block_with_parent(prev_header, slot)
+                    .expect("Failed to build block with parent");
```
danielSanchezQ, line 756: Should this really be a panic?

PR #1863, opened 2025-10-29
feat(tx-service): subscribe to tx service from da-sampling

pradovic:

When adding tx-service to sdp we encountered circular dependency sdp -> tx-service -> da-sampling - da-network - sdp. 
In order to break the cycle, the proposal is to subscribe to transactions from tx services and filter out blobs to sample (from sampling service). 

Also some types were in adapters (verifier, sampling) were cleaned up along the way.

Yes

@pradovic @bacv 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/nomos-services/data-availability/sampling/src/lib.rs
+++ b/nomos-services/data-availability/sampling/src/lib.rs
@@ -733,6 +733,12 @@ where
             BoxFuture<'static, (BlobId, Option<DaSharesCommitments>)>,
         > = FuturesUnordered::new();
 
+        let pending_tasks = &mut PendingTasks {
+            long_tasks: &mut long_tasks,
+            sampling_continuations: &mut sampling_continuations,
+            delayed_sdp_sampling_triggers: &mut delayed_sdp_sampling_triggers,
+        };
+
         loop {
             tokio::select! {
                     Some(service_message) = service_resources_handle.inbound_relay.recv() => {
@@ -743,11 +749,7 @@ where
                             &mut sampler,
                             commitments_wait_duration,
                             &share_verifier,
-                            &mut PendingTasks {
-                                long_tasks: &mut long_tasks,
-                                sampling_continuations: &mut sampling_continuations,
-                                delayed_sdp_sampling_triggers: &mut delayed_sdp_sampling_triggers,
-                            }
+                            pending_tasks,
                         ).await;
                     }
                     Some(sampling_message) = sampling_message_stream.next() => {
@@ -765,7 +767,7 @@ where
                         ).await;
                     }
                     // Handle completed sampling continuations
-                    Some((blob_id, commitments)) = sampling_continuations.next() => {
+                    Some((blob_id, commitments)) = pending_tasks.sampling_continuations.next() => {
                         if let Some(commitments) = commitments {
                             info_with_id!(blob_id, "Got commitments for triggered sampling");
                             sampler.add_commitments(&blob_id, commitments);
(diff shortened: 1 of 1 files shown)


nomos-services/data-availability/sampling/src/mempool/sampling.rs
```diff
@@ -0,0 +1,96 @@
...
+                        if let Op::ChannelBlob(blob_op) = op {
+                            Some(blob_op.blob)
+                        } else {
+                            None
+                        }
+                    })
+                    .collect();
+
+                tokio_stream::iter(blob_ids)
+            });
```
bacv, line 86:
```suggestion
            .flat_map(|tx: Self::Tx| {
                let blob_ids_iter = tx.mantle_tx.ops.into_iter().filter_map(|op| {
                    if let Op::ChannelBlob(blob_op) = op {
                        Some(blob_op.blob)
                    } else {
                        None
                    }
                });

                tokio_stream::iter(blob_ids_iter)
            });
```

nodes/nomos-executor/config.yaml
```diff
@@ -149,6 +149,9 @@ da_sampling:
   commitments_wait_duration:
     secs: 2
     nanos: 0
+  sdp_blob_trigger_sampling_delay:
+    secs: 5
```
danielSanchezQ, line 153: Why 5? what I want to know if its specified somewhere or we just consulted the glass ball.

nomos-services/data-availability/sampling/src/lib.rs
```diff
@@ -697,6 +779,33 @@ where
...
+
+                    Some(blob_id) = blob_stream.next() => {
+                        Self::handle_incoming_blob(
+                            blob_id,
+                            sdp_blob_trigger_sampling_delay,
+                            &PendingTasks {
+                                long_tasks: &mut long_tasks,
+                                sampling_continuations: &mut sampling_continuations,
+                                delayed_sdp_sampling_triggers: &mut delayed_sdp_sampling_triggers,
+                            }
```
danielSanchezQ: Wouldn't it be better to pack it before hand and just share the &mut to the created `PendingTasks` instead of creating an object each time?

nomos-services/data-availability/sampling/src/lib.rs
```diff
@@ -697,6 +779,33 @@ where
...
+                            &storage_adapter,
+                            &mut sampler,
+                            commitments_wait_duration,
+                            &share_verifier,
+                            &mut PendingTasks {
+                                    long_tasks: &mut long_tasks,
+                                    sampling_continuations: &mut sampling_continuations,
+                                    delayed_sdp_sampling_triggers: &mut delayed_sdp_sampling_triggers,
+                                }
+                        ).await;
```
danielSanchezQ: Same here as we are immediately awaiting here.

danielSanchezQ (approved):
Left some tiny comments. Otherwise good job. Thanks.

nodes/nomos-executor/config.yaml
pradovic: It used to be like this, just in mempool config, so I copied it. I see now that I forgot to remove it from mempool config, I will add this to the PR as well

pradovic: @bacv Do you know the origin of 5s delay maybe? It used to be in a mempool config before

bacv: well this is a configurable delay, if tx appears in the mempool while shares are not fully disseminated sampling will fail. Disperser has no control over when the TX is put into the mempool because it's done by the DA nodes. Also TX is required for shares to be accepted, so TX will always hit DA node first, then shares.

pradovic: Tnx!

PR #1871, opened 2025-10-30
chore(da): Channel blob session number

bacv:

Add current_session number to BlobOp and set it when dispersing.

Yes

@bacv

Yes

No

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-core/chain-defs/src/mantle/encoding.rs
+++ b/nomos-core/chain-defs/src/mantle/encoding.rs
@@ -113,8 +113,9 @@ fn decode_channel_inscribe(input: &[u8]) -> IResult<&[u8], InscriptionOp> {
 }
 
 fn decode_channel_blob(input: &[u8]) -> IResult<&[u8], BlobOp> {
-    // ChannelBlob = ChannelId BlobId BlobSize DaStorageGasPrice Parent Signer
-    // Signer = Ed25519PublicKey
+    // ChannelBlob = ChannelId Session BlobId BlobSize DaStorageGasPrice Parent
+    // Signer Session = Uint64
+    // Signer  = Ed25519PublicKey
     let (input, channel) = map(decode_hash32, ChannelId::from).parse(input)?;
     let (input, session) = decode_uint64(input)?;
     let (input, blob) = decode_hash32(input)?;


davidrusu (changes requested):
sorry to block, just want to quickly discuss this approach a bit before we merge

nomos-core/chain-defs/src/mantle/encoding.rs
```diff
@@ -116,6 +116,7 @@ fn decode_channel_blob(input: &[u8]) -> IResult<&[u8], BlobOp> {
     // ChannelBlob = ChannelId BlobId BlobSize DaStorageGasPrice Parent Signer
     // Signer = Ed25519PublicKey
```
davidrusu: Also update this spec comment here
```suggestion
// ChannelBlob = ChannelId Session BlobId BlobSize DaStorageGasPrice Parent Signer
// Session = Uint64
// Signer  = Ed25519PublicKey
```

nomos-services/data-availability/dispersal/src/adapters/session/mod.rs
```diff
@@ -0,0 +1,28 @@
...
+    Channel(#[from] DynError),
+}
+
+#[async_trait::async_trait]
+pub trait SessionAdapter {
+    type Service: ServiceData;
+
+    fn new(relay: OutboundRelay<<Self::Service as ServiceData>::Message>) -> Self;
+
+    async fn subscribe(&self) -> Result<SessionStream, SessionAdapterError>;
```
pradovic, line 27: Just a nitpick, I saw in a codebase we started using new(overwatch_handle) instead, or even move the logic to subscribe(overwatch_handle) and let the adapter handle take the actual relay. 

I don't mind either way, it's really not a blocker.

pradovic (approved):

nomos-services/data-availability/dispersal/src/adapters/session/mod.rs
bacv: I've seen that too, it works when the adapter only provides one method (like in this case), but if we want the adapter to send more than one message to the service, then we'll need to revert back to this solution. I would leave this as is.

pradovic: I agree, if we need more then new method is mandatory, but it can still take the handle. I dont have a preference, I think this is good to goas is, no need to change!

PR #1873, opened 2025-10-31
feat(test): channel ops e2e

andrussal:

This PR adds channel ops e2e test.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/tests/Cargo.toml
+++ b/tests/Cargo.toml
@@ -43,7 +43,7 @@ nomos-da-sampling             = { workspace = true }
 nomos-da-verifier             = { workspace = true }
 nomos-executor                = { default-features = false, features = ["testing", "tracing"], workspace = true }
 nomos-http-api-common         = { workspace = true }
-nomos-ledger                 = { workspace = true }
+nomos-ledger                  = { workspace = true }
 nomos-libp2p                  = { workspace = true }
 nomos-network                 = { features = ["libp2p"], workspace = true }
 nomos-node                    = { default-features = false, features = ["testing"], workspace = true }


tests/src/tests/mantle/channel_ops.rs
```diff
@@ -0,0 +1,168 @@
...
+        Duration::from_secs(10),
+    )
+    .await
+    .expect("Channel should exist after inscription");
+
+    assert_ne!(
+        state.tip,
+        MsgId::root(),
+        "Channel tip should not be root after inscription"
+    );
```
davidrusu, line 91: Can we add an assert that the tip is in fact the inscription msg id?

```rust
assert_eq!(state.tip, signed_inscribe_tx.mantle_tx.ops[0].id())
```

tests/src/tests/mantle/channel_ops.rs
```diff
@@ -0,0 +1,168 @@
...
+    assert_eq!(
+        &*channel_state.keys,
+        &[new_authorized_key],
+        "Channel keys should be updated to the new key"
+    );
+
+    assert_ne!(
+        channel_state.tip,
+        MsgId::root(),
+        "Channel tip should not be root before publishing blob"
```
davidrusu, line 133: blob has not been published yet?

tests/src/tests/mantle/channel_ops.rs
```diff
@@ -0,0 +1,168 @@
...
+        reqwest::Url::parse(&format!("http://{backend_address}")).expect("Valid executor URL");
+
+    let blob_id = ExecutorHttpClient::new(None)
+        .publish_blob(
+            exec_url,
+            channel_id,
+            channel_state.tip,
+            new_authorized_key,
+            blob_payload.to_vec(),
+        )
```
davidrusu, line 149: any chance we can get the `publish_blob` api to return the blob msg id? This message id should be generated client side before publishing.

That way we can check that the tip is indeed equal to the blob msg id

andrussal: > any chance we can get the `publish_blob` api to return the blob msg id? This message id should be generated client side before publishing.
> 
> That way we can check that the tip is indeed equal to the blob msg id

Yep, good idea, tighened the check to verfify against the generated id.

tests/src/tests/mantle/channel_ops.rs
andrussal: > blob has not been published yet?

Here idea was to check that changing keys doesn't update tip. Made this also to check against previous tip instead of `MsgId::root()`.

davidrusu: I see, makes sense

andrussal:
Closing as superseded by newer tests

Issue #1860, opened 2025-10-29
Activity proof is occationally not submitted if the proxy service kills the core service.


youngjoon-lee:
The proxy service kills the core service when the node is transitioning from EdgeAfterCore to Edge -- i.e. right after the session transition period (STP) is done. Then, core service is killed immediately before finishing submitting an activity proof. 

SOLUTION:
If Overwatch had a feature to gracefully shut a service down, we could have used it to add a tear-down logic to the core service (e.g. finishing STP and submitting an activity proof). However, Overwatch doesn't have the feature yet.
For now, the cleanest way is letting each service (core and edge) terminate by itself when they're ready to terminate. Then, the responsibility of the proxy service is just starting the other service without manually shutting the existing service down.
The core service will terminate by itself after a STP if it receives a new session where it's not core anymore. Whereas, the edge service will terminate by itself once it receives a new session where it's not edge anymore.

PR #1875, opened 2025-10-31
fix(blend): complete STP before terminating core service

### LINKED ISSUE #1855: Core service terminates by itself before transition is done

youngjoon-lee:
Closes #1855 

**Background:**
Previously, the core service has been terminating by itself once it receives a new session where it's not core anymore. That's wrong because the core service should also complete the last STP (session transition period) before shutdown. 

**Solution:**
In this PR, the core service still exits from the main event loop once it receives a non-core session. But, instead of terminating immediately, it runs another event loop only for the last STP. This event loop does less jobs than the main loop. Also, this stops once a STP has passed. Then, the core service finally terminates.

The test is coming as a next PR. Since the core service structure is not good for testing, I did some refactoring.

NOTE: This PR doesn't fix https://github.com/logos-co/nomos/issues/1860 (service lifecycle mgmt in proxy service). I will work on it separately.

yes

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/core/mod.rs
+++ b/nomos-services/blend/src/core/mod.rs
@@ -514,20 +514,57 @@ where
                 }
                 Some(session_event) = remaining_session_stream.next() => {
                     match handle_session_event(session_event, &blend_config, crypto_processor, current_public_info, current_recovery_checkpoint, &mut blending_token_collector, &mut backend).await {
-                        Ok((new_crypto_processor, new_public_info, new_recovery_checkpoint)) => {
-                            crypto_processor = new_crypto_processor;
-                            current_public_info = new_public_info;
-                            current_recovery_checkpoint = new_recovery_checkpoint;
+                        Ok(output) => {
+                            crypto_processor = output.crypto_processor;
+                            current_public_info = output.public_info;
+                            current_recovery_checkpoint = output.recovery_checkpoint;
                         }
-                        Err(e) => {
+                        Err((e, output)) => {
                             tracing::error!(
                                 target: LOG_TARGET,
                                 "Terminating the '{}' service as the new membership does not satisfy the core node condition: {e:?}",
                                 <RuntimeServiceId as AsServiceId<Self>>::SERVICE_ID
                             );
-                            return Err(e.into());
+                            crypto_processor = output.crypto_processor;
+                            current_public_info = output.public_info;
+                            current_recovery_checkpoint = output.recovery_checkpoint;
+                            break;
                         }
                     }
+                }
+            }
+        }
+
+        // The main event loop has ended because the node is no longer a core node
+        // in the new session.
+        // Before terminating the service, complete the old session during the
+        // session transition period.
+        //
+        // First, drop unnecessary streams to not block senders.
+        drop((
+            // Don't receive local data messages, as they shouldn't be sent in the old session.
+            inbound_relay,
+            // Don't receive secret PoL info updates, as new messages are not generated for the old
+            // session.
+            secret_pol_info_stream,
+        ));
+        loop {
+            tokio::select! {
+                Some(incoming_message) = blend_messages.next() => {
+                    current_recovery_checkpoint = handle_incoming_blend_message(incoming_message, &mut message_scheduler, &crypto_processor, &mut blending_token_collector, current_recovery_checkpoint);
+                }
+                Some(round_info) = message_scheduler.next() => {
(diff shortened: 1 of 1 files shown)


nomos-services/blend/src/core/mod.rs
```diff
@@ -514,23 +514,60 @@ where
...
+        // session transition period.
+        //
+        // First, drop unnecessary streams to not block senders.
+        drop((
+            // Don't receive local data messages, as they shouldn't be sent in the old session.
+            inbound_relay,
+            // Don't receive secret PoL info updates, as new messages are not generated for the old
+            // session.
+            secret_pol_info_stream,
+        ));
```
youngjoon-lee, line 550: I'd thought about this point multiple time, but ended up having two loops in the same place and having this `drop` explicitly between the loops.

I considered defining a function for each loop and removing the `drop` by making the 1st function consumes those resources, but I realized that it could cause potential misuse, and not good for readability. I want the the 2nd loop always comes after the 1st one.

nomos-services/blend/src/core/mod.rs
```diff
@@ -514,23 +514,60 @@ where
...
+        loop {
+            tokio::select! {
+                Some(incoming_message) = blend_messages.next() => {
+                    current_recovery_checkpoint = handle_incoming_blend_message(incoming_message, &mut message_scheduler, &crypto_processor, &mut blending_token_collector, current_recovery_checkpoint);
+                }
+                Some(round_info) = message_scheduler.next() => {
+                    current_recovery_checkpoint = handle_release_round(round_info, &mut crypto_processor, &mut rng, &backend, &network_adapter, current_recovery_checkpoint).await;
```
ntn-x2, line 557: I think we don't want to deal with release rounds anymore either, no? We only accept incoming messages, but should not be pushing any new messages out.

ntn-x2 (approved):
Looks good! Just left a comment about probably one more stream we should drop.

nomos-services/blend/src/core/mod.rs
youngjoon-lee: I think we still need the scheduler because we need to forward (or broadcast) any processed messages received during STP in the line 554. Please correct me if I'm missing anything.

youngjoon-lee: Hm. I think I know what you mean. We may not need to generate new cover messages during STP when its core role is fading out. I will open a discussion in Discord since it's not defined in the spec. If necessary, we should probably refactor the `message_scheduler` or introduce any better way.

I think it's okay to merge this PR for now because generating new cover messages here doesn't make any issue, I think. I will add more tests soon when fixing additional bugs. I anyway have to write tests to cover all the cases/behaviors of session transitions.

youngjoon-lee: As discussed in https://discord.com/channels/1111286067413405788/1434779184206712862, we keep the current implementation that generates new cover messages during the last STP because it doesn't break the definition of STP, but beneficial in terms of privacy.

Issue #1866, opened 2025-10-30
The updated crypto processor can't decapsulate messages from the old session


youngjoon-lee:
When a new session info arrives, the core service recreates the crypto process with the new public info. 
https://github.com/logos-co/nomos/blob/5bd5fe06d1c4db31e12c421e4a62697b60c59a3c/nomos-services/blend/src/core/mod.rs#L516-L516

Then, it cannot decapsulate messages received from the old session during STP (session transition period). We may need to keep the old crypto processor until STP is done. 

I'm not sure if this is a bug. I have to test it first.

This issue is related to https://github.com/logos-co/nomos/issues/1854.

Issue #1854, opened 2025-10-28
Collect blending tokens from previous session as well


youngjoon-lee:
Currently, after a session transition period starts, blending tokens received from connections in the previous session are collected to the token set for the current session. It is a bug. They must be collected for the previous session during the transition period by adding the following method, so that we can compute an activity proof including them. 
```rust
impl BlendingTokenCollector {
    /// Collects a blending token from the old session.
    pub fn collect_from_old_session(&mut self, token: BlendingToken) {
```

In the service layer, we currently cannot check if a message was received from the previous session or the current one, because the Swarm doesn't provide the information. We should think about how to resolve this.

youngjoon-lee:
> In the service layer, we currently cannot check if a message was received from the previous session or the current one, because the Swarm doesn't provide the information. We should think about how to resolve this.

Actually, I think we don't need to refactor Swarm to provide the origin of messages. Instead, we can try to decapsulate messages (returned from Swarm) using both current and old crypto processor. Only one of them would be successful. Then, we can collect the blending token from the messages, accordingly.

For that, we also need to fix https://github.com/logos-co/nomos/issues/1866

PR #1876, opened 2025-10-31
refactor(blend): add testing framework for core service

youngjoon-lee:

Closes https://github.com/logos-co/nomos/issues/1855

This is a follow-up PR of https://github.com/logos-co/nomos/pull/1875

I added a test for #1875, by refactoring the core service as below:
- Split the `ServiceCore::run` into 3 functions: `initialize`, `post_initialize`, and `run_event_loop`.
- In the tests, we can call those 3 functions with `Stream`s that we manually create, instead of implementing/creating all the dependencies (services). We can manually send events through the `Stream`s at the time we want, and check if the core service reacts as expected.

The diff looks complicated, but the key changes are what I listed up above. I didn't change any logic. I just split functions. It might be easier to open a file in your editor, rather than reading diffs. 

This PR contains only one test for #1875. Still there are many cases not covered yet. I will add tests soon, but I need to fix other bugs (https://github.com/logos-co/nomos/issues/1854 and https://github.com/logos-co/nomos/issues/1866) as well to make all tests work.

Yes

@youngjoon-lee @ntn-x2 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/ledger/nomos-ledger/src/cryptarchia/mod.rs
+++ b/ledger/nomos-ledger/src/cryptarchia/mod.rs
@@ -339,7 +339,7 @@ impl core::fmt::Debug for LedgerState {
 
 #[cfg(test)]
 pub mod tests {
-    use std::{num::NonZero, sync::Arc};
+    use std::num::NonZero;
 
     use cryptarchia_engine::EpochConfig;
     use groth16::Field as _;
@@ -350,7 +350,7 @@ pub mod tests {
             gas::MainnetGasConstants, ledger::Tx as LedgerTx, ops::leader_claim::VoucherCm,
         },
         proofs::zksig::DummyZkSignature,
-        sdp::{ServiceParameters, ServiceType},
+        sdp::ServiceParameters,
     };
     use num_bigint::BigUint;
     use rand::{RngCore as _, thread_rng};
@@ -465,6 +465,28 @@ pub mod tests {
 
     #[must_use]
     pub fn config() -> Config {
+        let mut service_params = std::collections::HashMap::new();
+        service_params.insert(
+            nomos_core::sdp::ServiceType::DataAvailability,
+            ServiceParameters {
+                lock_period: 10,
+                inactivity_period: 20,
+                retention_period: 100,
+                timestamp: 0,
+                session_duration: 10,
+            },
+        );
+        service_params.insert(
+            nomos_core::sdp::ServiceType::BlendNetwork,
+            ServiceParameters {
+                lock_period: 10,
+                inactivity_period: 20,
+                retention_period: 100,
+                timestamp: 0,
+                session_duration: 10,
+            },
+        );
+
         Config {
             epoch_config: EpochConfig {
                 epoch_stake_distribution_stabilization: NonZero::new(4).unwrap(),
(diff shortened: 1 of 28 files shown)


nomos-services/blend/src/core/processor.rs
```diff
@@ -97,7 +97,7 @@ pub enum Error {
 }
 
 #[cfg(test)]
-mod tests {
+pub mod tests {
```
ntn-x2: If stuff from this module is needed elsewhere, I think it's better to move those into `test_utils` and leave this module private. I don't think we should make the test module public.

nomos-services/blend/src/core/tests/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+/// it's still core. Also, check if it stops after the session transition period
+/// if it receives another new session that doesn't meet the core node
+/// conditions.
+#[test_log::test(tokio::test)]
+async fn complete_old_session_after_main_loop_done() {
+    let minimal_network_size = 2;
+    let (membership, local_private_key) = new_membership(minimal_network_size);
+
+    // Create settings.
+    let (settings, _recovery_file) = settings(
```
ntn-x2, line 40:
```suggestion
    let (settings, _) = settings(
```

nomos-services/blend/src/core/tests/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    // Send a new session with a new membership smaller than minimal size
+    membership_info.membership = new_membership(minimal_network_size.checked_sub(1).unwrap()).0;
+    membership_info.session_number += 1;
+    membership_sender.send(membership_info).await.unwrap();
+
+    // Since the network is smaller than the minimal size,
+    // the service must stop after a session transition period.
+    timeout(
+        settings.time.session_transition_period().mul_f64(1.5),
```
ntn-x2: Can we test that any logic in the session transition session is executed, somehow? With this condition we are testing, we only know that after 1.5x the transition period the function exits, but we don't know exactly when. We can use, for instance, `backend.complete_session_transition()` to be notified when the function is called. Or use the blending token collector to make sure it is called before quitting.

ntn-x2 (commented):
Thanks for this refactoring. I agree we might want to make things nicer later on, but this makes the service much more testable.

nomos-services/blend/src/core/tests/mod.rs
youngjoon-lee: Ahah, right. This test is not correct. Let me think how to make this correct.

nomos-services/blend/src/core/tests/mod.rs
youngjoon-lee: If we use `_`, the `recovery_file: NamedTempFile` instance is dropped and the corresponding temp file is deleted from disk. Although this test currently doesn't produce any state to be stored, but I think it's better to not drop the file instance for the future.

nomos-services/blend/src/core/tests/mod.rs
youngjoon-lee: Thank you, and sorry for this mistake. I fixed it 02bb804ac28155a83eff4ce035e4129e8c858dc8. I also identified and fixed a bug in the logic. Now I think this PR can be merged.

nomos-services/blend/src/core/processor.rs
youngjoon-lee: Thank you. I was lazy at that time. I fixed it f07cc6ee073c4bbc1b85ea539d197f7a5a147e8e.

PR #1879, opened 2025-11-02
feat(ledger): Add Da rewards

zeegomo:
Implement DA rewards as specified in https://www.notion.so/nomos-tech/NomosDA-Rewarding-203261aa09df80af8c77dfb3dc593673?d=29c261aa09df803790ed001cda56c5ca#223261aa09df803197f0f6fdfa11ff39.
There are a few documented discrepancies but only in what is considered a valid activity proof and not in rewards calculation. In particular, the code accepts a few more cases than described in the spec. 
Also, we don't actually distribute rewards yet, but that will be added in a coming PR.

I'll also improve tests tomorrow

Yes

> [!IMPORTANT]  
@madxor 
Yes
yes, as documented, we can resolve discrepancies before merging if necessary
## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/ledger/nomos-ledger/src/mantle/sdp/mod.rs
+++ b/ledger/nomos-ledger/src/mantle/sdp/mod.rs
@@ -222,7 +222,7 @@ impl SessionState {
 impl<R: Rewards> ServiceState<R> {
     fn try_apply_header(mut self, block_number: u64, config: &ServiceParameters) -> Self {
         // TODO: remove expired declarations based on retention_period
-        let current_session = block_number / config.session_duration;
+        let current_session = config.session_for_block(block_number);
         // shift all session!
         if current_session == self.active.session_n + 1 {
             // Update rewards with current session state and distribute rewards
@@ -235,6 +235,10 @@ impl<R: Rewards> ServiceState<R> {
                 session_n: self.forming.session_n + 1,
             };
         } else {
+            assert!(
+                current_session < self.active.session_n + 1,
+                "Nomos isn't ready for time travel yet"
+            );
             self.forming = self.forming.update(&self, block_number, config);
         }
 
--- a/ledger/nomos-ledger/src/mantle/sdp/rewards.rs
+++ b/ledger/nomos-ledger/src/mantle/sdp/rewards.rs
@@ -130,13 +130,10 @@ impl DaRewards {
             });
         }
 
-        // Validation is performed a little bit different from the spec even though the
-        // result is the same
         // * self opinion is not checked
         // * we only check opinions up to the number of validators, without checking the
         //   rest of the opinion bits are zero
 
-        // bits beyond n_validators should be zero but we ignore them
         Ok((0..n_validators)
             .map(|i| Self::get_opinion_bit(opinions, i))
             .collect())
@@ -144,9 +141,6 @@ impl DaRewards {
 
     /// Calculate expected byte length for opinion vector: ⌈log₂(Ns + 1) / 8⌉
     const fn calculate_opinion_vector_length(node_count: usize) -> usize {
-        if node_count == 0 {
-            return 0;
-        }
         let bits_needed = (node_count + 1).next_power_of_two().trailing_zeros() as usize;
         bits_needed.div_ceil(8)
     }
--- a/nomos-core/chain-defs/src/sdp/mod.rs
+++ b/nomos-core/chain-defs/src/sdp/mod.rs
@@ -39,6 +39,13 @@ pub struct ServiceParameters {
     pub session_duration: BlockNumber,
 }
 
+impl ServiceParameters {
+    #[must_use]
+    pub const fn session_for_block(&self, block_number: BlockNumber) -> SessionNumber {
+        block_number / self.session_duration
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
 #[serde(transparent)]
 pub struct Locator(pub Multiaddr);


ledger/nomos-ledger/src/mantle/sdp/rewards.rs
```diff
@@ -56,23 +56,547 @@ pub struct NoopRewards;
...
+        if opinions.len() != expected_current_len {
+            return Err(Error::InvalidOpinionLength {
+                expected: expected_current_len,
+                got: opinions.len(),
+            });
+        }
+
+        // Validation is performed a little bit different from the spec even though the
+        // result is the same
+        // * self opinion is not checked
```
madxor, line 133: I agree, we don't need to check this. I will update the spec.

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
```diff
@@ -56,23 +56,547 @@ pub struct NoopRewards;
...
+        // * we only check opinions up to the number of validators, without checking the
+        //   rest of the opinion bits are zero
```
madxor, line 135: What is the motivation to not to check that?

zeegomo: It's more work to do for something we don't care about, like, we would specifically need to check for those values but then ignore the result anyway and only return an error if they are set, why not ignoring this all together?

madxor: Just to filter out nodes that are not complying with the protocol and for example are sending truly random opinions? If they do not comply, their activity message will be rejected so they should not get a reward. I know it's not a strong way of verifying the compliance but it's pretty much the only thing that we can do.

ledger/nomos-ledger/src/mantle/sdp/mod.rs
```diff
@@ -212,14 +205,30 @@ impl SessionState {
         }
         self.clone()
     }
+
+    /// Get provider ID by index using the stable iteration order of the map.
+    ///
+    /// The `RedBlackTrieMapSync` provides a stable iteration order based on its
+    /// internal structure.
```
davidrusu, line 212: Is this order insertion order? or ordered by the ord defined on `ProviderId`?

ledger/nomos-ledger/src/mantle/sdp/mod.rs
```diff
@@ -212,14 +205,30 @@ impl SessionState {
...
+            .values()
+            .nth(index)
+            .map(|decl| decl.provider_id)
+    }
 }
 
 impl<R: Rewards> ServiceState<R> {
     fn try_apply_header(mut self, block_number: u64, config: &ServiceParameters) -> Self {
+        // TODO: remove expired declarations based on retention_period
         let current_session = block_number / config.session_duration;
```
davidrusu: nit:
```suggestion
        let current_session = config.session_for_block(block_number);
```

ledger/nomos-ledger/src/mantle/sdp/mod.rs
davidrusu, line 237: can you add a panic if `current_session > self.active.session_n + 1`? something has gone wrong if that's the case and we should halt the node (rather than continuing to run with SDP stuck on an old session)

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
```diff
@@ -56,23 +56,547 @@ pub struct NoopRewards;
...
+        Ok((0..n_validators)
+            .map(|i| Self::get_opinion_bit(opinions, i))
+            .collect())
+    }
+
+    /// Calculate expected byte length for opinion vector: ⌈log₂(Ns + 1) / 8⌉
+    const fn calculate_opinion_vector_length(node_count: usize) -> usize {
+        if node_count == 0 {
+            return 0;
+        }
```
davidrusu: The `node_count=0` case is handled just fine in the code below, we don't need this special case.

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
madxor: I thought about this, and I agree, we have a very permissive design so to save on complexity we can just disregard trailing bits.

ledger/nomos-ledger/src/mantle/sdp/mod.rs
zeegomo: order by `ProviderId`

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
```diff
@@ -21,11 +24,11 @@ pub trait Rewards: Clone + PartialEq + Eq + Send + Sync + std::fmt::Debug {
     /// Called when a provider submits an active message with metadata
     /// (e.g., activity proofs containing opinions about other providers).
     fn update_active(
-        &mut self,
-        declaration_id: DeclarationId,
+        &self,
+        declaration_id: ProviderId,
```
madxor, line 28: Is this `update_active` some intermediary call that somehow retrieves `DeclarationId` from `ProviderId` before sending the active message? If that's not the case then I don't think that we can use `ProviderId` as it's uniqueness is not guaranteed. The same `ProviderId` can be reused in multiple declarations.

zeegomo: Good catch, does this mean a provider with multiple declarations can send multiple activity proofs with possibly different values for other providers activity?

danielSanchezQ: ccing @bacv To this convo. Most probably the actual implementation does it (?)

bacv: > does this mean a provider with multiple declarations can send multiple activity proofs

As of now a node will maintain one provider id per service, and use that provider to send the activity proof, but if different node instances have same provider id, but multiple declarations (lets say different Locators for the same service), the same provider will post activity proofs for different declarations.

PR #1880, opened 2025-11-03
test(chain): increase timeout for `test_orphan_handling`

youngjoon-lee:

The `test_orphan_handling` has been frequently failing with a timeout, because the 3rd node didn't finish the catch-up (via orphan handling).

I found the reasons:
- The 3rd node didn't even received any orphan block.
- Why? Two initial nodes hadn't proposed any block since the 3rd block started. 
- Since the 3rd node didn't receive any block from other two nodes, orphan handling logic wasn't triggered.

Why didn't the two initial nodes propose any block for a few seconds?
It's because the total stake is high. Since the recent change https://github.com/logos-co/nomos/commit/6d3683816fb3f7927f16c4a6d2a3931820b3f0a1, we've been creating 3 UTXOs for each node, but only one of them is eligible for leadership. (1 for leadership, 1 for DA, and 1 for Blend). This means that we create 9 UTXOs (3 * 3 nodes) in total for the genesis. That is, the total stake is 9. Therefore, the block proposal interval could be quite long.

Even more, after the test spawns the 3rd node, it sometimes take much time than expected until a block is proposed by one of two initial nodes, because the 3rd node also has one eligible UTXO for leadership.

Therefore, this PR sets timeout for the test very long enough. I confirmed that the current test failure doesn't tell us anything wrong in the orphan handling. So, it's okay to set a long timeout to wait until the scenario that we want happens.

This test failure was reproduced in my local. After increasing the timeout, I ran the test more than 20 times, and didn't see any failure.

Yes

@youngjoon-lee @andrussal 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/tests/src/common/sync.rs
+++ b/tests/src/common/sync.rs
@@ -66,11 +66,13 @@ pub async fn wait_for_validators_mode(
 }
 
 fn print_validators_info(infos: &[CryptarchiaInfo]) {
-    println!(
-        "   Validators: {:?}",
-        infos
-            .iter()
-            .map(|info| format!("{:?}/{:?}", info.height, info.mode))
-            .collect::<Vec<_>>(),
-    );
+    println!("   Validators: {:?}", format_cryptarhica_info(infos));
+}
+
+#[must_use]
+pub fn format_cryptarhica_info(infos: &[CryptarchiaInfo]) -> Vec<String> {
+    infos
+        .iter()
+        .map(|info| format!("Height({})/{:?}/{:?}", info.height, info.slot, info.mode))
+        .collect()
 }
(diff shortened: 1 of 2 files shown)


andrussal (approved):
Ah, this makes complete sense—thanks for untangling it. LGTM!

bacv:
> It's because the total stake is high

Would it be possible to increase the value of leadership notes for node 1 and 2, making the stake even higher, but concentrated between these two nodes. Wouldn't that allow to reduce the time required for the test?

bacv:
> That makes sense. I tried similar way but your suggestion is better. Let me try.

We could keep the high timeout but it would be nice to have the test finish faster (on a good day).

Issue #1867, opened 2025-10-30
Fix Blend message release logic


ntn-x2:
Currently, Blend core service stores all decapsulated messages received from other nodes in the message scheduler. Then, at the next release round, it releases ALL these messages plus an optional cover message to the network. This logic is actually wrong.

What is supposed to happen, is that:

* cover message generation happens independently of everything else
* a separate component keeps track of two queues: one for processed messages and one for cover messages
* when a release round happens, the node knows it has to release at most `N` messages, where `N` is calculated with [this formula](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df817e9fb1c108c8bfe8f2). It then picks at most `N` items, first from the processed messages queue, and then if more items are needed, from the cover messages queue. This means that each release round will include at most `N` messages.

Discord discussion for more context: https://discord.com/channels/1111286067413405788/1433027500765614100.

madxor:
Looks good. The relevant section of the spec was clarified: https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81a08689e031f2a202ee

PR #1881, opened 2025-11-03
fix(blend): update release logic

ntn-x2:

Fixes https://github.com/logos-co/nomos/issues/1867. The release schedule needed a little bit of tweaking. The main logic changes introduced are that:

1. Data messages are not blended immediately, but they are scheduled for release at the next round, regardless of any other message types that might be released as well.
2. Scheduling a data message removes a random pre-scheduled cover message from the unreleased ones, instead of skipping the next cover message (to maintain statistical indistinguishability). This is achieved by having a threshold formula that makes releasing a cover message more or less likely depending on how many unprocessed data messages there are (the more, the more likely a cover message is skipped) and how many remaining cover messages are scheduled (the more, the less likely the next cover release round will skip the cover message), with the trivial cases that if the number of processed messages is greater than the number of scheduled release slots, no more cover message will be released. I found this solution to be much simpler and more efficient than having some additional data structure that would allow removal of random elements.

Another minor change I introduced, is for `RoundInfo` to explicitly state what round type we are dealing with: whether data only, data + cover, data + processed, data + cover + processed. I found this clearer to follow, while still mapping each variant to the right value upon new round ticks.

Yes.

@ntn-x2 dev, @madxor spec, @youngjoon-lee review

Yes.

Yes. See linked GH issue for more context.

## Checklist

* [x] 2. Description added.


nomos-blend/scheduling/src/cover_traffic.rs
youngjoon-lee:
```suggestion
/// Notify the scheduler that a new data message has been scheduled, which
   /// will affect the generation of the next scheduled cover message.
   pub fn notify_new_data_message(&mut self) {
```
Do we need to update this comment?: `emitted` -> `scheduled`. 
(Github doesn't show this diff well)

nomos-blend/scheduling/src/cover_traffic.rs
```diff
@@ -53,28 +56,69 @@ impl<RoundClock> SessionCoverTraffic<RoundClock> {
...
+            let threshold =
+                self.unprocessed_data_messages as f64 / current_scheduled_messages_count as f64;
+            self.rng.gen_range(0f64..=1f64) <= threshold
+        };
+        // If we are not emitting a cover message because of an unprocessed data
+        // message, "burn" one quota.
+        if should_skip {
+            self.unprocessed_data_messages = self.unprocessed_data_messages.saturating_sub(1);
+            trace!(target: LOG_TARGET, "Pre-schedule release skipped because of data message.");
         }
```
youngjoon-lee, line 112: I still can't quite understand it, even though I've read the PR description, the issue, and the spec.

Could you help me understanding:
- What does the `threshold` mean?
- What does the coin flipping mean? I'm trying to understand how this probabilistic manner achieves our goal.

I couldn't find this from the spec, if I don't miss anything. 
I think I have to understand the following description but not intuitive. I need your help. 
> This is achieved by having a threshold formula
…
I found this solution to be much simpler and more efficient than having some additional data structure that would allow removal of random elements.

And, at the line 98, I'd like to understand why we decrease `unprocessed_data_messages` instead of `scheduled_message_rounds`. We've had this logic before, but I suddenly feel this is not intuitive. Does this module decrease this value, under an assumption that a data message will be released on this round by the upper layer?

youngjoon-lee: Also, I think the comment should be updated:
```diff
-            // Threshold will be in the range (0, unprocessed_data_messages], since we
+            // Threshold will be in the range (0, 1], since we
```

ntn-x2: You are right. The coin flipping is not part of the spec. It is used to implement the bit of the spec that requires us to _randomly_ unschedule an unreleased cover message.

We currently have a hashset for looking up whether a new round is a cover release round. Hashsets do not have random access, hence we can't just remove a random element from it in O(1). We are then left with a few options:

* We use an auxiliary vector that we use to remove random elements. This was actually my first approach, which works if we use `swap_remove` instead of `remove` to avoid shifting the whole vector everytime, but we make the logic more complicated because we now have to update the index of the element that was swapped from last place to the place of the element we removed. As I said, I took this approach first before coming up with the coin flipping solution which, in my opinion, is much more elegant and does not rely on any external dependency nor any auxiliary data structure besides the set of scheduled rounds
* Use an auxiliary structure like `IndexMap`, which internally does the same as the point above, just hiding it behind an abstraction
* Use the coin-flipping method.

For the coin-flipping method, the reasoning is actually quite simple. We calculate `threshold` which is the ratio of unprocessed data messages over the number of remaining cover messages to schedule. Now, there are three cases here:
1. `Unprocessed data messages >= remaining cover messages` -> the ratio will be `>= 1`, since we are basically supposed to ignore all remaining cover messages until the end of the session. This works since generating a random value in the range `[0, 1]` will always be `< 1`, hence we will always skip cover messages.
2. `Unprocessed data message == 0` -> We explicitly check for this, and we do not skip any future cover message.
3. `0 < threshold < 1` -> This is the case we solve with the coin flipping. Basically we calculate how many data messages we have not processed, and how many cover messages we have to release until the end of the session. At each release round, we run the coin flip, and skip the message if the coin flip is above the threshold. The threshold is the ratio between unprocessed data messages and remaining cover messages, so that means that the more unprocessed data messages there are, the more likely we are to skip the next cover message, and the more cover messages are scheduled, the less likely it is that the next one will be skipped. Of course, in the end if no cover message is skipped, there will be a point in which the number of unprocessed data messages will reach the number of remaiing cover messages, which brings us back to point `1`. If, on the other side, the next `N` cover messages are skipped and we consume all the unprocessed data messages, then we are back at point `2`.

In short, the coin-flipping is used to decide whether the next cover message that should be emitted will actually be emitted, based on number of unprocessed data messages and remaining number of cover messages still to generate. This is used instead of randomly remove items from the set of pre-scheduled rounds for the reasons I mentioned above.

ntn-x2: > Also, I think the comment should be updated.

I hope it's clear now from the comment above that this is indeed correct. Maybe "threshold" is not the right term, but the range of values it can take goes from `0`, if the number of processed data messages is `0`, to the number of processed data messages, if the number of remaining cover messages is `1` (we check that the set is not empty at the beginning of the function, so this value can never be `0`). We only sample in the range `[0, 1]` but the threshold can be larger than 1, which means that we should always ignore the next cover message no matter what.

nomos-blend/scheduling/src/cover_traffic.rs
ntn-x2: Well, this is one thing that changed in the spec. We are not releasing data messages immediately, but we queue them until the next round. But as soon as we queue them, we notify the cover traffic module. So when this function is called, the data message has not been emitted yet, just scheduled. Nevertheless, we call this function already when the data message is queued since we want to give it a change to skip the next cover round, if it happened to be the next one when also the data message will be released.

youngjoon-lee: Ah, yeah. For that reason, I think we need to update the comment in the code.

The comment is still saying that `a new data message has been emitted`.
I think we should update it to `a new data message has been scheduled`.

nomos-blend/scheduling/src/cover_traffic.rs
youngjoon-lee: Thank you for your detailed explanations. All is clear now!

youngjoon-lee (approved):
Great!

PR #1891, opened 2025-11-04
feat(blend): optimize message decapsulation

ntn-x2:

Since the intended recipient of a given encapsulation of a message is chosen at random, it can happen that more than one consecutive layer is addressed to the same node. We introduced a small optimization in the spec that accounts for this, with a [new bullet point](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#2a1261aa09df806b83b0e18fb2b42da9).

Furthermore, not within the scope of the spec but an issue for our tests, if we have test environments with only two nodes, and node A generates a message which has the outermost layer that can only be decapsulated by node A, and this message is sent to node B, the message will be lost, because node B cannot decapsulate it since it's not the intended recipient, but at the same time it won't send the message back to A, since it was the original sender. This is only an issue in networks with 2 nodes, since if we add a third node, the third node would relay messages back to node A, which would be able to decapsulate the outermost layer. Anyway, instead of changing our test cases to deploy 3 nodes, and since it is actually economically advantageous for a node to collect blending tokens anyway, I introduced logic where local data messages are processed, to check whether after encapsulating the message, we can already peel one or more layer off from it, instead of changing the proof generation logic to skip proofs that included the sender in any encapsulation layer. This would allow us to keep the current testing infrastructure in place, and also to avoid problems in the future if we were to ever deploy a 2-node network, for some reason. This will also remove the blocker to switch to actual core proofs, since right now they don't work in our tests because of the aforementioned issue.

So this PR introduces a feature in the core service cryptographic processor to decapsulate multiple layers and collect blending tokens on the way, and it integrates this mechanism in the blend core service in the following places:
* when a data message is received by cryptarchia leader. In this case, after encapsulation, we try to decapsulate one or more layers.
* when a cover message should be emitted. Same as above: we encapsulate it, then try to decapsulate one or more layers. If we decapsulate all layers, we collect the blending tokens and don't send the cover message out. This is perfectly fine since it is unlikely but could still happen.
* when a new Blend message is received: we don't limit ourselves to a single layer, but we try to batch-decapsulate more than one, and we stop when we can't anymore or when there's no more layers left.

We can remove the additional logic in steps 1 and 2 once we move away from having the minimum network size a client-side config, and we make it either a const in the code or we put it in the genesis or in some genesis-related information. That will happen soon.

With this new feature in, and the fix to the merkle logic merged with https://github.com/logos-co/nomos/pull/1878, there's no blocker to switching to actual core proofs. That will happen in https://github.com/logos-co/nomos/pull/1895, which is based on top of this PR.

Yes. I tried to keep the changes are concise and localized as possible.

@ntn-x2 dev, @madxor spec author, @youngjoon-lee reviewer.

Yes.

Yes. Please check the link above.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/core/mod.rs
+++ b/nomos-services/blend/src/core/mod.rs
@@ -992,6 +992,8 @@ where
     // Before blending the data message, we try to peel off any outer layers that
     // are addressed to us. In this case, we collect the blending tokens and we
     // blend only the remaining layers.
+    // TODO: Remove this logic once we don't have tests that deploy less than 3
+    // Blend nodes, or when we start using a minimum network size of 3.
     let self_decapsulation_output = cryptographic_processor.decapsulate_message_recursive(
         IncomingEncapsulatedMessageWithValidatedPublicHeader::from_message_unchecked(
             wrapped_message.clone(),
@@ -1220,6 +1222,7 @@ where
         .collect::<Vec<_>>();
 
     if should_generate_cover_message
+        // TODO: Remove this logic once we don't have tests that deploy less than 3 Blend nodes, or when we start using a minimum network size of 3.
         && let Some(encapsulated_cover_message) = generate_and_try_to_decapsulate_cover_message(
             cryptographic_processor,
             blending_token_collector,


nomos-blend/message/src/encap/validated.rs
```diff
@@ -33,6 +33,18 @@ impl<const ENCAPSULATION_COUNT: usize>
...
+    /// header has been verified.
+    ///
+    /// This function should only be called from context where it's guaranteed
+    /// the message has a valid public header.
+    #[must_use]
+    pub const fn from_message_unchecked(
+        encapsulated_message: EncapsulatedMessage<ENCAPSULATION_COUNT>,
     ) -> Self {
         Self(encapsulated_message)
     }
```
youngjoon-lee, line 50: Can't we have only the `from_message_unchecked` function which is `pub` (not `pub(super`))? Or, do we have any reason to keep the `from_message` function?

nomos-services/blend/src/core/mod.rs
```diff
@@ -984,13 +987,57 @@ where
...
 
-    scheduler.queue_data_message(wrapped_message.clone());
-
     let mut state_updater = current_recovery_checkpoint.start_updating();
+
+    // Before blending the data message, we try to peel off any outer layers that
+    // are addressed to us. In this case, we collect the blending tokens and we
+    // blend only the remaining layers.
+    // TODO: Remove this logic once we don't have tests that deploy less than 3
+    // Blend nodes, or when we start using a minimum network size of 3.
```
youngjoon-lee, line 996: Yeah, I agree to remove this logic later, because having this logic reduces traffic on the entire network, though this is an edge case. In other words, this logic indirectly reduces the number of blend layers that this node uses.
But, at the same time, I don't feel it's safe to rely on the const parameter (if we change it to 3). After long long time, we may forget about this edge case, and we may change the minimum network size back to 2 for some reason.
So, I think we may need to introduce a logic for nodes to exclude themselves when encapsulating messages somehow, at some time. Right now, I'm not sure if it would be possible since you mentioned that it's based on the generated PoQ. So, currently, I agree with having this decapsulation logic here right now.

nomos-services/blend/src/core/mod.rs
```diff
@@ -1023,24 +1070,28 @@ fn handle_incoming_blend_message<
 ) -> ServiceState<BackendSettings, BroadcastSettings>
 where
     NodeId: 'static,
-    BroadcastSettings: Serialize + for<'de> Deserialize<'de> + Eq + Hash + Clone + Send,
+    BroadcastSettings: Serialize + for<'de> Deserialize<'de> + Debug + Eq + Hash + Clone + Send,
     BackendSettings: Clone,
     ProofsVerifier: ProofsVerifierTrait,
 {
-    let Ok(decapsulated_message) = cryptographic_processor.decapsulate_message(validated_encapsulated_message).inspect_err(|e| {
+    let Ok(multi_layer_decapsulation_output) = cryptographic_processor.decapsulate_message_recursive(validated_encapsulated_message).inspect_err(|e| {
```
youngjoon-lee, line 1077: I agree that this is a beneficial behavior as a core node. Any core node can, of course, try to decapsulate layers as many as possible to maximize this rewards that they can collect right now.
This behavior may reduce the traffic of the entire network, but I think it's fine. Core nodes can do whatever is the best for them. If a message sender doesn't want for this scenario to happen, he can encapsulate messages by choosing `N` different nodes somehow, though I'm not sure if it'd be possible (as mentioned in https://github.com/logos-co/nomos/pull/1891/files#r2501243756). 
So, anyway, I agree with having this logic here, and we don't need to remove this later even if we define the minimal network parameter `3` as a const (or a genesis param).

nomos-blend/message/src/encap/validated.rs
ntn-x2: The idea is that `from_message_unchecked` does not perform any checks, while `from_message` might in the future add additional checks, so I would rather keep it like this not to avoid the API in the future. That's also the reason why I simply did not implement `From<EncapsulatedMessage>` for this type, as I wanted the API to be explicit about what it assumes.

nomos-services/blend/src/core/mod.rs
ntn-x2: Unfortunately, having this logic is not trivial. The reason is that the only parameter when generating a PoQ that results in a different key nullifier (hence potentially a different recipient) is the key index. Even using a different ephemeral key won't affect the resulting key nullifier (I just wrote a test for it and it will be included in the PR that introduces actual core proofs). This means that the only way to effectively skip messages to self is by skipping a given key index altogether, which I don't think is a good thing to do since that would effectively reduce the node's rewards. More "exotic" approaches would involve use a leadership proof for when a core proof would be addressed to oneself,  but the leadership proof might actually lead to the same issue, although it exponentially reduces the chance of it happening, yet it's a case we should handle. At the moment, if we assume that we don't want to rely on a minimum network size of `3`, I don't see any other way that is also less complicated than this. From a privacy perspective, from my conversation with @madxor I understood this is not a big issue. Can you confirm this Marcin? That if we try to peel off layers from a locally generated data or cover message it won't be a big privacy issue? Especially since we treat the peeled off message as processed and we will then randomly delay it until the next release round.

nomos-services/blend/src/core/mod.rs
ntn-x2: Given my comment above, a node has limited flexibilty in choosing the intended recipients for its messages. It all boils down to the key nullifier of the resulting PoQ, which is only directly affected by the key index used, since all other public and private inputs are fixed for the whole duration of a session and the ephemeral key used does not affect the resulting key nullifier. As I mentioned above, skipping key indices or introducing even more complicated logic than this like switching to a leadership proof might do more harm than good, in my opinion. So unless it's a huge privacy risk, from implementation perspective I think this is the easiest way we can solve this. But maybe @madxor will have some remarks about this. For now, I will go ahead and merge, and we will re-evaluate if new information comes up.

youngjoon-lee: Ah, yes, nodes have limited flexibility due to PoSel. I forgot. Thank you.

PR #1893, opened 2025-11-04
feat(zk): update to new nomos-circuits structure

davidrusu:

nomos-circuits releases now have this structure:
```
nomos-circuits-{version}-{os}-{arch}/
├── VERSION
├── prover[.exe]
├── verifier[.exe]
├── pol/
│   ├── witness_generator[.exe]
│   ├── witness_generator.dat
│   ├── proving_key.zkey
│   └── verification_key.json
├── poq/
│   ├── witness_generator[.exe]
│   ├── witness_generator.dat
│   ├── proving_key.zkey
│   └── verification_key.json
├── zksign/
│   ├── witness_generator[.exe]
│   ├── witness_generator.dat
│   ├── proving_key.zkey
│   └── verification_key.json
└── poc/
    ├── witness_generator[.exe]
    ├── witness_generator.dat
    ├── proving_key.zkey
    └── verification_key.json
```

We now place this folder in ~/.nomos-circuits. This simplifies the process of finding the associated proving/verifying keys and executables.

We now provide a `scripts/setup-nomos-circuits.sh` script that will install nomos-circuits so that doing work with zk should be easier for developers.

yes

NA

NA

NA

## Checklist

* [x] 2. Description added.

### DIFF
--- a/.github/workflows/code-check.yml
+++ b/.github/workflows/code-check.yml
@@ -64,6 +64,10 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@85e6279cec87321a52edac9c87bce653a07cf6c2  # Version 4.2.2
+      - name: Setup Circuits
+        uses: ./.github/actions/setup-circuits
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Install cargo-hack
         uses: taiki-e/install-action@5651179950649c44da31d346537e20c0534f0f25  # Version 2.49.35
         with:


.github/workflows/code-check.yml
```diff
@@ -168,9 +168,6 @@ jobs:
 
       - name: Setup Circuits
         uses: ./.github/actions/setup-circuits
-        with:
```
ntn-x2, line 171: If we're interacting with GH, which we do in the script, we might need to keep this, else we get rate limited very fast.

zk/circuits/prover/src/rapidsnark.rs
ntn-x2, line 1: Is this file supposed to be here or is it a leftover?

davidrusu: It's showing up as a `binary` because we used the null byte \0 in code directly instead of '\0'. 

Any file that uses \0 is considered "binary" by git.

There is changes to this file but you just can't see them rendered unfortunately.

I fixed this, I replaced the \0 with the escaped `'\0'` so future diffs will render correctly, but this one is not visible :/

.github/workflows/code-check.yml
davidrusu: We no longer use a release downloader action, we just download the public release via curl in the setup script, so using the token is slightly more complicated. I could make it work but how about we simply cache the release instead of using the token?

ntn-x2: Not sure if from GH perspective there is a difference between the two operations. But we can try caching, or we could even try to leave it like this. If we see the action starts failing, we could take action.

davidrusu: https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-unauthenticated-users

Seems we get something like 60req/hr before we start hitting rate limits. I think we should be fine, if it's an issue I'll add a cache.

zk/proofs/poc/src/verification_key/mod.rs
```diff
@@ -13,9 +15,15 @@ impl AsRef<Groth16PreparedVerificationKey> for PoCVerifyingKey {
...
+    let vk_json = std::fs::read(&vk_path).unwrap_or_else(|error| {
+        panic!(
+            "Failed to read verification key from {}: {error}",
+            vk_path.display()
+        )
+    });
     let groth16_vk_json: Groth16VerificationKeyJsonDeser =
-        serde_json::from_slice(vk_json).expect("Key should always be valid");
+        serde_json::from_slice(&vk_json).expect("Key should always be valid");
     let groth16_vk: Groth16VerificationKey = groth16_vk_json
```
danielSanchezQ, line 20: This is wrong. There is a reason that we used `include_bytes` here and it is to avoid distributing external files if possible. The VKs should be part of the binary at the moment of building.

zk/proofs/poc/src/verification_key/verification_key.json
danielSanchezQ, line 1: Good to see it gone, but it should be part of the build system.

zk/circuits/utils/src/lib.rs
```diff
@@ -1,285 +1,117 @@
...
+
+    if proving_key_path.is_file() {
+        proving_key_path
+    } else {
+        panic!(
+            "Proving key not found at expected path: {}\n\
+             Please ensure your nomos-circuits directory has the correct structure for circuit '{circuit_name}'",
+            proving_key_path.display()
+        )
+    }
```
danielSanchezQ, line 91: I see your point but we just replaced general code for hardcoded values and duplicated code. There is barely no difference in all this calls (?).

zk/proofs/poc/src/verification_key/mod.rs
davidrusu: Does this extend to the zkey's as well? or just the verification key's?

zk/circuits/utils/src/lib.rs
davidrusu: The external api for using any one of the zk proofs doesn't change. But the details of how to setup a dev environment that will work with these proofs is where I think the biggest change is.

In nomos-circuits v0.1.0, you had to download each artifact, one by one, and then place them each in a location that the `find_file` function will look at.

With the change, we are less flexible about where individual binaries/zkey/vk's (we enforce that they are all in a single folder NOMOS_CIRCUITS) but that allows us to give a clear setup process and clear errors when something goes wrong.

zk/proofs/poc/src/verification_key/mod.rs
danielSanchezQ: > Does this extend to the zkey's as well? or just the verification key's?

Yes, the zkey too!

zk/circuits/utils/src/lib.rs
danielSanchezQ: Yeah, I think its fine. I would have reused what we had and just changed this for the binaries (as keys should be included in the build and they should fail at compile time).

strinnityk: I think the point @danielSanchezQ is trying to make is that these functions are coupled with circuits' stuff while the previous ones were parametrised (and the circuits stuff passed as parameters).

.github/workflows/code-check.yml
davidrusu: ok you're right, decided to just go ahead and do it properly. I've added the GITHUB_TOKEN back to the release downloads

danielSanchezQ (approved):
Nice build changes, this is perfect, thanks!

zk/proofs/poc/src/verification_key/mod.rs
davidrusu: discussed offline, zkeys are currently not embedded in the binary, and given that we shell out to a prover, the zkey would have to be written back out to a file in order to give it to the prover.

We decided to keep the existing behavior and leave zkeys as external files that are referenced at runtime.

pradovic (approved):

PR #1895, opened 2025-11-05
feat(blend): actual generation and verification logic for core proofs

ntn-x2:

We have real session data now, so we can run actual core PoQ and PoSel generation and verification logic. We still need to mock leader PoQs and related PoSels because we don't have PoL info. We will also switch to that once they are integrated in cryptarchia leader service.

I also changed the way proofs are generated. Instead of spawning a task and have complex teardown logic, we now create a stream and poll from it, with a buffer of 10 proofs, which is currently a `const`, but we can change it later on maybe based on the number of expected proofs per message.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nodes/nomos-node/node/src/generic_services/blend/proofs.rs
+++ b/nodes/nomos-node/node/src/generic_services/blend/proofs.rs
@@ -192,7 +192,8 @@ mod core_to_core_tests {
         core_and_leader::CoreAndLeaderProofsGenerator as _,
     };
     use nomos_blend_service::{ProofsVerifier as _, merkle::MerkleTree};
-    use nomos_core::{crypto::ZkHash, mantle::keys::SecretKey};
+    use nomos_core::crypto::ZkHash;
+    use zksign::SecretKey;
 
     use crate::generic_services::blend::{BlendProofsVerifier, CoreProofsGenerator};
 


nodes/nomos-node/node/src/generic_services/blend/proofs.rs
```diff
@@ -0,0 +1,487 @@
...
+    fn verify_proof_of_quota(
+        &self,
+        proof: ProofOfQuota,
+        signing_key: &Ed25519PublicKey,
+    ) -> Result<ZkHash, Self::Error> {
+        let key_nullifier = proof.key_nullifier();
+        tracing::debug!(target: LOG_TARGET, "Verifying PoQ with key nullifier: {key_nullifier:?}");
+        if proof.key_nullifier() == DUMMY_POQ_ZK_NULLIFIER {
+            tracing::debug!(target: LOG_TARGET, "Mocked PoL PoQ proof received (automatically verified successfully).");
+            Ok(DUMMY_POQ_ZK_NULLIFIER)
```
youngjoon-lee, line 141: When is this true? Is there still any case where this is true, even though we now use the real PoQ generator?

ntn-x2: When generating random leadership proofs, we always use the dummy nullifier and use random bytes for the proof part: https://github.com/logos-co/nomos/pull/1895/files#diff-9c19c24e25d1d55b1fff58803dee1b921473b122adeacfcd88cb570098eb93aeR89. So all mock proofs will always have the same dummy nullifier, and we can check that here.

ntn-x2: We generate real core proofs, but we still generate mock leadership proofs, since real data is not yet coming from the chain leader service.

Issue #1664, opened 2025-09-09
Fix bugs in rs.rs: test decode call, insufficient points check, and misleading encode docstring


megonen:
nomos-da/kzgrs/src/rs.rs

While auditing rs.rs, we identified the following issues:

**1. Test does not re-run decode after erasures**

In `test_encode_decode`, after setting half of the encoded evaluations to `None`, the code compares `bytes` against the old `decoded` result, without calling `decode` again on the modified input.

**2. `decode` does not check if enough points are present**

`decode` attempts to interpolate regardless of how many points are provided. If fewer than `original_chunks_len` points are available, the interpolation silently produces an incorrect polynomial.

**3. Misleading encode docstring**

The docstring for encode says:

"Extend a polynomial over some factor `polynomial.len()*factor` and return the original points plus the extra ones."

But the function does not accept a factor parameter; it simply evaluates over the provided domain.

Suggestion:

"Evaluate the polynomial over domain (`length = domain.size()`), producing a codeword of that length."

Optionally, assert that `domain.size() >= polynomial.degree() + 1` with a friendly error.

PR #1897, opened 2025-11-05
fix(da): Align rs decode domain

megonen:

> Solves issue #1664  by making decode_unchecked evaluate the reconstructed polynomial on the original (coefficient) domain instead of the extended (encode) domain.

> Yes. The issue describes the mismatch that happens when we encode on a larger domain (e.g. 32) but originally built the polynomial on a smaller domain (e.g. 16).

> @danielSanchezQ @megonen 

> Yes

> No. It aligns the code with the specification by ensuring decoding uses the same domain that was used to build the polynomial.

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-da/kzgrs-backend/benches/reconstruct.rs
+++ b/nomos-da/kzgrs-backend/benches/reconstruct.rs
@@ -51,9 +51,8 @@ fn reconstruct<const SIZE: usize>(bencher: Bencher, column_size: usize) {
                         .collect::<Vec<Option<Fr>>>()
                 });
                 let domain = GeneralEvaluationDomain::<Fr>::new(SIZE).unwrap();
-                let expanded_domain = GeneralEvaluationDomain::<Fr>::new(SIZE * 2).unwrap();
                 let data: Vec<u8> = rows
-                    .map(|row| decode_unchecked(SIZE, &row, domain, expanded_domain))
+                    .map(|row| decode_unchecked(SIZE, &row, domain, domain))
                     .flat_map(|evals| points_to_bytes::<31>(&evals.evals))
                     .collect();
                 assert_eq!(data, encoded.data);
--- a/nomos-da/kzgrs-backend/src/encoder.rs
+++ b/nomos-da/kzgrs-backend/src/encoder.rs
@@ -349,7 +349,6 @@ pub mod test {
                 assert_eq!(c1, c2);
             }
         }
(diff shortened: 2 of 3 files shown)


megonen:
## What changed

- `decode_unchecked(...)` now takes **two** domains:
  - `encode_domain`: the larger domain used for the extended / RS-style encoding.
  - `coeff_domain`: the original, smaller domain used when doing `bytes -> polynomial`.

## How it works now

1. Collect the available points from the **encode** domain (skipping `None`s).
2. Truncate to the original number of chunks.
3. Lagrange-interpolate to recover the polynomial.
4. Re-evaluate that polynomial on the **original (coeff)** domain.
5. Return those evaluations.

## Why

Before this change, when some encoded points were missing, we could still interpolate the right polynomial, but we then evaluated it again on the **large** encode domain and just sliced the result. That slice didn’t match the original byte layout, so the test failed.

Now we always re-evaluate on the original domain, so the decoded bytes match.

## Tests

- Tests now build two domains: `32` (encode) and `16` (coeff). For FFT, these must be the power of 2.
- They check the full roundtrip.
- They also drop every other point (simulate missing shares) and decode successfully.

danielSanchezQ (approved):

nomos-da/kzgrs-backend/benches/reconstruct.rs
```diff
@@ -51,8 +51,9 @@ fn reconstruct<const SIZE: usize>(bencher: Bencher, column_size: usize) {
                         .collect::<Vec<Option<Fr>>>()
                 });
                 let domain = GeneralEvaluationDomain::<Fr>::new(SIZE).unwrap();
+                let expanded_domain = GeneralEvaluationDomain::<Fr>::new(SIZE * 2).unwrap();
                 let data: Vec<u8> = rows
-                    .map(|row| decode(SIZE, &row, domain))
+                    .map(|row| decode_unchecked(SIZE, &row, domain, expanded_domain))
```
megonen: In decode_unchecked, the call order is: decode_unchecked(16, &encoded, *ENCODE_DOMAIN, *COEFF_DOMAIN);
so the smaller domain is the coeff domain. But in here, the second one is expanded domain which is the bigger one. Probably this must be map(|row| decode_unchecked(SIZE, &row, expanded_domain, domain)) ?
@danielSanchezQ

PR #1902, opened 2025-11-07
fix(blend): attempt to decapsulate messages with the old session crypto processor during STP

youngjoon-lee:

Closes https://github.com/logos-co/nomos/issues/1866

Please see details in the issue. 

In this PR, we maintain up to two processors at the same time. One is for the current session, and the other is for the old session (maintained only during STP).
Whenever we want to decapsulate a message, we try with both processors. Since PoQ and PoSel verifications are based on the session number (i.e. key nullifier), only one (or none) processor will successfully decapsulate it.
Messages successfully decapsulated by either of the two processors are sent to the message scheduler, as before.
(Currently, we uses the same scheduler for both sessions, but I will split it into two at a next PR.)

I also refactored the `run_event_loop` function by extracting the last logic as the `retire` function, because it's natural for this PR.

Diffs look large because of tests and refactoring, but please focus on the following parts:
- Maintaining `old_session_crypto_processor` in the `run_event_loop`.
- Trying to decapsulate messages with both processors in the `handle_incoming_blend_message`.
- Returning three different kinds of output from the `handle_session_event`.

The code structure might be not optimal because I'm thinking about refactorting with state machine. Nevertheless, I tried to make the code simple, safe, and testable.

I believe so

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/core/mod.rs
+++ b/nomos-services/blend/src/core/mod.rs
@@ -1176,23 +1176,21 @@ where
         ),
         Err(e) => {
             tracing::debug!(target: LOG_TARGET, "Failed to decapsulate received message with the current session crypto processor: {e:?}");
-            if let Some(old_crypto_processor) = old_session_cryptographic_processor {
-                match old_crypto_processor
-                    .decapsulate_message_recursive(validated_encapsulated_message)
-                {
-                    Ok(output) => schedule_decapsulated_incoming_message_and_collect_tokens(
-                        output,
-                        scheduler,
-                        blending_token_collector,
-                        current_recovery_checkpoint,
-                    ),
-                    Err(e) => {
-                        tracing::debug!(target: LOG_TARGET, "Failed to decapsulate received message with the old session crypto processor: {e:?}");
-                        current_recovery_checkpoint
-                    }
+            let Some(old_crypto_processor) = old_session_cryptographic_processor else {
+                return current_recovery_checkpoint;
+            };
+            match old_crypto_processor.decapsulate_message_recursive(validated_encapsulated_message)
+            {
+                Ok(output) => schedule_decapsulated_incoming_message_and_collect_tokens(
+                    output,
+                    scheduler,
+                    blending_token_collector,
+                    current_recovery_checkpoint,
+                ),
+                Err(e) => {
+                    tracing::debug!(target: LOG_TARGET, "Failed to decapsulate received message with the old session crypto processor: {e:?}");
+                    current_recovery_checkpoint
                 }
-            } else {
-                current_recovery_checkpoint
             }
         }
     }


nomos-services/blend/src/core/mod.rs
```diff
@@ -348,23 +348,54 @@ where
...
         let secret_pol_info_stream = post_initialize::<PolInfoProvider, _>(overwatch_handle).await;
 
-        // Run the main event loop of the service.
-        run_event_loop(
+        let mut blend_messages = backend.listen_to_incoming_messages();
+
+        // Run the main event loop while the node is a core node across multiple
+        // sessions. When the node becomes a non-core node in a new session, the
+        // components for the last session transition period are returned.
+        //
```
ntn-x2:
```suggestion
```

nomos-services/blend/src/core/mod.rs
```diff
@@ -695,77 +730,132 @@ async fn run_event_loop<
...
-                        crypto_processor = output.crypto_processor;
-                        current_public_info = output.public_info;
-                        current_recovery_checkpoint = output.recovery_checkpoint;
-                    }
-                    Err((e @ (Error::LocalIsNotCoreNode | Error::NetworkIsTooSmall(_)), output)) => {
-                        tracing::info!(
-                            target: LOG_TARGET,
-                            "New membership does not satisfy the core node condition: {e:?}:  Exiting from the main event loop",
+                match handle_session_event(session_event, blend_config, crypto_processor, public_info, recovery_checkpoint, blending_token_collector, backend).await {
+                    HandleSessionEventOutput::Transitioning { new_crypto_processor, old_crypto_processor, new_public_info, new_recovery_checkpoint } => {
```
ntn-x2, line 760: If we are passing the `new_public_info` here, we should avoid passing components that are also created from those public info, and we should instead create them inside using the public info received, else we risk getting out of sync between those components and the public info themselves. Let's just create everything internally and have the output just return the public info, and perhaps the stuff belonging to the old session, although we already have a reference to it in here, so I don't understand why we need it to be part of the event payload?

nomos-services/blend/src/core/mod.rs
```diff
@@ -695,77 +730,132 @@ async fn run_event_loop<
...
+                        crypto_processor = new_crypto_processor;
+                        old_session_crypto_processor = Some(old_crypto_processor);
+                        public_info = new_public_info;
+                        recovery_checkpoint = new_recovery_checkpoint;
+                    },
+                    HandleSessionEventOutput::TransitionCompleted { current_crypto_processor, current_public_info, current_recovery_checkpoint } => {
+                        crypto_processor = current_crypto_processor;
```
ntn-x2, line 767: Why do we need this if we are still in the same session?

nomos-services/blend/src/core/mod.rs
ntn-x2: Also I don't think the output should have knowledge about recovery state. Can't it just tell us what the new session values are?

nomos-services/blend/src/core/mod.rs
```diff
@@ -997,26 +1087,71 @@ where
...
+) -> ServiceState<BackendSettings, BroadcastSettings>
+where
+    NodeId: 'static,
+    BroadcastSettings: Serialize + for<'de> Deserialize<'de> + Eq + Hash + Clone + Send,
+    BackendSettings: Clone,
+    ProofsVerifier: ProofsVerifierTrait,
+{
+    // First, try to decapsulate with the current session crypto processor.
+    // If that fails, try with the old session crypto processor, if any.
+    match cryptographic_processor.decapsulate_message(validated_encapsulated_message.clone()) {
```
ntn-x2: Why don't we do what we do with other components and have this logic live inside the cryptographic processor by exposing `start_session_transition` and `complete_session_transition` functions? It would make all the codebase uniform in that case, and we should do the same for all components we want to keep around. I guess only for the retire case we would not call `rotate_session` but simply keep the old one around until the STP has elapsed.

ntn-x2 (commented):
Good to fix these issues, thank you very much for that. I think we should aim to have a uniform approach to how we handle session transitions. In our codebase we currently do let internal components handle the old and new state, and we should do the same for the cryptographic processor, and also for the message scheduler. So that once we move away from that to passing the stream directly, we would apply the change everywhere at the same level and not in the service for some levels and in the components in other places. For this reason, if we expose a consistent API on the cryptographic processor, I don't think we need the `ProcessedMessageScheduler` trait, in my opinion. Also, why does the output return data that we already have available in the context where we parse this output? We should either consume the input data and return it in the output, or let the output only return data we don't have access to from the context where such output is processed. What do you think?

nomos-services/blend/src/core/mod.rs
```diff
@@ -695,77 +730,132 @@ async fn run_event_loop<
     ProofsVerifier: ProofsVerifierTrait,
     RuntimeServiceId: Sync,
 {
-    // Main event loop that persists while the node is a core node.
-    // This can span across multiple sessions.
+    // An optional crypto processor to handle the old session during transition
+    // period.
+    let mut old_session_crypto_processor: Option<
+        CoreCryptographicProcessor<NodeId, ProofsGenerator, ProofsVerifier>,
+    > = None;
```
danielSanchezQ, line 739: I think it would be better to merge both crypto processors in a cryptor processor that handles sessions. It would be a wrapper of arbitrary cryptroprocessors and tries for each (?)

nomos-services/blend/src/core/mod.rs
```diff
@@ -695,77 +730,132 @@ async fn run_event_loop<
...
-                            "New membership does not satisfy the core node condition: {e:?}:  Exiting from the main event loop",
+                match handle_session_event(session_event, blend_config, crypto_processor, public_info, recovery_checkpoint, blending_token_collector, backend).await {
+                    HandleSessionEventOutput::Transitioning { new_crypto_processor, old_crypto_processor, new_public_info, new_recovery_checkpoint } => {
+                        crypto_processor = new_crypto_processor;
+                        old_session_crypto_processor = Some(old_crypto_processor);
+                        public_info = new_public_info;
+                        recovery_checkpoint = new_recovery_checkpoint;
+                    },
+                    HandleSessionEventOutput::TransitionCompleted { current_crypto_processor, current_public_info, current_recovery_checkpoint } => {
+                        crypto_processor = current_crypto_processor;
```
danielSanchezQ, line 767: I think this should be encapsulated in the crypto_processor itself (as per my comment above).

nomos-services/blend/src/core/mod.rs
ntn-x2: Yup, that's also what I suggested in https://github.com/logos-co/nomos/pull/1902#discussion_r2509388344. We do that for all our other components, so we should be consistent and do the same here. In the future, we will replace the explicit API with an input stream and let each component handle transitions as they need.

nomos-services/blend/src/core/mod.rs
youngjoon-lee: > Let's just create everything internally and have the output just return the public info, and perhaps the stuff belonging to the old session, although we already have a reference to it in here, so I don't understand why we need it to be part of the event payload?

I'm asking just to understand what you suggest. What do you mean about `internally`?
Inside the `handle_session_event`, we create all the components that are necessary for the new and old sessions. The `HandleSessionEventOutput` contains all of them. Of course, some components (such as crypto processor) can be made of `new_public_info` outside of the `handle_session_event`. But, to avoid the out-of-sync issue you mentioned, I created all the componenets inside the `handle_session_event`. 

> Also I don't think the output should have knowledge about recovery state. Can't it just tell us what the new session values are?

Acutally, the `new_recovery_checkpoint` was already there as the return value of the `handle_session_event`. I just include it into `HandleSessionEventOutput`.

nomos-services/blend/src/core/mod.rs
youngjoon-lee: It's because the `handle_session_event` consumes the existing crypto processor. Outside of the `handle_session_event`, we don't check which type of `SessionEvent` has been arrived. It's the role of `handle_session_event`. So, we pass all the components that may be recreated for the new/old sessions to the `handle_session_event`. Then, the `handle_session_event` recreates/returns components in some cases, or returns components as they are in some other cases.

nomos-services/blend/src/core/mod.rs
youngjoon-lee: As shortly discussed in Discord, I felt it's clearer to make the crypto processor live only for a single session because then we can keep the internal logic of the crypto processor simple. I thought the session transition is not what the crypto processor should handle. Its job is encapsulating/decapsulating messages, using the ingredients it has. 
Actually, I'm not against to having the `rotate_session` API for the crypto processor, because I already implemented it for the `BlendingTokenCollector`. But, I realized that the `MessageScheduler` takes a different approach which consumes a stream. As we discussed, in the future, we may be able to try to apply the stream-based approach to all components. But, for now, I'm not sure if it's the best way. And, I'm not sure if it won't have any timing issue. Maybe, I'm missing something from what you suggested. I think it's worth trying pair-programming for the refactoring to avoid miscommunications.

For now, I'd like to go with this PR. I think the approach is fine. After having everything in place, we can find the best structure.

nomos-services/blend/src/core/mod.rs
```diff
@@ -348,23 +348,53 @@ where
...
+        retire(
+            blend_messages,
+            remaining_clock_stream,
             remaining_session_stream,
             &blend_config,
             backend,
             network_adapter,
             epoch_handler,
+            // TODO: Use a scheduler that doesn't generate new cover messages
             message_scheduler,
```
ntn-x2, line 395: What is the point of hiding this behind a trait then? Don't we know at this point that when calling the `retire` we must have a scheduler that is only able to schedule remaining messages? That could be created with a method on the original session scheduler, consuming it and returning a different scheduler for the STP only.

nomos-services/blend/src/core/mod.rs
```diff
@@ -997,26 +1086,71 @@ where
...
+    match cryptographic_processor.decapsulate_message(validated_encapsulated_message.clone()) {
+        Ok(decapsulated_message) => schedule_decapsulated_incoming_message(
+            decapsulated_message,
+            scheduler,
+            blending_token_collector,
+            current_recovery_checkpoint,
+        ),
+        Err(e) => {
+            tracing::debug!(target: LOG_TARGET, "Failed to decapsulate received message with the current session crypto processor: {e:?}");
+            if let Some(old_crypto_processor) = old_session_cryptographic_processor {
```
ntn-x2: We can save on some indentation :)

```suggestion
            let Some(old_crypto_processor) = old_session_cryptographic_processor else {
```

ntn-x2 (approved):
After the discussion we had, I think we can go ahead with the current approach. I am then approving. But I think we could fine a way to split the `lib.rs` file into smaller modules, if possible, since it's definitely way too large now. Good job tho, thank you!

nomos-services/blend/src/core/mod.rs
youngjoon-lee: Yes. The PR https://github.com/logos-co/nomos/pull/1913 resolves this TODO by changing this `retire` function to accept the `OldSessionMessageScheduler` explicitly. Also, `OldSessionMessageScheduler` is  created by `SessionMessageScheduler::to_old_session()`.
But, we still need the `MessageScheduler` trait because it's used for the `schedule_decapsulated_incoming_message_and_collect_tokens` function. That function sometimes accepts the current scheduler, and sometimes accepts the old scheduler. We could refactor the core service to clearly split logics between current and current+old, but in the current codebase, using this trait is the simplest way, I think.

nomos-services/blend/src/core/mod.rs
youngjoon-lee: Done d33725be

PR #1899, opened 2025-11-06
feat(da-sampling): support multiple sesssions in historic sampling

pradovic:
## 1. What does this PR implement?

This PR adds support for multiple sessions in historical sampling requests, as blobs might be from several (probably at most 2) different sessions.

## 2. Does the code have enough context to be clearly understood?

Yes

## 3. Who are the specification authors and who is accountable for this PR?

@pradovic 

## 4. Is the specification accurate and complete?

Yes

## 5. Does the implementation introduce changes in the specification?

No

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [ ] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [ ] 2. Description added.
* [ ] 3. Context and links to Specification document(s) added.
* [ ] 4. Main contact(s) (developers and specification authors) added
* [ ] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [ ] 6. Link PR to a specific milestone.

PR #1907, opened 2025-11-07
fix(tests): Serialize blob and session hashmap as vec

bacv:

In https://github.com/logos-co/nomos/commit/a29aa048a3279a28c3ce693821a9c4637ce57392 historical sampling test was unignored, and in https://github.com/logos-co/nomos/pull/1899 it was changed to a request type that contain blob id as hashmap key, json does not allow non strings as a map key and test started failing.

The workaround is to encode historical sampling post request as a `Vec<(blob_id, session)>`.

Yes

@pradovic 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nodes/nomos-node/node/src/api/testing/handlers.rs
+++ b/nodes/nomos-node/node/src/api/testing/handlers.rs
@@ -1,5 +1,4 @@
 use std::{
-    collections::HashMap,
     fmt::{Debug, Display},
     hash::Hash,
 };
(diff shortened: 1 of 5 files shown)


nodes/nomos-node/node/src/api/testing/handlers.rs
```diff
@@ -66,12 +67,14 @@ where
     >(handle, session_id))
 }
 
+#[serde_as]
```
pradovic: Maybe: `use serde_with::{serde_as, Same};`

bacv: Just change it directly to `Vec<(BlobId, SessionNumber)>` as it's only used for the post request.

PR #1908, opened 2025-11-07
feat(da-sampling): sampling falls back to historic sampling on session change

pradovic:

This PR attempts to solve the edge case:
1. Tx is posted by mempool
2. Trigger sampling in sampling service
3. Session changes before sampling starts

In this case we compare the session from the blob (in the sampling behavior) and return error session mismatch. Then, sampling service falls back to historical sampling to sample and verify this blob.

The edge case that is not solved is get_commitments call which gets commitments with provided session, but expects current session. The sampling would return mismatch in this case. Maybe we should also expand historic behavior to support fetching only commitments for specified session, in this case.

Yes

@pradovic @bacv 

Yes

No
## Checklist

> [!WARNING]  

* [x] 2. Description added.


nodes/nomos-node/node/src/api/handlers.rs
```diff
@@ -644,7 +651,7 @@ where
         SamplingMempoolAdapter,
         SamplingStorage,
         RuntimeServiceId,
-    >(&handle, blob_id))
+    >(&handle, req.blob_id, req.session))
```
danielSanchezQ, line 649: Is this in this pr because of testing?

nomos-services/data-availability/sampling/src/lib.rs
```diff
@@ -287,61 +299,155 @@ where
         sampler: &mut SamplingBackend,
         storage_adapter: &SamplingStorage,
         verifier: &ShareVerifier,
-    ) {
+        network_adapter: &SamplingNetwork,
+    ) -> Option<(
+        BoxFuture<'static, ()>,
+        BoxFuture<'static, HistoricFallbackResult>,
```
danielSanchezQ: Nit: this boxed futures should be a named typealias

nomos-services/data-availability/sampling/src/lib.rs
```diff
@@ -34,20 +37,24 @@ use serde::{Deserialize, Serialize};
...
+
 struct PendingTasks<'a> {
     long_tasks: &'a mut FuturesUnordered<BoxFuture<'static, ()>>,
     sampling_continuations:
-        &'a mut FuturesUnordered<BoxFuture<'static, (BlobId, Option<DaSharesCommitments>)>>,
-    delayed_sdp_sampling_triggers: &'a mut FuturesUnordered<BoxFuture<'static, BlobId>>,
+        &'a mut FuturesUnordered<BoxFuture<'static, (Blob, Option<DaSharesCommitments>)>>,
+    delayed_sdp_sampling_triggers: &'a mut FuturesUnordered<BoxFuture<'static, Blob>>,
+    historic_fallback_continuations:
+        &'a mut FuturesUnordered<BoxFuture<'static, HistoricFallbackResult>>,
```
danielSanchezQ, line 59: We should use the same named type alias for futures here too.

nomos-services/data-availability/sampling/src/lib.rs
```diff
@@ -287,61 +299,155 @@ where
...
+    ) {
+        info_with_id!(blob_id, "SamplingSuccess");
+
+        let Some(commitments) = sampler.get_commitments(&blob_id) else {
+            error_with_id!(blob_id, "Error getting commitments for blob");
+            sampler.handle_sampling_error(blob_id).await;
+            return;
+        };
+
+        if verifier.verify(&commitments, &light_share).is_err() {
```
danielSanchezQ, line 342: We need to check on the shares verifier, its a quick operation, but we may need to use the blocking pool for them at some point if we have a lot, so we do not block the async thread pool.
This is something to consider for the future.

nomos-services/data-availability/sampling/src/lib.rs
```diff
@@ -287,61 +299,155 @@ where
...
+                    sampler,
+                    network_adapter,
+                    verifier,
+                )
+                .await;
+            }
+            _ => {
+                error!("Error while sampling: {error}");
+            }
+        }
```
danielSanchezQ, line 383: We should log the session mismatch and the retry here with the blob id and the session itself. For sanity on debugging.

nomos-services/data-availability/sampling/src/lib.rs
```diff
@@ -287,61 +299,155 @@ where
...
+        blob_id: BlobId,
+        session: SessionNumber,
+        sampler: &mut SamplingBackend,
+        network_adapter: &SamplingNetwork,
+        verifier: &ShareVerifier,
+    ) -> Option<(
+        BoxFuture<'static, ()>,
+        BoxFuture<'static, HistoricFallbackResult>,
+    )> {
+        tracing::warn!("Session mismatch for {blob_id:?}, falling back to historic sampling");
```
danielSanchezQ: Ah, log was here, good. Lets add the intended session number as well (?)

nomos-services/data-availability/sampling/src/lib.rs
```diff
@@ -447,11 +559,106 @@ where
...
+        timeout: Duration,
+    ) -> Option<BoxFuture<'static, ()>> {
+        let mut blob_ids = HashMap::new();
+        blob_ids.insert(blob_id, session);
+        let blobs: HashSet<BlobId> = [blob_id].into();
+
+        let Ok(historic_stream) = network_adapter.listen_to_historic_sampling_messages().await
+        else {
+            let _ = result_sender.send(None);
+            return None;
```
danielSanchezQ, line 621: Lets log the failure here (?)

nodes/nomos-node/node/src/api/handlers.rs
pradovic: @danielSanchezQ  No. This is the consequence of the sampling now requiring the session so we can compare it with current session. This GetCommitments call remains unsolved, where system will return an MismatchError to the user in the edge-case of the wrong session. 
My plan is to open the next PR with the fallback to historical in the commitments case as well. In order to do this I need to expand historic sampling behavior, as currently it is fetching everything together (commitments and shares), and in this case we would only need to fetch commitments.

Does that makes sense?

nomos-services/data-availability/sampling/src/lib.rs
pradovic: Ah yes, good catch, there was a TODO comment about this before which I accidentally erased while refactoring. I will bring back the comment for now

PR #1909, opened 2025-11-08
refactor(blend): non-optional membership for network behaviours

youngjoon-lee:

We had `Option<Membership>` in `NetworkBehaviuor`s. We set `None` to it only when testing. But, having it optional is not intuitive and not convenient to add more logic. 
So, this PR makes it non-optional by refactoring tests. 

Diffs look large, but don't worry. Please focus on test utils that build memberships, and the `NetworkBehaviour`s that now hold the mandatory `Membership`.

I checked that all the tests in the `nomos-blend-network` and `nomos-blend-service` crates pass.

Yes

@youngjoon-lee @ntn-x2 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.


nomos-blend/network/src/core/with_core/behaviour/tests/utils.rs
```diff
@@ -45,41 +48,63 @@ impl IntervalProviderBuilder {
...
+    count: usize,
+) -> (impl Iterator<Item = Keypair>, Vec<Node<PeerId>>) {
+    let mut identities: Vec<Keypair> = repeat_with(Keypair::generate_ed25519).take(count).collect();
+    identities.sort_by_key(|id| PeerId::from(id.public()));
+    let nodes = identities
+        .iter()
+        .map(|identity| Node {
+            id: identity.public().into(),
+            address: Multiaddr::empty(),
+            public_key: Ed25519PrivateKey::generate().public_key(),
```
ntn-x2: Should we not use the public key of the identity instead?

nomos-blend/network/src/core/with_core/behaviour/tests/utils.rs
```diff
@@ -45,41 +48,63 @@ impl IntervalProviderBuilder {
...
+        }
     }
 
-    pub fn with_local_peer_id(mut self, local_peer_id: PeerId) -> Self {
-        assert!(
-            self.identity.is_none(),
-            "Cannot set local peer ID when identity keypair is also set."
-        );
-        self.local_peer_id = Some(local_peer_id);
+    pub fn with_membership(mut self, nodes: &[Node<PeerId>]) -> Self {
```
ntn-x2, line 99: I think we should rather make this explicit and split it into two different methods. One for membership that we can assume has the local node in it (and panic if it does not), and another which we assume does not (and panic if it does). It would make tests also easier to read, in my opinion.

nomos-blend/network/src/core/with_edge/behaviour/tests/utils.rs
```diff
@@ -22,9 +22,13 @@ pub struct BehaviourBuilder {
 }
 
 impl BehaviourBuilder {
-    pub fn with_core_peer_membership(mut self, core_peer_id: PeerId) -> Self {
-        self.core_peer_ids.push(core_peer_id);
-        self
+    pub fn new(core_peer_id: PeerId) -> Self {
```
ntn-x2, line 24: I guess if we have `new` we should remove the `Default` impl then?

nomos-blend/network/src/core/tests/utils.rs
```diff
@@ -38,12 +39,36 @@ impl<Behaviour> TestSwarm<Behaviour>
...
+    /// Creates [`TestSwarm`] with an ephemeral [`Keypair`] generated randomly.
+    pub fn new_ephemeral<BehaviourConstructor>(behaviour_fn: BehaviourConstructor) -> Self
     where
         BehaviourConstructor: FnOnce(Keypair) -> Behaviour,
     {
         Self(Swarm::new_ephemeral_tokio(behaviour_fn))
     }
+
+    /// Creates [`TestSwarm`] with a specific [`Keypair`].
+    pub fn new<BehaviourConstructor>(identity: &Keypair, behaviour_fn: BehaviourConstructor) -> Self
```
ntn-x2: Do we have the need to pre-generate a public key instead of letting `new_ephemeral_tokio` generate one and pass it to the constructor?

nomos-services/blend/src/core/backends/libp2p/swarm.rs
```diff
@@ -531,7 +533,9 @@ where
         + 'static,
 {
     #[cfg(test)]
+    #[expect(clippy::too_many_arguments, reason = "necessary for testing")]
     pub fn new_test<BehaviourConstructor>(
+        identity: &identity::Keypair,
```
ntn-x2: Personal preference, instead of feature-gating the import, we could just use the fully-qualified name here instead.

nomos-services/blend/src/core/backends/libp2p/tests/network_maintenance.rs
```diff
@@ -1,56 +1,48 @@
...
         swarm: mut second_swarm,
         ..
-    } = SwarmBuilder::default()
-        .build(|id| BlendBehaviourBuilder::new(&id, MockProofsVerifier).build());
-    let (membership_entry, _) = second_swarm.listen_and_return_membership_entry(None).await;
+    } = SwarmBuilder::new(identities.next().unwrap(), &nodes).build(|id, membership| {
+        BlendBehaviourBuilder::new(id, MockProofsVerifier, membership).build()
+    });
+    let (second_node, _) = second_swarm.listen_and_return_membership_entry(None).await;
+    update_nodes(&mut nodes, &second_node.id, second_node.address);
```
ntn-x2, line 34: I am curious. Why do you think this approach is better than what we had before?

ntn-x2 (commented):
Removing the optional membership is something we had to do, so thank you very much! I am not sure I understand why we needed to refactor all utility test functions to pre-generate identities instead of relying on the randomly generated ones by `neq_ephemeral_tokio` constructor?

nomos-blend/network/src/core/with_core/behaviour/tests/utils.rs
youngjoon-lee: `identity` is for deriving `PeerId`. The `public_key` is for the message encapsulation/decapsulation.

nomos-blend/network/src/core/tests/utils.rs
youngjoon-lee: Yes. Now, our NetworkBehaviour implementation requires a membership when it's initialized, because this PR makes it non-optional. An important point here is that, once a membership is passed to the NetworkBehaviour, it can't be updated/replaced. So, at the beginning of each test, we should build a membership with all nodes we need, in advance. If we rely on auto-generated keys by `new_ephemeral_tokio`, we cannot build a membership before initiating any Swarm.

nomos-blend/network/src/core/with_core/behaviour/tests/utils.rs
youngjoon-lee: Could you help me understand which part of the tests felt hard to read for you? In each test, we’ll have a list of core nodes and a local peer ID, and given those values, I just want to build a membership regardless of whether the local node is included or not. I didn’t see any particular need to split this function, but I’d appreciate it if you could share a bit more about your intention.

nomos-blend/network/src/core/with_edge/behaviour/tests/utils.rs
youngjoon-lee: Yes. I forgot to remove it. 237f65b6

nomos-services/blend/src/core/backends/libp2p/tests/network_maintenance.rs
youngjoon-lee: I didn't make this change because I think it's better than before. I did it because I had to. 
As explained in https://github.com/logos-co/nomos/pull/1909/files#r2512582671, we have to build a membership before initiating any Swarm. But, we can get a listening address of a certain Swarm only after the Swarm is initiated. So, I build a membership with empty addresses beforehands, pass it to the 1st Swarm, and wait for the listening address. And then, I update the membership with the listening address received, and pass it to the 2nd Swarm, so that the 2nd Swarm can know the address of the 1st Swarm.

nomos-services/blend/src/core/backends/libp2p/swarm.rs
youngjoon-lee: I agree. 5ee30c5b

nomos-blend/network/src/core/with_core/behaviour/tests/utils.rs
ntn-x2: It's just a matter of expressing intentions with function names. We are dealing with a builder, which we initialize with an initial local peer ID, and we then pass a membership to it. Having explicit functions would allow us indeed to panic in case we are trying to build a membership assuming the local node would be present, but it is not. Yes, we would also catch it in the tests most likely, but it was just a matter of clarity. But no need to do this, it was a personal recommendation.

nomos-services/blend/src/core/backends/libp2p/tests/network_maintenance.rs
ntn-x2: Ok, thanks for the explanation.

nomos-blend/network/src/core/with_core/behaviour/tests/utils.rs
youngjoon-lee: Now I got what you mean. Yes. Let's use the same key: 1bcdbe50a9adb93dbf8ff9a87e7ae752d64428e9

nomos-blend/network/src/core/with_core/behaviour/tests/utils.rs
youngjoon-lee: I simplified the `with_membership` even more d03ddfbbe802c9f906fb5cccc0bef8f49d2a9eab. Thanks.

Issue #1886, opened 2025-11-04
Release processed messages for the old session through old connections


youngjoon-lee:
Currently, we receive messages from old connections during STP, but they are released through the connections of the new session after they are processed. 
They should be released through old connections.

For that, we need to change two things:
- Keep the message scheduler of the old session, and use it for messages received from the old session during STP.
- Add a feature to the Backend to release messages through the old session

PR #1911, opened 2025-11-10
fix(blend): release old session messages via old connections

youngjoon-lee:

Closes https://github.com/logos-co/nomos/issues/1886
(Stacked on the PR #1902)

Previously, all (processed) messages have been sent through the connections in the current (new) session even during STP. But, processed messages originated from the old session must be released through the old session connections. This PR fixes it. See the [`validate_and_public_message`](https://github.com/logos-co/nomos/commit/af1777b01972c21f0b66410bb0cae6976854856e#diff-2a6abfe9a1ac517407672a6214af0382d69d53489549213c5a41f927b07c5dbdR790) in the `core/with_core/behaviour`.

Also, I refactored [`forward_validated_message`](https://github.com/logos-co/nomos/commit/af1777b01972c21f0b66410bb0cae6976854856e#diff-2a6abfe9a1ac517407672a6214af0382d69d53489549213c5a41f927b07c5dbdR298) just for readability. 

I improved the tests in the `core/with_core/behaviour/tests/session.rs` to test the changes of this PR.

Yes

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.


nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -677,11 +678,11 @@ impl<ProofsVerifier, ObservationWindowClockProvider>
     ) -> Option<NegotiatedPeerState> {
         let peer_details = self.negotiated_peers.get_mut(&peer_id)?;
         // We double check we are dealing with the expected connection.
-        debug_assert!(
-            peer_details.connection_id == connection_id,
-            "Stored connection ID and provided connection ID do not match for the provided peer ID."
-        );
-
+        // This could be false if `connection_id` is from the old session.
+        if peer_details.connection_id != connection_id {
```
ntn-x2, line 683: Don't we want to update the state for a peer in the old connection as well? How do we deal in general with connection monitoring for the old session during the transition period?

ntn-x2: Since we always try with both current and old session, I think we could do that here as well.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -856,8 +861,10 @@ where
             return;
         };
         // Verify the message public header, or else mark the peer as malicious: https://www.notion.so/nomos-tech/Blend-Protocol-Version-1-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81859cebf5e3d2a5cd8f.
-        let Ok(validated_message) =
-            self.validate_encapsulated_message_public_header(deserialized_encapsulated_message)
+        let Ok(validated_message) = self
```
ntn-x2, line 865: Why don't we need to try to verify it with the old session? We still want to receive messages from the old session, right?

nomos-blend/network/src/core/with_core/behaviour/tests/utils.rs
```diff
@@ -88,7 +112,20 @@ impl BehaviourBuilder {
...
+        self.session = Some(session);
+        self
+    }
+}
+
+impl<ProofsVerifier> BehaviourBuilder<ProofsVerifier>
+where
+    ProofsVerifier: encap::ProofsVerifier,
+{
+    pub fn build(self) -> Behaviour<ProofsVerifier, IntervalProvider> {
```
ntn-x2: I think we can specify the proof verifier generic here instead of having it in the struct and basically never use it. What do you think?

ntn-x2: Cause we use it as phantom data, so I am not sure what we use use it for other than returning a behaviour that contains it, so a generic on this function should be enough.

nomos-blend/network/src/core/with_core/behaviour/tests/utils.rs
```diff
@@ -112,7 +149,19 @@ impl BehaviourBuilder {
                 .minimum_network_size
                 .unwrap_or_else(|| 1usize.try_into().unwrap()),
             old_session: None,
-            poq_verifier: AlwaysTrueVerifier,
+            poq_verifier: ProofsVerifier::new(PoQVerificationInputsMinusSigningKey {
```
ntn-x2: Do you think we should expose a method on the builder to specify these inputs, and use the default if no value is passed? Else this assumes that all verifiers will only work with `ZkHash::ZERO` values, but that is not necessarily true. So we should expose a function to specify these values.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: This is actually what we need to discuss (maybe with Marcin). In this PR, we ignore monitoring events from the old session. One big reason is that, even if we detects a malicious connection and closes it, we cannot create another connection in our current codebase, because our NetworkBehaviour implementation rejects incoming connection establishments if the dialer is not a core node in the "current" session. In other words, we currently don't support establishing new connections for the old session.
So, I think it's not meaningful to monitor connections in the old session. It might be meaningful to close malicious connections, but I wanted to address this topic as a separate discussion and as a separate PR. Before making any decision, I didn't want to add any temporary logic in this PR.

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: We try to verify and handle the message with the old session at the beginning of this function (line 825). When making this change, I also felt that this function is not easy to read. We can refactor this later after discussing the best structure that we're talking about.

nomos-blend/network/src/core/with_core/behaviour/tests/utils.rs
youngjoon-lee: You're right. Fixed b93479b874022177c0814665a4bb3547b2a3a234

nomos-blend/network/src/core/with_core/behaviour/tests/utils.rs
youngjoon-lee: I did like this to make tests short, but I agree with your points. Fixed 30eac958

nomos-blend/network/src/core/with_core/behaviour/mod.rs
youngjoon-lee: Also, the other parts of the existing codebase also doesn't perform monitoring for the old session. For example, we monitor spammy connections for the current session in the following code:
https://github.com/logos-co/nomos/blob/30eac958fe84c612053a722ef38e2bcb58c4c688/nomos-blend/network/src/core/with_core/behaviour/mod.rs#L847
But we don't do it for the old session:
https://github.com/logos-co/nomos/blob/blend-release-old-msgs-to-old-conns/nomos-blend/network/src/core/with_core/behaviour/old_session.rs#L227

If we want to do monitoring for the old session as well, we can implement it as a separate PR. I think it's fine to not monitor connections for the old session for now because the connections will be closed after STP anyway.

Issue #1912, opened 2025-11-10
Use a separate message scheduler for old session during STP


youngjoon-lee:
Currently we use the same message scheduler for both current and old sessions during STP. And, the scheduler resets scheduled processed messages when a new session starts. 

So, we should create a separate message scheduler dedicated to the old session when a new session starts (when a STP starts). This dedicated scheduler should just handle processed messages received from the old session. This shouldn't generate any new message.

PR #1913, opened 2025-11-10
fix(blend): use a separate message scheduler for old session

youngjoon-lee:

Closes https://github.com/logos-co/nomos/issues/1912
(Stacked on the PR #1902)

Previously we used the same message scheduler for both current and old sessions during STP. And, the scheduler reset scheduled processed messages when a new session starts. That's wrong because the messages that were processed/scheduled at the old session are discarded even though we have to process a STP. 

So, this PR creates a separate message scheduler dedicated to the old session when a new session starts (when a STP starts). This dedicated scheduler handles processed messages received from the old session. This never generate any new message.

Also, I renamed `MessageScheduler` -> `SessionMessageScheduler` by refactoring it to live only during a single session. The previous `MessageScheduler` consumed new session info from the stream, and I remember what we discussed last week about this. But, while implementing this PR, I thought it's simpler to make it session-specific. Also, I think it's safer, considering the case in which we may forget what to be implemented for STP inside the `MessageScheduler`. Actually, that's the case that this PR tries to resolve.

I believe so.

@youngjoon-lee @ntn-x2 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/core/mod.rs
+++ b/nomos-services/blend/src/core/mod.rs
@@ -749,6 +749,7 @@ where
                         crypto_processor = current_crypto_processor;
                         old_session_crypto_processor = None;
                         message_scheduler = current_scheduler;
+                        old_session_message_scheduler = None;
                         public_info = current_public_info;
                         recovery_checkpoint = current_recovery_checkpoint;
                     },


nomos-services/blend/src/core/mod.rs
```diff
@@ -735,42 +703,58 @@ where
...
                 handle_new_secret_epoch_info(pol_info.poq_private_inputs, &mut crypto_processor);
             }
             Some(session_event) = remaining_session_stream.next() => {
-                match handle_session_event(session_event, blend_config, crypto_processor, public_info, recovery_checkpoint, blending_token_collector, backend).await {
-                    HandleSessionEventOutput::Transitioning { new_crypto_processor, old_crypto_processor, new_public_info, new_recovery_checkpoint } => {
+                match handle_session_event(session_event, blend_config, crypto_processor, message_scheduler, public_info, recovery_checkpoint, blending_token_collector, backend, rng.clone()).await {
+                    HandleSessionEventOutput::Transitioning { new_crypto_processor, old_crypto_processor, new_scheduler, old_scheduler, new_public_info, new_recovery_checkpoint } => {
                         crypto_processor = new_crypto_processor;
                         old_session_crypto_processor = Some(old_crypto_processor);
+                        message_scheduler = new_scheduler;
```
ntn-x2, line 743: I guess the same comment as for the crypto processor applies here: why do we need to keep track of two schedulers instead of having all the logic internally, as we do for, for instance, our libp2p swarm and behaviours? Or our proof verifiers?

nomos-services/blend/src/core/mod.rs
```diff
@@ -868,11 +848,15 @@ async fn retire<
 /// on a new session with its new membership. It also creates new public inputs
 /// for `PoQ` verification in this new session. It ignores the transition period
 /// expiration event and returns the previous cryptographic processor as is.
+#[expect(clippy::too_many_arguments, reason = "necessary for session handling")]
 async fn handle_session_event<
     NodeId,
     ProofsGenerator,
     ProofsVerifier,
     Backend,
+    Scheduler,
```
ntn-x2, line 859: As for the previous PR (but please correct me if my understanding is incorrect), I think having traits for this feature is overkill and makes the code less easy to follow. Unless there is a valid reason to do this (that I have missed), I am wondering why we cannot go with the typical approach of having a component exposing methods for managing sessions and exposing a clearly defined API that the service can consume.

ntn-x2 (commented):
I left some comments which are closely related to the PR updating the crypto processor. I would like to discuss the approach taken, and then once we have a shared understanding and agreement I can give a second round of review for the remaining bits.

nomos-services/blend/src/core/mod.rs
youngjoon-lee: I did like this with the same reason I described in https://github.com/logos-co/nomos/pull/1902#discussion_r2511160379. IMO, this looks more intuitive, but we can discuss the best structure soon later. Sorry for the structure of `MessageScheduler` you implemented, but I wanted to apply the same pattern to this, crypto processor (https://github.com/logos-co/nomos/pull/1902), and token collector (https://github.com/logos-co/nomos/pull/1914)

nomos-services/blend/src/core/mod.rs
youngjoon-lee: Actually, these generic params (`Scheduler` and `OldScheduler`) are for injecting mock schedulers when testing this function `handle_session_event`. No reason other than that.

ntn-x2 (approved):
Now I understand much better your intentions. Thank you!

PR #1914, opened 2025-11-10
fix(blend): collect old session tokens with the correct collector

youngjoon-lee:

Closes https://github.com/logos-co/nomos/issues/1854
(Stacked on the PR https://github.com/logos-co/nomos/pull/1902).

Previously, blending tokens from the old session have been collected to the set for the new session.

This PR fixed it by refactoring the `BlendingTokenCollector` into two separate structs (`SessionBlendingTokenCollector` and `OldSessionBlendingTokenCollector`), which are intuitive and follow the same pattern introduced in other PRs (https://github.com/logos-co/nomos/pull/1902 and https://github.com/logos-co/nomos/pull/1913).

You may feel that the `run_event_loop` is verbose. I will refactor it by introducing state machine or any other way soon.

I believe so.

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/message/src/reward/mod.rs
+++ b/nomos-blend/message/src/reward/mod.rs
@@ -8,131 +8,109 @@ pub use activity::ActivityProof;
 use nomos_core::sdp::SessionNumber;
 pub use session::SessionInfo;
 pub use token::BlendingToken;
-use tracing::warn;
 
-use crate::reward::session::SessionRandomness;
+pub use crate::reward::session::SessionRandomness;
 
 const LOG_TARGET: &str = "blend::message::reward";
 
-/// Collects blending tokens for the current session, while retaining those from
-/// the previous session which are used to produce an activity proof.
-pub struct BlendingTokenCollector {
-    current_session_tokens: BlendingTokensForSession,
-    current_session_randomness: SessionRandomness,
-    previous_session_tokens: Option<BlendingTokensForSession>,
+pub trait BlendingTokenCollector {
+    fn collect(&mut self, token: BlendingToken);
 }
 
-impl BlendingTokenCollector {
-    #[must_use]
-    pub fn new(current_session_info: &SessionInfo) -> Self {
-        Self {
-            current_session_tokens: BlendingTokensForSession::new(
-                current_session_info.session_number,
-                current_session_info.token_count_byte_len,
-                current_session_info.activity_threshold,
-            ),
-            current_session_randomness: current_session_info.session_randomness,
-            previous_session_tokens: None,
-        }
-    }
-
-    /// Collects a blending token from the current session.
-    pub fn insert(&mut self, token: BlendingToken) {
-        self.current_session_tokens.insert(token);
-    }
-
-    /// Computes an activity proof for the previous session by consuming
-    /// blending tokens collected during that session.
-    ///
-    /// It returns `None` if there was no blending token collected during
-    /// the previous session, or if there is no blending token satisfying the
-    /// activity threshold.
-    pub fn compute_activity_proof_for_previous_session(&mut self) -> Option<ActivityProof> {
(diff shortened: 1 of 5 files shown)


nomos-blend/message/src/reward/mod.rs
```diff
@@ -8,132 +8,110 @@ pub use activity::ActivityProof;
...
 
 const LOG_TARGET: &str = "blend::message::reward";
 
-/// Collects blending tokens for the current session, while retaining those from
-/// the previous session which are used to produce an activity proof.
-pub struct BlendingTokenCollector {
-    current_session_tokens: BlendingTokensForSession,
-    current_session_randomness: SessionRandomness,
-    previous_session_tokens: Option<BlendingTokensForSession>,
+pub trait BlendingTokenCollector {
```
ntn-x2, line 16: Do we need to define a trait for this behaviour? We know what we are dealing with here. What is your opinion on this?

nomos-services/blend/src/core/mod.rs
```diff
@@ -1120,12 +1144,15 @@ where
         ),
         Err(e) => {
             tracing::debug!(target: LOG_TARGET, "Failed to decapsulate received message with the current session crypto processor: {e:?}");
-            if let Some(old_crypto_processor) = old_session_cryptographic_processor {
+            if let (Some(old_crypto_processor), Some(old_session_blending_token_collector)) = (
+                old_session_cryptographic_processor,
+                old_session_blending_token_collector,
+            ) {
                 match old_crypto_processor.decapsulate_message(validated_encapsulated_message) {
```
ntn-x2: Yes I think I am more and more convinced that if we need to differentiate between new and old sessions, we should probably do so inside the component itself. I.e., the cryptographic processor should return information or should expose in the API a way to differentiate whether an operation should be done in the current or in the old session, or maybe return a type that internally has this info, so that when it is passed back to it from the service it already knows what to do with it, without the service needed to know. E.g., it returns a decapsulated message that belongs to an old session, so when the service decapsulates it and sends it back to the processor, the processor knows it belongs to the old session. But, in general I am conflicted about this. Your approach is mostly type safe, but very very hard to follow, because of our current service structure. So I am also open to follow this approach instead, but refactor the whole service into a state machine so it's clear what each state does, as we already discussed.

ntn-x2: But in general, unless we need the service to know and differentiate between the two components, I would rather see us take the same approach we have taken for other components about session transition, and then refactor everything into some other coherent version that relies on a state machine to handle the different states. I can see how that will look like, and it would make sense to have different types for different component states, but at the moment this is a mixed approach which I think is quite confusing. At the same time, I am also aware there are not many options for us to tackle this session transition properly at the moment, so it might also be ok to leave things as you have designed them to be.

nomos-blend/message/src/reward/mod.rs
youngjoon-lee: Now we have two kinds of collector. `SessionBlendingTokenCollector` and `OldSessionBlendingTokenCollector`. Their common feature is the `collect` method. 
 In the upper layer (core service), I wanted to pass them to the `schedule_decapsulated_incoming_message` function that accepts a collector and calls only its `collect` method, regardless of its actual type.

nomos-services/blend/src/core/mod.rs
youngjoon-lee: Well. I took this approach for the reason I described in https://github.com/logos-co/nomos/pull/1902#discussion_r2511160379. The stream-based approach or the `rotate_session` approach looked good to me at the beginning, but now I feel having each component very simple and dealing with transitions in the upper layer look intuitive. But, let's discuss further and make a conclusion after merging this fix.

PR #1916, opened 2025-11-10
feat(zksign): Replace dummy zk signatures with real zk signature

davidrusu:

This PR drops DummyZkSignature, replacing it with the real `zksign::Signature` everywhere in the codebase

The change is fairly mechanical, but it touches a lot of code.

@davidrusu 

NA

NA

## Checklist

> [!WARNING]  

* [x] 2. Description added.


nomos-services/key-management-system/src/keys/zk.rs
```diff
@@ -21,16 +21,13 @@ impl SecuredKey for ZkKey {
         keys: &[&Self],
         payload: &Self::Payload,
     ) -> Result<Self::Signature, Self::Error> {
-        if keys.len() != 32 {
+        if keys.len() > 32 {
```
danielSanchezQ: Actually it should be 32 or it would fail. Either we fill with ZERO ourselves here or we expect 32 keys should come.

nomos-services/wallet/src/lib.rs
```diff
@@ -416,10 +416,8 @@ where
...
                         ))?;
 
-                    let zk_sig = Self::sign_zksig(
-                        tx_hash,
-                        [locked_note.pk.into(), declare_op.zk_id.into_inner()],
-                    );
+                    let zk_sig =
+                        Self::sign_zksig(tx_hash, [locked_note.pk, declare_op.zk_id], kms).await?;
                     let ed25519_sig =
                         Self::sign_ed25519(tx_hash, declare_op.provider_id.0, kms).await?;
```
danielSanchezQ, line 422: nitpick: This can probably be enqueued together (?)

nomos-services/wallet/src/lib.rs
```diff
@@ -504,11 +500,31 @@ where
...
+        kms: &KmsServiceApi<Kms, RuntimeServiceId>,
+    ) -> Result<zksign::Signature, WalletServiceError> {
+        // Use hex-encoded public key as key_id for now
+        let key_ids: Vec<_> = pks
+            .into_iter()
+            .map(|pk| hex::encode(fr_to_bytes(&pk.into_inner())))
+            .collect();
+
+        let payload =
+            key_management_system::keys::PayloadEncoding::Ed25519(tx_hash.as_signing_bytes());
```
danielSanchezQ: I think we can bring `PayloadEncoding` in scope.

nomos-services/wallet/src/lib.rs
```diff
@@ -504,11 +500,31 @@ where
...
+        let signature = kms
+            .sign_multiple(key_ids, payload)
+            .await
+            .map_err(WalletServiceError::KmsApi)?;
+
+        let key_management_system::keys::SignatureEncoding::Zk(zk_sig) = signature else {
```
danielSanchezQ: Same for `SignatureEncoding` (?)

zk/proofs/zksign/src/private.rs
```diff
@@ -11,6 +13,18 @@ pub struct ZkSignPrivateKeysInputs(pub(crate) [Groth16Input; 32]);
...
+    pub fn new(sks: impl IntoIterator<Item = SecretKey>) -> Self {
+        let mut secret_keys = [Fr::ZERO; 32];
+        for (i, sk) in sks.into_iter().enumerate() {
+            assert!(i < 32, "ZkSign supports signing with at most 32 keys");
+            secret_keys[i] = sk.into_inner();
+        }
+
+        Self(secret_keys)
+    }
+}
```
danielSanchezQ: Ok, so you abstracted it here to fill with Zeros. 
nitpick: It would have been better to implement `TryFromIterator` so we could use `collect`?

Also Im not so fond of this API, I would prefer something like:
```rust
pub fn try_from_slice(sks: &[SecretKey]) -> Result<Self, Error> {
        if sks.len() > 32 {
            return Err();
        }
        let mut secret_keys = [Fr::ZERO; 32];
        for (i, sk) in sks.into_iter().enumerate() {
            secret_keys[i] = sk.into_inner();
        }

        Self(secret_keys)
    }
```

zk/proofs/zksign/src/public.rs
```diff
@@ -2,12 +2,34 @@ use circuits_utils::dev_mode::DevModeProof;
...
 
+#[derive(Debug)]
 pub struct ZkSignVerifierInputs {
     pub public_keys: [Groth16Input; 32],
     pub msg: Groth16Input,
 }
 
 impl ZkSignVerifierInputs {
+    #[must_use]
+    pub fn new_variable_pks(
```
danielSanchezQ: Same issue with this method 🤔

nomos-core/chain-defs/src/proofs/zksig.rs
danielSanchezQ, line 1: I'm not so sure we do not need this. Maybe we should have just renamed the things here 🤔 . Or do we actually have the types exported from mantle keys? Dont recall atm.

nomos-services/key-management-system/src/keys/zk.rs
davidrusu: This is a detail of the spec that had not been implemented in the zksign crate yet.

I first made the mistake of thinking that when verifying you would also pad the public key list with ZERO's.

This was wrong, you need to pad with the _public key of the zero secret key_! So I think we really need to encapsulate this detail about how padding is done inside the zksign library.

nomos-services/wallet/src/lib.rs
davidrusu: future improvement?

zk/proofs/zksign/src/private.rs
davidrusu: Yes, good point, it should return an error rather than panic, this could be a DOS attack vector where someone can construct a transaction with more then 32 keys, and a panic here would let the attacker crash the nodes

nomos-services/key-management-system/src/keys/zk.rs
danielSanchezQ: > This is a detail of the spec that had not been implemented in the zksign crate yet.
> 
> I first made the mistake of thinking that when verifying you would also pad the public key list with ZERO's.
> 
> This was wrong, you need to pad with the _public key of the zero secret key_! So I think we really need to encapsulate this detail about how padding is done inside the zksign library.

I got lost on the padding explanation :sweat_smile: , but yes, I saw it was abstracted later in the lib which is better. Thanks.

zk/proofs/zksign/src/private.rs
davidrusu: done

nomos-core/chain-defs/src/proofs/zksig.rs
davidrusu: I don't think it would have helped much, still would need to update every call site to deal with errors

zk/proofs/zksign/src/private.rs
davidrusu: `TryFromIterator` is not a trait, but I used a `TryFrom<&[SecretKey]>` impl

nomos-core/chain-defs/src/proofs/zksig.rs
danielSanchezQ: Yes. What I wanted to express is that all this zksignatures should be re-exported in the nomos-core crate. As we move everything to the zksign crate.
TLDR: Core components are agreed to be available in nomos-core

davidrusu: discussed offline, `zksign` mirrors our usage of `ed25519_dalek`, we don't currently re-export `ed25519_dalek` types from `nomos-core` but when we do the cleanup work in the future, we can centralize these core components in nomos-core so that other crates just depend on nomos-core for the common packages.

danielSanchezQ (approved):

PR #1918, opened 2025-11-10
chore(kms): improve API surface

ntn-x2:

* Change the `Backend` trait definition to take references to key IDs instead of taking ownership of the key, since in many cases a reference is enough, and ownership can be taken by cloning the input key if needed
* The KMS itself should not panic if a key with the same ID exists or if the public key for a key with a given key does not exist, or also if a signature operation fails for some reason. That should be at the discretion of the calling context whether to panic, change ID, or do something else. Hence, I also changed those to be `Result`s which are sent back to the caller via the response channel.

Tried to touch as little as possible, but this should be already better for more ergonomics at the calling site.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-services/key-management-system/src/lib.rs
+++ b/nomos-services/key-management-system/src/lib.rs
@@ -180,33 +180,18 @@ where
                     }
                     return;
                 }
-                let key_public_key = match backend.public_key(&key_id) {
-                    Err(e) => {
-                        if reply_channel.send(Err(e)).is_err() {
-                            error!("Could not send backend public key retrieval error to caller.");
-                        }
-                        return;
-                    }
-                    Ok(key_public_key) => key_public_key,
-                };
-                if reply_channel.send(Ok((key_id, key_public_key))).is_err() {
-                    error!("Could not reply key_id for register request");
+
+                let pk_bytes_result = backend.public_key(&key_id).map(|pk| (key_id.clone(), pk));
+                if reply_channel.send(pk_bytes_result).is_err() {
+                    error!("Could not reply to the public key request channel");
                 }
             }
             KMSMessage::PublicKey {
                 key_id,
                 reply_channel,
             } => {
-                let pk_bytes = match backend.public_key(&key_id) {
-                    Err(e) => {
-                        if reply_channel.send(Err(e)).is_err() {
-                            error!("Could not send backend public key retrieval error to caller.");
-                        }
-                        return;
-                    }
-                    Ok(pk_bytes) => pk_bytes,
-                };
-                if reply_channel.send(Ok(pk_bytes)).is_err() {
+                let pk_bytes_result = backend.public_key(&key_id);
+                if reply_channel.send(pk_bytes_result).is_err() {
                     error!("Could not reply to the public key request channel");
                 }
             }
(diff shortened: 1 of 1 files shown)


ledger/nomos-ledger/src/mantle/sdp/rewards.rs
```diff
@@ -296,11 +296,9 @@ pub enum Error {
 
 #[cfg(test)]
 mod tests {
-    use nomos_core::{
-        mantle::keys::PublicKey,
-        sdp::{DaActivityProof, Declaration, DeclarationId},
-    };
+    use nomos_core::sdp::{DaActivityProof, Declaration, DeclarationId};
     use num_bigint::BigUint;
+    use zksign::PublicKey;
```
ntn-x2, line 301: This is just to make it compile. It will disappear once rebased on `master` after the fix is merged.

nomos-services/key-management-system/src/lib.rs
```diff
@@ -172,24 +174,39 @@ where
...
+                            error!("Could not send backend public key retrieval error to caller.");
+                        }
+                        return;
+                    }
+                    Ok(pk_bytes) => pk_bytes,
                 };
-                if let Err(_pk_bytes) = reply_channel.send(pk_bytes) {
+                if reply_channel.send(Ok(pk_bytes)).is_err() {
                     error!("Could not reply to the public key request channel");
                 }
```
danielSanchezQ:
```suggestion
                let pk = backend.public_key(&key_id);
                if let Err(e) = reply_channel.send(pk) {
                    match e {
                        Ok(_) => error!("Could not reply to the public key request channel"),
                        Err(_) => error!("Could not send backend public key retrieval error to caller."),
                    }
                }
```

nomos-services/key-management-system/src/lib.rs
```diff
@@ -172,24 +174,39 @@ where
...
+                            error!("Could not send backend public key retrieval error to caller.");
+                        }
+                        return;
+                    }
+                    Ok(key_public_key) => key_public_key,
                 };
-                if let Err(_key_descriptor) = reply_channel.send((key_id, key_public_key)) {
+                if reply_channel.send(Ok((key_id, key_public_key))).is_err() {
                     error!("Could not reply key_id for register request");
                 }
```
danielSanchezQ:
```suggestion
                let key_public_key = backend.public_key(&key_id);
                if let Err(e) = reply_channel.send(key_public_key.map(|key| (key_id, key))){
                    match e {
                        Ok(_) => error!("Could not reply to the public key request channel"),
                        Err(e) => error!("Could not send backend public key retrieval error to caller."),
                    }
}
```

nomos-services/key-management-system/src/lib.rs
```diff
@@ -198,21 +215,30 @@ where
...
+                        }
+                        return;
+                    }
+                    Ok(signature) => signature,
+                };
 
-                if let Err(_signature) = reply_channel.send(signature) {
+                if let Err(_signature) = reply_channel.send(Ok(signature)) {
                     error!("Could not reply to the public key request channel");
                 }
```
danielSanchezQ: Same patterns here.

nomos-services/key-management-system/src/lib.rs
ntn-x2: I am not sure I understand this suggestion? You are suggesting to factor out the result of `backend.public_key(&key_id)` into its own variable?

danielSanchezQ: > I am not sure I understand this suggestion? You are suggesting to factor out the result of `backend.public_key(&key_id)` into its own variable?

Is basically making a single reply send instead of two. Swapping the conditions as we send either if its ok or if its err.

ntn-x2: Oh you mean we can send the result directly, I see. Yeah that looks good. I'll fix it.

ntn-x2: Not sure what was I thinking :D This looks much cleaner, thanks for the tip: https://github.com/logos-co/nomos/pull/1918/commits/ea84ffbcf190ebf97240f47349720cf2fd3593ef.

danielSanchezQ (approved):

PR #1919, opened 2025-11-11
test(blend): wait for inbound upgrades as well

youngjoon-lee:

This PR fixes a test that runs forever in the PR https://github.com/logos-co/nomos/pull/1911. 
```
test core::with_core::behaviour::tests::session::forward_message has been running for over 60 seconds
```

Previously, the test was waiting for the outbound connection to be negotiated successfully after dialing to a peer. But, we also have to wait until the inbound connection (on the peer side) is upgraded successfully. If we start other jobs without waiting for it, we may observe unexpected behaviors.
In the test above, we send a new session info to all the swarms, right after the outbound connection was upgraded. But, at that time, the listening peer didn't get/handle the event of the inbound connection upgrade. When a new session starts, each swarm clears all the pending connection upgrades (HashMap). So, even if the listening peer receives the event lately, it ignores it because there's no corresponding pending upgrades that it manages in the HashMap.
So, I think it's much safer to wait until both ends are upgraded successfully.

p.s.
It was hard to find this edge case because the failure is not reproduced when we run only the single `forward_message` test. It is reproduced only when we run multiple tests, as below.
```
cargo test -p nomos-blend-network core::with_core::behaviour::tests::session --all-features -- --nocapture`
```
I still don't get why it's failed only in this case. I think the https://github.com/logos-co/nomos/pull/1911 doesn't change anything related to this test. But anyway, it's good to have this fix that I found from debugging.

I believe so. 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/network/src/core/with_core/behaviour/tests/utils.rs
+++ b/nomos-blend/network/src/core/with_core/behaviour/tests/utils.rs
@@ -1,4 +1,4 @@
-use core::{fmt::Debug, num::NonZeroUsize, ops::RangeInclusive, time::Duration};
+use core::{num::NonZeroUsize, ops::RangeInclusive, time::Duration};
 use std::collections::{HashMap, VecDeque};
 
 use async_trait::async_trait;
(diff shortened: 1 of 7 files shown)


nomos-blend/network/src/core/with_core/behaviour/tests/utils.rs
```diff
@@ -164,29 +167,46 @@ pub fn default_poq_verification_inputs_for_session(
...
 #[async_trait]
 pub trait SwarmExt: libp2p_swarm_test::SwarmExt {
-    async fn connect_and_wait_for_outbound_upgrade<T>(&mut self, other: &mut Swarm<T>)
-    where
-        T: NetworkBehaviour<ToSwarm: Debug> + Send;
+    async fn connect_and_wait_for_upgrade<ListenerBehaviour>(
+        &mut self,
+        other: &mut Swarm<ListenerBehaviour>,
+    ) where
+        ListenerBehaviour: NetworkBehaviour<ToSwarm = Event> + Send;
```
ntn-x2, line 174: Do we need this bound in the trait definition?

danielSanchezQ (approved):

PR #1921, opened 2025-11-11
fix(blend): wait for core service to be stopped before killing it

youngjoon-lee:

Closes https://github.com/logos-co/nomos/issues/1860

Please see the issue to understand a bug that this PR fixes.

In this PR, instead of killing the core service right after its last STP has passed, the proxy service waits 3s for the core service to be stopped by itself. If it's not stopped in time, we kill it forcefully in the end. 3s is for the core service to perform its tear-down logic (computing/submitting an activity proof). 
In most cases, the tear-down will be done very quickly. Even if it wouldn't, I think spending 3s is fine.

Unlike for the core service, we wait 1s for the edge service because it doesn't have any tear-down yet. 

Initially (as noted in the issue), I remember that we shortly discussed about changing the proxy service to not kill or wait for the underlying services at all, but just focusing on starting new services. But after some experiments, I realized that it's not that safe. 
In summary, I think it's better to make sure that all the unnecessary services are cleanly terminated before moving forward. If we just fire-and-forget about services, some services may persist for a long time. If we try to start the same service after some time, Overwatch will ignore our request and allow us to interact with the existing service (This is what I found by experiments. It might be different depending on Overwatch settings). If that service will stop soon, the entire logic will be broken. 
We could probably avoid those edge cases by strictly limiting the cases of transitions. Nevertheless, I want to explicitly make sure that the retired services doesn't exist anymore.

Yes

@youngjoon-lee @ntn-x2 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/instance.rs
+++ b/nomos-services/blend/src/instance.rs
@@ -178,7 +178,7 @@ where
         match self {
             Self::Core(mode) => Ok(Self::Core(mode)),
             mode => {
-                mode.shutdown().await;
+                mode.wait_until_stopped_or_kill().await;
                 Ok(Self::Core(Self::new_core_mode(overwatch_handle).await?))
             }
         }
@@ -199,11 +199,11 @@ where
             }),
             Self::Edge(mode) => Ok(Self::Edge(mode)),
             Self::EdgeAfterCore { mode, prev } => {
-                prev.shutdown().await;
+                prev.wait_until_stopped_or_kill().await;
                 Ok(Self::Edge(mode))
             }
             mode => {
-                mode.shutdown().await;
+                mode.wait_until_stopped_or_kill().await;
                 Ok(Self::Edge(Self::new_edge_mode(overwatch_handle).await?))
             }
         }
@@ -225,11 +225,11 @@ where
             }),
             Self::Broadcast(mode) => Ok(Self::Broadcast(mode)),
             Self::BroadcastAfterCore { mode, prev } => {
-                prev.shutdown().await;
+                prev.wait_until_stopped_or_kill().await;
                 Ok(Self::Broadcast(mode))
             }
             mode => {
-                mode.shutdown().await;
+                mode.wait_until_stopped_or_kill().await;
                 Ok(Self::Broadcast(
                     Self::new_broadcast_mode(overwatch_handle).await?,
                 ))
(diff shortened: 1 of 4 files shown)


danielSanchezQ (approved):

nomos-services/blend/src/instance.rs
```diff
@@ -199,11 +199,11 @@ where
             }),
             Self::Edge(mode) => Ok(Self::Edge(mode)),
             Self::EdgeAfterCore { mode, prev } => {
-                prev.shutdown().await;
+                prev.wait_until_stopped_or_kill().await;
```
ntn-x2, line 202: Am I wrong, or this case should never happen?

nomos-services/blend/src/instance.rs
```diff
@@ -225,11 +225,11 @@ where
             }),
             Self::Broadcast(mode) => Ok(Self::Broadcast(mode)),
             Self::BroadcastAfterCore { mode, prev } => {
-                prev.shutdown().await;
+                prev.wait_until_stopped_or_kill().await;
```
ntn-x2, line 228: This should also never happen, since the first time we call this function we move to `Broadcast`. At the very least a warning line should be printed, but maybe even a panic?

ntn-x2 (approved):
I think this approach looks reasonable. Thanks.

nomos-services/blend/src/instance.rs
youngjoon-lee: Yesss, you're right. As long as the STP is shorter than the session duration, this shouldn't happen.
When I wrote this code at the very beginning, I wanted for our code to handle that weird case, even if the membership stream yields new memberships very fast unexpectedly. But, after having many discussions with you, now I think it's better and safer to strictly reject these cases.
This is actually what I shortly mentioned in the PR description:
> We could probably avoid those edge cases by strictly limiting the cases of transitions

I'll open another PR soon to address it. We may be able to panic in these cases or consider other ways. I'll check different aspects soon. Thanks.

nomos-services/blend/src/instance.rs
youngjoon-lee: Yeah, exactly. As I noted in https://github.com/logos-co/nomos/pull/1921#discussion_r2516843819, I will address this in a nexst PR.

PR #1923, opened 2025-11-11
chore(blend): remove node_key in favor of cryptographic processor key

ntn-x2:

Tests were passing by chance since we were using the same seed for both keys, but in reality `node_key` is not needed, since using a different value than the cryptographic processor would result in a different peer ID and hence in failed handshakes. I tested that locally, by using a different `node_key` value compared to the crypto processor and that's the result.

Yes.

@ntn-x2 @youngjoon-lee 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nodes/nomos-node/config.yaml
+++ b/nodes/nomos-node/config.yaml
@@ -30,7 +30,6 @@ blend:
   core:
     backend:
       listening_address: /ip4/127.0.0.1/udp/3001/quic-v1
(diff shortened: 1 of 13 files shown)


youngjoon-lee (approved):
Oh, yes, thank you. I also realized this (from your comments and messages) and updated a test util in https://github.com/logos-co/nomos/pull/1909#discussion_r2518074291. Actually, these two keys are leftovers from the very beginning when we were not sure how to authenticate connections. Now we shouldn't have two separate keys. Thanks.

PR #1925, opened 2025-11-11
feat(blend): integrate KMS into PoQ generation flow

ntn-x2:

Integrate KMS into core PoQ generation. We don't integrate it with the other private key used (the non-ephemeral signing key) since it's also used in libp2p and we need a global strategy to do the same in other services as well.

In order to do that, we need to move away from passing the secret key around, and instead I adapted the proof generators to take an inner PoQ generator implementation, which in the binary is implemented by the KMS adapter: each request to generate a PoQ is transformed into an `Execute` message and the result sent back to the caller. I tried to keep the changes to both KMS and Blend as small as possible, and as a result I had to take a few shortcuts along the way.

I introduced a `kms.rs` module inside the Blend core module that is responsible for defining the adapter trait and provide an implementation for the already existing KMS API type.

As for the KMS module, we decided that at the moment we go ahead and just add the functionality we need (in this case PoQ generation). Nevertheless, there's a draft PR that aims to make this general: https://github.com/logos-co/nomos/pull/1928. We will explore its potential after this PR is merged, to provide an existing functionality to migrate to the new implementation.

I think and hope so.

@ntn-x2 @youngjoon-lee 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/core/kms.rs
+++ b/nomos-services/blend/src/core/kms.rs
@@ -1,7 +1,7 @@
 use core::fmt::{Debug, Display};
 
 use async_trait::async_trait;
-use futures::future::ready;
+use futures::{executor::block_on, future::ready};
 use key_management_system::{
     KMSService,
     api::KmsServiceApi,
@@ -13,7 +13,7 @@ use nomos_blend_scheduling::message_blend::CoreProofOfQuotaGenerator;
 use nomos_core::crypto::ZkHash;
 use overwatch::services::AsServiceId;
 use poq::CorePathAndSelectors;
-use tokio::sync::oneshot;
+use tokio::{sync::oneshot, task::spawn_blocking};
 
 const LOG_TARGET: &str = "blend::service::core::kms-poq-generator";
 
@@ -69,7 +69,7 @@ impl<RuntimeServiceId> CoreProofOfQuotaGenerator
     for PreloadKMSBackendCorePoQGenerator<RuntimeServiceId>
 where
     RuntimeServiceId:
-        AsServiceId<PreloadKmsService<RuntimeServiceId>> + Debug + Display + Send + Sync,
+        AsServiceId<PreloadKmsService<RuntimeServiceId>> + Debug + Display + Send + Sync + 'static,
 {
     fn generate_poq(
         &self,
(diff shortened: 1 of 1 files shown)


nomos-blend/scheduling/src/message_blend/provers/core/mod.rs
```diff
@@ -120,28 +126,30 @@ fn create_proof_stream(
...
+                        &PublicInputs {
+                            signing_key: ephemeral_signing_key.public_key(),
+                            core: public_inputs.core,
+                            leader: public_inputs.leader,
+                            session: public_inputs.session,
+                        },
+                        key_index,
+                    )
+                    .await
+                    .ok()?;
```
youngjoon-lee, line 144: So, now we don't spawn a task for generating proofs?

nomos-services/blend/src/core/mod.rs
```diff
@@ -450,14 +490,19 @@ where
     Backend: BlendBackend<NodeId, BlakeRng, ProofsVerifier, RuntimeServiceId> + Sync,
     NetAdapter: NetworkAdapter<RuntimeServiceId, BroadcastSettings: Eq + Hash>,
     ChainService: ChainApi<RuntimeServiceId> + Sync,
-    ProofsGenerator: CoreAndLeaderProofsGenerator,
+    ProofsGenerator: CoreAndLeaderProofsGenerator<KmsAdapter::CorePoQGenerator>,
     ProofsVerifier: ProofsVerifierTrait,
-    RuntimeServiceId: Sync,
+    // To avoid bubbling up generics everywhere in the configs (current Overwatch limitation), we
+    // know the final key ID type is a `String`, so we constraint the trait impl here instead.
```
youngjoon-lee, line 499: 👍

nomos-services/blend/src/core/settings.rs
```diff
@@ -101,13 +101,6 @@ pub struct MessageDelayerSettings {
...
-    pub sk: SecretKey,
-}
-
-impl ZkSettings {
-    #[must_use]
-    pub fn public_key(&self) -> PublicKey {
-        self.sk.to_public_key()
-    }
+    #[serde(rename = "secret_key_kms_id")]
+    pub sk_kms_id: KeyId,
```
youngjoon-lee: nit: now I think we can use the same name `secret_key_kms_id` in the struct, but not important.

nomos-blend/scheduling/src/message_blend/provers/core/mod.rs
youngjoon-lee: Ah, now I understand this after reading the implementation of `PreloadKMSBackendCorePoQGenerator`.

Then, what do you think about spawning a task in the KMS service when it runs an `execute`? https://github.com/logos-co/nomos/blob/d28012ed9a171a245543afcfd900a37580256b54/nomos-services/key-management-system/src/lib.rs#L152-L153
This might not be the scope of this PR. But if an `execute` is heavy, it will block other messages scheduled for the KMS service.
I guess that was why we've had `spawn_blocking` in this `create_proof_stream` function so far.

nomos-services/blend/src/core/kms.rs
```diff
@@ -0,0 +1,121 @@
...
+                .expect("Should not fail to get PoQ generation result from KMS.")
+                .map_err(|e| {
+                    let KeyError::PoQGeneration(poq_err) = e else {
+                        panic!("PoQ generation from KMS should return a PoQ generation error, if any.");
+                    };
+                    tracing::error!(target: LOG_TARGET, "KMS-based PoQ generation failed with error {poq_err:?}.");
+                    poq_err
+                })?;
+            tracing::debug!(target: LOG_TARGET, "KMS-based PoQ generation succeeded.");
+            Ok(poq)
```
danielSanchezQ, line 102: This is not gonna change too much if we change to the new schema. We will encapsulate this computation in a struct which will be send to the execute instead.

nomos-blend/scheduling/src/message_blend/provers/core/mod.rs
ntn-x2: Yeah I thought about it as well. I don't think having the spawn blocking INSIDE the KMS execute is the proper solution, since execute is meant to allow for arbitrary computation. If anything, the blocking task should be spawned by the calling context. In this case we could spawn it in the `generate_poq` impl for `PreloadKMSBackendCorePoQGenerator`. This is my attempt at it: https://github.com/logos-co/nomos/pull/1925/commits/37c84fb741d1c339ec82efff3cb7a06f47a93981.

ntn-x2: It's a bit tricky since `execute` is an `async` function, so we need to generate some `async` closures to glue everything together, but I think this does the job. I'll see if the tests keep passing, then will merge.

nomos-services/blend/src/core/settings.rs
ntn-x2: Yes makes sense, thanks. https://github.com/logos-co/nomos/pull/1925/commits/cea9bb9edbf4a8d7e89522ccc8c7de2ffa4857b1

nomos-blend/scheduling/src/message_blend/provers/core/mod.rs
ntn-x2: Macos tests failed as usual, but Linux ones passed, so I'll go ahead and merge this.

PR #1928, opened 2025-11-13
feat(kms): Excute bind to SecuredKey

danielSanchezQ:

This PR introduce a different abstraction for operating with keys in the KMS. Not it uses a double dispatch approach and its included in the `SecuredKey` trait. 
It also adds the extensible functionality for the macros. So the proper types are generated and implemented for an enumeration used keys.
It also adapts the `PoQ` so it is computed in the KMS instead.

Yes, it's much simpler than previous version actually.

@danielSanchezQ @strinnityk @ntn-x2 

Not related

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.


nomos-services/key-management-system-macros/src/lib.rs
```diff
@@ -52,7 +53,7 @@ fn validate_variants(key_enum_variants: &Punctuated<Variant, Comma>) -> Option<E
     None
 }
 
-fn build_encoding_enums(key_enum_variants: &Punctuated<Variant, Comma>) -> TokenStream {
+fn build_associated_types(key_enum_variants: &Punctuated<Variant, Comma>) -> TokenStream {
```
strinnityk: This one shouldn't be renamed.

nomos-services/key-management-system/src/backend/mod.rs
```diff
@@ -33,6 +33,6 @@ pub trait KMSBackend {
     async fn execute(
         &mut self,
         key_id: &Self::KeyId,
-        operator: KMSOperatorBackend<Self>,
+        operator: <Self::Key as SecuredKey>::Operations,
```
strinnityk: Nitpick: There's a mix between `operation` and `operator` throughout this PR, I'd stick to one.

nomos-services/key-management-system/src/keys/errors.rs
```diff
@@ -14,6 +14,8 @@ pub enum KeyError {
     UnsupportedMultisignatureSize(usize, usize),
     #[error(transparent)]
     ZkSignError(#[from] zksign::ZkSignError),
+    #[error("Unsupported operation for key type `{0}`")]
```
strinnityk: Could we go for something like: ```#[error("Unsupported operation `{0}` for key type `{1}`")]```

nomos-services/key-management-system/src/keys/secured_key.rs
```diff
@@ -1,11 +1,42 @@
+use std::marker::PhantomData;
+
 use zeroize::ZeroizeOnDrop;
 
+#[async_trait::async_trait]
+pub trait SecureKeyOperations {
```
strinnityk: Name should be singular?

strinnityk (approved):
Some minor suggestions.

nomos-services/key-management-system/src/keys/secured_key.rs
```diff
@@ -1,11 +1,42 @@
...
+}
+
 /// A key that can be used within the Key Management Service.
+#[async_trait::async_trait]
 pub trait SecuredKey: ZeroizeOnDrop {
     type Payload;
     type Signature;
     type PublicKey;
     type Error;
+    type Operations: SecureKeyOperations<Key = Self, Error = Self::Error> + Send + Sync + 'static;
```
ntn-x2: What's the need for the `Key` associated type? We assume the trait is implemented for a key already, so we can probably just remove `Key` and use `execute(&self)` only? Also, if we were to provide e.g., two different capabilities for a key, such as PoQ and PoL generation for a ZK key, how would that work? This can only be implemented once per key, unless we define a collection type that itself implements `SecureKeyOperations`? We would then need to provide an operation ID of some sort?

ntn-x2: I think what we could do is to change this to be a generic, then in the place where we use the key, we constrain it to be of a specific type. If we need two operations at the same time for the same key, we would require the key to implement `SecuredKey<Op1>` and `SecuredKey<Op2>`.

nomos-services/key-management-system/src/keys/secured_key.rs
```diff
@@ -1,11 +1,42 @@
+use std::marker::PhantomData;
+
 use zeroize::ZeroizeOnDrop;
 
+#[async_trait::async_trait]
+pub trait SecureKeyOperations {
+    type Key;
+    type Error;
```
ntn-x2, line 8: What if we also define a `Output` type on this trait?

nomos-services/key-management-system/src/keys/secured_key.rs
danielSanchezQ: > What's the need for the `Key` associated type? We assume the trait is implemented for a key already, so we can probably just remove `Key` and use `execute(&self)` only? Also, if we were to provide e.g., two different capabilities for a key, such as PoQ and PoL generation for a ZK key, how would that work? This can only be implemented once per key, unless we define a collection type that itself implements `SecureKeyOperations`? We would then need to provide an operation ID of some sort?

Uhm, I'm not sure I follow. The `SecureKeyOperations` is not implemented by the key buth rather use a double dispatch system. I think best way is to show to you how it was intended por PoQ for example?

When implementing for PoQ and Pol for example we would have to have a wrapper enum with each operations that dispatch it. Sadly I could't make it work with dinamic dispatch.

nomos-services/key-management-system/src/keys/secured_key.rs
danielSanchezQ: > What if we also define a `Output` type on this trait?

I tried, but the problem is that then we need to make the `execute` method from the `SecuredKey` trait to be generic which breaks some v-table rules. This is why I had to do it through the static enum dispatch.
Also It is not really necessary, as we can treat this as an async computation, the result extraction is responsability of the operator itself.

nomos-services/key-management-system/src/backend/mod.rs
danielSanchezQ: Yeah, its gone now anyway.

nomos-services/key-management-system/src/keys/errors.rs
danielSanchezQ: I would have to make Operator to impl Debug or Display :thinking:

nomos-services/key-management-system/src/keys/ed25519.rs
```diff
@@ -38,11 +35,13 @@ impl<'de> Deserialize<'de> for Ed25519Key {
     }
 }
 
+#[async_trait::async_trait]
 impl SecuredKey for Ed25519Key {
     type Payload = Bytes;
     type Signature = Signature;
     type PublicKey = VerifyingKey;
     type Error = KeyError;
+    // type Operations = NoKeyOperator<Self, Self::Error>;
```
ntn-x2: I guess we want to uncomment this? And what if we make this `Option` instead? Would that be ugly?

danielSanchezQ: No need, we do not define it like this anymore. Its a leftover. I already removed. Thanks!

PR #1933, opened 2025-11-17
fix(blend): update calculating the bit length of expected token count for activity proof

youngjoon-lee:

Recently, the spec has been updated (simplified).
Now we don't need `message_frequency_per_round` for calculating `token_count_bit_len` (which is necessary for activity proof).
I agree that this spec update makes more sense (it's actually correct now).

https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df815492efec16a0e9021b

image

Yes

@youngjoon-lee @ntn-x2 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/message/src/reward/session.rs
+++ b/nomos-blend/message/src/reward/session.rs
@@ -140,9 +140,5 @@ mod tests {
         let total_core_quota = 0;
         // ceil(log2(0 + 1))
         assert_eq!(token_count_bit_len(total_core_quota).unwrap(), 0);
-
-        let total_core_quota = 10;
-        // ceil(log2(10 + 1))
-        assert_eq!(token_count_bit_len(total_core_quota).unwrap(), 0);
     }
 }


nomos-blend/message/src/reward/session.rs
```diff
@@ -99,19 +96,14 @@ pub enum Error {
...
-    Ok(
-        F64Ge1::try_from(total_core_quota.mul_add(message_frequency_per_round.get(), 1.0))
-            .expect("must be >= 1.0")
-            .log2()
-            .ceil() as u64,
-    )
+    Ok(F64Ge1::try_from(total_core_quota.add(1.0))
+        .expect("must be >= 1.0")
+        .log2()
+        .ceil() as u64)
```
danielSanchezQ, line 106: Why do we transform into u64 to make it F64 later on again?

youngjoon-lee: It’s because we want to do `log2` + `ceil`, as per spec. 

`u64` has [`ilog2`](https://doc.rust-lang.org/std/primitive.u64.html#method.ilog2) but it rounds the output down. That’s not what we want. 

So I had to convert the value to f64 first, do log2 + ceil, and convert it back to u64

PR #1934, opened 2025-11-17
feat(da-sampling): fallback to historic commitments request on session mismatch

pradovic:

This PR solves the edge case when get commitments is request from behavior but session changes before the fetching starts. The behavior returns session mismatch in that case and we fallback to historic behavior to fetch the commitments.
The implementation follows the existing pattern and avoids major refactoring. We will probably need to refactor this code in the future to make it more elegant, now that we know more edge cases.

Yes

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


nomos-da/network/core/src/protocols/sampling/mod.rs
```diff
@@ -143,6 +144,11 @@ pub enum SamplingEvent {
         shares: HashMap<BlobId, Vec<DaLightShare>>,
         commitments: HashMap<BlobId, DaSharesCommitments>,
     },
+    HistoricCommitmentsSuccess {
+        block_id: HeaderId,
```
pradovic, line 148: I think we should rename block_id to something like request_id, in both historic sampling and historic commitments requests, as this is really used only to group result events. It does not have to be block_id.

bacv: If I remember correctly we used block_id only for error messages. We could leave it as is, just add a comment that it acts as request id, but also allows to have more descriptive errors.

PR #1935, opened 2025-11-17
feat(chain-network): fix chain-service cycles once and for all!

davidrusu:

I was attempting to integrate the wallet with SDP and DA, but ran into a cyclic dependency that I could not break using the `chain/broadcast-service` since the Wallet service is not just listening for new blocks, it needs to query the chain state on demand.

To resolve the cycle, this PR splits the chain-service into two service:

1. `chain-service` which is a "simple" wrapper of cryptarchia_engine
2. `chain-network` which receives blocks from the network, validates DA sampling and then forwards them to `chain-service`.

That this means is that `chain-service` has a very light dependency tree now. It does not depend on any services that rely on chain state. This means that if you need to read the chain state, you can take a direct dependency on `chain-service` and not worry about cycles. (e.g. the `chain/broadcast-service` can now be folded back into `chain-service`). All heavy dependencies have been shifted to chain-network.

There were a few decisions made here:
1. we now always do DA validation before appying the block. Previously we would first attempt to apply the block, and the if that succeeds, run DA validation.
2. chain-sync events are handled by chain-service, so we had to introduce some indirection here because we don't want chain-service to take a dependency on the network. `chain-network` now listens for `chain-sync` events and forwards them to `chain-service`.

hopefully the explanation above helps.

NA

NA

NA

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-services/chain/chain-service/src/lib.rs
+++ b/nomos-services/chain/chain-service/src/lib.rs
@@ -237,14 +237,23 @@ impl Cryptarchia {
         let parent = header.parent();
         let slot = header.slot();
         // A block number of this block if it's applied to the chain.
-        let ledger = self.ledger.try_update::<_, MainnetGasConstants>(
-            id,
-            parent,
-            slot,
-            header.leader_proof(),
-            VoucherCm::default(), // TODO: add the new voucher commitment here
-            block.transactions(),
-        )?;
+        let ledger = self
+            .ledger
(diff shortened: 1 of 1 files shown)


danielSanchezQ (approved):
Just what we need. Thanks for the extra effort @davidrusu !

youngjoon-lee (approved):
Great! Nice structure.

PR #1936, opened 2025-11-17
perf(merkle_proof): improve zk-proving complexity

thomaslavaur:

It change the Poseidon2 hash function to use the compression mode in Merkle proofs, decreasing the generation time of poq, pol and poc.

Yes

@thomaslavaur 

No but I added the details in the Cryptographic Component about the SAFE padding usage.

Yes: https://www.notion.so/nomos-tech/Common-Cryptographic-Components-1fd261aa09df81ac8ebbe0111e2c2d84?source=copy_link#2ae261aa09df8071ab60f1acf59f821f

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/nomos-blend/message/src/crypto/proofs/mod.rs
+++ b/nomos-blend/message/src/crypto/proofs/mod.rs
@@ -135,13 +135,18 @@ where
     }
 }
 
-impl<T> ZkCompressExt for T
-where
-    T: AsRef<[ZkHash; 2]>,
-{
+impl ZkCompressExt for [ZkHash; 2] {
     fn compress(&self) -> ZkHash {
         let mut hasher = ZkHasher::new();
-        hasher.compress(self.as_ref());
+        hasher.compress(self);
+        hasher.finalize()
+    }
+}
+
+impl ZkCompressExt for &[ZkHash; 2] {
+    fn compress(&self) -> ZkHash {
+        let mut hasher = ZkHasher::new();
+        hasher.compress(self);
         hasher.finalize()
     }
 }


zk/poseidon2/src/lib.rs
```diff
@@ -9,6 +9,8 @@ pub type ZkHash = Fr;
 pub trait Digest {
     fn digest(inputs: &[Fr]) -> ZkHash;
 
+    fn compress(inputs: &[Fr; 2]) -> ZkHash;
```
danielSanchezQ, line 22: It would be nice to add some documentation on the difference between this two calls

danielSanchezQ (approved):
Lets add some docs. Otherwise really nice catch. Also it would be nice to add the explanation on the compressed form and improvements to the PR, for posterity.

zk/poseidon2/src/lib.rs
thomaslavaur: added

scripts/setup-nomos-circuits.sh
```diff
@@ -5,18 +5,18 @@
...
 #
 # Arguments:
-#   VERSION      - Optional. Version to install (default: v0.2.0)
+#   VERSION      - Optional. Version to install (default: v0.3.1)
 #   INSTALL_DIR  - Optional. Installation directory (default: $HOME/.nomos-circuits)
 #
 # Examples:
 #   ./setup-nomos-circuits.sh                    # Install default version to default location
-#   ./setup-nomos-circuits.sh v0.3.0             # Install specific version to default location
+#   ./setup-nomos-circuits.sh v0.2.0             # Install specific version to default location
```
ntn-x2, line 13: I guess we want to use `0.3.1` here as well?

nomos-blend/message/src/crypto/proofs/mod.rs
```diff
@@ -129,4 +130,11 @@ where
         hasher.update(self.as_ref());
         hasher.finalize()
     }
+
+    fn compress(&self) -> ZkHash {
```
ntn-x2, line 147: I think this does not belong here. We should have a different trait that expose the `compress` function and only implement it for `T: AsRef<[Fr; 2]>`. We can call the new trait `ZkCompressExt`.

ntn-x2: I can help you with that if you need.

scripts/setup-nomos-circuits.sh
thomaslavaur: Here it’s to download a specific version I was supposing it makes more sense to not use the default one in this case

nomos-blend/message/src/crypto/proofs/mod.rs
thomaslavaur: But we also have hash defined here. In this case isn’t both not supposed ro be here?

thomaslavaur: In zk we also have Poseidon2Hasher defined already

ntn-x2: The problem is that we can only call `compress` on slices of two elements, while we can call `hash` on slices of any number of elements. So we can implement this trait for slices of two elements, but then we lose `hash` impls for slices of any elements. Splitting the trait up allows for different slices to implement the trait independently, so we would implement the compress trait only for slices of two elements.

PR #1937, opened 2025-11-18
fix(blend): avoid dividing core_quota by num_blend_layers when unnecessary

youngjoon-lee:

In the `session_quota` function (which calculate the core quota `Q_c`), we were dividing the calculated core quota by `num_blend_layers`. I guess it was because we need to know the allowed number of "messages" (not "all PoQs") for the message scheduler.
But, we were also using this function for the proof generator and the token collector, which require the `Q_c` (the allowed number of PoQs). 

So, this PR changes the `session_quota` function to return the core quota `Q_c` as it is. And, now we divide `Q_c` by `num_blend_layers` only when we pass it to the message scheduler.

Plus, this PR changes the type of `num_blend_layers` from `u64` to `NonZeroU64`, because it is used for the div operation. I think this change is fine, because it's meaningless to use the blend service if we want to set 0 to `num_blend_layers`.

I believe so.

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nodes/nomos-node/node/src/config/blend.rs
+++ b/nodes/nomos-node/node/src/config/blend.rs
@@ -1,3 +1,5 @@
+use std::num::NonZeroU64;
+
 use nomos_libp2p::Multiaddr;
 use overwatch::services::ServiceData;
 use serde::{Deserialize, Serialize};
@@ -25,7 +27,7 @@ impl BlendConfig {
         self.0.core.backend.listening_address = listening_address;
     }
 
-    pub const fn set_blend_layers(&mut self, layers: u64) {
+    pub const fn set_blend_layers(&mut self, layers: NonZeroU64) {
         self.0.common.crypto.num_blend_layers = layers;
     }
 }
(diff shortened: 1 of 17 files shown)


nomos-blend/scheduling/src/cover_traffic.rs
```diff
@@ -177,20 +177,20 @@ where
 /// Pre-compute the rounds that will contain a cover message, given the amount
 /// of messages to generate and the total number of rounds.
 fn schedule_message_rounds<Rng>(
-    mut total_message_count: usize,
+    mut total_message_count: u64,
```
ntn-x2, line 180: We should probably leave this to `usize` since we use it where a `usize` is expected.

youngjoon-lee: Currently, there's no place where a `usize` is passed to this function. I think we can safely define this as `u64`.

ntn-x2: What do you mean? We have to cast to define the capacity of the set of slots. The idea was that anything that is serialized uses `u64` since `usize` is platform dependent. Anything else uses `usize`, especially if it's used in place where a `usize` is expected, i.e., the size of the set of slots. But sure, it was a nitpick anyway.

PR #1938, opened 2025-11-18
chore: support deployment-bound config and integrate it for Blend

ntn-x2:

This PR introduces the concept of deployments. A _deployment_ includes a set of pre-defined parameters that the user does not need to be aware of nor should manually modify and hence should not be present in the node's config file, unless they need to tweak some values, for some reasons, or need to deploy a Nomos instance that is not one of the supported ones. 

As a test case, I specified a `mainnet` variant for now, which includes the set of Blend parameters we will use in our mainnet. These include:
* Expected maximum number of encapsulations per message. This is a network-wide parameter, as specifying a higher value will introduce a risk for a node to be marked as spammy by its peers.
* Time configuration, including session durations, session and epoch transition periods, etc.
* Minimum Blend network size. A user tweaking this value simply won't have any concrete impact since other nodes will ignore Blend messages, hence we don't see any reason for a user to change this from the default value. This is the closest to a genesis-level value, but since we don't have any way to include it in genesis we need to rely on a deployment-bound configuration
* Cover traffic and message delaying settings (only for core nodes): this are all protocol-defined values, and changing any of them can result in the node being excluded from the p2p network, so all nodes should use the same set of values
* Blend core network libp2p protocol name: this is also spec-defined so the user should not be required to change this unless they are purposefully spinning up a new network. This applies to both core and edge nodes since they are logically part of the same libp2p network.

I also fixed the e2e config generation from Blend by removing a seemingly useless `signer` and return a deployment-agnostic config, which is then "augmented" with deployment-specific information before each binary is started.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nodes/nomos-node/node/src/config/blend/deployment.rs
+++ b/nodes/nomos-node/node/src/config/blend/deployment.rs
@@ -16,7 +16,6 @@ pub struct Settings {
     #[serde(flatten)]
     pub common: CommonSettings,
     pub core: CoreSettings,
-    pub edge: EdgeSettings,
 }
 
 impl From<DeploymentSettings> for Settings {
@@ -37,6 +36,7 @@ pub struct CommonSettings {
     pub num_blend_layers: NonZeroU64,
     pub timing: TimingSettings,
     pub minimum_network_size: NonZeroU64,
+    pub protocol_name: StreamProtocol,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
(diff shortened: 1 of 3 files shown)


bacv:
> Did I miss anything?

No, I see that the previous method of overwriting from cli (and env) as you said is still applied here, so from my side it looks great and could also be applied to DA to simplify.

youngjoon-lee (approved):
I like this approach.

I think this is also useful for what I'm working on. To make `nomos-ledger` calculate blend rewards, I have to pass some blend parameters to the `nomos-ledger` through the `chain-*` services. Instead of specifying duplicate values in TOML, we could load values from `DeploymentSettings`.

ntn-x2:
@youngjoon-lee the idea was to separate service-specific settings between user and deployment config. But I guess if a service needs a deployment-bound value that strictly belongs to a different service, we can still do that in the binary, avoid duplication, as you mention. Glad to hear that is a second use case we can fulfil with this new approach.

PR #1939, opened 2025-11-18
feat(wallet): HTTP API

strinnityk:
HTTP API and Client for wallet.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/nodes/nomos-node/node/src/api/handlers.rs
+++ b/nodes/nomos-node/node/src/api/handlers.rs
@@ -67,9 +67,7 @@ use {
     tokio_stream::StreamExt as _,
 };
 
-use crate::api::{
-    backend::DaStorageBackend, responses, responses::overwatch::get_relay_or_response,
-};
+use crate::api::{backend::DaStorageBackend, responses, responses::overwatch::get_relay_or_500};
 
 #[macro_export]
 macro_rules! make_request_and_return_response {
@@ -521,7 +519,7 @@ where
     RuntimeServiceId:
         AsServiceId<StorageService<RocksBackend, RuntimeServiceId>> + Debug + Sync + Display,
 {
-    let relay = match get_relay_or_response(&handle).await {
+    let relay = match get_relay_or_500(&handle).await {
         Ok(relay) => relay,
         Err(error_response) => return error_response,
     };
@@ -610,7 +608,7 @@ where
         + Display
         + 'static,
 {
-    let relay = match get_relay_or_response(&handle).await {
+    let relay = match get_relay_or_500(&handle).await {
         Ok(relay) => relay,
         Err(error_response) => return error_response,
     };
@@ -645,7 +643,7 @@ where
         + Display
         + 'static,
 {
-    let relay = match get_relay_or_response(&handle).await {
+    let relay = match get_relay_or_500(&handle).await {
         Ok(relay) => relay,
         Err(error_response) => return error_response,
     };
@@ -680,7 +678,7 @@ where
         + 'static
         + AsServiceId<StorageService<RocksBackend, RuntimeServiceId>>,
 {
-    let relay = match get_relay_or_response(&handle).await {
+    let relay = match get_relay_or_500(&handle).await {
         Ok(relay) => relay,
         Err(error_response) => return error_response,
     };
@@ -1051,7 +1049,7 @@ pub mod wallet {
         RuntimeServiceId,
     >(
         State(handle): State<OverwatchHandle<RuntimeServiceId>>,
-        Path(public_key): Path<PublicKey>,
+        Path(address): Path<PublicKey>,
         Query(query): Query<TipQuery>,
     ) -> Response
     where
@@ -1080,7 +1078,7 @@ pub mod wallet {
             + AsServiceId<Cryptarchia<RuntimeServiceId>>,
     {
         let wallet_api = {
-            let wallet_relay = match get_relay_or_response::<WalletService, _>(&handle).await {
+            let wallet_relay = match get_relay_or_500::<WalletService, _>(&handle).await {
                 Ok(relay) => relay,
                 Err(error_response) => return error_response,
             };
@@ -1103,10 +1101,19 @@ pub mod wallet {
             }
         };
 
-        let balance = wallet_api.get_balance(tip, public_key).await;
+        let balance = wallet_api.get_balance(tip, address).await;
         match balance {
-            Ok(Some(balance)) => WalletBalanceResponseBody { balance }.into_response(),
-            Ok(None) => (StatusCode::NOT_FOUND, "Wallet not found").into_response(),
+            Ok(Some(balance)) => WalletBalanceResponseBody {
+                tip,
+                balance,
+                address,
+            }
+            .into_response(),
+            Ok(None) => (
+                StatusCode::NOT_FOUND,
+                "The requested address could not be found in the wallet.",
+            )
+                .into_response(),
             Err(error) => (StatusCode::INTERNAL_SERVER_ERROR, error.to_string()).into_response(),
         }
     }
@@ -1160,7 +1167,7 @@ pub mod wallet {
             + AsServiceId<Cryptarchia<RuntimeServiceId>>,
     {
         let wallet_api = {
-            let wallet_relay = match get_relay_or_response::<WalletService, _>(&handle).await {
+            let wallet_relay = match get_relay_or_500::<WalletService, _>(&handle).await {
                 Ok(relay) => relay,
                 Err(error_response) => return error_response,
             };
--- a/nodes/nomos-node/node/src/api/responses/overwatch.rs
+++ b/nodes/nomos-node/node/src/api/responses/overwatch.rs
@@ -7,7 +7,7 @@ use overwatch::{
     services::{AsServiceId, ServiceData, relay::OutboundRelay},
 };
 
-pub async fn get_relay_or_response<Service, RuntimeServiceId>(
+pub async fn get_relay_or_500<Service, RuntimeServiceId>(
     handle: &OverwatchHandle<RuntimeServiceId>,
 ) -> Result<OutboundRelay<<Service as ServiceData>::Message>, Response>
 where
--- a/nodes/api-common/src/bodies/wallet.rs
+++ b/nodes/api-common/src/bodies/wallet.rs
@@ -3,12 +3,15 @@ pub mod balance {
         http::StatusCode,
         response::{IntoResponse, Response},
     };
-    use nomos_core::mantle::Value;
+    use nomos_core::{header::HeaderId, mantle::Value};
     use serde::{Deserialize, Serialize};
+    use zksign::PublicKey;
 
     #[derive(Serialize, Deserialize)]
     pub struct WalletBalanceResponseBody {
+        pub tip: HeaderId,
         pub balance: Value,
+        pub address: PublicKey,
     }
 
     impl IntoResponse for WalletBalanceResponseBody {


nodes/api-common/Cargo.toml
```diff
@@ -22,10 +22,13 @@ pprof = { default-features = false, features = [
...
+serde_json = { workspace = true, optional = true }
 serde_with = { default-features = false, features = ["macros"], workspace = true }
 tokio = { default-features = false, features = ["time"], optional = true, version = "1" }
 tower_governor = { default-features = false, features = ["axum"], version = "0.3" }
 tracing = { default-features = false, optional = true, workspace = true }
+zksign = { workspace = true, optional = true }
 
 [features]
 profiling = ["pprof", "tokio", "tracing"]
+wallet    = ["serde_json", "zksign"]
```
davidrusu: can we drop the feature gate?

nodes/api-common/src/paths.rs
```diff
@@ -24,7 +25,10 @@ pub const SDP_POST_WITHDRAWAL: &str = "/sdp/withdrawal";
 pub const BLOCKS: &str = "/cryptarchia/blocks";
 pub const BLOCKS_STREAM: &str = "/cryptarchia/blocks/stream";
 
-pub const MANTLE_SDP_DECLARATIONS: &str = "/mantle/sdp/declarations";
+pub mod wallet {
+    pub const BALANCE: &str = "/wallet/:public_key/balance";
+    pub const TRANSACTIONS_TRANSFER_FUNDS: &str = "/wallet/transactions/transfer-funds";
+}
```
danielSanchezQ, line 31: Why is this in a mod?

nodes/api-common/Cargo.toml
danielSanchezQ: Yes, no needed, better its just available.

nodes/nomos-node/node/src/api/handlers.rs
```diff
@@ -1148,3 +1160,198 @@ where
         Err(error) => (StatusCode::INTERNAL_SERVER_ERROR, error.to_string()).into_response(),
     }
 }
+
+#[cfg(feature = "wallet")]
+pub mod wallet {
```
danielSanchezQ, line 1026: this is big enough to be its own file

nodes/nomos-node/node/src/api/handlers.rs
```diff
@@ -1148,3 +1160,198 @@ where
...
+        SamplingBackend,
+        SamplingNetworkAdapter,
+        SamplingStorage,
+        StorageAdapter,
+        TimeBackend,
+        RuntimeServiceId,
+    >(
+        State(handle): State<OverwatchHandle<RuntimeServiceId>>,
+        Path(public_key): Path<PublicKey>,
+        Query(query): Query<TipQuery>,
```
danielSanchezQ, line 1053: Do we need the tip? I guess we want the balance from the current one (?) cc @davidrusu

nodes/nomos-node/node/src/api/handlers.rs
```diff
@@ -1148,3 +1160,198 @@ where
...
+                Ok(relay) => relay,
+                Err(error_response) => return error_response,
+            };
+            WalletApi::<WalletService, RuntimeServiceId>::new(wallet_relay)
+        };
+        let tip = {
+            if let Some(tip) = query.tip {
+                tip
+            } else if let Ok(info) = consensus::cryptarchia_info(&handle).await {
+                info.tip
```
danielSanchezQ, line 1091: Ok, so we default to current tip if none is provided 🤔 , sounds good then.

nomos-services/wallet/src/api.rs
```diff
@@ -88,6 +88,20 @@ where
...
+        change_pk: PublicKey,
+        funding_pks: Vec<PublicKey>,
+        recipient_pk: PublicKey,
+        amount: Value,
+    ) -> Result<nomos_core::mantle::SignedMantleTx, DynError> {
+        let mantle_tx_builder =
+            MantleTxBuilder::new().add_ledger_output(Note::new(amount, recipient_pk));
+        self.fund_and_sign_tx(tip, mantle_tx_builder, change_pk, funding_pks)
+            .await
+    }
```
danielSanchezQ, line 103: Nice, this is exactly whats needed later for the logos-core exposure.

wallet-http-client/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+        wallet_address: PublicKey,
+        tip: Option<HeaderId>,
+    ) -> Result<Url, Error> {
+        let Ok(wallet_address) = wallet_address.to_bytes() else {
+            return Err(Error::Client(String::from(
+                "The wallet address is not a valid public key.",
+            )));
+        };
+        let wallet_address = hex::encode(wallet_address.iter().as_slice());
+        let path = paths::wallet::BALANCE.replace(":public_key", wallet_address.as_str());
```
danielSanchezQ, line 39: I'm pretty sure there should be a Url construction for this instead of plain text replacement.

nodes/nomos-node/node/src/api/handlers.rs
strinnityk: I thought about that but that woudl've meant refactoring some files around, I'd rather do that in different PR so it's easier to read.

nodes/api-common/src/paths.rs
strinnityk: It seemed like a good idea to namespace them, since it's starting to get a bit large and they were getting a bit unordered.

wallet-http-client/src/lib.rs
strinnityk: There isn't. I implemented a simple one but it seems a excessive for how things are now.

nodes/api-common/Cargo.toml
davidrusu: I just checked, adding this wallet feature flag would have added 50 new combinations to check in cargo hack 🤯. I think we really should avoid flags unless there's some strong need to compile some code out, it's killing the CI runtimes

nodes/nomos-node/node/src/api/responses/overwatch.rs
```diff
@@ -0,0 +1,22 @@
+use std::fmt::{Debug, Display};
+
+use axum::response::{IntoResponse as _, Response};
+use http::StatusCode;
+use overwatch::{
+    overwatch::OverwatchHandle,
+    services::{AsServiceId, ServiceData, relay::OutboundRelay},
+};
+
+pub async fn get_relay_or_response<Service, RuntimeServiceId>(
```
davidrusu: nit:
```suggestion
pub async fn get_relay_or_fail<Service, RuntimeServiceId>(
```
"response" could be both success or failure

nodes/nomos-node/node/src/api/handlers.rs
davidrusu: I think the current behaviour is good, clients can query their historical wallet state, or they can use latest

nodes/nomos-node/node/src/api/handlers.rs
```diff
@@ -1148,3 +1160,197 @@ where
...
+                    String::from("Couldn't retrieve a valid tip"),
+                )
+                    .into_response();
+            }
+        };
+
+        let balance = wallet_api.get_balance(tip, public_key).await;
+        match balance {
+            Ok(Some(balance)) => WalletBalanceResponseBody { balance }.into_response(),
+            Ok(None) => (StatusCode::NOT_FOUND, "Wallet not found").into_response(),
```
davidrusu: `None` signals that the `public_key` was not found in the wallet, ie. they requested the balance for a non-existent public key.

nodes/nomos-node/node/src/api/handlers.rs
```diff
@@ -1148,3 +1160,197 @@ where
...
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    String::from("Couldn't retrieve a valid tip"),
+                )
+                    .into_response();
+            }
+        };
+
+        let balance = wallet_api.get_balance(tip, public_key).await;
+        match balance {
+            Ok(Some(balance)) => WalletBalanceResponseBody { balance }.into_response(),
```
davidrusu: can we include the `tip` in the balance response? If they didn't provide a tip, the latest would have been used, it'd be good for the caller to know for what block this balance corresponds to.

nodes/nomos-node/node/src/api/handlers.rs
```diff
@@ -1148,3 +1160,197 @@ where
...
+                body.funding_public_keys,
+                body.recipient_public_key,
+                body.amount,
+            )
+            .await;
+
+        match transfer_funds {
+            Ok(transaction) => WalletTransferFundsResponseBody::from(transaction).into_response(),
+            Err(error) => (StatusCode::INTERNAL_SERVER_ERROR, error.to_string()).into_response(),
+        }
```
davidrusu, line 1207: few q's
1. can we have the WalletTransferFundsResponseBody contain the json-encoded transaction rather than just a hash? That way a CLI can present the final transaction to the user for approval
2. this api is a POST, I guess your thought was that we would directly publish the transaction to mempool? I think there should be a `GET /prepare_transfer` api that constructs a funded transfer tx and then a second `POST /sign_and_publish_tx` that receives a MantleTxBuilder from the user and calls the wallet sign api and then publishes it to the mempool.

davidrusu (commented):
Heads up, there's some conflicts with master that are not being detected by git, you'll need a rebase

nodes/nomos-node/node/src/api/handlers.rs
danielSanchezQ: Imo this is a local trusted API, there should be no need for confirmation after you send the request. I see the point on having it as two endpoints but I do not think it is really necessary at this point.

nodes/api-common/Cargo.toml
danielSanchezQ: > I just checked, adding this wallet feature flag would have added 50 new combinations to check in cargo hack 🤯. I think we really should avoid flags unless there's some strong need to compile some code out, it's killing the CI runtimes

I think this is really a CI issue. We should probably migrate cargo hack to be run on a nightly process instead of per PR.

nodes/nomos-node/node/src/api/handlers.rs
danielSanchezQ: > I think the current behaviour is good, clients can query their historical wallet state, or they can use latest

Yeah, I saw the behaviour later on during the review and forgot removing this comment.

nodes/nomos-node/node/src/api/handlers.rs
davidrusu: My thought was to confirm the transaction fee, because as it stands this api will calculate a fee for the user and spend their money without their confirmation.

danielSanchezQ: > My thought was to confirm the transaction fee, because as it stands this api will calculate a fee for the user and spend their money without their confirmation.

Uhm, true, that makes sense, in this case we can do this:
* Have a dry-run method that returns the pre-built tx.
* Have the sign and publish endopoint
* Make the current endpoint an strategy which compose inmmidiaty publish or a limit fee to abort publishing.

WDYT?

We can have both worlds, why not.

davidrusu: my preference is to keep the api surface small, just
- build tx
- sign and publish tx

Each additional variant is another thing to test no?

And It's how most wallets work these days anyways, they present the final transaction to the user, the user authorizes it, it gets signed and sent to the mempool.

I don't think we will have a usecase where we just blindly fund and publish transactions?

danielSanchezQ: Yeah, let's keep it simple. As we cannot know the fee before hand.

strinnityk: 1. We certainly can. Shall I just cast the `SignedMantleTx` to json and return it?
2. Is the confirmation step just a "yes/no" thing? If so, can the fees change between in the amount of time it takes to send request 1, check the response, and send request 2?

nodes/nomos-node/node/src/api/responses/overwatch.rs
strinnityk: Fair. But `fail` is also quite generic. I've gone for `get_relay_or_500`.

nodes/nomos-node/node/src/api/handlers.rs
davidrusu: 1. I hope its as simple as using serde_json
2. yes, just yes/no. and no the fees won't change, they are hardcoded in the transaction that would be presented to the user. So as long as you send back that same json encoded transaction, it will be the same fees. (i.e. don't regenerate the transaction after the user confirms, just send back the tx json to the `sign` api)

danielSanchezQ: I mean, there is no confirmation at all, if its submitted to the endpoint that's it (?)

davidrusu: Given that there will be two API's (build the funded transaction and a second to sign and publish the tx), the thought was that the caller would present a confirmation dialogue to the user.

danielSanchezQ: > Given that there will be two API's (build the funded transaction and a second to sign and publish the tx), the thought was that the caller would present a confirmation dialogue to the user.

Yes, thats correct. What I meant is that from the API side of things there is no validation, if its submitted to the endpoint its accepted inmidiately (we need to validate its correct tho).

davidrusu: right, the sign endpoint should just publish without confirming

danielSanchezQ: @strinnityk , are this proposed changes applied? I can't see the flow atm.

strinnityk: Most of them are, except the last one about splitting the functionality.
I talked with David about this, it might be a good idea to leave that split for a different PR, so as not to clutter this one.

PR #1940, opened 2025-11-19
feat(sdp): define BlendActivityProof in nomos-core

youngjoon-lee:

This is a foundational PR before implementing the blend reward calculation in `nomos-ledger`.

This PR fills in the placeholder in the `nomos_core::sdp`, by defining `BlendActivityProof` which can be converted into/from activity metadata bytes using `nom` (like `DABlendActivityProof`).

Plus, I split `nomos_core::sdp` into two new submodules: `nomos_core::sdp::da` and `nomos_core::sdp::blend`.

The ledger will be implemented in a next PR.

@ntn-x2 
Even though we already have `ActivityProof` in the `nomos-blend-message` crate, I had to re-define the `BlendActivityProof` in the `nomos_core::sdp::blend` because of its architecture and cyclic dependency. ~~But, to minimize duplication, I defined `BlendActivityProof` simpler than `ActivityProof`, by moving only several constants to the new module `nomos_core::blend`. I think this is a reasonable approach we can take for now. Later, we could consider moving more types from `nomos-blend-message` to `nomos_core::blend` to remove duplication completely. Or, we could even move `nomos_core::sdp` out of `nomos_core`, but I didn't want to change the structure in this PR.~~

I believe so. 

- implementer @youngjoon-lee 
- blend reviewer @ntn-x2 
- reward reviewer @zeegomo 
- sdp reviewer @bacv 
- spec @madxor 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/message/src/reward/activity.rs
+++ b/nomos-blend/message/src/reward/activity.rs
@@ -51,22 +51,8 @@ impl From<&ActivityProof> for nomos_core::sdp::blend::ActivityProof {
     fn from(proof: &ActivityProof) -> Self {
         Self {
             session: proof.session_number,
-            proof_of_quota: proof.token.proof_of_quota().into(),
-            proof_of_selection: proof.token.proof_of_selection().into(),
+            proof_of_quota: (*proof.token.proof_of_quota()).into(),
+            proof_of_selection: (*proof.token.proof_of_selection()).into(),
         }
-    }
-}
-
-impl TryFrom<&nomos_core::sdp::blend::ActivityProof> for ActivityProof {
-    type Error = Box<dyn std::error::Error>;
-
-    fn try_from(proof: &nomos_core::sdp::blend::ActivityProof) -> Result<Self, Self::Error> {
-        Ok(Self::new(
-            proof.session,
-            BlendingToken::new(
-                (&proof.proof_of_quota).try_into()?,
-                (&proof.proof_of_selection).try_into()?,
-            ),
-        ))
     }
 }
--- a/nomos-core/chain-defs/src/blend/mod.rs
+++ b/nomos-core/chain-defs/src/blend/mod.rs
@@ -1,8 +1,0 @@
-use poq::PoQProof;
-use poseidon2::ZkHash;
-
-pub const KEY_NULLIFIER_SIZE: usize = size_of::<ZkHash>();
-pub const PROOF_CIRCUIT_SIZE: usize = size_of::<PoQProof>();
-pub const PROOF_OF_QUOTA_SIZE: usize = KEY_NULLIFIER_SIZE.checked_add(PROOF_CIRCUIT_SIZE).unwrap();
-
-pub const PROOF_OF_SELECTION_SIZE: usize = size_of::<ZkHash>();
--- a/nomos-blend/proofs/src/quota/mod.rs
+++ b/nomos-blend/proofs/src/quota/mod.rs
@@ -73,6 +73,33 @@ impl PartialEq<VerifiedProofOfQuota> for ProofOfQuota {
     }
 }
 
+impl From<&ProofOfQuota> for [u8; PROOF_OF_QUOTA_SIZE] {
+    fn from(proof: &ProofOfQuota) -> Self {
+        let mut bytes = [0u8; PROOF_OF_QUOTA_SIZE];
+        bytes[..KEY_NULLIFIER_SIZE].copy_from_slice(&fr_to_bytes(&proof.key_nullifier));
+        bytes[KEY_NULLIFIER_SIZE..].copy_from_slice(&proof.proof.to_bytes());
+        bytes
+    }
+}
+
+impl TryFrom<[u8; PROOF_OF_QUOTA_SIZE]> for ProofOfQuota {
+    type Error = Box<dyn std::error::Error>;
+
+    fn try_from(bytes: [u8; PROOF_OF_QUOTA_SIZE]) -> Result<Self, Self::Error> {
+        let (key_nullifier_bytes, proof_circuit_bytes) = bytes.split_at(KEY_NULLIFIER_SIZE);
+        let key_nullifier = fr_from_bytes(key_nullifier_bytes).map_err(Box::new)?;
+        let (pi_a, pi_b, pi_c) = split_proof_components::<
+            <Bn254 as CompressSize>::G1CompressedSize,
+            <Bn254 as CompressSize>::G2CompressedSize,
+        >(proof_circuit_bytes.try_into().map_err(Box::new)?);
+
+        Ok(Self {
+            key_nullifier,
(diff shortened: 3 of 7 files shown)


nomos-core/chain-defs/src/blend/mod.rs
```diff
@@ -0,0 +1,9 @@
+use poq::PoQProof;
+use poseidon2::ZkHash;
+
+pub const KEY_NULLIFIER_SIZE: usize = size_of::<ZkHash>();
+pub const PROOF_CIRCUIT_SIZE: usize = size_of::<PoQProof>();
+pub const PROOF_OF_QUOTA_SIZE: usize = KEY_NULLIFIER_SIZE.checked_add(PROOF_CIRCUIT_SIZE).unwrap();
+
+const SELECTION_RANDOMNESS_SIZE: usize = size_of::<ZkHash>();
+pub const PROOF_OF_SELECTION_SIZE: usize = SELECTION_RANDOMNESS_SIZE;
```
ntn-x2: Why two different consts?

nomos-blend/message/src/crypto/proofs/quota/mod.rs
```diff
@@ -122,6 +121,33 @@ impl ProofOfQuota {
...
+        >(proof_circuit_bytes.try_into().map_err(Box::new)?);
+
+        Ok(Self {
+            key_nullifier,
+            proof: PoQProof::new(pi_a, pi_b, pi_c),
+        })
+    }
+}
+
+impl From<&ProofOfQuota> for [u8; PROOF_OF_QUOTA_SIZE] {
```
ntn-x2: Personal preference: do you think it's better to change this to be a `as_bytes` method instead?

nomos-core/chain-defs/src/sdp/mod.rs
```diff
@@ -350,18 +354,89 @@ fn parse_length_prefixed_bytes(input: &[u8]) -> IResult<&[u8], Vec<u8>> {
...
+
+fn parse_blend_activity_proof(input: &[u8]) -> IResult<&[u8], BlendActivityProof> {
+    let (input, version) = nom_u8(input)?;
+    if version != BLEND_ACTIVE_METADATA_VERSION_BYTE {
+        return Err(nom::Err::Error(nom::error::Error::new(
+            input,
+            nom::error::ErrorKind::Verify,
+        )));
+    }
+    let (input, session) = parse_session_number(input)?;
```
ntn-x2: Why noy storing the actual types instead of byte arrays in `BlendActivityProof`? Here all we know is that we can parse byte arrays with a given size. We don't know whether what's inside is actually a valid proof. Why not take the extra step of trying to convert the parsed bytes into actual `PoQ` and `PoSel`?

ntn-x2 (commented):
The split-up seems reasonable. I have a couple of small nits, and one bigger point about how we go from byte arrays to Blend activity proofs.

nomos-core/chain-defs/src/sdp/mod.rs
youngjoon-lee: Yeah. I thought about it. But, this `BlendActivityProof` is defined in the `nomos-core`. Whereas, `ProofOfQuota` and `ProofOfSelection` are defined in the `nomos-blend-message`. I thought about moving them into the `nomos-core` to remove duplication and byte arrays, but also didn’t want to move all the associated methods.

nomos-blend/message/src/crypto/proofs/quota/mod.rs
youngjoon-lee: I thought about it as well. At the beginning, I defined `to_bytes` and `from_bytes` (similar as the `from_bytes_unchecked` we already had). But, I realized that those names are conflict with the methods generated by [`nomos-core::codec::SerializeOp`](https://github.com/logos-co/nomos/blob/blend-activity-proof-in-nomos-core/nomos-core/chain-defs/src/codec/mod.rs#L12-L12) and [`DeserializeOp`](https://github.com/logos-co/nomos/blob/3ea9814f2044939effdcb0ab5f7bf13713477171/nomos-core/chain-defs/src/codec/mod.rs#L17). If you have any reason to prefer having methods, please let me know. I will think about other names.

nomos-core/chain-defs/src/blend/mod.rs
youngjoon-lee: 989fb61b I removed the duplicate.

I copied what was in the `nomos-blend-message` crate to here. I remember I defined `SELECTION_RANDOMNESS_SIZE` since I think it's more readable (personal preference).
Anyway, we can remove `SELECTION_RANDOMNESS_SIZE` since it's not used anywhere.

nomos-core/chain-defs/src/sdp/mod.rs
youngjoon-lee: I think I can move minimum components from `nomos-blend-message` to `nomos-core` to minimize duplication. To not make this PR too larger for other reviewers, I will do it as a separate PR. I think that change can be reviewed between only you and me.

youngjoon-lee: @ntn-x2 I opened a draft PR https://github.com/logos-co/nomos/pull/1950 to show how we could remove this duplication, but I personally prefer the approach in this PR. Please let me know your opinion.

nomos-blend/message/src/reward/activity.rs
```diff
@@ -46,3 +46,27 @@ pub fn activity_threshold(token_count_bit_len: u64, network_size_bit_len: u64) -
...
+impl From<&ActivityProof> for nomos_core::sdp::BlendActivityProof {
+    fn from(proof: &ActivityProof) -> Self {
+        Self {
+            session: proof.session_number,
+            proof_of_quota: proof.token.proof_of_quota().into(),
+            proof_of_selection: proof.token.proof_of_selection().into(),
+        }
+    }
+}
+
```
zeegomo: not necessarily a fan of the dependencies / duplicated structs here, but I think you're already aware of that :)

nomos-core/chain-defs/src/sdp/mod.rs
```diff
@@ -350,18 +366,98 @@ fn parse_length_prefixed_bytes(input: &[u8]) -> IResult<&[u8], Vec<u8>> {
     Ok((input, data.to_vec()))
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct BlendActivityProof {
```
zeegomo: maybe we could split those into separate modules?

nomos-blend/message/src/reward/activity.rs
youngjoon-lee: Yes. I will resolve this duplication after having all the pieces in place, because it requires moving many components across crates.

PR #1941, opened 2025-11-19
feat(ledger): blend reward calculation

youngjoon-lee:

(Stacked on the PR https://github.com/logos-co/nomos/pull/1940)

This PR implements the `nomos_ledger::sdp::rewards::blend` module that replaces `NoopRewards`. 
Now, the ledger calculates rewards for each blend provider. This PR doesn't implement distributing the rewards yet.

Diff looks large, but the key changes are in the `nomos-ledger::sdp::rewards`.

Spec:
- The activity proof is defined in [here](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df8124bd88cb58d2171ccf).
- Like DA, this activity proof is packed with a [activity message](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df817882dfc5ca30ec3c1c).
- The ledger should [calculate rewards](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81449b30f2b84d8d7cfa).
  - The ledger should validate activity proofs:
    - PoQ and PoSel in a activity proof must be [valid](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81b6ab5eea4e715193f1).
      - My PR doesn't implement this yet, because we need to inject more inputs to the `nomos-ledger::sdp`. I'm preparing a next PR for this.
    - The hamming distance of the activity proof must be [<= activity proof](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df818fbcb7d9decb5387d7).

I have several things to highlight:
- I split `nomos-leger::sdp::rewards.rs` into submodules under `nomos-ledger::sdp::rewards::`.
- Unlike `da::Rewards`,
  - `blend::Rewards` requires some parameters that are also used for the blend service. So, this PR defines a struct for those parameters. This is why this PR has many files changed, but diffs are small. Also, because of this, the `config.toml` contains some duplicates. In a next PR, I will remove duplication once @ntn-x2's PR [#1938](https://github.com/logos-co/nomos/pull/1938) is merged.
    - I tried other approaches (extending `ServiceParameters`) as well, but this turned out to be the most reasonable one.
  - `blend::Rewards` requires the epoch information when a new session starts. So, I updated the `Rewards::update_session` method.
  - Most of the logic required by `blend::Rewards` has been implemented in the `nomos-blend-message` crate, not in the `nomos-core` crate. So, now the `nomos_ledger` crate has `nomos-blend-message` as a dependency.
- This PR doesn't implement verifying PoQ and PoSel included in an activity proof. That will be implemented in a next PR, because I don't want to make this PR large. We need a fair amount of engineering work to implement it.

I believe so.

- dev @youngjoon-lee 
- reviewer
  - blend @ntn-x2 
  - sdp @bacv 
  - rewards @zeegomo 
- @madxor 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.


ledger/nomos-ledger/src/mantle/sdp/rewards.rs
```diff
@@ -275,8 +262,209 @@ impl Rewards for DaRewards {
...
+        session_state: SessionState,
+        /// Session randomness for the session `s`.
+        next_session_randomness: SessionRandomness,
+        /// Proofs submitted by providers in the session corresponding to
+        /// `session_state`.
+        submitted_proofs: HashTrieMapSync<ProviderId, u64>,
+        /// Tracking the minimum Hamming distance among submitted proofs.
+        min_hamming_distance: u64,
+        /// Settings that don't change per session.
+        settings: BlendRewardsParameters,
```
ntn-x2: Just a nit (feel free to ignore), but if there's values that don't depend on sessions, as is `BlendRewardsParameters` that is used for both `Initialized` and `Uninitialized`, then perhaps we should use a struct with the common fields in there, and an enum field that includes state-specific elements. For the `Uninitialized` case, that would probably be a `None` then. In my opinion that is 1. easier to read, and 2. less prone to introduce mistakes going forward.

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
```diff
@@ -275,8 +262,209 @@ impl Rewards for DaRewards {
...
+                    });
+                }
+
+                let proof = nomos_blend_message::reward::ActivityProof::try_from(proof)
+                    .map_err(|_| Error::InvalidProofType)?;
+
+                // TODO: Validate PoQ and PoSel by holding all epoch infos that
+                // spanned across the `session_state.session_n` session.
+
+                let hamming_distance = proof.token().hamming_distance(
```
ntn-x2: Don't we need to check the distance is smaller than the node's activity threshold?

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
```diff
@@ -275,8 +262,209 @@ impl Rewards for DaRewards {
...
+                ..
+            } => {
+                // TODO: Calculate base rewards when session_income is added to config
+                // For now using placeholder value of 0
+                let session_income = 0;
+
+                // Identify premium providers with the minimum Hamming distance
+                let premium_providers = submitted_proofs
+                    .iter()
+                    .filter_map(|(&id, &hamming_distance)| {
```
ntn-x2: Why not keeping track directly of the providers with the smallest distance at the time when we evaluate a new active message instead?

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
youngjoon-lee: Sorry, I'm not sure if I understand your comment. Do you suggest something like:
```
struct BlendRewards {
  settings: BlendRewardsParameters
  state: BlendRewardsState
}

enum BlendRewardsState {
  Uninitialized,
  Initialized {
    ..
  }
}
```

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
youngjoon-lee: Good point. I forgot about it since it's not mentioned in the [spec](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81449b30f2b84d8d7cfa). I will ask Marcin to update the spec and implement it in this PR.

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
ntn-x2: Correct. Except that we can use an `Option::None` for the `Uninitialized` state, I guess.

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
ntn-x2: It is mentioned [here](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df8124bd88cb58d2171ccf) tho, that's how I noticed it wasn't implemented.

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
youngjoon-lee: We may be able to have another hashmap `<distance, hashset<providerId>>` to not iterate the original hashmap every time the min_hamming_distance is updated. But, I didn't do it since this operation (finding premium providers) happen once only when a session ends. 

But, let me try to optimize it

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
youngjoon-lee: Yes. It's implemented in the blend service side, but not in the ledger side yet. I will add it.

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
ntn-x2: We don't need another map, we just need to store the smallest distance and the list of providers that match that distance.

youngjoon-lee: > We don't need another map, we just need to store the smallest distance and the list of providers that match that distance.

But, then, whenever the smallest distance is updated (when a new active message is processed), we need to iterate the original hashmap to find the list of providers that have the smallest distance. That's why I store only the smallest distance, and find premium providers only at the end.

ntn-x2: Mhh I am not sure I understand. When we submit a new proof, we check if it's valid and we add it to the set of submitted proofs. Then, if the new proof is equal to the current smallest distance, we add the provider to the list of premium reward winners. If it's smaller, we replace the old distance and list with the new value. At the end, we will still have the list of winning nodes, and a list of premium winnings. I understand only the nodes with EXACTLY the same smallest distance win the premium, right? Why do we need to iterate the original hashmap again when we update the smallest distance? If we update the smallest distance it means that only this new node will win a premium reward, and all the nodes at the previous smallest distance will not.

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
youngjoon-lee: Implemented 933a17acb338f3a634adc09241b3b405f80272da

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
youngjoon-lee: ah, you're right. i was confused. i will optimize it.

youngjoon-lee: I implemented it 81b775cdd. Sorry for making confusions. My brain was totally stuck for a moment.

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
youngjoon-lee: I agree that has more advantages. I will try it, but after others finish their reviews.

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
```diff
@@ -275,8 +258,255 @@ impl Rewards for DaRewards {
...
+
+impl MinHammingDistance {
+    fn new() -> Self {
+        Self {
+            distance: u64::MAX,
+            providers: HashTrieSetSync::new_sync(),
+        }
+    }
+
+    fn update(&self, distance: u64, provider: ProviderId) -> Self {
```
ntn-x2: I can't find the comment I just left. I wrote that this is confusing. We either mutate self with `&mut self`, or we consume `self` and return `Self`, in my opinion. Either way it's fine for me.

danielSanchezQ: But we want the previous state @ntn-x2 as well. Maybe `with_update` instead (?)

youngjoon-lee: I see your point. I thought about it as well, but followed the same convention of the `Rewards` trait, which has the following method:
```rust
fn update_active(&self, ...) -> Result<Self, Error>
```

Anyway I renamed it to `with_update` which I agree it's clearer, regardless of the convention. d03d1716

ntn-x2: Ok if it's done the same way in other places, I guess we can maintain a consistent structure, then update things later on. Was not aware of it.

ledger/nomos-ledger/src/cryptarchia/mod.rs
```diff
@@ -556,21 +569,24 @@ pub mod tests {
...
+        config: &Config,
     ) -> HeaderId {
         let id = update_ledger(ledger, parent, slot, utxo_proof).unwrap();
         // we still don't have transactions, so the only way to add a commitment to
         // spendable utxos and test epoch snapshotting is by doing this
         // manually
         let mut block_state = ledger.states[&id].clone().cryptarchia_ledger;
         block_state.utxos = block_state.utxos.insert(utxo_add.id(), utxo_add).0;
-        ledger.states.insert(id, full_ledger_state(block_state));
+        ledger
```
zeegomo, line 579: unless the config changes could you take it from the existing state?

nomos-blend/message/src/reward/session.rs
```diff
@@ -106,39 +131,56 @@ fn token_count_bit_len(total_core_quota: u64) -> Result<u64, Error> {
...
 
+pub fn activity_threshold(token_count_bit_len: u64, num_core_nodes: u64) -> Result<u64, Error> {
+    let network_size_bit_len = F64Ge1::try_from(
+        num_core_nodes
+            .checked_add(1)
+            .ok_or(Error::NetworkSizeTooLarge(num_core_nodes))?,
+    )
+    .map_err(|()| Error::NetworkSizeTooLarge(num_core_nodes))?
+    .log2()
+    .ceil() as u64;
```
zeegomo, line 140: is this a double log?  see also lines 55 where we already apply it

nomos-core/chain-defs/src/blend/mod.rs
```diff
@@ -6,3 +9,22 @@ pub const PROOF_CIRCUIT_SIZE: usize = size_of::<PoQProof>();
...
+    membership_size: usize,
+) -> u64 {
+    // `C`: Expected number of cover messages that are generated during a session by
+    // the core nodes.
+    let expected_number_of_session_messages =
+        rounds_per_session.get() as f64 * message_frequency_per_round.get();
+
+    // `Q_c`: Messaging allowance that can be used by a core node during a single
+    // session. We assume `R_c` to be `0` for now, hence `Q_c = ceil(C * (ß_c
+    // + 0 * ß_c)) / N = ceil(C * ß_c) / N`.
```
zeegomo, line 19: why assuming 0?

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
zeegomo, line 1: we could also split da and blend rewards in different files

nomos-core/chain-defs/src/blend/mod.rs
ntn-x2: We do not use this parameter at the moment, at least not in Blend. The formula includes it, but then we say this is not used for v1, so since if this were to be different than `0` we would still not do anything with it, we removed from our formula implementation and use `0` instead. Cc @madxor

nomos-blend/message/src/reward/session.rs
youngjoon-lee: Oh, thank you. I made this bug when refactoring this module in this PR. I fixed it and added a test: c8413ae96d5c1e31421a81126851b9777f030683

ledger/nomos-ledger/src/cryptarchia/mod.rs
youngjoon-lee: True. Thanks. Resolved 56e47ce97a65c56fdda5fe6752ba39353a4b54e5

nomos-core/chain-defs/src/blend/mod.rs
youngjoon-lee: Yeah, Antonio explained it nicely. I actually moved this logic from `nomos-services/blend/src/core/settings.rs`, but forgot to add a PR comment here about it.

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
youngjoon-lee: Yeah, I was going to split them after you finish your review to not make this PR larger. But, now that you finished your first round, I split them now: 07ee3225
I hope this doesn't make your second round of review too hard.

nomos-core/chain-defs/src/blend/mod.rs
zeegomo: Just want to make sure spec and implementation are in sync, as long as @madxor is happy I'm fine, but we might have to update one of the two

ntn-x2: Yes. V1 will not really use this parameter, it's only defined but assumed to be `0`. Hence we simplified the code by removing a parameter we were moving around without really using it, even if it was given any other value than `0`.

madxor: Yes, I'm fine with this code simplification, as long as we have that nice comment :)

ledger/nomos-ledger/src/mantle/sdp/rewards.rs
youngjoon-lee: In the next PR, I will introduce a nicer structure. I already implemented it but waiting until this PR is merged

danielSanchezQ (approved):

PR #1943, opened 2025-11-19
chore(blend): remove const encapsulation count from Blend stack

ntn-x2:

Number of blending layer is a dynamic config, so we can't have `const`s in our codebase. Also, the `nomos-blend-message` stack is not really concerned about encapsulation layers at all, since that is something available to the cryptographic processor, hence we removed all such logic from somewhere so deep in the stack.

As a minor optimization, we also validate the payload body to blend before requesting the proof generator to generate and return a set of `PoQ` proofs, in case the payload is invalid (i.e., too large).

Yes.

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/message/Cargo.toml
+++ b/nomos-blend/message/Cargo.toml
@@ -13,22 +13,21 @@ version     = { workspace = true }
 workspace = true
 
 [dependencies]
-blake2          = { workspace = true }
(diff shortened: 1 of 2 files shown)


youngjoon-lee (approved):
thank you

PR #1949, opened 2025-11-20
feat(PoL): Implement and integrate slot secrets

danielSanchezQ:

This PR implements the slot secret merkle tree generation and integrates into the leader module. **It does not include** secret erasure or secret renewal which will be included in a further pr (I prefer to keep it short as we iterated a lot till having this version with "proper" times).

Yes, latest version is quite simple to read. Basic idea is to have a splitted tree with a cached part and a part that need to be recalculated, for ram saving purposes. Generation side of things is just a reduce operation and some index calculations to chose the proper merkle paths.

@thomaslavaur @davidrusu @danielSanchezQ 

We should change the spec of the secrets generation as the safely optimal version of secrets leafs chain was **way too slow** to compute.

No changes.

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/chain/chain-leader/benches/pol_sk.rs
+++ b/nomos-services/chain/chain-leader/benches/pol_sk.rs
@@ -1,5 +1,5 @@
 use blake2::Digest as _;
-use chain_leader::pol::merkle::{MerklePolCache, MerklePolSubtree};
+use chain_leader::pol::merkle::{CachedPoLMerkleTree, PolMerkleSubtree};
 use cryptarchia_engine::Slot;
 use divan::{Bencher, black_box};
 use groth16::{Fr, fr_from_bytes};
@@ -18,7 +18,7 @@ fn precompute_slot_secret(bencher: Bencher, (tree_depth, cache_depth): (usize, u
             BlakeRng256Seed::from(seed_bytes)
         })
         .bench_values(|seed| {
-            black_box(MerklePolCache::new(
+            black_box(CachedPoLMerkleTree::new(
                 seed,
                 Slot::new(0),
                 tree_depth,
@@ -34,7 +34,7 @@ fn compute_non_cached_subtree(bencher: Bencher) {
             let seed: Fr = fr_from_bytes(b"1987").unwrap();
             seed
         })
-        .bench_values(|seed| black_box(MerklePolSubtree::new(seed, 5).merkle_path_for_index(0)));
+        .bench_values(|seed| black_box(PolMerkleSubtree::new(seed, 5).merkle_path_for_index(0)));
 }
 
 #[divan::bench(sample_count = 10, sample_size = 10)]
@@ -48,7 +48,7 @@ fn precompute_leaves(bencher: Bencher) {
         })
         .bench_values(|seed| {
             black_box(
-                MerklePolCache::leaves_from_seed(seed)
+                CachedPoLMerkleTree::leaves_from_seed(seed)
                     .take(2usize.pow(20))
                     .collect::<Vec<_>>(),
             )
--- a/nomos-services/chain/chain-leader/src/leadership.rs
+++ b/nomos-services/chain/chain-leader/src/leadership.rs
@@ -13,14 +13,14 @@ use serde::{Deserialize, Serialize};
 use tokio::sync::watch::Sender;
 
 #[cfg(not(feature = "pol-dev-mode"))]
-use crate::pol::{MAX_TREE_DEPTH, merkle::MerklePolCache};
+use crate::pol::{MAX_TREE_DEPTH, merkle::CachedPoLMerkleTree};
 
 #[derive(Clone)]
 pub struct Leader {
     sk: UnsecuredZkKey,
     config: nomos_ledger::Config,
     #[cfg(not(feature = "pol-dev-mode"))]
-    merkle_pol_cache: MerklePolCache,
+    merkle_pol_cache: CachedPoLMerkleTree,
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone)]
@@ -51,7 +51,7 @@ impl Leader {
                 .collect::<Vec<_>>()
                 .try_into()
                 .expect("Size should match");
-            MerklePolCache::new(
+            CachedPoLMerkleTree::new(
                 seed.into(),
                 starting_slot,
                 MAX_TREE_DEPTH as usize,
@@ -184,7 +184,7 @@ impl Leader {
         let slot_secret: Fr = {
             #[cfg(not(feature = "pol-dev-mode"))]
             {
-                *self.merkle_pol_cache.root_slot_secret().as_ref()
+                self.merkle_pol_cache.slot_secret_for_slot(slot)
             }
             #[cfg(feature = "pol-dev-mode")]
             {
--- a/nomos-services/chain/chain-leader/src/pol/merkle.rs
+++ b/nomos-services/chain/chain-leader/src/pol/merkle.rs
@@ -1,6 +1,6 @@
 use blake2::Digest as _;
 use cryptarchia_engine::Slot;
-use groth16::{Fr, fr_from_bytes_unchecked, fr_to_bytes};
+use groth16::{Fr, GROTH16_SAFE_BYTES_SIZE, fr_from_bytes_unchecked, fr_to_bytes};
 use nomos_core::{
     crypto::{ZkDigest, ZkHasher},
     utils::merkle::{MerkleNode, MerklePath},
@@ -17,14 +17,14 @@ use crate::pol::SlotSecret;
 pub type CachedTree = Vec<Vec<Fr>>;
 
 #[derive(Clone)]
-pub struct MerklePolCache {
+pub struct CachedPoLMerkleTree {
     pub cached_tree: CachedTree,
-    pub lower_sub_trees: Vec<MerklePolSubtree>,
+    pub lower_sub_trees: Vec<PolMerkleSubtree>,
     pub tree_depth: usize,
     pub starting_slot: Slot,
 }
 
-impl MerklePolCache {
+impl CachedPoLMerkleTree {
     #[must_use]
     pub fn new(
         seed: BlakeRng256Seed,
@@ -32,14 +32,15 @@ impl MerklePolCache {
         tree_depth: usize,
         cache_depth: usize,
     ) -> Self {
-        let sub_tree_depth = tree_depth - cache_depth;
+        let sub_tree_depth = tree_depth.saturating_sub(cache_depth);
         let number_of_subtrees = 2usize.pow(cache_depth as u32);
         // we have to pre-cache the leaves so we can parallelize the subtree computation
         let lower_sub_trees: Vec<_> = Self::leaves_from_seed(seed)
             .take(number_of_subtrees)
-            .map(|seed| MerklePolSubtree::new(seed, sub_tree_depth))
+            .map(|seed| PolMerkleSubtree::new(seed, sub_tree_depth))
             .collect();
 
+        // get the roots for the subtrees
         let lower_tree_roots: Vec<Fr> = lower_sub_trees
             .par_iter()
             .map(|subtree| {
@@ -62,10 +63,11 @@ impl MerklePolCache {
         }
     }
 
+    /// Compute the leaves (seeds of the subtrees) from a seed
     pub fn leaves_from_seed(seed: BlakeRng256Seed) -> impl Iterator<Item = Fr> {
         let mut rng = BlakeRng256::from_seed(seed);
         std::iter::repeat_with(move || {
-            let mut bytes = [0u8; 31];
+            let mut bytes = [0u8; GROTH16_SAFE_BYTES_SIZE];
             rng.fill_bytes(&mut bytes);
             fr_from_bytes_unchecked(&bytes)
         })
@@ -79,46 +81,128 @@ impl MerklePolCache {
 
     #[must_use]
     pub const fn cache_depth(&self) -> usize {
-        self.cached_tree.len()
+        self.cached_tree.len().saturating_sub(1)
     }
 
-    #[must_use]
-    pub fn merkle_path_for_slot(&self, slot: Slot) -> MerklePath<Fr> {
-        let index = usize::try_from(slot.into_inner() - self.starting_slot.into_inner())
-            .expect("Slot difference should always fit in usize");
-        let mut cached_path = get_merkle_path(&self.cached_tree, index, self.cached_tree.len());
+    /// Compute the location of a slot secret in reference to the internal
+    /// location. # Returns
+    /// 1) The index of the slot in the internal tree.
+    /// 2) The index of the subtree in the internal tree.
+    /// 3) The index of the leaf in the subtree.
+    fn indexes_from_slot(&self, slot: Slot) -> (usize, usize, usize) {
+        debug_assert!(
+            slot >= self.starting_slot,
+            "Slots should not be checked for the past"
+        );
+        let index = usize::try_from(
+            slot.into_inner()
+                .saturating_sub(self.starting_slot.into_inner()),
+        )
+        .expect("Slot difference should always fit in usize");
         let subtree_leaf_length = 2usize.pow((self.tree_depth - self.cache_depth()) as u32);
         let subtree_index = index / subtree_leaf_length;
         let leaf_index = index % subtree_leaf_length;
+        (index, subtree_index, leaf_index)
+    }
+
+    /// Generates the Merkle path for a given `slot`, **it does not contain the
+    /// root**.
+    #[must_use]
+    pub fn merkle_path_for_slot(&self, slot: Slot) -> MerklePath<Fr> {
+        let (index, subtree_index, leaf_index) = self.indexes_from_slot(slot);
+        let mut cached_path = get_merkle_path(&self.cached_tree, index, self.cached_tree.len());
         let current_path = self.lower_sub_trees[subtree_index].merkle_path_for_index(leaf_index);
         cached_path.extend(current_path);
         cached_path
     }
+
+    #[must_use]
+    pub fn slot_secret_for_slot(&self, slot: Slot) -> Fr {
+        let (_, subtree_index, leaf_index) = self.indexes_from_slot(slot);
+        self.lower_sub_trees[subtree_index].slot_secret_for_index(leaf_index)
+    }
 }
 
+/// Represents a partial Merkle subtree.
+///
+/// This structure encapsulates the initial seed and the depth of the subtree,
+/// enabling the re-computation of the subtree's leafs secrets.
 #[derive(Clone)]
-pub struct MerklePolSubtree {
+pub struct PolMerkleSubtree {
+    /// A field element (`Fr`) that serves as the initial seed for constructing
+    /// the Merkle subtree
     seed: Fr,
+    /// The depth of the subtree, i.e. the number of levels in the tree
     tree_depth: usize,
 }
 
-impl MerklePolSubtree {
+impl PolMerkleSubtree {
     #[must_use]
     pub const fn new(seed: Fr, tree_depth: usize) -> Self {
         Self { seed, tree_depth }
     }
 
+    /// Generates a Merkle path for a given leaf index in the Merkle tree.
+    ///
+    /// # Arguments
+    ///
+    /// * `index` - The index of the leaf node for which the Merkle path is to
+    ///   be generated.
+    ///
+    /// # Returns
+    ///
+    /// A `MerklePath` object containing the path from the specified leaf index
+    /// to the root of the Merkle tree.
     #[must_use]
     pub fn merkle_path_for_index(&self, index: usize) -> MerklePath<Fr> {
+        let hashed_leafs = self.compute_leafs();
+        let merkle_tree = compute_cached_tree_from_leafs(&hashed_leafs);
+        get_merkle_path(&merkle_tree, index, self.depth())
+    }
+
+    /// Computes the leaf nodes of a Merkle tree using a pseudorandom hashing
+    /// process.
+    ///
+    /// This function generates leaf nodes based on the `seed` initialized in
+    /// the struct and iteratively hashes it to produce a sequence of
+    /// pseudorandom values. The hashing process employs the `Blake2b256`
+    /// algorithm and constrains the output size to `GROTH16_SAFE_BYTES_SIZE` so
+    /// it fits in a `Fr`.
+    ///
+    /// # Returns
+    /// A `Vec<Fr>` containing all the computed leaf nodes of the Merkle tree.
+    /// The number of leaf nodes is determined by the tree depth:
+    /// `2^(tree_depth)`.
+    fn compute_leafs(&self) -> Vec<Fr> {
         let hashed_leafs: Vec<_> = std::iter::successors(Some(self.seed), |seed| {
             Some(fr_from_bytes_unchecked(
-                &Blake2b256::digest(fr_to_bytes(seed))[..31],
+                &Blake2b256::digest(fr_to_bytes(seed))[..GROTH16_SAFE_BYTES_SIZE],
             ))
         })
         .take(2usize.pow(self.tree_depth as u32))
         .collect();
-        let merkle_tree = compute_cached_tree_from_leafs(&hashed_leafs);
-        get_merkle_path(&merkle_tree, index, self.depth())
+        hashed_leafs
+    }
+
+    /// Retrieves the secret associated with the specified slot index.
+    ///
+    /// This function computes the hashed leaf nodes of the corresponding data
+    /// and returns the secret value at the given `index`.
+    ///
+    /// # Parameters
+    /// - `index`: The position of the slot for which the secret is needed. Must
+    ///   be within the bounds of the computed hashed leaf nodes array.
+    ///
+    /// # Returns
+    /// - `Fr`: The secret value located at the specified index.
+    ///
+    /// # Panics
+    /// This function will panic if `index` is out of bounds of the
+    /// `hashed_leafs` array.
+    #[must_use]
+    pub fn slot_secret_for_index(&self, index: usize) -> Fr {
+        let hashed_leafs = self.compute_leafs();
+        hashed_leafs[index]
     }
 
     #[must_use]
@@ -128,6 +212,7 @@ impl MerklePolSubtree {
 }
 
 fn compute_cached_tree_from_leafs(tree_leafs: &[Fr]) -> Vec<Vec<Fr>> {
+    // reduce a tree from the leafs hashes
     let mut cached_tree: Vec<Vec<Fr>> = std::iter::successors(Some(tree_leafs.to_vec()), |leafs| {
         if leafs.len() <= 1 {
             return None;
@@ -158,17 +243,18 @@ pub fn get_merkle_path(
     for level_idx in (1..tree_depth).rev() {
         // sibling is the adjacent index: flip the lowest bit
         let sibling_index = current_index ^ 1;
-
-        if sibling_index < cached_tree[level_idx].len() {
-            let sibling_value = cached_tree[level_idx][sibling_index];
-            // Orientation is relative to the current node position
-            let node = if sibling_index.is_multiple_of(2) {
-                MerkleNode::Left(sibling_value)
-            } else {
-                MerkleNode::Right(sibling_value)
-            };
-            path.push(node);
-        }
+        assert!(
+            sibling_index < cached_tree[level_idx].len(),
+            "Tree should be constructed properly"
+        );
+        let sibling_value = cached_tree[level_idx][sibling_index];
+        // Orientation is relative to the current node position
+        let node = if sibling_index.is_multiple_of(2) {
+            MerkleNode::Left(sibling_value)
+        } else {
+            MerkleNode::Right(sibling_value)
+        };
+        path.push(node);
 
         // move to index for the next level up
         current_index /= 2;
@@ -180,6 +266,10 @@ pub fn get_merkle_path(
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    fn get_test_seed() -> BlakeRng256Seed {
+        BlakeRng256Seed::from([42u8; 32])
+    }
 
     // Helper to build a small deterministic tree and verify paths
     fn build_cached_tree() -> CachedTree {
@@ -246,6 +336,40 @@ mod tests {
                 assert_node_match(a, b);
                 assert_eq!(*a.item(), *b.item());
             }
+        }
+    }
+
+    #[test]
+    fn cache_depth_is_correct() {
+        let tree_depth = 4;
+        let cache_depth = 2;
+        let merkle_cache =
+            CachedPoLMerkleTree::new(get_test_seed(), Slot::new(0), tree_depth, cache_depth);
+        assert_eq!(merkle_cache.cache_depth(), cache_depth);
+    }
+
+    #[test]
+    fn slot_secret_for_slot_is_correct() {
+        let tree_depth = 4;
+        let cache_depth = 2;
+        let merkle_cache =
+            CachedPoLMerkleTree::new(get_test_seed(), Slot::new(0), tree_depth, cache_depth);
+
+        for slot_number in 0..16 {
+            let slot = Slot::new(slot_number);
+            // Get slot secret for slot
+            let slot_secret = merkle_cache.slot_secret_for_slot(slot);
+
+            // Calculate expected values
+            let subtree_leaf_length = 2usize.pow((tree_depth - cache_depth) as u32);
+            let expected_subtree_index = slot_number as usize / subtree_leaf_length;
+            let expected_leaf_index = slot_number as usize % subtree_leaf_length;
+
+            // Get expected slot secret
+            let expected = merkle_cache.lower_sub_trees[expected_subtree_index]
+                .slot_secret_for_index(expected_leaf_index);
+
+            assert_eq!(slot_secret, expected, "Failed for slot {slot_number}");
         }
     }
 }


nomos-services/chain/chain-leader/src/pol.rs
```diff
@@ -0,0 +1,61 @@
...
+pub fn pol_sk_generator(slot_start: Slot, seed: ZkHash) -> (SlotSecret, MerklePath<ZkHash>) {
+    let mut frontier_nodes: Mmr = Mmr::new();
+    let mut path = MerklePath::new();
+    let mut slot_secret = seed.into();
+
+    for i in 0..2u64.pow(u32::from(MAX_TREE_DEPTH)) {
+        frontier_nodes = frontier_nodes.push(slot_secret);
+        path.push(nomos_core::utils::merkle::MerkleNode::Right(
+            frontier_nodes.frontier_root(),
+        ));
```
danielSanchezQ: @davidrusu this was autocompleted, is not right, but should not mind for performance.

nomos-services/chain/chain-leader/src/pol.rs
```diff
@@ -0,0 +1,61 @@
...
+        slot_secret = ZkHasher::digest(&[seed]).into();
+    }
+
+    let root = frontier_nodes.frontier_root();
```
danielSanchezQ: Is the call to `frontier_root` ok here? Frontier root is actually a really heavy computation and it is done in each iteration.

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,242 @@
...
+    pub cache_depth: usize,
+    pub tree_depth: usize,
+    pub current_slot: Slot,
+}
+
+impl MerklePolCache {
+    #[must_use]
+    pub fn new(seed: Fr, tree_depth: usize, cache_depth: usize) -> Self {
+        let sub_tree_depth = tree_depth - cache_depth;
+        let numer_of_trees = 2usize.pow(sub_tree_depth as u32);
```
thomaslavaur: Isn't the number of trees 2.pow(cache_depth)? if you have a cache depth of 8 you will have 2^8 sub trees

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,242 @@
...
+        let total_leaves = 2usize.pow(tree_depth as u32);
+        let seed_jump = total_leaves / numer_of_trees;
+        // we have to pre-cache the leaves so we can parallelize the subtree computation
+        let seeds: Vec<Fr> = Self::leaves_from_seed(seed).take(total_leaves).collect();
```
thomaslavaur: I think It's not the correct way to do it. Here we still have all the 2^tree_depth secrets related to each other. That takes a lot of time to generate the first chain of hashes. IMO We should just get one random number for each subtree. That is you need one different seed (random from each other) for each seed

thomaslavaur: What I mean here is that you created a big iterator of length 2^tree_depth instead of creating number_of_trees iterators each of length 2^sub_tree_depth

thomaslavaur: Using this you can even parallelize the iterator derivation for each subtree

thomaslavaur: Yes I think it get harder than I expected like this. What I had in mind was you take 2^cache_depth seeds and just call the new_function of MerklePoLSubtree with this

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,242 @@
...
+
+        // compute the root slot secret
+        let cached_tree = Self::compute_cached_tree_from_leafs(lower_tree_roots);
+
+        Self {
+            cached_tree,
+            lower_merkle_pol: first_tree,
+            cache_depth,
+            tree_depth,
+            current_slot: Slot::new(0),
```
thomaslavaur: I think it would be nice to use a current_index and a starting index variable instead. So here the slot proven during PoL would be starting_slot + current_index

nomos-services/chain/chain-leader/src/pol/merkle.rs
danielSanchezQ: Yes, you are right, this is wrong!

nomos-services/chain/chain-leader/src/pol/merkle.rs
danielSanchezQ: > I think it would be nice to use a current_index and a starting index variable instead. So here the slot proven during PoL would be starting_slot + current_index

Yes, I agree. thought about it but finally didnt implemented it. Will do.

nomos-services/chain/chain-leader/src/pol/merkle.rs
danielSanchezQ: > I think It's not the correct way to do it. Here we still have all the 2^tree_depth secrets related to each other. That takes a lot of time to generate the first chain of hashes. IMO We should just get one random number for each subtree. That is you need one different seed (random from each other) for each seed

Yeah, I think we discussed this 🤔 , but that would break the slot chain from the spec right? Let's sync about this, just in case I'm not getting the idea properly.

thomaslavaur: But we can update the specs

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    #[must_use]
+    pub const fn cache_depth(&self) -> usize {
+        self.cached_tree.len()
+    }
+
+    #[must_use]
+    pub fn merkle_path_for_index(&self, index: usize) -> MerklePath<Fr> {
+        let mut cached_path = get_merkle_path(&self.cached_tree, index, self.cached_tree.len());
```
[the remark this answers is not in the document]
thomaslavaur: I've watch the code it should be the same

nomos-services/chain/chain-leader/src/leadership.rs
```diff
@@ -1,28 +1,67 @@
...
+    )]
+    pub fn new(
+        sk: UnsecuredZkKey,
+        starting_slot: Slot,
+        cache_depth: usize,
+        config: nomos_ledger::Config,
+    ) -> Self {
+        #[cfg(not(feature = "pol-dev-mode"))]
+        let merkle_pol_cache = {
+            let seed = Blake2b512::digest(fr_to_bytes(sk.to_public_key().as_fr()));
```
thomaslavaur: This is wrong you cannot use the secret key as a seed. Otherwise, you are high likely to be able to recover the seeds of all subtrees.
Most of the time the sk will be derived from a mnemonic and to be erasable you shouldn't be able to memorize it

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub fn root_slot_secret(&self) -> SlotSecret {
+        // this should always be there
+        self.cached_tree[0][0].into()
+    }
+
+    #[must_use]
+    pub const fn cache_depth(&self) -> usize {
+        self.cached_tree.len()
+    }
+
```
thomaslavaur, line 86: Would be nice to add a function merkle_path_for_slot() which return the merkle_path_for_index using as index slot-self.starting_slot

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,173 @@
...
+                .map(|pair| <[Fr; 2]>::try_from(pair).unwrap())
+                .map(|pair| <ZkHasher as ZkDigest>::compress(&pair))
+                .collect(),
+        )
+    })
+    .collect();
+    cached_tree.reverse();
+    cached_tree
+}
+
```
thomaslavaur, line 232: Can we add some test to check the merkle path is correct?

Just create a tree take the root and check that the path of a random index is valid

mmr/src/lib.rs
```diff
@@ -22,84 +22,92 @@ const EMPTY_VALUE: Fr = Fr::ZERO;
...
 
-impl<T, Hash, const MAX_HEIGHT: u8> Eq for MerkleMountainRange<T, Hash, MAX_HEIGHT> {}
+impl<T, Hash> Eq for MerkleMountainRange<T, Hash> {}
 
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-struct Root {
+pub struct Root {
     #[cfg_attr(feature = "serde", serde(with = "serde_fr"))]
     root: Fr,
```
ntn-x2, line 45: Not part of this PR, but this type is already `Root`. Maybe `value` would be a better name for this property? Else we would have `tree.root().root`.

mmr/src/lib.rs
```diff
@@ -22,84 +22,92 @@ const EMPTY_VALUE: Fr = Fr::ZERO;
...
+}
+
+impl<T, Hash> Default for MerkleMountainRange<T, Hash>
 where
     T: AsRef<Fr>,
     Hash: Digest,
 {
     fn default() -> Self {
-        Self::new()
+        Self::new(32)
```
ntn-x2: Should this be a constant (?)

mmr/src/lib.rs
```diff
@@ -22,84 +22,92 @@ const EMPTY_VALUE: Fr = Fr::ZERO;
...
+            height: 1,
+        };
         let mut roots = self.roots.clone();
+        let max_height = self.max_height;
 
         while let Some(root) = roots.peek().copied() {
+            // we want the frontier root to have a fixed height, so each individual root
+            // must be less than MAX_HEIGHT
+            assert!(
+                root.height <= self.max_height,
```
ntn-x2: Are we supposed to use `max_height` here instead?

mmr/src/lib.rs
```diff
@@ -22,84 +22,92 @@ const EMPTY_VALUE: Fr = Fr::ZERO;
...
             _hash: std::marker::PhantomData,
+            max_height: tree_depth,
         }
     }
 
     #[must_use]
-    pub fn push(&self, elem: T) -> Self {
-        let root = Hash::digest(&[*elem.as_ref()]);
-        let mut last_root = Root { root, height: 1 };
+    pub fn push(&self, root: T) -> Self {
```
ntn-x2: Are we pushing roots? Or we are pushing values which then becomes the new root?

nomos-services/chain/chain-leader/src/lib.rs
```diff
@@ -336,7 +339,27 @@ where
...
+
+        let SlotTick { slot, .. } = slot_timer
+            .next()
+            .await
+            .expect("Slot timer should emit at least one slot tick");
+
+        let leader = Leader::new(
+            leader_config.sk,
+            slot,
+            leader_config.cache_depth,
```
ntn-x2, line 360: Isn't `leader_config` already passed? We can probably not pass also one of its fields.

nomos-services/chain/chain-leader/src/leadership.rs
```diff
@@ -121,8 +168,28 @@ impl Leader {
     ) -> LeaderPrivate {
         // TODO: Get the actual witness paths and leader key
         let aged_path = Vec::new(); // Placeholder for aged path, aged UTXO tree is included in `EpochState`.
-        let latest_path = Vec::new();
-        let slot_secret = *self.sk.as_fr();
+        let latest_path: MerklePath<Fr> = {
```
ntn-x2: If `latest_path` does not use `latest_tree`, can `latest_tree` go?

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,173 @@
...
+impl MerklePolCache {
+    #[must_use]
+    pub fn new(
+        seed: BlakeRng256Seed,
+        starting_slot: Slot,
+        tree_depth: usize,
+        cache_depth: usize,
+    ) -> Self {
+        let sub_tree_depth = tree_depth - cache_depth;
+        let numer_of_sub_trees = 2usize.pow(cache_depth as u32);
```
ntn-x2:
```suggestion
        let number_of_subtrees = 2usize.pow(cache_depth as u32);
```

mmr/src/lib.rs
danielSanchezQ: yeah, I agree, but better in a different PR, not to add too much stuff to this already complicated one.

mmr/src/lib.rs
danielSanchezQ: Probably yes

mmr/src/lib.rs
danielSanchezQ: This is dinamic now, ill change the docs actually.

nomos-services/chain/chain-leader/src/lib.rs
danielSanchezQ: the configuration on this left me a bit uneasy as well, but didn't want to touch too much on the structure.

nomos-services/chain/chain-leader/src/leadership.rs
danielSanchezQ: true, ill have to check, maybe i actually used the wrong one here. cc @thomaslavaur

nomos-services/chain/chain-leader/src/leadership.rs
```diff
@@ -121,8 +168,28 @@ impl Leader {
...
+            // TODO: use proper u64 instead
+            #[cfg(not(feature = "pol-dev-mode"))]
+            {
+                self.merkle_pol_cache
+                    .merkle_path_for_index(slot.into_inner() as usize)
```
ntn-x2: Should't this be

```
slot.into_inner() - self.merkle_pol_cache.starting_slot.into_inner() as usize
```

nomos-services/chain/chain-leader/src/leadership.rs
danielSanchezQ: Yeah, I just forgot to remove it. I used stdrng, is not reproducible and its platfform dependant. I think it should be fine to use.

mmr/src/lib.rs
danielSanchezQ: kinda both? pushing a new elem makes as the root of the leave and then it modifies the overall roots. I'll revert to its original name. Not sure when was this renamed tho.

nomos-services/chain/chain-leader/src/leadership.rs
danielSanchezQ: indeed! good catch!

danielSanchezQ:
> Also, perhaps we should have some unit tests against some well-known values to make sure the implementation is sound?
Definitely

> Also, to confirm, we re-compute the lower trees for each path request to save memory space sacrificing computation time, correct? 
Yes

> Else this is repeated work that we could also somehow cache. We know we will be querying slots in order, so we could fully compute the subtree of a bunch of slots, keeping only a few full subtrees for the next `N` slots, for instance. We could pre-compute the subtree at the beginning of each 32-slot chunk, if feasible, and maintain only 2 full subtrees, so for 64 slots in total. This has modest increase in memory footprint (2 KB) but we would avoid re-calculating the subtree for each query, speeding up path computation after the first slot of each 32-chunk. Do you think that would be possible?

So, this is what we do with the cache already if I understood correctly. Essentially we just recompute the lower trees (which are quite fast, so query should be fine to be done JIT style). Anyway what you suggest may not even be worth, as you may recompute a slot for what you have not win. So its better to compute just for the ones you know you win and do the thing each time. The only major problem is when computing the overall root which is much slower. But if you still think it may be worth to try we can do it (as I may have not understood your proposition well).

nomos-services/chain/chain-leader/src/pol/merkle.rs
danielSanchezQ: yes, fixed

ntn-x2:
Keeping the computed sub-trees means the memory footprint will grow over time unless the binary is restarted, right? I think this is a separate problem related more to memory. My suggestion was to avoid wasting compute by generating a full sub-tree in advance instead of re-generating it for ever slot. Even better, if we know the next winning slot (and winning slots are always increasing), then we can pre-generate sub-trees for only the winning slots, keeping a window of a few "batches" of slots in advance. Then, second step would be to remove the old ones from memory once not used. But regardless of this, I think we can still implement pre-generation logic so we can avoid repeating the same work over and over again on path generation.

mmr/src/lib.rs
youngjoon-lee: It seems that we can remove the variable `max_height` and use `self.max_height` directly?

youngjoon-lee: Also, what do you think about moving this assertion to the previous location? If I understand correctly, we need to check if `last_root.height + 1 <= self.max_height`?

mmr/src/lib.rs
youngjoon-lee, line 171: Maybe, out of scope of this PR, but don't we need to replace all the magic numbers `32` with the `max_height` parameter?

nomos-services/chain/chain-leader/src/leadership.rs
```diff
@@ -1,28 +1,70 @@
...
     pub sk: UnsecuredZkKey,
+    pub cache_depth: usize,
 }
 
 impl Leader {
-    pub const fn new(sk: UnsecuredZkKey, config: nomos_ledger::Config) -> Self {
-        Self { sk, config }
+    #[cfg_attr(
+        feature = "pol-dev-mode",
+        expect(unused, reason = "Variables used in dev mode only"),
```
youngjoon-lee:
```suggestion
        expect(unused, reason = "Variables used in non dev mode only"),
```
I think we can modify the reason like this, because `starting_slot` and `cache_depth` are used only in **non**-dev mode?

nomos-services/chain/chain-leader/src/leadership.rs
```diff
@@ -110,9 +153,14 @@ impl Leader {
         None
     }
 
+    #[cfg_attr(
+        feature = "pol-dev-mode",
+        expect(unused, reason = "Variables ar only used in non-dev mode")
```
youngjoon-lee:
```suggestion
        expect(unused, reason = "Variables are only used in non-dev mode")
```

youngjoon-lee (commented):
Just left minor comments after the 1st round of review. I will take a closer look at `MerklePolCache` after revisiting the PoL spec.

nomos-services/chain/chain-leader/src/leadership.rs
youngjoon-lee: It seems that the latest_path should be obtained from the latest_tree, and that is not in the scope of this PR, if I understand correctly, as mentioned in the Iteration 17 note? If so, I think we can revert this to Vec::new() for now.

Instead, I think we need to compute `slot_secret_path` (probably, by calling `self.merkle_pol_cache.merkle_path_for_slot(slot)`), and pass it to `LeaderPrivate::new`.

nomos-services/chain/chain-leader/src/leadership.rs
youngjoon-lee, line 215: I think we should now implement this function properly using `_slot`?

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,252 @@
...
+    let mut path = MerklePath::new();
+
+    // Iterate from leaves level (tree_depth - 1) up to, but not including, the root
+    // level (0)
+    for level_idx in (1..tree_depth).rev() {
+        // sibling is the adjacent index: flip the lowest bit
+        let sibling_index = current_index ^ 1;
+
+        if sibling_index < cached_tree[level_idx].len() {
+            let sibling_value = cached_tree[level_idx][current_index];
```
youngjoon-lee:
```suggestion
            let sibling_value = cached_tree[level_idx][sibling_index];
```
I think we should use the sibling index.

nomos-services/chain/chain-leader/src/leadership.rs
danielSanchezQ: uhm, this looks like something bad went when merging, I recall implementing it (or maybe not :sweat_smile: ), but nice catch!

nomos-services/chain/chain-leader/src/leadership.rs
danielSanchezQ: I'm coming back to this after some time. I think I already did for both dev and non-dev modes.

mmr/src/lib.rs
danielSanchezQ: Deifnetely 51da5acd203a12df0a9c84da7209acba0dfb886a

nomos-services/chain/chain-leader/src/pol/merkle.rs
danielSanchezQ: I think the naming is unfortunate here. It is the value for the current index. I changed `sibling_balue` => `value` instead.

nomos-services/chain/chain-leader/src/leadership.rs
danielSanchezQ: actually this is just missnamed and have the wrong signature. This is only used for the winning ticket which needs the sk. No slots computation involved.

nomos-services/chain/chain-leader/src/pol/merkle.rs
youngjoon-lee: I think the merkle path should contain sibling values, not current values.
image
In this picture, I think the circled nodes should be included in the merkle path, if the target is `L0[0]`. So, the merkle path should be like below:
```
Right(L0[1])
Right(L1[1])
```
But, I think the current implementation computes a path like below, which I think it's not correct:
```
Left(L1[0])
Left(L0[0])
```

I made a similar implementation for `utxotree`: https://github.com/logos-blockchain/logos-blockchain/blob/311f164037bd5f12b79017a6ce0dbc1363945132/utxotree/src/merkle.rs#L202

Also, wrote tests to check if merkle paths can be actually used for the proof of membership: https://github.com/logos-blockchain/logos-blockchain/blob/311f164037bd5f12b79017a6ce0dbc1363945132/utxotree/src/merkle.rs#L675

I think we can adopt this practice for this PR, if I'm not missing anything.

nomos-services/chain/chain-leader/src/leadership.rs
youngjoon-lee: I think that the variable `latest_path` is not for $slot\\_secret\\_path$ (defined in the [spec](https://www.notion.so/nomos-tech/Proof-of-Leadership-Specification-21c261aa09df819ba5b6d95d0fe3066d?source=copy_link#21c261aa09df812badc8f579b798ad7b)). Instead, I think it's for $noteid\\_latest\\_path$ ([spec](https://www.notion.so/nomos-tech/Proof-of-Leadership-Specification-21c261aa09df819ba5b6d95d0fe3066d?source=copy_link#25a261aa09df80a081ffdb4f3fc41118)), as I implemented in the PR https://github.com/logos-blockchain/logos-blockchain/pull/2007.
IIUC, we need to introduce a new variable `slot_secret_path`.

nomos-services/chain/chain-leader/src/pol/merkle.rs
danielSanchezQ: Let me review the logic, did you check the test for this? I'll ping you later on.

danielSanchezQ: I had a big misconception. Was doing the leaf path instead of using the siblings for the markle one. Please take a look at f2ea26b00563b71c4e399ab69442487ba8551e0b
And thanks a lot!!

nomos-services/chain/chain-leader/src/leadership.rs
danielSanchezQ: > I think that the variable `latest_path` is not for s l o t _ s e c r e t _ p a t h (defined in the [spec](https://www.notion.so/nomos-tech/Proof-of-Leadership-Specification-21c261aa09df819ba5b6d95d0fe3066d?source=copy_link#21c261aa09df812badc8f579b798ad7b)). Instead, I think it's for n o t e i d _ l a t e s t _ p a t h ([spec](https://www.notion.so/nomos-tech/Proof-of-Leadership-Specification-21c261aa09df819ba5b6d95d0fe3066d?source=copy_link#25a261aa09df80a081ffdb4f3fc41118)), as I implemented in the PR #2007. IIUC, we need to introduce a new variable `slot_secret_path`.

yes you are right! just located the hole, thanks!
Edit: Should be fine now at 498faafddb5fcc64d1240ddacc101d256bce2e5e @youngjoon-lee

mmr/src/lib.rs
danielSanchezQ: Reverted and fixed in b6ce4fe9f7a146375b7da7402896ee134d86bf5e @youngjoon-lee

mmr/src/lib.rs
```diff
@@ -176,22 +193,18 @@ mod test {
...
     // bytes to poseidon field element
     fn b2p(b: &[u8]) -> Fr {
-        let mut repr = [0u8; 32];
-        assert!(b.len() <= 32);
-        let len = b.len().min(32);
+        let mut repr = [0u8; DEFAULT_MMR_DEPTH as usize];
+        assert!(b.len() <= DEFAULT_MMR_DEPTH as usize);
+        let len = b.len().min(DEFAULT_MMR_DEPTH as usize);
         repr[..len].copy_from_slice(&b[..len]);
         Fr::from_le_bytes_mod_order(&repr)
```
thomaslavaur, line 200: Should we check here that repr fitting in Fr field? because including 32 bytes it may overpass p

nodes/nomos-executor/config.yaml
```diff
@@ -178,6 +178,7 @@ cryptarchia:
   leader:
     pk: "0000000000000000000000000000000000000000000000000000000000000000"
     sk: 3cf3335017304dcfce45124a0633829f00000000000000000000000000000000
+    cache_depth: 25
```
thomaslavaur: It's a big cache depth! Do we want that much?

nodes/nomos-node/config.yaml
```diff
@@ -167,6 +167,7 @@ cryptarchia:
   leader:
     pk: "0000000000000000000000000000000000000000000000000000000000000000"
     sk: 3cf3335017304dcfce45124a0633829f00000000000000000000000000000000
+    cache_depth: 25
```
thomaslavaur: ditto

nomos-services/chain/chain-leader/benches/pol_sk.rs
```diff
@@ -0,0 +1,56 @@
...
+        .with_inputs(|| {
+            let seed = Blake2b256::digest(b"1987");
+            let seed_bytes: [u8; 32] = seed.into();
+
+            BlakeRng256Seed::from(seed_bytes)
+        })
+        .bench_values(|seed| {
+            black_box(
+                MerklePolCache::leaves_from_seed(seed)
+                    .take(2usize.pow(20))
```
thomaslavaur, line 52: Should we set it to 25? Because the real application requires you to directly generate the 2^25 leaves at the start

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,251 @@
...
+            let node = if sibling_index.is_multiple_of(2) {
+                MerkleNode::Left(sibling_value)
+            } else {
+                MerkleNode::Right(sibling_value)
+            };
+            path.push(node);
+        }
+
+        // move to index for the next level up
+        current_index /= 2;
```
thomaslavaur, line 260: nit: current_index << 2?

nomos-services/chain/chain-leader/src/leadership.rs
```diff
@@ -121,8 +169,28 @@ impl Leader {
...
+            #[cfg(not(feature = "pol-dev-mode"))]
+            {
+                self.merkle_pol_cache.merkle_path_for_slot(slot)
+            }
+            #[cfg(feature = "pol-dev-mode")]
+            {
+                Vec::new()
+            }
+        };
+        let slot_secret: Fr = {
```
thomaslavaur, line 184: No the slot secret shouldn't be the root of the cached tree nor the sk. It should be the deepest leaf value of the subtree.

tests/src/topology/configs/consensus.rs
```diff
@@ -143,6 +143,7 @@ pub fn create_consensus_configs(
                 leader: LeaderConfig {
                     pk,
                     sk: sk.into_unsecured(),
+                    cache_depth: 20,
```
thomaslavaur, line 146: still a lot of depth :0

nodes/nomos-executor/config.yaml
danielSanchezQ: no, this should be 20, it was the wrong value actually

mmr/src/lib.rs
danielSanchezQ: This is just for testing purposes, so it should be ok.

nomos-services/chain/chain-leader/src/pol/merkle.rs
danielSanchezQ: it probably makes no difference at the end as the compiler should actually do that :thinking:

nomos-services/chain/chain-leader/src/leadership.rs
danielSanchezQ: good catch!

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,251 @@
...
+
+impl MerklePolCache {
+    #[must_use]
+    pub fn new(
+        seed: BlakeRng256Seed,
+        starting_slot: Slot,
+        tree_depth: usize,
+        cache_depth: usize,
+    ) -> Self {
+        let sub_tree_depth = tree_depth - cache_depth;
```
ntn-x2: We do not perform any input validation, so this can very easily panic in debug and underflow in release. Since cache depth is always smaller than tree depth, we could use a range type, or then use a saturating subtraction instead.

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,251 @@
...
+pub struct MerklePolCache {
+    pub cached_tree: CachedTree,
+    pub lower_sub_trees: Vec<MerklePolSubtree>,
+    pub tree_depth: usize,
+    pub starting_slot: Slot,
+}
+
+impl MerklePolCache {
+    #[must_use]
+    pub fn new(
```
ntn-x2, line 29: Fine if part of a separate PR, but all these components should have clear comments and, if possible, some basic tests that both validate their logic and show how to use them.

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,251 @@
...
+    #[must_use]
+    pub const fn new(seed: Fr, tree_depth: usize) -> Self {
+        Self { seed, tree_depth }
+    }
+
+    #[must_use]
+    pub fn merkle_path_for_index(&self, index: usize) -> MerklePath<Fr> {
+        let hashed_leafs: Vec<_> = std::iter::successors(Some(self.seed), |seed| {
+            Some(fr_from_bytes_unchecked(
+                &Blake2b256::digest(fr_to_bytes(seed))[..31],
```
ntn-x2: I guess this is a constant for the tree height, perhaps it could be declared as a constant. Or at least `32` should, and this would be derived from that.

nomos-services/chain/chain-leader/src/leadership.rs
```diff
@@ -121,8 +169,28 @@ impl Leader {
     ) -> LeaderPrivate {
         // TODO: Get the actual witness paths and leader key
         let aged_path = Vec::new(); // Placeholder for aged path, aged UTXO tree is included in `EpochState`.
-        let latest_path = Vec::new();
-        let slot_secret = *self.sk.as_fr();
+        let latest_path = Vec::new(); // Placeholder for latest path.
+        let slot_secret_path: MerklePath<Fr> = {
```
ntn-x2, line 173: This is always fixed-length, right? I guess we can encode that knowledge in the `MerklePath` type, to enforce compile-time optimizations.

mmr/src/lib.rs
```diff
@@ -22,59 +23,66 @@ const EMPTY_VALUE: Fr = Fr::ZERO;
...
     #[must_use]
-    pub fn push(&self, elem: T) -> Self {
-        let root = Hash::digest(&[*elem.as_ref()]);
-        let mut last_root = Root { root, height: 1 };
+    pub fn push(&self, value: T) -> Self {
+        let mut last_root = Root {
+            root: *value.as_ref(),
+            height: 1,
+        };
         let mut roots = self.roots.clone();
```
ntn-x2, line 86: I guess we don't really need this line here.

mmr/src/lib.rs
```diff
@@ -22,59 +23,66 @@ const EMPTY_VALUE: Fr = Fr::ZERO;
...
 
         while let Some(root) = roots.peek().copied() {
```
ntn-x2, line 88: It's not a constant anymore.

```suggestion
            // must be less than `max_height`
```

nodes/nomos-executor/config.yaml
ntn-x2: Is this user-configurable? I believe so, right?

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,251 @@
...
+    iter::{IntoParallelRefIterator as _, ParallelIterator as _},
+    prelude::ParallelSlice as _,
+};
+
+use crate::pol::SlotSecret;
+
+pub type CachedTree = Vec<Vec<Fr>>;
+
+#[derive(Clone)]
+pub struct MerklePolCache {
```
ntn-x2: I think this naming is confusing. The fact that there is a cache underneath is an implementation detail. Why not call this `PolMerkleTree` directly? Or, since we pass information about the cache in the `new` function, why not `CachedPolMerkleTree`? What I am suggesting is that the tree is cached, but this is in itself not a cache, but a tree.

nomos-services/chain/chain-leader/src/pol/merkle.rs
ntn-x2: Also, isn't Merkle PoL tree of fixed-length?

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,251 @@
...
+            lower_sub_trees,
+            tree_depth,
+            starting_slot,
+        }
+    }
+
+    pub fn leaves_from_seed(seed: BlakeRng256Seed) -> impl Iterator<Item = Fr> {
+        let mut rng = BlakeRng256::from_seed(seed);
+        std::iter::repeat_with(move || {
+            let mut bytes = [0u8; 31];
```
ntn-x2: Why `31`?

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,251 @@
...
+
+    #[must_use]
+    pub fn root_slot_secret(&self) -> SlotSecret {
+        // this should always be there
+        self.cached_tree[0][0].into()
+    }
+
+    #[must_use]
+    pub const fn cache_depth(&self) -> usize {
+        self.cached_tree.len()
```
ntn-x2: I think this is not correct. If the input `cache_depth` value was `3` with a tree depth of `5`, now here this function will return `4` because it adds the root of the subtrees, so I think we should have a `-1` here.

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    #[must_use]
+    pub fn merkle_path_for_slot(&self, slot: Slot) -> MerklePath<Fr> {
+        let index = usize::try_from(slot.into_inner() - self.starting_slot.into_inner())
```
ntn-x2: This is also a regular `-`, but we don't know whether the input slot is greater than the starting slot, so we should have a checked subtraction here.

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,251 @@
...
+            .expect("Slot difference should always fit in usize");
+        let mut cached_path = get_merkle_path(&self.cached_tree, index, self.cached_tree.len());
+        let subtree_leaf_length = 2usize.pow((self.tree_depth - self.cache_depth()) as u32);
```
ntn-x2, line 102: This again I think it's not correct, since `self.cache_depth()` returns the wrong value.

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,251 @@
...
+#[must_use]
+pub fn get_merkle_path(
+    cached_tree: &CachedTree,
+    mut current_index: usize,
+    tree_depth: usize,
+) -> MerklePath<Fr> {
+    let mut path = MerklePath::new();
+
+    // Iterate from leaves level (tree_depth - 1) up to, but not including, the root
+    // level (0)
```
ntn-x2, line 242: Should we mention in the function name that the root is not part of the path?

nomos-services/chain/chain-leader/src/pol/merkle.rs
```diff
@@ -0,0 +1,251 @@
...
+    for level_idx in (1..tree_depth).rev() {
+        // sibling is the adjacent index: flip the lowest bit
+        let sibling_index = current_index ^ 1;
+
+        if sibling_index < cached_tree[level_idx].len() {
```
ntn-x2: What happens if this condition is not met?

nomos-services/chain/chain-leader/src/pol/merkle.rs
danielSanchezQ: Yes, its good catch! I realize when adding little testing.

nomos-services/chain/chain-leader/src/pol/merkle.rs
danielSanchezQ: > Also, isn't Merkle PoL tree of fixed-length?

Depth? it is 25 iirc.

nomos-services/chain/chain-leader/src/pol/merkle.rs
danielSanchezQ: 31 is not related to the tree height. But ill name it so its clear. Thanks.

nomos-services/chain/chain-leader/src/leadership.rs
danielSanchezQ: Yes merkle paths will be fixed size as the tree depth.

mmr/src/lib.rs
danielSanchezQ: `86`? We do as we return a new instance (?)
Not getting it probably.

nomos-services/chain/chain-leader/src/pol/merkle.rs
danielSanchezQ: Used constant instead.

nomos-services/chain/chain-leader/src/pol/merkle.rs
danielSanchezQ: Yes, realized with some test. Good catch, thanks!

nomos-services/chain/chain-leader/src/pol/merkle.rs
danielSanchezQ: it should panic actually, it should never happen. I will replace with an assert instead.

mmr/src/lib.rs
danielSanchezQ: I don't get this.

nomos-services/chain/chain-leader/src/pol/merkle.rs
danielSanchezQ: Yeah, was working on tests too.

nomos-services/chain/chain-leader/src/pol/merkle.rs
danielSanchezQ: Added a debug assert as well.

nomos-services/chain/chain-leader/src/pol/merkle.rs
ntn-x2: Yeah my point is that it should probably not be a parameter but rather a constant, which would then unlock both compiler optimizations and fixed-length merkle paths, which is what we want anyway.

mmr/src/lib.rs
ntn-x2: Yeah no somehow my VSCode must have messed stuff up. This is not meant to be here, sorry.

mmr/src/lib.rs
ntn-x2: Same as above comment, sorry. Not meant to be here.

nodes/nomos-executor/config.yaml
ntn-x2: If it is, and if we make the tree a fixed-depth tree, then upon deserialization we should fail if the cache value is grater than the tree depth. In that case, trying to run the node will fail instantly since the config file would fail to be deserialized.

nomos-services/chain/chain-leader/src/pol/merkle.rs
ntn-x2: I don't think a debug assert is enough, because on debug builds we already panic. The problem is in release builds, where we will wrap around `0`. So this should rather be a regular assert, valid both in debug and release builds. Or, even better, a checked subtraction.

danielSanchezQ: > I don't think a debug assert is enough, because on debug builds we already panic. The problem is in release builds, where we will wrap around `0`. So this should rather be a regular assert, valid both in debug and release builds. Or, even better, a checked subtraction.

yeah, I meant I would add the checked substration and the dbg assert, so it doesnt panic unless we are debugging. It was added yesterday already.

nomos-services/chain/chain-leader/src/pol/merkle.rs
danielSanchezQ: > Yeah my point is that it should probably not be a parameter but rather a constant, which would then unlock both compiler optimizations and fixed-length merkle paths, which is what we want anyway.

Yes would be better, I will do the refactor in a different PR tho.

PR #1950, opened 2025-11-21
refactor(blend): move minimum components from `nomos-blend-message` to `nomos-core`

youngjoon-lee:

This PR is a draft to show how to resolve https://github.com/logos-co/nomos/pull/1940#discussion_r2540897638. I personally doesn't prefer this solution over the PR #1940, but want to hear your opinion. At the end of this PR description, I left my opinion in detail.

In the PR https://github.com/logos-co/nomos/pull/1940, we had defined two duplicate structs:
- `nomos-blend-message::reward::ActivityProof`
- `nomos-core::sdp::BlendActivityProof`.

It was because all the components necessary for `ActivityProof` were defined in the `nomos-blend-message` (which has `nomos-core` as a dependency). To avoid cyclic dependency, I had to define `nomos-core::sdp::BlendActivityProof`, which is the semi-serialized version of `nomos-blend-message::reward::ActivityProof`.
This is similar as what DA does: https://github.com/logos-co/nomos/blob/78774b15bf12500d7d7adeaef2e4b8770d0c5492/nomos-services/data-availability/network/src/opinion_aggregator.rs#L41-L41.

But, as pointed out in https://github.com/logos-co/nomos/pull/1940#discussion_r2540897638, it's not so good to have this duplication. So, this PR removes the duplication by moving some components from `nomos-blend-message` to `nomos-core`. I moved only the data structures itself. Core logic (which isn't required by `nomos-core::sdp`) still lives in the `nomos-blend-message`.
- We have only `nomos-core::sdp::BlendActivityProof`. The `nomos-blend-message::reward::ActivityProof` has been removed.
- The following structs have been moved to the new `nomos-core::blend` module.
  - reward 
    - `BlendingToken` and its logic
    -  `SessionRandomness`: required by `BlendingToken`
  - proof  
    - `ProofOfQuota` and `ProofOfSelection`: required by `BlendingToken`
- The following structs and logic remain in the `nomos-blend-message`.
  - reward
    - `BlendingTokenCollector`: required only by `nomos-blend-service`
    - `SessionInfo`: required by `BlendingTokenCollector`
  - proof
    - Generating/verifying proofs

If you think this is not so better than having duplicate structs as defined in PR #1940, please let me know. We can close this PR and go back. I personally don't prefer having some components in the `nomos-core`, even though we already have `nomos-blend-message`. But I think this is the only way to remove duplication. 

Even if we merge this PR, we cannot remove a dependency graph `nomos-ledger::sdp --> nomos-blend-message` introduced in another PR https://github.com/logos-co/nomos/pull/1941 (this is another topic. though). It's because we also have to integrate the PoQ and PoSel verification to the `nomos-ledger::sdp` soon. Those verification logic lives in the `nomos-blend-message`. So, in my opinion, it's better to not move anything between crates for now. After having all the logic in place, we can consider the optimal crate structure.
 

Yes

@youngjoon-lee @ntn-x2 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.


youngjoon-lee:
> you have valid points not to overhaul the structure at the moment. I still think tho that we need to ensure that the byte array is not only of the right size (as we already do), but that it's also a valid PoQ. After all, once we work on actually verifying those proofs, don't we still need SDP to depend on nomos-blend-message, which would be a cyclic dependency? How do you think to solve that?

The PoQ verification will be implemented in the `nomos-ledger`, which already has `nomos-blend-message` as a dependency, as introduced in the PR #1941. In that PR, we're already checking if the byte array can be deserialized to the PoQ. In a next PR, I will implement verifying the PoQ deserialized. This is actually what I mentioned in the last paragraph of the PR description.

PR #1951, opened 2025-11-21
chore(blend): distinguish verified from unverified components

ntn-x2:

This PR introduces a clear distinction between verified PoQ, PoSel, public header and encapsulated messsage, and unverified counterparts. Locally-generated proofs and messages, are always assumed to be verified. Messages received over the network are treated as unverified, and we expose a clear API to move them from unverified all the way to decapsulated. They both have the same serialization format, so sending a verified encapsulated message is treated as an unverified one by the receiver, which can call the methods we expose over unverified components: I added some tests to make sure this assumption is never broken.

Doing this exposed a minor but subtle bug in how we collect blending tokens: we were collecting proofs linking the PoQ and PoSel of the **same** layer, while we know that the **outer** PoQ is linked to the **inner** PoSel after decapsulation.

Once a component is verified, it can at any time be treated as an unverified if needed, e.g., in tests.

Because we have multiple sessions, we cannot expose a super clear API from the libp2p swarm, hence we need expose a `publish(EncapsulatedMessage)` method and let the swarm's verifier decide whether this is an old or a new session message. Currently, this is the only limitation of this codebase: we are not able to distinguish between old and new session messages. I will try to make the distinction clear in a follow-up PR, to see how it looks, as I am sure it will avoid us potential new bugs in the future. But for now, this is already a good enough improvement over the status-quo.

The changeset is big, but in reality it's MOSTLY about renamings.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-services/key-management-system-macros/src/lib.rs
+++ b/nomos-services/key-management-system-macros/src/lib.rs
@@ -90,7 +90,6 @@ fn build_operators_enum(
     let operations_impl =
         build_operations_impl_for_enum(&operator_enum_ident, base_enum_ident, variants);
 
-    // let operator_variants = operators_variants.iter().map(|i| i);
     quote! {
         #[derive(Debug)]
         pub enum #operator_enum_ident {
@@ -110,7 +109,7 @@ fn build_operations_impl_for_enum(
         .iter()
         .map(|variant| {
             let variant = &variant.ident;
-            quote! { (Self::#variant(inner), Self::Key::#variant(key)) => inner.execute(key).await }
+            quote! { (Self::#variant(inner), #keys_idents::#variant(key)) => inner.execute(key).await }
         })
         .collect();
 
@@ -119,10 +118,10 @@ fn build_operations_impl_for_enum(
         impl crate::keys::secured_key::SecureKeyOperator for #enum_ident {
             type Key = #keys_idents;
             type Error = crate::keys::errors::KeyError;
-            async fn execute(&mut self, key: &Self::Key) -> Result<(), Self::Error> {
-                match (self, key) {
+            async fn execute(self: Box<Self>, key: &Self::Key) -> Result<(), Self::Error> {
+                match (*self, key) {
                     #(#variants,)*
-                    (operator, key) => Err(KeyError::UnsupportedKeyOperator {
+                    (operator, key) => Err(crate::keys::errors::KeyError::UnsupportedKeyOperator {
                         operator: format!("{operator:?}"),
                         key: format!("{key:?}"),
                     }),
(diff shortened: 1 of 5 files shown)


nomos-blend/message/src/encap/validated.rs
```diff
@@ -18,32 +26,86 @@ pub struct RequiredProofOfSelectionVerificationInputs {
...
-/// An incoming encapsulated Blend message whose public header has been
-/// verified.
-///
-/// It can be decapsulated, but before being sent out as-is, it needs to be
-/// converted into its outgoing variant.
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct IncomingEncapsulatedMessageWithValidatedPublicHeader(EncapsulatedMessage);
+/// A received encapsulated message whose public header has been verified.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
+pub struct EncapsulatedMessageWithVerifiedPublicHeader {
```
youngjoon-lee:
```suggestion
/// A encapsulated message whose public header has been verified.
#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
pub struct EncapsulatedMessageWithVerifiedPublicHeader {
```
I think we can modify the comment, because this struct is now used when we create an encapsulated message locally by `new()`

nomos-services/blend/src/core/processor.rs
```diff
@@ -157,9 +157,23 @@ where
...
+                    // schedule it. At the time of release, the message will be
+                    // ignored since its public header cannot be verified. This is not the most
+                    // efficient way, but it's the less invasive way since by decapsulation we
+                    // currently mean decrypting an encrypted Blend header. No additional checks are
+                    // performed on the nested public header. The spec simply ignores the message,
+                    // and so we do.
+                    let Ok(message_with_validated_public_header) = remaining_encapsulated_message
+                        .clone()
+                        .verify_public_header(self.verifier())
+                    else {
```
youngjoon-lee, line 171: 👍

nomos-blend/message/src/encap/validated.rs
```diff
@@ -87,10 +150,10 @@ impl IncomingEncapsulatedMessageWithValidatedPublicHeader {
             PartDecapsulationOutput::Incompleted {
                 encapsulated_part,
                 public_header,
-                proof_of_selection,
+                verified_proof_of_selection,
             } => {
                 let blending_token =
-                    BlendingToken::new(*public_header.proof_of_quota(), proof_of_selection);
+                    BlendingToken::new(verified_proof_of_quota, verified_proof_of_selection);
```
youngjoon-lee, line 156: Thank you for resolving this bug.

PR #1954, opened 2025-11-24
chore(blend): do not rely on previously-verified messages when propagating them

ntn-x2:

There might be cases where we verify a message coming from an old session, then the STP elapses, and then we send the previously-validated message out. In general, it's not "safe" to perform validation, and then at a later point assume such validation is still valid. Hence, I modified the lib2p swarm implementation to always verify outgoing messages using the latest session info, to avoid potential edge case bugs. This comes at the cost of an additional PoQ verification, but can save us from headaches down the road. I tried a few other approaches, but they all introduced additional complexity, for basically very little benefits. Incoming messages are still verified and bubbled up only if valid, using the appropriate verified type.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/network/src/core/with_core/behaviour/mod.rs
+++ b/nomos-blend/network/src/core/with_core/behaviour/mod.rs
@@ -748,6 +748,9 @@ where
     /// Forwards a message to all healthy connections except the [`except`]
     /// connection.
     ///
+    /// Before the message is forwarded, its public header is validated to
+    /// make sure the receiving peer won't mark us as malicious.
+    ///
     /// If the [`except`] connection is part of the old session, the message is
     /// forwarded to the connections in the old session.
     /// Otherwise, it is forwarded to the connections in the current session.


nomos-blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -791,6 +741,31 @@ where
     pub fn validate_and_publish_message(
         &mut self,
         message: EncapsulatedMessage,
+    ) -> Result<(), Error> {
+        self.validate_and_forward_message_except(message, None)
+    }
+
+    /// Forwards a message to all healthy connections except the [`except`]
+    /// connection.
```
youngjoon-lee, line 749: nit: now we can say that a message is forwarded only if it's valid

youngjoon-lee (approved):
I agree this approach is clean. I've noticed the same thing when I read the code, but I hadn't looked into it properly and just kept postponing it. Thank you for resolving it 🙏

PR #1958, opened 2025-11-26
feat(blend): introduce KMS core create and refer to it from Blend

ntn-x2:
## Problem
We previously defined our own Ed25519 key type in Blend, which means all "efforts" to secure it are duplicated since we do the same in our KMS module. As the linked issue highlighted, we were not zeroizing the key after use, which we already do for KMS keys.

## Solution
Remove our own definition and refer to the KMS key type.

### How the solution was implemented

Since we define key operators inside the KMS module, it needs to depend on blend PoQ crate to define the PoQ operator. Hence, the old `nomos-blend-message` cannot depend on the KMS service module directly: we need to break the cycle.

The cycle is broken in the following way:
* We introduce a new crate in Blend for common crypto operations, called `nomos-blend-crypto`. This is used mostly by `nomos-blend-message` and `nomos-blend-scheduling`.
* We introduce a second create called `nomos-blend-proofs` which only contains PoQ and PoSel definitions. From a quick discussion with @youngjoon-lee, this seems to be a very useful thing to do regardless of the rest of this PR.
* We stripped out crypto utility and proof definition from the `nomos-blend-message` crate, which now only has logic to encapsulate and decapsulate messages
* We take out of `key-management-system` all the key definitions, which are moved inside the new `key-management-system-keys` crate: here we define all key types and all supported operators. This crate now depends on `nomos-blend-proofs`, which means that `nomos-blend-message`, which needs to reference the `Ed25519` secret key type, can now do so without introducing a dependency cycle since `nomos-blend-message` depends on `key-management-system-keys`.
* `key-management-system` has been renamed to `key-management-system-service` and it depends on the new KMS keys module, at the same time re-exporting the keys and operators definitions in there, so that depending on `key-management-system-service` directly is enough to access all keys and operators

Because we have a proliferation of Blend-related crate, I also introduced a new umbrella crate, `nomos-blend`, which is simply a re-export of the other five crates: `nomos-blend-crypto`, `nomos-blend-proofs`, `nomos-blend-message`, `nomos-blend-scheduling`, and `nomos-blend-network`. They are re-exported as `crypto`, `proofs`, `message`, `scheduling` and `network`. So, importing `nomos_blend::crypto` is equivalent to `nomos_blend_crypto`, but it's all under a single namespace for convenience. We used this for instance in `nomos-node` and in the same `nomos-blend-service`.

The rest of the changes are a direct consequence of the decisions taken above. The KMS logic stays the same. The Blend logic stays the same, and we simply reference the KMS keys from there, while still maintaining the other Blend keys that are derived from an Ed25519 key internal to Blend, via an extension trait.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-utils/Cargo.toml
+++ b/nomos-utils/Cargo.toml
@@ -19,7 +19,7 @@ cipher      = { version = "0.4" }
 const-hex   = { default-features = false, version = "1" }
 humantime   = { optional = true, version = "2.1" }
 overwatch   = { workspace = true }
-rand        = { workspace = true }
+rand        = { workspace = true, features = ["getrandom"] }
 serde       = { workspace = true, features = ["derive"], optional = true }
 serde_with  = { optional = true, workspace = true }
 time        = { default-features = false, features = ["serde-human-readable"], optional = true, version = "0.3" }


danielSanchezQ (approved):
Looking good, thanks!

youngjoon-lee (approved):
looks great. Feel free to merge this PR before my PR. I'll rebase mine.

nomos-blend/proofs/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+version     = { workspace = true }
+
+[lints]
+workspace = true
+
+[dependencies]
+generic-array      = { default-features = false, version = "1.2" }
+groth16            = { workspace = true }
+nomos-blend-crypto = { workspace = true }
+nomos-core         = { workspace = true }
```
youngjoon-lee: As long as we have `nomos-core` as a dependency here, the duplication in the PR https://github.com/logos-co/nomos/pull/1940 can't be resolved, because this new crate can't be used by `nomos-core::sdp`. 
Anyway, this topic is not directly related to the issue that this PR tries to solve. So, I think we can merge this PR. The duplication in `nomos-core::sdp` can be discussed separately (FYI, I'm trying to extract `nomos-core::sdp` out of `nomos-core`).

ntn-x2: We only import it to use `ZkHash`. That can go away. If that means removing duplication, that's a good cause. Let me do that.

ntn-x2: I pushed an update: https://github.com/logos-co/nomos/pull/1958/commits/5152f3560a7ec061afdfb647ff78d9c301649018. 

`nomos-blend-proofs` now has only a dev dependency on `nomos_core` for serde stuff, otherwise it only depends on `groth16`, `poq` and `poseidon2` directly. Is this enough?

youngjoon-lee: thank you. it's enough :+1:

PR #1959, opened 2025-11-26
chore(da): Migrate configs to use  deployments and user configs paradigm

pradovic:

This PR migrates da configs to use the pattern presented in: https://github.com/logos-co/nomos/pull/1938.

- The config yamls from both validator and executor are migrated.
- Bunch of seemingly unused config values were erased. 
- For the list of user fields pls consult `config.yaml in nomos-node and nomos-executor`. 
- For the list of deployments fields pls consult: `nodes/nomos-node/node/src/config/da/deployment.rs`.

Yes

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/tests/Cargo.toml
+++ b/tests/Cargo.toml
@@ -35,10 +35,8 @@ nomos-api                     = { workspace = true }
 nomos-blend                   = { workspace = true }
 nomos-blend-service           = { features = ["libp2p"], workspace = true }
 nomos-core                    = { workspace = true }
-nomos-da-dispersal            = { workspace = true }
 nomos-da-network-core         = { workspace = true }
 nomos-da-network-service      = { workspace = true }
-nomos-da-sampling             = { workspace = true }
 nomos-da-verifier             = { workspace = true }
 nomos-executor                = { default-features = false, features = ["testing", "tracing"], workspace = true }
 nomos-http-api-common         = { workspace = true }


nodes/nomos-node/node/src/config/da/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+pub mod sampling;
+pub mod verifier;
+
+/// Parent DA configuration containing all DA services (network, verifier,
+/// sampling, and optionally dispersal for executors)
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Config {
+    pub network: NetworkUserConfig,
+    /// Dispersal config - only present for executor nodes
+    #[serde(skip_serializing_if = "Option::is_none")]
```
ntn-x2: What about defining a different config type inside `nomos-executor` for this? Else it would be possible for a node operator to specify stuff here. If it's not too much effort, I'd rather do that to make things clean and easier to follow.

ntn-x2 (approved):

nodes/nomos-node/node/src/config/da/mod.rs
pradovic: I will try to implement it without introducing a huge change. My logic was that this "might" be okay, as the nomos-node does not actually read this value, only the deployment values are set. The user values are in nomos-executor. 
But I agree it would be even cleaner, and I will try to make it happen!

ntn-x2: Yeah I would argue the main feature that the deployment-bound settings introduce is for us to actually understand better what is going on when spinning up a node. So the config itself should be as straightforward as possible. We have a way to achieve that with Rust, so I think we should utilize it :)

pradovic: Is this better: https://github.com/logos-co/nomos/pull/1959/commits/14e3fddb051f234816f3fa8ed4a270a9ae2123f7?

pradovic: I split both user and deployment executor/validator settings:https://github.com/logos-co/nomos/pull/1959/commits/9ed382d40042bb8aafd84214e407bb1a61f9d7d5. Hope this is better!

pradovic:
https://github.com/logos-blockchain/logos-blockchain/issues/2020.

PR #1967, opened 2025-12-01
feat(ledger): add `Rewards::update_epoch` and PoQ/PoSel verification

youngjoon-lee:
## 1. What does this PR implement?

This PR adds a new `update_epoch` method to the `Rewards` trait.
To verify PoQ and PoSel in `blend::ActivityProof`, we need to track all the epoch changes during the session. Whenever a new epoch is detected, the `update_epoch` stores the epoch info.
When a new session starts, the collected epoch info will be used to initiate PoQ/PoSel verifiers (will be implemented in a next PR).

Additionally, I refactored `blend::Rewards` into sub-structs, so that we can easily distinguish which components are for mutating, and why. Diffs look larger, but most of them are just moving functions into submodules.
- `TargetSessionState`
- `TargetSessionTracker`
- `CurrentSessionState`
- `CurrentSessionTracker` 

## 2. Does the code have enough context to be clearly understood?

Yes

## 3. Who are the specification authors and who is accountable for this PR?

- author: @youngjoon-lee 
- spec: @madxor 
- reviewers
  - blend: @ntn-x2 
  - sdp: @bacv 
  - rewards: @zeegomo  

## 4. Is the specification accurate and complete?
Yes

## 5. Does the implementation introduce changes in the specification?

No

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [x] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [x] 2. Description added.
* [x] 3. Context and links to Specification document(s) added.
* [x] 4. Main contact(s) (developers and specification authors) added
* [x] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 6. Link PR to a specific milestone.

PR #1968, opened 2025-12-01
feat(ledger): PoQ/PoSel verification for Blend rewards

youngjoon-lee:

Finally, the ledger verifies PoQ and PoSel in the `blend:ActivityProof`, as defined in the [spec](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#215261aa09df81b6ab5eea4e715193f1).

We've already implemented the `ProofsVerifiers` trait in the `nomos-blend-message` trait. This PR uses it.

In the previous PR #1967, we implemented collecting all the epoch changes during a session. When a next session starts, we creates a `ProofsVerifier` for each epoch change we've collected. These `ProofsVerifier`s are used to verify `blend::ActivityProofs` submitted for the target session.
Since the ledger cannot know which epoch the activity proof corresponds to, it attempts verification using all the verifiers it created. If any one of them succeeds, the proof is considered valid.

Yes

- author: @youngjoon-lee 
- spec: @madxor 
- reviewers
  - blend: @ntn-x2 
  - sdp: @bacv 
  - rewards: @zeegomo  

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-blend/message/src/reward/session.rs
+++ b/nomos-blend/message/src/reward/session.rs
@@ -186,12 +186,12 @@ mod tests {
 
     #[test]
     fn test_token_evaluation() {
-        let evaluation = BlendingTokenEvaluation::new(1000, 2).unwrap();
-        // token_count_bit_len = ceil(log2((1000*2) + 1)) = 11
+        let evaluation = BlendingTokenEvaluation::new(2000, 2).unwrap();
+        // token_count_bit_len = ceil(log2((2000*2) + 1)) = 12
         // token_count_byte_len = ceil(token_count_bit_len / 8) = 2
         assert_eq!(evaluation.token_count_byte_len, 2);
-        // token_count_bit_len - ceil(log2(2+1)) - 1 = 8
-        assert_eq!(evaluation.activity_threshold, 8.into());
+        // token_count_bit_len - ceil(log2(2+1)) - 1 = 9
+        assert_eq!(evaluation.activity_threshold, 9.into());
 
         let maybe_distance = evaluation.evaluate(
             &BlendingToken::new(
@@ -201,7 +201,7 @@ mod tests {
             ),
             SessionRandomness::from([0; 64]),
         );
-        assert_eq!(maybe_distance, None);
+        assert_eq!(maybe_distance, Some(8.into()));
 
         let maybe_distance = evaluation.evaluate(
             &BlendingToken::new(
@@ -211,6 +211,6 @@ mod tests {
             ),
             SessionRandomness::from([0; 64]),
         );
-        assert_eq!(maybe_distance, Some(7.into()));
+        assert_eq!(maybe_distance, None);
     }
 }
--- a/nomos-blend/message/src/reward/token.rs
+++ b/nomos-blend/message/src/reward/token.rs
@@ -148,7 +148,7 @@ mod tests {
     #[test]
     fn test_blending_token_hamming_distance() {
         let token = blending_token(1, 1, 2);
-        assert_eq!(token.hamming_distance(1, [3u8; 64].into()), 4.into());
+        assert_eq!(token.hamming_distance(1, [3u8; 64].into()), 2.into());
     }
 
     fn blending_token(


ledger/nomos-ledger/src/mantle/sdp/rewards/blend/target_session.rs
```diff
@@ -76,12 +87,25 @@ impl TargetSessionState {
...
         );
-        let &zk_id = self
+        let &(zk_id, index) = self
             .providers
             .get(provider_id)
             .ok_or_else(|| Error::UnknownProvider(Box::new(*provider_id)))?;
 
-        let proof = verify_activity_proof(proof);
+        // Try to verify the proof with each of the available verifiers
+        let proof = self
```
ntn-x2: I suggested it also in the other PR:
```suggestion
        let verified_proof = self
```

ledger/nomos-ledger/src/mantle/sdp/rewards/blend/current_session.rs
```diff
@@ -109,11 +123,56 @@ impl CurrentSessionTracker {
...
-        session_state
+    fn providers_and_zk_root(
+        session_state: &SessionState,
+    ) -> (HashTrieMapSync<ProviderId, (PublicKey, u64)>, ZkHash) {
+        let mut providers = session_state
             .declarations
             .values()
             .map(|declaration| (declaration.provider_id, declaration.zk_id))
+            .collect::<Vec<_>>();
+        providers.sort_by_key(|(_, zk_id)| *zk_id);
```
ntn-x2: This is duplicated logic now. Not really happy with it. Can we add a comment that we will try to tackle by unifying the logic of sorting the list of providers? Not sure where it would be best for it to live.

ledger/nomos-ledger/src/mantle/sdp/mod.rs
```diff
@@ -733,7 +733,7 @@ mod tests {
     fn gc_test_config() -> Config {
         let mut params = HashMap::new();
         params.insert(
-            ServiceType::BlendNetwork,
+            ServiceType::DataAvailability,
```
ntn-x2, line 736: Is this intentional?

nomos-blend/proofs/src/lib.rs
```diff
@@ -3,6 +3,7 @@ pub use poq::CorePathAndSelectors;
 pub(crate) type ZkHash = groth16::Fr;
 pub(crate) type ZkHasher = poseidon2::Poseidon2Bn254Hasher;
 
+pub mod merkle;
```
ntn-x2: I don't think this is the best place for this, since proofs do not really do anything Merkle-related. Can we move it to `nomos-blend-crypto` perhaps?

nomos-blend/message/src/reward/activity.rs
```diff
@@ -31,6 +35,32 @@ impl ActivityProof {
     pub const fn token(&self) -> &BlendingToken {
         &self.token
     }
+
+    pub fn verify_and_build<ProofsVerifier>(
+        proof: &nomos_core::sdp::blend::ActivityProof,
```
ntn-x2, line 40: Why not consuming the input proof?

nomos-blend/message/src/reward/session.rs
```diff
@@ -196,6 +196,7 @@ mod tests {
         let maybe_distance = evaluation.evaluate(
             &BlendingToken::new(
                 VerifiedProofOfQuota::from_bytes_unchecked([0; _]),
+                ed25519_dalek::SigningKey::from_bytes(&[0; _]).verifying_key(),
```
ntn-x2, line 198: I think we can also create a public key directly from bytes. Just a nit.

ledger/nomos-ledger/src/mantle/sdp/mod.rs
youngjoon-lee: Yes, this method is used by `test_garbage_collection_different_services`. Previously, the test checked if a blend declaration is expired after some time. But now, the test becomes complicated since we integrated PoQ verification. So, I modified the test to check a DA declaration, instead. For that, we also need to modify this `gc_test_config` method.
I also updated the code comments in the `test_garbage_collection_different_services` that explains how the test works.

nomos-blend/message/src/reward/activity.rs
youngjoon-lee: I also thought about it, but wanted to avoid `clippy::needless_pass_by_value`.

Also, all the methods in the `Rewards` trait and its child accept arguments by reference. If we design this function to accept arguments by value, we have to clone the object to pass it to this function. So, I think passing a reference here is fine.

ledger/nomos-ledger/src/mantle/sdp/rewards/blend/target_session.rs
youngjoon-lee: resolved in the previous PR

ledger/nomos-ledger/src/mantle/sdp/rewards/blend/current_session.rs
youngjoon-lee: resolved 9d1ad60439182002109c039a6b7e95cbc3f62bbf. I think this is the minimal unification we can do.

nomos-blend/proofs/src/lib.rs
youngjoon-lee: resolved d35f0d180aa35318000fbfa7096771babbf0cdc9. I had to move some tests in the `merkle.rs` to the `nomos-blend-proofs::quota::tests` because they depend on some components in the `nomos-blend-proofs` (to avoid cyclic dep). I think it's reasonable because those tests don't test `MerkleTree` itself, but test if PoQ is verified correctly with `MerkleTree`.

nomos-blend/message/src/reward/session.rs
youngjoon-lee: I prefer creating a privkey first because creating a pubkey directly from bytes is fallible.

nomos-blend/proofs/src/lib.rs
ntn-x2: Makes sense, thanks.

nomos-blend/message/src/reward/activity.rs
ntn-x2: Oh, right. Makes sense then.

nomos-blend/message/src/reward/session.rs
ntn-x2: Sure. But if we use the same input it will never fail. Anyway, this is also ok.

PR #1970, opened 2025-12-02
fix(tests): Balancer refresh subnets quick fix + tests wait for balancer before dispersing

pradovic:
The tests that uses dispersal were failing often with `RetryLimitReached` error. Logs showed Executor constantly returning `InsufficientSubnetworkConnections`.

This PR fixes a race condition in the balancer that caused intermittent `InsufficientSubnetworkConnections` errors during dispersal. The difference between before is that tests don't use mocked initial membership now, but genesis tx, which might come later then initial connections are established.

Root Cause

When a connection is established, the balancer updates subnet stats based on current membership:
```
ConnectionEvent::OpenInbound(peer) => {
    self.connected_peers.insert(peer);
    for subnetwork in self.membership.membership(&peer) {
        self.update_subnetwork_stats(subnetwork, 1, 0);
    }
}
```

The race condition occurs when:
1. Validator has membership and opens outbound connection to executor
2. Executor receives inbound connection **before** it has membership
3. Executor adds peer to `connected_peers`, but `membership.membership(&peer)` returns `[]` → no subnet stats updated
4. Executor later receives membership, but subnet stats are **never refreshed**
5. Executor has valid connections but empty subnet stats → always fails with `InsufficientSubnetworkConnections`

This would also happen on session change, as subnets in balancer would never be refreshed.

### Evidence

Added debug logging confirmed the issue:
```
Validator: BALANCER: OpenOutbound peer=..., subnets=[0, 1]
Executor:  BALANCER: OpenInbound peer=..., subnets=[]
```

Result:
```
validator#0: 2/2 subnets connected, Details: [subnet_0: in=0, out=1, subnet_1: in=0, out=1]
executor#0:  0/2 subnets connected, Details: []

```

Fix:

Refresh subnet attributions in the balancer when membership changes (on subnet refresh signal from da-network), re-evaluating all connected peers against current membership:
```
fn refresh_subnet_attributions(&mut self) {
    self.subnetwork_stats.clear();
    for peer_id in &self.connected_peers {
        for subnetwork_id in self.membership.membership(peer_id) {
            self.update_subnetwork_stats(subnetwork_id, 0, 1);
        }
    }
}
```
Additional:

- Added `wait_da_network_ready()` check in tests to wait for balancer subnets to be populated
- Improved logging in balancer for easier debugging

Future:
- This is a quick fix, we should probably try to provide more robust solution if needed.
- Dispersal and replication behaviors inbound connection needs to be handled on refresh signal. It used to return dummy connection handler, and was not updated when connected peer actually becomes a member. This code is commented with fixme comment to test if this makes the test pass on CI. If it does, we should solve this problem in the separate PR.

Yes

@pradovic 

Yes

NO

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/nomos-da/network/core/src/protocols/dispersal/validator/behaviour.rs
+++ b/nomos-da/network/core/src/protocols/dispersal/validator/behaviour.rs
@@ -192,14 +192,15 @@ impl<M: MembershipHandler<Id = PeerId, NetworkId = SubnetworkId> + 'static> Netw
         role_override: Endpoint,
         port_use: PortUse,
     ) -> Result<THandler<Self>, ConnectionDenied> {
+        // FIXME:
         // Sampling or replication behaviour might open connection to a member peer.
         // During the lifetime of a connection the remote peer might decide to
         // disperse data via existing connection - in such case the connection
         // needs to already have a handler that accepts DA_DISPERSAL_PROTOCOL
         // messages.
-        if !self.membership.is_allowed(&peer) {
-            return Ok(Either::Right(libp2p::swarm::dummy::ConnectionHandler));
-        }
+        // if !self.membership.is_allowed(&peer) {
+        //     return Ok(Either::Right(libp2p::swarm::dummy::ConnectionHandler));
+        // }
         self.stream_behaviour
             .handle_established_outbound_connection(
                 connection_id,
--- a/nomos-da/network/core/src/protocols/sampling/requests/request_behaviour.rs
+++ b/nomos-da/network/core/src/protocols/sampling/requests/request_behaviour.rs
@@ -606,9 +606,10 @@ where
         local_addr: &Multiaddr,
         remote_addr: &Multiaddr,
     ) -> Result<THandler<Self>, ConnectionDenied> {
-        if !self.membership.is_allowed(&peer) {
-            return Ok(Either::Right(libp2p::swarm::dummy::ConnectionHandler));
-        }
+        // FIXME:
+        // if !self.membership.is_allowed(&peer) {
+        //     return Ok(Either::Right(libp2p::swarm::dummy::ConnectionHandler));
+        // }
         self.stream_behaviour
             .handle_established_inbound_connection(connection_id, peer, local_addr, remote_addr)
             .map(Either::Left)
@@ -622,9 +623,10 @@ where
         role_override: Endpoint,
         port_use: PortUse,
     ) -> Result<THandler<Self>, ConnectionDenied> {
-        if !self.membership.is_allowed(&peer) {
-            return Ok(Either::Right(libp2p::swarm::dummy::ConnectionHandler));
-        }
+        // FIXME:
+        // if !self.membership.is_allowed(&peer) {
+        //     return Ok(Either::Right(libp2p::swarm::dummy::ConnectionHandler));
+        // }
         self.connections.register_connect(peer);
         self.try_peer_sample_share(peer);
         self.stream_behaviour


nomos-da/network/core/src/protocols/dispersal/executor/behaviour.rs
```diff
@@ -596,9 +596,12 @@ where
         // decide to disperse data via existing connection - in such case the
         // connection needs to already have a handler that is able to open streams.
         self.connected_peers.insert(peer, connection_id);
-        if !self.membership.is_allowed(&peer) {
-            return Ok(Either::Right(libp2p::swarm::dummy::ConnectionHandler));
-        }
+        // FIXME: We need to update this connection when membership is updated
+        // (on refresh signal). This is commented to see if the integration tests
```
pradovic, line 600: @bacv @danielSanchezQ, there is more context in the PR description.

After adding the refresh signal fix to the balancer, tests started failing in a different place. After inspecting the logs, I believe it's related to the same scenario - but this time because a dummy ConnectionHandler was returned for an incoming connection that arrived before session 0 membership was set. Once this happens, the connection handler is never updated.

I think we need to handle this case in DispersalExecutorBehaviour as well - updating connections on session change so we don't have dummyhandlers forever.  I've commented out this check to isolate further fails so I can inspect the logs. 

I think we should address the proper fix in a separate PR, as it might not be trivial

Additional:
In the case when dummy handler is used these logs appear: "Not connected to peer, initiating dial", which initiates a dial, but dial goes to network swarm port not da swarm, and it does not look it manages to establish a meaningful connection again. 
So the problem seems to be, if I understand correctly :
  1. DA swarm has one inbound connection with dummy handler
  2. libp2p_stream thinks it's not connected, initiates dial
  3. But the dial goes to the network swarm address (wrong swarm entirely)
  4. No new DA swarm connection is made
  
 Same scenario seems to apply to replication protocol (on incoming connection before genesis membership is set)

danielSanchezQ: I would have to check this flow live. But either it should be impossible or we messed up big.
By impossible I mean step 3 `But the dial goes to the network swarm address (wrong swarm entirely)`. what network? At the end connections are handled independently of protocols. Protocols handlers should resolve if they want a *stream* for that connection or not.

pradovic: @danielSanchezQ I am investigating this now, this is as far as I got yesterday. First I am trying to solve it only for tests by commenting code and inspecting further and then I will see what is happening. If you like we can have a call, I can show you what I have so far, that additional comment is a work in progress but the scenario in the PR seems to happen and we need to handle the session update and connections/streams. I need to check how is libp2p initiating dials

danielSanchezQ: I have less context atm. So go for it, this things takes time and it is our job for the next months. So no hurries but steady advance.

pradovic: @bacv, @danielSanchezQ Ok, so here is the new breakdown, the CI is green now.

The dial attempts were actually noise from Kademlia. The dummy connection handler was actually blocking new dials and just returning `DeniedUpgrade`. So the flow is:

  1. Connection established → handle_established_inbound_connection called ONCE
  2. Dummy handler returned with `DeniedUpgrade`
  3. Connection stays open with dummy handler forever
  4. When remote peer tries to open a stream (protocol negotiation), `DeniedUpgrade` rejects it
  5. Membership arriving later has no effect - the handler is immutable once assigned to a connection, there's no libp2p mechanism to replace it after connection establishment

I have commented out dummy connection handlers in all behaviors, and now tests pass on CI as well, even the extremely slow ones. The macOS one is now really slow, compiling for more than 20 minutes, so the test lasts around 50 minutes but passes. I also cannot make them fail on my machine, but my machine usually behaves well (maybe it is fast).

I would merge this PR to unblock CI and then we can see what to do with incoming connection handlers in behaviors in separate PR if needed.

I think we should actually remove the dummy handler altogether (as it is in this PR now), and handle missing membership at the protocol level in behaviors. It might already be handled in all cases, but I can check.

The other solution would be to track dummy connection handlers, but since there's no way to switch the connection handler for an already existing connection, we would need to disconnect peers that have dummy handlers and make them reconnect. We would also need to disconnect peers that fall out of membership on session change. Given this complexity, I think handling the membership check in actual behavior logic is a  better solution.

Let me know what you think. I'm trying to be as clear as possible - I could also explain in a call if that helps.

bacv: We had the same issue in validator behaviour ([disussion](https://github.com/logos-blockchain/nomos/pull/1378/files#r2156370827)), I think we should remove all dummy handlers one by one, because on every session change, the established connection can become usable for dispersal, or sampling, or replication. 

So lets go with this change in executor behaviour as it solves immediate issue with the node startup, then we should revisit other behaviours.

pradovic: I think I have removed it (commented) in all behaviours in order to make it work, as replication had the same issue, same pattern. I have commented all dummy handlers in this PR, and left fixme comments, so I can remove it completely if we agree on the approach.

bacv: You are right, my eyes didn't register those small changes in other behaviours. I think this solves all concerns I had in mind.

danielSanchezQ: What dummy handlers are we using? Supposedly the dummy handlers from libp2p (default ones) should not keep connection or reject anything.

Good to have it green now @pradovic , nice job! But I would like to understand this properly.

nomos-da/network/core/src/protocols/dispersal/validator/behaviour.rs
```diff
@@ -192,14 +192,15 @@ impl<M: MembershipHandler<Id = PeerId, NetworkId = SubnetworkId> + 'static> Netw
...
         // During the lifetime of a connection the remote peer might decide to
         // disperse data via existing connection - in such case the connection
         // needs to already have a handler that accepts DA_DISPERSAL_PROTOCOL
         // messages.
-        if !self.membership.is_allowed(&peer) {
-            return Ok(Either::Right(libp2p::swarm::dummy::ConnectionHandler));
-        }
+        // if !self.membership.is_allowed(&peer) {
+        //     return Ok(Either::Right(libp2p::swarm::dummy::ConnectionHandler));
+        // }
```
danielSanchezQ, line 203: But this means we accept **ANY** connection for this protocol now right?

nomos-da/network/core/src/protocols/replication/behaviour.rs
```diff
@@ -577,10 +577,11 @@ where
...
     ) -> Result<THandler<Self>, ConnectionDenied> {
-        if !self.is_neighbour(&peer_id) {
-            trace!("refusing connection to {peer_id}");
-            return Ok(Either::Right(libp2p::swarm::dummy::ConnectionHandler));
-        }
+        // FIXME: same as in dispersal protocol
+        // if !self.is_neighbour(&peer_id) {
+        //     trace!("refusing connection to {peer_id}");
+        //     return Ok(Either::Right(libp2p::swarm::dummy::ConnectionHandler));
+        // }
```
danielSanchezQ, line 584: How do we filter out connections to protocols we want do not want to have?

nomos-da/network/core/src/protocols/dispersal/executor/behaviour.rs
pradovic: This is the whole `libp2p::swarm::dummy::ConnectionHandler`:
```
pub struct ConnectionHandler;

impl crate::handler::ConnectionHandler for ConnectionHandler {
    type FromBehaviour = Infallible;
    type ToBehaviour = Infallible;
    type InboundProtocol = DeniedUpgrade;
    type OutboundProtocol = DeniedUpgrade;
    type InboundOpenInfo = ();
    type OutboundOpenInfo = ();

    fn listen_protocol(&self) -> SubstreamProtocol<Self::InboundProtocol> {
        SubstreamProtocol::new(DeniedUpgrade, ())
    }

    fn on_behaviour_event(&mut self, event: Self::FromBehaviour) {
        // TODO: remove when Rust 1.82 is MSRV
        #[allow(unreachable_patterns)]
        libp2p_core::util::unreachable(event)
    }

    fn poll(
        &mut self,
        _: &mut Context<'_>,
    ) -> Poll<ConnectionHandlerEvent<Self::OutboundProtocol, (), Self::ToBehaviour>> {
        Poll::Pending
    }

    fn on_connection_event(
        &mut self,
        event: ConnectionEvent<Self::InboundProtocol, Self::OutboundProtocol>,
    ) {
        match event {
            // TODO: remove when Rust 1.82 is MSRV
            #[allow(unreachable_patterns)]
            ConnectionEvent::FullyNegotiatedInbound(FullyNegotiatedInbound {
                protocol, ..
            }) => libp2p_core::util::unreachable(protocol),
            // TODO: remove when Rust 1.82 is MSRV
            #[allow(unreachable_patterns)]
            ConnectionEvent::FullyNegotiatedOutbound(FullyNegotiatedOutbound {
                protocol, ..
            }) => libp2p_core::util::unreachable(protocol),
            // TODO: remove when Rust 1.82 is MSRV
            #[allow(unreachable_patterns)]
            ConnectionEvent::DialUpgradeError(DialUpgradeError { info: _, error }) => match error {
                // TODO: remove when Rust 1.82 is MSRV
                #[allow(unreachable_patterns)]
                StreamUpgradeError::Timeout => unreachable!(),
                StreamUpgradeError::Apply(e) => libp2p_core::util::unreachable(e),
                StreamUpgradeError::NegotiationFailed | StreamUpgradeError::Io(_) => {
                    unreachable!("Denied upgrade does not support any protocols")
                }
            },
            // TODO: remove when Rust 1.82 is MSRV
            #[allow(unreachable_patterns)]
            ConnectionEvent::AddressChange(_)
            | ConnectionEvent::ListenUpgradeError(_)
            | ConnectionEvent::LocalProtocolsChange(_)
            | ConnectionEvent::RemoteProtocolsChange(_) => {}
        }
    }
}
```
```
fn listen_protocol(&self) -> SubstreamProtocol<Self::InboundProtocol> {
      SubstreamProtocol::new(DeniedUpgrade, ())
  }
```

It uses `DeniedUpgrade` for both inbound and outbound protocols. So it doesn't close the connection, but it rejects all protocol upgrades - meaning no streams can be opened on that connection for the protocol that returned the dummy handler. Also, this is never updated on session change. This reflects tests as some connections can be established before the genesis session is initialized.

  The problem is, if I understand correctly:
  - The swarm-level connection stays open (so is_connected() returns true)
  - But libp2p_stream never registers the connection in its senders map (because we returned dummy handler instead of calling stream_behaviour.handle_established_inbound_connection())
  - So when we try to open a dispersal/replication stream, libp2p_stream thinks there's no connection and the protocol negotiation fails with `UnsupportedProtocol`

danielSanchezQ: Ok, makes sense @pradovic . I think in this case we may need a custom handler that awaits for a protocol upgrade indefinetely until the connection is closed. So if the upgrade is done we can use the protocol. But lets have a chat for this later on. For now...nice analysis on this and thanks for bringing it 🟢 again!

pradovic: Thanks, I agree! Either this or handle it on a protocol level (returning errors if node is not in membership).

PR #1973, opened 2025-12-03
feat(blend): submit activity proof to SDP service

youngjoon-lee:

Closes https://github.com/logos-blockchain/nomos/issues/1930

Finally, the blend core service submits activity proofs to the SDP service.

Yes

dev: @youngjoon-lee @ntn-x2 
spec: @madxor 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.


danielSanchezQ (approved):
🚢

ntn-x2 (commented):
I understand we might want to have a crate-local trait for unit testing activity proof submission, but why adding one more generic to the Blend service? I think it's unnecessary since we depend on SDP anyway, and as far as I see we are not mocking anything anywyay. So I would suggest we remove the service-level generic, and if we need to, we can simply implement the trait we need in the tests to test the service `run` logic. Do you agree that would make sense? Other than that, the logic looks good! If we want to have an adapter for better ergonomics, that's also fine, but I think it should not bubble up at the service level and be provided by the binary.

Issue #1961, opened 2025-11-27
Keystream reuse across private-header items (and payload) enables XOR leakage


megonen:
In here: https://github.com/logos-co/nomos/blob/e09a442c6fa87414b10657dbf7cca096bf4f4dbc/nomos-blend/message/src/crypto/mod.rs#L35 "SharedKey::encrypt" seeds BlakeRng only from the shared key and then XORs that keystream with the input. Because we call it multiple times per encapsulation with the same SharedKey and no nonce/context/counter, the same keystream is reused for:

- every blending header within the private header for that layer, and

- the payload for that layer as well.

This is a classic two-time pad problem: for any two equal-length plaintext chunks encrypted with the same keystream, C1 ⊕ C2 = P1 ⊕ P2, leaking structure and enabling offline recovery with partial knowledge/guesses.

PR #1975, opened 2025-12-04
fix(blend): fix XOR leakage by introducing `Cipher`

youngjoon-lee:

Resolves https://github.com/logos-blockchain/nomos/issues/1961

Previously, when encrypting [blending header](https://www.notion.so/nomos-tech/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#215261aa09df812da31de7731b566203)s in a [private header](https://www.notion.so/nomos-tech/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#215261aa09df81d28398e0426d6af865), we XOR-ed them with the pseudo-random bytes generated from the same seed (the same shared key), as described in the [spec](https://www.notion.so/nomos-tech/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#215261aa09df812882dce60652a8c73e): 
image

However, this is a security vulnerability, as explained in #1961.
For example, if someone receives an encapsulated message with 2 layers (i.e. `encrypted_header1` and `double_encrypted_header2`), and if he knows `plain_header1` "somehow" (while not knowing the key for decrypting `encrypted_header1`), he can easily derive `encrypted_header2` by `XOR(encrypted_header1, double_encrypted_header2)`. In other words, if he is the one who has the key for `header2`, he can bypass the 1st decapsulation that should be done by any other node. This is only one example. 

To fix this vulnerability, we should encrypt each blending header using a distinct pseudo-random bytes generated from a different seed. For that, this PR introduces a new struct `Cipher` which is initialized from a `SharedKey`. Using this repeatedly, the underlying PRNG is advanced internally, so we can encrypt each data using a distinct pseudo-random bytes.
We use `Cipher` for encrypting multiple blending headers in a private header using a `SharedKey`. We can decrypt the blending headers in the same order by using the same `Cipher` which can be initialized from the same `SharedKey`, but without having the security vulnerability. I wrote a test for this, but the following pseudo-code shows how it works.
```rust
let cipher = shared_key.cipher();
// `header1` and `encrypted_header2` are encrypted using different pseudo-random bytes.
let encrypted_header1 = cipher.encrypt(header1);
let double_encrypted_header2 = cipher.encrypt(encrypted_header2);

// Using the same cipher, we can decrypt them in the same order.
let cipher = shared_key.cipher();
assert_eq!(cipher.decrypt(encrypted_header1), header1);
assert_eq!(cipher.decrypt(double_encrypted_header2), encrypted_header2);
``` 

As well as blending headers, we also encrypt a [payload](https://www.notion.so/nomos-tech/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#215261aa09df81e59306e2dfbaa88647) in a [message](https://www.notion.so/nomos-tech/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#215261aa09df819a9e60d8839e97a160). Unlike blending headers, we have only one payload per message. So, when encrypting a payload using a `SharedKey` (i.e. adding an encapsulation), we use the 1st pseudo-random bytes generated from a `Cipher`.
This is actually the same behavior as before, because I think this doesn't have the security vulnerability, but please correct me if I'm wrong.
```rust
let cipher = shared_key.cipher();
let encrypted_payload = cipher(payload);
``` 

One thing I want to highlight is that the [Message Initialization](https://www.notion.so/nomos-tech/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#215261aa09df811aa232c1097eafd867) process becomes a bit more complicated now. This process is for generating random dummy blending headers in a "reconstructable" way for the correct signature verification. But, now that we use the "advancing" PRNG (i.e. `Cipher`), this process must be updated to use the correct PRNG seeds. I updated the doc comment of `EncapsulatedPrivateHeader::initialize` to explain the updated logic. But, I think we also need to update the spec to explain this clearly.

I hope so.

author: @youngjoon-lee @ntn-x2 
spec: @madxor
security advisor: @megonen  

Needs to be updated

The following sections should be updated, according to this PR's changes:
- [`CSPRBG()`](https://www.notion.so/nomos-tech/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#232261aa09df80939e32dbe515d9a473) function should accept a new argument `sequence` starting from 0. This is the behavior of `Cipher` introduced in this PR.
- [`Ek(x)`](https://www.notion.so/nomos-tech/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#215261aa09df81fbad9ee6cec3ef6e19) and [`Dk(x)`](https://www.notion.so/nomos-tech/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#215261aa09df814780bedb40c83a7e01) also should accept a new argument `sequence` and pass it to `CSPRBG()`.
- [Message Encapsulation](https://www.notion.so/nomos-tech/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#215261aa09df81e3920adedf09f0805a) should be updated to use `sequence`s when encrypting blending headers using a shared key (in the [step 6](https://www.notion.so/nomos-tech/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#215261aa09df812882dce60652a8c73e)).
- [Message Decapsulation](https://www.notion.so/nomos-tech/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#215261aa09df8159b1dcfdcda8043b92) should be updated to use `sequence`s when decrypting blending headers using a shared key (in the [step 2](https://www.notion.so/nomos-tech/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#215261aa09df81e1b528e61b00597b5f)).
- [Message Initialization](https://www.notion.so/nomos-tech/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#215261aa09df811aa232c1097eafd867) should be updated to use the correct `sequence`s when encrypting the last `h` blending headers in a reconstructable manner (in the [step 4](https://www.notion.so/nomos-tech/Message-Encapsulation-Mechanism-215261aa09df81309d7fd7f1c2da086b?source=copy_link#215261aa09df815ebb1ffe89a71c7b11)), as described in the doc comment of `EncapsulatedPrivateHeader::initialize` in this PR.

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,7 @@ members = [
   "tests",
   "utxotree",
   "wallet",
+  "wallet-http-client",
   "zk/circuits/prover",
   "zk/circuits/utils",
   "zk/circuits/verifier",
@@ -142,6 +143,7 @@ tests                         = { default-features = false, path = "./tests" }
 tx-service                    = { default-features = false, path = "./nomos-services/tx-service" }
 utxotree                      = { default-features = false, path = "./utxotree" }
 wallet                        = { default-features = false, path = "./wallet" }
+wallet-http-client            = { default-features = false, path = "./wallet-http-client" }
 witness-generator             = { default-features = false, path = "zk/circuits/witness-generator" }
 zksign                        = { default-features = false, path = "./zk/proofs/zksign" }
 
@@ -150,6 +152,7 @@ blake2            = { default-features = false, version = "0.10" }
 bytes             = { default-features = false, version = "1.3" }
 cached            = { default-features = false, version = "0.55.1" }
 fork_stream       = { default-features = false, version = "0.1.0" }
+hex               = { default-features = false, version = "0.4" }
 libp2p            = { default-features = false, version = "0.55" }
 libp2p-stream     = { default-features = false, version = "0.3.0-alpha" }
 overwatch         = { default-features = false, git = "https://github.com/logos-co/Overwatch", rev = "f5a9902" }
--- a/nodes/api-common/Cargo.toml
+++ b/nodes/api-common/Cargo.toml
@@ -22,10 +22,12 @@ pprof = { default-features = false, features = [
   "protobuf-codec",
 ], optional = true, version = "0.15" }
 serde = { workspace = true, features = ["alloc", "derive"] }
+serde_json = { workspace = true }
 serde_with = { default-features = false, features = ["macros"], workspace = true }
 tokio = { default-features = false, features = ["time"], optional = true, version = "1" }
 tower_governor = { default-features = false, features = ["axum"], version = "0.3" }
 tracing = { default-features = false, optional = true, workspace = true }
+zksign = { workspace = true }
 
 [features]
 profiling = ["pprof", "tokio", "tracing"]
--- a/nodes/api-common/src/bodies/dispersal.rs
+++ b/nodes/api-common/src/bodies/dispersal.rs
@@ -2,7 +2,7 @@ use nomos_core::mantle::ops::channel::{ChannelId, Ed25519PublicKey, MsgId};
 use serde::{Deserialize, Serialize};
 
 #[derive(Serialize, Deserialize)]
-pub struct DispersalRequest {
+pub struct DispersalRequestBody {
     pub channel_id: ChannelId,
     pub parent_msg_id: MsgId,
     pub signer: Ed25519PublicKey,
--- a/nodes/api-common/src/bodies/mod.rs
+++ b/nodes/api-common/src/bodies/mod.rs
@@ -1,0 +1,12 @@
+use serde::{Deserialize, Serialize};
+
+pub mod dispersal;
+pub mod wallet;
+
+/// A no-operation body for endpoints that do not require a request or response
+/// body.
+///
+/// This is a workaround used to satisfy the type system in niche cases that
+/// should be addressed in the future. E.g.: A `get` requiring a request body.
+#[derive(Serialize, Deserialize)]
+pub struct NoopBody;
--- a/nodes/api-common/src/bodies/wallet.rs
+++ b/nodes/api-common/src/bodies/wallet.rs
@@ -1,0 +1,63 @@
+pub mod balance {
+    use axum::{
+        http::StatusCode,
+        response::{IntoResponse, Response},
+    };
+    use nomos_core::{header::HeaderId, mantle::Value};
+    use serde::{Deserialize, Serialize};
+    use zksign::PublicKey;
+
+    #[derive(Serialize, Deserialize)]
+    pub struct WalletBalanceResponseBody {
+        pub tip: HeaderId,
+        pub balance: Value,
+        pub address: PublicKey,
+    }
+
+    impl IntoResponse for WalletBalanceResponseBody {
+        fn into_response(self) -> Response {
+            (StatusCode::OK, serde_json::to_string(&self).unwrap()).into_response()
+        }
+    }
+}
+
+pub mod transfer_funds {
+    use axum::{
+        http::StatusCode,
+        response::{IntoResponse, Response},
+    };
+    use nomos_core::{
+        header::HeaderId,
+        mantle::{SignedMantleTx, Transaction as _, Value},
+    };
+    use serde::{Deserialize, Serialize};
+    use zksign::PublicKey;
+
+    #[derive(Serialize, Deserialize)]
+    pub struct WalletTransferFundsRequestBody {
+        pub tip: Option<HeaderId>,
+        pub change_public_key: PublicKey,
+        pub funding_public_keys: Vec<PublicKey>,
+        pub recipient_public_key: PublicKey,
+        pub amount: Value,
+    }
+
+    #[derive(Serialize, Deserialize)]
+    pub struct WalletTransferFundsResponseBody {
+        pub hash: nomos_core::mantle::tx::TxHash,
+    }
+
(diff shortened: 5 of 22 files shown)


nomos-blend/message/src/encap/encapsulated.rs
```diff
@@ -141,7 +142,7 @@ impl EncapsulatedPart {
         );
 
         // Encrypt the payload.
-        let payload = self.payload.encapsulate(shared_key);
+        let payload = self.payload.encapsulate(&mut shared_key.cipher());
```
ntn-x2: I know this was not in the scope of this PR, but I think it would be clearer.
```suggestion
        let encapsulated_payload = self.payload.encapsulate(&mut shared_key.cipher());
```

nomos-blend/message/src/encap/encapsulated.rs
```diff
@@ -168,7 +169,7 @@ impl EncapsulatedPart {
                 public_header,
                 verified_proof_of_selection,
             } => {
-                let payload = self.payload.decapsulate(key);
+                let payload = self.payload.decapsulate(&mut key.cipher());
```
ntn-x2:
```suggestion
                let decapsulated_payload = self.payload.decapsulate(&mut key.cipher());
```

nomos-blend/message/src/encap/encapsulated.rs
```diff
@@ -189,7 +190,7 @@ impl EncapsulatedPart {
...
```
ntn-x2:
```suggestion
                let decapsulated_payload = self.payload.decapsulate(&mut key.cipher());
```

ntn-x2 (approved):
It looks good to me

nomos-blend/message/src/encap/encapsulated.rs
```diff
@@ -271,26 +272,38 @@ impl EncapsulatedPrivateHeader {
...
                 .rev()
-                .map(|rng_key| {
+                .map(|(input_index, rng_key)| {
                     let mut header = EncapsulatedBlendingHeader::initialize(
                         &BlendingHeader::pseudo_random(rng_key.as_slice()),
                     );
                     // Encapsulate the blending header with the shared key of each input
                     // until the shared key equal to the `rng_key` is encountered
                     // (inclusive).
+                    let mut blocks_to_advance = inputs.len() - input_index;
```
madxor, line 295: Can we start initialize the `blocks_to_advance` from the `2*ENCAPSUALTION_COUNT`? This way we can have a completely separated sequence that will not overlap with the headers sequence.

youngjoon-lee: If I understood the rationale of this Message Initialization step correctly, I think, this step becomes reconstructable only if the sequence overlaps with the headers sequence. 

Step 0: Initialization:
```
D[1] = Enc(K[1], 2, Enc(K[0], 1, RND(K[1], 0)))
D[0] =              Enc(K[0], 2, RND(K[0], 0))
```

Step 1: Encapsulation with `K[0]` after shifting right:
```
H[0] =                                        Enc(K[0], 0, H[0]) 
D[1] = Enc(K[0], 1, Enc(K[1], 2, Enc(K[0], 1, RND(K[1], 0))))
           --------                  -------
          canceled out              canceled out

(discarded)
D[0]
```

Step 2: Encapsulation with `K[1]` after shifting right:
```
H[1] =              Enc(K[1], 0, H[1])
H[0] = Enc(K[1], 1, Enc(K[0], 0, H[0]))

(discarded)
D[0]
D[1]
```

Step 3: Decapsulation with `K[1]` and shift left:
```
(fully decapsulated)
H[1]

H[0] = Dec(K[1], 1, Enc(K[1], 1, Enc(K[0], 0, H[0])))
           -------      ------- 
        canceled out  canceled out
D[1] = Enc(K[1], 2, RND(K[1], 0))
```
Now, `D[1]` has been successfully reconstructed by the holder of `K[1]`, and it matches the `D[1]` that was produced in Step 1 by the holder of `K[0]`. So, the signature can be verified successfully.

If we start `blocks_to_advance` from `2 * L` in the Initialization step, the decrypting node is supposed to **jump** the sequence at some point while decrypting the blending headers, but it has no way of knowing **when** to perform this jump. In the example above where `L=2`, the node can simply jump the sequence when decrypting the last blending header. However, for `L=3` (or larger), the node can't decide whether to jump the sequence at the last header or at the second-to-last one, for instance. This is because the decrypting node has no way of knowing whether it's the 1st, 2nd, or 3rd blend node.

madxor: You are right that a node could not know which blending header it is decrypting. My point was to use the sequence number outside of the range that is used for the header encryption, so that there would be no overlap. That shifted sequence number would only be use for the reconstruction (and must be fixed to a single value as you correctly noted). However, with domain separation this doesn't matter as the seed (and keystream) is going to be different for headers, payload and initialization. So this discussion can be resolved. Thank you!

PR #1985, opened 2025-12-08
feat(da): da procotols remove dummy connection handler

pradovic:

Context:
https://github.com/logos-blockchain/logos-blockchain/pull/1970 - this was a quick fix for following scenario.

Da behaviors would add dummy connection handlers based on membership on inbound/outbound connection handlers. The problem with this approach is that membership is dynamic, and does not map 1-1 to connections. The quick fix was to propagate session change to balancer as well so it would properly update subnetwork stats for already connected peers that become part of the membership. This prompted problems in da behaviors because they would already assign dummy connection handlers to these connections, which would make impossible opening the stream to the updated membership peers.  The quick fix was to comment the checks in da behaviors.

This PR removes FIXME comments for dummy connection handlers in da protocol. It adds per message or per stream filtering instead in the places where it was needed or just cleans up where everything was already handled by the protocol logic. 

Please look at the added comments as they should describe the choice of adding or removing a check.

Yes

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


nomos-da/network/core/src/protocols/replication/behaviour.rs
```diff
@@ -411,6 +415,28 @@ where
     {
         match self.incoming_tasks.poll_next_unpin(cx) {
             Poll::Ready(Some(Ok((peer_id, message, read_half)))) => {
+                // For Share messages, only accept from neighbours (peers in the same subnetwork).
+                // TX messages are accepted from any connected peer to allow broad propagation.
+                // Receiving Share from non-neighbour indicates sender misbehavior.
```
pradovic: @danielSanchezQ , I have talked with @bacv  and replication was propagating tx to ALL connected peers to accelerate the propagation. Because of this, I have added the explicit message level check that DOES NOT allow share request from non-members (drops the stream), but ALLOWS tx streams from all connected peers.

danielSanchezQ: Well, I guess it makes sense as this is only DA network connections, its an early optimization, thanks.

danielSanchezQ (approved):
Thanks, lets watch CI for 🟢

PR #1987, opened 2025-12-10
chore(kms): Zk key migration into KMS crate

ntn-x2:

Removed the ZK key definition from the `zksign` crate, which now contains raw types only. The key definition beneath the `ZkKey` variant of the KMS key is now moved inside the KMS crate itself, so that we can avoid exposing methods to access its private components. Some places still require access to the underlying private key, so we followed the same approach used for the Ed25519 key: we expose an unsecured type, that can be converted into the secured type, if needed, but only after enabling KMS's `unsafe` feature. So now all crates depend on `key-management-system-keys`, which in turn is the only crate that depends on `zksign`. Follow-up PRs will look for more key definition duplication and will eliminate those as well, if any.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-kms/keys/src/keys/zk/mod.rs
+++ b/nomos-kms/keys/src/keys/zk/mod.rs
@@ -64,12 +64,13 @@ impl ZkKey {
 
 impl Debug for ZkKey {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        let private_key = if cfg!(feature = "unsafe") {
-            format!("{:?}", self.0)
-        } else {
-            "<redacted>".to_owned()
-        };
-        write!(f, "ZkKey({private_key})")
+        #[cfg(feature = "unsafe")]
+        write!(f, "ZkKey({:?})", self.0)?;
+
+        #[cfg(not(feature = "unsafe"))]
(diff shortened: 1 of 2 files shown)


youngjoon-lee (approved):
looks good

nomos-kms/keys/src/keys/zk/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+impl Debug for ZkKey {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        let private_key = if cfg!(feature = "unsafe") {
+            format!("{:?}", self.0)
+        } else {
+            "<redacted>".to_owned()
+        };
```
danielSanchezQ: I know is not from this PR, but this should be featured gated instead? or does the compiler optimize the branch already with the `if` statement?

ntn-x2: Good point, I changed it in https://github.com/logos-blockchain/logos-blockchain/pull/1987/commits/f7707f585eb042f36eaec4f4932a1bf121de92c4.

PR #1989, opened 2025-12-10
feat(blend): save/load unsubmitted blending tokens

### LINKED ISSUE #1848: Store and retrieve unclaimed blending tokens upon node startup
Storing blending tokens that have been collected so far (for current & previous sessions). They can be restored once the node is restarted. Of course, we should handle the case in which the session has passed before restart.

youngjoon-lee:

Closes #1848

### Key Changes

Now, the blend core service saves/loads blending tokens that are not yet used for computing the activity proof. 

First, the service saves `SessionBlendingTokenCollector` (for the **current session**), so that it can continue collecting tokens if it's restarted within the same session.

Additionally, the service saves `Option<OldSessionBlendingTokenCollector>` as well for the **old session**. It's saved as `Some` during STP and updated to `None` after STP.
You may be curious why we save it unlike how we don't save unsent messages for the old session (https://github.com/logos-blockchain/logos-blockchain/pull/1986). If the node shut down during the STP (`s` -> `s+1`), and if the node is restarted before the session `s+1` is over, it can immediately compute/submit the activity proof using tokens that have been collected so far for the session `s`. This is the best effort that the service can do to not waste tokens collected so far.

### Notes

Unlike unsent messages, the `ServiceState` stores token collectors directly instead of storing  `Vec<BlendingTokens>`, so that it can load the collectors and use it immediately. If we store `Vec`s, we need to prepare ingredients for recreating collectors when the service loads the state. Recreating the current session token collector is simple because the service fetches the current membership. But, the service doesn't fetch the old session membership, which is necessary for recreating the old session token collector. So, I think it's better to store the collectors directly.

As part of this change, I also introduced `SessionMismatchError` to explicitly handle cases where the provided `TokenCollector.session_number` does not match `ServiceState::last_seen_session`.

Lastly, I slightly adjusted the behavior of `ServiceState::new`. The instance is now saved **immediately** via the `state_updater` upon creation. Previously, we **delayed** saving the state until a new message was scheduled. However, since `ServiceState` now owns the token collectors which contain meaningful values from the beginning, saving the state eagerly at creation time is a cleaner behavior, in my opinion.

I believe so.

dev: @youngjoon-lee @ntn-x2 
spec: @madxor 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nomos-services/blend/src/core/mod.rs
+++ b/nomos-services/blend/src/core/mod.rs
@@ -25,7 +25,7 @@ use nomos_blend::{
             validated::EncapsulatedMessageWithVerifiedPublicHeader,
         },
         reward::{
-            self, ActivityProof, BlendingTokenCollector, OldSessionBlendingTokenCollector,
+            self, ActivityProof, BlendingToken, OldSessionBlendingTokenCollector,
             SessionBlendingTokenCollector,
         },
     },
@@ -367,7 +367,6 @@ where
             mut remaining_clock_stream,
             current_public_info,
             crypto_processor,
-            blending_token_collector,
             current_recovery_checkpoint,
             message_scheduler,
             mut backend,
@@ -428,7 +427,6 @@ where
             &mut epoch_handler,
             message_scheduler.into(),
             &mut rng,
-            blending_token_collector,
             crypto_processor,
             current_public_info,
             current_recovery_checkpoint,
@@ -504,7 +502,6 @@ async fn initialize<
         ProofsGenerator,
         ProofsVerifier,
     >,
-    SessionBlendingTokenCollector,
     ServiceState<Backend::Settings, NetAdapter::BroadcastSettings>,
     SchedulerWrapper<
         BlakeRng,
@@ -702,9 +699,6 @@ where
         remaining_clock_stream,
         current_public_info,
         crypto_processor,
-        current_recovery_checkpoint
-            .current_session_token_collector()
-            .clone(),
         current_recovery_checkpoint,
         message_scheduler,
         backend,
@@ -765,7 +759,6 @@ async fn run_event_loop<
         EncapsulatedMessageWithVerifiedPublicHeader,
     >,
     rng: &mut Rng,
-    mut blending_token_collector: SessionBlendingTokenCollector,
 
     mut crypto_processor: CoreCryptographicProcessor<
         NodeId,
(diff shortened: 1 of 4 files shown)


nomos-services/blend/src/core/state.rs
```diff
@@ -21,23 +27,31 @@ mod serde {
         ))]
         unsent_processed_messages: HashSet<ProcessedMessage<BroadcastSettings>>,
         unsent_data_messages: HashSet<EncapsulatedMessageWithVerifiedPublicHeader>,
+        token_collector: SessionBlendingTokenCollector,
```
ntn-x2: Just a bit clearer?

```suggestion
        current_session_token_collector: SessionBlendingTokenCollector,
```

nomos-services/blend/src/core/state.rs
```diff
@@ -163,6 +227,62 @@ mod service {
...
+        pub(super) fn update_old_session_token_collector(
+            &mut self,
+            collector: OldSessionBlendingTokenCollector,
+        ) -> Result<(), SessionMismatchError> {
+            Self::validate_old_session_token_collector(self.last_seen_session, &collector)?;
+            self.old_session_token_collector = Some(collector);
+            Ok(())
+        }
+
+        const fn validate_old_session_token_collector(
```
ntn-x2: Do we need to separate functions, or can we simply use a smaller session number for the previous session collector?

nomos-services/blend/src/core/mod.rs
```diff
@@ -649,9 +639,32 @@ where
...
+                ).expect("Reward session info must be created successfully. Panicking since the service cannot continue with this session")
+            ),
+            None,
+            state_updater,
+        ).expect("service state should be created successfully")
     };
 
+    // If there is the old session token collector loaded from `last_saved_state`,
+    // compute/submit its activity proof because we won't collect more tokens for
+    // the old session after this initialization step.
```
ntn-x2: Can we add to the comment that we won't collect because we are not establishing connections with peers from the old session?

nomos-services/blend/src/core/mod.rs
```diff
@@ -688,7 +701,7 @@ where
         remaining_clock_stream,
         current_public_info,
         crypto_processor,
-        blending_token_collector,
+        current_recovery_checkpoint.token_collector().clone(),
```
ntn-x2: Do we need to clone here? Can't we, down below, use references to the collector from the recovery checkpoint?

nomos-services/blend/src/core/mod.rs
```diff
@@ -1054,37 +1064,43 @@ where
...
             let (new_scheduler, old_scheduler) = current_scheduler
                 .rotate_session(new_scheduler_session_info, settings.scheduler_settings());
             HandleSessionEventOutput::Transitioning {
                 new_crypto_processor: new_processor,
                 old_crypto_processor: current_cryptographic_processor,
                 new_scheduler,
                 old_scheduler,
-                new_token_collector: new_session_blending_token_collector,
-                old_token_collector: old_session_blending_token_collector,
+                new_token_collector: new_session_blending_token_collector.clone(),
```
ntn-x2: Also here, why do we need to return the collectors on their own and also as part of the recovery state?

nomos-services/blend/src/core/state.rs
youngjoon-lee: it's because the type of old session token collector (`OldSessionTokenCollector`) is different from the one for the current session (`SessionTokenCollector`).

youngjoon-lee: Let me share one more context. I initially defined `validate_token_collector_session_number(session_number, last_seen_session)`, but I felt this signature is not so readable. Also, this function should be used in two places: `ServiceState::new` and `ServiceState::update_old_session_token_collector`. I didn't want to duplicate the `+1` logic (or `-1`) in two places, because it's easy to forget when we may change logic in the future. So, I decided to have `validate_old_session_token_collector` which performs `+1` inside.

ntn-x2: Fair enough, thanks for the explanation.

nomos-services/blend/src/core/mod.rs
youngjoon-lee: Resolved c00a9a14b2d30650e6c53916f44fcef706d1b90a

nomos-services/blend/src/core/mod.rs
youngjoon-lee: Resolved 6a8eca68e29c8b8fa3010e24e6a0c66e7aee2f8c. Now, token collectors live only in the `ServiceState`.
`ServiceState` doesn't expose collectors as reference (or mut). Instead, `StateUpdater` exposes methods to insert tokens to the collectors, ensuring that all updates are properly committed in the end.

Now the implementation is much cleaner, I think. Thank you.

nomos-services/blend/src/core/mod.rs
youngjoon-lee: Resolved in https://github.com/logos-blockchain/logos-blockchain/pull/1989#discussion_r2612789866

nomos-services/blend/src/core/state.rs
youngjoon-lee: Resolved f0dd563a4084ee5b7a3cfae4b61702cdda63b702, but it'd be more worth reviewing https://github.com/logos-blockchain/logos-blockchain/pull/1989#discussion_r2612789866 than this commit.

nomos-services/blend/src/core/mod.rs
ntn-x2: Yes, this was also an option, and it indeed looks much cleaner. Thanks!

PR #1996, opened 2025-12-12
chore(PoL): Use blake2b256 instead of blake2b512 for chain of secrets

thomaslavaur:
Informal PR to Daniel's branch

### DIFF
--- a/nomos-blend/scheduling/src/message_scheduler/tests.rs
+++ b/nomos-blend/scheduling/src/message_scheduler/tests.rs
@@ -5,7 +5,7 @@ use core::{
 use std::collections::HashSet;
 
 use futures::{StreamExt as _, task::noop_waker_ref};
-use nomos_utils::blake_rng::BlakeRng;
+use nomos_utils::blake_rng::BlakeRng512;
 use rand::SeedableRng as _;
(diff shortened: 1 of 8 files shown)


nomos-utils/src/blake256_rng.rs
```diff
@@ -0,0 +1,259 @@
+use std::iter::{Enumerate, FlatMap, Repeat};
+
+use blake2::{Blake2b, Digest as _};
+use blake2::digest::consts::U32;
+pub use cipher::StreamCipher;
+use cipher::{BlockSizeUser, StreamCipherError, inout::InOutBuf};
+use rand::Error;
+pub use rand::{CryptoRng, RngCore, SeedableRng};
+
+pub type Blake2b256 = Blake2b<U32>;
```
danielSanchezQ: Uhm, this is not really the way to go 🤔 , we should have made the original blakrng generic on output size, then export a couple of typealiases instead of having duplicated implementations. I'll update this branch myself with it if you dont mind.

thomaslavaur: yep works for me

ntn-x2 (approved):

PR #1997, opened 2025-12-12
feat(ci): Cache-aware Cargo hack check

strinnityk:
Implement a cache-aware cargo hack check script with the objective of reducing cargo hack runtime.

This PR contains:
- A Bash script that hashes a crate directory (excluding `.git` and `target`).
- A Python script that:
  - Retrieves all workspace members via `cargo metadata`.
  - Sorts them topologically (ignoring dev-only dependencies).
  - Retrieves the commands cargo hack would run.
  - Reorders those commands based on the workspace dependency graph.
  - Runs each cargo hack command with per-crate cache awareness and dependent invalidation.

Also important to note, I'd like to address the listed TODOs before merging:
- Parametrize the python script.
- Update the cargo hack invocation to match the current CLI.

Other than that, this is still a v1 but feel free to go hard at it.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/.github/actions/cargo-hack-check-cached/action.yml
+++ b/.github/actions/cargo-hack-check-cached/action.yml
@@ -25,6 +25,7 @@ runs:
           cargo-hack-check/${{ runner.os }}/main
           cargo-hack-check/${{ runner.os }}
 
+    # TODO: Improvement: Restore keys should not be `main` but the branch we're merging into.
     - name: Restore Cargo Hack Keys (PR)
       if: github.event_name == 'pull_request'
       uses: actions/cache/restore@5a3ec84eff668545956fd18022155c47e93e2684  # Version 4.2.3
--- a/scripts/cargo-hack-check/script.py
+++ b/scripts/cargo-hack-check/script.py
@@ -322,7 +322,7 @@ class CargoHackCheckEntry:
         # Changing this doesn't void the cache.
         # TODO: Consider adding the command line options to the cache key.
         return [
-            "cargo", "hack", "check", "--feature-powerset", "--all-targets", "--no-dev-deps", "--manifest-path", self.manifest_path_posix
+            "cargo", "hack", "check", "--feature-powerset", "--all-targets", "--manifest-path", self.manifest_path_posix
         ]
 
 


scripts/cargo-hack-check/script.py
```diff
@@ -0,0 +1,442 @@
...
+        capture_output=True,
+        text=True,
+        check=True,
+    )
+    return result.stdout.strip()
+
+
+def list_cargo_hack_check_commands() -> Set[CargoHackCheckEntry]:
+    result = subprocess.run(
+        ["cargo", "hack", "check", "--no-dev-deps", "--print-command-list"],
```
ntn-x2, line 401: Is `--feature-powerset` missing here or it does not matter at this stage?

.github/actions/cargo-hack-check-cached/action.yml
```diff
@@ -0,0 +1,56 @@
+name: Cargo Hack Check (Cache-Aware)
```
ntn-x2, line 1: Somewhere we need to specify `-D warnings` or then `RUSTFLAGS` as an env to fail on warnings.

scripts/cargo-hack-check/script.py
```diff
@@ -0,0 +1,442 @@
...
+        return f"CargoHackCommand({self.manifest_path})"
+
+    def __repr__(self):
+        return f"CargoHackCommand(cmd_line={self.cmd_line}, manifest_path={self.manifest_path})"
+
+    def __hash__(self):
+        return hash(self.manifest_path_posix)
+
+    def as_feature_powerset_command(self) -> List[str]:
+        # TODO: Current command used: "cargo hack check --feature-powerset --keep-going --all-targets"
```
ntn-x2: Why don't we use `--all-targets` or `--keep-going` anymore?

danielSanchezQ:
> I will say one last time that if we were to apply mitigations other than having a full bash script for caching (i.e., partitioning, dealing properly with cache among runs - which will be useful also for other jobs besides cargo-hack, etc.), this custom, somewhat complicated solution might not be needed. This said, since general consensus is that this is a good thing to have regardless of whatever else we can do and not only as a last resort, I guess we can go ahead. Despite this being a very good job, I am personally not happy to see this solution applied before having dealt with other issues in a more "canonical" way by using existing tools. So I am leaving a review but I will let someone else approve this PR :)

I agree that custom stuff is not really nice. BUT, in this case we lose nothing to try, we see how it works and if its worth it. If it gives more problems than not we remove it. If it improves our times a lot we keep. IMHO that is the best route now. Otherwise we will have to go to pruning features which may not even be possible (or open other problem paths like bigger build times).

scripts/cargo-hack-check/script.py
strinnityk: It does not matter at this stage. We're just using this as the source of truth, to get the list of crates cargo hack will run on.

scripts/cargo-hack-check/script.py
strinnityk: Good question. Can't remember why I left it that way. Updated.
EDIT: I remembered: While I was developing I had issues due to development dependencies interfering with crate sorting. It's a leftover.

PR #2005, opened 2026-01-06
fix(ci): Cache pruning

strinnityk:
- Fix and extend the cache pruning action:
  - Fix cache pruning API calls.
  - Add support for multiple prune modes (e.g. `keep-latest`) and additional parameters.

- Rework cache key formats for `setup-cargo-cache` and `update-cargo-cache`:
  - Align them with the cargo-hack caching key scheme.
  - Improve key scheme format by placing broader scope components earlier to enable better filtering and reuse.

- Move cargo cache pruning from the setup phase to the update phase:
  - Allows concurrent workflows to reuse the same cache.
  - Still ensures only the most recent cache is retained via `mode: keep-latest`.

- Add automatic cleanup of code-check caches when a PR is closed.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/.github/workflows/code-check.yml
+++ b/.github/workflows/code-check.yml
@@ -57,37 +57,6 @@ jobs:
       - name: Run cargo-deny
         run: cargo-deny --locked --all-features check --hide-inclusion-graph -c .cargo-deny.toml --show-stats -D warnings
 
-  debug-common-cache:
-    name: "Self hosted lints"
-    runs-on: [self-hosted, Linux]
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@85e6279cec87321a52edac9c87bce653a07cf6c2  # Version 4.2.2
-
-      - name: Set up Cargo cache
-        uses: ./.github/actions/setup-cargo-cache
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          key-prefix: code-check/feature-combinations/${{ matrix.platform.label }}  # Get cache from feature combinations on self-hosted
-
-      - name: Setup Circuits
-        uses: ./.github/actions/setup-circuits
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Run cargo clippy on workspace
-        uses: actions-rs/cargo@9e120dd99b0fbad1c065f686657e914e76bd7b72  # Version 1.0.1
-        with:
-          command: clippy
-          args: --locked --all --all-targets --all-features
-
-      - name: Update Cargo cache
-        if: success() || failure()
-        uses: ./.github/actions/update-cargo-cache
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          key-prefix: test/code-check/lints/${{ matrix.os }}
-
   features:
     name: Check packages feature combinations
     strategy:


.github/actions/prune-cache/action.yml
```diff
@@ -1,34 +1,134 @@
...
+  ref:
+    description: Optional ref to scope cache deletion (omit for all)
+    required: false
+  limit:
+    # NOTE: Uses a single page. Add pagination if the cache count grows.
+    description: Max number of caches to fetch (page size)
+    required: false
+    default: "100"
+  mode:
+    description: "Prune mode. Available options: 'delete-all' (default), 'keep-latest'"
```
ntn-x2, line 19: What does `keep-latest` do?

.github/workflows/code-check.yml
```diff
@@ -118,25 +121,30 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@85e6279cec87321a52edac9c87bce653a07cf6c2  # Version 4.2.2
+
       - name: Set up Cargo cache
         uses: ./.github/actions/setup-cargo-cache
         with:
-          key: ${{ github.ref }}->${{ github.workflow }}->${{ github.job }}->${{ matrix.os }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
+          key-prefix: code-check/lints/${{ matrix.os }}
```
ntn-x2, line 136: Doesn't this keep a single cache per CI job? I think we want one cache per PR, not a shared one for everyone, since PRs can differ greatly from each other and this would just lead to continuous "contention", or do I understand this wrong?

scripts/cargo-hack-check/script.py
```diff
@@ -23,6 +23,7 @@
 # - WORKSPACE_ROOT pointing to the root of the workspace.
 # - HASH_SCRIPT being in the same directory as this script.
 # If these prerequisites are not met, the script will not behave as expected.
+# Moving these to parameters would be safer.
```
ntn-x2, line 26: I guess this is unrelated to this PR?

.github/workflows/code-check.yml
ntn-x2: Ok nevermind, we namespace by PR number, so all good.

scripts/cargo-hack-check/script.py
strinnityk: Hmmm, might be a leftover line I didn't commit when I should've. I'll leave it here since I want an actual explicit statement about it.

.github/workflows/code-check.yml
strinnityk: Yes, we still do.
The reason I handled it this way is to allow for broader restore-keys.

.github/actions/prune-cache/action.yml
strinnityk: Deletes all caches that start with a given key prefix except the latest one.

ntn-x2: Do you think it's worth specifying?

strinnityk: Certainly. It's how we're enabling a cache to be used concurrently.

PR #2009, opened 2026-01-08
feat(chain): reject blocks from future slots

youngjoon-lee:

Implements the [block header validation rule #6](https://www.notion.so/nomos-tech/Cryptarchia-v1-Protocol-Specification-21c261aa09df810cb85eff1c76e5798c?source=copy_link#21c261aa09df81ffaf4ff3168252a097): reject blocks where `block.slot > current_slot`.

The validation is implemented in the chain-service (not cryptarchia-engine) to keep the engine time-independent and avoid changing all engine callers (including tests) to pass `current_slot`. Block validation logic is already distributed across cryptarchia-engine and nomos-ledger, so I think it's fine to implement this check in the chain-service which can subscribe to time-service for slots.

This PR doesn't include tests because the chain-service currently lacks tests for its block processing logic. Setting up the testing infra would significantly expand this PR. A follow-up PR will introduce testing for the entire logic.

Yes

spec: @davidrusu 
dev: @youngjoon-lee 

Need to be improved as below.

As discussed in the Notion [comment](https://www.notion.so/Cryptarchia-v1-Protocol-Specification-21c261aa09df810cb85eff1c76e5798c?d=2e1261aa09df80668983001cfed0f9ba&source=copy_link#21c261aa09df81ffaf4ff3168252a097), I'll update the spec to compare slots instead of times.

## Checklist

> [!WARNING]  

* [x] 2. Description added.


nomos-services/chain/chain-network/src/lib.rs
```diff
@@ -710,7 +710,6 @@ where
 {
     debug!("received proposal {:?}", block);
 
-    // TODO: filter on time?
```
youngjoon-lee, line 713: Removing this because the time (slot) check is handled by chain-service (as implemented in this PR), not by this chain-network.

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -236,6 +243,15 @@ impl Cryptarchia {
         let id = header.id();
         let parent = header.parent();
         let slot = header.slot();
+
+        // Reject blocks from future slots
+        if slot > current_slot {
```
davidrusu, line 248: Don't forget to update the cryptarchia spec to compare by slot rather than time

PR #2012, opened 2026-01-09
chore: basic metrics

andrussal:

This PR add metrics to the code locations which seem to provide the most useful initial insight how system behaves. It covers most services. Previously we didn't have almost any metrics, so this is attempt to get some coverage. 

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/testnet/tracing/grafana/dashboards/testnet_metrics.json
+++ b/testnet/tracing/grafana/dashboards/testnet_metrics.json
@@ -1,226 +1,1650 @@
 {
   "annotations": {
-    "list": [
-      {
-        "builtIn": 1,
-        "datasource": {
-          "type": "grafana",
-          "uid": "-- Grafana --"
-        },
-        "enable": true,
-        "hide": true,
-        "iconColor": "rgba(0, 211, 255, 1)",
-        "name": "Annotations & Alerts",
-        "type": "dashboard"
-      }
-    ]
+    "list": []
   },
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 1,
+  "id": null,
   "links": [],
   "panels": [
     {
-      "datasource": {
-        "type": "prometheus",
-        "uid": "PBFA97CFB590B2093"
-      },
-      "fieldConfig": {
-        "defaults": {
-          "color": {
-            "mode": "palette-classic"
-          },
-          "custom": {
-            "axisBorderShow": false,
-            "axisCenteredZero": false,
-            "axisColorMode": "text",
-            "axisLabel": "",
-            "axisPlacement": "auto",
-            "barAlignment": 0,
-            "barWidthFactor": 0.6,
-            "drawStyle": "line",
-            "fillOpacity": 0,
-            "gradientMode": "none",
-            "hideFrom": {
-              "legend": false,
-              "tooltip": false,
-              "viz": false
-            },
-            "insertNulls": false,
-            "lineInterpolation": "linear",
-            "lineWidth": 1,
-            "pointSize": 5,
-            "scaleDistribution": {
-              "type": "linear"
-            },
-            "showPoints": "auto",
-            "spanNulls": false,
-            "stacking": {
-              "group": "A",
-              "mode": "none"
-            },
-            "thresholdsStyle": {
-              "mode": "off"
-            }
-          },
-          "mappings": [],
-          "thresholds": {
-            "mode": "absolute",
-            "steps": [
-              {
-                "color": "green",
-                "value": null
-              },
-              {
-                "color": "red",
-                "value": 80
-              }
-            ]
-          }
-        },
-        "overrides": []
-      },
+      "id": 1,
+      "type": "timeseries",
+      "title": "blend_inbound_messages_failed_total",
       "gridPos": {
         "h": 8,
         "w": 12,
         "x": 0,
         "y": 0
       },
-      "id": 2,
-      "options": {
-        "legend": {
-          "calcs": [],
-          "displayMode": "list",
-          "placement": "bottom",
-          "showLegend": true
-        },
-        "tooltip": {
-          "mode": "single",
-          "sort": "none"
-        }
-      },
-      "pluginVersion": "11.4.0",
-      "targets": [
-        {
-          "disableTextWrap": false,
-          "editorMode": "builder",
-          "expr": "da_mempool_pending_items",
-          "fullMetaSearch": false,
-          "includeNullMetadata": true,
-          "legendFormat": "__auto",
-          "range": true,
-          "refId": "A",
-          "useBackend": false
-        }
-      ],
-      "title": "Mempool: Pending DA blobs",
-      "type": "timeseries"
-    },
-    {
       "datasource": {
         "type": "prometheus",
         "uid": "PBFA97CFB590B2093"
       },
-      "fieldConfig": {
-        "defaults": {
-          "color": {
-            "mode": "palette-classic"
-          },
-          "custom": {
-            "axisBorderShow": false,
-            "axisCenteredZero": false,
-            "axisColorMode": "text",
-            "axisLabel": "",
-            "axisPlacement": "auto",
-            "barAlignment": 0,
-            "barWidthFactor": 0.6,
-            "drawStyle": "line",
-            "fillOpacity": 0,
-            "gradientMode": "none",
-            "hideFrom": {
-              "legend": false,
-              "tooltip": false,
-              "viz": false
-            },
-            "insertNulls": false,
-            "lineInterpolation": "linear",
-            "lineWidth": 1,
-            "pointSize": 5,
-            "scaleDistribution": {
-              "type": "linear"
-            },
-            "showPoints": "auto",
-            "spanNulls": false,
-            "stacking": {
-              "group": "A",
-              "mode": "none"
-            },
-            "thresholdsStyle": {
-              "mode": "off"
-            }
-          },
-          "mappings": [],
-          "thresholds": {
-            "mode": "absolute",
-            "steps": [
-              {
-                "color": "green",
-                "value": null
-              },
-              {
-                "color": "red",
-                "value": 80
-              }
-            ]
-          }
-        },
-        "overrides": []
+      "targets": [
+        {
+          "refId": "A",
+          "expr": "increase(blend_inbound_messages_failed_total[$__range])",
+          "legendFormat": "{{service_name}} {{instance}} {{action}} {{origin}} {{reason}} {{kind}} {{result}} {{method}} {{endpoint}} {{status}} {{status_class}} {{mode}} {{strategy}} {{message_type}}"
+        }
+      ]
+    },
+    {
+      "id": 2,
+      "type": "timeseries",
+      "title": "blend_messages_received_total",
+      "gridPos": {
+        "h": 8,
+        "w": 12,
+        "x": 12,
+        "y": 0
       },
+      "datasource": {
+        "type": "prometheus",
+        "uid": "PBFA97CFB590B2093"
+      },
+      "targets": [
+        {
+          "refId": "A",
+          "expr": "increase(blend_messages_received_total[$__range])",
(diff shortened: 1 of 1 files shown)


nomos-da/network/core/src/swarm/executor.rs
```diff
@@ -348,35 +345,29 @@ where
...
             }
             ExecutorBehaviourEvent::ValidatorDispersal(event) => {
-                tracing::info!(
-                    counter.behaviour_events_received = 1,
-                    event = EVENT_VALIDATOR_DISPERSAL,
-                    share_size = event.share_size()
-                );
+                let share_size = event.share_size().unwrap_or(0);
+                metrics::da_behaviour_event_received(EVENT_VALIDATOR_DISPERSAL);
+                metrics::da_behaviour_share_size_bytes(EVENT_VALIDATOR_DISPERSAL, share_size);
```
ntn-x2: Should we rather only log this if the share size is not `None`, i.e., if the event is a blob dispersal event? Although DA is getting removed, so maybe all this code will be deleted I think.

nomos-da/network/core/src/swarm/executor.rs
```diff
@@ -348,35 +345,29 @@ where
...
             }
             ExecutorBehaviourEvent::Replication(event) => {
-                tracing::info!(
-                    counter.behaviour_events_received = 1,
-                    event = EVENT_REPLICATION,
-                    share_size = event.share_size()
-                );
+                let share_size = event.share_size().unwrap_or(0);
+                metrics::da_behaviour_event_received(EVENT_REPLICATION);
+                metrics::da_behaviour_share_size_bytes(EVENT_REPLICATION, share_size);
```
ntn-x2: Same as above.

nomos-da/network/core/src/swarm/validator.rs
```diff
@@ -319,28 +320,24 @@ where
...
             }
             ValidatorBehaviourEvent::Dispersal(event) => {
-                tracing::info!(
-                    counter.behaviour_events_received = 1,
-                    event = EVENT_VALIDATOR_DISPERSAL,
-                    share_size = event.share_size()
-                );
+                let share_size = event.share_size().unwrap_or(0);
+                metrics::da_behaviour_event_received(EVENT_VALIDATOR_DISPERSAL);
+                metrics::da_behaviour_share_size_bytes(EVENT_VALIDATOR_DISPERSAL, share_size);
```
ntn-x2: Same, if you decide to apply my suggestion.

nomos-da/network/core/src/lib.rs
```diff
@@ -5,6 +5,7 @@ pub mod addressbook;
 )]
 pub mod behaviour;
 pub mod maintenance;
+mod metrics;
```
ntn-x2: Nitpick: perhaps we could separate pub from local mods?

nomos-tracing/src/metrics/emit.rs
```diff
@@ -0,0 +1,159 @@
...
+use opentelemetry::{
+    KeyValue, Value, global,
+    metrics::{Counter, Gauge, Histogram, Meter},
+};
+
+fn meter() -> Meter {
+    global::meter("nomos-node")
+}
+
+static U64_COUNTERS: OnceLock<Mutex<HashMap<&'static str, Counter<u64>>>> = OnceLock::new();
```
ntn-x2: Why not a `LazyLock` instead? I think in this case it would be a simpler tool to use.

nomos-tracing/src/metrics/emit.rs
```diff
@@ -0,0 +1,159 @@
...
+        .lock()
+        .expect("f64 histogram lock poisoned");
+
+    histograms
+        .entry(name)
+        .or_insert_with(|| meter().f64_histogram(name).init())
+        .clone()
+}
+
+pub fn counter_u64(name: &'static str, value: impl IntoMetricU64, attributes: &[KeyValue]) {
```
ntn-x2: I find the naming of this group of functions confusing. Can we call it `increase_u64_counter` or `increase_counter_u64` instead?

nomos-tracing/src/lib.rs
```diff
@@ -3,3 +3,45 @@ pub mod logging;
 pub mod metrics;
 pub mod panic;
 pub mod tracing;
+
+pub use opentelemetry;
+
+#[macro_export]
+macro_rules! metric_counter_u64 {
```
ntn-x2: Also here, I think a clearer name would greatly help since this macro is used everywhere in the codebase. What about using the same naming we use in the `emit` module? Something like `increase_u64_counter`?

nomos-services/wallet/src/metrics.rs
```diff
@@ -0,0 +1,31 @@
...
+        WalletMsg::GetBalance { .. } => "get_balance",
+        WalletMsg::FundAndSignTx { .. } => "fund_and_sign_tx",
+        WalletMsg::GetLeaderAgedNotes { .. } => "get_leader_aged_notes",
+    };
+
+    nomos_tracing::metric_counter_u64!(wallet_requests_total, 1, kind = kind);
+}
+
+pub fn wallet_response_send_failed_get_balance() {
+    nomos_tracing::metric_counter_u64!(wallet_response_send_failed_total, 1, kind = "get_balance");
```
ntn-x2: These strings should be the same used above. So I suggest making them constants instead.

nomos-services/storage/src/metrics.rs
```diff
@@ -0,0 +1,12 @@
+use std::time::Instant;
+
+pub fn storage_request_failed() {
+    nomos_tracing::metric_counter_u64!(storage_request_failed_total, 1);
```
ntn-x2: We don't want to log regardless how long processing took in this case?

nomos-services/blend/src/metrics.rs
```diff
@@ -0,0 +1,31 @@
+pub fn mix_packets_processed_total() {
+    nomos_tracing::metric_counter_u64!(blend_mix_packets_processed_total, 1);
+}
+
+pub fn peers_connected(count: usize) {
+    nomos_tracing::metric_gauge_u64!(blend_peers_connected, count as u64);
+}
+
+pub fn outbound_publish_ok() {
+    nomos_tracing::metric_counter_u64!(blend_messages_sent_total, 1, action = "publish");
```
ntn-x2: This also should be a constant, as soon as we use it in more than a single place.

nomos-services/blend/src/metrics.rs
```diff
@@ -0,0 +1,31 @@
...
+pub fn outbound_forward_err() {
+    nomos_tracing::metric_counter_u64!(blend_outbound_messages_failed_total, 1, action = "forward");
+}
+
+pub fn inbound_message_ok() {
+    nomos_tracing::metric_counter_u64!(blend_messages_received_total, 1);
+}
+
+pub fn inbound_message_err() {
+    nomos_tracing::metric_counter_u64!(blend_inbound_messages_failed_total, 1);
```
ntn-x2: Can we log the messsage type as well? Would that be possible?

services/chain/chain-network/src/bootstrap/download.rs
```diff
@@ -233,6 +246,7 @@ where
             Poll::Ready(Some(result)) => match result {
                 Ok((id, block)) => {
                     self.last = Some(id);
+                    self.blocks_downloaded = self.blocks_downloaded.saturating_add(1);
```
hansieodendaal, line 245: It may be worthwhile to also keep a moving average of the time it takes to download blocks, say with a window of between 20 to 50.

services/chain/chain-network/src/sync/orphan_handler.rs
```diff
@@ -250,11 +275,17 @@ where
...
+
                         if download.orphan_info.orphan_id == block_id {
+                            metrics::orphan_observe_parent_fetch_ok(
+                                download.download_started_at.elapsed(),
+                            );
                             self.state = DownloaderState::Idle;
                         }
 
                         self.pending_orphans_queue.remove(&block_id);
+                        metrics::orphan_blocks_pending(self.pending_orphans_queue.len());
```
hansieodendaal, line 312: Is tracking re-org depths maybe a thing? I quickly looked, did not see any metrics in that regard, but I may have missed it. 

In the same breath, and excuse my ignorance, is a fork chain at all possible, and what metrics would we track for that?

nodes/api-common/src/metrics.rs
```diff
@@ -0,0 +1,75 @@
...
+
+fn get_status_class(status: StatusCode) -> &'static str {
+    match status {
+        status if status.is_success() => "2xx",
+        status if status.is_redirection() => "3xx",
+        status if status.is_client_error() => "4xx",
+        status if status.is_server_error() => "5xx",
+        _ => "unknown",
+    }
+}
```
danielSanchezQ, line 64: This method looks so wrong 😅 
either we use plain `ifs`, or you match on on the status itself (not sure is possible). But not this in the middle.

nomos-tracing/src/metrics/emit.rs
danielSanchezQ: is this updated from runtime?

hansieodendaal (commented):
These look great, just some comments to consider above.

Now a related but maybe out of scope question, what about collecting some OS vitals, which could be opt in? For seed/other public nodes it is good to know processing load, heap and stack memory size and remaining disk space. It may come from a different path than the Rust code.

nomos-services/storage/src/metrics.rs
andrussal: Initially I actually added but removed later. I thought that mixing failures with successes would give wrong data. Here in my opinion most important info is to quickly see if we have failures.

nomos-da/network/core/src/swarm/validator.rs
andrussal: Makes sense

nomos-tracing/src/metrics/emit.rs
andrussal: Here it is stored once when created at runtime and later reused. Just to avoid creating same metric instances multiple times.

services/chain/chain-network/src/bootstrap/download.rs
andrussal: Actually I have another metrics that sends time of downloading all blocks: `chainsync_download_blocks_seconds`. Then in Garafana we can configure dashboard to show these I think.

nomos-services/storage/src/metrics.rs
ntn-x2: What would be confusing about adding also the processing time?

services/chain/chain-network/src/sync/orphan_handler.rs
andrussal: Yes, we currently emit forks count by `consensus_forks_count`. Orphan downloader just downloads any chain that has missing parents, it doesn't know actual fork length.

andrussal:
> These look great, just some comments to consider above.
> 
> Now a related but maybe out of scope question, what about collecting some OS vitals, which could be opt in? For seed/other public nodes it is good to know processing load, heap and stack memory size and remaining disk space. It may come from a different path than the Rust code.

Yes, these would be good things to have. But I think system metrics are usually collected by other means, not in code tho. Maybe something Rust specific might be worth adding at some point tho.

hansieodendaal:
> Yes, these would be good things to have. But I think system metrics are usually collected by other means, not in code tho. Maybe something Rust specific might be worth adding at some point tho.

Ah, one thing I could think of is the internal db(s) size(s), which should be readily available. This has always been a thing in my previous blockchain project... "How big is the blockchain db currently?"

services/chain/chain-network/src/sync/orphan_handler.rs
hansieodendaal: Theoretical question: We have the consensus security parameter; could that be challenged?

danielSanchezQ:
Btw, this bulk changes I think we should introduce them little by little. First the functionality and later the integration.
image
Big prs are more difficult to review.

andrussal:
Was replaced by another PR few weeks ago

PR #2014, opened 2026-01-12
chore: improve tests for beginners

hansieodendaal:

This PR:
- reduced spam prints in tests;
- added `POL_PROOF_DEV_MODE` runtime environment variable check;
- added environment variables to choose between running the release or debug process binaries; and
- updated the test README.

Yes

n/a

n/a

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/tests/src/lib.rs
+++ b/tests/src/lib.rs
@@ -3,9 +3,10 @@ pub mod common;
 pub mod nodes;
 pub mod topology;
 
+use std::{env, ops::Mul as _, sync::LazyLock, time::Duration};
+
 use nomos_core::proofs::leader_proof::POL_PROOF_DEV_MODE;
 use nomos_libp2p::{Multiaddr, PeerId, multiaddr};
-use std::{env, ops::Mul as _, sync::LazyLock, time::Duration};
 
 static IS_SLOW_TEST_ENV: LazyLock<bool> =
     LazyLock::new(|| env::var("SLOW_TEST_ENV").is_ok_and(|s| s == "true"));
--- a/tests/src/nodes/mod.rs
+++ b/tests/src/nodes/mod.rs
@@ -1,8 +1,7 @@
 pub mod executor;
 pub mod validator;
 
-use std::path::PathBuf;
-use std::sync::LazyLock;
+use std::{path::PathBuf, sync::LazyLock};
 
 use reqwest::Client;
 use tempfile::TempDir;
--- a/tests/src/nodes/executor.rs
+++ b/tests/src/nodes/executor.rs
@@ -1,3 +1,11 @@
+use std::{
+    collections::HashSet,
+    net::SocketAddr,
+    process::{Child, Command, Stdio},
+    str::FromStr as _,
+    time::Duration,
+};
+
 use broadcast_service::BlockInfo;
 use chain_service::CryptarchiaInfo;
 use common_http_client::CommonHttpClient;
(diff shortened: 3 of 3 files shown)


tests/src/nodes/validator.rs
```diff
@@ -408,9 +419,10 @@ impl Validator {
     ) -> Vec<Option<HeaderId>> {
         let mut results = vec![None; tx_hashes.len()];
 
+        let mut tick = 0u8;
         let _ = tokio::time::timeout(timeout, async {
             loop {
-                let headers = self.get_headers(None, None).await;
+                let headers = self.get_headers(None, None, tick == 0).await;
```
andrussal, line 425: Is the goal here to print at every 256 iteration? (I didn't check in practice which number is reasonable tho).

tests/src/nodes/mod.rs
```diff
@@ -29,3 +33,47 @@ fn persist_tempdir(tempdir: &mut TempDir, label: &str) -> std::io::Result<()> {
...
+            if std::fs::exists(&release_binary).unwrap() {
+                release_binary
+            } else {
+                panic!(
+                    "Could not find nomos binary in release path '{}'",
+                    release_binary.display()
+                );
+            }
+        }
+        (_, _) => {
```
andrussal: Would it be more explicit to panic if both are specified? Otherwise it feels like misconfiguration...

tests/src/tests/cryptarchia/happy.rs
```diff
@@ -37,30 +37,34 @@ async fn happy_test(topology: &Topology) {
...
             {
-                println!(
-                    "waiting... {}",
-                    stream::iter(nodes)
-                        .then(async |n| { format!("{}", n.consensus_info().await.height) })
-                        .collect::<Vec<_>>()
-                        .await
-                        .join(" | ")
-                );
+                if tick.is_multiple_of(10) {
```
andrussal, line 47: I trust your decision about the output. I don't really have strong preference.

tests/src/lib.rs
```diff
@@ -57,3 +57,10 @@ pub fn secret_key_to_provider_id(
     )
     .unwrap()
 }
+
+pub fn verify_pol_proof_dev_mode() {
+    assert!(
+        env::var(POL_PROOF_DEV_MODE).is_ok(),
```
andrussal, line 64: Should we also check then if it is `true`?

tests/src/nodes/validator.rs
hansieodendaal: The goal is to limit header state prints because they overpower the interesting test output - 
I used u8 here so it can print again when/if it cycles. Ideally, we only want to see this at test edges.

tests/src/nodes/mod.rs
hansieodendaal: I thought about that and opted for _better user experience_, but maybe you are right. I will change it to be stricter.

tests/src/lib.rs
hansieodendaal: Ah, interesting. The code will accept any value, just as long as the env var is present. I changed the README to make this explicit.

andrussal (approved):
LGTM!

(PS! Just curios why do create branches in your personal github account? Is this common practise? But I guess nothing wrong with it:))

hansieodendaal:
> (PS! Just curios why do create branches in your personal github account? Is this common practise? But I guess nothing wrong with it:))

It tends to keep the main repo cleaner, and yes, this is fairly common practice. I usually create branches in my personal fork so I can work independently without affecting the main repository, experiment freely, and keep a clean history while iterating.

PR #2023, opened 2026-01-15
feat(blend): remove plaintext Ed25519 key from Blend config

ntn-x2:

First step towards a deeper integration of KMS into Blend components: removing the plaintext key from Blend config and replacing it with a KMS key ID. As a temporary workaround until we remove the need to access the secret key inside our libp2p stack, we have a new operator to extract a Ed25519 key from the KMS, behind the `unsafe` feature flag. We could still leave this operator in the future, but from Blend we will at some point move away from this.

Also, the X25519 logic has been moved inside the KMS crate, and a new Ed25519 operator added that derives the X25519 secret key. For the time being, this operator is not used anywhere but again, once Blend won't require access to the secret Ed25519 key anymore, this will be used to derive the key to pass to the cryptographic processor.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nomos-kms/keys/src/keys/ed25519/private.rs
+++ b/nomos-kms/keys/src/keys/ed25519/private.rs
@@ -3,7 +3,7 @@ use nomos_utils::serde::{deserialize_bytes_array, serialize_bytes_array};
 use rand_core::CryptoRngCore;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use subtle::ConstantTimeEq as _;
-use zeroize::ZeroizeOnDrop;
+use zeroize::{ZeroizeOnDrop, Zeroizing};
 
 use crate::{
     keys::{Ed25519PublicKey, Ed25519Signature},
@@ -48,8 +48,8 @@ impl UnsecuredEd25519Key {
     }
 
     #[must_use]
-    pub fn to_bytes(&self) -> [u8; SECRET_KEY_LENGTH] {
-        self.0.to_bytes()
+    pub fn to_bytes(&self) -> Zeroizing<[u8; SECRET_KEY_LENGTH]> {
+        self.0.to_bytes().into()
     }
 
     #[must_use]
@@ -72,7 +72,7 @@ impl<'de> Deserialize<'de> for UnsecuredEd25519Key {
     where
         D: Deserializer<'de>,
     {
-        let bytes = deserialize_bytes_array::<KEY_SIZE, _>(deserializer)?;
+        let bytes = Zeroizing::new(deserialize_bytes_array::<KEY_SIZE, _>(deserializer)?);
         Ok(Self(SigningKey::from_bytes(&bytes)))
     }
 }
--- a/nomos-services/blend/src/edge/backends/libp2p/mod.rs
+++ b/nomos-services/blend/src/edge/backends/libp2p/mod.rs
@@ -46,7 +46,7 @@ impl<RuntimeServiceId> BlendBackend<PeerId, RuntimeServiceId> for Libp2pBlendBac
         let (swarm_command_sender, swarm_command_receiver) = mpsc::channel(CHANNEL_SIZE);
         let swarm_identity = {
             let mut non_ephemeral_signing_key_bytes = non_ephemeral_signing_key.to_bytes();
-            Keypair::try_from_bytes(&mut non_ephemeral_signing_key_bytes)
+            Keypair::try_from_bytes(&mut non_ephemeral_signing_key_bytes[..])
                 .expect("Cryptographic secret key should be a valid Ed25519 private key.")
         };
         let swarm = BlendSwarm::new(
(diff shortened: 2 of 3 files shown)


nodes/nomos-node/node/src/config/blend/serde/mod.rs
```diff
@@ -14,13 +14,11 @@ pub mod edge;
 ///
 /// This includes all values that are not strictly related to any specific
 /// deployment and that users have to specify when starting up the node.
-#[derive(Clone, Derivative, Serialize, Deserialize)]
-#[derivative(Debug)]
+#[derive(Clone, Debug, Derivative, Serialize, Deserialize)]
```
youngjoon-lee:
```suggestion
#[derive(Clone, Debug, Serialize, Deserialize)]
```
it seems that we can now remove `Derivative` from here and `Cargo.toml`.

youngjoon-lee (approved):
👍

nomos-kms/keys/src/operators/ed25519/exfiltrate_secret_key.rs
```diff
@@ -0,0 +1,39 @@
+use core::fmt::{self, Debug, Formatter};
+
+use tokio::sync::oneshot;
+use tracing::error;
+
+use crate::keys::{
+    Ed25519Key, UnsecuredEd25519Key, errors::KeyError, secured_key::SecureKeyOperator,
+};
+
+pub struct ExfiltrateSecretKeyOperator {
```
danielSanchezQ: I think we can call it `LeakSecretKeyOperator`, `Exfiltrate` is too much southern european 😜

danielSanchezQ (approved):

nomos-kms/keys/src/keys/ed25519/private.rs
```diff
@@ -44,10 +47,14 @@ impl UnsecuredEd25519Key {
         self.0.as_bytes()
     }
 
-    /// Return a reference to the inner secret key.
     #[must_use]
-    pub const fn as_inner(&self) -> &SigningKey {
-        &self.0
+    pub fn to_bytes(&self) -> [u8; SECRET_KEY_LENGTH] {
```
hansieodendaal: I think we could harden this and similar functions returning raw secret bytes
```suggestion
    pub fn to_bytes(&self) -> Zeroizing<[u8; SECRET_KEY_LENGTH]> {
```

nomos-services/blend/src/edge/backends/libp2p/mod.rs
```diff
@@ -33,21 +34,27 @@ impl<RuntimeServiceId> BlendBackend<PeerId, RuntimeServiceId> for Libp2pBlendBac
...
         rng: Rng,
+        non_ephemeral_signing_key: UnsecuredEd25519Key,
     ) -> Self
     where
         Rng: RngCore + Send + 'static,
     {
         let (swarm_command_sender, swarm_command_receiver) = mpsc::channel(CHANNEL_SIZE);
+        let swarm_identity = {
+            let mut non_ephemeral_signing_key_bytes = non_ephemeral_signing_key.to_bytes();
+            Keypair::try_from_bytes(&mut non_ephemeral_signing_key_bytes)
```
hansieodendaal: Using the zeroizing array suggestion, these `non_ephemeral_signing_key_bytes` raw secret bytes would automatically be zeroized once they go out of scope.

nomos-services/blend/src/edge/mod.rs
```diff
@@ -192,17 +202,38 @@ where
...
+                settings.non_ephemeral_signing_key_id,
+                KeyOperators::Ed25519(Box::new(ExfiltrateSecretKeyOperator::new(sender))),
+            )
+            .await
+            .expect("Failed to interact with KMS to fetch non-ephemeral signing key.");
+            receiver
+                .await
+                .expect("Failed to retrieve non-ephemeral signing key from KMS.")
+        }
+        .await;
```
hansieodendaal: - I am missing the reasoning for the outside async block here, as we are already in an async fn and each statement is awaited on.
- Why not propagate the errors instead of the panics?

nomos-services/blend/src/edge/backends/libp2p/mod.rs
danielSanchezQ: Yeah, it could be introduced here.

nomos-services/blend/src/edge/mod.rs
ntn-x2: Good catch. I fixed the unnecessary `async` here: https://github.com/logos-blockchain/logos-blockchain/pull/2023/commits/0be22e31c4742cf7220ef5efe534f7b01783fc54.

As for the panic, there is little point in propagating errors here. The service as such (which is spawned on its own thread) can do very little if it is not able to fetch the key, i.e., there is no fallback mechanism at the moment. So we are assuming the operation to never fail for now. If it does, the service thread will panic and it will be up to the overwatch runner to decide what to do.

PR #2027, opened 2026-01-16
feat(chain): remove txs from mempool only after added to honest chain

youngjoon-lee:

Based on https://github.com/logos-blockchain/logos-blockchain/pull/2022

Each node should try their best to include txs (in mempool) to the honest chain. This is the ultimate goal of this PR, even though it originally started from a [task](https://www.notion.so/nomos-tech/Iteration-17-2c3261aa09df80e6a00bfc9120b77c77?source=copy_link#2e1261aa09df80ddaaeec634eedb7d7b) in the iteration note.

This feature may not be very urgent because we've decided to not implement the adaptive adversary protection at this moment. But, I think we should have this feature in the end. The PR is not small, but not difficult to read, I think. Let me explain changes in detail.

This PR contains 3 commits. All of them can be reviewed together, but you can review them separately if you prefer.
1. When we receive a block, and if the block is added to a fork (not the honest chain), now we don't remove txs (of the block) from mempool, so that those txs can be selected/proposed by this node later. Of course, if the block is added to the honest chain, we remove its txs from mempool, as before.
2. When `chain-leader` proposes a block, it has been applying the block to the engine before publishing it to the network. We keep that behavior, but in this PR, we remove txs of the block from mempool, so that the leader doesn't select the same txs again if he wins the next lottery before the published block is arrived back. This change is probably not the main focus of this PR, but I think it's under the same context.
3. Now we re-insert reorged txs back into mempool, as described in the [iteration note](https://www.notion.so/nomos-tech/Iteration-17-2c3261aa09df80e6a00bfc9120b77c77?source=copy_link#2e1261aa09df80ddaaeec634eedb7d7b). Regardless whether blocks are finalized or not, whenever the honest chain is switched, now we bring all txs from the previous honest branch back to the mempool, so that we can select/propose those txs again based on the new honest chain later.

Because of the 1, there may be some txs that will never be deleted from the mempool, in the case where the fork that contains those txs becomes the new honest chain at some point and never be reorged again. (because those txs were not in the honest chain and weren't be removed from mempool). An upcoming PR will handle this case by removing all txs of each new LIB from mempool. I didn't do it here to not make this PR bigger.

I hope so.

@youngjoon-lee 

This behavior is not defined in the spec because this is a behavior that we recommend.

N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.


consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -72,11 +78,21 @@ impl State {
 // paper k defines the forking depth of chain we accept without more
 // analysis s defines the length of time (unit of slots) after the fork
 // happened we will inspect for chain density
-fn maxvalid_bg<Id>(local_chain: Branch<Id>, branches: &Branches<Id>, k: u64, s: u64) -> Branch<Id>
+//
+// It returns the new local chain and the lowest common ancestor with the
+// previous.
```
danielSanchezQ: This should be part of docs `\\\` instead of `\\`, it comes from before but I just realize.

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -94,29 +111,40 @@ where
...
+    (cmax, lowest_common_ancestor_with_cmax)
 }
 
 // Implementation of the fork choice rule as defined in the Ouroboros Praos
-// paper k defines the forking depth of chain we can accept
-fn maxvalid_mc<Id>(local_chain: Branch<Id>, branches: &Branches<Id>, k: u64) -> Branch<Id>
+// paper k defines the forking depth of chain we can accept.
+//
+// It returns the new local chain and the lowest common ancestor with the
+// previous.
```
danielSanchezQ: Likewise, lets make it par of docs

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -586,6 +660,23 @@ where
...
+    }
+}
+
+impl<Id> Deref for ReorgedBlocks<Id> {
+    type Target = [Id];
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
danielSanchezQ: `Deref` should't be use for this cases. Usually it is used for the kind of smart pointers. It is better to implement `AsRef` instead which makes it explicit.

nomos-services/chain/chain-leader/src/lib.rs
```diff
@@ -658,6 +662,34 @@ where
...
+            )
+            .await
+        {
+            error!(
+                "failed to remove txs included in block {:?} from mempool: {e:?}",
+                block.header().id()
+            );
+        }
+    }
 }
```
danielSanchezQ, line 698: Shouldn't this be part of the mempool adapter for the chain leader instead?

nomos-services/chain/chain-network/src/lib.rs
```diff
@@ -716,18 +716,29 @@ where
...
-                .collect::<Vec<_>>(),
-        )
-        .await
-        .unwrap_or_else(|e| error!("Could not mark transactions in block: {e}"));
+    // Re-insert reorged txs back into the mempool.
+    for tx in reorged_txs {
+        if let Err(e) = mempool_adapter.add_transaction(tx).await {
+            error!("Could not reinsert a reorged tx into mempool: {e:?}");
+        }
+    }
```
danielSanchezQ: I think we can do this in an asynchronous way, no need to be sequential.

nomos-services/chain/chain-service/src/lib.rs
```diff
@@ -861,7 +867,14 @@ where
             .await;
         }
 
-        Ok((cryptarchia, pruned_blocks))
+        let mut reorged_txs = Vec::new();
+        for id in reorged_blocks.iter() {
+            if let Some(block) = relays.storage_adapter().get_block(id).await {
+                reorged_txs.extend(block.transactions().cloned());
+            }
+        }
```
danielSanchezQ: This can be completely asynchronous as well, even if you want to keep the ordering. Also you can probably just consume the block and not cloned the txs that are gonna be droped inmediately after.

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: Thanks. Done 310cb2ef2bdca38fdfdc4244adde0887efa9ea99

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: Thanks. Done 310cb2ef2bdca38fdfdc4244adde0887efa9ea99

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: Thanks. Done 035de33a75af67441d172030836bc956a58358ca. Instead of using `AsRef`, I defined some specific methods explicitly, which I think is better.

nomos-services/chain/chain-leader/src/lib.rs
youngjoon-lee: I thought about it, but ended up with defining this helper function because I didn't want to make the caller more verbose. If we define `MempoolAdapter::remove_transactions_in_block`, we could simplify the `&block.transactions().map(...)...` part here, but the error logging is still verbose. So, I chose to keep only the unit operation `MempoolAdapter::remove_transactions` (which is used in other places as well), and define this helper function. Wdyt?

nomos-services/chain/chain-network/src/lib.rs
youngjoon-lee: Yeah, I was lazy. Thank you. 51bf943fd669ae3935a64c860e3ea6ca1ccf0bc5
Actually, I'm working on improving the mempool API to support multiple insertions. The mempool already has an ability to do that (as it supports multiple removals already), but its API doesn't support it yet. I will introduce it in a next API and replace this fix with the new API.

nomos-services/chain/chain-service/src/lib.rs
youngjoon-lee: Thank you. Done 51bf943fd669ae3935a64c860e3ea6ca1ccf0bc5

consensus/cryptarchia-engine/src/lib.rs
danielSanchezQ: Why not implement `Iterator` instead of the custom method?

nomos-services/chain/chain-leader/src/lib.rs
danielSanchezQ: Ok, fine with this. We should add this as a explanation there in the method (so we know why it wasn't included in the adapter in the first place).

nomos-services/chain/chain-service/src/lib.rs
danielSanchezQ: Superb, thanks!

danielSanchezQ (approved):

consensus/cryptarchia-engine/src/lib.rs
youngjoon-lee: yeah, done [fcdd262](https://github.com/logos-blockchain/logos-blockchain/pull/2027/commits/fcdd26264b858f66098475c297e0a92df7ee22d9) (i still had to have the `iter` method, even though now we implement `IntoIterator`, because clippy complains.)

nomos-services/chain/chain-leader/src/lib.rs
youngjoon-lee: i agree. done [edb6f45](https://github.com/logos-blockchain/logos-blockchain/pull/2027/commits/edb6f45f2eb152a9bc151b8c0345ca5b1c15fdb4)

PR #2028, opened 2026-01-16
chore: migrate nomos* to logos-blockchain*

ntn-x2:

No more No mos.....

### What I have done

* Renamed folders to be less redundant (i.e., removing the former `nomos` prefix altogether from all the places it appeared)
* Removed the `book/` folder that was completely outdated
* Renamed all services to have the `_service` suffix in their crate name
* Prefixed all crate names with `logos-blockchain` in their cargo file (for external consumers) and with `lb_` in our root `Cargo.toml` (for internal consumption). This make it immediate to understand whether a dependency is an internal one or an external one.
* Removed project structure from `README` cause that is also very much outdated

### Things I noticed will need to change in the specs

* **All** libp2p protocol names

### Things that need to happen after this PR is merged

Will make tickets for these items once this PR is merged.

* Update the testing framework, then uncomment the cucumber tests from this repo
* Fire extinguish broken stuff due to the renaming
* Once the Notion is also migrated from nomos to logos, update the links in this repo as well

In the end, there should be no mentions of `nomos` left in this repo 🚀🚀🚀

### What's relevant to check

* All files in the root of the crate, e.g., Docker compose files, `Cargo.toml`, `README.md`
* Workflows and actions inside `.github`
* Logic inside `bundler` and `c-bindings`
* Scripts folder
* Anything else that is not directly checked by `cargo` and that might contain inconsistent information

Well...

I bear full responsibility: @ntn-x2 

Will need changes for libp2p protocol names and DSTs.

See point above.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/scripts/setup-logos-blockchain-circuits.ps1
+++ b/scripts/setup-logos-blockchain-circuits.ps1
@@ -88,10 +88,9 @@ function Download-And-Extract {
         [string]$InstallDir
     )
 
-    # TODO: Rename these artifacts to `logos-blockchain-circuits-...`
     $artifacts = @(
-        "nomos-circuits-$Version-$Platform.tar.gz",
-        "nomos-circuits-$Version-$Platform.zip"
+        "logos-blockchain-circuits-$Version-$Platform.tar.gz",
+        "logos-blockchain-circuits-$Version-$Platform.zip"
     )
 
     $tempDir = Join-Path ([System.IO.Path]::GetTempPath()) ([System.Guid]::NewGuid().ToString())
--- a/scripts/setup-logos-blockchain-circuits.sh
+++ b/scripts/setup-logos-blockchain-circuits.sh
@@ -101,8 +101,7 @@ check_existing_installation() {
 # Download and extract the release
 download_release() {
     local platform="$1"
-    # TODO: Rename these artifacts to `logos-blockchain-circuits-...`
-    local artifact="nomos-circuits-${VERSION}-${platform}.tar.gz"
+    local artifact="logos-blockchain-circuits-${VERSION}-${platform}.tar.gz"
     local url="https://github.com/${REPO}/releases/download/${VERSION}/${artifact}"
     local temp_dir=$(mktemp -d)
 


zk/circuits/prover/Cargo.toml
```diff
@@ -4,14 +4,14 @@ description = { workspace = true }
 edition     = { workspace = true }
 keywords    = { workspace = true }
 license     = { workspace = true }
-name        = "circuits-prover"
+name        = "logos-blockchain-circuits-prover"
```
danielSanchezQ, line 7: Why this renames? I think something went wrong.

ntn-x2: I will explain why. It's to namespace crates that we develop/maintain from third-party dependencies. I want to namespace our crates with `lb-` prefix, so it would be `lb-circuits-prover`. Third-party dependencies would still use it as `logos-blockchain-*`.

ntn-x2: I find it much cleaner, looking at the end result. But we can discuss whether it's something we want to do. In the end, we are doing a massive amount of renaming in this PR anyway, so I thought changing these to be what I envisioned would not be a big issue.

danielSanchezQ: Ok, I'm not against it, I just didn't understand it. Let's discuss this in the channel with everyone tho. 
Also, imo, we should go for something shorter :sweat_smile: `lb` sounds better to me.

Cargo.toml
```diff
@@ -83,73 +83,72 @@ resolver = "2"
...
+lb-tracing-service               = { default-features = false, path = "./services/tracing", package = "logos-blockchain-tracing-service" }
+lb-tx-service                    = { default-features = false, path = "./services/tx-service", package = "logos-blockchain-tx-service" }
+lb-utils                         = { default-features = false, path = "./utils", package = "logos-blockchain-utils" }
+lb-utxotree                      = { default-features = false, path = "./utxotree", package = "logos-blockchain-utxotree" }
+lb-wallet                        = { default-features = false, path = "./wallet", package = "logos-blockchain-wallet" }
+lb-wallet-http-client            = { default-features = false, path = "./wallet-http-client", package = "logos-blockchain-wallet-http-client" }
+lb-wallet-service                = { default-features = false, path = "./services/wallet", package = "logos-blockchain-wallet-service" }
+lb-witness-generator             = { default-features = false, path = "./zk/circuits/witness-generator", package = "logos-blockchain-witness-generator" }
```
danielSanchezQ, line 150: I see it now, it is nice as all of our deps will be grouped as well by cargo fmt. Will check that now on the files themselves.

c-bindings/src/api/types/value.rs
```diff
@@ -1,19 +1,19 @@
...
+//! `lb_core::mantle::Value` is a type alias defined in an external
 //! crate. Since `cbindgen` cannot reliably emit typedefs for
 //! re-exported aliases, we define the FFI-facing `Value` locally and
 //! enforce at compile time that it is exactly the same type as the
 //! upstream one.
 
-// Upstream Nomos type
-use nomos_core::mantle::Value as NomosValue;
+// Upstream Logos blockchain type
+use lb_core::mantle::Value as LogosBlockchainValue;
```
danielSanchezQ, line 10: This probably could use the lb shortener instead. But can be done later in future prs related to logos-core.

ntn-x2: What do you mean?

ntn-x2: Ah you mean the `LogosBlockchainValue` type? If so, yes I agree we can push that in a different PR.

PR #2030, opened 2026-01-18
feat(DA): Disable DA 

pradovic:

This PR disables DA (Data Availability) from the nomos-node for testnet/mainnet deployment, as per the project-wide decision. The goal was to remove DA code deploying, while keeping the services code still in the project for maintenance and documentation. Also, DA related types are kept, per spec (for example in SDP definitions), but the service is guarded not to process or produce DA tx or DA sdp declarations.

  ### Changes

  Executor Binary Removal
  - Completely removed the nomos-executor binary and its HTTP client
  - Removed executor-related CLI commands from nomos-cli
  - Removed executor from CI workflows (code-check, codecov)

  DA Removal from Node
  - Removed DA services (DaVerifier, DaSampling, DaNetwork) from the Nomos struct
  - Removed DA types and dependencies from chain-leader and chain-network services
  - Removed DA HTTP endpoints from the node API
  - Removed DA configs from node configuration

  Transaction Filtering
  - Added mempool-level rejection for transactions containing blob operations (Op::ChannelBlob)
  - Added block proposal filtering to reject DA blob transactions (defensive measure)

  Test Cleanup
  - Removed all DA-specific tests (tests/src/tests/da/)
  - Removed DA test helpers and executor test node
  - Removed DA topology configs
  - Cleaned up validator test methods (removed DA-related methods)

  Testnet/Cfgsync Cleanup
  - Removed DA parameters from cfgsync configuration
  - Removed DA network port from host configuration
  - Updated provider creation to only include Blend providers

  What's NOT Removed

  - DA service crates still exist in the workspace for compilation (keeping codebase maintained)
  - DA type definitions in nomos-core remain (follows the spec)
  - common-http-client DA methods remain (used by DA service crates)

  ### Follow-up Work Needed

  1. Testing Framework Update: The smoke tests (smoke_local.rs, cucumber_smoke_local.rs) reference the executor binary in comments. The testing framework (testing-framework-runner-local) needs to be updated to work without DA and to remove exutor.
  2. SDP E2E Test Rewrite: The SDP operations test (sdp_ops.rs) currently tests DA service declaration which are rejeceted now. It was not trivial to rewrite it to use Blend transaction so I would tackle this in a separate PR and then enable the enable the test.

Yes

@pradovic @danielSanchezQ @bacv 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/tests/src/topology/configs/mod.rs
+++ b/tests/src/topology/configs/mod.rs
@@ -7,7 +7,7 @@ pub mod time;
 pub mod tracing;
 
 use blend::GeneralBlendConfig;
-use consensus::{create_genesis_tx_with_declarations, GeneralConsensusConfig, ProviderInfo};
+use consensus::{GeneralConsensusConfig, ProviderInfo, create_genesis_tx_with_declarations};
 use key_management_system_service::backend::preload::PreloadKMSBackendSettings;
 use network::GeneralNetworkConfig;
 use nomos_core::{
@@ -15,7 +15,7 @@ use nomos_core::{
     sdp::{Locator, ServiceType},
 };
 use nomos_utils::net::get_available_udp_port;
-use rand::{thread_rng, Rng as _};
+use rand::{Rng as _, thread_rng};
 use tracing::GeneralTracingConfig;
 
 use crate::topology::configs::{


nomos-services/tx-service/src/backend/mod.rs
```diff
@@ -12,10 +18,40 @@ pub enum MempoolError {
...
+pub trait DaOpsCheck {
+    /// Returns true if this item contains DA-related operations.
+    fn has_da_ops(&self) -> bool;
+}
+
+impl DaOpsCheck for SignedMantleTx {
+    fn has_da_ops(&self) -> bool {
+        self.mantle_tx.ops.iter().any(|op| match op {
+            Op::ChannelBlob(_) => true,
+            Op::SDPDeclare(decl) => decl.service_type == ServiceType::DataAvailability,
```
ntn-x2, line 38: Why not simply removing these variants from those enums? I think it's ok to leave the DA services and crates so we can keep developing them, but all other mentions of DA should, in my opinion, go. This is one such instance. It's just much less prone to mis-use or errors in the future, if we were to use this service type elsewhere.

danielSanchezQ: I think everything will be removed as per our [convo](https://discordapp.com/channels/973324189794697286/1460228968803471433)

ntn-x2: Even better, in my opinion.

pradovic: @danielSanchezQ I thought we agreed to leave the enums in SDP and general types (per spec) and just guard on the tx side. This is why this "hack" is introduced to filter out tx there are DA related. Does this makes sense or did you guys want to remove the DA stuff from all types as well (SDP, etc..)? 

cc. @bacv  as well

pradovic: I don't mind removing it from the types as well, it will be a bit bigger change, and we might have some compilation conflicts in the DA code (for opinions for example) etc... These would need to be handled, but probably doable

danielSanchezQ: > @danielSanchezQ I thought we agreed to leave the enums in SDP and general types (per spec) and just guard on the tx side. This is why this "hack" is introduced to filter out tx there are DA related. Does this makes sense or did you guys want to remove the DA stuff from all types as well (SDP, etc..)?
> 
> cc. @bacv as well

Sorry, I didn't realize. We should not remove them from defs, otherwise some code will just do not compile. In this case we can go with this. Also this is not written in stone, we can modify or remove later on if we consider it needed.

pradovic: No worries! I think it might make sense to leave them in definitions because we still mention them in specs, just that node is not supporting it yet. If we would remove from all definitions we would probably need to update all the specs as well, or document that those are missing. The PR should be ready for review now then

PR #2034, opened 2026-01-20
feat(cryptarchia): Implement stake inference

danielSanchezQ:

This PR introduces stake inference integration into cryptarchia ledger as per the [specification](https://www.notion.so/nomos-tech/Total-Stake-Inference-22d261aa09df8051a454caa46ec54b34)

It's pretty small and contained.

@davidrusu @youngjoon-lee @danielSanchezQ 

I think specification should contain specific precision for floating point operations and types, as all nodes should reach the same conclusions. That part is missing atm.

It just extracts the block density out of the computation, but its a minor change in the flow rather than the computation itself.

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/ledger/src/cryptarchia/block_density.rs
+++ b/ledger/src/cryptarchia/block_density.rs
@@ -103,8 +103,8 @@ mod tests {
     fn test_large_slot_jump() {
         let mut inference = create_inference(5, 0);
         inference.increment_block_density(Slot::from(100));
-        // 100 empty slots pushed (only last 4 remain in window) + 1 filled = 1 block
-        assert_eq!(inference.current_block_density(), 1);
+        // 100 empty slots pushed, more than period, so block density is 0
+        assert_eq!(inference.current_block_density(), 0);
     }
 
     #[test]


ledger/src/cryptarchia/mod.rs
```diff
@@ -312,10 +324,17 @@ impl LedgerState {
             .iter()
             .map(|(_, (utxo, _))| utxo.note.value)
             .sum::<Value>();
+        let slot: Slot = 0.into();
+        let stake_inference = Arc::new(StakeInference::new(
+            LEARNING_RATE,
+            config.consensus_config.active_slot_coeff,
+            config.consensus_config.security_param.get().into(),
+        ));
+        let block_density_inference = BlockDensity::new(stake_inference.period(), slot);
```
danielSanchezQ: This is one of the initialization of the inference objects.

ledger/src/cryptarchia/mod.rs
```diff
@@ -107,14 +108,20 @@ impl LedgerState {
...
-        // TODO: update once supply can change
-        let total_stake = self.epoch_state.total_stake;
+        // increment density for new slot
+        let mut block_density_inference = self.block_density.clone();
+        block_density_inference.increment_block_density(slot);
+        // infere new total stake
+        let total_stake = self.stake_inference.total_stake_inference(
+            self.epoch_state.total_stake,
+            block_density_inference.current_block_density(),
+        );
```
danielSanchezQ, line 132: This is probably the most important part of the integration. In this case I'm updating it before processing any new header or txs, which I'm not really sure is ok. Should it be here, or after/while processing the new block txs?

ledger/src/cryptarchia/mod.rs
```diff
@@ -142,6 +164,7 @@ impl LedgerState {
                 slot,
                 next_epoch_state,
                 epoch_state,
+                block_density: block_density_inference,
```
youngjoon-lee: I think, this behavior is different from the spec.

[Total stake inference spec](https://www.notion.so/nomos-tech/Total-Stake-Inference-22d261aa09df8051a454caa46ec54b34?source=copy_link#22d261aa09df80d782f0e69a48fa4d43) defines the density observation period as  $6\lfloor \frac{k}{f} \rfloor$. But, precisely, this must be the same as `StakeDistributionSnapshot + BufferPhase` period in the [Epoch Schedule](https://www.notion.so/nomos-tech/Cryptarchia-v1-Protocol-Specification-21c261aa09df810cb85eff1c76e5798c?source=copy_link#21c261aa09df81a6ad71e403bbc9fa4f) in the Cryptarchia spec. **That is, I think we don't need the sliding window in this PR for tracking the block density.**

This fact is explained in the total stake inference spec, but **not explicitly**. In the [Algorithm](https://www.notion.so/nomos-tech/Total-Stake-Inference-22d261aa09df8051a454caa46ec54b34?source=copy_link#22d261aa09df8015bf22fb28aa0c0ea1), we call `density_over_slots(epoch_slot, PERIOD)` where `epoch_slot` is the 1st slot of the current epoch and `PERIOD` is $6\lfloor \frac{k}{f} \rfloor$. This says that we should accumulate the block density during the first $6\lfloor \frac{k}{f} \rfloor$ slots in the current epoch, and stop accumulating it after the slot (i.e., once [LotteryConstantFinalization](https://www.notion.so/nomos-tech/Cryptarchia-v1-Protocol-Specification-21c261aa09df810cb85eff1c76e5798c?source=copy_link#21c261aa09df81a6ad71e403bbc9fa4f) period starts). I think we should make it more explicit by linking the total stake spec to the Cryptarchia spec (like $f$ and $k$).

This logic is similar as accumulating the epoch nonce $\eta$, but slightly different.

In contrast, in this PR, block density is accumulated until right before a new epoch starts, and the new total stake is inferred from its final value. I believe this differs from the specification. Please let me know if I'm wrong. 

If my understanding above is correct, we can simply remove the `BlockDensity` struct, and add a new simple field `block_density` (or any better name) to the `LedgerState`, similar as `LedgerState.nonce`:
```diff
diff --git a/ledger/src/cryptarchia/mod.rs b/ledger/src/cryptarchia/mod.rs
index 77c0568d..28e9f1ba 100644
--- a/ledger/src/cryptarchia/mod.rs
+++ b/ledger/src/cryptarchia/mod.rs
@@ -89,12 +89,20 @@ impl EpochState {
 pub struct LedgerState {
     // All available Unspent Transtaction Outputs (UTXOs) at the current slot
     pub utxos: UtxoTree,
-    // randomness contribution
+    // Randomness contribution.
+    // This must be accumulated continuously even after the Lottery Constants Finalization starts,
+    // but the value accumulated right before the Lottery Constants Finalization starts must be
+    // used for constructing the next epoch state.
     #[cfg_attr(feature = "serde", serde(with = "lb_groth16::serde::serde_fr"))]
     pub nonce: Fr,
+    // Num of blocks accumulated from the beginning of the current epoch.
+    // This accumulation must stop when Lottery Constants Finalization period starts,
+    // and the value must be used for constructing the next epoch state.
+    pub block_density: u64,
     pub slot: Slot,
     // rolling snapshot of the state for the next epoch, used for epoch transitions
     pub next_epoch_state: EpochState,
+    // Current epoch state
     pub epoch_state: EpochState,
 }
 ```
Then, the `LedgerState::update_epoch_state` could be implemented as below:
```diff
diff --git a/ledger/src/cryptarchia/mod.rs b/ledger/src/cryptarchia/mod.rs
index 28e9f1ba..652e0450 100644
--- a/ledger/src/cryptarchia/mod.rs
+++ b/ledger/src/cryptarchia/mod.rs
@@ -115,8 +115,14 @@ impl LedgerState {
             });
         }
 
-        // TODO: update once supply can change
-        let total_stake = self.epoch_state.total_stake;
+        // Accumulate block density if we are before Lottery Constants Finalization period.
+        if self.slot < config.total_stake_snapshot(self.slot) {
+            self.block_density += 1;
+        }
+        // Infer new total stake just in case for new epoch.
+        let total_stake =
+            infer_total_stake(self.epoch_state.total_stake, self.block_density, config);
+
         let current_epoch = config.epoch(self.slot);
         let new_epoch = config.epoch(slot);
 
@@ -135,6 +141,7 @@ impl LedgerState {
             Ok(Self {
                 slot,
                 next_epoch_state,
+                // The newly inferred total stake is ignored in this case.
                 ..self
             })
         } else if new_epoch == current_epoch + 1 {
```

As I mentioned above, this logic is similar but different from the epoch nonce accumulation.
Epoch nonce is accumulated continuously regardless of the Epoch Schedule by `update_nonce` in the following code. In contrast, block density must stop accumulating at some point.
 https://github.com/logos-blockchain/logos-blockchain/blob/47ae18e95f643bde563b4769212b37f6f018fed3/ledger/src/cryptarchia/mod.rs#L203-L206

Also, the epoch nonce is applied to the `EpochState` conditionally by the following function: https://github.com/logos-blockchain/logos-blockchain/blob/47ae18e95f643bde563b4769212b37f6f018fed3/ledger/src/cryptarchia/mod.rs#L33-L39
It's possible to have this logic inside `EpochState` because the `EpochState` owns the `nonce` itself. In contrast, `EpochState` doesn't have `block_density` --it only has `total_stake`. Therefore, block density cannot be applied to `EpochSatte` within this function in the same way as the nonce. That's why I implemented the separate logic in `LedgerState::update_epoch_state` above.

This is my understanding. Please correct me if I'm wrong. Now I feel we have to refactor this module much clearer and safer in near future. It is not so easy to read since we've updated it for a long time.

danielSanchezQ: Ah I see. Thanks for the detailed explanation.
Indeed this was confusing:
image
We should definetely explain it better or link to the other spec because by itself it made me reach the sliding window decision.
Btw, if you had the changes you could have open a PR on top of this one with the changes instead, they are always welcome!
Also, I agree, we may have to refactor this at some point. But lets come back to it in a couple months 😉

ledger/src/cryptarchia/block_density.rs
```diff
@@ -0,0 +1,117 @@
+use std::ops::RangeInclusive;
+
+use lb_cryptarchia_engine::Slot;
+
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Clone)]
+pub struct BlockDensity {
+    period_range: RangeInclusive<Slot>,
+    density: u64,
+}
```
danielSanchezQ, line 10: @youngjoon-lee I kept this to keep the logic encapsulated, now it is actually really simple.

ledger/src/cryptarchia/mod.rs
```diff
@@ -127,10 +148,12 @@ impl LedgerState {
             Ok(Self {
                 slot,
                 next_epoch_state,
+                block_density: block_density_inference,
                 ..self
             })
         } else if new_epoch == current_epoch + 1 {
             // case 2)
+            let block_density = BlockDensity::new(self.stake_inference.period(), slot);
```
danielSanchezQ, line 156: Now this is reset for cases 2 and 3. So it will only track for the next epoch.

ledger/src/cryptarchia/mod.rs
youngjoon-lee: IIUC, it's fine. The spec doesn't explain this boundary case explicitly. But in the case 2 of the current implementation, we are replacing `epoch_state` with `new_epoch_state` **without** updating `next_epoch_state.nonce` using the current block. The `nonce` is updated after this function is returned. 
So, I think, what you implemented for the total stake in this PR is consistent with how/when epoch nonce is updated. As we discussed, we ofc need to refactor this module more nicely in the future.

ledger/src/cryptarchia/stake.rs
```diff
@@ -0,0 +1,140 @@
...
+            security_parameter,
+        }
+    }
+
+    pub fn period(&self) -> u64 {
+        const PERIOD_CONSTANT: u64 = 6;
+        ((self.security_parameter * PERIOD_CONSTANT) as f64)
+            .div(self.slot_activation_coefficient)
+            .floor() as u64
+    }
```
youngjoon-lee, line 34: This function calculates $\lfloor \frac{6k}{f} \rfloor$, but I think we need $6\lfloor \frac{k}{f} \rfloor$, as we [calculate `s`](https://github.com/logos-blockchain/logos-blockchain/blob/df1ad3738341eac0b2e79b0448a751cfa16e16c3/consensus/cryptarchia-engine/src/config.rs#L25).

ledger/src/cryptarchia/mod.rs
youngjoon-lee: I think, now we can rename `block_density_inference` variable to `block_density`.

ledger/src/cryptarchia/stake.rs
youngjoon-lee: or. I think you can simply do `2s` if it’s accessible from this module.

danielSanchezQ: You are right, It should not make a difference I think, but I made it explicit as per the spec. Thanks!

PR #2039, opened 2026-01-21
fix(utxotree): increase tree height (except root): 31 -> 32

youngjoon-lee:

(This is the 1st fix to make PoL work. More PRs are coming soon. The last PR in this series will introduce an actual PoL test. Until then, each PR contains only unit tests.)

The circuit expects a merkle tree with 32 layers “under” the root. But, our implementation has had the tree with 31 layers under the root since the very beginning. We must fix this mismatch to make PoL work.

This PR increases the tree height from 31 to 32.
All Merkle paths now have a length of 32. (Previously it was 31).

I hope so.

spec: @thomaslavaur 
dev: @youngjoon-lee 

The spec should explicitly mention that the "tree height" means the number of layers under the root (i.e. except the root). I made a [suggestion](https://www.notion.so/v1-1-Proof-of-Leadership-Specification-2e9261aa09df80058244c902defc6da2?d=2ef261aa09df8073aeea001c3fe71a46&source=copy_link#2e9261aa09df80dc94c8f28b225e0a36), and it has been applied to the [spec](https://www.notion.so/nomos-tech/v1-1-Proof-of-Leadership-Specification-2e9261aa09df80058244c902defc6da2?source=copy_link#2e9261aa09df80dc94c8f28b225e0a36).

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/utxotree/src/merkle.rs
+++ b/utxotree/src/merkle.rs
@@ -12,20 +12,19 @@ use rpds::RedBlackTreeSetSync;
 
 use crate::CompressedUtxoTree;
 
-const TREE_HEIGHT: usize = 32;
-const PATH_LENGTH: usize = TREE_HEIGHT - 1; // Exclude the root
+const TREE_HEIGHT_EXCEPT_ROOT: usize = 32;
 
 const EMPTY_VALUE: Fr = <Fr as Field>::ZERO;
 
 fn empty_subtree_root<Hash: Digest>(height: usize) -> Fr {
-    static PRECOMPUTED_EMPTY_ROOTS: OnceLock<[Fr; TREE_HEIGHT]> = OnceLock::new();
+    static PRECOMPUTED_EMPTY_ROOTS: OnceLock<[Fr; TREE_HEIGHT_EXCEPT_ROOT + 1]> = OnceLock::new();
     assert!(
-        height < TREE_HEIGHT,
-        "Height must be less than {TREE_HEIGHT}: {height}"
+        height <= TREE_HEIGHT_EXCEPT_ROOT,
+        "Height{height} must be <={TREE_HEIGHT_EXCEPT_ROOT}"
     );
     PRECOMPUTED_EMPTY_ROOTS.get_or_init(|| {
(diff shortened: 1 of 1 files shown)


utxotree/src/merkle.rs
```diff
@@ -215,13 +214,13 @@ impl<Item: AsRef<Fr>> Node<Item> {
                 if index < left.capacity() {
                     // Going down left subtree, store right sibling hash
                     let mut path = left.path::<Hash>(index)?;
-                    assert!(path.len() < PATH_LENGTH, "Path length exceeded");
+                    assert!(path.len() < TREE_HEIGHT_EXCEPT_ROOT, "Path length exceeded");
```
thomaslavaur, line 217: TREE_HEIGHT_EXCEPT_ROOT +1? Because the path should be of length 32 too

utxotree/src/merkle.rs
```diff
@@ -215,13 +214,13 @@ impl<Item: AsRef<Fr>> Node<Item> {
...
                     path.push(MerkleNode::Right(right.value::<Hash>()));
                     Some(path)
                 } else {
                     // Going down right subtree, store left sibling hash
                     let mut path = right.path::<Hash>(index - left.capacity())?;
-                    assert!(path.len() < PATH_LENGTH, "Path length exceeded");
+                    assert!(path.len() < TREE_HEIGHT_EXCEPT_ROOT, "Path length exceeded");
```
thomaslavaur, line 223: ditto

utxotree/src/merkle.rs
youngjoon-lee: TREE_HEIGHT_EXCEPT_ROOT is defined as 32. So, this assertion checks that the path length is 32.

thomaslavaur: ho my bad

danielSanchezQ:
Nice @youngjoon-lee , I was expecting the path till (plus) the root to be 32 instead.

danielSanchezQ (approved):

PR #2044, opened 2026-01-21
feat: separate user from deployment configs

ntn-x2:

This PR splits up the deployment configuration and the user configuration into two distinct sets. The deployment settings can now be specified with the optional `--deployment` CLI argument. If not given, the default `mainnet` deployment is used. If provided, it can be either the name of a well-known deployment, or it falls back to the path containing the description of a custom deployment (as is the case for the `one-node-custom-deployment.yaml` file).

This gives us a much easier way to let users specify whatever values they want for their node, while at the same time making it possible to define and circulate deployment files that are completely independent from those user settings. This will be very useful once we start having dev and test deployments.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.


c-bindings/src/api/lifecycle.rs
```diff
@@ -14,6 +22,9 @@ pub type InitializedLogosBlockchainNodeResult = PointerResult<LogosBlockchainNod
 ///
 /// - `config_path`: A pointer to a string representing the path to the
 ///   configuration file.
+/// - `deployment`: A pointer to a string representing either a well-known
+///   deployment name (e.g., "mainnet") or a path to a deployment YAML file. If
+///   null, defaults to "mainnet".
```
hansieodendaal: I think the default should be some testnet, not mainnet, at least for now.

danielSanchezQ: Agreed, we do not want people playing with mainnet accidentally. So testnet should be a better default

c-bindings/src/api/lifecycle.rs
```diff
@@ -22,45 +33,42 @@ pub type InitializedLogosBlockchainNodeResult = PointerResult<LogosBlockchainNod
 #[unsafe(no_mangle)]
 pub extern "C" fn start_lb_node(
     config_path: *const c_char,
+    deployment: *const c_char,
```
danielSanchezQ, line 36: `deployment_path` ?

c-bindings/src/api/lifecycle.rs
```diff
@@ -22,45 +33,42 @@ pub type InitializedLogosBlockchainNodeResult = PointerResult<LogosBlockchainNod
...
+/// - `deployment`: A pointer to a string for the deployment (can be null).
 ///
 /// # Returns
 ///
 /// A `Result` containing either the initialized `LogosBlockchainNode` or an
 /// error code.
-fn initialize_lb_node(config_path: *const c_char) -> Result<LogosBlockchainNode, OperationStatus> {
+fn initialize_lb_node(
+    config_path: *const c_char,
+    deployment_arg: *const c_char,
```
danielSanchezQ: `deployment_path` too?

c-bindings/src/api/lifecycle.rs
ntn-x2: Yeah but there won't be any mainnet running, so by default it won't do anything as the node won't find any peers. I think ideally in the long term mainnet should be the default. But if you think for now testnet is a better default, I can change. But long term I think it should be mainnet so the CLI is as simple as possible.

danielSanchezQ: > Yeah but there won't be any mainnet running, so by default it won't do anything as the node won't find any peers. I think ideally in the long term mainnet should be the default. But if you think for now testnet is a better default, I can change. But long term I think it should be mainnet so the CLI is as simple as possible.

Once we have mainnet the idea is that users will use an app. We will setup the app so the users do not need to touch anything and packaging should include the proper deployment config. But lets discuss it when we have to do it (?)

danielSanchezQ (approved):

c-bindings/src/api/lifecycle.rs
ntn-x2: This is either a well-known deployment name or a path, so not just a path. You have a better naming suggestion?

danielSanchezQ: Ok, I missed that it could have the name only. Then deployment should be fine, thanks.

ntn-x2: I used the same name for the arg in both functions, as it means the same thing in both contexts: https://github.com/logos-blockchain/logos-blockchain/pull/2044/commits/8391ca98418c8869c660d78443c3c80d2beb6082.

c-bindings/src/api/lifecycle.rs
ntn-x2: Introduced a new testnet definition which just returns the mainnet definition (which itself is also not officially defined yet). The new testnet deployment is used by default: https://github.com/logos-blockchain/logos-blockchain/pull/2044/commits/052bfb19796534ec603c5805050f5553fe7b9580.

PR #2068, opened 2026-01-28
chore(cryptarchia): Remove active slot coefficient loading from configuration

danielSanchezQ:

* It removes f from config on load, but it replaces it with featured gated values. 
* It also refactors the variable naming
* It feature gate `P`, `T0` and `T1` values from pol zk crate for testing values

Should be simple to track.

@youngjoon-lee @danielSanchezQ @thomaslavaur 

Yes

No

## Checklist

* [x] 2. Description added.

### DIFF
--- a/services/wallet/Cargo.toml
+++ b/services/wallet/Cargo.toml
@@ -27,8 +27,8 @@ lb-key-management-system-service = { workspace = true }
 lb-ledger                        = { workspace = true }
 lb-services-utils                = { workspace = true }
 lb-storage-service               = { features = ["mock"], workspace = true }
-lb-wallet                        = { workspace = true }
 lb-utxotree                      = { workspace = true }
+lb-wallet                        = { workspace = true }
 overwatch                        = { workspace = true }
 
 [lints]


zk/proofs/pol/src/lottery.rs
```diff
@@ -5,23 +5,44 @@ use std::{
...
+        BigUint::from_str_radix(
+            "30644e72e131a029b85045b68181585d2833e84879b9709143e1f593f0000001",
+            16,
+        )
+        .expect("P constant should parse")
+    }
+    #[cfg(feature = "high-active-slot-coefficient")]
+    {
+        BigUint::from_str_radix(
+            "30644e72e131a029b85045b68181585d2833e84879b9709143e1f593f0000001",
```
danielSanchezQ: I still need to add the proper values. Waiting for them.

consensus/cryptarchia-engine/src/config.rs
```diff
@@ -8,15 +8,36 @@ pub struct Config {
     // trigger the additional fork selection rule, which is however only expected to be used
     // during bootstrapping.
     pub security_param: NonZero<u32>,
-    // f, the rate of occupied slots
-    pub active_slot_coeff: f64,
+    #[serde(skip, default = "Config::active_slot_coefficient")]
```
ntn-x2: I think this needs to be feature-gated behind the `serde` feature.

consensus/cryptarchia-engine/Cargo.toml
```diff
@@ -17,15 +17,17 @@ ignored = ["logos-blockchain-utils"]
...
+thiserror = "1"
+time = { default-features = false, version = "0.3" }
+tokio = { default-features = false, features = ["time"], optional = true, version = "1" }
+tracing = { workspace = true }
 
 [features]
 serde = ["dep:cfg_eval", "dep:lb-utils", "dep:serde", "dep:serde_with"]
 tokio = ["dep:tokio"]
+high-active-slot-coefficient = []
+default = []
```
ntn-x2: I think this is default behaviour so we don't need to specify it.

consensus/cryptarchia-engine/Cargo.toml
```diff
@@ -17,15 +17,17 @@ ignored = ["logos-blockchain-utils"]
 workspace = true
 
 [dependencies]
-cfg_eval   = { optional = true, version = "0.1" }
-lb-utils   = { features = ["time"], optional = true, workspace = true }
-serde      = { optional = true, workspace = true }
+cfg_eval = { optional = true, version = "0.1" }
```
ntn-x2: Something is wrong with formatting here 👀

consensus/cryptarchia-engine/src/config.rs
```diff
@@ -8,15 +8,36 @@ pub struct Config {
...
+    }
+    #[must_use]
+    const fn active_slot_coefficient() -> f64 {
+        #[cfg(not(feature = "high-active-slot-coefficient"))]
+        {
+            1f64
+        }
+        #[cfg(feature = "high-active-slot-coefficient")]
+        {
+            1f64 / 30f64
```
ntn-x2: This should be reverse, no?

consensus/cryptarchia-engine/src/config.rs
```diff
@@ -8,15 +8,36 @@ pub struct Config {
...
 impl Config {
+    #[must_use]
+    pub const fn new(security_param: NonZero<u32>) -> Self {
+        Self {
+            security_param,
+            active_slot_coefficient: Self::active_slot_coefficient(),
+        }
+    }
+    #[must_use]
+    const fn active_slot_coefficient() -> f64 {
```
ntn-x2: I see that this function is only used in this file. Do you think it would make things easier to simply pass `base_period_length` to the config instead? Or at least should this config here take `base_period_length` so we avoid defining a new feature for the crate, and we define a different config where the active slot coefficient is needed? If that place already depends on the `pol` crate, then we would simply re-use the same feature. Would that work? And does that make sense?

danielSanchezQ: Yeah, this is a good idea. But I don't know how much will it spread. Let me take a look.

consensus/cryptarchia-engine/src/config.rs
danielSanchezQ: yes, thanks!

consensus/cryptarchia-engine/src/config.rs
```diff
@@ -1,28 +1,51 @@
...
-    pub active_slot_coeff: f64,
+    security_param: NonZero<u32>,
+    base_period_length: NonZero<u64>,
 }
 
 impl Config {
     #[must_use]
-    pub fn base_period_length(&self) -> NonZero<u64> {
-        NonZero::new((f64::from(self.security_param.get()) / self.active_slot_coeff).floor() as u64)
+    pub fn new(security_param: NonZero<u32>, active_slot_coefficient: f64) -> Self {
```
ntn-x2: Ok I meant we pass the base period length directly, so this component does not need to know how to derive it from an active slot coefficient, but this also works I guess.

danielSanchezQ: We should not, because its dependant on the security param. If we inyect both is due to make us fail at some point. That is why I isolated the security param as well so it cannot be mutated, but the whole config should be replace instead.

youngjoon-lee (approved):
Awesome. Thank you! Now I will work on removing `pol-dev-mode` features from everywhere, wdyt?

youngjoon-lee:
I am changing the constants and making tests work. Please hold on.

PR #2069, opened 2026-01-28
feat(blend): add redundancy to data message blending

ntn-x2:

We re-introduce the concept of data message redundancy: a value of `0` means a single encapsulation is performed for a data message, while any value greater than `0` involves a higher leadership quota and a full encapsulation cycle for each copy.

Since proof generation is random and PoSel is what determines the intended recipient of a given message copy at a given encapsulation level, I simply wrapped the message encapsulation in both core and edge services without introducing any more complex logic deep down in the stack.

On the side, given the current structure of proof generation, I introduced panicking if any layer of proof generation fails, since failing would "consume" one quota and invalidate few other invariants. We will work on a more reliable system once we enable real proofs and see how they behave. Nevertheless, it should be fairly unlikely for the proof generation step to fail, given we have valid public inputs and we always use valid private inputs as well.

Yes.

Spec: @madxor, implementation: @ntn-x2.

Yes.

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/services/blend/src/core/mod.rs
+++ b/services/blend/src/core/mod.rs
@@ -764,7 +764,6 @@ where
 // Run the main event loop that persists while the node is a core node.
 // This can span across multiple sessions.
 #[expect(clippy::too_many_arguments, reason = "categorize args")]
-#[expect(clippy::cognitive_complexity, reason = "TODO: Address at some point.")]
 async fn run_event_loop<
     NodeId,
     Backend,
--- a/services/blend/src/edge/mod.rs
+++ b/services/blend/src/edge/mod.rs
@@ -329,7 +329,6 @@ where
 ///   handler.
 /// - If the initial secret `PoL` info is not yielded immediately by the `PoL`
 ///   info provider.
-#[expect(clippy::too_many_lines, reason = "TODO: Address at some point.")]
 async fn run<Backend, NodeId, ProofsGenerator, ChainService, PolInfoProvider, RuntimeServiceId>(
     session_stream: UninitializedSessionEventStream<
         impl Stream<Item = MembershipInfo<NodeId>> + Unpin,


services/blend/src/core/mod.rs
```diff
@@ -843,7 +845,20 @@ where
     loop {
         tokio::select! {
             Some(local_data_message) = inbound_relay.next() => {
-                recovery_checkpoint = handle_local_data_message(local_data_message, &mut crypto_processor, &mut message_scheduler, recovery_checkpoint).await;
+                let ServiceMessage::Blend(message_payload) = local_data_message;
```
danielSanchezQ: nitpick: you can nest the match:
```rust
Some(ServiceMessage::Blend(local_data_message))
```

services/blend/src/core/mod.rs
```diff
@@ -843,7 +845,20 @@ where
...
+
+                // We serialize here, outside of the handler function, so that we can serialize only once for all replicas.
+                let serialized_data_message = NetworkMessage::<NetAdapter::BroadcastSettings>::to_bytes(&message_payload).expect("NetworkMessage should be able to be serialized");
+                // If there's at least 2 more nodes besides us, try to replicate the message according to the replication factor.
+                if public_info.session.membership.size() > 2 {
+                    let message_copies = blend_config.data_replication_factor.checked_add(1).unwrap();
```
danielSanchezQ: Why `checked_add` just to `unwrap`?

services/blend/src/core/mod.rs
```diff
@@ -1211,7 +1226,11 @@ enum HandleSessionEventOutput<
 /// encapsulate its payload is performed. If encapsulation is successful, the
 /// message is queued with the Blend scheduler and blended during the next
 /// round.
-async fn handle_local_data_message<
+#[expect(
+    clippy::cognitive_complexity,
+    reason = "TODO: Address this at some point."
+)]
```
danielSanchezQ, line 1224: stfu clippy, all this cognity complexity is data feed for ai 😂

services/blend/src/core/mod.rs
```diff
@@ -1289,9 +1302,16 @@ where
...
+        DecapsulatedMessageType::Completed(fully_decapsulated_message) => {
+            assert!(
+                fully_decapsulated_message.payload_type() == PayloadType::Data,
+                "Locally-generated and fully-decapsulated message should be a data message."
+            );
+            let deserialized_data_message =
+                NetworkMessage::from_bytes(fully_decapsulated_message.payload_body())
+                    .expect("Locally-generated and serialized message should be deserializable.");
+            tracing::debug!(target: LOG_TARGET, "Locally generated data message {deserialized_data_message:?} had all the {} layers addressed to this same node. Propagating only the fully decapsulated message.", blending_tokens.len());
+            ProcessedMessage::from(deserialized_data_message)
```
danielSanchezQ, line 1306: Is this a special case? I don't fully grasp it, would you mind to explain?

danielSanchezQ (approved):

blend/scheduling/src/message_blend/provers/leader/mod.rs
```diff
@@ -108,7 +108,18 @@ fn create_leadership_proof_stream(
...
+            // indices `0` to `2` that belong to the first copy encapsulation, and indices
+            // `3` to `5` that belong to the second copy encapsulation.
+            // In the end, because the expected maximum message quota is `6` (if we take `3`
+            // as the blending operations per message), we end up with two,
+            // fully-encapsulated copies of the same original message, with valid proofs
+            // because within the expected index value.
+            // The logic on how these indices are mapped to each message + encapsulation
+            // layer is out of scope for this component, and will be up to the
+            // message scheduler.
+            let message_release_index = current_index % message_quota;
```
youngjoon-lee, line 122: Asking just to confirm my understanding,
So, is the `message_release_index` always in `[0, 5]` if `public_inputs.leader.message_quota == 6`?

I think the name `message_release_index` is not intuitive, but I can't find anything better. So, I just want to make sure that I understood it correctly.

services/blend/src/core/mod.rs
```diff
@@ -843,7 +845,20 @@ where
...
+
+                // We serialize here, outside of the handler function, so that we can serialize only once for all replicas.
+                let serialized_data_message = NetworkMessage::<NetAdapter::BroadcastSettings>::to_bytes(&message_payload).expect("NetworkMessage should be able to be serialized");
+                // If there's at least 2 more nodes besides us, try to replicate the message according to the replication factor.
+                if public_info.session.membership.size() > 2 {
+                    let message_copies = blend_config.data_replication_factor.checked_add(1).unwrap();
+                    for _ in 0..message_copies {
+                        recovery_checkpoint = handle_serialized_local_data_message(&serialized_data_message, &mut crypto_processor, &mut message_scheduler, recovery_checkpoint).await;
+                    }
+                // Else there is only a single other node (or even maybe just us), so just send it out once.
```
youngjoon-lee: Is this an optimization not described in the spec? I think it's reasonable, but doesn't this conflict with the `message_release_index` calculation in https://github.com/logos-blockchain/logos-blockchain/pull/2069/changes#r2740130987? Or, is it fine? Asking to check if I'm missing anything.

services/blend/src/core/mod.rs
ntn-x2: Because if we don't `checked_add`, in release mode this would be an overflow.

services/blend/src/core/mod.rs
ntn-x2: There is a slim, non-zero chance that after encapsulating a message, ALL layers are addressed to the same node that is performing the encapsulation. This is because we hash the key nullifier, which is derived per index, and the hash modulo is used to index into the membership. So here we take care of that case, checking that if all layers are peeled off by the sender, we just collect the blending tokens and schedule the message for release via gossipsub directly.

This is more likely to happen for testnets with small number of nodes, but if we really enforce the lower limit of 32 nodes, then this is highly unlikely.

blend/scheduling/src/message_blend/provers/leader/mod.rs
ntn-x2: Yes, I also could not find a better naming for it. And yes, the index for PoQ verification is 0-based (meaning hte circuit checks that `index < quota`), so a quota of `6` will verify proofs with index `[0-5]`.

services/blend/src/core/mod.rs
ntn-x2: Yes, that is a good catch. The optimization does indeed affect the index generation logic, given the current code structure. I have reverted the optimization, and will think about a more robust approach to proof generation - probably after testnet launch, that allows pre-generating and buffering proofs, that also plays well with message redundancy.

https://github.com/logos-blockchain/logos-blockchain/pull/2069/commits/5d5add6ae1316c5bcc4140283bd050e8f6aabc57

services/blend/src/core/mod.rs
ntn-x2: https://github.com/logos-blockchain/logos-blockchain/pull/2069/commits/4afcc9c0ace9334c668d8fe6fcc23b8d40e9b79d

PR #2074, opened 2026-01-29
fix(wallet): filter aged notes by leader_pk

youngjoon-lee:

Wallet service has a `GetLeaderAgedNotes`, but it was returning **all** aged notes regardless of `leader_pk`.
If the wallet hold 3 notes, but only two of them are corresponding to the `leader_pk`, but wallet service must return only those two, because the name of API is Get**Leader**AgedNotes.

So far, this issue hasn't been discovered because of `pol-dev-mode`. Once we enable PoL for all integration tests, we will see this issue, because each node in the integration tests holds multiple notes that aren't for leadership.

So, in this PR, I added the `leader_pk` argument for the API.
Also, now the chain-leader service double checks that a note is owned by the leader when checking winning slots.

Yes

spec: @thomaslavaur 
dev: @youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/services/chain/chain-leader/src/leadership.rs
+++ b/services/chain/chain-leader/src/leadership.rs
@@ -126,7 +126,7 @@ impl Leader {
                 utxo.note.value,
                 note_id,
                 *secret_key.as_fr(),
-                self.config.consensus_config.active_slot_coeff,
+                slot_activation_coefficient(),
             ),
             public_inputs.check_winning(utxo.note.value, note_id, *secret_key.as_fr())
         )


davidrusu:
@youngjoon-lee I'm thinking about this now, and I'm wondering if we need this.

Previously we had the leader sk inside the ChainLeader because it had the complicated slot merkle tree that would need to be maintained by the leader.

Now *any* aged note that the leader knows the sk for is eligible for leadership. So, I think the existing behaviour is fine, the only new thing is that we need to fetch the leader sk from KMS

PR #2076, opened 2026-01-29
feat(epoch-state): support computing epoch state after empty epoch

davidrusu:

We have an annoying behavior right now that if the chain-start is not recent, it will halt the chain because we cannot compute the epoch state if there was no blocks produced last epoch.

This PR introduces a change to how we compute epoch state so that we can skip over empty epochs, while also adjusting total stake inference downwards just in case the lottery difficulty is the reason why no blocks were produced.

when skipping over epochs, we log a warning:
```
2026-01-30T08:11:17.700932Z  WARN logos_blockchain_ledger::cryptarchia: EpochState skipping epochs 1..576, adjusting total stake: 101001 -> 1
```

yes

@davidrusu 

yes

This is the implementation in the executable specification: https://github.com/logos-blockchain/logos-blockchain-specs/blob/master/cryptarchia/test_stake_relativization.py#L40-L68

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nodes/node/config.yaml
+++ b/nodes/node/config.yaml
@@ -120,14 +120,3 @@ key_management:
     89b7314f5184bae9eb7fa511098eb321b97b8fbbdec611dda591e0c01e18331d: !Zk '3930000000000000000000000000000000000000000000000000000000000000'
     # Ed25519 signing key (key_id is the public key)
     43974097c741b4831ba37499135a1bbdb449346cc2ca66cd00d3324703a4c881: !Ed25519 af610d87a6a59ac0a533c452b93c578a39616d74bdad7752c0446799af0206ef
-testing_http:
-  backend_settings:
-    address: 127.0.0.1:50897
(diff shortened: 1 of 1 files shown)


nodes/node/config-one-node.yaml
```diff
@@ -2,8 +2,8 @@ network:
   backend:
     swarm:
       host: 0.0.0.0
-      port: 58072
-      node_key: 639c9a4a96b667ceee5fb8e871822f5de870734d55a7b30ff023cf1e84663a35
+      port: 63663
+      node_key: 5cd4636d9c4c36a37a9aeabcaa332c3ec796226af0af48a0b2e70167205af749
```
danielSanchezQ, line 6: I think this config part was not meant to be added (?)

davidrusu: it was, old config was not working with latest circuits

PR #2080, opened 2026-01-30
feat(chain-leader): remove the designated leader_sk

youngjoon-lee:

The chain-leader doesn't need to hold only a single `leader_sk`. Actually, it should be able to use its ALL aged note for leadership.
Previously, the chain-leader service was configured with a single `leader_sk` and ignored all other notes even if they're aged enough. This decreases the winning rate very low unless the `leader_sk` note holds a significant amount of value.

This PR implements the right way:
- chain-leader should fetch its all aged notes from wallet.
- chain-leader should use the secret key of each note for lottery and PoL.
  - chain-leader should be able to access secret keys from KMS. 
    - Ideally, it **shouldn't** access it. Instead, KMS should provide operations that never reveal secret keys. But, I will implement it in a next PR. It requires lots of refactoring. I don't want to make this PR bigger.

To implement this, I introduced additional changes:
- Removed `cryptarchia.leader` from `config.yaml`. Now we don't need any config for the chain-leader service.
- Updated `lb-tests` crate to put note keys to both `wallet.known_keys` and `key_management.keys`.
- Updated `WalletMessage::GetLeaderAgedNotes` to return `KeyId`s along with `Utxo`s, allowing chain-leader to request the corresponding keys from KMS.

If you feel it's not easy to review this PR, I recommend you to review each commit separately. I couldn't split this PR into smaller units.

i believe so

spec: @thomaslavaur 
dev: @youngjoon-lee 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/tests/src/topology/mod.rs
+++ b/tests/src/topology/mod.rs
@@ -383,6 +383,10 @@ pub fn create_kms_configs(
                         blend_conf.core.zk.secret_key_kms_id.clone(),
                         zk_secret_key.clone().into(),
                     ),
+                    (
+                        key_id_for_preload_backend(&consensus_configs[i].known_key.clone().into()),
+                        consensus_configs[i].known_key.clone().into(),
+                    ),
                     // SDP funding secret key - used by wallet for signing SDP transactions
                     (
                         key_id_for_preload_backend(&consensus_configs[i].funding_sk.clone().into()),


nodes/node/config-one-node.yaml
```diff
@@ -529,13 +526,14 @@ storage:
...
+    2e03b2eff5a45478e7e79668d2a146cf2c5c7925bce927f2b1c67f2ab4fc0d26: !Zk 6c645cd4636d9c4c36a37a9aeabcaa3300000000000000000000000000000000
     64249c75c2cb813578b75d05b215fc95f67cea5862fff047228183f22e63460e: !Zk 83c851cf4436e8d2fdac33d56d2b235f66431be97e2a20bf241d431713dc720f
     ed266e6e887b9b97059dc1aa1b7b2e19b934291753c6336a163fe4ebaa28e717: !Zk 7364705cd4636d9c4c36a37a9aeabcaa00000000000000000000000000000000
     86c8519f00178e9eb1fe5f4247e4bed77d4c9f6da2fb10e8a1fdd7ba6bc79fa0: !Ed25519 5cd4636d9c4c36a37a9aeabcaa332c3ec796226af0af48a0b2e70167205af749
 wallet:
   known_keys:
-  - 2e03b2eff5a45478e7e79668d2a146cf2c5c7925bce927f2b1c67f2ab4fc0d26
-  - ed266e6e887b9b97059dc1aa1b7b2e19b934291753c6336a163fe4ebaa28e717
+    2e03b2eff5a45478e7e79668d2a146cf2c5c7925bce927f2b1c67f2ab4fc0d26: "2e03b2eff5a45478e7e79668d2a146cf2c5c7925bce927f2b1c67f2ab4fc0d26"
+    ed266e6e887b9b97059dc1aa1b7b2e19b934291753c6336a163fe4ebaa28e717: "ed266e6e887b9b97059dc1aa1b7b2e19b934291753c6336a163fe4ebaa28e717"
```
davidrusu, line 536: The duplication here is pretty ugly. Hopefully we can drop the concept of KeyId in KMS soon, and just refer to key's by their PK directly.

danielSanchezQ: Yeah, we use the pk as Id anyway.

kms/keys/src/operators/zk/leader.rs
```diff
@@ -0,0 +1,47 @@
+// TODO: Make this secure by embedding actual logic
```
danielSanchezQ, line 1: Aaahg, cycle deps. Maybe we can extract a key-operators crate.

danielSanchezQ (approved):

kms/keys/src/operators/zk/leader.rs
youngjoon-lee: Yes. Actually I tried it but it also cause another cycle deps :), but I think we can resolve it easily.
I also have another idea. I will discuss with you soon and implement it

PR #2085, opened 2026-01-31
feat(leadership): Use kms operators in leader

danielSanchezQ:

This PR makes leader to use kms operators without exfiltrating the leader sk.

~~Yes, it uses some kind of callback approach. Which is not perfect but is better than directly leaking the keys.~~
I refactored `operators` out of `kms-keys` which removes the circular dep. Also some things were moved around (for example the *winning* check). Overall is almost the same and code got a bit simpler. 
One drawback is that ZkKeys now has leaking methods public, because operators need to access them.

@youngjoon-lee @thomaslavaur 

Not completely related, so yes.

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7226,6 +7226,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "logos-blockchain-zone-sdk"
+version = "0.1.0"
+dependencies = [
+ "futures",
+ "logos-blockchain-common-http-client 0.1.0",
+ "logos-blockchain-core 0.1.0",
+ "logos-blockchain-key-management-system-service 0.1.0",
+ "reqwest 0.12.28",
+ "thiserror 2.0.18",
+ "tokio",
+ "tracing",
+]
+
+[[package]]
 name = "loki-api"
 version = "0.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,10 @@ readme      = "README.md"
 repository  = "https://github.com/logos-blockchain/logos-blockchain"
 version     = "0.1.0"
 
+[profile.release]
+codegen-units = 1     # (Much) slower build time, much better performances
+lto           = "fat" # Same as above
+
 [workspace]
 # Make sure any excluded crates are still clipped in our CI. Also, make sure they use the same Clippy and rustc lints we specify in this file down below.
 members = [
@@ -67,6 +71,7 @@ members = [
   "zk/proofs/pol",
   "zk/proofs/poq",
   "zk/proofs/zksign",
+  "zone-sdk",
 ]
 resolver = "2"
 
@@ -127,6 +132,7 @@ lb-wallet-http-client              = { default-features = false, package = "logo
 lb-wallet-service                  = { default-features = false, package = "logos-blockchain-wallet-service", path = "./services/wallet" }
 lb-witness-generator               = { default-features = false, package = "logos-blockchain-witness-generator", path = "./zk/circuits/witness-generator" }
 lb-zksign                          = { default-features = false, package = "logos-blockchain-zksign", path = "./zk/proofs/zksign" }
+lb-zone-sdk                        = { default-features = false, package = "logos-blockchain-zone-sdk", path = "./zone-sdk" }
 
 # External
 blake2            = { default-features = false, version = "0.10" }
--- a/zone-sdk/Cargo.toml
+++ b/zone-sdk/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+categories  = { workspace = true }
+description = "Logos blockchain Zone SDK - client library for building zone sequencers and indexers"
+edition     = { workspace = true }
+keywords    = { workspace = true }
+license     = { workspace = true }
+name        = "logos-blockchain-zone-sdk"
+publish     = false
+readme      = { workspace = true }
+repository  = { workspace = true }
+version     = { workspace = true }
+
+[lints]
+workspace = true
+
+[dependencies]
+futures                          = { default-features = false, version = "0.3" }
+lb-common-http-client            = { workspace = true }
+lb-core                          = { workspace = true }
+lb-key-management-system-service = { workspace = true }
+reqwest                          = { workspace = true }
+thiserror                        = { workspace = true }
+tokio                            = { default-features = false, features = ["macros", "rt", "sync", "time"], version = "1" }
+tracing                          = { workspace = true }
--- a/zone-sdk/src/indexer.rs
+++ b/zone-sdk/src/indexer.rs
@@ -1,0 +1,74 @@
+use futures::{Stream, StreamExt as _};
+use lb_common_http_client::{BasicAuthCredentials, CommonHttpClient};
+use lb_core::mantle::ops::{Op, channel::ChannelId};
+use reqwest::Url;
+use tracing::warn;
+
+use crate::ZoneBlock;
+
+/// Zone indexer — follows finalized (LIB) blocks and yields zone blocks
+/// for a given channel.
+pub struct ZoneIndexer {
+    channel_id: ChannelId,
+    node_url: Url,
+    http_client: CommonHttpClient,
+}
+
+impl ZoneIndexer {
+    #[must_use]
+    pub fn new(channel_id: ChannelId, node_url: Url, auth: Option<BasicAuthCredentials>) -> Self {
+        Self {
+            channel_id,
+            node_url,
+            http_client: CommonHttpClient::new(auth),
+        }
+    }
+
+    /// Follow finalized zone blocks.
+    pub async fn follow(
+        &self,
+    ) -> Result<impl Stream<Item = ZoneBlock> + '_, lb_common_http_client::Error> {
+        let lib_stream = self
+            .http_client
+            .get_lib_stream(self.node_url.clone())
+            .await?;
+
+        let channel_id = self.channel_id;
+        let stream = lib_stream.filter_map(move |block_info| {
+            let http_client = self.http_client.clone();
+            let node_url = self.node_url.clone();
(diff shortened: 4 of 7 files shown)


kms/keys/src/operators/zk/leader.rs
```diff
@@ -1,47 +1,76 @@
...
+            tracing::error!("Failed to send result via channel");
         }
+        Ok(())
+    }
+}
+
+pub struct BuildWithLeaderKey<F, R> {
+    result_channel: tokio::sync::oneshot::Sender<R>,
+    f: F,
+}
```
youngjoon-lee: If this operator is almost the same as the other operator, we can have only one and use it for both use cases.

youngjoon-lee (approved):
Thank you! As you mentioned in the PR description, this approach is not perfect yet, because the secret key is still leaked from KMS. The private input (`LeaderPrivate`) built from KMS contains the secret key itself. If we want to do it more perfectly, KMS should build a `LeaderPrivate` internally and build/return a PoL as a final result. But I agree that this approach is better than the previous. We can improve it gradually 👍

kms/keys/src/operators/zk/leader.rs
```diff
@@ -1,47 +1,76 @@
...
+use std::fmt::{Debug, Formatter};
 
-    use tokio::sync::oneshot;
+use crate::keys::{UnsecuredZkKey, ZkKey, errors::KeyError, secured_key::SecureKeyOperator};
 
-    use crate::keys::{
-        UnsecuredZkKey, ZkKey,
-        secured_key::{SecureKeyOperator, SecuredKey},
-    };
+pub struct CheckConditionWithLeaderKey<F> {
```
ntn-x2:
```suggestion
pub struct CheckConditionWithLeaderKey<CheckFn> {
```

kms/keys/src/operators/zk/leader.rs
```diff
@@ -1,47 +1,76 @@
...
+    result_channel: tokio::sync::oneshot::Sender<bool>,
```
ntn-x2: Should the channel return `Result` where `Result` is the result type of `F`?

kms/keys/src/operators/zk/leader.rs
```diff
@@ -1,47 +1,76 @@
...
-                .is_err()
-            {
-                tracing::error!("Failed to send result via channel");
-            }
-            Ok(())
+#[cfg(feature = "unsafe")]
+#[async_trait::async_trait]
+impl<F> SecureKeyOperator for CheckConditionWithLeaderKey<F>
+where
+    F: Fn(&UnsecuredZkKey) -> bool + Send + Sync + 'static,
```
ntn-x2: How is this approach better than accessing the unsecured key directly? Is it worth introducing a new operator?

kms/keys/src/keys/zk/mod.rs
```diff
@@ -60,6 +60,12 @@ impl ZkKey {
     pub fn into_unsecured(self) -> UnsecuredZkKey {
         self.0.clone()
     }
+
+    #[cfg(feature = "unsafe")]
+    #[must_use]
+    pub const fn as_unsecured(&self) -> &UnsecuredZkKey {
```
ntn-x2, line 67: If this is used only internally, we can keep it `pub(crate)` for now, I'd say.

services/chain/chain-leader/src/kms.rs
```diff
@@ -28,12 +51,57 @@ where
...
+        let (output_tx, output_rx) = oneshot::channel();
+        // clone to send
+        let utxo = *utxo;
+        let public_inputs = *public_inputs;
+        let () = self
+            .execute(
+                key_id,
+                KeyOperators::Zk(Box::new(CheckConditionWithLeaderKey::new(
+                    output_tx,
+                    move |key: &UnsecuredZkKey| check_winning(utxo, public_inputs, key),
```
ntn-x2: Why don't we run this computation inside the KMS, so we don't need to expose the `UnsecuredZkKey`?

ntn-x2 (commented):
I personally think this approach might not be worth the gains, since we keep exposing the secret key anyway. We should rather move the computation inside the KMS. I think that would lead to a dependency cycle, which we can break by moving the logic itself out of `lb-core` and into a lower-level crate, kinda what we do for Blend PoQs, and depend on that crate from kms-keys crate, while lb-core would keep depending on kms-keys. I personally think that would then make the changes worth being introduced.

services/chain/chain-leader/src/kms.rs
danielSanchezQ: It is because of cycle deps.

core/src/proofs/leader_proof.rs
```diff
@@ -134,6 +135,14 @@ pub struct LeaderPublic {
...
 }
 
+/// Check if the given note is owned by the leader and wins the lottery with
+/// the given public inputs.
+#[must_use]
+pub fn check_winning(utxo: Utxo, public_inputs: LeaderPublic, secret_key: &ZkKey) -> bool {
+    utxo.note.pk == secret_key.to_public_key()
+        && public_inputs.check_winning(utxo.note.value, utxo.id().0, *secret_key.as_fr())
+}
+
```
danielSanchezQ, line 150: I moved this here to avoid deps issues, but maybe some other place would be better (?) cc @youngjoon-lee

kms/operators/src/zk/leader.rs
```diff
@@ -0,0 +1,126 @@
...
+            result_channel,
+            utxo,
+            public_inputs,
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl SecureKeyOperator for CheckLotteryWinning {
+    type Key = ZkKey;
```
youngjoon-lee, line 42: I have a foundational question. 
I have been thinking about making ZkKey un-clonable. Then, even if this operator accepts any type of closure, it’ll be safe because the closure can’t clone the key and send it via channel or so. I think we also need to remove Serialize. 
I think this way is more flexible, and we don’t need to move any core logic from other crates to here. I’ve been feeling that it is not natural to have consensus/blend-specific logic in the kms crate (even though we are modularizing logics)
What do you think about this idea?

core/src/proofs/leader_proof.rs
youngjoon-lee: i think it’s fine

kms/operators/src/zk/leader.rs
danielSanchezQ: We are undergoing conversations with @ntn-x2 as well. What you say its true, serialize and clone should not be needed. But the other part is not accurate. The intention of having the operators as part of kms-keys was to keep access to the keys private, that way we ensured that no one would access the inner secret by mistak. We are losing that here (because we had to extract to a third, operators, crate). This makes that we have to take care of what we do. 
But at some point we need to check with care how to structure so we can keep the operators operating with the keys together as well. 
For now I think this is as close as we can get to the idea of how kms should work, we just need to make sure, manually, that we do not leak the keys. So the only place where the `unsafe` feature should be present is inside the kms itself.

youngjoon-lee: hm. no. it’s not enough.
we still need to expose some methods for ZkKey, so that other crates can write closures. 
Never mind this for now.

danielSanchezQ: > hm. no. it’s not enough. we still need to expose some methods for ZkKey, so that other crates can write closures. Never mind this for now.

There is no closures anymore. I removed them and everything is done in the operator itself. So its only controlled in a kms related crate.

youngjoon-lee: > The intention of having the operators as part of kms-keys was to keep access to the keys private, that way we ensured that no one would access the inner secret by mistak

Yes, right. Thank you for explaining. Closing this thread. I will review new changes in detail tomorrow

kms/keys/src/keys/zk/mod.rs
danielSanchezQ: Yes, we will have to revisit this later. I'm resolving the convo as per what we talked offline.

blend/message/Cargo.toml
```diff
@@ -21,13 +21,14 @@ lb-blend-proofs               = { workspace = true }
...
-tracing                       = { workspace = true }
-zeroize                       = { workspace = true }
+
+lb-utils        = { features = ["rng"], workspace = true }
+serde           = { workspace = true }
+serde-big-array = { workspace = true }
+serde_with      = { default-features = false, features = ["alloc"], version = "3" }
+thiserror       = { default-features = false, version = "1.0.69" }
+tracing         = { workspace = true }
+zeroize         = { workspace = true }
```
youngjoon-lee, line 31: i think we can revert this file

kms/operators/src/zk/leader.rs
```diff
@@ -0,0 +1,131 @@
...
+            leader_pk,
+        } = *self;
+        if result_channel
+            .send(LeaderPrivate::new(
+                leader_public,
+                utxo,
+                &aged_path,
+                &latest_path,
+                *key.as_fr(),
+                &leader_pk,
```
youngjoon-lee, line 123: NOTE:
As I mentioned in https://github.com/logos-blockchain/logos-blockchain/pull/2085#pullrequestreview-3737049798, this leaks the note secret key because `LeaderPrivate` contains it as a `Fr`. But, I understand it's very hard to change this operator right now, because `LeaderPrivate` is sent across chain->blend services. So, I think it's good to go with this approach for now.

danielSanchezQ: Yes, it leaks, we can't do anything here. The only thing is that the secret key field is private in `LeaderPrivate`, so after building it it should not be accessible by code (so we prevent missuse). A further option is to move the creation of the different proofs to the kms and return a task from it (as we did in the voucher). But we can do that little by little later.

PR #2088, opened 2026-02-01
feat(lb-zone-sdk): simple mvp e2e of lb-zone-sdk (sequencer, indexer, state)

pradovic:

This PR introduces a minimal end-to-end Zone SDK (sequencer + indexer + in-memory state) to enable simple zone publishing and reading over channel inscriptions.

Sequencer design:
- Actor-based background loop owns all mutable state.
- Fire & forget publishing: publish_block returns immediately even if post_tx fails, and schedules it for resubmittion.
- Pending txs are resubmitted every 30s until finalized in LIB.
- LIB stream auto-reconnect prevents silent death.

Indexer:
- Follows LIB blocks and extracts channel inscriptions.
- Stateless, simple stream wrapper.

Scope / Non-Goals (v0):
- No persistence / checkpoints / restart / recovery .
- No “safe” tracking (LIB only).
- Resubmit optimizations intentionally deferred

Future:
- Track “processed” / “safe” status (not only LIB) to reduce pending set and resubmit noise.
- Persistence/restart: expose a checkpoint API so callers can persist pending state and restore it on init.
- Resilience: timeouts/backoff/retry policy; handle stream expiration and edge cases more robustly.
- Improve resubmit efficiency if it becomes a bottleneck.

Yes

@danielSanchezQ , @davidrusu , @pradovic 

This is an MVP version that shows e2e flow but does not handle unfinalized blocks, persistence, durability etc... This will be the scope of the following PRs

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/zone-sdk/src/sequencer.rs
+++ b/zone-sdk/src/sequencer.rs
@@ -18,9 +18,29 @@ use tracing::{debug, info, warn};
 
 use crate::state::State;
 
-const RESUBMIT_INTERVAL: Duration = Duration::from_secs(30);
-const RECONNECT_DELAY: Duration = Duration::from_secs(5);
-const PUBLISH_CHANNEL_CAPACITY: usize = 256;
+const DEFAULT_RESUBMIT_INTERVAL: Duration = Duration::from_secs(30);
+const DEFAULT_RECONNECT_DELAY: Duration = Duration::from_secs(5);
+const DEFAULT_PUBLISH_CHANNEL_CAPACITY: usize = 256;
+
+/// Configuration for the zone sequencer.
+pub struct SequencerConfig {
+    /// How often to resubmit pending transactions to the mempool.
+    pub resubmit_interval: Duration,
+    /// Delay before retrying a failed LIB stream connection.
+    pub reconnect_delay: Duration,
+    /// Capacity of the internal publish request channel.
+    pub publish_channel_capacity: usize,
+}
+
+impl Default for SequencerConfig {
+    fn default() -> Self {
+        Self {
+            resubmit_interval: DEFAULT_RESUBMIT_INTERVAL,
+            reconnect_delay: DEFAULT_RECONNECT_DELAY,
+            publish_channel_capacity: DEFAULT_PUBLISH_CHANNEL_CAPACITY,
+        }
+    }
+}
 
 /// Result of a `publish_block` call.
 pub struct PublishResult {
@@ -62,7 +82,7 @@ pub struct ZoneSequencer {
 }
 
 impl ZoneSequencer {
-    /// Initialize a new sequencer.
+    /// Initialize a new sequencer with default configuration.
     ///
     /// Spawns a background actor that owns all mutable state, connects to
     /// the node's LIB stream, tracks finalized transactions, and resubmits
@@ -74,8 +94,26 @@ impl ZoneSequencer {
         node_url: Url,
         auth: Option<BasicAuthCredentials>,
     ) -> Self {
+        Self::init_with_config(
+            channel_id,
+            signing_key,
+            node_url,
+            auth,
+            SequencerConfig::default(),
+        )
+    }
+
+    /// Initialize a new sequencer with custom configuration.
+    #[must_use]
+    pub fn init_with_config(
+        channel_id: ChannelId,
+        signing_key: Ed25519Key,
+        node_url: Url,
+        auth: Option<BasicAuthCredentials>,
+        config: SequencerConfig,
+    ) -> Self {
         let http_client = CommonHttpClient::new(auth);
-        let (request_tx, request_rx) = mpsc::channel(PUBLISH_CHANNEL_CAPACITY);
+        let (request_tx, request_rx) = mpsc::channel(config.publish_channel_capacity);
 
         tokio::spawn(run_loop(
             request_rx,
@@ -83,6 +121,7 @@ impl ZoneSequencer {
             signing_key,
             node_url.clone(),
             http_client.clone(),
+            config,
         ));
 
         Self {
@@ -92,13 +131,13 @@ impl ZoneSequencer {
         }
     }
 
-    /// Publish a zone block.
+    /// Publish an inscription to the zone's channel.
     ///
     /// Sends the data to the background actor which creates the inscription
     /// transaction and tracks it as pending. Then posts the transaction to
     /// the mempool inline. If the post fails the transaction remains pending
     /// and will be retried by the periodic resubmit.
-    pub async fn publish_block(&self, data: Vec<u8>) -> Result<PublishResult, Error> {
+    pub async fn publish_inscription(&self, data: Vec<u8>) -> Result<PublishResult, Error> {
         let (reply_tx, reply_rx) = oneshot::channel();
         let request = PublishRequest {
             data,
@@ -136,10 +175,11 @@ async fn run_loop(
     signing_key: Ed25519Key,
     node_url: Url,
     http_client: CommonHttpClient,
+    config: SequencerConfig,
 ) {
     let mut state = State::new();
     let mut last_msg_id = MsgId::root();
-    let mut resubmit_interval = tokio::time::interval(RESUBMIT_INTERVAL);
+    let mut resubmit_interval = tokio::time::interval(config.resubmit_interval);
     let mut resubmit_active = false;
     let mut in_flight: FuturesUnordered<BoxFuture<'static, InFlight>> = FuturesUnordered::new();
 
@@ -147,8 +187,11 @@ async fn run_loop(
         let lib_stream = match http_client.get_lib_stream(node_url.clone()).await {
             Ok(stream) => stream,
             Err(e) => {
-                warn!("Failed to connect to LIB stream: {e}, retrying in {RECONNECT_DELAY:?}");
-                tokio::time::sleep(RECONNECT_DELAY).await;
+                warn!(
+                    "Failed to connect to LIB stream: {e}, retrying in {:?}",
+                    config.reconnect_delay
+                );
+                tokio::time::sleep(config.reconnect_delay).await;
                 continue;
             }
         };


zone-sdk/src/sequencer.rs
```diff
@@ -0,0 +1,311 @@
...
+
+fn create_inscribe_tx(
+    channel_id: ChannelId,
+    signing_key: &Ed25519Key,
+    data: Vec<u8>,
+    parent: MsgId,
+) -> (SignedMantleTx, MsgId) {
+    let verifying_key_bytes = signing_key.public_key().to_bytes();
+    let verifying_key =
+        Ed25519PublicKey::from_bytes(&verifying_key_bytes).expect("valid ed25519 public key");
```
davidrusu: This is the only way to get a verifying key?

Maybe we should have a `.to_ed25519()` and `.to_zk()` on the the KMS `PublicKey` enum?

zone-sdk/src/sequencer.rs
```diff
@@ -0,0 +1,311 @@
...
+        storage_gas_price: 0,
+        execution_gas_price: 0,
+    };
+
+    let tx_hash = inscribe_tx.hash();
+    let signature_bytes = signing_key
+        .sign_payload(tx_hash.as_signing_bytes().as_ref())
+        .to_bytes();
+    let signature =
+        lb_key_management_system_service::keys::Ed25519Signature::from_bytes(&signature_bytes);
```
davidrusu: These conversions through `bytes()` are pretty ugly, but it'll do for now

zone-sdk/src/sequencer.rs
```diff
@@ -0,0 +1,311 @@
...
+        }
+    }
+
+    /// Publish a zone block.
+    ///
+    /// Sends the data to the background actor which creates the inscription
+    /// transaction and tracks it as pending. Then posts the transaction to
+    /// the mempool inline. If the post fails the transaction remains pending
+    /// and will be retried by the periodic resubmit.
+    pub async fn publish_block(&self, data: Vec<u8>) -> Result<PublishResult, Error> {
```
davidrusu: Thinking about the naming of this again.

I think we should call this method `publish_inscription(block)`. In the future we can have a second API for `publish_blob(block)`.

"inscriptions" are pretty core to our design so, exposing the name to the user is a good IMO.

Not pressing though, we can change in the next PR.

zone-sdk/src/sequencer.rs
```diff
@@ -0,0 +1,311 @@
...
+            tokio::select! {
+                Some(request) = request_rx.recv() => {
+                    let (signed_tx, new_msg_id) =
+                        create_inscribe_tx(channel_id, &signing_key, request.data, last_msg_id);
+                    let tx_hash = signed_tx.mantle_tx.hash();
+
+                    state.submit_pending(tx_hash, signed_tx.clone());
+                    last_msg_id = new_msg_id;
+
+                    drop(request.reply.send((signed_tx, tx_hash)));
```
davidrusu, line 211: just wondering, `drop` here is used in place of the `let _ =` pattern?

zone-sdk/src/sequencer.rs
```diff
@@ -0,0 +1,311 @@
...
+    let verifying_key_bytes = signing_key.public_key().to_bytes();
+    let verifying_key =
+        Ed25519PublicKey::from_bytes(&verifying_key_bytes).expect("valid ed25519 public key");
+
+    let inscribe_op = InscriptionOp {
+        channel_id,
+        inscription: data,
+        parent,
+        signer: verifying_key,
+    };
```
davidrusu, line 325: nit: can rename `data` to `inscription`, `verifying_key` to `signer`.

zone-sdk/src/state.rs
```diff
@@ -0,0 +1,152 @@
...
+
+use lb_core::mantle::{SignedMantleTx, Transaction as _, tx::TxHash};
+
+/// Pure in-memory state tracker for zone transactions.
+///
+/// Tracks pending transactions (submitted but not yet finalized).
+/// When a LIB block is processed, any pending txs found in that block
+/// are removed from pending and returned as newly finalized.
+pub struct State {
+    pending: HashMap<TxHash, SignedMantleTx>,
```
davidrusu, line 11: I'm wondering if a more natural modelling would be as a `VecDeque`? since pending txs would finalize in a the channel parent enforced ordering.

davidrusu: this is fine though, I can't imagine this map growing too large

zone-sdk/src/sequencer.rs
pradovic: Yes, it is a clippy suggestion. We have a lint enabled that asks for it.

zone-sdk/src/state.rs
pradovic: Good point. I considered `VecDeque/IndexMap` initially, since within a channel the txs form a parent→child chain and typically finalize in that order.

I kept `HashMap<TxHash, SignedMantleTx>` for v0 mainly for operational simplicity: finalization/removal is driven by “tx hash appears in LIB block”, and we also need fast lookup/removal by hash (and may see gaps/retries/reorg-ish situations before we add fuller tracking).

We can definitely switch to a deque (or keep both: `VecDeque` for order + `HashMap` for O(1) lookup) once we decide on the resubmission strategy (e.g. parent-first iteration, bounded window, etc.). Even if we store them in order, it doesn’t by itself solve resubmission; we still need a retry/backoff policy for parent→child dependencies, so I kept the simplest representation for now.

davidrusu: right, lets leave it as is for now, no need to think too hard about this right now, can optimize it later once the rest of the features are ready.

zone-sdk/src/sequencer.rs
pradovic: Is this maybe better for now:
https://github.com/logos-blockchain/logos-blockchain/pull/2088/changes/9bdcd9e7ed36ae015afcab9148c87fea7eeb6ad1. 

I can also add method to kms enum as well.

zone-sdk/src/sequencer.rs
pradovic: Same as the above: https://github.com/logos-blockchain/logos-blockchain/commit/9bdcd9e7ed36ae015afcab9148c87fea7eeb6ad1. Does it  look better?

zone-sdk/src/sequencer.rs
pradovic: Makes sense, changing it right away.

zone-sdk/src/sequencer.rs
```diff
@@ -0,0 +1,305 @@
...
+};
+use lb_key_management_system_service::keys::{Ed25519Key, ZkKey};
+use reqwest::Url;
+use tokio::sync::{mpsc, oneshot};
+use tracing::{debug, info, warn};
+
+use crate::state::State;
+
+const RESUBMIT_INTERVAL: Duration = Duration::from_secs(30);
+const RECONNECT_DELAY: Duration = Duration::from_secs(5);
```
danielSanchezQ: This should rather be a config (?)

zone-sdk/src/sequencer.rs
```diff
@@ -0,0 +1,305 @@
...
+    node_url: Url,
+    http_client: CommonHttpClient,
+) {
+    let mut state = State::new();
+    let mut last_msg_id = MsgId::root();
+    let mut resubmit_interval = tokio::time::interval(RESUBMIT_INTERVAL);
+    let mut resubmit_active = false;
+    let mut in_flight: FuturesUnordered<BoxFuture<'static, InFlight>> = FuturesUnordered::new();
+
+    loop {
```
danielSanchezQ, line 186: This should be a single outer loop to get the stream. Why nested them?

zone-sdk/src/state.rs
```diff
@@ -0,0 +1,152 @@
...
+}
```
danielSanchezQ, line 12: I think this will not make it (or will turn everything more complicated) when handling re-orgs.

zone-sdk/src/sequencer.rs
```diff
@@ -0,0 +1,305 @@
...
+    pub fn init(
+        channel_id: ChannelId,
+        signing_key: Ed25519Key,
+        node_url: Url,
+        auth: Option<BasicAuthCredentials>,
+    ) -> Self {
+        let http_client = CommonHttpClient::new(auth);
+        let (request_tx, request_rx) = mpsc::channel(PUBLISH_CHANNEL_CAPACITY);
+
+        tokio::spawn(run_loop(
```
danielSanchezQ, line 118: This is probably a bad signal, There should not be need to spawn tasks. For the future we should consider using `Sink`-`Stream` and keep this things connection agnostic if possible.

pradovic: Good point. I’m ok to keep this as-is for v0 (internal task) since we’ll soon need to multiplex streams + timers anyway. Let’s treat Sink/Stream + “no spawn” as a follow-up once safe tracking/persistence is in. Does that makes sense?

zone-sdk/src/sequencer.rs
pradovic: Should we expose this to the user? User might do weird things if they don't understand the implication how these intervals are used in the code and what the implications are. I am not sure, just thinking out loud

pradovic: Actually it makes sense, I would just add some validation and sane defaults. No worries, sorry for the noise

pradovic: https://github.com/logos-blockchain/logos-blockchain/pull/2088/changes/b91bf7380da1ed534620ba5b8686c8b2694a9081

zone-sdk/src/state.rs
pradovic: Yes, I would improve the state once the re-orgs stuff is implemented, as it will be influenced by implementation

zone-sdk/src/sequencer.rs
pradovic: This is a simple reconnect (when the stream drops). It will be improved once we also track processing events, as then we will have a correct flow to reason about.

PR #2089, opened 2026-02-02
feat(wallet): derive leader vouchers from master key in KMS

youngjoon-lee:

So far, we had been holding voucher secrets in memory (in the `lb-wallet` crate).
In this PR, we derive voucher secrets using KMS and never store them anywhere. Of course, KMS should do more sophisticated jobs rather than returning voucher secrets, but I will implement it as a separate PR to not make this PR so big. 

Flow:
1. Wallet service loads the voucher indices and the next index from the state recovery. 
2. These information is kept in memory in the `lb-wallet` crate, so that they can be added/removed when new blocks are processed. Whenever they are added/removed, they are written in the state of the wallet service. 
3. Whenever we need a new voucher, the wallet service requests it for KMS, by passing the master key ID and the next index. The master key must be one of keys in KMS.

Note: 
The concept of index and master key is not defined in the spec. It’s fine even if each node uses different derivations. That’s why I implemented the logic in the wallet service (KMS later), not in the `lb-wallet`. I abstracted this concept (as `VoucherId`) so that the lb-wallet does not need to be aware of it. You may understand this easily during the review. 

I believe so

spec: @thomaslavaur 
dev: @youngjoon-lee 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/.github/release/release-content.md
+++ b/.github/release/release-content.md
@@ -16,11 +16,31 @@ For the former, please reach out to the Logos Blockchain team on [Discord](https
 For the latter, you can download the example config from this release and tweak it to your needs.
 Please check the docs for info on what each field means.
 
+## Devnet setup
+
+If you wish to join the devnet at https://devnet.blockchain.logos.co, you can automatically generate and download your node configuration using the following command:
+
+```bash
+curl -X POST -L --location-trusted https://devnet.blockchain.logos.co/node/0/cfgsync/config/custom-node \
+     -u "username:password" \
+     -H "Content-Type: application/json" \
+     -d '{
+            "ip": "192.168.6.7",
+            "identifier": "marcins-anonymous-node",
+            "network_port": 3000,
+            "blend_port": 4000,
+            "api_port": 8080
+         }' \
+     -o my_logos_node_config.yaml
+```
+
 ## Run the binary
 
 After obtaining the deployment file for the network you want to join and tweaking the node config file to fit your needs, including specifying the list of bootnodes for the network you are joining, you can run the node binary.
 
 For example: `logos-blockchain-node-macos-aarch64-0.0.1 --deployment deployment.yaml node-config.yaml`. See the repo's `README.md` for more info.
+
+To verify that your node is running correctly and connected, visit http://localhost:{api_port_in_user_config}/cryptarchia/info. The slot and height should both be constently increasing.
 
 ## Checklist
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,10 +9,9 @@ repository  = "https://github.com/logos-blockchain/logos-blockchain"
 version     = "0.1.0"
 
 [profile.release]
-codegen-units = 1       # (Much) slower build time, much better performances
-lto           = "fat"   # Same as above
-panic         = "abort" # Remove panic handler to reduce binary size
-strip         = true    # Remove symbols to reduce binary size
+codegen-units = 1     # (Much) slower build time, much better performances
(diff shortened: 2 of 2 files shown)


services/wallet/src/lib.rs
```diff
@@ -731,36 +749,58 @@ where
...
+        kms: &KmsServiceApi<Kms, RuntimeServiceId>,
         resp_tx: oneshot::Sender<VoucherCm>,
     ) {
-        let mut voucher_secret_bytes = [0u8; 31];
-        OsRng.fill_bytes(&mut voucher_secret_bytes);
-        let voucher_secret = fr_from_bytes(&voucher_secret_bytes)
-            .expect("voucher secret bytes must be a valid Fr")
-            .into();
-        let voucher_cm = VoucherCm::from_secret(voucher_secret);
+        let index = state.get_and_inc_next_new_voucher_index();
```
thomaslavaur, line 759: TODO: when KMS is also generating the secret key (not loaded from config), voucher secret should be generated in the same way instead of being derived from a master key

wallet/src/lib.rs
```diff
@@ -377,15 +366,23 @@ mod tests {
         TxHash::from(BigUint::from(v))
     }
 
-    fn voucher_secret(i: u64) -> VoucherSecret {
-        VoucherSecret(BigUint::from(i).into())
+    fn voucher(key_id: u64, idx: u64) -> (VoucherCm, VoucherNullifier) {
```
thomaslavaur, line 369: ditto

services/wallet/src/lib.rs
danielSanchezQ: What do you mean? Voucher generated from randomness?

kms/operators/src/zk/voucher.rs
```diff
@@ -1,23 +1,37 @@
 use lb_groth16::Fr;
-use lb_poseidon2::{Digest as _, Poseidon2Bn254Hasher as ZkHasher, ZkHash};
+use lb_poseidon2::{Digest as _, Poseidon2Bn254Hasher as ZkHasher};
 use tokio::sync::oneshot;
 use tracing::error;
 
 use crate::keys::{ZkKey, errors::KeyError, secured_key::SecureKeyOperator};
 
-pub struct VoucherOperator {
+/// Derives/returns a voucher secret from key and index.
```
danielSanchezQ, line 9: We can dump the voucher creation here after we merge the #2085 PR. I can do it.

youngjoon-lee: Yes, I approved #2085. I think it's better to merge this PR as it is, and improve the KMS part as a separate PR, so that we can unblock next tasks. In this PR, we focus on only how the wallet manages indices.

danielSanchezQ:
@youngjoon-lee I rebased the branch to remove the conflicts. Should be good to go now.

services/wallet/src/lib.rs
thomaslavaur: yes. Here voucher is derived from a master key and index. It works but I think it would be a better practice to not link voucher between each other (in case the master key leaks).

thomaslavaur: so the generation of a secret voucher should be exactly the same as generating a master secret key. But it's also ok to do it this way. The process of generating a random secret voucher isn't specified and we are free to chose how to do it

youngjoon-lee: yeah, let's make this more secure later.

PR #2093, opened 2026-02-02
fix(tests): enforce deployment config

ntn-x2:

We were generating a default e2e deployment config but we were not passing it to the spawned binaries, which ended up using the default config for the testnet, which is MOSTLY similar to the e2e one, but it was ignoring the `CONSENSUS_SLOT_TIME` env variable, which is now rendered unnecessary since we use `1` (which was the value we were, without knowing, using already previously) as the default value in case the variable is not specified. This PR fixes that.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nodes/node/binary/src/config/blend/deployment.rs
+++ b/nodes/node/binary/src/config/blend/deployment.rs
@@ -62,7 +62,7 @@ fn mainnet_settings() -> Settings {
             data_replication_factor: 0,
         },
         core: CoreSettings {
-            minimum_messages_coefficient: 3.try_into().unwrap(),
+            minimum_messages_coefficient: 1.try_into().unwrap(),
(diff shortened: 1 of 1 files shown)


nodes/node/binary/src/config/blend/deployment.rs
```diff
@@ -62,7 +62,7 @@ fn mainnet_settings() -> Settings {
             data_replication_factor: 0,
         },
         core: CoreSettings {
-            minimum_messages_coefficient: 3.try_into().unwrap(),
+            minimum_messages_coefficient: 1.try_into().unwrap(),
```
ntn-x2, line 65: This is just to match the (working) deployment we have been using so far.

hansieodendaal (approved):
LGTM

PR #2099, opened 2026-02-03
fix(chain-leader): remove assertion that proposed block always becomes honest chain tip

youngjoon-lee:

Closes https://github.com/logos-blockchain/logos-blockchain/issues/2096

As described in the issue, the chain-leader was applying the block proposal to the chain first before publishing it to the network. In this step, it had `assert!` to check that the block is always added to the honest chain. But, this assumption is not always true. Although the chain-leader picks the current honest tip as the parent for a new block, the chain can be reorged while the chain-leader builds/proves the block. Then, the block will be added to the fork. We should accept this case.

In this PR,
- Removed `assert!`.
- Instead of having the same logic in chain-leader, I introduced `ChainNetworkServiceApi` because the chain-network service already has a function to apply a block to the chain and reconcile mempool accordingly. This API is now called by chain-leader.

NOTE:
Yeah, this PR doesn't have any test. As I've been saying, I will write tests for all `chain-*` services by little refactoring after having everything work first. I promise.

I believe so.

@youngjoon-lee 

n/a

n/a

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/services/chain/chain-leader/src/lib.rs
+++ b/services/chain/chain-leader/src/lib.rs
@@ -330,6 +330,7 @@ where
             TxMempoolService<_, _, _, _>,
             TimeService<_, _>,
             CryptarchiaService,
+            ChainNetwork,
             Wallet,
             PreloadKmsService<_>
         )


services/chain/chain-network/src/api.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+pub trait ChainNetworkServiceData:
+    ServiceData<Message = Message<Self::Tx>> + Send + 'static
+{
+    type Tx;
+}
+
+impl<T, Tx> ChainNetworkServiceData for T
```
ntn-x2, line 62: What is `T`? Can you use a more descriptive name?

ntn-x2 (approved):

services/chain/chain-network/src/api.rs
youngjoon-lee: I think T is fine here because we're implementing the trait to any type that implements ServiceData. I adopted this convention from other `*ServiceData` traits. I thought about any other names, but they didn't make this clearer so much.

PR #2103, opened 2026-02-03
perf: further optimize release builds

ntn-x2:

Apply optimizations to third-party dependencies, and remove unnecessary symbols from the final binary, making it smaller in size. This is as far as we can go without adding CPU-specific flags, which I don't think we need for now. For reference, check the [Cargo profiles reference](https://doc.rust-lang.org/cargo/reference/profiles.html).

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,8 +11,8 @@ version     = "0.1.0"
 [profile.release]
 codegen-units = 1       # (Much) slower build time, much better performances
 lto           = "fat"   # Same as above
-panic         = 'abort' # Smaller binary size
-strip         = true
+panic         = "abort" # Remove panic handler to reduce binary size
+strip         = true    # Remove symbols to reduce binary size
 
 [profile.release.package."*"]
 # LTO cannot be overridden for downstream dependencies.


Cargo.toml
```diff
@@ -9,8 +9,15 @@ repository  = "https://github.com/logos-blockchain/logos-blockchain"
 version     = "0.1.0"
 
 [profile.release]
-codegen-units = 1     # (Much) slower build time, much better performances
-lto           = "fat" # Same as above
+codegen-units = 1       # (Much) slower build time, much better performances
+lto           = "fat"   # Same as above
+panic         = "abort" # Remove panic handler to reduce binary size
```
hansieodendaal, line 14: @ntn-x2 something to consider..
For testnet, `panic = "abort"` might be too aggressive. With `panic = "unwind"`, a panic inside a `tokio::spawn` task typically only kills that task (Tokio catches it); the runtime and other tasks continue. With `panic = "abort"`, any panic aborts the entire node; for example, a panic in a peer connection handler wouldn't crash the whole node.

ntn-x2: Fair point, we don't want that happening indeed, probably not even on our mainnet. Overwatch gives us independent sub-processes, and crashing the whole node if any of them panics is probably not the way to go. I will remove this configuration and ask you for your review. Good catch!

danielSanchezQ: Uhm, a panic means that is not recoverable. I think we actually take care a lot on where to panic and where not. Other question are panics from third party crates :thinking: 
Let's go with the most useful for now (not aborting), but at some point aborting should be what we have to do.

ntn-x2: @danielSanchezQ ideally Overwatch would handle these panics and try to recover the panicking service, if it makes sense. Aborting does not even give Overwatch a chance. But yeah, let's re-introducing unwinding for now, I agree...

ntn-x2: Just for the record: https://github.com/logos-blockchain/logos-blockchain/pull/2106.

PR #2107, opened 2026-02-03
chore(release): Devnet setup instructions

bacv:

Devnet setup instructions in the release template

Yes

@bacv

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/.github/release/release-content.md
+++ b/.github/release/release-content.md
@@ -34,18 +34,13 @@ curl -X POST -L --location-trusted https://devnet.blockchain.logos.co/node/0/cfg
      -o my_logos_node_config.yaml
 ```
 
-Once the configuration file is downloaded, you can start your node with:
-```bash
-logos-blockchain-node-macos-aarch64-0.0.1 --deployment devnet my_logos_node_config.yaml 
-```
-
-To verify that your node is running correctly and connected, visit http://localhost:8080/cryptarchia/info. The slot and height should both be constantly increasing.
-
 ## Run the binary
 
 After obtaining the deployment file for the network you want to join and tweaking the node config file to fit your needs, including specifying the list of bootnodes for the network you are joining, you can run the node binary.
 
 For example: `logos-blockchain-node-macos-aarch64-0.0.1 --deployment deployment.yaml node-config.yaml`. See the repo's `README.md` for more info.
+
+To verify that your node is running correctly and connected, visit http://localhost:{api_port_in_user_config}/cryptarchia/info. The slot and height should both be constently increasing.
 
 ## Checklist
 


.github/release/release-content.md
```diff
@@ -22,6 +22,32 @@ After obtaining the deployment file for the network you want to join and tweakin
 
 For example: `logos-blockchain-node-macos-aarch64-0.0.1 --deployment deployment.yaml node-config.yaml`. See the repo's `README.md` for more info.
 
+
+## Devnet setup
+
+If you wish to join the devnet at https://devnet.blockchain.logos.co, you can automatically generate and download your node configuration using the following command:
+
+```bash
+curl -X POST -L --location-trusted https://devnet.blockchain.logos.co/node/0/cfgsync/config/custom-node \
```
ntn-x2, line 24: Do we need to reference a specific node in the address?

.github/release/release-content.md
```diff
@@ -22,6 +22,32 @@ After obtaining the deployment file for the network you want to join and tweakin
...
+     -u "username:password" \
+     -H "Content-Type: application/json" \
+     -d '{
+            "ip": "192.168.6.7",
+            "identifier": "marcins-anonymous-node",
```
ntn-x2, line 29: Who's Marcin? 🙈 What else could go here?

.github/release/release-content.md
```diff
@@ -22,6 +22,32 @@ After obtaining the deployment file for the network you want to join and tweakin
 
 For example: `logos-blockchain-node-macos-aarch64-0.0.1 --deployment deployment.yaml node-config.yaml`. See the repo's `README.md` for more info.
 
+
+## Devnet setup
```
ntn-x2, line 19: Shouldn't this section go before the one about running the binary? We can set up the config file then follow the instructions to run the binary.

ntn-x2: Also, the command to run the node you wrote below only specifies a user config, while it uses the default deployment which is the testnet one. If the call below downloads both files, we should specify both of them.

.github/release/release-content.md
bacv: Unfortunately at the moment we need to have this specific `/node/0` part.

We'll need to coordinate with infra to change cfgsync url. For this weeks release I suggest we use what we have. Once we made the address without specific node, we'll update this document.

.github/release/release-content.md
bacv: > Who's Marcin?

Nobody knows, he's anonymous :D But suggestions are welcome.

.github/release/release-content.md
bacv: > If the call below downloads both files
 
This technically downloads the `RunConfig`, with deployment and user config, but since it's flattened, the node only uses the user config. Will fix this in the cfgsync PR as it's not related to this change. 

Lets use the testnet deployment config.

ntn-x2: Ok I will download the deployment config and add a `--deployment devnet` flag that uses those values, then we can tell people to use that flag instead.

bacv: > Shouldn't this section go before the one about running the binary? We can set up the config file then follow the instructions to run the binary.

I'm not sure since it has both setup and run instructions. I can move it, but after looking at it for a while, pushed down because it will get less important with future releases

ntn-x2: Once not relevant, we will move it or delete it later on. I think for now it would make more sense to run the setup step before the run step, in my opinion.

bacv: > Ok I will download the deployment config and add a --deployment devnet 

Added the flag in https://github.com/logos-blockchain/logos-blockchain/pull/2107/commits/c7cb8c44f6fce8e72decd698f47d4c8ebc18e358.

> Once not relevant, we will move it or delete it later on.

Pushed above "run" in the same commit.

.github/release/release-content.md
```diff
@@ -16,6 +16,31 @@ For the former, please reach out to the Logos Blockchain team on [Discord](https
...
+            "ip": "192.168.6.7",
+            "identifier": "marcins-anonymous-node",
+            "network_port": 3000,
+            "blend_port": 4000,
+            "api_port": 8080
+         }' \
+     -o my_logos_node_config.yaml
+```
+
+Once the configuration file is downloaded, you can start your node with:
```
ntn-x2: I would just remove this section since we have a section below that instructs on how to run the binary, and we could add there how to check that it's running ok. I can also do in the PR that introduces the devnet flag, but I think it could be done here, if you agree.

bacv: Removed run instructions from devnet setup section and moved verification instructions to run section https://github.com/logos-blockchain/logos-blockchain/pull/2107/commits/b5ecdffcd275c2fe7d324653425a8e33b55be652

PR #2115, opened 2026-02-04
fix(chain-leader): do not propose blocks while chain is in Bootstrapping mode

### LINKED ISSUE #2100: Don't propose blocks duing the Bootstrap mode
The chain-leader service should wait until the chain-service switches its mode from Bootstrap to Onilne. Until then, the chain-leader shouldn't propose any block.

youngjoon-lee:

Closes #2100 

### Problems

The chain-service was proposing blocks even while the chain is in the Bootstrapping mode (but after IBD completed). 
However, as discussed in [Discord](https://discord.com/channels/973324189794697286/1468073076943224892), this is not good because the node could have a permanent fork, if it proposes `k` blocks during Bootstrapping mode very quickly and doesn't find any better fork.

We have the density-based fork choice rule which protects us from this case to some extent.
But, if your node is connected only with other **new** nodes that are also bootstrapping, the IBD will be completed very quickly and we will enter PBP (while the mode is still Boostrapping). In this example, if the node proposes `k` blocks, and if PBP has passed before it discovers other nodes in the network, the node ends up having a permanent fork. It is because the node enters the Online mode after PBP is done and we ignore forks deeper than `k` in the Online mode.

Connecting to only nodes that are **not** in the Bootstrapping mode doesn't solve this problem, because we cannot trust any information provided by other nodes, which are not validated by the consensus algorithm.

Therefore, we should not propose blocks during Bootstrapping.

### Changes 
In this PR,
- `chain` service provides a new API `SubscribeChainOnline`. It notifies subscribers when the chain becomes Online mode after IBD+PBP.
- `chain-leader` subscribes it. After it is notified, it marks itself ready. 
  - Services that depend on `chain-leader` readiness must use a zero timeout because the wait can be very lengthy. Currently, only the `blend` service does this (via [binary impl](https://github.com/logos-blockchain/logos-blockchain/blob/2392190d88e8ae8271fa9321014ea33324be7c28/nodes/node/binary/src/generic_services/blend/mod.rs#L93)) @ntn-x2 .
- The genesis(?) nodes should set PBP=0 (or very short), allowing them to switch to Online and start proposing blocks immediately. 

I recommend to review 3 commits separately.

I believe so

spec: @davidrusu @youngjoon-lee 
dev: @youngjoon-lee 

This behavior should be specified in the spec.

I think the bootstrap spec is the best place to mention this behavior. I will update the [diagram](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df8113baa7c8c751f7d8e3) in the bootstrap spec, and add a new section right after the [Prolonged Bootstrap Period](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df8162be49e5aa02199378) section which is describing when the node is switched to the Online rule. In the new section, I can mention that the node can finally start proposing blocks after switching to Online. 
And we can link it to the [Leadership Lottery](https://www.notion.so/nomos-tech/Cryptarchia-v1-Protocol-Specification-21c261aa09df810cb85eff1c76e5798c?source=copy_link#21c261aa09df81d6a165ed07bd639dae) section in the main spec, in one sentence, because that section is only one place in the main spec which mentions about proposing blocks.

## Checklist

> [!WARNING]  

* [x] 2. Description added.


services/chain/chain-leader/src/lib.rs
youngjoon-lee, line 333: This change is not highly related to the main topic, but is indirectly necessary. We should have had this from the beginning, as mentioned in the code comment.

services/chain/chain-leader/src/lib.rs
```diff
@@ -322,7 +320,8 @@ where
 
         wait_until_services_are_ready!(
             &self.service_resources_handle.overwatch_handle,
-            Some(Duration::from_secs(60)),
+            // No timeout since ChainNetwork service becomes ready only after IBD is complete.
+            None,
```
danielSanchezQ: Instead we should remove the `ChainNetwork` from this call, and add a non-waiting call only on it (?) or split services between the ones that should be ready and hence have a timeout and the ones that not.

services/chain/chain-service/src/api.rs
```diff
@@ -243,4 +243,21 @@ where
...
+            .send(ConsensusMsg::SubscribeChainOnline { sender })
+            .await
+            .map_err(|(relay_error, _)| {
+                ApiError::CommsFailure(format!("{relay_error} while sending SubscribeChainOnline"))
+            })?;
+
+        receiver.await.map_err(|relay_error| {
+            ApiError::CommsFailure(format!("{relay_error} while recving SubscribeChainOnline"))
+        })
+    }
```
danielSanchezQ: Couldn't we override the overwatch service status instead of a custom notifier? Not sure atm.

services/chain/chain-service/src/lib.rs
```diff
@@ -432,6 +439,7 @@ where
     service_resources_handle: OpaqueServiceResourcesHandle<Self, RuntimeServiceId>,
     new_block_subscription_sender: broadcast::Sender<ProcessedBlockEvent>,
     lib_subscription_sender: broadcast::Sender<LibUpdate>,
+    chain_online_subscription_channel: watch::Sender<Option<()>>,
```
ntn-x2: I think using a `bool` would probably be easier to understand.

services/chain/chain-service/src/api.rs
```diff
@@ -243,4 +243,21 @@ where
...
+
+        self.relay
+            .send(ConsensusMsg::SubscribeChainOnline { sender })
+            .await
+            .map_err(|(relay_error, _)| {
+                ApiError::CommsFailure(format!("{relay_error} while sending SubscribeChainOnline"))
+            })?;
+
+        receiver.await.map_err(|relay_error| {
+            ApiError::CommsFailure(format!("{relay_error} while recving SubscribeChainOnline"))
```
ntn-x2:
```suggestion
            ApiError::CommsFailure(format!("{relay_error} while receiving SubscribeChainOnline"))
```

services/chain/chain-service/src/api.rs
ntn-x2: I think it would need to support a generic state when it's in the `Started` state, no?

ntn-x2 (approved):

services/chain/chain-service/src/api.rs
danielSanchezQ: > I think it would need to support a generic state when it's in the `Started` state, no?

Yeah, I think this is something we could add to the system in the future. So that started holds an internal state defined by the service :thinking:

services/chain/chain-service/src/lib.rs
danielSanchezQ: :heavy_plus_sign: **1**

services/chain/chain-service/src/api.rs
youngjoon-lee: ChainService becomes ready even before the chain enters the Online mode because it should be able to process blocks during Bootstrapping mode. So, other services (e.g. chain-leader) who want to wait until chain becomes Online cannot rely on checking the Overwatch service state, if I understand your comment correctly

services/chain/chain-service/src/lib.rs
youngjoon-lee: 934fff779519136dbad3f1363b9d8a8f37c03c29 Good point. I changed the type to `bool`, and also simplified the API.

services/chain/chain-leader/src/lib.rs
youngjoon-lee: Thanks. 6c54bddbc2640d3d88b82ca53029975c73d3328b Done

PR #2118, opened 2026-02-04
chore(cryptarchia-engine): cryptarchia engine branches minor refactor

pradovic:

I am implementing unfinalized blocks tracking (pending, safe tx) in `zone-sdk`. In order to not duplicate work it looks like `Branches<Id>` structure can be reused, but to be complete it is missing a few methods so that `zone-sdk` does not need to access private fields in order to use it. This PR expands only those methods and makes cryptarchia use it as well

Yes

@pradovic 

> [!IMPORTANT]Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/consensus/cryptarchia-engine/src/lib.rs
+++ b/consensus/cryptarchia-engine/src/lib.rs
@@ -374,54 +374,43 @@ where
         removed_blocks
     }
 
-    /// Get an iterator over the prunable forks that diverged before
-    /// the `max_div_depth`-th block from the given local chain tip.
+    /// Get an iterator over the prunable forks that diverged before LIB.
     pub fn prunable_forks(
         &self,
         local_chain: &Branch<Id>,
-        max_div_depth: u64,
     ) -> impl Iterator<Item = ForkDivergenceInfo<Id>> + '_ {
         let local_chain = *local_chain;
-        let Some(deepest_div_block) = local_chain.length.checked_sub(max_div_depth) else {
-            tracing::debug!(
-                target: LOG_TARGET,
-                "No prunable fork, the canonical chain is not longer than the provided depth. Canonical chain length: {}, provided max_div_depth: {}", local_chain.length, max_div_depth
-            );
-            return Box::new(core::iter::empty())
-                as Box<dyn Iterator<Item = ForkDivergenceInfo<Id>>>;
-        };
+        let lib_length = self.lib_branch().length;
         let local_tip = local_chain.id;
         Box::new(
             self.branches()
                 .filter(move |fork_tip| fork_tip.id != local_tip)
                 .filter_map(move |fork| {
                     let lca = self.lca(&local_chain, &fork);
-                    (lca.length < deepest_div_block)
-                        .then_some(ForkDivergenceInfo { tip: fork, lca })
+                    (lca.length < lib_length).then_some(ForkDivergenceInfo { tip: fork, lca })
                 }),
-        )
+        ) as Box<dyn Iterator<Item = ForkDivergenceInfo<Id>>>
     }
 
-    /// Prune all blocks that are included in forks that diverged before
-    /// the `max_div_depth`-th block from the current local chain tip.
+    /// Prune all blocks that are included in forks that diverged before LIB.
     /// It returns the block IDs that were part of the pruned forks.
     ///
     /// For example,
-    /// Given a block tree:
+    /// Given a block tree where LIB is `b2`:
     ///               b6
     ///             /
     /// G - b1 - b2 - b3 - b4 - b5 == local chain tip
     ///                  \
     ///                    b7
-    /// Calling `prune_forks(2)` will remove `b6` because it is diverged from
-    /// `b2`, which is deeper than the 2nd block `b3` from the local chain tip.
-    /// The `b7` is not removed since it is diverged from `b3`.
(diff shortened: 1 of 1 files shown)


consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -333,6 +333,121 @@ where
...
+        forks
+            .into_iter()
+            .flat_map(|info| self.prune_fork(info.tip.id, info.lca.id))
+            .collect()
+    }
+
+    /// Prunes all immutable blocks (excluding LIB) that are deeper than LIB,
+    /// and returns the slots and IDs of the pruned blocks.
+    pub fn prune_immutable_blocks(&mut self) -> Vec<(Slot, Id)> {
+        let mut block = self.lib_branch().parent;
```
youngjoon-lee, line 423: These functions were previously returning Iterator, but this PR changes them to return Vec. I think we can keep it as Iterator, so that callers can use it as they want.
If we really need to change it to Vec, we should also remove redundant `collect()` from `Cryptarchia::update_lib` below.

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -418,15 +533,22 @@ where
...
+                .into_iter()
+                .collect();
+            let immutable_blocks = self.branches.prune_immutable_blocks().into_iter().collect();
             PrunedBlocks {
-                // TODO: Eliminate the need of `lib_depth` by refactoring `prune_stale_forks`,
-                //       similar as `prune_immutable_blocks`.
-                stale_blocks: self.prune_stale_forks(self.lib_depth()).collect(),
-                immutable_blocks: self.prune_immutable_blocks().collect(),
+                stale_blocks,
+                immutable_blocks,
```
youngjoon-lee, line 551: I'm not against exposing the `Branches` struct itself because it has been already `public` and used in outside modules. I don't think it's optimal but it has been like that from the ancient :)

Nevertheless, I think we need to rethink about the changes in this PR. Previously, the `Branches` struct was designed as an **immutable** struct. It didn't provide any method with `&mut self` (whether public or private). Only the `Cryptarchia` struct could re-create `Branches` with updated states. 
The same policy applies to `Cryptarchia`. It's also an immutable struct and doesn't provide 'public' methods with `&mut self`.

But, in this PR, we are introducing several public methods for `Branches` with `&self mut`. I think we need to be careful about how `Branches` will be used by callers. Let me share some concerns I have.

We are adding `Branches::set_lib` function, which changes LIB **without** pruning forks. This can cause intermediate state (i.e. it's not atomic).
The previous way was not also 100% perfect, but at least, `Cryptarchia::update_lib` was the only place where the `self.branches.lib` was updated. But, if we define `Branches::set_lib` and make it public, callers could misuse it.

Instead, we can probably remove `Branches::set_lib` and keep only `Branches::advance_lib` (which also you added in this PR). It seems that `Branches::advance_lib`  is not used anywhere in this repo, but I think we can just rename it to `Branches::update_lib` and use it in `Cryptarchia::update_lib`. To do that, we also have to change `advance_lib` to prune stale forks as well as immutable blocks, IIUC. 

Another thing I'm concerning is that we are giving callers freedom to set **any** LIB to `Branches` by calling `Branches::set_lib` or `Branches::advance_lib`. I don't think it is safe. Previously, callers couldn't do it because `Cryptarchia::update_lib` doesn't accept any argument and compute the current LIB by itself.

Plus, I'm curious how callers will obtain a `Branch` object. Currently, even in this PR, `Cryptarchia::branches` provides `&Branches`. If we really want to expose `Branches` to callers, `Cryptarchia` needs to exposes a cloned `Branches`, I guess.

I'm not sure how `Branches` will be used for `zone-sdk`, but I think we may need to minimize public methods if possible.
Could you please let me know which features `zone-sdk` requires from `cryptarchia-engine`? I can help supporting it. Maybe we could add more public methods to `Cryptarchia`, not `Branches`.

consensus/cryptarchia-engine/src/lib.rs
```diff
@@ -333,6 +333,121 @@ where
         }
         *current
     }
+
+    /// Returns the current LIB (last irreversible block).
```
danielSanchezQ, line 337:
```suggestion
    /// Returns the current LIB (last immutable block).
```

consensus/cryptarchia-engine/src/lib.rs
danielSanchezQ: I agree with @youngjoon-lee , this is core, we cannot fuck-up here. Imo exposing methods to be public so we can reuse them is ok, but changing the fundations not. What we should do instead, is re-wrapp this `Branches` in a specialized struct for the `zones-sdk` which expands/modify the functionality if needed exposing the same API more or less.

pradovic: Thanks for the feedback, I agree with all the points!

I’ll implement the logic in `zone-sdk first by copying the minimal functionality, so we don’t change `cryptarchia-engine` prematurely.

Once I have a working version and a clearer picture of what’s actually needed, I can come back with a smaller proposal — either a thin wrapper in `zone-sdk` (as Daniel suggested) or a minimal, safe API surface in core — if there’s a stable abstraction worth extracting.

Closing the PR for now, thanks!

Issue #2098, opened 2026-02-02
Remove known vouchers in `lb-wallet::Wallet` after claim txs are finalized


youngjoon-lee:
`lb-wallet::Wallet` holds a set of known voucher commitments and nullifiers but it doesn’t prune them currently.

Whenever it sees immutable blocks (LIB and any deeper blocks), it should prune vouchers if the corresponding voucher nullifiers are found in the transactions in the immutable blocks. 

The Wallet already has a function for pruning WalletStates. We can extend the function.

PR #2120, opened 2026-02-04
feat(wallet): Remove lib processed vouchers from wallet

danielSanchezQ:

This PR introduces voucher pruning trying to follow up #2098. I retrieved the pruned blocks from storage, but I'm not sure if there is a better way of getting the claim operations from the structures we had in place (I couldn't find a better way). Otherwise the API should be simple. 

:+1: 

@youngjoon-lee @davidrusu @danielSanchezQ 

N/A

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/services/wallet/src/lib.rs
+++ b/services/wallet/src/lib.rs
@@ -129,6 +129,12 @@ pub enum WalletMsg {
     GenerateNewVoucherSecret {
         resp_tx: oneshot::Sender<VoucherCm>,
     },
+    GetClaimableVoucher {
+        tip: Option<HeaderId>,
+        resp_tx: oneshot::Sender<
+            Result<TipResponse<Option<VoucherCommitmentAndNullifier>>, WalletServiceError>,
+        >,
+    },
 }
 
 #[derive(Debug)]
@@ -143,6 +149,12 @@ pub struct UtxoWithKeyId {
     pub key_id: KeyId,
 }
 
+#[derive(Debug)]
+pub struct VoucherCommitmentAndNullifier {
+    pub commitment: VoucherCm,
+    pub nullifier: VoucherNullifier,
+}
+
 impl WalletMsg {
     /// Returns [`HeaderId`] of the tip if the message is associated
     /// with a specific tip.
@@ -152,7 +164,8 @@ impl WalletMsg {
             Self::GetBalance { tip, .. }
             | Self::FundTx { tip, .. }
             | Self::SignTx { tip, .. }
-            | Self::GetLeaderAgedNotes { tip, .. } => *tip,
+            | Self::GetLeaderAgedNotes { tip, .. }
+            | Self::GetClaimableVoucher { tip, .. } => *tip,
             Self::GenerateNewVoucherSecret { .. } => None,
         }
     }
@@ -456,6 +469,9 @@ where
                 )
                 .await;
             }
+            WalletMsg::GetClaimableVoucher { tip, resp_tx } => {
+                Self::get_claimable_voucher(tip, resp_tx, state.wallet(), cryptarchia).await;
+            }
         }
     }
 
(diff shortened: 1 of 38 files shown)


services/wallet/src/lib.rs
```diff
@@ -890,6 +895,28 @@ where
         );
 
         state.prune_states(lib_update.pruned_blocks.all());
+        let pruned_blocks: Vec<Block<Tx>> =
```
youngjoon-lee:
```suggestion
        let immutable_blocks: Vec<Block<Tx>> =
```
I think it's better to call this `immutable_blocks`, because it's a subset of `pruned_blocks`.

services/wallet/src/lib.rs
```diff
@@ -890,6 +895,28 @@ where
...
+            futures::stream::iter(lib_update.pruned_blocks.immutable_blocks.values())
+                .filter_map(async |header_id: &HeaderId| storage_adapter.get_block(header_id).await)
+                .collect::<Vec<_>>()
+                .await;
+        let pruned_nullifiers: Vec<VoucherNullifier> = pruned_blocks
```
youngjoon-lee:
```suggestion
        let claimed_nullifiers: Vec<VoucherNullifier> = pruned_blocks
```
I think, `claimed_nullifiers` is a clearer name, because we are collecting vouchers in the claim operations in the immutable blocks.

services/wallet/src/states.rs
```diff
@@ -89,6 +89,13 @@ impl<'u> ServiceState<'u> {
         self.update_state();
     }
 
+    pub fn prune_vouchers(
+        &mut self,
+        pruned_nullifiers: impl IntoIterator<Item = VoucherNullifier>,
+    ) {
+        self.wallet.prune_vouchers(pruned_nullifiers);
```
youngjoon-lee, line 92:
```suggestion
        self.wallet.prune_vouchers(pruned_nullifiers);
        self.update_state();
```
Since we are mutating `self.wallet`, we should store it to the recovery state. We have the `self.update_state` function for it. We could rename the function clearer, though. Or we can consider having any better way to not forget calling `update_state` (e.g. wrapper, or so), but later after merging this PR.

services/wallet/src/lib.rs
```diff
@@ -881,7 +882,11 @@ where
         Ok((block, ledger))
     }
 
-    fn handle_lib_update(lib_update: &LibUpdate, state: &mut ServiceState<'_>) {
+    async fn handle_lib_update(
+        lib_update: &LibUpdate,
+        storage_adapter: &StorageAdapter<Storage, Tx, RuntimeServiceId>,
+        state: &mut ServiceState<'_>,
+    ) {
```
youngjoon-lee, line 949: > I retrieved the pruned blocks from storage, but I'm not sure if there is a better way of getting the claim operations from the structures we had in place (I couldn't find a better way). Otherwise the API should be simple.

I agree that using storage is the only way to get transactions in the immutable blocks, currently.

services/wallet/src/lib.rs
danielSanchezQ: Renamed at [2e6f480](https://github.com/logos-blockchain/logos-blockchain/pull/2120/commits/2e6f4800b81d0148351b5709f385081969eea821)

services/wallet/src/lib.rs
danielSanchezQ: Renamed at [2e6f480](https://github.com/logos-blockchain/logos-blockchain/pull/2120/commits/2e6f4800b81d0148351b5709f385081969eea821)

services/wallet/src/states.rs
danielSanchezQ: Added the call at [2e6f480](https://github.com/logos-blockchain/logos-blockchain/pull/2120/commits/2e6f4800b81d0148351b5709f385081969eea821)

PR #2124, opened 2026-02-05
feat(api): remove rate-limits from Node API

davidrusu:

Removes Rate limits from node API.

why?
1. we are hitting rate limits consistently in local testing
2. rate limits are actually implemented incorrectly (`rate_limit_per_second:1000` actually means 1 req per 1000 seconds)
3. these API's are not to be exposed externally (they expose wallet functionality so should only be used in a trusted environment)
4. don't want these rate limits to be the reason we can't push our testnet too hard (lets find the physical rate limits instead)

yes

NA

NA

NA

## Checklist

* [X] 2. Description added.


danielSanchezQ (approved):
🚢

PR #2132, opened 2026-02-05
feat(zone-sdk): implement v0 non-finalized block tracking

pradovic:

This PR introduces v0 transaction/inscription non-finalized state tracking in zone-sdk.

It tracks pending, safe, and finalized transactions based on processed blocks, handles basic reorgs, and prunes old blocks below LIB to keep memory bounded. The implementation is similar to the wallet service.

Includes periodic re-submission of non-finalized transactions and basic tests.

The main assumption is a continuous stream of events from the node. Retry logic, resilience, persistence, and other durability concerns are not yet covered and will be addressed in future PRs.

A follow-up PR will likely focus on restart handling and backfilling.

Yes

@pradovic @davidrusu @danielSanchezQ 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/zone-sdk/src/state.rs
+++ b/zone-sdk/src/state.rs
@@ -80,10 +80,9 @@ impl TxState {
 
         // When lib advances: finalize txs and prune
         if lib != self.current_lib {
-            // Finalize txs in all blocks from old lib to new lib (inclusive)
-            let mut block = lib;
-
-            loop {
+            // Finalize txs in all blocks from new lib back to old lib (inclusive)
+            let mut block_opt = Some(lib);
+            while let Some(block) = block_opt {
                 let block_safe = self
                     .block_states
                     .get(&block)
@@ -99,21 +98,14 @@ impl TxState {
                     break;
                 }
 
-                block = self
-                    .parent_map
-                    .get(&block)
-                    .copied()
-                    .expect("parent_map should contain block while finalizing LIB advance");
+                block_opt = self.parent_map.get(&block).copied();
             }
 
-            // Prune ancestors of new lib (up to and including old lib)
+            // Prune ancestors of new lib (but not lib itself)
             let mut prune_cursor = self.parent_map.get(&lib).copied();
             while let Some(b) = prune_cursor {
                 self.block_states.remove(&b);
                 prune_cursor = self.parent_map.remove(&b);
-                if b == self.current_lib {
-                    break;
-                }
             }
 
             self.prune_orphans(lib);


zone-sdk/src/state.rs
```diff
@@ -1,66 +1,179 @@
...
+
+        // Build cumulative safe set from parent
+        let parent_state = self.block_states.get(&parent_id).cloned();
+
+        // TODO: backfill missing blocks
+        debug_assert!(
+            parent_state.is_some(),
+            "Missing parent state for {parent_id:?}, cumulative safe set will be incomplete"
+        );
+        let mut safe_set = parent_state.unwrap_or_else(HashTrieSetSync::new_sync);
```
davidrusu: until backfilling is implemented, we should panic on this unwrap, so just do `.expect("parent state should exist")`

zone-sdk/src/state.rs
```diff
@@ -1,66 +1,179 @@
...
+
+            // Prune ancestors of new lib (up to and including old lib)
+            let mut cursor = self.parent_map.get(&lib).copied();
+            while let Some(block) = cursor {
+                self.block_states.remove(&block);
+                cursor = self.parent_map.remove(&block);
+                if block == self.current_lib {
+                    break;
+                }
+            }
```
davidrusu: If LIB advances by more than one block, we should also finalize the intermediate blocks as well.

```
X <- CURRENT_LIB <- Y <- LIB
```

This logic would only finalize transactions in `LIB`, we need to also finalize transactions in Y

pradovic: Oups, good catch! Here is a simple fix: https://github.com/logos-blockchain/logos-blockchain/pull/2132/changes/0c2a86b63ebdf2a5ae9915cbb944cff6db52a312. Does this look correct now?

zone-sdk/src/state.rs
```diff
@@ -1,66 +1,194 @@
...
+            // Prune ancestors of new lib (up to and including old lib)
+            let mut prune_cursor = self.parent_map.get(&lib).copied();
+            while let Some(b) = prune_cursor {
+                self.block_states.remove(&b);
+                prune_cursor = self.parent_map.remove(&b);
+                if b == self.current_lib {
+                    break;
+                }
+            }
+
```
davidrusu: I was looking why we don't merge this loop with the one above, but the reason for the two loops here is because you don't want to prune lib, but you still want to finalize txs in lib?

Still, I think this while loop structure is better than the `loop {.. break;}` structure, can we use this above?

Moreover, we don't need the `if b==self.current_lib {break;}` check here, we should iterate until there is no more parent in the parent map

zone-sdk/src/state.rs
```diff
@@ -1,66 +1,194 @@
...
+                if block == self.current_lib {
+                    break;
+                }
+
+                block = self
+                    .parent_map
+                    .get(&block)
+                    .copied()
+                    .expect("parent_map should contain block while finalizing LIB advance");
```
davidrusu: Can replace the bare loop with:
```rust
let mut block_opt = Some(lib)
while let Some(block) = block_opt  {
    //....
    block_opt = self.parent_map.get(&block).copied();
}
```

zone-sdk/src/state.rs
pradovic: Yes, I tried to merged them as well, but ended up splitting for the same reason. Makes sense about the loop, switching back to if let: https://github.com/logos-blockchain/logos-blockchain/pull/2132/changes/51ce4a6e3b43f4af15dc44ae2b2ef6efbc862053.

PR #2149, opened 2026-02-09
feat: promote genesis state to deployment config

ntn-x2:

Promote the genesis state to be part of deployment setting instead of user setting. Change is reflected in e2e and partially in cfgsync binaries, which need further updates, I think.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nodes/node/binary/src/config/cryptarchia/deployment.rs
+++ b/nodes/node/binary/src/config/cryptarchia/deployment.rs
@@ -81,10 +81,7 @@ fn devnet_settings() -> Settings {
             .into(),
         },
         security_param: 20.try_into().unwrap(),
-        // We use a dummy genesis state here, since the deployment setting will be overridden by
-        // `cfgsync` binary when spinning up devnets in the beginning. Once we will have a
-        // stable devnet, this will be a proper genesis state with real initial validators and their
-        // stakes.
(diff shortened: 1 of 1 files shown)


core/src/mantle/genesis_tx.rs
```diff
@@ -64,6 +64,21 @@ impl GenesisTx {
...
+        use lb_key_management_system_keys::keys::ZkSignature;
+
+        use crate::mantle::tx_builder::MantleTxBuilder;
+
+        Self(SignedMantleTx::new_unverified(
+            MantleTxBuilder::new().build(),
+            vec![],
+            ZkSignature::new(CompressedGroth16Proof::from_bytes(&[0; _])),
+        ))
+    }
```
danielSanchezQ, line 81: Why do we need to mock this? Just to generate an fake config?

ntn-x2: Yeah, we will replace this with an actual state before the next release. This is a temporary placeholder. There is currently no way to create a genesis state skipping verification. We either expose a `new_unverified` method, expose a method behind the `mock` feature (like here), or then I need to come up with a valid structure to declare an empty state, in which case I could use `GenesisTx::from_tx().unwrap()`.

PR #2150, opened 2026-02-09
test: improve cucumber tests

hansieodendaal:

- Improved cucumber test artefacts by persisting them to the test output location, grouped by `feature>scenario>attempt>node`.
   
  image

- Removed old `NOMOS_` const references.
- Re-pointed to the updated test framework commit.
- Ensured that required test artefacts are uploaded in CI.
- Update test framework tag.
- Fixed `test_ibd_behind_nodes` locally, CI still has some funny behaviour.
- Curious difference with e2e tests using debug vs. release binaries - see e2e local runs timing difference below:
  
  image

- Compiling and running without `high-active-slot-coefficient` feature flag is even more interesting compared with the previous:

  image

~~**Note:** `Orphan staggered fork start 1` and `Orphan staggered fork start 2` has fundamental issues - forks are created when multiple nodes join the leader, and they never converge.~~ _Fixed with using consistent chain start time._

Yes

@hansieodendaal 

n/a

n/a

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/tests/src/cucumber/world.rs
+++ b/tests/src/cucumber/world.rs
@@ -238,8 +238,8 @@ impl CucumberWorld {
             .map_err(|source| StepError::ScenarioBuild { source })
     }
 
-    /// Perform preflight checks to ensure the world is properly configured for the
-    /// expected deployer kind.
+    /// Perform preflight checks to ensure the world is properly configured for
+    /// the expected deployer kind.
     pub fn preflight(&self, expected: DeployerKind) -> Result<(), StepError> {
         let actual = self.deployer.ok_or(StepError::MissingDeployer)?;
         if actual != expected {


.github/workflows/cucumber.yml
ntn-x2: Since you're here, can we also remove this env variable from the e2e tests? It's not really needed in there but only for cucumber tests.

.github/workflows/cucumber.yml
```diff
@@ -110,14 +110,25 @@ jobs:
...
-            or @undefined_behaviour)"
+            --test cucumber -- --tags "@smoke_ci and not (@broken or @undefined_behaviour)"
+
+      # On failure, upload cucumber logs and artifacts for debugging
+      - name: Upload cucumber logs on failure
+        if: failure()
+        continue-on-error: true
+        uses: actions/upload-artifact@6027e3dd177782cd8ab9af838c04fd81a07f1d47
+        with:
+          name: cucumber-artifacts-${{ runner.os }}  # ✅ Unique per OS
```
ntn-x2: Per PR as well? I think we typically do OS + PR or branch name + workflow run.

.github/workflows/end-to-end-with-cucumber.yml
```diff
@@ -127,16 +138,19 @@ jobs:
...
+        continue-on-error: true
+        uses: actions/upload-artifact@6027e3dd177782cd8ab9af838c04fd81a07f1d47
         with:
-          command: clean
+          name: cucumber-artifacts-${{ runner.os }}  # ✅ Unique per OS
+          include-hidden-files: true
+          path: ${{ github.workspace }}/tests/temp/cucumber_tests/cucumber_artefacts/Cryptarchia/**
 
       # Always upload JUnit test results for reporting, even if tests fail
       - name: Upload cucumber JUnit results
```
ntn-x2, line 156: We know `if: always()` is not a good condition to specify :)

.github/workflows/cucumber.yml
```diff
@@ -145,29 +159,17 @@ jobs:
...
-        with:
-          name: cucumber-artifacts-${{ runner.os }}  # ✅ Unique per OS
-          include-hidden-files: true
-          path: |
-            cucumber_tests/temp/**
-            cucumber_tests/temp/node-logs/**
-
       # Always publish a test report for visibility, do not fail workflow if this step fails
       - name: Publish cucumber test report
         if: always()
```
ntn-x2: Also here, no `if: always()` please.

tests/Cargo.toml
```diff
@@ -53,14 +53,10 @@ serde_yaml                       = "0.9"
...
-#testing-framework-runner-compose = { branch = "master", git = "https://github.com/logos-blockchain/logos-blockchain-testing.git" }
-#testing-framework-runner-local   = { branch = "master", git = "https://github.com/logos-blockchain/logos-blockchain-testing.git" }
-#testing-framework-workflows      = { branch = "master", git = "https://github.com/logos-blockchain/logos-blockchain-testing.git" }
-testing-framework-core           = { git = "https://github.com/logos-blockchain/logos-blockchain-testing.git", tag = "test_ref_04" }
-testing-framework-runner-compose = { git = "https://github.com/logos-blockchain/logos-blockchain-testing.git", tag = "test_ref_04" }
-testing-framework-runner-local   = { git = "https://github.com/logos-blockchain/logos-blockchain-testing.git", tag = "test_ref_04" }
-testing-framework-workflows      = { git = "https://github.com/logos-blockchain/logos-blockchain-testing.git", tag = "test_ref_04" }
+testing-framework-core           = { git = "https://github.com/logos-blockchain/logos-blockchain-testing.git", tag = "test_ref_06" }
```
ntn-x2: Don't we have a more "stable" tag for the testing framework?

tests/src/cucumber/defaults.rs
```diff
@@ -21,24 +24,67 @@ fn set_default_env(key: &str, value: &str) {
...
-    set_default_env("RUST_LOG", "info");
-}
+    set_default_env("LOGOS_BLOCKCHAIN_TESTS_KEEP_LOGS", "true");
+    set_default_env("LOGOS_BLOCKCHAIN_TESTS_TRACING", "true");
 
-pub fn init_node_log_dir_defaults(deployer: DeployerKind) {
-    if std::env::var_os("NOMOS_LOG_DIR").is_some() {
-        return;
-    }
+    std::env::var_os("CUCUMBER_LOG_LEVEL").map_or_else(
```
ntn-x2: This is repeated in more than one place, so it should rather be a constant I think.

tests/src/cucumber/defaults.rs
```diff
@@ -21,24 +24,67 @@ fn set_default_env(key: &str, value: &str) {
...
+        || {
+            set_default_env("LOGOS_BLOCKCHAIN_LOG_LEVEL", "info");
```
ntn-x2: Same for these envs.

tests/src/cucumber/defaults.rs
```diff
@@ -50,6 +96,37 @@ pub fn init_tracing() {
         .try_init();
 }
 
+/// Get the number of retries for failed scenarios from the `CUCUMBER_RETRIES`
+/// environment variable. If the variable is not set, defaults to 2 retries. If
+/// the variable is set to 0, returns None.
+pub fn get_retries() -> Result<Option<usize>, String> {
+    std::env::var_os("CUCUMBER_RETRIES").map_or_else(
```
ntn-x2: Also here. These can very easily get out of sync.

ntn-x2 (commented):
Did not get into the weeds of cucumber, but I left some comments nonetheless.

.github/workflows/cucumber.yml
hansieodendaal: Ok yes, adding the commit sha as well

tests/cucumber_tests/cucumber.rs
```diff
@@ -64,16 +93,34 @@ async fn main() {
...
+                    world.set_scenario_base_dir(&scenario_dir, &deployer);
                 }
             })
         });
+    if let Some(retries) = get_retries()
+        .inspect_err(|e| println!("{e}"))
+        .expect("should parse retries")
+    {
+        // Makes failed Scenarios being retried the specified number of times.
+        world = world.retries(retries);
```
andrussal, line 125: I am not completely sure about this. Sometimes it is slow CI but I guess sometimes could be an real issue?

tests/Cargo.toml
hansieodendaal: I created a tag with a different name, so it does not look like an unstable tag.

tests/cucumber_tests/cucumber.rs
hansieodendaal: Node performance is not consistent - mostly a single run will do, but occasionally, it only passes the 3rd time. Those runs will then be plagued by nodes battling to gain the required chain height.

andrussal: > Node performance is not consistent - mostly a single run will do, but occasionally, it only passes the 3rd time. Those runs will then be plagued by nodes battling to gain the required chain height.

Probably should track somewhere which tests are flaky and at some point look more closely.

tests/src/topology/configs/deployment.rs
```diff
@@ -25,6 +32,16 @@ use time::OffsetDateTime;
 
 use crate::topology::configs::time::{CONSENSUS_SLOT_TIME_VAR, DEFAULT_SLOT_TIME_IN_SECS};
 
+static CHAIN_START_TIMES: LazyLock<Mutex<HashMap<TxHash, OffsetDateTime>>> =
```
ntn-x2: As discussed privately, I don't think we need this synchronization primitive because each test runs in isolation, so it's sufficient for all nodes **within** the same test to use the same time. I don't think we need to keep track of all times ever returned for all threads. This said, I will approve the PR regardless, but perhaps a second opinion on this would be preferred.

tests/cucumber_tests/cucumber.rs
hansieodendaal: Agreed, that is what we can use the tags for. like adding `@flaky`.

youngjoon-lee (approved):
thank you.

PR #2154, opened 2026-02-10
chore(docker): Node docker image use release binaries

bacv:

Node docker file is updated to use release binaries. A github workflow is added to trigger the docker image build process on a new release.

Yes

N/A

N/A

N/A

## Checklist

* [X] 2. Description added.

### DIFF
--- a/.github/workflows/publish-node-image.yml
+++ b/.github/workflows/publish-node-image.yml
@@ -6,7 +6,7 @@ on:
 
 env:
   REGISTRY: ghcr.io
-  IMAGE_NAME: ${{ github.repository }}
+  IMAGE_NAME: logos-blockchain-node
 
 jobs:
   push_to_registries:
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ Currently the `"profiling"` feature is not supported in Windows builds.
 To build the Logos blockchain Docker image, run:
 
 ```bash
-docker build -t logos-blockchain .
+docker build -t logos-blockchain-node .
 ```
 
 #### Command line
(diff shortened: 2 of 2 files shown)


.github/workflows/code-check.yml
```diff
@@ -224,25 +224,3 @@ jobs:
           name: integration-test-artifacts-${{ runner.os }}-${{ github.sha }}
           include-hidden-files: true
           path: tests/.tmp*
-  
-  build-docker:
```
ntn-x2, line 228: I always assumed this job was here to check whether our Docker building process was working fine, and not to actually push images?

.github/workflows/publish-node-image.yml
```diff
@@ -0,0 +1,66 @@
+name: Publish Logos Blockchain Node Docker Image
+
+on:
+  release:
+    types: [published, created]
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
```
ntn-x2: Should we hard-code this instead? We should have control over this in my opinion.

.github/workflows/publish-node-image.yml
ntn-x2, line 1: I think we want to link the Docker image in the release notes, so why not making this a job of the prepare-release workflow? It would run in parallel with the others, and would complete around the same time, so we don't waste any time in that regard. But then when we do manual refinement of the automatically-created release, we have the Docker image to link to already.

.github/workflows/publish-node-image.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Build and push multi-platform Docker images
+        id: push
+        uses: docker/build-push-action@84ad562665bb303b549fec655d1b64f9945f3f91 # Version 6.15.0
+        with:
+          context: .
+          file: ./Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: |
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.ref_name }}
```
ntn-x2, line 57: If we move this to the release workflow, this could be the `VERSION` number that we prepare in the setup job in there. I think it would make things more consistent and all grouped within the same workflow as they are all related to each other.

bacv:
> We dispatch this and the binary building, and then gather all the results in the release notes. What do you think?

It's totally possible to put an image building as a step into the release workflow, but then we'll have a one big job. If we look at the docker image as a release "consumer", then it makes sense to have it separate, and trigger after the release is public.

Is it possible to have couple of workflows for the release?

.github/workflows/publish-node-image.yml
bacv: Do we really want to couple Docker image to the binary release? The release link won't have this in a list of released files for download. This will be in the packages page. How about we consider a docker image as something that is published after a release?

bacv: > I think we want to link the Docker image in the release notes

How about we release the "Dockerfile", but our docker image will be published with the job that is triggered after a release. Just thinking out loud about the possibilities and semantically correct solution.

.github/workflows/publish-node-image.yml
bacv: In this case the version would be the tag, so it should be the same as in the release?

.github/workflows/publish-node-image.yml
bacv: Now it's `logos-blockchain-node`, fixed in https://github.com/logos-blockchain/logos-blockchain/pull/2154/commits/b6ab80036fa5eb53c7b8e893e434e5aedd7f9d77

PR #2155, opened 2026-02-10
feat(pol): derive t0/t1_constant from `f`

youngjoon-lee:

This PR revives the `cryptarchia_engine::Config::slot_activation_coeff` paramater, which is now used to derive [`t0_constant` and `t1_constant`](https://www.notion.so/nomos-tech/v1-1-Proof-of-Leadership-Specification-2e9261aa09df80058244c902defc6da2?source=copy_link#2e9261aa09df8010b506cea919608b8b).

So, now we don't need the feature gate `high-slot-activation-coefficient`.

In this PR, we declare `LotteryConstants` (which contains the derived `t0_constant` and `t1_constant`) as a **static** value using `OnceLock`, in the `lb-pol` crate. This is initialized once before a node binary is started.
This replaces the static `T0_CONSTANT` and `T1_CONSTANT` hardcoded previously.
I also tried the other way: creating a `LotteryConstants` at startup and passing it all the way around (even to the blend service). But it was not so clean.
So, I think, the approach in this PR is simpler and safer.

This PR looks big, but most of diffs are test files.
You can focus on:
- `consensus/cryptarchia-engine/src/config.rs`
- `core/src/proofs/leader_proof.rs`
- `nodes/node/binary/src/config/cryptarchia/deployment.rs`
- `nodes/node/binary/src/lib.rs`
- `nodes/node/standalone-deployment-config.yaml`
- `zk/proofs/pol/src/lottery.rs`

I couldn't split this PR smaller.

I believe so.

spec: @thomaslavaur @davidrusu 
dev: @youngjoon-lee 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,12 +245,6 @@ target/debug/logos-blockchain-node nodes/node/standalone-node-config.yaml
 Node stores its state inside the `store` directory. If there are any issues when restarting the node, please try removing 
 `store` directory.
 
-If you would like to test a node with a more frequent blocks per slot, please usea `--features high-active-slot-coefficient` feature flag when building:
-```bash
-cargo build --all-features --all-targets --features high-active-slot-coefficient
-```
-
-
 #### Running Logos Blockchain Node with integration test
 
 To run the node programatically, one can use `local_testnet_one_node` integration test.


zk/proofs/pol/src/lottery.rs
```diff
@@ -25,52 +15,230 @@ pub static P: LazyLock<BigUint> = LazyLock::new(|| {
...
+            .unwrap(),
+        );
+        assert_eq!(
+            constants.t1_constant,
+            BigUint::from_str_radix(
+                "71e790b4199113a9a00298d823c5716ddac764a110a45fe3b770bbb3e8a57",
+                16,
+            )
+            .unwrap()
+        );
```
youngjoon-lee, line 164: NOTE: These constants have been slightly updated in the [spec](https://www.notion.so/nomos-tech/v1-1-Proof-of-Leadership-Specification-2e9261aa09df80058244c902defc6da2?source=copy_link#2e9261aa09df8010b506cea919608b8b), based on the [code](https://www.notion.so/nomos-tech/v1-1-Proof-of-Leadership-Specification-2e9261aa09df80058244c902defc6da2?source=copy_link#2f7261aa09df8068b842ff0721bfff7b) attached in the spec.

Image

This test checks that the derived constants are the same as what are defined in the latest spec.

utils/src/math.rs
```diff
@@ -252,6 +252,31 @@ impl TryFrom<u64> for F64Ge1 {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(::serde::Serialize, ::serde::Deserialize))]
+pub struct NonNegativeRatio {
+    pub numerator: u32,
+    pub denominator: u32,
+}
+
+impl NonNegativeRatio {
+    pub const fn new(numerator: u32, denominator: u32) -> Result<Self, &'static str> {
+        if denominator == 0 {
```
ntn-x2: I think we can simply define the denominator as a `NonZeroU32`?

zk/proofs/pol/src/lottery.rs
```diff
@@ -25,52 +16,230 @@ pub static P: LazyLock<BigUint> = LazyLock::new(|| {
...
+
+/// Initializes the lottery constants with the given slot activation
+/// coefficient. This must be called exactly once before using
+/// [`lottery_constants`].
+///
+/// # Panics
+/// Panics if called more than once.
+pub fn init_lottery_constants(f: NonNegativeRatio) {
+    LOTTERY_CONSTANTS
+        .set(LotteryConstants::new(f))
```
ntn-x2: What about using `get_or_init` instead here? So we avoid panicking if called more than once?

ntn-x2: Or we can already return the `&'static LotteryConstants` here in case it's needed? Or maybe a `Result` which is an `Err` if called more than once? It's still better than a panic in my opinion.

zk/proofs/pol/src/lottery.rs
```diff
@@ -25,52 +16,230 @@ pub static P: LazyLock<BigUint> = LazyLock::new(|| {
...
-            "44c2a290c72d4dc7d6e514a4b9683cc6e7c15e64f0f59482ec7d3f5906784a",
-            16,
-        )
-        .expect("Constant should parse")
+
+    /// Computes the lottery values t₀ and t₁ for a given total stake.
+    pub fn compute_lottery_values(&self, total_stake: u64) -> (BigUint, BigUint) {
+        let total_stake = BigUint::from(total_stake);
+
+        let lottery_0 = &self.t0_constant / &total_stake;
```
ntn-x2, line 134: We're diving here, should `total_stake` be a `NonZeroU64` instead?

utils/src/math.rs
ntn-x2: Or at least the struct field could be defined as `NonZero`.

ntn-x2 (commented):
Nice! Left some potential improvement suggestions!

zk/proofs/pol/src/lottery.rs
youngjoon-lee: IMO, `set` is safer than `get_or_init`, to prevent any case where we call this function more than once with different `f`. 
Currently, I think, panicking is fine because we should fail early if this function is misused.
Also, we don't have any usecase to get a return value from this function right now.

ntn-x2: My suggestion was to rather return a `Result` in case the function is called more than once instead of panicking directly, but if you think panicking straight away is better, let's go with it.

zk/proofs/pol/src/lottery.rs
```diff
@@ -25,52 +16,230 @@ pub static P: LazyLock<BigUint> = LazyLock::new(|| {
...
+/// coefficient. This must be called exactly once before using
+/// [`lottery_constants`].
+///
+/// # Panics
+/// Panics if called more than once.
+pub fn init_lottery_constants(f: NonNegativeRatio) {
+    LOTTERY_CONSTANTS
+        .set(LotteryConstants::new(f))
+        .expect("Lottery constants can only be initialized once");
+}
```
danielSanchezQ: Imo if we do this they are not constants anymore and should be computed and pass around. I understand that now its complicated and a bigger refactor, so lets go with it. But we should plan to rework this.

zk/proofs/pol/src/lottery.rs
```diff
@@ -25,52 +16,230 @@ pub static P: LazyLock<BigUint> = LazyLock::new(|| {
...
+    LOTTERY_CONSTANTS
+        .get()
+        .expect("Lottery constants must be initialized before use")
+}
+
+/// Lottery approximation constants used for computing t₀ and t₁.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct LotteryConstants {
+    pub t0_constant: BigUint,
+    pub t1_constant: BigUint,
```
danielSanchezQ, line 24: `t0` and `t1` should be enough
```suggestion
    pub t0: BigUint,
    pub t1t: BigUint,
```

zk/proofs/pol/src/lottery.rs
```diff
@@ -25,52 +16,230 @@ pub static P: LazyLock<BigUint> = LazyLock::new(|| {
...
+                hex_str.push('0');
+            }
+        }
+
+        if hex_str.is_empty() {
+            BigUint::ZERO
+        } else {
+            BigUint::parse_bytes(hex_str.as_bytes(), 16)
+                .expect("hex string should parse to BigUint")
+        }
```
danielSanchezQ: Where does this algorithm comes from? Is there not a simplest way to achieve this?

zk/proofs/pol/src/lottery.rs
youngjoon-lee: Actually, these should be `*_constant` as defined in the [spec](https://www.notion.so/nomos-tech/v1-1-Proof-of-Leadership-Specification-2e9261aa09df80058244c902defc6da2?source=copy_link#2e9261aa09df8010b506cea919608b8b). These are different from `t0` and `t1`.

zk/proofs/pol/src/lottery.rs
youngjoon-lee: David shared this code with me, and I checked details. Unfortunately, there's no simpler way in astro-float.

zk/proofs/pol/src/lottery.rs
danielSanchezQ: > Actually, these should be `*_constant` as defined in the [spec](https://www.notion.so/nomos-tech/v1-1-Proof-of-Leadership-Specification-2e9261aa09df80058244c902defc6da2?source=copy_link#2e9261aa09df8010b506cea919608b8b). These are different from `t0` and `t1`.

Ah, I see, thanks!

utils/src/math.rs
youngjoon-lee: Thank you. Done 76b12f385e223d8884616275f883d35463b4232b

zk/proofs/pol/src/lottery.rs
danielSanchezQ: This seems pretty AI to me. Not against it, but would be nice to understand what is going on around. Which is not that easy :sweat_smile:

zk/proofs/pol/src/lottery.rs
youngjoon-lee: Good point. I opened a issue https://github.com/logos-blockchain/logos-blockchain/issues/2166 for this, because I think it's better to address it as a separate PR.

zk/proofs/pol/src/lottery.rs
youngjoon-lee: I will try to eliminate the static variable again and will open a PR on top of this. Since this PR is not so urgent, we can try the ultimate way.

zk/proofs/pol/src/lottery.rs
youngjoon-lee: I will try it as a next PR now without merging this. https://github.com/logos-blockchain/logos-blockchain/pull/2155#discussion_r2792634340

zk/proofs/pol/src/lottery.rs
youngjoon-lee: I just simplified the algorithm db80d0a8 and added comments to explain each step. I think this is the simplest approach we can take right now. Thanks!

utils/src/math.rs
```diff
@@ -252,6 +253,28 @@ impl TryFrom<u64> for F64Ge1 {
...
+            denominator,
+        }
+    }
+
+    #[must_use]
+    pub const fn as_f64(&self) -> f64 {
+        self.numerator as f64 / self.denominator.get() as f64
+    }
+}
+
```
danielSanchezQ, line 277: This would have come in handy for #2199 😢

PR #2159, opened 2026-02-10
chore: move chain start time to deployment config

ntn-x2:

1. Move chain start time from user config to deployment config
2. Remove one of the example config YAML files, and rename the other two to `standalone-{node|deployment}-config.yaml`.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.


nodes/node/binary/src/config/mod.rs
```diff
@@ -414,19 +378,6 @@ pub fn update_http(
     Ok(())
 }
 
-pub fn update_time(time: &mut TimeConfig, time_args: &TimeArgs) -> Result<()> {
```
bacv, line 417: Couldn't these settings be applied to the deployment config? In case starting and stopping the node multiple times for debugging purposes it would be less cumbersome.

ntn-x2: Mhh I am skeptical of adding these flags. Deployments are meant to be shared by all nodes. Would it be too much effort to update the deployment file and provide the path instead?

bacv: The one node config was created specifically for people to be able to run the node locally easily. For debugging purposes, seems that it would be better for people not to join any well known environments, just use one node setup locally. Maybe we could even explain this in the readme.
To iterate quickly, the time flag would allow restart the node without modifying the config on every restart.

Another option would be to remove this flag and env vars, and then generate user and deployment configs with the soon to be implemented `--gen-config` command.

ntn-x2: Yes, that sounds like a better approach. Anything deployment-specific that needs to be regenerated should not happen with a simple CLI flag. I understand your point of view, but at most we can put it behind a flag, if anything. But the real binary should not make this possible. I agree using `gen-config` subcommand would be a good option tho. This PR already removes the flag and I updated the README to remove mentions of it, is that enough?

bacv: Okay, lets go with this approach.

PR #2160, opened 2026-02-10
chore: decouple service configs from node config

ntn-x2:

This PR removes any explicit links between the node-level configuration and the settings of each service.

I left all the service-side code mostly unchanged, although in the future we should remove any `serde` derivations for those, to make sure we don't directly use service-level configs as node configs, so at the moment there is a lot of repeated code, but in the long term all serialization logic + any composition logic will be in the node, with services only ever expecting the data they will use. In my opinion, this is the right thing to do, as a service is mostly concerned about its own settings and is not aware of other related values for other services, and should not really deal with (de)serialization logic or default values. I followed the existing convention to split between user and deployment configs to maintain the same familiar structure.

To ease the transition, I basically copy-pasted the removed serde derivations into the new settings, and I will work on improving these in a follow-up PR, where I plan to modify the new configs to allow for as many sensible defaults as we can, without affecting the code of the downstream services anymore. At the same time, if we see some settings are repeated between services, we can specify them just once in the node config and then pass them down to each service where needed. For instance, we could let each service define the KMS key ID + actual key, and we would convert it behind the scenes into an entry in the settings for the KMS service, so we don't risk specify keys which are not used anywhere. Or still, entries like `cryptarchia.leader.wallet` is semantically equivalent to `sdp.wallet` entries, so those should be specified only once and transferred to the other, as relevant.

I also renamed the `http` config entry to `api`, and added a nested `testing` field that is only required if the `testing` feature is enabled, and allows to keep both testing and non-testing mode under the same key. Once we have default, both of these values would not be required anymore, or any subset of the default values can be specified by the user.

This PR is based on the (personal) opinion that serialization, deserialization and default values should be a concern of the binary and not of each service, since they are meant to simply represent a self-encapsulated piece of logic that takes some settings in input in order to work. Do you guys agree?

Yes. See description above for more context.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nodes/node/http-client/src/lib.rs
+++ b/nodes/node/http-client/src/lib.rs
@@ -2,7 +2,8 @@ use std::sync::Arc;
 
 use futures::{Stream, StreamExt as _};
 use lb_chain_broadcast_service::BlockInfo;
-use lb_chain_service::{CryptarchiaInfo, Slot};
+use lb_chain_service::CryptarchiaInfo;
+pub use lb_chain_service::Slot;
 use lb_core::{
     block::Block,
     header::{ContentId, HeaderId},
--- a/tests/src/tests/zone_sdk/e2e.rs
+++ b/tests/src/tests/zone_sdk/e2e.rs
@@ -93,6 +93,7 @@ async fn test_sequencer_publish_and_indexer_read() {
         node_url.clone(),
         None,
         sequencer_config,
+        None, // Fresh start, no checkpoint
     );
 
     // Publish inscriptions (with retry until sequencer is initialized)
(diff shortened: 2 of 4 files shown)


nodes/node/binary/src/config/kms/serde.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+#[derive(Clone, Debug, Deserialize)]
+#[cfg_attr(any(test, feature = "testing"), derive(serde::Serialize))]
+pub struct PreloadKmsBackendSettings {
+    pub keys: HashMap<KeyId, Key>,
+}
+
+impl From<PreloadKmsBackendSettings>
+    for lb_key_management_system_service::backend::preload::PreloadKMSBackendSettings
```
hansieodendaal: This would read more easily if added to the use statements on top (and below)
```suggestion
    for PreloadKMSBackendSettings
```

nodes/node/binary/src/config/kms/serde.rs
```diff
@@ -0,0 +1,70 @@
+use std::collections::HashMap;
+
+use lb_key_management_system_service::{backend::preload::KeyId, keys::Key};
+use serde::Deserialize;
+
+#[derive(Clone, Debug, Deserialize)]
+#[cfg_attr(any(test, feature = "testing"), derive(serde::Serialize))]
+pub struct Config {
+    pub backend: PreloadKmsBackendSettings,
```
hansieodendaal, line 13: I am curious, why the indirection and not just have `pub keys: HashMap<KeyId, Key>,` directly?

nodes/node/binary/src/config/kms/serde.rs
ntn-x2: Thanks for the review, albeit this is still a draft :) We are using the same name for the local type as well, so the two conflict and we need to disambiguate.

nodes/node/binary/src/config/kms/serde.rs
ntn-x2: To keep it consistent with other places where we groups service-specific settings in the top level, and backend-specific settings behind the `backend` key. So I wanted to keep the same here.

nodes/node/binary/src/config/kms/serde.rs
hansieodendaal: `use PreloadKMSBackendSettings As PreloadKMSBackendSettingsBackend`

ntn-x2: I removed such `From` implementation. Please give another review when you can 😊

PR #2161, opened 2026-02-10
feat(zone-sdk): v0 indexer/sequencer e2e

pradovic:

This PR aims to stabilize the v0 API for the indexer and sequencer. It adds polling cursor based API as well along side the streaming `follow` call. 

It is still based on finalized blocks, but the goal is to expand opaque cursor later to support non finalized blocks as well.

The integration test is added to test the sequencer/indexer flow. Couple of bugs were fixed as well. Also, there is a fix to potential bug in indexing blocks in chain-def.

Yes

@pradovic @davidrusu 

Yes
No
## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/services/chain/chain-service/src/lib.rs
+++ b/services/chain/chain-service/src/lib.rs
@@ -923,19 +923,14 @@ where
             .await
             .map_err(|e| Error::Storage(format!("Failed to store block: {e}")))?;
 
-        let mut immutable_blocks = pruned_blocks.immutable_blocks().clone();
-        // The new LIB is also immutable and should be immediately queryable by slot.
-        // prune_immutable_blocks() only returns blocks older than the new LIB,
-        // so we explicitly add the new LIB here.
-        if prev_lib != new_lib {
-            let new_lib_slot = cryptarchia.consensus.lib_branch().slot();
-            immutable_blocks.insert(new_lib_slot, new_lib);
-        }
-        relays
-            .storage_adapter()
-            .store_immutable_block_ids(immutable_blocks)
-            .await
-            .map_err(|e| Error::Storage(format!("Failed to store immutable block ids: {e}")))?;
+        Self::store_immutable_blocks_index(
+            &pruned_blocks,
+            Some(prev_lib),
+            new_lib,
+            cryptarchia.consensus.lib_branch().slot(),
+            relays.storage_adapter(),
+        )
+        .await?;
 
         let processed_block_event = ProcessedBlockEvent {
             block_id: header.id(),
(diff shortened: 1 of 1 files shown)


services/chain/chain-service/src/lib.rs
```diff
@@ -923,7 +923,14 @@ where
             .await
             .map_err(|e| Error::Storage(format!("Failed to store block: {e}")))?;
```
pradovic, line 925: Hey @danielSanchezQ , @youngjoon-lee   , I found a potential boundary bug in immutable slot indexing when LIB advances. The finalized blocks were not appearing in my integration tests when i was using `cryptarchia/blocks?from=X&to=Y` call.

Proposed fix: when LIB advances, also insert the new LIB into the immutable slot map so it’s immediately queryable by slot. This should be an indexing/storage change only (no consensus semantics).

Is it correct that new LIB should be indexed too?

If this should not be touched now because of the devnet testing, I can probably switch to using just `get_block` and traverse by parent relationship, but I think this call would be useful for back-filling up to LIB.  I can also extract it to a separate PR, if needed.

Thanks so much!

cc. @davidrusu

youngjoon-lee: Thank you. I think this is fine because the `immutable_blocks` variable is only for inserting them to the storage. I think it's also good to extract the line 926-938 as a function (e.g. `store_immutable_blocks`), to not be confused between `immutable_blocks` and `pruned_blocks.immutable_blocks` at the rest of this function. But, it's not important, imo. This change looks good to me.

pradovic: Sounds good, I have applied the suggested refactor: https://github.com/logos-blockchain/logos-blockchain/pull/2161/changes/6716d66a0f997ae4814195e95e2b39e393fdc5a6. 

Thank you!

nodes/node/http-client/src/lib.rs
```diff
@@ -221,6 +222,23 @@ impl CommonHttpClient {
...
+    ) -> Result<Vec<ApiBlock>, Error> {
+        let mut request_url = base_url
+            .join(BLOCKS.trim_start_matches('/'))
+            .map_err(Error::Url)?;
+        request_url
+            .query_pairs_mut()
+            .append_pair("slot_from", &slot_from.to_string())
+            .append_pair("slot_to", &slot_to.to_string());
+        self.get::<(), Vec<ApiBlock>>(request_url, None).await
+    }
```
danielSanchezQ, line 240: We should expose the same API for the c-ffi crate probably. To be done in a different PR ofc.

zone-sdk/src/indexer.rs
```diff
@@ -71,4 +98,231 @@ impl ZoneIndexer {
...
+                        if let Op::ChannelInscribe(inscribe) = op
+                            && inscribe.channel_id == self.channel_id
+                            && let Some(done) =
+                                scan.push_msg(block_slot, inscribe.id(), &inscribe.inscription)
+                        {
+                            return Ok(done);
+                        }
+                    }
+                }
+            }
```
danielSanchezQ, line 177: Nitpick: This could use iterators. But its ok like this too.

zone-sdk/src/sequencer.rs
```diff
@@ -176,6 +176,50 @@ impl ZoneSequencer {
...
+            }
+            Err(e) => {
+                warn!(
+                    "Failed to fetch consensus info: {e}, retrying in {:?}",
+                    reconnect_delay
+                );
+                tokio::time::sleep(reconnect_delay).await;
+            }
+        }
+    }
```
danielSanchezQ, line 201: I think we kinda abuse looping and sleep, when we could transform everything in a stream. But its ok as well.

danielSanchezQ (approved):

zone-sdk/src/sequencer.rs
pradovic: Good point! I will do it in the different PR, I will expose `Sink-Stream` and also eliminate the task spawn.

nodes/node/http-client/src/lib.rs
pradovic: I'm on it (after the zone-sdk PRs done)

PR #2162, opened 2026-02-10
chore: prevent potential panics

hansieodendaal:

This PR prevents potential panics, such as
```
2026-02-09T12:10:37.310768Z ERROR logos_blockchain_tracing::panic: A panic occurred panic.payload="LazyLock instance has previously been poisoned" panic.location="library\\std\\src\\sync\\lazy_lock.rs:415:5" panic.backtrace="   0: <unknown>\n   1: <unknown>\n   2: <unknown>\n   3: <unknown>\n   4: <unknown>\n   5: <unknown>\n   6: <unknown>\n   7: <unknown>\n   8: <unknown>\n   9: <unknown>\n  10: <unknown>\n  11: <unknown>\n  12: <unknown>\n  13: <unknown>\n  14: <unknown>\n  15: BaseThreadInitThunk\n  16: RtlUserThreadStart\n"
```

Yes

@hansieodendaal 

n/a

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/zk/proofs/poc/Cargo.toml
+++ b/zk/proofs/poc/Cargo.toml
@@ -18,7 +18,7 @@ num-bigint           = "0.4"
 serde                = { features = ["derive", "std"], workspace = true }
 serde_json           = { version = "1.0" }
 thiserror            = "2.0"
-tracing              = { default-features = false,  workspace = true }
+tracing              = { default-features = false, workspace = true }
 
 [build-dependencies]
 lb-circuits-utils = { workspace = true }


zk/proofs/poc/src/lib.rs
```diff
@@ -93,7 +96,7 @@ pub fn prove(inputs: &PoCWitnessInputs) -> Result<(PoCProof, PoCVerifierInput),
         serde_json::from_slice(&verifier_inputs).map_err(ProveError::Json)?;
     let proof: Groth16Proof = proof.try_into().map_err(ProveError::Groth16JsonProof)?;
     Ok((
-        CompressedGroth16Proof::try_from(&proof).unwrap(),
```
danielSanchezQ, line 96: No, this should be a expect. If this is panicinc then we are doing something wrong. cc @thomaslavaur

zk/proofs/pol/src/lib.rs
```diff
@@ -93,7 +96,7 @@ pub fn prove(inputs: &PolWitnessInputs) -> Result<(PoLProof, PolVerifierInput),
         serde_json::from_slice(&verifier_inputs).map_err(ProveError::Json)?;
     let proof: Groth16Proof = proof.try_into().map_err(ProveError::Groth16JsonProof)?;
     Ok((
-        CompressedGroth16Proof::try_from(&proof).unwrap(),
+        CompressedGroth16Proof::try_from(&proof)?,
```
danielSanchezQ: Same here, this should never fail. It is an error on our side (coding).

services/tracing/src/lib.rs
```diff
@@ -48,11 +48,23 @@ pub struct SharedWriter {
...
 
     fn flush(&mut self) -> std::io::Result<()> {
-        self.inner.lock().unwrap().flush()
+        self.inner
+            .lock()
+            .unwrap_or_else(|poisoned| {
+                eprintln!("WARNING: Tracing writer mutex poisoned on flush, recovering");
+                poisoned.into_inner()
+            })
+            .flush()
```
danielSanchezQ, line 67: This is nice, thanks.

zk/proofs/poc/src/lib.rs
hansieodendaal: Ok, let's go with that, and evaluate the complete function:
```rust
pub fn prove(inputs: &PoCWitnessInputs) -> Result<(PoCProof, PoCVerifierInput), ProveError> {
    let witness = witness::generate_witness(inputs).map_err(ProveError::Io)?;
    let (proof, verifier_inputs) =
        lb_circuits_prover::prover_from_contents(POC_PROVING_KEY_PATH.as_path(), witness.as_ref())
            .map_err(ProveError::Io)?;
    let proof: Groth16ProofJsonDeser = serde_json::from_slice(&proof).map_err(ProveError::Json)?;
    let verifier_inputs: PoCVerifierInputJson =
        serde_json::from_slice(&verifier_inputs).map_err(ProveError::Json)?;
    let proof: Groth16Proof = proof.try_into().map_err(ProveError::Groth16JsonProof)?;
    Ok((
        CompressedGroth16Proof::try_from(&proof)?,
        verifier_inputs
            .try_into()
            .map_err(ProveError::Groth16JsonInput)?,
    ))
}
```
Are we saying that every `?` should be changed into an `expect` ? Then the function should not return an error, but panic explicitly. 

_Counter arguments:_
- A possible attack vector is that an attacker could create carefully crafted inputs and bring the entire node down.
- I would rather let the call site decide what to do with an error.
- The current approach is safer for production code.

zk/poseidon2/src/hasher.rs
```diff
@@ -10,9 +10,7 @@ pub struct Poseidon2Hasher {
 
 impl Clone for Poseidon2Hasher {
     fn clone(&self) -> Self {
-        Self {
-            state: self.state.to_vec().try_into().unwrap(),
-        }
+        Self { state: self.state }
```
davidrusu: Can this be a derive(Clone)?

hansieodendaal: Yes, quite right, good pick-up - `Fp` implements `Clone`. 
```
pub type Fr = Fp256<MontBackend<FrConfig, 4>>;

pub type Fp256<P> = Fp<P, 4>;

#[derive(Derivative)]
#[derivative(
    Default(bound = ""),
    Hash(bound = ""),
    Clone(bound = ""),
    Copy(bound = ""),
    PartialEq(bound = ""),
    Eq(bound = "")
)]
pub struct Fp<P: FpConfig<N>, const N: usize>(
    pub BigInt<N>,
    #[derivative(Debug = "ignore")]
    #[doc(hidden)]
    pub PhantomData<P>,
);
```

zk/proofs/poc/src/lib.rs
danielSanchezQ: No, just this line with the `unwrap`.  As discussed in discord.

> A possible attack vector is that an attacker could create carefully crafted inputs and bring the entire node down.
There is no attack vector, the current `unwrap` is on purpose, data types transformation should not occur, is an error on our side that could be **avoided**.
> I would rather let the call site decide what to do with an error.
The function should still keep the error, data types that comes from IO are errors. Inner transformations should be infallible.
> The current approach is safer for production code.
This is somehow true, but it may silence error that should just not be there.

zk/proofs/poc/src/lib.rs
```diff
@@ -93,7 +93,9 @@ pub fn prove(inputs: &PoCWitnessInputs) -> Result<(PoCProof, PoCVerifierInput),
...
+        CompressedGroth16Proof::try_from(&proof)
+            .inspect_err(|e| println!("\nSerialization error: {e}\n"))
```
danielSanchezQ: Is this a leftover? These `inspects` should be a debug log instead.
```suggestion
            .inspect_err(|e| debug("Serialization error: {e}"))
```

nodes/api-common/src/bodies/wallet.rs
```diff
@@ -57,7 +63,13 @@ pub mod transfer_funds {
 
     impl IntoResponse for WalletTransferFundsResponseBody {
         fn into_response(self) -> Response {
-            (StatusCode::CREATED, serde_json::to_string(&self).unwrap()).into_response()
+            match serde_json::to_string(&self) {
+                Ok(json) => (StatusCode::CREATED, json).into_response(),
+                Err(e) => {
+                    eprintln!("Failed to serialize WalletTransferFundsResponseBody: {e}");
```
danielSanchezQ: Again, at most this should be a debug log. But I would keep using `expect` instead. It should be impossible for this to fail. It comes from an inner type, if this doesn't work we should now and fix it!

PR #2172, opened 2026-02-12
feat(devnet): Use volumes in docker compose for setup and run

bacv:

It was decided to have two steps when spawning an initial logos environment with a new genesis tx:
- Setup - nodes registers as initial nodes, provide required information, genesis tx is created
- Run - nodes are spawned with a geneis tx from previous step, restarting "Run" step will result in progressing the chain further.

Cfgsync still creates keys for init nodes, it will result in the same genesis if registration params are the same. Registration params will be changed in the future to contain public key information, and cfgsync will only return the genesis_tx instead of full config.

Users can download the deployment settings with current genesis from http://cfgsync/deployment-settings

Yes

N/A

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/testnet/cfgsync/src/repo.rs
+++ b/testnet/cfgsync/src/repo.rs
@@ -28,7 +28,7 @@ pub struct ConfigRepo {
     waiting_hosts: Mutex<HashMap<Host, Sender<RepoResponse>>>,
     generated_user_configs: Mutex<HashMap<Host, GeneralConfig>>,
     deployment_settings: Mutex<Option<DeploymentSettings>>,
-    deployment_settings_storage_path: PathBuf,
+    pub deployment_settings_storage_path: PathBuf,
     n_hosts: usize,
     faucet_settings: FaucetSettings,
     tracing_settings: TracingConfig,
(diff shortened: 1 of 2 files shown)


compose.devnet.setup.yml
```diff
@@ -0,0 +1,83 @@
...
+    image: ghcr.io/logos-blockchain/logos-blockchain:devnet
+    volumes:
+      - node-data:/node-data
+      - ./testnet:/etc/logos-blockchain
+    ports:
+      - "4400:4400/tcp"
+    environment:
+      - CHAIN_START_TIME=${CFG_CHAIN_START_TIME}
+    entrypoint: /etc/logos-blockchain/scripts/setup_cfgsync_server.sh
+
```
hansieodendaal, line 14: Will `CHAIN_START_TIME=$(date -u +"%Y-%m-%dT%H:%M:%SZ")` in `setup_cfgsync_server.sh` not override the `- CHAIN_START_TIME=${CFG_CHAIN_START_TIME}` above?

bacv: It will, the idea with this change was, if we are running the "setup" stage, we set the current time for newly generated genesis. Need to check this part again after merging master, I think location for time is changed now. Then will clean it up, thanks for noticing it.

PR #2173, opened 2026-02-12
chore: use sensible defaults for user config

ntn-x2:

Based on top of https://github.com/logos-blockchain/logos-blockchain/pull/2160, which shows the endgame I wanted to get to. We define defaults for almost all fields, which means the set of required values for the user config goes down dramatically. We can discuss whether these values make sense, otherwise we can go ahead with this for now.

To get an idea, check the `standalone-node-config.yaml` file diff: https://github.com/logos-blockchain/logos-blockchain/pull/2173/changes#diff-d14a2f8aa702edc909edf06a0951640bcd99368bb8a8bc3dbd193d123f854365.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nodes/node/binary/src/config/api/mod.rs
+++ b/nodes/node/binary/src/config/api/mod.rs
@@ -15,11 +15,11 @@ impl ServiceConfig {
     pub fn into_backend_settings(self) -> ApiServiceSettings<AxumBackendSettings> {
         ApiServiceSettings {
             backend_settings: AxumBackendSettings {
-                listen_address: self.user.backend.listen_address,
+                address: self.user.backend.listen_address,
                 cors_origins: self.user.backend.cors_origins,
                 timeout: self.user.backend.timeout,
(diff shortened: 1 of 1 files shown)


nodes/node/binary/src/config/mod.rs
```diff
@@ -410,7 +417,7 @@ pub fn update_api(api: &mut ApiConfig, args: ApiArgs) -> Result<()> {
     let ApiArgs { addr, cors_origins } = args;
 
     if let Some(addr) = addr {
-        api.backend.address = addr;
+        api.backend.listen_address = addr;
```
davidrusu, line 420: was this a bug?

ntn-x2: No, I just renamed the field from `address` to `listen_address`, which makes it clearer what we are dealing with.

PR #2174, opened 2026-02-12
feat(genesis): add CryptarchiaParameter to genesis block

youngjoon-lee:

As defined in the [Genesis Block spec](https://www.notion.so/Bedrock-Genesis-Block-21d261aa09df80bb8dc3c768802eb527?source=copy_link), this PR defines [CryptarchiaParameter](https://www.notion.so/Bedrock-Genesis-Block-21d261aa09df80bb8dc3c768802eb527?source=copy_link#229261aa09df809e9174cb5356d35507) and includes it into the [Genesis Mantle Transaction](https://www.notion.so/Bedrock-Genesis-Block-21d261aa09df80bb8dc3c768802eb527?source=copy_link#229261aa09df80c699e8ec68050230a0).

Now, we don't have `chain_start_time` in the `deployment.yaml`. Instead, we use `CryptarchiaParameter::genesis_time` to initialize the time service.

Also, the chain service uses the `CryptarchiaParameter::epoch_nonce` to initialize the genesis epoch state.

yes

spec: @davidrusu 
dev: @youngjoon-lee 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.


core/src/mantle/genesis_tx.rs
```diff
@@ -157,8 +198,93 @@ impl<'de> Deserialize<'de> for GenesisTx {
...
+    }
+
+    /// Decode the inscription from the ad-hoc binary format.
+    pub fn decode(data: &[u8]) -> Result<Self, Error> {
+        let chain_id_len = u64::from_le_bytes(
+            data.get(..8)
+                .ok_or_else(|| Error::InvalidCryptarchiaParameter("inscription too short".into()))?
+                .try_into()
+                .expect("8-bytes fits in u64"),
+        ) as usize;
```
hansieodendaal: After this we could count the expected number of bytes and error out if it does not match.

Error if data.len() != 8 + chain_id_len + 8 + size_of::<FrBytes>()

nodes/node/standalone-deployment-config.yaml
```diff
@@ -48,7 +48,13 @@ cryptarchia:
...
           payload:
             channel_id: "0000000000000000000000000000000000000000000000000000000000000000"
-            inscription: [ 103, 101, 110, 101, 115, 105, 115 ]  # "genesis" in bytes
+            inscription: [  # CryptarchiaParameter
+              12, 0, 0, 0, 0, 0, 0, 0,                          # chain_id_len = 12 (little-endian u64)
+              108, 111, 103, 111, 115, 45, 100, 101, 118, 110, 101, 116,  # "logos-devnet" (UTF-8)
+              44, 4, 98, 105, 0, 0, 0, 0,                       # genesis_time = 2026-01-10T07:47:56Z (little-endian u64)
+              0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,  # epoch_nonce = [0u8; 32]
+              0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
+            ]
```
hansieodendaal: Could we not specify this in human readable struct format? Would serde not convert it automatically?

hansieodendaal (commented):
Looking good - just some one or two small remarks.

core/src/mantle/genesis_tx.rs
```diff
@@ -157,8 +198,93 @@ impl<'de> Deserialize<'de> for GenesisTx {
...
+
+        let mut buf = Vec::with_capacity(
+            chain_id_len.len() + chain_id.len() + genesis_time.len() + epoch_nonce.len(),
+        );
+        buf.extend_from_slice(&chain_id_len);
+        buf.extend_from_slice(chain_id);
+        buf.extend_from_slice(&genesis_time);
+        buf.extend_from_slice(&epoch_nonce);
+        buf
+    }
```
danielSanchezQ, line 269: We could have use nom for this, and we should, at least to be consisten on how we encode/decode things. Now its fine, but we should open an issue to migrate later on.

core/src/mantle/genesis_tx.rs
```diff
@@ -157,8 +198,93 @@ impl<'de> Deserialize<'de> for GenesisTx {
...
+
+    /// Decode the inscription from the ad-hoc binary format.
+    pub fn decode(data: &[u8]) -> Result<Self, Error> {
```
danielSanchezQ, line 272: Likewise

nodes/node/standalone-deployment-config.yaml
danielSanchezQ: It depends the type @hansieodendaal , if it is bin we cannot do anything, as this is the json/yaml representation

danielSanchezQ (approved):

core/src/mantle/genesis_tx.rs
youngjoon-lee: I replaced this logic with NOM https://github.com/logos-blockchain/logos-blockchain/pull/2185

nodes/node/binary/src/config/time/deployment.rs
```diff
@@ -10,5 +10,5 @@ use time::OffsetDateTime;
 pub struct Settings {
     #[serde_as(as = "MinimalBoundedDuration<1, SECOND>")]
     pub slot_duration: Duration,
-    pub chain_start_time: OffsetDateTime,
+    pub genesis_time: OffsetDateTime,
```
ntn-x2: Mhh if the genesis time is part of the genesis block, then we must remove this config from here altogether. That info will be included in `genesis_state.mantle_tx`.

ntn-x2: If we need the value to create the epoch config, we must parse the genesis block instead. This config here should go as it is not relevant and I'd say dangerous as well.

ntn-x2 (commented):
In general, what adventage does having the start time embedded in the genesis block compared to having it in the deployment config make? It's not a value that can be updated over time. We assume the same deployment config must be used by all nodes. I would think what we want to inscribe on chain are values that can be upgraded over time, so that the binary can parse the new values and update its logic. But stuff that is only used for initialization and values that are not supposed to change over time should probably stay in a deployment file. But of course I do not have context into how we reached this decision.

nodes/node/binary/src/config/time/deployment.rs
youngjoon-lee: Yes, we must remove this from here. I am refactoring the time service for that. I will ping you after marking this PR ready again soon.

hansieodendaal: In my mind, critical blockchain parameters like blockchain start time should not be possible to override by applying a different config; it must be embedded in the binary.

ntn-x2: Well, but what we are doing here is no different than putting it in a deployment config. The deployment config would contain the genesis tx definition, which would include the chain start time. How would you embed the chain start time into a binary that is supposed to be run for different networks?

hansieodendaal: One way to do that is to have a set of consensus parameters for each target network, so the node is compiled for say `testnet-harry-potter` or `mainnet`. For example, `CryptarchiaParameter` will be uniquely defined for each network. The consensus parameters are selected based on the target network. When you start the node, you could have `./logos-blockchain-node ./config/user_config --network testnet-harry-potter`. The network you want to run has to align with the compiled binary.

Opposed to this is having a single binary, and the deployment config defines the network.

ntn-x2: So you mean having one feature (and hence one binary) per network?

hansieodendaal: Yes, more or less. Such a scheme could be:
- `cargo build --release --target-network testnet` -> can run `--network testnet-harry-potter`, `--network testnet-new-amsterdam`, etc.
- `cargo build --release --target-network stagenet` -> can run `--network stagenet` (_e.g. maybe this will be the next mainnet_)
- `cargo build --release --target-network mainnet` can run `--network mainnet` only

If ever we need to introduce domain-separated hashes, those can be keyed to the `--target-network`, so a signature produced in one cannot be validated in the other.

youngjoon-lee: I think it's better to discuss the spec-related topics in Discord: https://discord.com/channels/973324189794697286/1473270782401450227

youngjoon-lee: done [accbc2f](https://github.com/logos-blockchain/logos-blockchain/pull/2174/commits/accbc2fd5da8bb6f768b1081a3a2031d9925525d)

danielSanchezQ (approved):

danielSanchezQ:
> In general, what advantage does having the start time embedded in the genesis block compared to having it in the deployment config make? It's not a value that can be updated over time. We assume the same deployment config must be used by all nodes. I would think what we want to inscribe on chain are values that can be upgraded over time, so that the binary can parse the new values and update its logic. But stuff that is only used for initialization and values that are not supposed to change over time should probably stay in a deployment file. But of course I do not have context into how we reached this decision.

Usually you would want to track this kind of setting in the genesis block. If you do not even agree in genesis you will follow your own fork. I see your point but the few items that we embedded in the block are the bare minimum the network should agree at. Also you can trust and verify the genesis block in a running chain, but not a configuration file.

core/src/mantle/genesis_tx.rs
```diff
@@ -213,8 +268,60 @@ impl<'de> Deserialize<'de> for GenesisTx {
...
+        Ok(Self::decode_by_nom(data)
+            .map_err(|e| Error::InvalidCryptarchiaParameter(format!("Decoding error: {e}")))?
+            .1)
+    }
+
+    fn decode_by_nom(data: &[u8]) -> IResult<&[u8], Self> {
+        let (data, chain_id_len) = decode_uint64(data)?;
```
hansieodendaal, line 279:
```suggestion
    assert!(chain_id_len <= 8, "Fatal error decoding 'CryptarchiaParameter ', chain_id_len out of bounds '{chain_id_len}', expected <= '8'");
    let (data, chain_id_len) = decode_uint64(data)?;
```

hansieodendaal (commented):
Looks good, just one suggestion to guard against memory over-allocation

core/src/mantle/genesis_tx.rs
youngjoon-lee: Thanks. but chain id could be longer than 8 chars. https://www.notion.so/nomos-tech/1-1-0-Bedrock-Genesis-Block-330261aa09df809ab143f87766b8d053?source=copy_link#330261aa09df80fc9ad8f51d2736dded

hansieodendaal (approved):
LGTM

Issue #2170, opened 2026-02-12
Terminate node if IBD is failed


youngjoon-lee:
As per [the diagram in spec](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81f6bf8cf7246f99d36e), the node must be terminated if IBD is not successful. It is also mentioned [here](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81b3b690d33c0aabf9c3) in the spec.

We consider that IBD is successful if all blocks have been fully downloaded from **at least one of the IBD nodes** specfied.

If all downloads from all nodes were failed, we consider that IBD was not successful.

youngjoon-lee:
> [@youngjoon-lee](https://github.com/youngjoon-lee) What is the implication if we set the IBD peer set to the empty set? does that mean that the node should refuse to start?

@davidrusu  I'm also thinking about it. I haven't specified it in the [spec](https://www.notion.so/nomos-tech/Cryptarchia-v1-Bootstrapping-Synchronization-1fd261aa09df81ac94b5fb6a4eff32a6?source=copy_link#1fd261aa09df81bd899ff05e97d66375) yet. Our current implementation skips IBD if the configured IBD peer set is empty, so that genesis nodes can skip IBD without any error.

PR #2179, opened 2026-02-13
fix(ibd): fail IBD if no peer is successfully caught up

youngjoon-lee:

This PR fixes IBD to cover the corner case described in https://github.com/logos-blockchain/logos-blockchain/issues/2170. 

In short,
Even if there is no download remaining, we should fail IBD if no peer has been successfully caught up.

For details of the corner case, please see #2170. I explained the case in detail.

I hope so.

spec: @youngjoon-lee 
dev: @youngjoon-lee 

yes, but need to explicit explain this case.

I will update it now.

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/services/chain/chain-network/src/bootstrap/ibd.rs
+++ b/services/chain/chain-network/src/bootstrap/ibd.rs
@@ -13,7 +13,7 @@ use lb_core::{
 use lb_cryptarchia_sync::GetTipResponse;
 use lb_tx_service::backend::RecoverableMempool;
 use overwatch::DynError;
-use tracing::{debug, error};
+use tracing::{debug, error, warn};
 
 use crate::{
     Error as ChainError, IbdConfig,
@@ -71,18 +71,15 @@ where
     }
 }
 
-// TODO: Replace ProcessBlock closures with a trait
-//       that implements block processing.
-//       https://github.com/logos-blockchain/logos-blockchain/issues/1505
 pub struct InitialBlockDownload<NetAdapter, BlockProcessor, RuntimeServiceId>
 where
     NetAdapter: NetworkAdapter<RuntimeServiceId>,
     NetAdapter::PeerId: Clone + Eq + Hash,
     BlockProcessor: IbdBlockProcessor<NetAdapter::Block>,
 {
-    config: IbdConfig<NetAdapter::PeerId>,
     block_processor: BlockProcessor,
     network: NetAdapter,
+    synced_peers: HashSet<NetAdapter::PeerId>,
     _phantom: PhantomData<RuntimeServiceId>,
 }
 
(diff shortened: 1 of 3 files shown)


services/chain/chain-network/src/bootstrap/ibd.rs
```diff
@@ -71,18 +71,15 @@ where
     }
 }
 
-// TODO: Replace ProcessBlock closures with a trait
-//       that implements block processing.
-//       https://github.com/logos-blockchain/logos-blockchain/issues/1505
```
youngjoon-lee, line 76: this TODO has been already implemented.

services/chain/chain-network/src/bootstrap/ibd.rs
```diff
@@ -121,36 +114,37 @@ where
     /// It downloads blocks from the peers, and applies them to the
     /// [`Cryptarchia`].
     ///
-    /// An updated [`Cryptarchia`] is returned after downloads from
-    /// **all** peers are completed.
+    /// An updated [`Cryptarchia`] is returned after all downloads
+    /// have completed from all peers except the failed ones.
```
youngjoon-lee, line 118: I just polished this comment. I didn't change the logic.

services/chain/chain-network/src/bootstrap/ibd.rs
```diff
@@ -220,15 +217,6 @@ where
         Ok(Some(Download::new(peer, target, stream)))
     }
 
-    async fn should_download(
-        &self,
-        target: &HeaderId,
-        targets_in_progress: &HashSet<HeaderId>,
-    ) -> Result<bool, Error> {
-        Ok(!self.block_processor.has_processed_block(*target).await?
-            && !targets_in_progress.contains(target))
```
youngjoon-lee, line 229: We don't actually need `!targets_in_progress.contains(target)`. This is being already checked by `Downloads::add`.

services/chain/chain-network/src/bootstrap/ibd.rs
```diff
@@ -373,20 +364,32 @@ where
         NetAdapter::Block: 'a,
     {
         match self
-            .initiate_download(peer, latest_downloaded_block, downloads.targets())
+            .initiate_download(peer, latest_downloaded_block)
```
hansieodendaal, line 367: I feel we might have an endless loop somewhere here, although I cannot put my finger on it. Maybe we should add a check like
```rust
    // Check if peer is already synced before trying
    if self.synced_peers.contains(&peer) {
        return Ok(());  // Don't add delay for synced peers
    }
```
when `async fn try_initiate_download<'a>(` starts?

youngjoon-lee: Even if a peer is already in sync, we should try to initiate download **again** after some time, to check if the peer has advanced while we were catching up other peers. That's why we put the peer in the delay queue (by `downloads.add_delay`) at line 376 after marking it as 'synced' inside the `initiate_download` function. After some delay, this `try_initiaite_download` function will be called for this peer again.

The final termination condition is that all peers are in the delay queue.
Then, IBD will return Ok or Err. 
If at least one peer in the delay queue is marked as synced, IBD returns Ok.
Otherwise, it returns Err.

danielSanchezQ (approved):

PR #2199, opened 2026-02-16
chore(settings): Stake inference beta config

danielSanchezQ:
This adds the learning rate (*B*) as a deployment config instead of a constant.

Should be simple

@davidrusu @danielSanchezQ @ntn-x2 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/consensus/cryptarchia-engine/Cargo.toml
+++ b/consensus/cryptarchia-engine/Cargo.toml
@@ -18,7 +18,7 @@ workspace = true
 
 [dependencies]
 cfg_eval   = { optional = true, version = "0.1" }
-lb-utils   = { features = ["time"], workspace = true }
+lb-utils   = { features = ["serde", "time"], workspace = true }
 serde      = { optional = true, workspace = true }
 serde_with = { features = ["macros"], optional = true, workspace = true }
 thiserror  = "1"


ledger/src/cryptarchia/stake.rs
```diff
@@ -1,20 +1,18 @@
...
 
-/// Current learning rate as per [especification](https://nomos-tech.notion.site/Total-Stake-Inference-22d261aa09df8051a454caa46ec54b34), this is not configurable.
-pub const LEARNING_RATE: u64 = 1;
 pub const PRECISION: u64 = 1000;
 
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[derive(Copy, Clone)]
 pub struct StakeInference {
-    learning_rate: u64,
+    learning_rate: f64,
```
ntn-x2, line 8: If this cannot be zero, we have dedicated `F64` types to enforce that. At the very least we should use the `NonNegative` one, but even the `greater-than-zero` one if it makes sense.

nodes/node/binary/src/config/cryptarchia/deployment.rs
```diff
@@ -14,6 +14,7 @@ use serde::{Deserialize, Serialize};
 pub struct Settings {
     pub epoch_config: EpochConfig,
     pub security_param: NonZeroU32,
+    pub learning_rate: f64,
```
ntn-x2: Same as above: do we want to enforce additional constraints, so we fail at deserialization time?

nodes/node/binary/src/config/deployment/devnet.rs
```diff
@@ -34,6 +34,7 @@ cryptarchia:
     epoch_period_nonce_buffer: 3
     epoch_period_nonce_stabilization: 4
   security_param: 20
+  learning_rate: 0.1
```
ntn-x2, line 37: Is it `0.1` or `1.0`?

davidrusu: we settled on 0.1 for devnet

ledger/src/cryptarchia/stake.rs
davidrusu: IMO, if we want to add more constraints to this, don't do it in this PR, lets log a techdebt issue as to not delay the release for devnet today.

danielSanchezQ: Yes, I was searching for those, but only found the ones for integers. I'll check again. Thanks.

nodes/node/binary/src/config/deployment/devnet.rs
danielSanchezQ: its `0.1` for devnet iirc

ledger/src/cryptarchia/stake.rs
danielSanchezQ: > IMO, if we want to add more constraints to this, don't do it in this PR, lets log a techdebt issue as to not delay the release for devnet today.

I left it like that here, ideally we should have a NonZeroF64. Inputs from config uses the NonNegativeF64 now, so it should be fine.

PR #2202, opened 2026-02-16
chore(ledger): Add tracing for epoch state changes

pradovic:
This PR adds simple traces to follow epoch state changes from logs.

Yes

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/ledger/src/cryptarchia/stake.rs
+++ b/ledger/src/cryptarchia/stake.rs
@@ -57,6 +57,17 @@ impl StakeInference {
             / i128::from(PRECISION);
         let new_total_stake_estimate =
             (total_stake_estimate_with_precision - correction) / i128::from(PRECISION);
+
+        tracing::debug!(
+            old_total_stake = total_stake_estimate,
+            new_total_stake = new_total_stake_estimate,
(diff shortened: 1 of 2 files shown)


ledger/src/cryptarchia/stake.rs
```diff
@@ -57,6 +57,17 @@ impl StakeInference {
...
+        tracing::debug!(
+            old_total_stake = total_stake_estimate,
+            new_total_stake = new_total_stake_estimate,
+            measured_density = measured_block_density,
+            expected_density = expected_density_with_precision / i128::from(PRECISION),
+            learning_rate = self.learning_rate,
+            period = self.period(),
+            "TSI update"
+        );
+
```
danielSanchezQ, line 70: I think all this is not necessary, if we log the old stimate and new stimate in the epoch state change should be enough. cc @davidrusu

davidrusu: I think each individual values will be interesting (if there is any problems in TSI, we will have wanted these values printed so we can reconstruct what happened)

So, lets keep it!

PR #2211, opened 2026-02-17
chore: small updates to logging

ntn-x2:

This PR is not meant to be a comprehensive overhaul of our logging logic (which needs to be restructured in terms of what we log, when we log it, and what log level we should use). Nevertheless, these small fixes will make inspecting logs easier as we reduce a lot of noise.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/services/chain/chain-network/src/lib.rs
+++ b/services/chain/chain-network/src/lib.rs
@@ -38,7 +38,7 @@ use overwatch::{
 use serde::{Deserialize, Serialize, de::DeserializeOwned};
 use thiserror::Error;
 use tokio::sync::oneshot;
-use tracing::{Level, debug, error, info, instrument, span};
+use tracing::{Level, debug, error, info, instrument, span, trace};
 use tracing_futures::Instrument as _;
 
 pub use crate::{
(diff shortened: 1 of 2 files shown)


core/src/proofs/leader_proof.rs
```diff
@@ -27,6 +28,20 @@ pub struct Groth16LeaderProof {
...
 
+impl Debug for Groth16LeaderProof {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("Groth16LeaderProof")
+            .field(
+                "proof",
+                &format_args!("{} bytes", size_of::<lb_pol::PoLProof>()),
+            )
+            .field("entropy_contribution", &self.entropy_contribution)
+            .field("leader_key", &self.leader_key)
```
danielSanchezQ, line 38: I think we should not log anything sensible. We could hash the leader key at least.

ntn-x2: The default impl was logging everything already. But yeah I can change that to be a hash.

ntn-x2: Wait this is a public key, so not sensitive.

danielSanchezQ: > Wait this is a public key, so not sensitive.

Ah ok, if it was pk is good to go. I didn't realize, thanks.

PR #2219, opened 2026-02-18
fix(autonat): do not panic if externally confirmed address differs from local address

ntn-x2:

Fix NAT state machine panic when the external IP differs from the local listen address.

The node listens on `0.0.0.0`, so the first `NewListenAddr` event seeds the state machine with `127.0.0.1`. When Kademlia/identify later discovers the real public IP and emits `ExternalAddrConfirmed`, the state machine panicked on the mismatch.

Changes:

- **`TestIfPublic` / `TestIfMappedPublic`** (the two "testing" states): accept a different confirmed address via `into_retarget()` — stay in the same state but re-target to the new address instead of panicking.
- **`Public` / `MappedPublic`** (the two "confirmed" states): downgrade address-mismatch panics to `warn!` + ignore, since a confirmed state should not crash on a stale/duplicate event for a different address.
- **`inner.rs` `poll()`**: remove the `local_address` filter on autonat results — once the state machine retargets to the public IP, the old filter (still holding `127.0.0.1`) would silently drop the autonat result we actually need. The state machine's own match guards handle filtering now.
- **`fixtures.rs`**: expand `all_events()` with all mismatch variants so `other_events_are_ignored` tests cover them.

`Uninitialized`, `TryMapAddress`, and `Private` are unchanged — `TryMapAddress` keeps its panics (mismatches there indicate a mapper bug).

Yes.

Spec - Chris, original implementation - @andrussal, fix - @ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/libp2p/src/behaviour/nat/state_machine/transitions/mapped_public.rs
+++ b/libp2p/src/behaviour/nat/state_machine/transitions/mapped_public.rs
@@ -1,5 +1,3 @@
-use tracing::warn;
-
 use crate::behaviour::nat::state_machine::{
     Command, CommandTx, OnEvent, State, event::Event, states::MappedPublic,
 };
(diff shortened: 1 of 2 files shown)


ntn-x2:
@bacv no worries, I am not going to merge this today. I am also pushing new stuff so it's all good.

PR #2224, opened 2026-02-18
fix(cfgsync): Use cfgsync blend configuration

bacv:

Cfgsync was using test topology and ran blend on localhost.

Run blend on 0.0.0.0

@ntn-x2 @youngjoon-lee 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/testnet/cfgsync/src/config/blend.rs
+++ b/testnet/cfgsync/src/config/blend.rs
@@ -1,5 +1,4 @@
-use core::time::Duration;
-use std::{num::NonZeroU64, str::FromStr as _};
+use std::str::FromStr as _;
 
 use lb_key_management_system_service::keys::{Ed25519Key, ZkKey};
 use lb_libp2p::Multiaddr;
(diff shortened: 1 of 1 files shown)


testnet/cfgsync/src/config/blend.rs
```diff
@@ -0,0 +1,58 @@
...
+        .map(|(id, port)| {
+            let private_key = Ed25519Key::from_bytes(id);
+            // We need unique ZK secret keys, so we just derive them deterministically from
+            // the generated Ed25519 public keys, which are guaranteed to be unique because
+            // they are in turned derived from node ID.
+            let secret_zk_key =
+                ZkKey::from(BigUint::from_bytes_le(private_key.public_key().as_bytes()));
+            (
+                blend::Config {
+                    non_ephemeral_signing_key_id: key_id_for_preload_backend(
```
ntn-x2, line 24: Should we use the defaults for everything else? There's a `with_required_values` method that lets you specify only required inputs, and uses default for everything else, so we make sure the cfg is in sync with what other people get.

ntn-x2: On the built object, we can then update the port number to the one passed in the mapping closure.

PR #2233, opened 2026-02-19
fix(blend): panic on empty membership before core and edge services are gracefully shut down

ntn-x2:

We partially fixed Blend panics on empty membership when the service was started, but we are using the same stream also for future session updates, which means that when the time comes that the membership becomes empty, the proxy service transitions from core or edge to broadcast mode, but core and edge modes are kept around to handle the session transition period. Since they also poll the same stream, we cannot `expect` the stream to be non-empty. This is a fix for it, but the whole thing needs to be thought out again after testnet.

Also, I saw some logs that needed fixing because they were spitting out way too much info, so I did that here as well. Hope that's fine.

I hope so.

@ntn-x2 @youngjoon-lee 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/services/blend/src/membership/mod.rs
+++ b/services/blend/src/membership/mod.rs
@@ -34,8 +34,8 @@ impl<NodeId> MembershipInfo<NodeId> {
         };
         Self {
             membership,
-            session_number,
             zk,
+            session_number,
         }
     }
 }


blend/proofs/src/quota/inputs/prove/private.rs
```diff
@@ -50,12 +52,21 @@ impl Inputs {
...
 
+impl Debug for ProofType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::CoreQuota(_) => f.write_str("ProofType::CoreQuota"),
+            Self::LeadershipQuota(_) => f.write_str("ProofType::LeadershipQuota"),
+        }
+    }
+}
+
```
davidrusu, line 69: 🙏

blend/scheduling/src/message_blend/crypto/core_and_leader/send.rs
```diff
@@ -302,6 +302,6 @@ mod test {
 
         processor.set_epoch_private(new_private_inputs.clone());
 
-        assert_eq!(processor.proofs_generator.1, Some(new_private_inputs));
+        assert!(processor.proofs_generator.1 == Some(new_private_inputs));
```
davidrusu, line 305: 🤔 why?

ntn-x2: Because private inputs does not implement `Debug` anymore, so we cannot use `assert_eq`.

PR #2234, opened 2026-02-19
feat(faucet): exclude faucet from leadership

davidrusu:

Faucet was influencing consensus, we now exclude it from leadership.

1. all core nodes now have the same faucet sk in their kms
2. we exclude faucet form initial total stake calculations
3. faucet server can route to any core node to send funds

@bacv 

NA

NA

## Checklist

* [x] 2. Description added.

### DIFF
--- a/testnet/cfgsync/Cargo.toml
+++ b/testnet/cfgsync/Cargo.toml
@@ -33,7 +33,7 @@ tokio                            = { default-features = false, features = ["macr
 tracing                          = { workspace = true }
 
 [dev-dependencies]
-hex       = { default-features = false, features = ["alloc"], workspace = true }
+hex        = { default-features = false, features = ["alloc"], workspace = true }
 lb-groth16 = { workspace = true }
 
 [[bin]]


ledger/src/cryptarchia/mod.rs
```diff
@@ -430,6 +430,7 @@ impl LedgerState {
         let total_stake = utxos
             .utxos()
             .iter()
+            .filter(|(_, (utxo, _))| config.faucet_pk.is_none_or(|fpk| utxo.note.pk != fpk))
```
danielSanchezQ, line 433: Is there no other way of doing this? This should be part of the faucet code instead and should definitely not interfere with cryptarchia at all.

ledger/src/config.rs
```diff
@@ -9,6 +10,8 @@ pub struct Config {
     pub epoch_config: lb_cryptarchia_engine::EpochConfig,
     pub consensus_config: lb_cryptarchia_engine::Config,
     pub sdp_config: crate::mantle::sdp::Config,
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub faucet_pk: Option<ZkPublicKey>,
```
danielSanchezQ, line 14: I know we are running out of time, but this should absolutely not be the solution.

danielSanchezQ (changes requested):
@bacv Can't we block faucet from http api instead if it interferes or something else. I would rather not touch consensus at all for a patch like this.

PR #2237, opened 2026-02-19
fix(blend): more Blend fixes and chores

ntn-x2:
Fixes logic for epoch transition handling in Blend.

* The secret PoL info stream now yields once per epoch, and it yields not only the secret info but also the public inputs and the epoch number
* The public epoch info stream (provided by the epoch handler) also returns the epoch number, for easier comparison with the secret PoL info stream
* The Blend edge service logic has been much simplified, and is able to handle both public epoch info coming in before secret epoch info and the reverse, which was not the case before (it was assumed the epoch clock would yield before the secret info stream, which we observed was not always true)
* The Blend core edge service has been updated to also handle secret PoL info coming in before the epoch ticking.
* Other small QOL improvement changes

### DIFF
--- a/services/blend/src/core/mod.rs
+++ b/services/blend/src/core/mod.rs
@@ -905,27 +905,10 @@ where
                 (public_info, epoch) = handle_clock_event(clock_tick, blend_config, epoch_handler, &mut crypto_processor, backend, public_info, epoch).await;
             }
             Some(pol_info) = secret_pol_info_stream.next() => {
-                // Update the crypto processor first (which conditionally rotates
-                // the epoch if the PoL info is ahead of the tracked epoch), then
-                // synchronize the backend verifier, tracked epoch, and public info.
-                handle_new_secret_epoch_info(blend_config, &pol_info, &mut crypto_processor, epoch);
-                let new_leader_inputs = LeaderInputs {
-                    pol_ledger_aged: pol_info.poq_public_inputs.aged_root,
-                    pol_epoch_nonce: pol_info.poq_public_inputs.epoch_nonce,
-                    message_quota: blend_config.session_leadership_quota(),
-                    lottery_0: pol_info.poq_public_inputs.lottery_0,
-                    lottery_1: pol_info.poq_public_inputs.lottery_1,
-                };
-                // Keep the backend verifier in sync so it can verify messages
-                // with new-epoch proofs. Without this, the verifier would be
-                // stuck on the old epoch if the PoL info arrives before the
-                // clock tick (since handle_clock_event guards both the crypto
-                // processor and the backend behind `new_epoch > current_epoch`).
-                if pol_info.epoch > epoch {
-                    backend.rotate_epoch(new_leader_inputs).await;
+                if let Some(new_leader_inputs) = handle_new_secret_epoch_info(blend_config, &pol_info, &mut crypto_processor, backend, epoch).await {
+                    epoch = pol_info.epoch;
+                    public_info.epoch = new_leader_inputs;
                 }
-                epoch = pol_info.epoch;
-                public_info.epoch = new_leader_inputs;
             }
             Some(session_event) = remaining_session_stream.next() => {
                 match handle_session_event(session_event, blend_config, crypto_processor, message_scheduler, public_info, recovery_checkpoint, backend, sdp_relay, epoch).await {
@@ -1925,6 +1908,12 @@ where
             new_epoch,
         )) => {
             tracing::debug!(target: LOG_TARGET, "New epoch {new_epoch:?} with nonce {pol_epoch_nonce:?} started");
+            if new_epoch <= current_epoch {
+                return (current_public_info, current_epoch);
+            }
+
+            // Only rotate if the PoL info handler hasn't already advanced
+            // the crypto processor and backend verifier to this epoch.
             let new_leader_inputs = LeaderInputs {
                 message_quota: settings.session_leadership_quota(),
                 pol_epoch_nonce,
@@ -1939,10 +1928,8 @@ where
 
             // Only rotate if the PoL info handler hasn't already advanced
             // the crypto processor and backend verifier to this epoch.
-            if new_epoch > current_epoch {
-                cryptographic_processor.rotate_epoch(new_leader_inputs, new_epoch);
-                backend.rotate_epoch(new_leader_inputs).await;
-            }
+            cryptographic_processor.rotate_epoch(new_leader_inputs, new_epoch);
+            backend.rotate_epoch(new_leader_inputs).await;
 
             (new_public_info, new_epoch)
         }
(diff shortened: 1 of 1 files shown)


blend/scheduling/src/message_blend/provers/core_and_leader/mod.rs
```diff
@@ -83,27 +91,78 @@ where
...
+                tracing::debug!(target: LOG_TARGET, "Stopping old epoch leadership proofs generator until new secret PoL info is provided.");
+            }
+            Ordering::Equal => {
+                tracing::debug!(target: LOG_TARGET, "Leadership proofs generator already on the new epoch, ignoring the new public epoch info received.");
+                self.leader_proofs_generator = Some(leader_proofs_generator);
+            }
+            Ordering::Greater => {
+                panic!(
+                    "Secret PoL info for new epoch should never provide an epoch greater than what the public epoch info returns, as the two should always be yielded together, or at most the secret PoL info would lag behind if the node has no or close to no stake."
+                );
```
youngjoon-lee, line 128: Sorry, I don't fully understand what this log explains. Don't this log need to be similar as the one in line 108? I think the "Secret PoL info" doesn't present in this function. This function only accepts the public epoch info, and we're comparing its epoch with `current_epoch()`, iiuc.

nodes/node/binary/src/generic_services/blend/pol.rs
```diff
@@ -0,0 +1,114 @@
...
+        // processed epoch.
+        Some(Box::new(
+            WatchStream::new(pol_winning_slot_receiver)
+                .filter_map(ready)
+                .scan(None, |processed_epoch, (leader_private, leader_public, epoch)| {
+                    let should_yield_new_epoch = processed_epoch.is_none_or(|processed_epoch| processed_epoch < epoch);
+                    if !should_yield_new_epoch {
+                        return ready(Some(None));
+                    }
+
```
youngjoon-lee, line 71: It's not easy to see the diffs because this file has been moved. I guess this part is the key change? (yielding only for new epoch)

services/blend/src/core/backends/libp2p/swarm.rs
```diff
@@ -360,8 +361,14 @@ where
...
-            tracing::error!(target: LOG_TARGET, "Failed to forward message to blend network: {e:?}");
-            tracing::trace!(counter.failed_outbound_messages = 1);
+            // If we have a single connection, then we will always hit the `NoPeers` error.
+            // In this case it's ok not to log such error, since this function is only
+            // called on FORWARDED messages, not on PUBLISHED ones, for which we want to
+            // know if that is the issue.
+            if !matches!(e, Error::NoPeers) {
+                tracing::error!(target: LOG_TARGET, "Failed to forward message to blend network: {e:?}");
+                tracing::trace!(counter.failed_outbound_messages = 1);
+            }
```
youngjoon-lee, line 371: >  If we have a single connection, then we will always hit the `NoPeers` error.

Is it because the single connection is for the `except` peer?

services/blend/src/core/mod.rs
```diff
@@ -883,13 +902,33 @@ where
...
+                // Keep the backend verifier in sync so it can verify messages
+                // with new-epoch proofs. Without this, the verifier would be
+                // stuck on the old epoch if the PoL info arrives before the
+                // clock tick (since handle_clock_event guards both the crypto
+                // processor and the backend behind `new_epoch > current_epoch`).
+                if pol_info.epoch > epoch {
+                    backend.rotate_epoch(new_leader_inputs).await;
+                }
+                epoch = pol_info.epoch;
+                public_info.epoch = new_leader_inputs;
```
youngjoon-lee: Wdyt about moving this logic into `handle_new_secret_epoch_info`? because they have to be always performed together.

blend/scheduling/src/message_blend/provers/core_and_leader/mod.rs
ntn-x2: Yes so the assumption is that there is always a new epoch tick for new epochs, and there is at most one secret PoL info for new epochs (could be 0 if the node has no stake). This panic catches the impossible case where we get two epoch ticks for secret info, but no epoch tick from the clock tick, which is impossible and we should panic if that ever happens. Should I make the log clearer?

nodes/node/binary/src/generic_services/blend/pol.rs
ntn-x2: Yes sorry, moving it to a different file was easier to follow for me because there was too much in the `mod.rs` file. But yeah, we keep state with `scan` and yield only once per epoch, else we yield `Some(None)` and filter it out later.

services/blend/src/core/backends/libp2p/swarm.rs
ntn-x2: Yes correct. We used to see this log very often if the log has a single connection established. We get the message from that peer, and we try to forward it to all other peers, but since there is no other peer we used to generate this error log. Now this is conditional to whether we are forwarding a message (should not be an error) or publishing a message (should be an error).

blend/proofs/src/quota/inputs/prove/private.rs
```diff
@@ -89,7 +89,7 @@ impl From<ProofOfCoreQuotaInputs> for ProofType {
     }
 }
 
-#[derive(Clone, PartialEq, Eq)]
```
danielSanchezQ, line 92: Nipick: why does this need to be clone or copy? if it contains a secret it would be better it doesnt dot it (if possible.)

nodes/node/binary/src/generic_services/blend/pol.rs
```diff
@@ -0,0 +1,114 @@
...
+            .await
+            .ok()?;
+        let pol_winning_slot_receiver = receiver.await.ok()?;
+        // Return a `WatchStream` that filters out `None`s (i.e., at the very beginning
+        // of chain leader start), and any leader info that belongs to an already
+        // processed epoch.
+        Some(Box::new(
+            WatchStream::new(pol_winning_slot_receiver)
+                .filter_map(ready)
+                .scan(None, |processed_epoch, (leader_private, leader_public, epoch)| {
```
danielSanchezQ, line 66: I think we could extract a method out of this. Can be done on a different pr, we have other priorities. But feels like too much.

blend/scheduling/src/message_blend/provers/core_and_leader/mod.rs
youngjoon-lee: I understand the case. Thank you

services/blend/src/core/mod.rs
ntn-x2: Good point. I moved it https://github.com/logos-blockchain/logos-blockchain/pull/2237/commits/e4e7c6d3d085c226e7991bf5281df08b65e9d064. I also conditionally return the public inputs if they were not already processed by the epoch ticket. Can you double check this one please? Wanna make sure I am not missing anything.

blend/proofs/src/quota/inputs/prove/private.rs
ntn-x2: If it's `Clone`, it's not an issue that it's also `Copy`, I'd say. But yeah, I will re-check the logic later on :)

services/blend/src/core/mod.rs
youngjoon-lee: This looks correct. Thank you.

blend/proofs/src/quota/inputs/prove/private.rs
danielSanchezQ: Yeah, this is something that was not filtered out in a previous review I guess. Thanks.

PR #2239, opened 2026-02-19
fix(cryptarchia): use existing `update_epoch_state` for `epoch_state_for_slot`

### LINKED ISSUE #2229: Use the common logic for skipping empty epochs
We have recently implemented `LedgerState::epoch_state_for_slot` for skipping empty epochs with zero block density. However, we’ve also had another logic in the `LedgerState::update_epoch_state`, which is different from the one we added recently. 

We should have the same logic for both. 

`LedgerState::epoch_state_for_slot` is just for querying the epoch state and doesn’t build a new `LedgerState`. 
Whereas, `LedgerState::update_epoch_state` is called when a block is received, and build a new `LedgerState` for accepting the block. 

This may affect chain-leader service. The chain-leader fetches the epoch state from the chain service, and use it to build a PoL. This epoch state is synthesized by `edgerState::epoch_state_for_slot`.
After that it tries to apply the block to the chain before publishing it to the network. But, applying the block will be failed because the epoch state used in this step is made by `LedgerState::update_epoch_state`.
These two epoch states are different, which result in PoL verification failure.

youngjoon-lee:

Closes #2229 

### Background

Recently, we implemented `LedgerState::epoch_state_for_slot` #2076 to query the epoch state for a given slot, supporting skipping empty epochs. It is very useful, but I realized that we already have `LedgerState::update_epoch_state` which is a superset of `epoch_state_for_slot`.

Moreover, each function had its own bugs. 
- `epoch_state_for_slot`
  - It was ignoring the block density accumulated so far for the current epoch.
- `update_epoch_state`
  - It was not adjusting total stake with zero density for the skipped epochs.

### Solution 

I replaced the function body of `epoch_state_for_slot` with a function call of `update_epoch_state`, because it is what we have heavily used. This technique has been already used in the tests. I guess this is what is intended by the original author. https://github.com/logos-blockchain/logos-blockchain/blob/60ed77d5779af5c4d9c977e6dfbeced8cc48d8b3/ledger/src/cryptarchia/mod.rs#L574

You may be worried about the name `update_...`, but it's fine because all functions in the module create a fresh object.

I also fixed `update_epoch_state` to adjust total stake with zero density for skipped epochs.

### Note

As I mentioned in the issue #2229, this bug can prevent block proposals. 
We use `epoch_state_for_slot` for building a PoL. But, we use `update_epoch_state` for applying the block header (i.e., verifying the PoL). If those two functions return different epoch states, PoL verification will be failed.
In this PR, I added a test for this case.

Also, `cryptarchia::LedgerState` is not so easy to read. I have an idea for refactoring (after Lisbon).

yes

spec: @davidrusu 
dev: @youngjoon-lee 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/ledger/src/cryptarchia/mod.rs
+++ b/ledger/src/cryptarchia/mod.rs
@@ -126,6 +126,11 @@ pub struct LedgerState {
 }
 
 impl LedgerState {
+    /// Synthesizes the epoch state for the given slot.
+    ///
+    /// This function must be called before any other function that updates
+    /// [`LedgerState`]. Otherwise, previously accumulated values (e.g. nonce
+    /// and block density) will be lost.
     fn update_epoch_state<Id>(self, slot: Slot, config: &Config) -> Result<Self, LedgerError<Id>> {
         if slot <= self.slot {
             return Err(LedgerError::InvalidSlot {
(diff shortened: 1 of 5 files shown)


ledger/src/cryptarchia/mod.rs
```diff
@@ -195,11 +199,28 @@ impl LedgerState {
             })
         } else {
             // case 3)
+
+            // First, infer total stake using block density of the current epoch
+            let mut total_stake = self.stake_inference.total_stake_inference::<PRECISION>(
+                self.epoch_state.total_stake,
+                self.block_density.current_block_density(),
+            );
```
davidrusu, line 207: Ah this was missed in the other code path, good catch

davidrusu (approved):
Nice find

PR #2245, opened 2026-02-20
chore(chain-network): increase default `max_orphan_cache_size` to 1000

youngjoon-lee:

As discussed in [Discord](https://discord.com/channels/973324189794697286/1474296507707560081/1474301892245848147), this PR increases the default value of `max_orphan_cache_size` from 5 to 1000.

yes

@youngjoon-lee 

n/a

n/a

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/services/chain/chain-network/src/sync/config.rs
+++ b/services/chain/chain-network/src/sync/config.rs
@@ -3,7 +3,7 @@ use std::num::NonZeroUsize;
 use serde::{Deserialize, Serialize};
 
 const MAX_ORPHAN_CACHE_SIZE: NonZeroUsize =
-    NonZeroUsize::new(5).expect("MAX_ORPHAN_CACHE_SIZE must be non-zero");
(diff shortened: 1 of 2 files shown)


services/chain/chain-network/src/sync/config.rs
```diff
@@ -3,7 +3,7 @@ use std::num::NonZeroUsize;
 use serde::{Deserialize, Serialize};
 
 const MAX_ORPHAN_CACHE_SIZE: NonZeroUsize =
```
ntn-x2, line 5: Out of scope, but do we need this here? If we need to enforce some bounds, we should do so at deserialization time, by having a type that enforces that. For future reference tho. Otherwise we spread out information over too many places.

ntn-x2 (approved):

davidrusu (approved):

services/chain/chain-network/src/sync/config.rs
youngjoon-lee: Right, I was also thinking about it. I will remove this later.

PR #2250, opened 2026-02-20
feat(chain-network): DRAFT::::: delay future block proposals within margin

youngjoon-lee:

Closes https://github.com/logos-blockchain/logos-blockchain/issues/2243
Please see the detailed background in the issue.

This PR introduces the following changes:
- The chain service still reject all future blocks **strictly** (no margin).
- But, the chain-network service checks if the block slot is within the acceptable margin, if it receives the `FutureBlock` error from the chain service.
- If the block slot is within the margin, it is stored in the `FutureProposalDelayer`. Otherwise, it is just rejected.
- `FutureProposalDelayer` returns blocks once their slots become the current slot. These blocks are rescheduled, so that the chain-network service can process them soon.
- This delaying mechanism works only for block proposals received from the network. We don't use this for IBD, orphan handling, and proposals from the local node. 

I believe so

spec: @davidrusu 
dev: @youngjoon-lee 

This delaying mechanism is not defined in the [Block Header Validation](https://www.notion.so/nomos-tech/Cryptarchia-v1-Protocol-Specification-21c261aa09df810cb85eff1c76e5798c?source=copy_link#21c261aa09df810bb539f80ba66dba13) spec. But, I don't think this should be added to the spec because this is a behavior that each implementer can choose.

I don't think so.

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/services/chain/chain-network/src/future.rs
+++ b/services/chain/chain-network/src/future.rs
@@ -1,0 +1,150 @@
+use std::time::Duration;
+
+use lb_chain_service::Slot;
+use tokio::sync::mpsc;
+use tracing::{error, warn};
+
+use crate::LOG_TARGET;
+
+pub struct FutureProposalDelayer<Proposal> {
+    future_slot_margin: Slot,
+    slot_duration: Duration,
+    sender: mpsc::Sender<Proposal>,
+}
+
+impl<Proposal> FutureProposalDelayer<Proposal> {
+    pub const fn new(
+        future_slot_margin: Slot,
+        slot_duration: Duration,
+        sender: mpsc::Sender<Proposal>,
+    ) -> Self {
+        Self {
+            future_slot_margin,
+            slot_duration,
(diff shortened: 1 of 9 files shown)


services/chain/chain-network/src/future.rs
```diff
@@ -0,0 +1,150 @@
...
+        let channel = self.sender.clone();
+        tokio::spawn(async move {
+            tokio::time::sleep(delay).await;
+            if let Err(err) = channel.send(proposal).await {
+                error!(target: LOG_TARGET, ?err, "Failed to reschedule delayed proposal");
+            }
+        });
+
+        true
+    }
```
danielSanchezQ: I think its better if instead of a mpsc channel per se, we wrapp it in a stream/sink which process this delay automatically. That way we should be able to avoid spawning tasks for this as well.

PR #2252, opened 2026-02-20
fix(libp2p): don't update static address in static mode

pradovic:

It looks like we were overwriting the static external address in NAT static mode. This PR removes the branch where the external address was being overwritten, potentially by local address (if behind a NAT, for example). 

More context:
```
192.168.x.x (private LAN):
  Dialing address: /ip4/192.168.5.77/udp/3002/quic-v1/p2p/12D3KooWBGAS8KazCze9xukok6iFEVBZmro88rqY2kRZWGcAxFZJ
  Failed to dial using libp2p_core::transport::map::Map<libp2p_quic...

  127.0.0.1 (localhost):
  Dialing address: /ip4/127.0.0.1/udp/3000/quic-v1/p2p/12D3KooWNWgWAzkvH95ucQitsSmddm7JFtQAsufaBUVS7GB6s6qu
  Failed to dial using libp2p_core::transport::map::Map<libp2p_quic...

  100.x.x.x (CGNAT/Tailscale):
  Dialing address: /ip4/100.107.205.121/udp/3000/quic-v1/p2p/12D3KooWNWgWAzkvH95ucQitsSmddm7JFtQAsufaBUVS7GB6s6qu
  Failed to dial using libp2p_core::transport::map::Map<libp2p_quic...

  10.x.x.x (private):
  Dialing address: /ip4/10.114.0.3/udp/3000/quic-v1/p2p/12D3KooWNWgWAzkvH95ucQitsSmddm7JFtQAsufaBUVS7GB6s6qu
  Failed to dial using libp2p_core::transport::map::Map<libp2p_quic...

  WrongPeerId (connecting to self via localhost):
  Connection attempt to peer failed with WrongPeerId { obtained: PeerId("12D3KooWJzcJ78thxyFyDbda7L5fkj5R2KdEmPveWSME6snyjd1A"), endpoint: Dialer { address: /ip4/127.0.0.1/udp/3000/quic-v1/p2p/12D3KooWKbgpSnw98ru3XgwFscwv4dYSkQnYXfq9tAwYtE6J2tZt
  ```
  
  Nodes were advertising their local addresses along with the public one. This should fix local peer advertisement noise.

Yes

@pradovic 

Yes
No
## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/libp2p/src/behaviour/nat/mod.rs
+++ b/libp2p/src/behaviour/nat/mod.rs
@@ -5,8 +5,8 @@ use libp2p::{
     Multiaddr, PeerId, autonat,
     core::{Endpoint, transport::PortUse},
     swarm::{
-        ConnectionDenied, ConnectionId, FromSwarm, NetworkBehaviour, NewListenAddr, THandler,
-        THandlerInEvent, THandlerOutEvent, ToSwarm,
+        ConnectionDenied, ConnectionId, FromSwarm, NetworkBehaviour, THandler, THandlerInEvent,
+        THandlerOutEvent, ToSwarm,
         behaviour::toggle::{Toggle, ToggleConnectionHandler},
     },
 };
@@ -126,9 +126,9 @@ impl<Rng: RngCore + 'static> NetworkBehaviour for Behaviour<Rng> {
     fn on_swarm_event(&mut self, event: FromSwarm) {
         if let Some(inner_behaviour) = self.inner_behaviour.as_mut() {
             inner_behaviour.on_swarm_event(event);
-        } else if let FromSwarm::NewListenAddr(NewListenAddr { addr, .. }) = event {
-            self.static_listen_addr = Some(addr.clone());
         }
+        // Static mode: don't overwrite configured external_address with listen
+        // addresses.
     }
 
     fn on_connection_handler_event(

### CODE REVIEW

File: libp2p/src/behaviour/nat/mod.rs [hunk 1]
```diff
@@ -126,9 +126,9 @@ impl<Rng: RngCore + 'static> NetworkBehaviour for Behaviour<Rng> {
     fn on_swarm_event(&mut self, event: FromSwarm) {
         if let Some(inner_behaviour) = self.inner_behaviour.as_mut() {
             inner_behaviour.on_swarm_event(event);
-        } else if let FromSwarm::NewListenAddr(NewListenAddr { addr, .. }) = event {
-            self.static_listen_addr = Some(addr.clone());
         }
```
danielSanchezQ, line 131: I think this could be improved in 2 ways. Check if there was a previous listen_addr before setting. Also checking if the new address is a valid public ip (at least manually filtering the addr masks we know are local.)
Because the problem here is that we completely remove the nat behaviour otherwise?

pradovic: Just to clarify the context: this path is for static NAT mode (we already have a configured external address). The issue I was trying to avoid is that we also pick up Identify observed_addr candidates and end up setting/advertising them one-by-one (via the poll()/take() flow), which in practice included ~3 extra private/loopback addrs in addition to the configured static one.

I can make the change safer in the two ways you mention:
1. only update static_listen_addr if it wasn’t set yet (or if we explicitly decide to allow replacing it), and
2. filter out known non-public ranges/masks (loopback/private/link-local/etc.) before accepting an addr.

Question on intent though: in static mode, do we actually want to advertise any Identify-derived observed address if we already have an explicit external address configured? My worry is that `observed_addr` is “what the remote saw for the outbound connection”, not “publicly reachable on my listen port”, so it can be misleading (ephemeral port / non-reachable), unless we explicitly validate it (AutoNAT-style probing).

Happy to implement whichever policy you prefer (keep only configured static, or allow Identify-derived but filtered + only-if-empty, etc.).

danielSanchezQ: Ah, no! in static mode it should always be the same static address, thanks!

PR #2257, opened 2026-02-23
test(framework): fork detection

andrussal:

1. Block feed upgraded from single-block view to cluster-aware view by querying all nodes 
2. New fork-monitor expectation
3. New test with fork-monitor expectation

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -678,8 +678,10 @@ impl<ProofsVerifier, ObservationWindowClockProvider>
             .entry(peer_id)
             .or_default();
         if !exchanged_message_identifiers.insert(*message_id) {
-            tracing::debug!(target: LOG_TARGET, "Neighbor {peer_id:?} on connection {connection_id:?} sent us a message previously already exchanged. Marking it as spammy.");
-            self.close_spammy_connection((peer_id, connection_id), SpamReason::DuplicateMessage);
+            tracing::debug!(
```
ntn-x2: Mhh the spec used to specify that a duplicate message must result in the neighbor being marked as spammy. Now this seems to not be the case in the spec anymore. I opened a thread https://discord.com/channels/973324189794697286/1475384197161160834.

ntn-x2: If messages are indeed ignored, we can also get rid of the `SpamReason::DuplicateMessage` variant.

tests/src/tests/testing_framework/fork_detector.rs
```diff
@@ -0,0 +1,139 @@
...
+    let swarm_port = run_config.user.network.backend.swarm.port;
+    let external_addr = format!("/ip4/127.0.0.1/udp/{swarm_port}/quic-v1")
+        .parse()
+        .expect("static NAT external address should be valid multiaddr");
+
+    run_config.user.network.backend.swarm.nat = NatConfig::Static {
+        external_address: external_addr,
+    };
+
+    "pool.ntp.org:123".clone_into(&mut run_config.user.time.backend.server);
```
ntn-x2: This is already the default value so no need to specify it here.

ntn-x2 (commented):
Looks good, minus the duplicate message logic which I want to clarify first. If I can ask, how did you jump to the conclusion that that piece of logic was causing forking?

ntn-x2:
`NoPeers` is actually expected. I changed that in an upcoming PR. If a node with a single established connection receives a message, it will try to forward it to all other peers. Since there is no other peer, it will log a `NoPeers` error, but it's not really an error in this case. As I said, it was fixed to only log it in case a message is broadcasted for the first time, and there's no peers.

Anyway, if the spec actually does not mention banning peers anymore, we can go ahead with this change anyway. But I want @madxor to confirm this first.

Cargo.toml
```diff
@@ -142,9 +142,9 @@ lb-witness-generator               = { default-features = false, package = "logo
 lb-zksign                          = { default-features = false, package = "logos-blockchain-zksign", path = "./zk/proofs/zksign" }
 lb-zone-sdk                        = { default-features = false, package = "logos-blockchain-zone-sdk", path = "./zone-sdk" }
 lb_network                         = { default-features = false, package = "logos-blockchain-network-service", path = "./services/network" }
-testing-framework-core             = { default-features = false, git = "https://github.com/logos-blockchain/logos-blockchain-testing.git", tag = "tagged_ref_04" }
+testing-framework-core             = { default-features = false, git = "https://github.com/logos-blockchain/logos-blockchain-testing.git", rev = "ef0f989" }
```
andrussal: > Where is this commit? We need a tag.

It is currently on PR branch - `scenario/fail-fast-option`. I'll tag after merge.

Cargo.toml
```diff
@@ -142,9 +142,9 @@ lb-witness-generator               = { default-features = false, package = "logo
...
 testing-framework-env              = { default-features = false, git = "https://github.com/logos-blockchain/logos-blockchain-testing.git", tag = "tagged_ref_04" }
-testing-framework-runner-local     = { default-features = false, git = "https://github.com/logos-blockchain/logos-blockchain-testing.git", tag = "tagged_ref_04" }
+testing-framework-runner-local     = { default-features = false, git = "https://github.com/logos-blockchain/logos-blockchain-testing.git", rev = "ef0f989" }
```
hansieodendaal: Where is this commit? We need a tag.

hansieodendaal:
> It seems like we get the same message back what we sent and mark it duplicate? Is this possible?

It could very well be the libp2p layer doing this. It is not uncommon in distributed systems that the same message will be relayed by another path, so duplicates must be handled and are not necessarily malicious.

What works well is if you count the duplicates, then it would be easy to determine if it is some sort of an attack.

PR #2261, opened 2026-02-23
fix(blend): race condition when evaluating duplicate messages

ntn-x2:

There is a race condition in which a peer A can send the same message to peers B and C, and peers B and C try to send that message to each other. The result will be that they flag each other as spammy, since peer B does not expect peer C to send the same message back and viceversa.

This issue is solved by introducing a small sensitivity threshold of (currently) 100ms to allow for this case to be gracefully handled: if the message from a peer is received within this window, the message is simply discarded. If it happens after this window, the peer is considered malicious. We can increase this window if we need, but for now it should fix most issues.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/blend/network/src/core/with_core/behaviour/tests/message_handling.rs
+++ b/blend/network/src/core/with_core/behaviour/tests/message_handling.rs
@@ -249,7 +249,6 @@ async fn duplicate_message_within_sensitivity_interval_is_not_spam() {
         .force_send_message_to_peer(&test_message, *dialing_swarm.local_peer_id())
         .unwrap();
 
-    // Peers disconnect from each other without being marked as spammy.
     loop {
         select! {
             () = sleep(Duration::from_secs(1)) => {


blend/network/src/core/with_core/behaviour/tests/message_handling.rs
```diff
@@ -193,6 +212,96 @@ async fn duplicate_message_received() {
...
+    // dialer. This simulates a race condition where both peers independently
+    // forward the same message to each other near-simultaneously. Because the
+    // duplicate arrives within the `SENSITIVITY_INTERVAL_FOR_DUPLICATES`, the
+    // dialer should silently drop it without flagging the listener as malicious.
+    listening_swarm
+        .behaviour_mut()
+        .force_send_message_to_peer(&test_message, *dialing_swarm.local_peer_id())
+        .unwrap();
+
+    // Peers disconnect from each other without being marked as spammy.
```
youngjoon-lee: I think this comment is not correct? Do peers disconnect from each other in this test?

ntn-x2: Ah yeah I was testing for that before but then changed to simply check the state of each swarm, which is faster. I removed the comment, thanks! I'll wait for this CI run to complete before pushing and merging.

PR #2272, opened 2026-02-24
fix: blend session rotation issues

ntn-x2:

When a new session starts, we were re-creating the message handler, which by default did not set any secret PoL info and was then unable to generate leadership proofs. This is now changed since we keep track of the latest secret PoL info and we use it when creating a new message handler.

As a separate issue, we deny incoming connections from peers from which we already have an incoming connection.

Yes.

@ntn-x2 @youngjoon-lee 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/blend/network/src/core/with_core/behaviour/mod.rs
+++ b/blend/network/src/core/with_core/behaviour/mod.rs
@@ -352,6 +352,7 @@ impl<ProofsVerifier, ObservationWindowClockProvider>
 
         let serialized_message = serialize_encapsulated_message(message);
         let mut at_least_one_receiver = false;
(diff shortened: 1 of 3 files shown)


blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -726,7 +741,19 @@ impl<ProofsVerifier, ObservationWindowClockProvider>
             .is_some_and(|remote| remote.role.is_listener())
     }
 
-    /// Return `true` if there is at least one outbound connection pending
+    /// Return `True` if there is at least one inbound connection pending
+    /// upgrade with the provided peer.
+    // TODO: Find a different data structure to be able to perform this check in
+    // O(1).
```
danielSanchezQ, line 748: This should be an issue.

danielSanchezQ (approved):

PR #2280, opened 2026-02-24
fix(sdp): bump declaration nonce correctly and use locked note for signing active msg

youngjoon-lee:

Previously, we have been submitting blend activity messages, but all of them were being rejected (without any logs).

This PR fixes the issues and adds logs.

**Issue 1**: Wrong nonce was used for activity messages.
I added the following log and confirmed that the 1st activity message was rejected because `message_nonce` was not larger than `declaration_nonce`.
```
failed to apply SDP active message err=Invalid sdp message nonce: message_nonce=0, declaration_nonce=0
```
The `message_nonce` must start from 1 because it is always expected to be larger than `declaration_nonce`.
Also, `declaration_nonce` must be updated every time a SDP message is accepted.
These are defined in the [spec](https://www.notion.so/nomos-tech/Service-Declaration-Protocol-1fd261aa09df819ca9f8eb2bdfd4ec1d?source=copy_link#21b261aa09df80cc82afe94cc3fcdea6).

**Issue 2**: Wrong keys were used for signing activity messages.
After fixing the issue 1, I saw another error, saying that the signature is invalid. It's because one key (locked note) was missing when signing on activity messages.
```
failed to apply SDP active message err=Invalid signature
```
I'm not sure if we really need the locked note key for signing on activity messages. It is not clearly defined in the [spec](https://www.notion.so/nomos-tech/Service-Declaration-Protocol-1fd261aa09df819ca9f8eb2bdfd4ec1d?source=copy_link#21b261aa09df80cc82afe94cc3fcdea6). If it is not really necessary, we can remove it from the signature verification logic instead of adding it to the signature generation. We need a confirmation from @madxor.
Plus, the locked note key was not registered in KMS in the testing environment (and cfgsync). This PR also fixes it. 

Anyway, because of these two issues, blend nodes were kicked out from the membership after 2 sessions (inactivity + retention periods).

I hope so

spec: @madxor 
dev: @youngjoon-lee 

Signing spec must be defined (if not already).
The spec of [Activity Message](https://www.notion.so/nomos-tech/Service-Declaration-Protocol-1fd261aa09df819ca9f8eb2bdfd4ec1d?source=copy_link#1fd261aa09df81f1a0b5cc8956c70660) says that the message must be signed by `zk_id`. It doesn't mention anything about locked note. But, the implementation of signature verification was expecting that the locked note key is also used for signing. I am not sure which one is correct.
Also, the spec of [Withdraw Message](https://www.notion.so/nomos-tech/Service-Declaration-Protocol-1fd261aa09df819ca9f8eb2bdfd4ec1d?source=copy_link#1fd261aa09df817bad97f88bc40ba0c9) says that the message must be signed by `zk_id`. But, the implementation of signature generation/verification is using both locked note and zk id. 

same as above

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/testnet/cfgsync/src/config/mod.rs
+++ b/testnet/cfgsync/src/config/mod.rs
@@ -424,13 +424,13 @@ mod cfgsync_tests {
 
         assert!(faucet_pk.is_some(), "Faucet PK should be set");
 
-        // All nodes should have 5 keys (blend_signing, blend_zk, known, funding,
-        // faucet) and the faucet key in other_keys, so the faucet service can
-        // route to any node.
+        // All nodes should have 6 keys (blend_signing, blend_zk, blend_locked_note,
+        // known, funding, faucet) and the faucet key in other_keys, so the
+        // faucet service can route to any node.
         for host in &hosts {
             let config = configs.get(host).expect("Config missing for host");
             let kms_keys = &config.kms_config.backend.keys;
-            assert_eq!(kms_keys.len(), 5, "Node should have faucet key in KMS");
+            assert_eq!(kms_keys.len(), 6, "Node should have faucet key in KMS");
 
             assert_eq!(
                 config.consensus_config.other_keys.len(),


services/sdp/src/lib.rs
```diff
@@ -323,4 +319,15 @@ where
...
+    /// Increments the nonce of the current declaration, and returns the
+    /// incremented nonce.
+    ///
+    /// Nonce must be incremented first because it is initialized to 0 with the
+    /// declaration, and each SDP message (activity or withdrawal) must have
+    /// nonce larger than the previous one.
+    const fn bump_nonce(&mut self) -> u64 {
+        self.nonce += 1;
+        self.nonce
+    }
```
hansieodendaal, line 332: If the nonce must always increase, should it not be persisted for restarts?

youngjoon-lee: As I mentioned in the TODO comment, I think it must be obtained from the declaration stored in ledger, but I'm not sure what the best way is. This PR doesn't change the architecture to not break anything else. But, yes, we have to refactor this module soon.

pradovic: So basically, in the current design we need to start the node once, post declaration, add this id to the settings and restart the node. I am not sure are we doing this step at the moment.

pradovic: Confirmed here: https://discord.com/channels/973324189794697286/1475892674081980498/1476144050242453524.

Regarding the potential fix I think the decent solution might be to query the ledger on run rather than persisting separately.

The ledger already tracks everything we need in Declaration:
  - zk_id, locked_note_id, nonce (updated on each accepted message)

 Proposed follow-up:
  1. Change SdpSettings.declaration to just store Option<DeclarationId>
  2. Add GetDeclaration query to WalletAdapter (wallet already has ledger access)
  3. On SdpService::run(), query ledger for full declaration + current nonce

This way restart automatically gets the correct nonce from ledger state - no separate persistence needed.
I am not sure if we should do it March, maybe better to not change anything at the moment? I can try to do it today in a separate PR if you think it makes sense.

pradovic (approved):
LGTM + comment about persisting nonce.

services/sdp/src/lib.rs
youngjoon-lee: Yes. That’s what is in my mind, eliminating `self.nonce` from SDP service, as I mentioned in the TODO comments.

I will do it now.

pradovic: We agreed on discord that we can go without the persistence fix for today. I will open the PR for persistence that we can review before another run.

danielSanchezQ (approved):

PR #2283, opened 2026-02-25
chore(blend): do not panic on unexpected case and add more logs to proof generation and verification for easier matching

ntn-x2:

* Sometimes, counterintuitively, a block proposal can come in before the new epoch PoL info is received by Blend. We will investigate why this happens, but for now we avoid panicking in Blend edge and there will simply be no block proposed. This is a very narrow window that can happen if a node wins the lottery for a very early slot of a new epoch. Nevertheless, this will be investigated.
* Let's add some more logs (which we will remove or demote to trace later on) to print what inputs are being used when generating and verifying a PoQ, so that we can match exactly the flow of a proof and link them across nodes thanks to the key nullifier.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/services/blend/src/edge/mod.rs
+++ b/services/blend/src/edge/mod.rs
@@ -393,6 +393,7 @@ where
                 }
             }
             Some(message) = incoming_message_stream.next() => {
+                // TODO: Investigate why secret PoL info at times arrives after the block proposal.
                 let Some(handler) = current_pol_info_and_message_handler.as_mut().map(|(_, handler)| handler) else {
                     tracing::warn!(target: LOG_TARGET, "Received a message to blend, but no active message handler is available to process it because the secret PoL info for the current epoch is not yet available. Ignoring the message.");
                     continue;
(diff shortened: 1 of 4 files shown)


danielSanchezQ (approved):

services/blend/src/edge/mod.rs
```diff
@@ -393,7 +393,11 @@ where
                 }
             }
             Some(message) = incoming_message_stream.next() => {
-                let handler = current_pol_info_and_message_handler.as_mut().map(|(_, handler)| handler).expect("Message handler should be available at the time a new message is propagated.");
+                // TODO: Investigate why secret PoL info at times arrives after the block proposal.
+                let Some(handler) = current_pol_info_and_message_handler.as_mut().map(|(_, handler)| handler) else {
+                    tracing::warn!(target: LOG_TARGET, "Received a message to blend, but no active message handler is available to process it because the secret PoL info for the current epoch is not yet available. Ignoring the message.");
```
hansieodendaal, line 398: If it was serious enough to panic previously, this should be `error`.
```suggestion
                    tracing::error!(target: LOG_TARGET, "Received a message to blend, but no active message handler is available to process it because the secret PoL info for the current epoch is not yet available. Ignoring the message.");
```

PR #2286, opened 2026-02-25
chore: remove explicit deployment setting for Blend session and round duration

ntn-x2:

At the moment, things are intertwined.

Cryptarchia deployment settings need to fill the SDP settings to evaluate service activity proofs. This setting takes a `blocks_per_session` value, which indicates after how many blocks a new session should be yielded to consumers of the session stream. This value makes a bunch of assumptions about the expected average block length and is directly dependent on the slot time.

Nevertheless, in Blend, we used to take the round duration and the round per sessions as separate values. These values are used to compute stuff like expected number of cover messages, but do not affect directly how often sessions are changed, from Blend perspective. This means that using a different and incompatible value for Blend session and round duration leads to undefined behaviour, since sessions would still follow the `blocks_per_session` schedule, but the Blend service believes sessions are longer/shorter, and this can get messy.

So this PR removes the explicit mention of  `round_duration` and `rounds_per_session` from the Blend deployment settings, and sets them respectively to `slot_duration` and to the result of computing how long a session would last divided by the duration of a round, which is `1` as well in this case.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nodes/node/binary/src/config/cryptarchia/deployment.rs
+++ b/nodes/node/binary/src/config/cryptarchia/deployment.rs
@@ -28,7 +28,7 @@ pub struct Settings {
 
 impl Settings {
     #[must_use]
-    pub fn slots_per_epoch(&self) -> u64 {
+    pub const fn slots_per_epoch(&self) -> u64 {
         epoch_length(
             self.epoch_config.epoch_stake_distribution_stabilization,
             self.epoch_config.epoch_period_nonce_buffer,
--- a/consensus/cryptarchia-engine/src/time.rs
+++ b/consensus/cryptarchia-engine/src/time.rs
@@ -1,8 +1,4 @@
-use std::{
-    num::{NonZero, NonZeroU64},
-    ops::Add,
-    time::Duration,
-};
+use std::{num::NonZero, ops::Add, time::Duration};
 
 #[cfg(feature = "serde")]
 use lb_utils::bounded_duration::{MinimalBoundedDuration, SECOND};
--- a/ledger/src/config.rs
+++ b/ledger/src/config.rs
@@ -26,7 +26,7 @@ impl Config {
     }
 
     #[must_use]
-    pub fn epoch_length(&self) -> u64 {
+    pub const fn epoch_length(&self) -> u64 {
         self.epoch_config
             .epoch_length(self.consensus_config.base_period_length())
     }


nodes/node/binary/src/config/blend/deployment.rs
```diff
@@ -1,48 +1,134 @@
...
+    #[must_use]
+    pub fn rounds_per_session_transition_period(
+        &self,
+        slots_per_block: u64,
+        slot_duration: &Duration,
+    ) -> NonZeroU64 {
+        self.rounds_per_interval(slots_per_block, slot_duration)
+    }
+
+    /// Number of rounds per session transition period.
```
youngjoon-lee:
```suggestion
    /// Number of rounds per epoch transition period.
```

nodes/node/binary/src/config/blend/deployment.rs
```diff
@@ -1,48 +1,134 @@
...
+                .maximum_release_delay_in_rounds
+                .get(),
+        )
+        .unwrap()
+    }
+
+    /// Number of rounds per session transition period.
+    ///
+    /// The Blend spec defines this as roughly the same time it takes to propose
+    /// a new block.
```
youngjoon-lee:
```suggestion
    /// The Blend spec defines this as roughly the same as [`rounds_per_interval`].
```
I think we can clarify this comment like this because the [spec](https://www.notion.so/nomos-tech/Blend-Protocol-215261aa09df81ae8857d71066a80084?source=copy_link#27f261aa09df800689d6d88bc2326613) is saying:
> we round up the transition period to a single interval

nodes/node/binary/src/config/blend/deployment.rs
```diff
@@ -1,48 +1,134 @@
...
+        slots_per_block: u64,
+        slot_duration: &Duration,
+    ) -> NonZeroU64 {
+        let rounds_per_session_transition_period =
+            self.rounds_per_session_transition_period(slots_per_block, slot_duration);
+        ((slot_duration.as_secs() * rounds_per_session_transition_period.get())
+            / self.round_duration(slot_duration).as_secs())
+        .try_into()
+        .expect("There must be at least one slot per epoch transition period.")
+    }
```
youngjoon-lee, line 97: I think this algorithm should be: `(round_duration * rounds_per_stp) / slot_duration`. The test passes now because the round duration is the same as the slot duration for now.

nodes/node/binary/src/config/cryptarchia/deployment.rs
```diff
@@ -24,6 +24,42 @@ pub struct Settings {
...
+    }
+
+    const fn epoch_schedule(&self) -> u64 {
+        (self.epoch_config.epoch_period_nonce_buffer.get()
+            + self.epoch_config.epoch_period_nonce_stabilization.get()
+            + self
+                .epoch_config
+                .epoch_stake_distribution_stabilization
+                .get()) as u64
+    }
```
youngjoon-lee, line 52: I think these functions should be modified. Epoch length is defined as $10 \lfloor \frac{k}{f} \rfloor$ (floor). 
We can actually reuse the functions in the cryptarchia-engine. Let me prepare a branch right now.

youngjoon-lee: Wdyt about c9c5bba0c7934bed77752a55aa41c911d83d7db7? which reuses the cryptarchia-engine algorithms that do `floor` properly.

youngjoon-lee (approved):
thank you!

PR #2287, opened 2026-02-25
feat(sdp): recover declaration info on restart

pradovic:

This is a fix that should enable the sdp to find a correct nonce after restart. 

Yes

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/tests/src/topology/mod.rs
+++ b/tests/src/topology/mod.rs
@@ -35,7 +35,7 @@ pub struct TopologyConfig {
     pub n_validators: usize,
     pub network_params: NetworkParams,
     pub extra_genesis_notes: Vec<GenesisNoteSpec>,
-    /// Override the SDP lock_period for this test topology.
+    /// Override the SDP `lock_period` for this test topology.
     /// If None, uses the default from deployment settings (10).
     pub lock_period_override: Option<u64>,
 }
@@ -68,7 +68,7 @@ impl TopologyConfig {
     }
 
     #[must_use]
-    pub fn with_lock_period(mut self, lock_period: u64) -> Self {
+    pub const fn with_lock_period(mut self, lock_period: u64) -> Self {
         self.lock_period_override = Some(lock_period);
         self
     }


tests/src/tests/mantle/sdp_ops.rs
```diff
@@ -25,13 +23,16 @@ use tokio::time::{sleep, timeout};
...
 ///   disappears.
+///
+/// Note: Activity testing requires the blend service to generate real proofs,
+/// which happens automatically for nodes that are declared as blend providers.
+/// This test focuses on declare/withdraw flow which doesn't require blend
+/// proofs.
 #[tokio::test]
 #[serial]
-#[ignore = "TODO: DA is not supported anymore, rewrite with Blend transaction"]
+#[ignore = "Slow due to lock_period wait - needs faster block production config"]
```
pradovic: @ntn-x2 , @youngjoon-lee  I have tried to rewrite this test using blend instead of DA, but I cannot make it fast enough. It needs 10-15 minutes on my machine to produce enough blocks.

The test waits for lock_period (10 blocks) before withdraw. Even with max_block_propagation_time for proper timeout calculation, it's inherently slow. Could potentially be fixed by reducing lock_period to 3 blocks in a test-specific config.

This test might be useful

ntn-x2: Can the lock period be reduced in this test? It's a deployment config so it's possible.

pradovic: i will try this, thanks!

nodes/node/binary/src/config/sdp/serde.rs
```diff
@@ -1,24 +1,16 @@
...
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct Config {
+    /// Declaration ID (if set, full declaration info will be fetched from
+    /// ledger on startup).
     #[serde(default)]
-    pub declaration: Option<Declaration>,
+    pub declaration_id: Option<DeclarationId>,
     pub wallet: WalletConfig,
 }
```
danielSanchezQ, line 12: Uhm, shouldn't this be part of state instead? 🤔 Or is it because is expected to be manually restarted once posted the declaration?

services/sdp/src/lib.rs
```diff
@@ -58,14 +59,16 @@ pub type BlockUpdateStream = Pin<Box<dyn Stream<Item = BlockEvent> + Send + Sync
 
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct SdpSettings {
-    /// Declaration info for this node (set after posting declaration and
-    /// restarting).
-    pub declaration: Option<Declaration>,
+    /// Declaration ID for this node (set after posting declaration).
+    /// On startup, the full declaration info (`zk_id`, `locked_note_id`, nonce)
+    /// will be fetched from the ledger.
+    pub declaration_id: Option<DeclarationId>,
```
danielSanchezQ, line 65: I see...can't it be done automatically as per latest declaration related to this node?

services/sdp/src/lib.rs
```diff
@@ -154,6 +160,51 @@ where
...
+                Err(e) => {
+                    tracing::error!(
+                        declaration_id = ?declaration_id,
+                        error = ?e,
+                        "Failed to fetch declaration from ledger"
+                    );
+                }
+            }
+        }
+
```
danielSanchezQ, line 172: This is enough to be a method itself.

nodes/node/binary/src/config/sdp/serde.rs
pradovic: Yes, I think we have a flow where we need to set the declaration id, and then restart the service with it in the config. It was chosen for simplicity I think.  Can we tackle this problem in a separate PR?

services/sdp/src/lib.rs
pradovic: Same as the above?

tests/src/tests/mantle/sdp_ops.rs
pradovic: Tried, but the test failed now for other reasons. As it is not the main point of this PR I will disable it again and open a special PR only for this

services/sdp/src/lib.rs
pradovic: Fixed.

tests/src/tests/mantle/sdp_ops.rs
```diff
@@ -25,28 +23,40 @@ use tokio::time::{sleep, timeout};
...
-/// - activate the declaration and poll the REST test endpoint until the
-///   `Active` height reflects the update;
-/// - advance past the lock period,`Withdraw`, and verify the declaration
+/// - advance past the lock period, `Withdraw`, and verify the declaration
 ///   disappears.
+///
+/// Note: Activity testing requires the blend service to generate real proofs,
+/// which happens automatically for nodes that are declared as blend providers.
+/// This test focuses on declare/withdraw flow which doesn't require blend
+/// proofs.
```
youngjoon-lee, line 32: Yes. I was also trying to write tests for activity messages, but it was not easy because the SDP ledger uses the real blend proof verifier. I also thought about making it generic but it was not smooth because the `ServiceState` is defined as enum.
Anyway, I think it's good to test only withdraw messages.

services/sdp/src/lib.rs
```diff
@@ -126,20 +136,16 @@ where
             .get_updated_settings();
 
         Ok(Self {
-            current_declaration: settings.declaration,
+            declaration_id: settings.declaration_id,
+            current_declaration: None, // Will be fetched from ledger in run()
             service_resources_handle,
-            nonce: 0, // TODO: can we get this from declaration?
+            nonce: 0, // Will be fetched from ledger in run()
```
youngjoon-lee, line 142: Later, I think we can even eliminate this field. Whenever we need the latest nonce, we can fetch it from the latest ledger. But anyway, the current implementation is good enough, I think, because we are not going to run multiple nodes that deal with the same declaration.

youngjoon-lee (commented):
Thank you so much. Sorry for late review.

services/sdp/src/lib.rs
pradovic: I agree, I tried to scope the change to a little as possible, because of the devnet. Let's reiterate this after the devnet. Thank you!

PR #2289, opened 2026-02-25
chore: update devnet spec before 0.1.8 release

ntn-x2:


nodes/node/binary/src/config/deployment/devnet.rs
```diff
@@ -5,10 +5,10 @@ blend:
   common:
     num_blend_layers: 1
     timing:
-      rounds_per_session: 2000
-      rounds_per_interval: 10
+      rounds_per_session: 120000
```
danielSanchezQ: Is this value fine? I may have lose the convo on this, but seems like a much larger increase.

ntn-x2: Yes you are right, the calculations were wrong.

Since slots and rounds are of the same length, we have sessions that are long `10 * k / f` slots (or rounds), so `10 * 30 * 20` in our case. I added one more `/ f` but that is actually not correct. I was using @davidrusu's calculating that `rounds_per_session = session_length / f`, but in reality `rounds_per_session = session_length`, since we already compute `session_length` to be in slots, which are equivalent to rounds.

ntn-x2: So this should be `6000` only. The same value would be `648_000` for mainnet, if we use `k=2_160` and `f=1/30`.

PR #2294, opened 2026-02-25
fix: filter private ip + remove unreachable addr from kad

andrussal:

Ideally we never add unreachable addresses to Kademlia. First quick step we can do is to filter out clearly unreachable local addresses.  Next best thing is to remove address from Kademlia after we fail to dial it. This PR addresses both of these.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/services/network/src/backends/libp2p/swarm/identify.rs
+++ b/services/network/src/backends/libp2p/swarm/identify.rs
@@ -49,6 +49,13 @@ impl<R: Clone + Send + RngCore + 'static> SwarmHandler<R> {
 }
 
 fn is_kademlia_candidate_address(addr: &Multiaddr) -> bool {
+    // Tests run entirely on local/private interfaces; keep production
+    // filtering enabled while allowing all identify addresses in test builds.
+    let filter_identify_addrs = !cfg!(test);
+    if !filter_identify_addrs {
+        return true;
+    }
+
     for protocol in addr {
         match protocol {
             Protocol::Ip4(ip) => {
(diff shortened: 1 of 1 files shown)


pradovic (approved):

nodes/node/standalone-node-config.yaml
```diff
@@ -26,6 +26,12 @@ wallet:
     ed266e6e887b9b97059dc1aa1b7b2e19b934291753c6336a163fe4ebaa28e717: "ed266e6e887b9b97059dc1aa1b7b2e19b934291753c6336a163fe4ebaa28e717"
   voucher_master_key_id: 2e03b2eff5a45478e7e79668d2a146cf2c5c7925bce927f2b1c67f2ab4fc0d26
 
+network:
+  backend:
+    swarm:
+      identify:
+        hide_listen_addrs: true
+
```
danielSanchezQ: This doesn't really affect anything as it is just for standalone deployments.

tests/src/topology/configs/network.rs
```diff
@@ -40,6 +40,10 @@ pub fn create_network_configs(
             network::SwarmConfig {
                 node_key,
                 port: get_available_udp_port().unwrap(),
+                identify: network::identify::Config {
+                    hide_listen_addrs: Some(true),
+                    ..Default::default()
+                },
```
danielSanchezQ: Same here, good to have for tests, but not really the thing.

bacv: This is used by cfgsync, thats why I wanted to move topology and refactor testnet dir to tools.

danielSanchezQ: Ah! ok then. But I feel we are missing it somewhere else. cc @ntn-x2

bacv: Another option would be to make it default in node/serde/configs.

libp2p/src/behaviour/kademlia/behaviour_ext.rs
```diff
@@ -23,6 +23,18 @@ impl<R: Clone + Send + RngCore + 'static> Behaviour<R> {
         }
     }
 
+    pub(crate) fn kademlia_remove_address(&mut self, peer_id: PeerId, addr: &Multiaddr) {
+        if self.kademlia.remove_address(&peer_id, addr).is_some() {
+            tracing::debug!("Removed address {:?} from peer {:?}", addr, peer_id);
+        } else {
+            tracing::debug!(
```
ntn-x2: Maybe a warning here?

nodes/node/standalone-node-config.yaml
ntn-x2: If it's required, then it must be added. But I would say all of these entries can have default value so there's no need to explicitly define them.

tests/src/topology/configs/network.rs
ntn-x2: Does this even have to be a config? Can't we just hide them regardless? If not, then I agree this should be part of the default impl and default to true.

danielSanchezQ: > Another option would be to make it default in node/serde/configs.

Yes, it should be a default there!

andrussal: This makes yes more sense, I am not sure why I hesitated, we have some other "defaults" in places where we build config. Should review these at some point and move up as well.

PR #2300, opened 2026-02-26
feat(chain-network): sample both discovered and connected peers explicitly to attempt block downloads

youngjoon-lee:
### Backgrounds

Previously, we had a constant `MAX_PEERS_TO_TRY_FOR_ORPHAN_DOWNLOAD = 3` which is too low. If the sampled 3 peers don't have blocks that we want to download, we cannot catch up the orphan blocks in the end.

Also, we were prioritizing sampling discovered peers over connected peers. But, this strategy must be analyzed in detail later.

### Solutions
This PR defines two parameters: max sampling limit for each set (connected and discovered).

That is, we don't have prioritization. We explicitly pick `N` peers from the connected set and `M` peers from the discovered-but-not-connected set.

Also, I increased the default values for those parameters to 16, each.

I believe so

@youngjoon-lee 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/tests/testing_framework/src/framework/local/provisioning.rs
+++ b/tests/testing_framework/src/framework/local/provisioning.rs
@@ -498,6 +498,10 @@ fn build_cryptarchia_user_config(
                     peers: HashSet::new(),
                 },
             },
+            network: network::NetworkConfig {
+                max_connected_peers_to_try_download: 16,
+                max_discovered_peers_to_try_download: 16,
+            },
             sync: network::SyncConfig {
                 orphan: network::OrphanConfig {
                     max_orphan_cache_size: NonZeroUsize::new(5)


services/chain/chain-network/src/network/adapters/libp2p.rs
```diff
@@ -275,99 +282,119 @@ where
...
+        .iter()
+        .copied()
+        .choose_multiple(&mut rng, max_connected_peers);
 
-    selected
+    discovered_selected.into_iter().chain(connected_selected)
 }
 
 #[cfg(test)]
 mod tests {
```
youngjoon-lee, line 318: I completely rewrote tests. It might be better to not review diffs. Just review the new tests.

services/chain/chain-network/src/network/adapters/libp2p.rs
```diff
@@ -275,99 +282,119 @@ where
...
-        selected.extend(indexes.into_iter().map(|i| connected_vec[i]));
-    }
+    // select from connected peers
+    let connected_selected = connected_peers
+        .iter()
+        .copied()
+        .choose_multiple(&mut rng, max_connected_peers);
 
-    selected
+    discovered_selected.into_iter().chain(connected_selected)
```
danielSanchezQ, line 314: eleganto!!!

PR #2312, opened 2026-02-27
chore: reduce log noise

ntn-x2:
It also pulls the latest overwatch version which itself had some log polishing done this morning.


services/chain/chain-leader/src/lib.rs
```diff
@@ -402,7 +402,7 @@ where
             loop {
                 tokio::select! {
                     Some(SlotTick { slot, epoch }) = slot_timer.next() => {
-                        info!("Received SlotTick for slot {}, ep {}", u64::from(slot), u32::from(epoch));
+                        trace!("Received SlotTick for slot {}, ep {}", u64::from(slot), u32::from(epoch));
```
ntn-x2, line 405: I am uncertain whether this is `trace` or `debug`. It's for sure not `info` tho. Making this `trace` would reduce log size by quite a bit`.

PR #2325, opened 2026-02-27
chore: node README 🧹🧹🧹 

ntn-x2:

Spruce up the README for our repo.

Table of content is already provided by GitHub so I removed it.

Project structure is risky to have because it's prone to change, but for now I left it in there.

I added links to the releases page to join devnet, and removed all mentions of Windows until it becomes an officially supported target (soon).

To see how it looks like, visit [this PR branch README](https://github.com/logos-blockchain/logos-blockchain/blob/aa/node-readme/README.md).

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,20 +76,13 @@ cargo build -p logos-blockchain-node --release
 
 ### 3. Run a standalone node
 
-Before launching, update `chain_start_time` to a recent timestamp:
-
-```bash
-# Generate a timestamp
-date -u +"%Y-%m-%d %H:%M:%S.000000 +00:00:00"
-```
-
-Edit [`nodes/node/standalone-deployment-config.yaml`](nodes/node/standalone-deployment-config.yaml) and paste the timestamp into the `time.chain_start_time` field, then run:
+To start a local standalone instance of a Logos Blockchain network, run:
 
 ```bash
 target/release/logos-blockchain-node --deployment standalone-deployment-config.yaml nodes/node/standalone-node-config.yaml
 ```
 
-The node stores state in the `state` directory. If you encounter issues on restart, try removing it.
+The node stores state in the `state` directory. If you encounter issues on restart, try removing it before starting the node again.
 
 ### Docker
 


README.md
```diff
@@ -1,341 +1,262 @@
...
 
-### Custom Installation
+## What is Logos Blockchain?
 
-#### Linux
+Logos Blockchain is a core component of the [Logos][logos-website] technology stack.
+It combines zero-knowledge proofs, a mix network for anonymity, and a modular service architecture to provide a foundation for sovereign digital communities.
 
-You can specify a custom version or installation directory:
+### Why Logos Blockchain?
```
hansieodendaal: Definitely worth mentioning that one can run several nodes on a Pi

README.md
```diff
@@ -1,341 +1,262 @@
...
 ```bash
-# For macOS/Linux
+# Generate a timestamp
 date -u +"%Y-%m-%d %H:%M:%S.000000 +00:00:00"
 ```
 
-Open `nodes/node/standalone-deployment-config.yaml` and locate the `time` section. Replace the `chain_start_time` value with the 
-output from the command above:
-YAML
+Edit [`nodes/node/standalone-deployment-config.yaml`](nodes/node/standalone-deployment-config.yaml) and paste the timestamp into the `time.chain_start_time` field, then run:
```
hansieodendaal: I do not think this should be a thing at all. What is described here is for everyone to create their own hard fork.

The emphasis should rather be on running one of the blockchains in the code. Now we have a testnet called devnet, but soon we may more testnets, and then one day, mainnet.

ntn-x2: The standalone deployment was added by @andrussal for local testing.

These are not hard forks, these are entirely isolated networks running locally on a single block proposer. Also, we do have a link to the README where we will include all the deployments currently live for people to join.

What we could do tho, is to remove the `chain_start_time` update step, since running locally on a single node won't every bring issues if the chain start time is too far in the past. What do you think?

ntn-x2:
> A bit more explanatory text (not marketing fluff) throughout would help fill in the story spaces and connect the components into an architectural narrative.

I based this on the Overwatch README, but I am thinking to remove at least he `Why Logos Blockchain?` section. We can iterate on the README later on. We also have blog posts which we could link to from the README. I don't think the README should contain everything.

README.md
ntn-x2: I removed this section as it sounded too much marketing lingo. There will be other places in our website or blog posts that are a better fit for that. I tried to keep the README closer to the repo content itself. Pushed a change in https://github.com/logos-blockchain/logos-blockchain/pull/2325/commits/0eb4097aeafd9b2321a5656de3f6b00a9deddb5a.

README.md
ntn-x2: I removed the section about `chain_start_time` since it's not really required when running a node locally: https://github.com/logos-blockchain/logos-blockchain/pull/2325/commits/8d5cc8af8913fd6fd0b265477eed7595a9f219c4.

PR #2328, opened 2026-03-02
fix(network): gossipsub max transmit size

andrussal:

- Increased libp2p gossipsub transmit size limit to unlimited
- Made inscription workload payload size configurable via LOGOS_INSCRIPTION_PAYLOAD_BYTES

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/tests/testing_framework/src/framework/mod.rs
+++ b/tests/testing_framework/src/framework/mod.rs
@@ -33,6 +33,8 @@ use crate::{
     workloads::{ConsensusLiveness, inscription, transaction},
 };
 
+const DEFAULT_PAYLOAD_BYTES: usize = 128;
+
 pub type ScenarioBuilder = CoreScenarioBuilder<LbcEnv>;
 pub type ScenarioBuilderWith = ScenarioBuilder;
 
@@ -203,7 +205,8 @@ impl ScenarioBuilderExt for ScenarioBuilderWith {
         InscriptionFlowBuilder {
             builder: self,
             channels: NonZeroUsize::MIN,
-            inscription_payload_bytes: NonZeroUsize::new(128).expect("constant is non-zero"),
+            inscription_payload_bytes: NonZeroUsize::new(DEFAULT_PAYLOAD_BYTES)
+                .expect("constant is non-zero"),
         }
     }
 


libp2p/src/behaviour/mod.rs
```diff
@@ -21,7 +21,7 @@ pub mod gossipsub;
 pub mod kademlia;
 pub mod nat;
 
-const DATA_LIMIT: usize = 1 << 16; // Do not serialize/deserialize more than 256 KiB
+const DATA_LIMIT: usize = usize::MAX; // Effectively unlimited at this layer (gossipsub default is 64 KiB)
```
danielSanchezQ: Shouldnt be better to remove this and the config set in the behaviour?

danielSanchezQ (changes requested):

libp2p/src/behaviour/mod.rs
andrussal: We probably don't want to allow anyone overwriting it in yaml(`gossipsub: libp2p::gossipsub::Behaviour)? I am not sure what is good location here for config default as our wrapper BehaviourConfig doesn't have default values.

danielSanchezQ: True, also the default is really small `2048` bytes. But let's use a reasonable size, something like 16MB ?

andrussal: > True, also the default is really small `2048` bytes. But let's use a reasonable size, something like 16MB ?

That's better yes, otherwise too easy to DoS.

tests/testing_framework/src/framework/mod.rs
```diff
@@ -203,7 +203,7 @@ impl ScenarioBuilderExt for ScenarioBuilderWith {
         InscriptionFlowBuilder {
             builder: self,
             channels: NonZeroUsize::MIN,
-            payload_bytes: NonZeroUsize::new(128).expect("constant is non-zero"),
+            inscription_payload_bytes: NonZeroUsize::new(128).expect("constant is non-zero"),
```
hansieodendaal: **Nit:** We could probably use the constant here, but not too important.

```suggestion
            inscription_payload_bytes: NonZeroUsize::new(DEFAULT_PAYLOAD_BYTES).expect("constant is non-zero"),
```

hansieodendaal (commented):
Looking good, 64 kB for gossip messages are way too low.

PR #2331, opened 2026-03-02
chore(block): Check block content max size

danielSanchezQ:

Check block content size on reconstruction. This [feature](https://www.notion.so/nomos-tech/Cryptarchia-v1-Protocol-Specification-21c261aa09df810cb85eff1c76e5798c?source=copy_link#21c261aa09df8141993acc87ccb1a6fe)  was missing.

Yes, is not that of a big change.

@thomaslavaur detected the missing feature. @youngjoon-lee leads and @davidrusu. @danielSanchezQ implemented this.

Yes: https://www.notion.so/nomos-tech/Cryptarchia-v1-Protocol-Specification-21c261aa09df810cb85eff1c76e5798c?source=copy_link#21c261aa09df8141993acc87ccb1a6fe

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/core/src/block/mod.rs
+++ b/core/src/block/mod.rs
@@ -86,7 +86,7 @@ impl<Tx> Block<Tx> {
         signing_key: &Ed25519Key,
     ) -> Result<Self, Error>
     where
-        Tx: Transaction<Hash = TxHash>,
+        Tx: Transaction<Hash = TxHash> + StorageSize,
     {
         let expected_public_key = proof_of_leadership.leader_key();
         let actual_public_key = signing_key.public_key();
@@ -98,6 +98,14 @@ impl<Tx> Block<Tx> {
             return Err(Error::TooManyTxs {
                 count: transactions.len(),
                 max: MAX_BLOCK_TRANSACTIONS,
+            });
+        }
+
+        let tx_size: usize = transactions.iter().map(StorageSize::storage_size).sum();
+        if tx_size > MAX_BLOCK_SIZE {
+            return Err(Error::ContentTooBig {
+                count: tx_size,
+                max: MAX_BLOCK_SIZE,
             });
         }
 
@@ -229,6 +237,7 @@ impl<Tx: Clone + Eq + Serialize + DeserializeOwned> TryFrom<Block<Tx>> for Bytes
 mod tests {
     use std::iter;
 
+    use ark_ff::Field as _;
     use lb_groth16::Fr;
     use lb_key_management_system_keys::keys::UnsecuredZkKey;
     use lb_pol::LotteryConstants;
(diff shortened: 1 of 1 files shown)


core/src/block/mod.rs
```diff
@@ -91,10 +94,10 @@ impl<Tx> Block<Tx> {
             return Err(Error::KeyMismatch);
         }
 
-        if transactions.len() > MAX_TRANSACTIONS {
+        if transactions.len() > MAX_BLOCK_TRANSACTIONS {
```
hansieodendaal, line 97: Why do we not check the cumulative size of the transactions here as well? For example, 10 transactions can bust the block size limit.

services/chain/chain-leader/src/lib.rs
```diff
@@ -623,7 +623,10 @@ where
 
         let valid_tx_stream = stream::iter(valid_txs);
         let selected_txs_stream = tx_selector.select_tx_from(valid_tx_stream);
-        let txs: Vec<_> = selected_txs_stream.take(MAX_TRANSACTIONS).collect().await;
+        let txs: Vec<_> = selected_txs_stream
+            .take(MAX_BLOCK_TRANSACTIONS)
+            .collect()
+            .await;
```
hansieodendaal, line 687: `txs: Vec<_>` should be bounded by block size limit as well - choose and calculate or re-choose if the block size limit go bust, but the selected transactions should fir into the size limit.

It could be:
- Select up to MAX_BLOCK_TRANSACTIONS transactions
- Create the block, if no error good, otherwise if block size out of bounds re-choose

hansieodendaal (commented):
If transaction storage size (as per `StorageSize::storage_size`) can vary, we may need to rethink this scheme. I am working on stress tests that will create many big (~250 outputs) and small transactions to fill up the mempool.

core/src/block/mod.rs
danielSanchezQ: We check both. Could we do it in a single run? Yes, but its easier to see the casuistic this way.

services/chain/chain-leader/src/lib.rs
danielSanchezQ: There is some info I think you are missing. The size of transactions from the stream will not surpass the 1MB limit already. But its true we could re-enforce it here.

hansieodendaal: Yes, I could not find the 1MB bounded tx stream in the code, so I am probably missing it, however, it would be nice to re-enforce the size limit here

danielSanchezQ: > Yes, I could not find the 1MB bounded tx stream in the code, so I am probably missing it, however, it would be nice to re-enforce the size limit here

Anyway, we will be refactoring this later as there are discussions on how to make everything a bit smarter. So not doing it for now.

core/src/block/mod.rs
```diff
@@ -25,6 +26,8 @@ pub enum Error {
     Signature,
     #[error("Too many transactions: {count} exceeds maximum of {max}")]
     TooManyTxs { count: usize, max: usize },
+    #[error("Block content too big: {count} exceeds maximum of {max}")]
+    ContentTooBig { count: usize, max: usize },
```
youngjoon-lee, line 30:
```suggestion
    ContentTooBig { size: usize, max: usize },
```
I think `size` or `bytes` is clearer, unlike `TooManyTxs`

PR #2333, opened 2026-03-02
chore: use fixed-size inputs for pol and poc proofs

ntn-x2:

Use fixed-size arrays for PoL and PoC proofs. Did not touch ZK signatures yet. Fixes https://github.com/logos-blockchain/logos-blockchain/issues/2315.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4562,7 +4562,6 @@ dependencies = [
  "logos-blockchain-libp2p",
  "logos-blockchain-network-service",
  "logos-blockchain-pol",
- "logos-blockchain-poq",
  "logos-blockchain-sdp-service",
  "logos-blockchain-services-utils",
  "logos-blockchain-storage-service",
--- a/nodes/node/binary/Cargo.toml
+++ b/nodes/node/binary/Cargo.toml
@@ -43,7 +43,6 @@ lb-ledger                        = { workspace = true }
 lb-libp2p                        = { workspace = true }
 lb-network-service               = { workspace = true }
 lb-pol                           = { workspace = true }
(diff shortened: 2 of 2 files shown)


core/src/proofs/leader_claim_proof.rs
```diff
@@ -108,8 +108,12 @@ impl LeaderClaimPrivate {
             merkle_path_to_witness(voucher_path);
         let wallet = lb_poc::PoCWalletInputsData {
             secret_voucher: secret_voucher.into(),
-            voucher_merkle_path,
-            voucher_merkle_path_selectors,
+            voucher_merkle_path_and_selectors: voucher_merkle_path
+                .into_iter()
+                .zip(voucher_merkle_path_selectors)
+                .collect::<Vec<_>>()
+                .try_into()
```
danielSanchezQ: Not sure atm, but I remember something like the stabilized collecting into a `Result<[_; SIZE], _>` or something like that.

ntn-x2: The problem I found is that we cannot `zip` two arrays directly. So we need to get their iterators. Once in iterator land, if we want to go back to an array, we need to `try_into` because we don't know how many elements the iterator has. There's nothing else that can fail.

ntn-x2: Found this little cutie ❤️ https://github.com/logos-blockchain/logos-blockchain/pull/2333/commits/07f24aa018f72cd426d96138ed4bb13688e89f60

youngjoon-lee (approved):

PR #2339, opened 2026-03-03
fix(tx-service): always gossip user txs, even if it's in mempool

davidrusu:

When a sequencer submits a tx to the local node's HTTP API and the tx is already in the mempool, `pool.add_item()` returns `Err(ExistingItem)`. The current code calls handle_add_error which just returns the error. So the tx is never gossiped.
Meanwhile, leader nodes may have already purged the tx from their mempools. The sequencer keeps resubmitting, but the tx never reaches leaders because the local node short-circuits on ExistingItem.

yes

@danielSanchezQ

NA

NA

## Checklist

* [x] 2. Description added.

### DIFF
--- a/services/tx-service/src/tx/service.rs
+++ b/services/tx-service/src/tx/service.rs
@@ -402,6 +402,17 @@ where
                     tx_broadcast,
                 );
             }
+            Err(MempoolError::ExistingItem) => {
+                // Tx already in pool, but since this came from a local submission
+                // (not gossip), re-gossip it so leader nodes can pick it up.
+                tokio::spawn(async move {
+                    let adapter = NetworkAdapter::new(settings, network_relay).await;
+                    adapter.send(item_for_broadcast).await;
+                });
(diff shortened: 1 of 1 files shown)


services/tx-service/src/tx/service.rs
```diff
@@ -402,6 +402,17 @@ where
...
                 );
             }
+            Err(MempoolError::ExistingItem) => {
+                // Tx already in pool, but since this came from a local submission
+                // (not gossip), re-gossip it so leader nodes can pick it up.
+                tokio::spawn(async move {
+                    let adapter = NetworkAdapter::new(settings, network_relay).await;
+                    adapter.send(item_for_broadcast).await;
+                });
+                if let Err(e) = reply_channel.send(Ok(())) {
```
bacv, line 412: We probably want to return `MempoolError::ExistingItem`, now we return "Success" even when an item already exists in the mempool.

PR #2360, opened 2026-03-05
chore(tests): Test restart with different initial peers

bacv:

Preparation for a "restart" step that will be addition to current "setup" and "run" steps.

In order for the devnet nodes to be upgraded with new release without performing new ceremony or losing state, the upgraded nodes need to be able to start with a initial peer set.

N/A

N/A

N/A

## Checklist

* [x] 2. Description added.


tests/src/nodes/validator.rs
```diff
@@ -207,6 +241,21 @@ impl Validator {
         }
     }
 
+    pub async fn wait_for_height(&self, target_height: u64, duration: Duration) -> Option<()> {
+        tokio::time::timeout(duration, async {
+            loop {
+                let info = self.consensus_info(false).await;
+                println!("{info:?}");
```
ntn-x2, line 254: Print every time?

tests/src/tests/cli/restart.rs
```diff
@@ -0,0 +1,96 @@
...
+async fn node_restart_w_init_peers() {
+    let topology = Topology::spawn(TopologyConfig::two_validators()).await;
+    topology.wait_network_ready().await;
+
+    let validator_1 = &topology.validators()[0];
+    let validator_1_config = validator_1.config();
+
+    let validator_1_multiaddr: Multiaddr = format!(
+        "/ip4/127.0.0.1/udp/{}/quic-v1",
+        validator_1_config.user.network.backend.swarm.port
```
ntn-x2, line 22: Don't we have the `user.network.backend.swarm.host` value to use here?

tests/src/tests/cli/restart.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    validator_1
+        .wait_for_height(1, height_timeout)
+        .await
+        .expect("validator should produce the first block");
+
+    validator_2
+        .wait_for_height(2, height_timeout)
+        .await
+        .expect("validator should produce the first block");
```
ntn-x2:
```suggestion
        .expect("validator should produce the first two blocks");
```

tests/src/nodes/validator.rs
```diff
@@ -133,6 +133,40 @@ impl Validator {
...
+
+        // Re-write config files (they were temporary and may have been cleaned up)
+        let mut user_config_file = NamedTempFile::new().unwrap();
+        let mut deployment_config_file = NamedTempFile::new().unwrap();
+
+        serde_yaml::to_writer(&mut user_config_file, &self.config.user).unwrap();
+        serde_yaml::to_writer(&mut deployment_config_file, &self.config.deployment).unwrap();
+
+        // Spawn new process with same config
+        let exe_path = get_exe_path();
```
ntn-x2, line 155: Can we re-use `self.spawn` after adapting to support optional arguments? Else we don't do everything we do in `spawn`, and I'm not sure that's intended?

tests/src/nodes/validator.rs
```diff
@@ -133,6 +133,40 @@ impl Validator {
         Ok(())
     }
 
+    /// Restarts with the same deployment and user configs, but attaches provided cli arguments.
+    pub async fn restart_with_args(&mut self, args: Vec<String>) -> Result<(), Elapsed> {
```
ntn-x2: Nit, but would it make sense to use the same signature the std library uses? Which is

`pub fn args<I, S>(&mut self, args: I) -> &mut Command
    where
        I: IntoIterator<Item = S>,
        S: AsRef<OsStr>,
`

tests/src/tests/cli/restart.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    println!(
+        "Final Heights -> V1: {}, V2: {}, V3: {}",
+        v1_info.height, v2_info.height, v3_info.height
+    );
+
+    assert!(
+        v3_info.height >= target_height,
+        "Third node failed to progress after restart"
+    );
```
hansieodendaal, line 94: We should also check the header hash, and wait for the nodes to converge on height and hash, not only a target height in the past. All nodes should be on the same chain.

tests/src/nodes/validator.rs
hansieodendaal: Yes maybe print 1 in 1000

bacv: All the prints are visible only when `--nocapture` is used, or the test fails.

tests/src/tests/cli/restart.rs
bacv: The host value sets the interface on which it will listen, in this case it's `0.0.0.0`, unfortunately such addr is too undefined for client to connect.

tests/src/nodes/validator.rs
bacv: In the restart case it's very similar to the spawn method, but we are not creating a new directory and names of the config paths. That's the key difference, I think in this case it's okay to have some code related to running a new process duplicated.

tests/src/tests/cli/restart.rs
bacv: The assumption here is like this, please validate:
- node1 and node2 are in genesis, and are block producers
- node3 connects to node1 as an initial 
- node3 is restarted with initial peer set to node2

If node1 and node2 are not in sync, then node3 won't be able to apply new blocks and increase in height.
We don't check the LIB because it stays the same at such low height, we don't check the tip, because we are not guaranteed they will match, and that's fine, we just need to be sure that node3 can continue building the chain after connecting to the node2.

tests/src/nodes/validator.rs
bacv: fixed in https://github.com/logos-blockchain/logos-blockchain/pull/2360/commits/736ba6a3f96b19d4ced7ad9604b648046969f9bc

tests/src/nodes/validator.rs
ntn-x2: Yes, but this is printing every half a second, I think it might be too much, but as long as this is not a left-over, it's ok. I wanted to make sure it's not one.

tests/src/tests/cli/restart.rs
hansieodendaal: What sometimes happens in the tests is that each node can be on their own chain, but I agree, the purpose of this test is just to verify that node3 can make progress, not if it is synced.

bacv: I think what helps in this case is that the node3 is not a block producer.

PR #2372, opened 2026-03-09
feat(chain): shrink the map of ledger states after bootstrapping

youngjoon-lee:

`HashMap` doesn't release memory eagerly after elements are removed from it.

We are removing a lot of `LedgerState`s from the `HashMap` when switching to Online mode (after bootstrapping is done). This PR adds shrinking the `HashMap` only once at that time to free up memory. 

We don't shrink it in any other place because it's expensive. Later we'll consider introducing the periodical shrinking (which should be designed carefully).

yes

@youngjoon-lee 

n/a

n/a

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/services/chain/chain-service/src/lib.rs
+++ b/services/chain/chain-service/src/lib.rs
@@ -320,7 +320,7 @@ impl Cryptarchia {
         self.consensus = consensus;
 
         // Prune the ledger states of all the pruned blocks.
-        self.prune_ledger_states(pruned_blocks.all(), false);
+        self.prune_ledger_states(pruned_blocks.all());
 
         Ok((pruned_blocks, reorged_blocks))
     }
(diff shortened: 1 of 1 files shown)


bacv (approved):

services/chain/chain-service/src/lib.rs
```diff
@@ -320,7 +320,7 @@ impl Cryptarchia {
         self.consensus = consensus;
 
         // Prune the ledger states of all the pruned blocks.
-        self.prune_ledger_states(pruned_blocks.all());
+        self.prune_ledger_states(pruned_blocks.all(), false);
```
danielSanchezQ: could we make the call to shrink explicit instead of hidding trhough a flag?

services/chain/chain-service/src/lib.rs
```diff
@@ -350,6 +357,10 @@ impl Cryptarchia {
             }
         }
         tracing::debug!(target: LOG_TARGET, "Pruned {pruned_states_count} old forks and their ledger states.");
+
+        if shrink {
+            self.ledger.shrink();
+        }
```
danielSanchezQ: Yeah, I think its better to offer prunning but not doing it here through a flag.

danielSanchezQ (changes requested):

services/chain/chain-service/src/lib.rs
youngjoon-lee: Yeah, I actually thought about the option when writing this. Changed it d016f147

services/chain/chain-service/src/lib.rs
youngjoon-lee: Done https://github.com/logos-blockchain/logos-blockchain/pull/2372#discussion_r2906594459

PR #2374, opened 2026-03-09
chore(tracing): basic metrics

bacv:

This is an update to https://github.com/logos-blockchain/logos-blockchain/pull/2012 where the conflicts with a master was resolved, this was tested with the `compose.tracing.yml` and DST lab.

Discussion https://github.com/logos-blockchain/logos-blockchain/pull/2012, the linked PR will be closed in favor of this one.

N/A

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/tracing/src/metrics/emit.rs
+++ b/tracing/src/metrics/emit.rs
@@ -103,49 +103,20 @@ impl IntoMetricU64 for i32 {
     }
 }
 
-fn u64_counter(name: &'static str) -> Counter<u64> {
-    let mut counters = U64_COUNTERS.lock().expect("u64 counter lock poisoned");
-
-    counters
-        .entry(name)
-        .or_insert_with(|| meter().u64_counter(name).build())
-        .clone()
-}
-
-fn f64_counter(name: &'static str) -> Counter<f64> {
-    let mut counters = F64_COUNTERS.lock().expect("f64 counter lock poisoned");
-
-    counters
-        .entry(name)
-        .or_insert_with(|| meter().f64_counter(name).build())
-        .clone()
-}
-
-fn u64_gauge(name: &'static str) -> Gauge<u64> {
-    let mut gauges = U64_GAUGES.lock().expect("u64 gauge lock poisoned");
-
-    gauges
-        .entry(name)
-        .or_insert_with(|| meter().u64_gauge(name).build())
-        .clone()
-}
-
-fn u64_histogram(name: &'static str) -> Histogram<u64> {
-    let mut histograms = U64_HISTOGRAMS.lock().expect("u64 histogram lock poisoned");
-
-    histograms
-        .entry(name)
-        .or_insert_with(|| meter().u64_histogram(name).build())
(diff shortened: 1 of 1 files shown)


tracing/src/metrics/emit.rs
```diff
@@ -0,0 +1,173 @@
...
+    U64_GAUGES.lock().expect("u64 gauge lock poisoned").clear();
+    U64_HISTOGRAMS
+        .lock()
+        .expect("u64 histogram lock poisoned")
+        .clear();
+    F64_HISTOGRAMS
+        .lock()
+        .expect("f64 histogram lock poisoned")
+        .clear();
+}
```
danielSanchezQ, line 44: If this fails, what would happen? 🤔 should we stop the node because of metrics broke?

tracing/src/metrics/emit.rs
```diff
@@ -0,0 +1,173 @@
...
+    let mut counters = F64_COUNTERS.lock().expect("f64 counter lock poisoned");
+
+    counters
+        .entry(name)
+        .or_insert_with(|| meter().f64_counter(name).build())
+        .clone()
+}
+
+fn u64_gauge(name: &'static str) -> Gauge<u64> {
+    let mut gauges = U64_GAUGES.lock().expect("u64 gauge lock poisoned");
```
danielSanchezQ: Same for all this calls. If we expect here an panic, where do we panic? This will end in terminating the node just because metrics do not work. We should not do this.

bacv: Since these might fail at runtime, I've changed them to just log the error if lock is unavailable.

tracing/src/metrics/emit.rs
bacv: Since this one is executed on the service startup (layer creation), I think it's fine to make it panic on error. If user didn't use metrics layer, this won't bring down the node, if it was specified, it will fail to startup the node.

hansieodendaal: We can recover gracefully from poisoned locks without panicking, example
```rust
    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
        self.inner
            .lock()
            .unwrap_or_else(|poisoned| {
                warn!("Tracing writer mutex poisoned on write, recovering");
                poisoned.into_inner()
            })
            .write(buf)
    }
```

bacv: But do we really want this during the node startup? My argument is that if a user chose to collect metrics and it failed on a startup (this step would happen during the service init) then its better stop the node entirely. This step should not fail in a normal environment.

hansieodendaal: It could be philosophical, so let's treat it as a NIT.

I prefer the node to keep running and show all the warnings, as it is possible that a poisoned lock can be recovered and life goes on without further/little incidents. If the error state is permanent, it will be evident in the console/logs. Both approaches have merit.

PR #2379, opened 2026-03-14
feat: add sequencer inscription lineage tracking

hansieodendaal:

Added sequencer inscription lineage tracking and delayed checkpoint persistence.

This is a WIP from the Lisbon All-Hands together with @ntn-x2. This PR manually merges additions from #2375, 
builds on top of that work and supplements it. However, the previously proposed blocking publishing model was 
replaced with asynchronous publishing.

This change upgrades the zone sequencer from tx-only tracking to inscription-aware chain tracking.

On the SDK side, `TxState` now tracks inscription lineage per block using `MsgIdState`. It resolves the latest deterministic inscription tip, detects ambiguous competing branches, and retains sufficient finalized lineage below LIB to continue publishing against the correct parent after pruning, restart, or backfill. Block processing, backfill, and restart bootstrap were updated to rebuild and maintain this lineage information. Publish parent selection now prefers the resolved on-chain inscription tip and rejects ambiguous forks.

On the TUI side, checkpoint handling was changed from immediate persistence on publish to persistence after the inscription is observed on-chain. The sequencer now notifies the application of inscription status in the background and saves checkpoints only once the chain state reflects the persisted inscription. Logging and terminal output were also improved for published, mined, delayed, and failed inscriptions.

When inscription collisions are detected with a peer sequencer, the affected inscriptions are recreated and resubmitted until they are successfully mined. This design depends on refreshing `pending_publications` promptly when new on-chain `MsgIdState` is observed. That refresh timing is critical to avoiding stale parent selection.

Overall, this makes the sequencer more robust across restarts, reorg windows, delayed inclusion, and concurrent local publishing while preserving correct inscription parent chaining.

System-level tests were performed where three sequencers were started concurrently and used to post inscriptions in random order. All inscriptions eventually resolved automatically. This also included clean starts without any saved state information.

Yes

@ntn-x2 and @hansieodendaal 

Yes

n/a

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/zone-sdk/src/sequencer.rs
+++ b/zone-sdk/src/sequencer.rs
@@ -6,7 +6,6 @@ use std::{
 use futures::{StreamExt as _, future::BoxFuture, stream::FuturesUnordered};
 use lb_common_http_client::{BasicAuthCredentials, CommonHttpClient, ProcessedBlockEvent, Slot};
 use lb_core::{
-    codec::SerializeOp as _,
     header::HeaderId,
     mantle::{
         MantleTx, SignedMantleTx, Transaction as _,
@@ -176,6 +175,7 @@ impl ZoneSequencer {
     /// Returns the inscription ID and a checkpoint for persistence.
     pub async fn publish(&self, data: Vec<u8>) -> Result<PublishResult, Error> {
         let (reply_tx, reply_rx) = oneshot::channel();
+        let msg = String::from_utf8_lossy(&data).into_owned();
         let request = ActorRequest::Publish {
             data,
             reply: reply_tx,
@@ -193,10 +193,10 @@ impl ZoneSequencer {
         })??;
 
         info!(
-            "Created inscription - hash: {}, this msg_id: {}, parent msg_id: {}",
-            hex::encode(result.inscription_id.to_bytes().unwrap_or_default()),
-            hex::encode(result.checkpoint.last_msg_id.as_ref()),
-            hex::encode(result.parent_msg_id.as_ref()),
+            "Created inscription - msg: {msg}, hash: '{}', this msg_id: '{}', parent msg_id: '{}'",
+            result.inscription_id.as_hex(),
+            result.checkpoint.last_msg_id.as_hex(),
+            result.parent_msg_id.as_hex(),
         );
 
         // Post to network (best effort, will be resubmitted/recreated if needed)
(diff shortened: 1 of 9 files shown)


zone-sdk/src/sequencer.rs
```diff
@@ -317,17 +412,46 @@ async fn run_loop(
...
+    let (state, current_tip, lib_slot, last_msg_id, restored_from_checkpoint) =
         initialize_from_checkpoint(&http_client, &node_url, config.reconnect_delay, checkpoint)
             .await;
     let mut state = Some(state);
     let mut current_tip = Some(current_tip);
     let mut lib_slot = lib_slot;
     let mut last_msg_id = last_msg_id;
 
+    if let (Some(s), Some(tip)) = (state.as_mut(), current_tip) {
+        bootstrap_history_on_clean_start(s, tip, channel_id, &http_client, &node_url).await;
```
pradovic, line 424: Is `bootstrap_history_on_clean_start` called every time then? Does it mean we would scan the whole chain even if we have a checkpoint?

hansieodendaal: A local checkpoint is good only if the local sequencer made the last inscription for the channel, so we have to scan the chain, but we do not scan the whole chain, only from the tip to the last inscription.

zone-sdk/src/sequencer.rs
```diff
@@ -243,34 +290,81 @@ async fn initialize_from_checkpoint(
...
+    };
+
+    let to: u64 = tip_slot.into();
+    if to == 0 {
+        return;
+    }
+
+    debug!("Bootstrapping inscription lineage from genesis to tip slot {to}",);
+
+    match http_client.get_blocks(node_url.clone(), 1, to).await {
```
pradovic, line 350: Related to: https://github.com/logos-blockchain/logos-blockchain/pull/2379/changes#r2974224325. This seems to scan from genesis up to tip, which can be very large?

I think we need to scan from checkpoint up to a new tip to reduce this (and also batch). If we have a checkpoint this means we already scanned this part of the chain, and no need to scan it again, right?
We can adapt checkpoint structure if needed.

zone-sdk/src/sequencer.rs
pradovic: I have added another comment to the exact line: https://github.com/logos-blockchain/logos-blockchain/pull/2379/changes#r2974277008.

I think the checkpoint should always be valid, as it should show the last scanned immutable inscription seen on the chain. Then we need to scan from checkpoint up to lib for finalized state. And then from tip back to lib for uninitialized to update the state. I think this logic was already there we just need to modify it to take into account foreign inscriptions . We can also change the structure of checkpoint if needed.

PR #2380, opened 2026-03-15
feat: add efficient inscription channel event/item streaming

hansieodendaal:

This PR adds **channel inscription item streaming and querying** end-to-end. The next step after this is to integrate it with the zone-sdk sequencer code, as provided in #2379 or similar. This PR will greatly enhance the ability of newly formed channel partners to scan historic channel content in "bootsrap" mode, for example there may exist a single inscription transaction near the genesis block.

### Scope
- Introduces typed API bodies for channel inscription responses/events.
- Exposes new channel inscription API paths.
- Adds/updates handlers and API wiring.
- Adds HTTP client support for:
  - channel inscription history retrieval;
  - channel inscription stream subscription.
- Adds integration coverage test.

### Functional behavior
- Supports fetching channel inscription items by channel and range/cursor semantics.
- Supports streaming `ChannelInscriptionsStreamEvent` updates (e.g. inscription observed and chain progress events).
- Includes test coverage for bootstrap/live stream transition and duplicate suppression expectations.

### Context that is now explicit in code paths
- `ChannelInscriptionItem` is the canonical transfer object for channel inscription materialization across API and client.
- Streaming API is intentionally modeled as event envelopes (`ChannelInscriptionsStreamEvent`) rather than raw transactions to support both data and progress signaling.
- The client mirrors server event structure directly for low-friction consumer integration.

### Idiosyncratic / implementation-specific details to call out
- **Bootstrap + live boundary behavior:** stream logic is designed to avoid duplicate emission when transitioning from initial snapshot to live tail.
- **Lineage-sensitive test behavior:** tests rely on valid channel message parent linkage (via `MsgId`) when submitting subsequent inscriptions.
- **Route parameter handling consistency:** API path parameter style and client substitution must stay aligned to avoid route miss/404 behavior.

Yes.

@hansieodendaal and **??**

[**TBD**]

[**TBD**]

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


nodes/node/binary/src/api/handlers.rs
```diff
@@ -495,6 +508,130 @@ where
...
+    channel_id: ChannelId,
+    include_inscription: bool,
+) -> Vec<ChannelInscriptionsStreamEvent> {
+    collect_channel_inscription_items(block, channel_id, include_inscription)
+        .into_iter()
+        .map(|item| ChannelInscriptionsStreamEvent::InscriptionObserved { item })
+        .collect()
+}
+
+async fn fetch_canonical_bootstrap_blocks<StorageBackend, RuntimeServiceId>(
```
pradovic, line 596: Should we make this a stream? This way we would guard against the huge payload case.

Then in `channel_inscriptions_stream` you could  .flat_map it into events and chain with live_events directly — no `bootstrap_events: Vec accumulator` either

nodes/api-common/src/bodies/channel_inscriptions.rs
```diff
@@ -0,0 +1,33 @@
...
+    pub tx_hash: TxHash,
+    pub parent_msg_id: MsgId,
+    pub this_msg_id: MsgId,
+    pub inscription: Option<Vec<u8>>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ChannelInscriptionsResponseBody {
+    pub items: Vec<ChannelInscriptionItem>,
+    pub next_cursor: Option<u64>,
```
pradovic, line 44: Is this currently hardcoded to None or is it used?

nodes/api-common/src/bodies/channel_inscriptions.rs
```diff
@@ -0,0 +1,33 @@
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(tag = "event", rename_all = "snake_case")]
+pub enum ChannelInscriptionsStreamEvent {
```
pradovic, line 54: Maybe the doc-comment on the stream endpoint should spell out something like:
    - Bootstrap items from from_slot..lib are finalized.
    - Bootstrap items from lib..tip and all live items are unfinalized until a subsequent LibAdvanced covers their header_id.
    - Clients are responsible for tracking the block tree to reconcile
  reorgs.

nodes/api-common/src/bodies/channel_inscriptions.rs
```diff
@@ -0,0 +1,33 @@
...
+    InscriptionObserved { item: ChannelInscriptionItem },
+    LibAdvanced { tip: HeaderId, lib: HeaderId },
+}
```
pradovic, line 57: This enum/API looks a bit odd to me.

Could/should we add either:
	•	an explicit reorg event here, or
	•	clearer docs that clients are expected to walk parent_block and reconcile the block tree themselves?

I think this gets to the broader question of what we want to keep in the SDK vs what we want to expose from the node. If we want this API to be fully reorg-aware, it may be a sign that channel handling belongs in a dedicated node-side service rather than growing this handler/API incrementally.

cc. @davidrusu , @hansieodendaal  - would be good to align on how much logic we want in the node vs in the SDK or clients, as it is about node public API as well. 

These handlers are starting to get fairly fat (handling bootstrap, live streaming, and partial reorg semantics). I’m not against that direction, but I think we should be deliberate about where we want this abstraction to live (sdk, clients, node, handlers/service).

nodes/node/binary/src/api/handlers.rs
```diff
@@ -495,6 +508,130 @@ where
...
+    };
+
+    if from_slot as u64 <= lib_slot && lib_slot > 0 {
+        for block in mantle::get_blocks(handle, from_slot, lib_slot as usize).await? {
+            seen_headers.insert(block.header().id());
+            blocks.push(block);
+        }
+    }
+
+    let mut canonical_headers =
```
pradovic, line 643: I believe that this re-queries consensus after the snapshot taken at line 762 (live_stream.next().await in channel_inscriptions_stream). Between those two .await points, consensus may have reorged — tip_id could now be on an orphaned branch or pruned, and the returned headers would be stale.

pradovic (commented):
I think this is useful and would make zone-sdk code more optimal and simplify it considerably, but let's sync how much should we offload to node vs what should be in the zone-sdk.

Left some comments to fix/discuss.

nodes/api-common/src/bodies/channel_inscriptions.rs
hansieodendaal: pagination is now implemented

danielSanchezQ (changes requested):
@hansieodendaal Why are we adding this to the node API? The node api is a generalized utility, not a specialized one. Whoever uses de sdk receives the blocks already and can track block content from there when needed.

hansieodendaal:
> @hansieodendaal Why are we adding this to the node API? The node api is a generalized utility, not a specialized one. Whoever uses de sdk receives the blocks already and can track block content from there when needed.

This is for improved sequencer efficiency. The node can filter inscriptions for a specific channel at virtually no cost, but the downstream effect is profound. When sequencers without state joins a specific channel, they need to be bootstrapped, and this will prevent them from needing to scan the entire blockchain to discover a handful of inscriptions. This will also drastically reduce the amount of data that needs to be sent across the "wire" (local or remote node).

PR #2389, opened 2026-03-16
feat(rewards): Block rewards calculation and distribution

thomaslavaur:

This PR implement the block reward calculation. When the content of a block is executed to the state, the block rewards are calculated. These block rewards are then added to the pending rewards for leader and blend. When a new epoch starts the pending leader rewards are added to the pool and when a new blend session starts the pending blend rewards are fixed and wait to be distributed.

Yes. It follows the block reward specification.

specification: @fredericosteixeira 
implementation: @thomaslavaur 

Yes: [Block rewards](<https://www.notion.so/nomos-tech/Block-Rewards-Specification-1fd261aa09df815b951ff7139cde3fd3>)

Yes: introducing reference implementation for precision management of the calculation of block rewards

## Checklist

* [x] 2. Description added.


ledger/src/cryptarchia/mod.rs
```diff
@@ -126,6 +126,11 @@ pub struct LedgerState {
     // Using an Arc wrapper here as this can be completely shared among instances of LedgerState
     #[derivative(PartialEq = "ignore")]
     stake_inference: Arc<StakeInference>,
+    // rolling fee window of 120 blocks, used to derive block rewards
+    #[cfg_attr(feature = "serde", serde(with = "serde_arrays"))]
+    pub fee_window: [f64; 120],
```
danielSanchezQ: `120` should be `const` probably. Also, are we sure f64 is safe here? Should't we use a fixed precision one?

ledger/src/mantle/leader.rs
```diff
@@ -68,6 +69,7 @@ impl LeaderState {
             claimable_vouchers_root: RewardsRoot::default(),
             n_claimable_vouchers: 0,
             nfs: rpds::HashTrieSetSync::new_sync(),
+            pending_rewards: Value::default(),
```
danielSanchezQ: nitpick: I think it could be nice to have explicit `0`

ledger/src/mantle/sdp/mod.rs
```diff
@@ -570,6 +574,14 @@ impl SdpLedger {
         Ok(self)
     }
 
+    pub fn add_blend_rewards(&mut self, rewards: Value) {
+        if let Some(service) = self.services.get_mut(&ServiceType::BlendNetwork) {
+            match service {
+                Service::BlendNetwork(state) => state.add_rewards(rewards),
+            }
```
danielSanchezQ: A single nested match is enough:
```suggestion
        if let Some(Service::BlendNetwork(state)) = self.services.get_mut(&ServiceType::BlendNetwork) {
                state.add_rewards(rewards);
        }
```

ledger/src/lib.rs
```diff
@@ -22,6 +22,18 @@ use lb_groth16::{Field as _, Fr};
...
+const KPI_FEE_TARGET: f64 = 10_000_000_000f64;
+const TGE_TOKEN: f64 = 10_000_000_000f64;
+const I_MAX: f64 = 0.01;
+const DELTA_T: f64 = 1_051_200f64;
+const F: f64 = 1f64;
+const ALPHA_D: f64 = 0.25;
+const ALPHA_A: f64 = 1f64;
+const I_MIN: f64 = 0f64;
+const T: f64 = 120f64;
+const W_I: f64 = 1f64;
```
danielSanchezQ: We usually tend to use explicit names instead of the math ones, as its designations are usually oblivious if you read them. But its fine if they do not actually have proper naming.

ledger/src/lib.rs
```diff
@@ -188,12 +202,56 @@ impl LedgerState {
         })
     }
 
+    /// For each block received, rewards are calculated based on the actual
+    /// total estimated stake and on the average of fees consumed per block over
+    /// the last T blocks
+    fn compute_block_rewards(mut self) -> Self {
```
danielSanchezQ, line 230: Lets add a link to the spec for this.

ledger/src/lib.rs
```diff
@@ -188,12 +202,56 @@ impl LedgerState {
...
+        emission_rate = emission_rate.clamp(0f64, 1f64);
+        let block_reward = (emission_rate * I_MAX * TGE_TOKEN).mul_add(
+            DELTA_T / F,
+            (1f64 - emission_rate) * self.cryptarchia_ledger.fee_window[window_index],
+        );
+
+        // blend get 60% of rewards while leaders get the 40% remaining.
+        // Casting truncate the floating points
+        let blend_reward = (0.6 * block_reward) as Value;
+        let leader_reward = (0.4 * block_reward) as Value;
```
danielSanchezQ: `0.4` and `0.6` should be named `const` values as well if possible.

ledger/src/cryptarchia/mod.rs
thomaslavaur: This should give the same result across implementations... Rust doesn't have a native fixed precision float what should we use then?

ledger/src/lib.rs
thomaslavaur: Done

ledger/src/lib.rs
thomaslavaur: Done

ledger/src/lib.rs
thomaslavaur: Renamed

ledger/src/mantle/sdp/mod.rs
thomaslavaur: Thanks!

ledger/src/mantle/leader.rs
thomaslavaur: Modified

ledger/src/cryptarchia/mod.rs
danielSanchezQ: There is not a fixed precision native one, but we already used one somewhere iirc. This should be discussed, as it should be part of the specification itself.

ledger/src/mantle/sdp/rewards/blend/mod.rs
```diff
@@ -183,6 +181,31 @@ where
             },
         }
     }
+
+    fn add_rewards(&self, rewards: Value) -> Self {
```
youngjoon-lee: Wdyt about renaming this to `add_income` or `add_block_rewards` (similar as `current_session_tracker.add_block_rewards`) to avoid confusion with service rewards (which are calculated by this module)?

ledger/src/mantle/leader.rs
```diff
@@ -32,6 +32,7 @@ pub struct LeaderState {
     // at the start of each epoch this is increased by the amount of rewards
     // that have been collected in the previous epoch.
     // unclaimed rewards are carried over to the next epoch.
+    pub pending_rewards: Value,
     claimable_rewards: Value,
```
youngjoon-lee, line 35: I think it's better to reorder these two fields because the comment explains about `claimable_rewards`. Also, it would be good to add a comment for `pending_rewards`, such as:
```
Rewards that are being collected during the current epoch.
This will be added to the `claimable_rewards` when a new epoch starts.
```

youngjoon-lee: Also, it would be better to remove `pub` for `pending_rewards` if possible.

ledger/src/lib.rs
```diff
@@ -188,12 +204,62 @@ impl LedgerState {
...
+    }
+
     /// Apply the contents of an update to the ledger state.
     pub fn try_apply_contents<Id, Constants: GasConstants>(
         mut self,
         config: &Config,
         txs: impl Iterator<Item = impl AuthenticatedMantleTx>,
     ) -> Result<Self, LedgerError<Id>> {
+        let fee_window_index = self.block_number as usize % WINDOW_SIZE;
+        self.cryptarchia_ledger.fee_window[fee_window_index] = 0f64;
```
youngjoon-lee: It would be better to make `fee_window` private and update it through a function. If possible, we could design the function to reset+update fees at once:
```rust
let mut total_gas_cost = 0;
for tx in txs {
  ...
  total_gas_cost += tx.gas_cost();
}
self.cryptarchia_ledger.update_fee_window(fee_window_index, total_gas_cost);
```

ledger/src/cryptarchia/mod.rs
youngjoon-lee: It seems that these fields can hold `u64`s instead of `f64`s because gas cost and stakes are `Value` (= `u64`).
Then, we can sum them and convert it into `f64` or a fixed precision value only when we compute rewards for each new block.

This might not be a huge difference in terms of memory size because both u64 and f64 are 8 bytes. But if we keep summing with f64, rounding errors may accumulated.

youngjoon-lee: Regarding f64 arithmetic in `compute_block_rewards`, could we probably revise the algorithm by using `i128` similar as [`StakeInference::total_stake_inference`](https://github.com/logos-blockchain/logos-blockchain/blob/220d49306b0ec2a84074a184376ae530800c5671/ledger/src/cryptarchia/stake.rs#L24)? ([spec](https://www.notion.so/nomos-tech/Total-Stake-Inference-22d261aa09df8051a454caa46ec54b34?source=copy_link#22d261aa09df8015bf22fb28aa0c0ea1))

If it's not possible, we can probably use the same practice as [`LotteryConstants::new(f: NonNegativeRatio)`](https://github.com/logos-blockchain/logos-blockchain/blob/220d49306b0ec2a84074a184376ae530800c5671/zk/proofs/pol/src/lottery.rs#L43). It accepts `NonNegativeRatio` which consists of `numerator` and `denominator`, and converts each of them into `BigFloat` using the `PRECISION` constant.

ledger/src/mantle/sdp/rewards/blend/mod.rs
thomaslavaur: I'll go for add_income as block rewards are dedicated to blend and leaders so we can stay generic for future services

ledger/src/cryptarchia/mod.rs
```diff
@@ -381,6 +386,34 @@ impl LedgerState {
         }
     }
 
+    pub const fn update_stake_window(&mut self, index: usize) {
+        self.stake_window[index] = self.epoch_state.total_stake;
+    }
+
+    pub const fn update_fee_window(&mut self, index: usize, total_fee: u64) {
+        self.stake_window[index] = total_fee;
```
youngjoon-lee:
```suggestion
    pub const fn update_fee_window(&mut self, index: usize, total_fee: u64) {
        self.fee_window[index] = total_fee;
```
I think we should update fee_window.

ledger/src/lib.rs
```diff
@@ -188,6 +193,41 @@ impl LedgerState {
...
+            + (BLOCK_REWARD_SUB_PRECISION - a_t_p)
+                * self.cryptarchia_ledger.get_fee_from_index(window_index)
+                * 657;
+
+        // blend get 60% of rewards while leaders get the 40% remaining.
+        // Casting as Value truncate the floating points
+        let blend_reward =
+            (i128::from(reward_numerator * 6) / i128::from(BLOCK_REWARD_PRECISION * 10)) as Value;
+        let leader_reward =
+            (i128::from(reward_numerator * 4) / i128::from(BLOCK_REWARD_PRECISION * 10)) as Value;
```
youngjoon-lee: I approved but we need to replace these magic numbers with better expressions. From the commit message, it seems that you are still working on it?

I understand that 657 is `BLOCK_REWARD_PRECISION / BLOCK_REWARD_SUB_PRECISION`, but i couldn't infer other magic numbers.

Also, the reason why we divide the final value with "BLOCK_REWARD_PRECISION **x 10**" is that we want to take only 60% of final value for blend and 40% for leader?

thomaslavaur: yes sorry I'm also updating the spec in parrallel I should have tell you.
Give a look to this but it may still be changing sorry https://www.notion.so/nomos-tech/v1-1-Block-Rewards-Specification-326261aa09df80579edddaf092057b3d

ledger/src/lib.rs
```diff
@@ -22,6 +22,30 @@ use lb_groth16::{Field as _, Fr};
...
+
+/// Numerator of 1/(`I_max` * `D1_target` * `Delta_t` * `T`)
+/// That correspond to `BLOCK_PER_YEAR` / (`MAX_INFLATION` * `KPI_FEE_TARGET` *
+/// `WINDOW_SIZE`)
+const FEE_AVG_NUM: u128 = 10_512;
+
+/// Numerator of `I_max` * `S_TGE` * `DELTA_t` / `f`
+/// It corresponds to `MAX_INFLATION` * `TOKEN_GENESIS` * `BLOCK_PER_BLOCK` /
+/// `BLOCK_PER_YEAR`
+const INFLATION_NUM: u128 = 62_500;
```
danielSanchezQ: Nitpick, we should use full naming: `INFLATION_NUMERATOR` its much clearer for everyone.

ledger/src/lib.rs
```diff
@@ -22,6 +22,30 @@ use lb_groth16::{Field as _, Fr};
...
+
+/// Numerator of `I_max` * `S_TGE` * `DELTA_t` / `f`
+/// It corresponds to `MAX_INFLATION` * `TOKEN_GENESIS` * `BLOCK_PER_BLOCK` /
+/// `BLOCK_PER_YEAR`
+const INFLATION_DEN: u128 = 657;
```
danielSanchezQ: Same here, what does `DEN` stands for?

ledger/src/lib.rs
```diff
@@ -188,6 +212,41 @@ impl LedgerState {
...
+        let reward_numerator = INFLATION_NUM * a_numerator
+            + INFLATION_DEN
+                * (A_SCALE - a_numerator)
+                * u128::from(self.cryptarchia_ledger.get_fee_from_index(window_index));
+        let reward_denominator = INFLATION_DEN * A_SCALE;
+
+        // blend get 60% of rewards while leaders get the 40% remaining.
+        // Casting as Value truncate the floating points
+        let blend_reward = (reward_numerator * 6 / (reward_denominator * 10)) as Value;
+        let leader_reward = (reward_numerator * 4 / (reward_denominator * 10)) as Value;
```
danielSanchezQ: The `6` and the `4` should be named consts too. I would say the `10` too.

PR #2391, opened 2026-03-17
chore: respect cucumber node log level in local tests

andrussal:

This fixes local cucumber node logging to actually respect `LOGOS_BLOCKCHAIN_LOG_LEVEL` instead of falling back to the node default `DEBUG`. This should also reduce space used by tests.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/tests/src/cucumber/defaults.rs
+++ b/tests/src/cucumber/defaults.rs
@@ -1,5 +1,6 @@
 use std::{fs, path::PathBuf};
 
+use lb_testing_framework::LOGOS_BLOCKCHAIN_LOG_LEVEL;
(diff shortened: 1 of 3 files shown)


hansieodendaal:
Setting `LOGOS_BLOCKCHAIN_LOG_LEVEL` does influence the node log level, I have verified this. So we are logging at the level we want, which is DEBUG. I propose to solve the high logging space hog in cucumber tests, we add an env var that will instruct deleting the work_dir for successful scenarios as part of `.after` function in cucumber.rs - I will make a PR for that.

PR #2392, opened 2026-03-17
fix: bypass Blend when proposing first block of a new session

ntn-x2:

As long as Blend uses block-based scheduling, we have an issue where block proposals fail to be accepted by the network if the proposer rotates to a new session when self-applying the new block. To avoid this, a temporary workaround is to bypass Blend for the first block of a new session, so that all nodes can rotate their session accordingly.

Nevertheless, this alone is not enough: after self-applying a block and rotating a session, a node will generate new PoQ proofs and start sending new cover messages, which are still invalid from the perspective of all other nodes until they also apply the new block. This is now less likely to happen, but if a cover message is generated and sent before the block proposal is received by all nodes, the node can still be marked as spammy and banned. Hence, we need to temporarily disable banning as well, which will happen in a PR on top of this one.

Yes.

@ntn-x2 

N/A

No.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/blend/network/src/core/with_core/behaviour/handler/mod.rs
+++ b/blend/network/src/core/with_core/behaviour/handler/mod.rs
@@ -184,7 +184,9 @@ where
         if let Poll::Ready(output) = self.monitor.poll(cx) {
             match output {
                 Some(ConnectionMonitorOutput::Spammy) => {
-                    self.close_substreams();
+                    // TODO: Re-enable this once we have fixed Blend observation
+                    // window range values.
+                    // self.close_substreams();
                     self.pending_events_to_behaviour
                         .push_back(ToBehaviour::SpammyPeer);
                 }
--- a/blend/network/src/core/with_core/behaviour/mod.rs
+++ b/blend/network/src/core/with_core/behaviour/mod.rs
@@ -605,25 +605,14 @@ impl<ProofsVerifier, ObservationWindowClockProvider>
 
     /// Mark the connection with the sender of a malformed message as malicious
     /// and instruct its connection handler to drop the substream.
-    #[expect(
-        clippy::needless_pass_by_ref_mut,
-        reason = "TODO: enable this logic after investigating session/epoch transition issues"
-    )]
-    #[expect(
-        clippy::unused_self,
-        reason = "TODO: enable this logic after investigating session/epoch transition issues"
-    )]
     fn close_spammy_connection(
         &mut self,
         (peer_id, connection_id): (PeerId, ConnectionId),
         reason: SpamReason,
     ) {
-        tracing::debug!(target: LOG_TARGET, ?peer_id, ?connection_id, ?reason, "Peer has been marked as spammy, but not closing the connection just for debugging");
-        // TODO: Enable this logic after investigating session/epoch transition
-        // issues tracing::debug!(target: LOG_TARGET, "Closing
-        // connection {connection_id:?} with spammy peer {peer_id:?}.");
-        // self.set_connection_to_spammy((peer_id, connection_id), reason);
-        // self.close_connection((peer_id, connection_id));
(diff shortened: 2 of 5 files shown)


services/chain/chain-leader/src/lib.rs
```diff
@@ -581,10 +621,23 @@ where
...
             .clone()
             .try_apply_header::<Groth16LeaderProof, HeaderId>(slot, &proof, ledger_config)?;
 
+        let blend_session_after = *ledger_state
+            .active_sessions()
+            .get(&ServiceType::BlendNetwork)
+            .unwrap();
+
+        let is_new_blend_session =
+            blend_session_after > blend_session_before && blend_session_after > 0;
```
youngjoon-lee, line 645: I think it's safer to use `unwrap_or` instead of `unwrap` because the chain-leader service doesn't know whether the ledger state actually contains the blend service or not.
```diff
diff --git a/services/chain/chain-leader/src/lib.rs b/services/chain/chain-leader/src/lib.rs
index 2d815bf9..a198f316 100644
--- a/services/chain/chain-leader/src/lib.rs
+++ b/services/chain/chain-leader/src/lib.rs
@@ -624,7 +624,7 @@ where
         let blend_session_before = *ledger_state
             .active_sessions()
             .get(&ServiceType::BlendNetwork)
-            .unwrap();
+            .unwrap_or(&0);
 
         ledger_state = ledger_state
             .clone()
@@ -633,10 +633,9 @@ where
         let blend_session_after = *ledger_state
             .active_sessions()
             .get(&ServiceType::BlendNetwork)
-            .unwrap();
+            .unwrap_or(&0);
 
-        let is_new_blend_session =
-            blend_session_after > blend_session_before && blend_session_after > 0;
+        let is_new_blend_session = blend_session_after > blend_session_before;
 
         let mut valid_txs = Vec::new();
         let mut invalid_tx_hashes = Vec::new();
```

services/chain/chain-leader/src/lib.rs
```diff
@@ -650,9 +708,26 @@ where
...
-        debug!(target: LOG_TARGET, "Successfully applied our own proposed block. Publishing it to the blend network: {:?}", block.header().id());
 
-        blend_adapter.publish_proposal(block.to_proposal()).await;
+        match proposal_strategy {
+            BlockProposalStrategy::Blend(blend_adapter) => {
+                debug!(target: LOG_TARGET, "Successfully applied our own proposed block. Publishing it to the blend network: {:?}", block.header().id());
+                blend_adapter.publish_proposal(block.to_proposal()).await;
+            }
+            BlockProposalStrategy::Broadcast { adapter, settings } => {
+                debug!(target: LOG_TARGET, "Successfully applied our own proposed block. First of a new session, broadcasting it directly: {:?}", block.header().id());
```
youngjoon-lee:
```suggestion
                debug!(target: LOG_TARGET, "Successfully applied our own proposed block. First of a new blend session, broadcasting it directly: {:?}", block.header().id());
```
I think we can make the log more explicitly like this.

services/chain/chain-leader/src/lib.rs
hansieodendaal: Agreed, missed that.

ntn-x2: Good point. Did it in https://github.com/logos-blockchain/logos-blockchain/pull/2392/commits/70d5483931f2aea78b8887bd7f4bc0a0c7de558b.

services/chain/chain-leader/src/lib.rs
ntn-x2: https://github.com/logos-blockchain/logos-blockchain/pull/2392/commits/70d5483931f2aea78b8887bd7f4bc0a0c7de558b

PR #2393, opened 2026-03-17
test(framework): improve fork monitor health reporting

andrussal:

ClusterForkMonitor was too weak for the nightly. It only caught proven LIB forks, so green runs could still hide long stalls, heavy lag, or unhealthy fragmentation. This change keeps the existing fork check, but also adds runtime health checks and a much more useful summary: tip/LIB stall detection, per-node lag detection, grouped final tips/LIBs, and ancestry-based relation output so we can tell "same chain but lagging" from an actual fork.`

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/tests/testing_framework/src/workloads/fork_monitor.rs
+++ b/tests/testing_framework/src/workloads/fork_monitor.rs
@@ -21,7 +21,7 @@ use crate::{
 
 const DEFAULT_TIP_STALL_THRESHOLD: Duration = Duration::from_secs(180);
 const DEFAULT_NODE_TIP_STALL_THRESHOLD: Duration = Duration::from_secs(360);
-const DEFAULT_LIB_STALL_THRESHOLD: Duration = Duration::from_secs(1200);
+const DEFAULT_LIB_STALL_THRESHOLD: Duration = Duration::from_secs(360);
 const PROGRESS_LOG_INTERVAL: Duration = Duration::from_secs(60);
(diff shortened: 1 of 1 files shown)


tests/testing_framework/src/workloads/fork_monitor.rs
```diff
@@ -1,15 +1,29 @@
...
+use crate::{
+    NodeHeadSnapshot,
+    framework::LbcEnv,
+    node::configs::{default_e2e_deployment_settings, deployment::TopologyConfig},
+    workloads::LbcBlockFeedEnv,
+};
+
+const DEFAULT_TIP_STALL_THRESHOLD: Duration = Duration::from_secs(180);
+const DEFAULT_NODE_TIP_STALL_THRESHOLD: Duration = Duration::from_secs(360);
+const DEFAULT_LIB_STALL_THRESHOLD: Duration = Duration::from_secs(1200);
```
hansieodendaal: `LIB` advances all the time, same as `TIP`, so `DEFAULT_LIB_STALL_THRESHOLD` should be in line with `DEFAULT_TIP_STALL_THRESHOLD`

andrussal: I made it shorter, but I still think it should be longer than the TIP stall threshold. At startup, LIB can stay at genesis for a while, and later during forks/divergence it can also lag behind TIP.

PR #2399, opened 2026-03-19
feat(zone-sdk): track finalized deposits

youngjoon-lee:
(This PR is not compiled yet. A PR for new `DepositOp` is coming soon by Alex. Before that, I want to open this PR and discuss about the approach.)

### NOTE

First of all, I want to mention two things:
- Since zone-sdk currently support subscribing to **finalized** inscriptions, this PR also focuses on **finalized** deposits. Subscribing to unfinalized parts is out of scope.
- This PR is **not** related to "atomic deposits" that we tried in Lisbon. Here, we touch only `ZoneIndexer` which just reads transactions (operations) finalized in Bedrock, regardless of how they submitted. 

### Main change

- Now, `ZoneIndexer::follow` returns `Stream<Item = ZoneMessage>` instead of `Stream<Item = ZoneBlock>`.
  - `ZoneMessage` is an enum of `ZoneBlock` and `Deposit`.
- Now, `ZoneIndexer::next_messages` returns `Stream<Item = ZoneMessage>` instead of `PollResult`. (FYI, this function is for bootstrapping a sequencer by catching up historic zone messages, unlike `follow` above.)
  - In other words, it doesn't work as "pagination" anymore. Instead, it returns a stream, so that the caller can stop when he wants.
  - I deprecated `Cursor`, but this function still provides a way to continue from the last `ZoneBlock` (not `ZoneMessage`) that the caller has.
  - Why did I change it? It's because deposits don't have IDs, unlike inscriptions. If the function returned page 7 consisting entirely  of deposits, it would be impossible to set the next cursor, and page 8 would return the same deposits again. It is an extreme case, but there are also other subtle issues that I didn't mention here. I've thought about many ways to keep the pagination approach, but ends up with the streaming approach and deprecating `Cursor`.
  - This function can be improved later according to the LEZ's needs, but I believe the new design is good enough.
  - It would be easier to review only the new version than comparing diffs.

### Refactoring

- I defined a trait `adapter::Node` which abstracts HTTP API calls, so that we can write tests with mocked nodes.
  - I tried to split this as a separate PR, but decided to include this to show you the working tests. Change looks big, but is not difficult to read.

Please let me know how you think about the new streaming approach of `next_messages`.

I believe so

@youngjoon-lee made these changes based on the original implementers written by @pradovic and @davidrusu.

n/a

n/a

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/zone-sdk/src/indexer.rs
+++ b/zone-sdk/src/indexer.rs
@@ -360,11 +360,12 @@ mod tests {
             slot_to: Slot,
             _channel_id: ChannelId,
         ) -> Result<impl Stream<Item = (ZoneMessage, Slot)>, lb_common_http_client::Error> {
-            Ok(self
-                .messages
-                .iter()
-                .filter(|(_, slot)| *slot >= slot_from && *slot <= slot_to)
-                .cloned())
+            Ok(futures::stream::iter(
+                self.messages
+                    .iter()
+                    .filter(move |(_, slot)| *slot >= slot_from && *slot <= slot_to)
+                    .cloned(),
+            ))
         }
     }
 }


consensus/cryptarchia-engine/src/time.rs
```diff
@@ -132,6 +144,22 @@ impl Add<u32> for Epoch {
...
+    fn sub(self, rhs: u64) -> Self::Output {
+        Self(self.0 - rhs)
+    }
+}
+
+impl Mul<u64> for Slot {
+    type Output = Self;
+
+    fn mul(self, rhs: u64) -> Self::Output {
+        Self(self.0 * rhs)
```
danielSanchezQ: Should they be checked or wrapped instead?

core/src/mantle/gas.rs
```diff
@@ -40,6 +43,7 @@ impl GasConstants for MainnetGasConstants {
     const TRANSFER: Gas = 2705;
     const CHANNEL_INSCRIBE: Gas = 22;
     const CHANNEL_SET_KEYS: Gas = 22;
+    const CHANNEL_DEPOSIT: Gas = 0;
```
danielSanchezQ: Are these temporary values? Why is it 0 gas?

zone-sdk/src/adapter.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    async fn lib_stream(&self) -> Result<impl Stream<Item = BlockInfo>, Error> {
+        self.client.get_lib_stream(self.base_url.clone()).await
+    }
+
+    async fn zone_messages_in_block(
+        &self,
+        id: HeaderId,
+        channel_id: ChannelId,
+    ) -> Result<Vec<ZoneMessage>, Error> {
```
danielSanchezQ: Why not use stream if it comes from request?

zone-sdk/src/adapter.rs
```diff
@@ -0,0 +1,130 @@
...
+            .filter_map(|op| op_to_zone_message(op, channel_id))
+            .collect())
+    }
+
+    async fn zone_messages_in_blocks(
+        &self,
+        slot_from: Slot,
+        slot_to: Slot,
+        channel_id: ChannelId,
+    ) -> Result<Vec<(ZoneMessage, Slot)>, Error> {
```
danielSanchezQ: Same here, is it better to have vec than stream in this interfaces?

zone-sdk/src/indexer.rs
```diff
@@ -16,315 +12,329 @@ pub enum Error {
...
-                                id: inscribe.id(),
-                                data: inscribe.inscription.clone(),
-                            })
-                        }
-                        _ => None,
-                    })
-                    .collect();
-
-                Some(futures::stream::iter(zone_blocks))
+                Some(futures::stream::iter(messages))
```
danielSanchezQ: Yeah, this already tells me that we should send a stream of blocks in those petitions and then just bubble up the transformed stream where needed instead.

core/src/mantle/gas.rs
youngjoon-lee: Gas for deposit is 0: https://www.notion.so/nomos-tech/v1-3-Mantle-Specification-330261aa09df80a899a6efd74f12a7c4?source=copy_link#330261aa09df81308acfc11b13716fab, because it doesn't require proof validation: https://www.notion.so/nomos-tech/v1-3-Gas-Cost-Determination-330261aa09df80338881c3924c48867f?source=copy_link#330261aa09df81dfb94bf0fc24f2d0ee.

consensus/cryptarchia-engine/src/time.rs
youngjoon-lee: Yes. This is actually a broken windows theory. 
[e9d4c63](https://github.com/logos-blockchain/logos-blockchain/pull/2399/commits/e9d4c63e1e0d7dc9a6afc1683587045db3e7865d) I reverted this change by using `saturating_add` and `checked_mul` explicitly in upper layers.

zone-sdk/src/adapter.rs
youngjoon-lee: Right, we can return a stream: [af74926](https://github.com/logos-blockchain/logos-blockchain/pull/2399/commits/af749262ffbb5688bb4efb4dafa89a4db7807240)

zone-sdk/src/adapter.rs
youngjoon-lee: done [af74926](https://github.com/logos-blockchain/logos-blockchain/pull/2399/commits/af749262ffbb5688bb4efb4dafa89a4db7807240)

zone-sdk/src/indexer.rs
youngjoon-lee: [af74926](https://github.com/logos-blockchain/logos-blockchain/pull/2399/commits/af749262ffbb5688bb4efb4dafa89a4db7807240) done

PR #2422, opened 2026-03-24
fix(blend): more Blend fixes

ntn-x2:
Catch-all PR for some more Blend fixes, hopefully the last ones. Sorry for this mixed PR. If it's too much, I can split it up into multiple, smaller ones that are semantically grouped.

* We used to bubble up to the Blend service all copies of the same message, which were then added to the queue of processed messages and tried to be published. Of course, the second time it did not work but was failing with a `NoPeers` error. Now we prevent duplicate messages from making it to the service by keeping a cache of processed messsages in the behaviour, so we can assert in the state recovery object that no two processed messages are equal. This also gets us rid of the `NoPeers` error which is not really meant to be there when publishing a message. The cache is passed on to the `old_session` object on session rotation and then freed when STP elapses
* We used to `publish` messages coming from the old session, meaning sending them back to the sender. This effectively did not happen because we still check for duplicates before sending out the message, but it was logically wrong at the higher level, so I changed the `publish` function for the `old_session` with a `forward` equivalent which takes an `except`.
* Because our Blend Swarm has (rightfully) a keepalive timeout of `0`, we used to immediately drop edge connections before they had a chance to send the message. This is now fixed by overriding the `keepalive` function of the edge behaviour's connection handler, which looks at the state of the handler to decide whether a connection is valid or not.
* Do not pre-generate a lot of proofs. We now use the number of encapsulation layers to pre-generate only enough proofs to cover the next message. This is easier on startup and also leads to less waste when we rotate sessions or epochs, invalidating previous proofs.
* Fixed a panic in the Blend edge service which was expecting a libp2p `Keypair` enum instead of a `ed25515::Keypair`, which does not have the enum tagging variant and can hence be deserialized from a 32 byte array
* More/improved logging for easier debugging

### DIFF
--- a/blend/network/src/core/with_core/behaviour/mod.rs
+++ b/blend/network/src/core/with_core/behaviour/mod.rs
@@ -811,7 +811,7 @@ where
     ProofsVerifier: encap::ProofsVerifier,
 {
     /// Publish an already-encapsulated message to all connected peers
-    /// in the current or old session.
+    /// in the current session.
     ///
     /// Before the message is propagated, its public header is validated to
     /// make sure the receiving peer won't mark us as malicious.


blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -821,7 +824,9 @@ where
         &mut self,
         message: EncapsulatedMessage,
     ) -> Result<(), Error> {
-        self.validate_and_forward_message_except(message, None)
+        let validated_message =
+            self.validate_encapsulated_message_public_header_with_current_session(message)?;
+        self.forward_validated_message_and_maybe_exclude(&validated_message, None)
```
youngjoon-lee, line 829: Q: Now we explicitly use validate a message using the **current** session because this function is for submitting **new** messages to the blend network? If so, yeah, the messages should be propagated through the current session. I think we need to update the doc comment: `current or old session` -> `current session`.

blend/network/src/core/with_core/behaviour/mod.rs
```diff
@@ -915,12 +907,20 @@ where
...
         ) else {
             return;
         };
+
+        // Exit early if we've processed this message already and we know it's a valid
+        // one, so no need to check it again to potentially mark the peer as malicious.
+        if self.message_cache.contains(&message_identifier) {
+            tracing::trace!(target: LOG_TARGET, "Message with id {message_identifier:?} already processed previously. Dropping it.");
+            return;
+        }
```
youngjoon-lee, line 922: It looks right to check `message_cache` here.

I want to check if I understood the comment correctly.
Without this logic, this `Behaviour` will bubble up the same messages to the upper layer, resulting in sending the same messages to the peers. Then, other peers who will receive those messages will mark **me** as malicious. 
Did I get it correctly? If so, I think we can modify the comment like: `mark the peer` -> `make this node`.

consensus/cryptarchia-engine/src/config.rs
```diff
@@ -105,10 +105,10 @@ pub const fn base_period_length(
 
 #[must_use]
 pub const fn average_slots_for_blocks(
-    num_blocks: NonZero<u32>,
+    security_param: NonZero<u32>,
     slot_activation_coeff: NonNegativeRatio,
 ) -> NonZero<u64> {
-    NonZero::new((num_blocks.get() as f64 / slot_activation_coeff.as_f64()).floor() as u64)
+    NonZero::new((security_param.get() as f64 / slot_activation_coeff.as_f64()).floor() as u64)
         .expect("base_period_length with proper configuration should never be zero")
```
youngjoon-lee, line 112: I think it's better to use the previous name `num_blocks` because this function is used for not only `security_param` but also any number of blocks: https://github.com/logos-blockchain/logos-blockchain/blob/e48b80992e9f79b6ef46111b26cc8f909c1f733b/nodes/node/binary/src/config/cryptarchia/deployment.rs#L46-L50

blend/network/src/core/with_core/behaviour/mod.rs
ntn-x2: No the behaviour has remained unchanged. But since the `validate_and_forward_message_except` function was only used one within this function, I decided to remove one level of indirection and include everything in this function. But, if you take the content of the old and new functions, they are exactly the same.

blend/network/src/core/with_core/behaviour/mod.rs
ntn-x2: No. Or rather, that also happens, but the comment in this case is about a different thing. Since we update the message cache only at the very end, before bubbling up the message to the service, when we know the message is valid, we also know that if we get a cache hit for another message with the same nullifier, that message will also be valid, so we don't need to re-verify it to decide whether the peer is malicious or not. The only thing we need to verifying before looking up the cache, is if the message is a duplicate from the same sender peer, which we already do, in which case we want to mark the peer as malicious because it's sending a duplicate message. Anyway the spec was changed yesterday to implement a different mechanism for this, so I will update the code accordingly once I am sure everything works as expected in the current version.

consensus/cryptarchia-engine/src/config.rs
ntn-x2: Yeah I found `num_blocks` to be too generic and not really descriptive, but anyway this is entirely out of scope, so I'll revert the change here. Sorry about that 😊

blend/network/src/core/with_core/behaviour/mod.rs
ntn-x2: Ah no you are right. The `publish` is concerned only with the current session, while the `forward` looks at the right session. I will fix the comment, thanks! I was confusing that function with another one when answering your comment.

PR #2424, opened 2026-03-24
feat(autonat): remove explicit setting of server mode

pradovic:

Removed forced Kademlia Server mode and rely on libp2p’s automatic mode switching based on confirmed external addresses (**static** mode still sets it manually, so behavior remains unchanged in that case), and added logging for the mode state changed event.

Yes

@pradovic 

Yes

No, it just adds a possibility for a user to add another candidate for probe in `traversal` mode.

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/libp2p/src/behaviour/nat/inner.rs
+++ b/libp2p/src/behaviour/nat/inner.rs
@@ -18,7 +18,7 @@ use libp2p::{
 };
 use rand::RngCore;
 use tokio::sync::mpsc::UnboundedReceiver;
-use tracing::{debug, error, info};
+use tracing::{error, info};
 
 use crate::{
     behaviour::nat::{
@@ -63,11 +63,6 @@ where
     autonat_client_tick_interval: Duration,
     /// Current local address that is being managed
     local_address: Option<Multiaddr>,
-    /// Optional external address candidate provided by the operator to be
-    /// verified via `AutoNAT`. Useful for nodes with manual port forwarding
-    /// where the correct public address may not be discovered automatically.
-    /// Emitted once as a `NewExternalAddrCandidate` on first poll.
-    external_address_candidate: Option<Multiaddr>,
 }
 
 pub type NatBehaviour<Rng> = InnerNatBehaviour<Rng, ProtocolManager, SystemGatewayDetector>;
@@ -113,7 +108,6 @@ where
             next_autonat_client_tick: OptionFuture::default().fuse(),
             autonat_client_tick_interval,
             local_address: None,
-            external_address_candidate: settings.external_address.clone(),
         }
     }
 }
@@ -210,11 +204,6 @@ where
         &mut self,
         cx: &mut Context<'_>,
     ) -> Poll<ToSwarm<Self::ToSwarm, THandlerInEvent<Self>>> {
-        if let Some(addr) = self.external_address_candidate.take() {
-            debug!("Emitting external address candidate for AutoNAT verification: {addr}");
-            return Poll::Ready(ToSwarm::NewExternalAddrCandidate(addr));
-        }
-
         if let Poll::Ready(to_swarm) = self.autonat_client_behaviour.poll(cx) {
             // Forward all autonat results to the state machine unconditionally.
             // The state machine's own guard clauses filter by the address it is
--- a/libp2p/src/config/nat/mod.rs
+++ b/libp2p/src/config/nat/mod.rs
@@ -9,11 +9,6 @@ pub struct TraversalSettings {
     pub autonat: autonat_client::Settings,
     pub mapping: mapping::Settings,
     pub gateway_monitor: gateway::Settings,
-    /// Optional external address candidate to verify via `AutoNAT`.
-    /// Useful for nodes with manual port forwarding where the correct
-    /// public address may not be discovered automatically.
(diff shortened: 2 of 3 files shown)


libp2p/src/behaviour/nat/inner.rs
```diff
@@ -63,6 +63,11 @@ where
     autonat_client_tick_interval: Duration,
     /// Current local address that is being managed
     local_address: Option<Multiaddr>,
+    /// Optional external address candidate provided by the operator to be
+    /// verified via `AutoNAT`. Useful for nodes with manual port forwarding
+    /// where the correct public address may not be discovered automatically.
+    /// Emitted once as a `NewExternalAddrCandidate` on first poll.
+    external_address_candidate: Option<Multiaddr>,
```
ntn-x2: We already have the `Static` variant to configure the NAT machine that simply skips NAT probes altogether. Can't we just use that to emit the event we want kad to process?

pradovic: This was added just as a convinience (nice to have) if you don't want to blindly trust the static setup, but still want to try the added public address. We can remove it but I would argue it can be useful as a convenience for some home users with weird mappings that would still like to provide additional candidate but not go into server mode without the check (probing). Let me know what you think.

The main change is removal of manual ` kademlia.set_mode(Some(Mode::Server));` and trusting the autonat in traversal mode, this is just nice to have.

Static mode still work as is was before, no change.

ntn-x2: Mhh but then why do we need the static mode? In my opinion we should only use one of them. If we go with this one, we should remove the static variant altogether, don't you agree?

pradovic: I agree that static mode already covers the case where we want to skip probing and trust the configuration.

At the same time, I think there is a small gap between the two modes:
	• static → fully trusted, no probing (can advertise a wrong address if misconfigured)
	• traversal → fully automatic, but may not discover the correct mapping in some edge cases

The optional external_address in traversal is just meant to bridge that gap:
	• allow users to provide a likely public address
	• but still verify it via AutoNAT before using it
	• so we avoid advertising incorrect addresses

So the intended usage would be:
	1. static → bootstrap nodes / fully trusted setups
	2. traversal → default for most users (should cover the majority of cases)
	3. traversal + external_address → edge cases where discovery is not enough, but we still want safety via probing

If probing does not work (e.g. restricted environments), static is still the fallback.

This is not meant to replace static, just to provide a safer intermediate option. For example, in some home setups it can be useful to start in traversal mode and only transition to server once reachability is actually confirmed.

Happy to remove it if you think the extra complexity is not worth it. I just did not want to switch completely to this as there might be some setup where static is the only solution (for example maybe bootnodes).

hansieodendaal: I like the addition of the optional `external_address_candidate` address here. I would use it.

ntn-x2: > but may not discover the correct mapping in some edge cases

What are such cases? In my opinion, if that is an issue, then that should be fixed instead of providing two ways that almost do the same thing.

ntn-x2: I am all in favor of having this feature you introduce here, but I would like it to be a full replacement of the current static configuration, if possible. Else why not use the default NAT configuration and let the NAT machine figure it out? If the NAT machine is broken or does not work properly, we should fix that in my opinion, rather than having what, in that case, feels like a hacky workaround? Maybe I am still missing something tho.

pradovic: I think the key difference is that static and traversal + external_address have different safety properties:
	• static → fully trusted, no probing (we immediately advertise the address, even if it is wrong)
	• traversal + external_address → user provides a hint, but it is still verified via AutoNAT before being used

So static is a trust override, while the candidate is a safe hint.

Also, static is still needed for cases like bootstrap nodes, where AutoNAT probing may not be reliable or even possible initially, but we still want those nodes to act as servers.

So I see it as:
	• traversal → default
	• traversal + candidate → safe extension for regular users with weird setups (different map ports, multiple NATs, where identify won't show correct observed address for probing)
	• static → explicit override, required for bootnodes I think (as they cannot be probed between themselves over public address, so they would switch to server only until nodes starts to appear if not for static mode)

I agree that ideally we converge to a simpler model, but I think static is still needed as an escape hatch, while the candidate approach improves the default path without removing safety.

pradovic: Also, I agree that it seems hacky, happy to change. Just showing my 2 c of why I added it as optional convenience. It was good for my home setup, for example, which might be the edge case as it would not be discoverable by autonat in any case but I don't like the idea of using static mode which is not checked by other nodes.

pradovic: @ntn-x2 As I need more time to setup a functional testing environment I have removed the optional external_address https://github.com/logos-blockchain/logos-blockchain/pull/2424/changes/6e88fc5e844ffd64af291ad79a880febe54d06e8. 

I will open a separate PR for this if needed, and we can discuss there, to avoid blocking this PR.

PR #2427, opened 2026-03-25
fix: slot ticks debugging

ntn-x2:
Make all components resilient to missed slot and round ticks. In most of the cases, this should not happen, but if it does, it's fair to expect all nodes to remain in sync and skip missed slots and rounds instead of get a burst of the missed ones, which can cause more problems than it tries to fix.

So main change is to the time service channel, which is not a broadcast (with a buffer size of `3`) anymore, but a watch channel, and the Blend stack is configured to skip missed round ticks, which is not bad since we would, in the worst case, simply generate less cover traffic, which was the behaviour also in case we were lagging behind, with the difference that in that case we were also leaking the fact that we were lagging behind.

### DIFF
--- a/blend/network/src/core/with_core/behaviour/tests/utils.rs
+++ b/blend/network/src/core/with_core/behaviour/tests/utils.rs
@@ -21,7 +21,7 @@ use libp2p::{
     identity::{PublicKey, ed25519},
 };
 use libp2p_swarm_test::SwarmExt as _;
-use tokio::time::interval;
+use tokio::time::{MissedTickBehavior, interval};
 use tokio_stream::wrappers::IntervalStream;
 
 use crate::core::{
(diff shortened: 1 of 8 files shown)


danielSanchezQ (approved):

blend/network/src/core/with_core/behaviour/tests/utils.rs
```diff
@@ -38,7 +38,13 @@ impl IntervalStreamProvider for IntervalProvider {
 
     fn interval_stream(&self) -> Self::IntervalStream {
         let range = self.1.clone();
-        Box::new(IntervalStream::new(interval(self.0)).map(move |_| range.clone()))
+        let interval = {
+            let mut interval = interval(self.0);
+            interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
```
hansieodendaal, line 43: I see we also have `impl IntervalStreamProvider for TestObservationWindowProvider {`
Worthwhile to set `interval.set_missed_tick_behavior(MissedTickBehavior::Skip);` there as well?

ntn-x2: That's used for tests. We have a bunch of other places where we use `tokio::interval` that do not have this set. Also, if the test starts failing, we could look into it, but I don't think it's necessary in this case since unit tests are quite lightweight.

PR #2428, opened 2026-03-25
refactor(ledger_tx): Make the Ledger Transaction an Operation

thomaslavaur:

This PR makes the Ledger Transaction becoming a Transfer Operation. This update is specified in the [[RFC] Improve Mantle Transaction hash](<https://www.notion.so/nomos-tech/RFC-Improve-Mantle-Transaction-hash-31e261aa09df80a08146e4978d2da3e0>)

Yes

@thomaslavaur 

No! The Genesis Block specification also need an update and will be done before merging.

Yes. The Genesis Block specification needs an udpate.

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/zone-sdk/src/sequencer.rs
+++ b/zone-sdk/src/sequencer.rs
@@ -781,11 +781,8 @@ fn create_set_keys_tx(
         keys,
     };
 
-    let ledger_tx = LedgerTx::new(vec![], vec![]);
-
     let set_keys_tx = MantleTx {
         ops: vec![Op::ChannelSetKeys(set_keys_op)],
-        ledger_tx,
         storage_gas_price: 0,
         execution_gas_price: 0,
     };
@@ -795,8 +792,6 @@ fn create_set_keys_tx(
 
     SignedMantleTx {
         ops_proofs: vec![OpProof::Ed25519Sig(signature)],
-        ledger_tx_proof: ZkKey::multi_sign(&[], tx_hash.as_ref())
-            .expect("multi-sign with empty key set"),
         mantle_tx: set_keys_tx,
     }
 }


core/src/block/mod.rs
```diff
@@ -401,17 +401,17 @@ mod tests {
     }
 
     #[derive(Clone, Copy, Debug)]
-    pub struct TestTx;
-    impl Transaction for TestTx {
-        const HASHER: TransactionHasher<Self> = |_tx| TxHash(Fr::ZERO);
+    pub struct TestTransferOp;
```
danielSanchezQ: I think for the sake of testing, maybe it would be better that this testing Tx has a single `TransferOp` embedded itself.

core/src/mantle/ops/mod.rs
```diff
@@ -35,14 +43,15 @@ use crate::{
 /// `#[serde(untagged)]` enums, binary deserialization is routed through
 /// [`OpWireVisitor`], which correctly handles `opcode` to select the
 /// appropriate variant.
-#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+#[derive(Clone, Debug, Eq, PartialEq)]
```
danielSanchezQ, line 46: Why did we removed `Hash`?

core/src/mantle/genesis_tx.rs
```diff
@@ -41,14 +44,19 @@ impl GenesisTx {
...
+            }
+            _ => return Err(Error::MissingTransfer),
         }
 
-        // Genesis transactions must contain exactly one inscription as the first op
-        // and then may contain other SDP declarations
-        let mut ops = mantle_tx.ops.iter();
         match ops.next() {
             Some(Op::ChannelInscribe(op)) => valid_cryptarchia_inscription(op)?,
             _ => return Err(Error::MissingInscription),
```
danielSanchezQ: We can actually reformat this to enforce types. We should be able to collect into a `[Op; 2]` to enforce the size. Then we can match:
```rust
match ops {
    [Op::ChannelInscribe(inscribe), Op::Transfer(transfer)] |  [Op::Transfer(transfer), Op::ChannelInscribe(inscribe),] => {
    ...
}
    _ => return Err(...)
}
```

The above is just an idea, haven't really tried.

core/src/mantle/genesis_tx.rs
```diff
@@ -122,9 +126,17 @@ impl GasCost for GenesisTx {
...
+        }
+    }
+
+    fn genesis_transfer(&self) -> &TransferOp {
+        // Safe to unwrap because we validated this in from_tx
+        match &self.mantle_tx().ops[0] {
+            Op::Transfer(op) => op,
+            _ => unreachable!("GenesisTx always has a valid transfer as first op"),
         }
     }
```
danielSanchezQ, line 141: We could use the same pattern as above and have it in a single method that returns a tuple.

ledger/src/cryptarchia/mod.rs
```diff
@@ -327,36 +330,40 @@ impl LedgerState {
...
+                    }
+                    let utxo;
+                    (self.utxos, utxo) = self
+                        .utxos
+                        .remove(input)
+                        .map_err(|_| LedgerError::InvalidNote(*input))?;
+                    balance = balance
+                        .checked_add(utxo.note.value.into())
+                        .ok_or(LedgerError::Overflow)?;
+                    pks.push(utxo.note.pk);
```
danielSanchezQ, line 349: This may be candidate or be extracted as a method.

ledger/src/lib.rs
```diff
@@ -495,7 +497,11 @@ mod tests {
...
 
         // Verify output was created
-        let output_utxo = tx.mantle_tx.ledger_tx.utxo_by_index(0).unwrap();
-        assert!(new_state.latest_utxos().contains(&output_utxo.id()));
+        if let Op::Transfer(transfer_op) = &tx.mantle_tx.ops[0] {
+            let output_utxo = transfer_op.utxo_by_index(0).unwrap();
+            assert!(new_state.latest_utxos().contains(&output_utxo.id()));
+        } else {
+            panic!("first op must be a transfer")
+        }
```
danielSanchezQ, line 505: Can we enforce that the first op is always a transfer in type level? I think if this always have to be the case we should enforce it. It can be outside the scope of this PR, but this can be better for sure.

services/wallet/src/lib.rs
```diff
@@ -617,13 +618,17 @@ where
                     Self::prove_leader_claim_op(claim_op.clone(), tx_hash, &ledger, wallet, kms)
                         .await?
                 }
+                // TODO: Mantle Tx Builder should come with different ledger inputs for each
```
danielSanchezQ: for *each* what?

wallet/src/lib.rs
```diff
@@ -27,18 +29,19 @@ pub use crate::voucher::Vouchers;
...
-    pub ledger_txs: Vec<LedgerTx>,
+    pub transfers: Vec<TransferOp>,
 }
 
 impl<Tx: AuthenticatedMantleTx> From<Block<Tx>> for WalletBlock {
     fn from(block: Block<Tx>) -> Self {
+        let mut transfers: Vec<TransferOp> = vec![];
+        for auth_tx in block.transactions() {
+            transfers.append(&mut auth_tx.mantle_tx().transfers());
+        }
```
danielSanchezQ: Why do you keep a `&mut` on them?
I think this is easier:

```suggestion
        let mut transfers: Vec<TransferOp> = block.transactions().flat_map(|auth_tx| auth_tx.mantle_tx().transfers().copied()).collect();
```

wallet/src/lib.rs
```diff
@@ -486,22 +494,26 @@ mod tests {
...
+            // ensure change was returned to alice
+            assert_eq!(
+                transfer_op.outputs,
+                vec![Note {
+                    value: 2075,
+                    pk: alice,
+                }]
+            );
+        } else {
+            panic!("last op must be a transfer")
```
danielSanchezQ, line 515: Wait why? We really need to enforce this code-wise. Do we have this as part of the encoding spec as well? or just in the mantle one?
Or is this just because of the cases of this test?

thomaslavaur: just this test

ledger/src/lib.rs
thomaslavaur: it's only in this test

ledger/src/cryptarchia/mod.rs
thomaslavaur: it was just moved from previous code already like this I can do it in next PR as this should be in Mantle Ledger I think?

core/src/mantle/ops/mod.rs
thomaslavaur: Because TransferOp cannot support Hash without introducing it to the note (and because it's not usefull in the current implem)?

thomaslavaur: Should I add it back and introduce it for note too?

core/src/mantle/genesis_tx.rs
thomaslavaur: Noted for next PR cleanup

core/src/mantle/genesis_tx.rs
thomaslavaur: noted for next PR

ledger/src/cryptarchia/mod.rs
thomaslavaur: I agree I'll do it in the cleanup

wallet/src/lib.rs
thomaslavaur: done

core/src/block/mod.rs
thomaslavaur: What do you mean? A Mantle Tx?

core/src/mantle/genesis_tx.rs
thomaslavaur: But the transaction can also include SDP declarations. Only the two first are enforced

core/src/mantle/ops/mod.rs
youngjoon-lee: I also think that it's fine to remove `Hash` since it's not used anywhere.

core/src/mantle/genesis_tx.rs
```diff
@@ -41,14 +44,19 @@ impl GenesisTx {
             return Err(Error::InvalidGenesisGasPrice);
         }
 
-        // Genesis transactions should not have any inputs
-        if !mantle_tx.ledger_tx.inputs.is_empty() {
-            return Err(Error::UnepectedInput);
+        // Genesis transactions must contain exactly one transfer and one inscription as
+        // the first op and then may contain other SDP declarations
```
youngjoon-lee:
```suggestion
        // Genesis transactions must contain exactly one transfer as the first op,
        // one inscription as the second op, and then may contain other SDP declarations
```
I think we can improve this comment like this.

core/src/mantle/ledger.rs
```diff
@@ -62,12 +57,6 @@ impl Note {
...
 
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-pub struct Tx {
-    pub inputs: Vec<NoteId>,
-    pub outputs: Vec<Note>,
-}
-
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
 pub struct Utxo {
     pub tx_hash: TxHash,
```
youngjoon-lee:
```suggestion
    pub transfer_hash: TxHash,
```
I think we can now consider renaming this field like this because we are eliminating Ledger **Tx** and this field will be filled with a hash from a Transfer operation. 
We can do this in a next PR after updating the spec once again. Wdyt?

services/wallet/src/lib.rs
youngjoon-lee: I'm reopening this thread to ask a question. Does this comment mean that the `tx_builder` should provide different `input_pks` for each transfer?

thomaslavaur: yes but I assumed that because the builder is in the wallet it's used only to make single user transaction containing a single transfer

thomaslavaur: We can remove this comment if it's enough

core/src/block/mod.rs
thomaslavaur: Yes I think this was wrong in master too... I'm updating it to use MantleTx instead but on master it uses LedgerTx which doesn't make sense here :/

PR #2431, opened 2026-03-26
perf(blend): actually pre-compute and buffer PoQ proofs

ntn-x2:
The old way of using a `buffered` stream was not really doing what it was expected to do. Since streams are lazy by default, the buffer is actually not filled in the background unless the stream is polled, which meant that for each new message, the first message, and very often also the second one, would still take full time before being returned to the requester. This is particularly visible in slow CI, where the old solution was very often taking more than a second to encapsulate a message, while now, because we do most of the generation ahead of time, it takes always `0 ms`, except for the very first data message for which a proof is requested right after setting the secret PoL info, which then kicks off the background generation of the proofs. But, after the first data message is encapsulated, successive ones are all encapsulated in `0 ms` because all proofs are ready by then.

Build on top of the slot tick fix PR: https://github.com/logos-blockchain/logos-blockchain/pull/2427. Together, they make CI tests much less flaky since we are able to actually exploit proof pre-generation, reducing the overall load on the system and concentrating it around session and epoch changes only, which in our current tests only happens once when the nodes are started up.

### DIFF
--- a/blend/scheduling/src/message_blend/provers/core/mod.rs
+++ b/blend/scheduling/src/message_blend/provers/core/mod.rs
@@ -142,11 +142,7 @@ fn spawn_proof_generation(
     let (proof_sender, proof_receiver) = mpsc::channel(buffer_size);
     let handle = tokio::spawn(async move {
         tokio::pin!(stream);
-        loop {
-            let proof = stream
-                .next()
-                .await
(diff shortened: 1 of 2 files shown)


danielSanchezQ (approved):
As per our convo. Next step is to keep the stream interface but using the channel underneath.

PR #2433, opened 2026-03-26
feat(ledger): channel deposit operation

youngjoon-lee:

Implemented `ChannelDeposit` operation, as per [spec](https://www.notion.so/nomos-tech/v1-2-1-Mantle-Specification-31e261aa09df8005988deef29a1286b4?source=copy_link#31e261aa09df81cc8d19f837db432029).

Next PRs (for withdraw, and so on) are coming soon.

Yes

@strinnityk @youngjoon-lee 
spec: @thomaslavaur 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/ledger/src/mantle/channel.rs
+++ b/ledger/src/mantle/channel.rs
@@ -30,6 +30,8 @@ pub enum Error {
     EmptyKeys { channel_id: ChannelId },
     #[error("Channel {channel_id:?} not found")]
     ChannelNotFound { channel_id: ChannelId },
+    #[error("Balance overflow")]
+    BalanceOverflow,
 }
 
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
@@ -136,7 +138,10 @@ impl Channels {
 
     pub fn deposit(mut self, op: &DepositOp) -> Result<Self, Error> {
         if let Some(channel) = self.channels.get_mut(&op.channel_id) {
-            channel.balance += op.amount;
+            channel.balance = channel
+                .balance
+                .checked_add(op.amount)
+                .ok_or(Error::BalanceOverflow)?;
             Ok(self)
         } else {
             Err(Error::ChannelNotFound {


ledger/src/mantle/mod.rs
```diff
@@ -181,6 +181,11 @@ impl LedgerState {
                     self.channels = self.channels.set_keys(op.channel, op, sig, &tx_hash)
                         .inspect_err(|err| error!(target: LOG_TARGET, %err, "failed to apply channel set-keys message"))?;
                 }
+                (Op::ChannelDeposit(op), Some(OpProof::NoProof)) => {
+                    self.channels = self.channels.deposit(op)
+                        .inspect_err(|err| error!(target: LOG_TARGET, %err, "Failed to apply the Channel Deposit message."))?;
+                    balance -= Balance::from(op.amount);
```
thomaslavaur: should we use the checked_sub() method like for the others?

youngjoon-lee: Good point. Fixed e92f9a1a7766d4cd24f685b2349dd2cc4a0d5fda which introduces a `Error::BalanceOverflow`. 

I didn't write a test because `Balance` is too big (i128) and `DepositOp::amount` is just u64 (`Value`). We need too many deposits to simulate a balance overflow. Instead, I added a basic test which we should have done before.

ledger/src/lib.rs
```diff
@@ -413,6 +415,18 @@ impl LedgerState {
...
+                (Op::ChannelDeposit(op), Some(OpProof::NoProof)) => {
+                    let deposit_balance;
+                    (self.mantle_ledger, deposit_balance) =
+                        self.mantle_ledger.try_apply_channel_deposit(op)?;
+                    assert!(
+                        deposit_balance <= 0,
+                        "deposit balance should be non-positive: {deposit_balance}"
+                    );
+                    balance = balance
+                        .checked_add(deposit_balance)
```
thomaslavaur: I think this should be a checked_sub() no?

thomaslavaur: it's decreasing the balance of the transaction and increading the one of the channel

youngjoon-lee: Yeah, I understand that this code is confusing.

The `try_apply_channel_deposit` returns a `Balance` which is `i128` (signed), following the convention of other functions, such as `try_apply_transfer`. It means that the function returns a negative value of Balance, and that's why we do `assert` and `checked_add` here. But, yes, it's ugly.

It would be better to make it return `Value` which is `u64` and do `checked_sub`. But then, it will follow a different convention with `try_apply_transfer` (which has to return `i128`). But I think it is not bad. Let me try some refactoring and ping you.

youngjoon-lee: 5e3fbf7c I refactored the code. Now `try_apply_channel_deposit` returns `Value` (`u64`) and we do `checked_sub`. The `try_apply_transfer` keeps returning `Balance` (`i128`) which is fine.

core/src/mantle/gas.rs
```diff
@@ -40,6 +43,7 @@ impl GasConstants for MainnetGasConstants {
     const TRANSFER: Gas = 2705;
     const CHANNEL_INSCRIBE: Gas = 22;
     const CHANNEL_SET_KEYS: Gas = 22;
+    const CHANNEL_DEPOSIT: Gas = 0;
```
danielSanchezQ, line 46: Is this correct? There is no gas fee for it?

ledger/src/mantle/channel.rs
```diff
@@ -121,13 +126,25 @@ impl Channels {
...
                 },
             );
         }
 
         Ok(self)
     }
 
+    pub fn deposit(mut self, op: &DepositOp) -> Result<Self, Error> {
+        if let Some(channel) = self.channels.get_mut(&op.channel_id) {
+            channel.balance += op.amount;
```
danielSanchezQ: The should be a checked add, it may overflow in production (?)

danielSanchezQ (approved):
A couple questions. After them it is good to go, thanks!

core/src/mantle/gas.rs
youngjoon-lee: Yes. Gas cost is 0 for deposit op.
https://www.notion.so/nomos-tech/v1-3-Mantle-Specification-330261aa09df80a899a6efd74f12a7c4?source=copy_link#330261aa09df81308acfc11b13716fab

ledger/src/mantle/channel.rs
youngjoon-lee: [f6407c6](https://github.com/logos-blockchain/logos-blockchain/pull/2433/commits/f6407c6343cad2c01e3cbf2ad4511710db2c0fb0) done

PR #2435, opened 2026-03-27
feat: clamp ntp jitter

hansieodendaal:

In recent e-2-e system-level tests it was observed that the current slot moved backwards. This can result in unwanted system-level behaviour.

- Clamped NTP jitter by ensuring that current slot, epoch and the slot timer cannot move backwards, ensuring monotonicity.
- Added unit test to confirm `NtpStream` works as expected when injecting it with jitter time stamps.
- Added unit tests to confirm that the `NtpTimeBackend` + `NtpStream` combination works as expected when polling a real world NTP server so fast that we are guaranteed to pick up time-stamp jitter, in the `ms` range.

Yes

@hansieodendaal 

n/a

No.

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/services/time/src/backends/ntp/mod.rs
+++ b/services/time/src/backends/ntp/mod.rs
@@ -143,21 +143,21 @@ impl NtpStream {
         let roundtrip = Duration::from_micros(timestamp.roundtrip());
         let ts_nanos_u128 = (seconds + nanos_fraction + roundtrip / 2).as_nanos();
 
-        let ts_nanos = match u64::try_from(ts_nanos_u128) {
+        let ts_nanos_i128 = match i128::try_from(ts_nanos_u128) {
             Ok(ts_nanos) => ts_nanos,
             Err(e) => {
                 tracing::warn!(
                     "Skipping invalid NTP timestamp {ts_nanos_u128} vs. {}: {e}",
-                    u64::MAX
+                    i128::MAX
                 );
                 return;
             }
         };
 
-        let date = match OffsetDateTime::from_unix_timestamp_nanos(i128::from(ts_nanos)) {
+        let date = match OffsetDateTime::from_unix_timestamp_nanos(ts_nanos_i128) {
             Ok(date) => date,
             Err(e) => {
-                tracing::warn!("Skipping invalid NTP timestamp: {e:?} (ts_nanos={ts_nanos})");
+                tracing::warn!("Skipping invalid NTP timestamp: {e:?} (ts_nanos={ts_nanos_i128})");
                 return;
             }
         };


services/time/src/backends/ntp/mod.rs
```diff
@@ -125,22 +129,412 @@ impl Stream for NtpStream {
...
-            let epoch_config = self.epoch_config;
-            let base_period_length = self.base_period_length;
-            (_, self.slot_timer) = slot_timer(
-                self.slot_config,
-                date,
-                current_slot,
-                epoch_config,
-                base_period_length,
-            );
+            let ts_nanos = (seconds + nanos_fraction + roundtrip / 2).as_nanos() as i128;
```
danielSanchezQ: Did you check about the roundtrip in the NTP specification? I remember there was something already there on how to actually compute this. I may be wrong tho.

services/time/src/backends/ntp/mod.rs
```diff
@@ -125,22 +129,412 @@ impl Stream for NtpStream {
...
+                // Clamp slot to never go backwards
+                if tick.slot < self.last_emitted_slot {
+                    tick.slot = self.last_emitted_slot;
+                } else {
+                    self.last_emitted_slot = tick.slot;
+                }
+                Poll::Ready(Some(tick))
+            }
+            other => other,
+        }
```
danielSanchezQ: This could be a method

services/time/src/backends/ntp/mod.rs
ntn-x2: We know this can never be negative, right? But for this to fit into an `i128`, it would have to fit into a `u64`. Also, additions can lead to overflow in release mode, so we should probably handle it better and log an error in case of overflow. Then after making sure it fits in a `u64`, we can convert it to a `i128`.

services/time/src/backends/ntp/mod.rs
```diff
@@ -125,22 +129,412 @@ impl Stream for NtpStream {
...
+                    } else {
+                        tracing::warn!(
+                            "NTP resync moved backwards: computed_slot={:?}, last_emitted_slot={:?}; clamping",
+                            current_slot,
+                            self.last_emitted_slot
+                        );
+                    }
+                }
+                Err(e) => {
+                    tracing::warn!("Skipping invalid NTP timestamp: {e:?} (ts_nanos={ts_nanos})");
```
ntn-x2: But we still don't return early, right? We should probably return here, in which case I would suggest to use a `let Ok(date)` and avoid one big indentation.

services/time/src/backends/ntp/mod.rs
```diff
@@ -125,22 +129,412 @@ impl Stream for NtpStream {
...
+                        let base_period_length = self.base_period_length;
+                        (_, self.slot_timer) = slot_timer(
+                            self.slot_config,
+                            date,
+                            current_slot,
+                            epoch_config,
+                            base_period_length,
+                        );
+                        self.last_emitted_slot = current_slot;
+                    } else {
```
ntn-x2: Also here don't we want to early return? In any case, probably specifying the error case beforehand and early returning is more readable than having a big indentation for the valid case. Just a stylistic choice, I agree, so these are my 2 cents.

services/time/src/backends/ntp/mod.rs
```diff
@@ -125,22 +129,412 @@ impl Stream for NtpStream {
...
+                }
+            }
         }
         // poll from internal last updated `SlotTick` stream
-        self.slot_timer.poll_next_unpin(cx)
+        match self.slot_timer.poll_next_unpin(cx) {
+            Poll::Ready(Some(mut tick)) => {
+                // Clamp slot to never go backwards
+                if tick.slot < self.last_emitted_slot {
+                    tick.slot = self.last_emitted_slot;
```
ntn-x2: But doesn't this mean that we would yield the same slot twice?

services/time/src/backends/ntp/mod.rs
danielSanchezQ: > We know this can never be negative, right? But for this to fit into an `i128`, it would have to fit into a `u64`. Also, additions can lead to overflow in release mode, so we should probably handle it better and log an error in case of overflow. Then after making sure it fits in a `u64`, we can convert it to a `i128`.

:100: nice catch indeed.

services/time/src/backends/ntp/mod.rs
danielSanchezQ: wasn't that the case already? we could tick for the same slot? :thinking:

ntn-x2: But that should not happen, right? We should just return `Poll::Pending` in that case, in my opinion.

danielSanchezQ: > But that should not happen, right? We should just return `Poll::Pending` in that case, in my opinion.

Yeah, it would be better to be consistent in that.

services/time/src/backends/ntp/mod.rs
hansieodendaal: We should not return here as the timer must be sanitized as well below.

services/time/src/backends/ntp/mod.rs
hansieodendaal: This does not emit the same slot twice, without this guard it does. Tested in `test_real_ntp_server_fast_poll() `.
But not a bad idea to return `Poll::Pending` if we regressed.

services/time/src/backends/ntp/mod.rs
hansieodendaal: Code flow improved in next commit

services/time/src/backends/ntp/mod.rs
hansieodendaal: > Did you check about the roundtrip in the NTP specification? I remember there was something already there on how to actually compute this. I may be wrong tho.

I could not find anything in our specs, and reading the web people suggest dividing roundtrip by 2 is a better estimate for NTP-style time sync. The usual assumption is that network delay is roughly symmetric, with the remote timestamp about half an RTT old by the time we process it. Adding `roundtrip / 2` is the standard “one-way delay” approximation.

services/time/src/backends/ntp/mod.rs
```diff
@@ -113,34 +112,458 @@ pub struct NtpStream {
...
+
+        let Poll::Ready(Some(timestamp)) = this.interval.as_mut().poll_next_unpin(cx) else {
+            return;
+        };
+
+        let seconds = Duration::from_secs(timestamp.sec().into());
+        let nanos_fraction =
+            Duration::from_nanos(fraction_to_nanoseconds(timestamp.sec_fraction()).into());
+        let roundtrip = Duration::from_micros(timestamp.roundtrip());
+        let ts_nanos_u128 = (seconds + nanos_fraction + roundtrip / 2).as_nanos();
```
ntn-x2, line 144: This can still overflow.

services/time/src/backends/ntp/mod.rs
```diff
@@ -113,34 +112,458 @@ pub struct NtpStream {
...
+
+        let ts_nanos = match u64::try_from(ts_nanos_u128) {
```
ntn-x2: This does not make much sense. Why would you try to convert to `u128` first and then to `u64`? Can't we convert to `u64` directly, and then from there into `i128`?

services/time/src/backends/ntp/mod.rs
hansieodendaal: No, 
- `let seconds = Duration::from_secs(timestamp.sec().into());` 
    => `Duration::from_secs(u32.into());` 
    => Maximum whole seconds is `u32::MAX`
- `let nanos_fraction = Duration::from_nanos(fraction_to_nanoseconds(timestamp.sec_fraction()).into());`
    => `Duration::from_nanos(u32.into());` 
    => Maximum whole seconds is < 1
- `let roundtrip = Duration::from_micros(timestamp.roundtrip());`
   => `Duration::from_micros(u64);` 
   => Maximum whole seconds is < `u64::MAX / 1_000_000`
- `(seconds + nanos_fraction + roundtrip / 2)`  gives a valid `Duration`
   => This can never overflow, maximum value is `< u32::MAX + u64::MAX / 2_000_000 + 1`
- `let ts_nanos_u128 = (seconds + nanos_fraction + roundtrip / 2).as_nanos();`   
   => This can never overflow for a valid `Duration`

PR #2439, opened 2026-03-27
feat: improve bincode serialization

hansieodendaal:

- Improved `bincode` serializaion by _simplifying the code and using the optimized implementation bincode already offers_ ~~removing an additional serialization step~~ and being consistent with using the custom `OPTIONS` for consistent configuration. ~~The improved approach is especially applicable when using `bincode` and `BytesMut` because it leverages their strengths: efficient, single-pass serialization into a buffer that can grow as needed, without unnecessary memory allocations or data traversals. This results in better performance and lower memory fragmentation.~~
- Many clippy fixes due to updated version.

Yes

@hansieodendaal 

n/a

No.

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/blend/network/src/core/with_core/behaviour/handler/mod.rs
+++ b/blend/network/src/core/with_core/behaviour/handler/mod.rs
@@ -157,10 +157,6 @@ where
 
     #[expect(deprecated, reason = "Self::OutboundOpenInfo is deprecated")]
     #[expect(clippy::too_many_lines, reason = "TODO: Address this at some point.")]
-    #[expect(
-        clippy::cognitive_complexity,
-        reason = "Singular fn with multiple branches to handle different events and futures."
-    )]
     fn poll(
         &mut self,
         cx: &mut Context<'_>,
@@ -334,10 +330,6 @@ where
     #[expect(
         deprecated,
         reason = "Self::InboundOpenInfo and Self::OutboundOpenInfo are deprecated"
-    )]
-    #[expect(
-        clippy::cognitive_complexity,
-        reason = "Singular fn with multiple branches to handle different events and futures."
     )]
     fn on_connection_event(
         &mut self,
--- a/blend/network/src/core/with_core/behaviour/mod.rs
+++ b/blend/network/src/core/with_core/behaviour/mod.rs
@@ -959,10 +959,6 @@ where
             .map_err(|_| Error::InvalidMessage)
     }
 
-    #[expect(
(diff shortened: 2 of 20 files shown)


core/src/codec/bincode/mod.rs
```diff
@@ -34,16 +34,19 @@ use serde::{Serialize, de::DeserializeOwned};
...
 use crate::codec::{Error as WireError, Result};
 
+const DEFAULT_SERIALIZATION_CAPACITY: usize = 16 * 1024; // 16 KB
+
 /// Serialize an object directly into bytes
 pub fn serialize<T: Serialize>(item: &T) -> Result<Bytes> {
-    let size = OPTIONS
-        .serialized_size(item)
-        .map_err(|e| WireError::Serialize(Box::new(e)))?;
-
```
youngjoon-lee, line 42: I just check how `serialized_size` is implemented in `bincode`. It seems to just walk through the data structure and compute how many bytes would produce, without allocating any output buffers. If so, I think we don't need this optimization. Wdyt?

youngjoon-lee: IMHO, we can simplify this function a lot:
```rust
pub fn serialize<T: Serialize>(item: &T) -> Result<Bytes> {
    Ok(OPTIONS
        .serialize(item)
        .map_err(|e| WireError::Serialize(Box::new(e)))?
        .into())
}
```

Actually, `OPTIONS.serialize` uses `serialized_size` internally. It traverses the data structure but I don't think it is a big burden. Also, we don't actually need to use `BytesMut`. `OPTIONS.serialize` allocates/returns a `Vec<u8>`. Then, we can convert it into `Bytes` without additional allocations. IMO, this function was initially implemented too complicatedly.

davidrusu (approved):
nice, if @youngjoon-lee is right, this saves us one traversal through the datastructure, and it's simper :)

youngjoon-lee:
> nice, if @youngjoon-lee is right, this saves us one traversal through the datastructure, and it's simper :)

IMO, it's safer to keep the data traversal for now, by using the simplified code I suggested above: https://github.com/logos-blockchain/logos-blockchain/pull/2439#discussion_r3015995169. I don't think it's a good idea to **always** allocate a 16KB buffer for any size of data, which this PR initially suggested. Later when we see any CPU-bounded issues, we can optimize this somehow. For now, I think it's better to calculate the size and allocate `Vec` with the size.

PR #2442, opened 2026-03-29
fix: mitigate panic in blend epoch info when e2e tests end

hansieodendaal:

This is a proposal to remove a hidden panic occurring in `epoch_info.rs`, in `pub async fn tick(&mut self, new_tick: SlotTick) -> Option<EpochEvent> `. The gist of this seems to be in line with the purpose of the function, but please evaluate.

When `pub async fn get_epoch_state(` cannot return the epoch state, an error is logged and `None` is returned, instead of panicking. If a panic is still desirable, it should be handled at the initial call site.

This panic has been observed when running multiple nextest e2e instances in parallel:
```rust
2026-03-26T15:37:53.459994Z ERROR logos_blockchain_chain_service: Could not notify new block to services 
channel closed
2026-03-26T15:37:55.978964Z ERROR logos_blockchain_node::panic: A panic occurred panic.payload="State for 
slot in current epoch should always be available: Ledger(InvalidSlot { parent: Slot(120), block: Slot(119) })" 
panic.location="services/blend/src/epoch_info.rs:69:14" panic.backtrace="disabled backtrace" panic.note="run 
with RUST_BACKTRACE=1 environment variable to display a backtrace"
```

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/services/blend/src/epoch_info.rs
+++ b/services/blend/src/epoch_info.rs
@@ -8,10 +8,7 @@ use core::{
 use async_trait::async_trait;
 use futures::Stream;
 use lb_blend::proofs::quota::inputs::prove::private::ProofOfLeadershipQuotaInputs;
-use lb_chain_service::{
-    Error as ChainServiceError,
-    api::{ApiError, CryptarchiaServiceApi, CryptarchiaServiceData},
-};
+use lb_chain_service::api::{CryptarchiaServiceApi, CryptarchiaServiceData};
 use lb_core::{crypto::ZkHash, proofs::leader_proof::LeaderPublic};
 use lb_cryptarchia_engine::{Epoch, Slot};
 use lb_groth16::Fr;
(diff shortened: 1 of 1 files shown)


services/blend/src/epoch_info.rs
```diff
@@ -48,11 +51,19 @@ pub trait PolInfoProvider<RuntimeServiceId> {
...
+    #[error("Chain service error: {0}")]
+    ChainServiceError(#[from] ChainServiceError),
+}
+
 /// A trait that provides the needed functionalities for the epoch stream to
 /// fetch the epoch state for a given slot.
 #[async_trait]
 pub trait ChainApi<RuntimeServiceId> {
-    async fn get_epoch_state_for_slot(&self, slot: Slot) -> EpochState;
+    async fn get_epoch_state_for_slot(&self, slot: Slot) -> Result<EpochState, EpochError>;
```
ntn-x2: I think it's just simpler to return `Option<EpochState>` instead, and handle the `None` at the higher level, as you suggested. We can still log the cause of the error here tho.

services/blend/src/epoch_info.rs
```diff
@@ -48,11 +51,19 @@ pub trait PolInfoProvider<RuntimeServiceId> {
 
 const LOG_TARGET: &str = "blend::service::epoch";
 
+#[derive(Debug, thiserror::Error)]
+pub enum EpochError {
```
ntn-x2: I guess we don't need this enum anymore, so let's get rid of it 🪓

PR #2445, opened 2026-03-30
chore(zone-sdk): zone sdk next_event API + prune finalized tx from rdbs

pradovic:

Exposes a `next_event` API that enables the user to drive the loop  and integrate with their existing async env. 

Also fixes pruning on finalizes states and rdbs so it does not grow forever

Yes

@pradovic 
Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/zone-sdk/src/sequencer.rs
+++ b/zone-sdk/src/sequencer.rs
@@ -426,12 +426,17 @@ impl ZoneSequencer {
                     .await;
 
                     // Update channel tip from backfill/block inscriptions.
-                    // Only when starting from root — checkpoint resume already
-                    // has correct last_msg_id which may be ahead of finalized.
-                    if let Some(tip) = result.channel_tip
-                        && self.last_msg_id == MsgId::root()
-                    {
-                        self.last_msg_id = tip;
+                    // Only when no pending inscriptions remain — if there are
+                    // pending txs, the checkpoint's last_msg_id may be ahead
+                    // of backfill (inscriptions above LIB, not yet finalized).
+                    if let Some(tip) = result.channel_tip {
+                        let has_pending = self
+                            .state
+                            .as_ref()
+                            .is_some_and(|s| s.unfinalized_count() > 0);
+                        if !has_pending {
+                            self.last_msg_id = tip;
+                        }
                     }
 
                     // Signal readiness after first block event processed


zone-sdk/src/state.rs
davidrusu: If we are missing the parent, I feel like we should be panicking here now rather than defaulting to the empty set.

We can take as an invariant the the parent must be processed before the child to keep our logic simple

pradovic: I agree! There are lot's of places like this as it is a draft. Let's review if general direction is good, you can check the e2e tests, and if it is good, I can polish the PR. Feel free to leave the comments though, appreciate them, I will address them all, thanks!

zone-sdk/src/state.rs
```diff
@@ -110,9 +96,25 @@ impl TxState {
...
+            // persistent structure.
+            if let Some(lib_safe_set) = self.block_states.get(&lib) {
+                let mut fresh = HashTrieSetSync::new_sync();
+                for hash in lib_safe_set.iter() {
+                    if self.pending.contains_key(hash) {
+                        fresh = fresh.insert(*hash);
+                    }
+                }
+                self.block_states.insert(lib, fresh);
+            }
```
davidrusu, line 105: Isn't this `lib_safe_set` the empty set? we removed all the transactions in `lib` from the pending set on line https://github.com/logos-blockchain/logos-blockchain/pull/2445/changes#diff-f805e7357053341a4d571eb932b9635687459426ed1bbc927b57b9c775208905L93

So this set should be the empty set no?

zone-sdk/src/state.rs
```diff
@@ -52,14 +36,14 @@ impl TxState {
         self.pending.insert(tx_hash, signed_tx);
     }
```
davidrusu, line 38: I think we should just take `signed_tx` as input, we can derive the hash with `signed_tx.mantle_tx.hash()`. (removes chances of providing wrong hash for some tx)

zone-sdk/src/state.rs
```diff
@@ -252,8 +228,11 @@ mod tests {
         // Process block containing our tx, lib stays at genesis
         state.process_block(b1, genesis, genesis, vec![hash]);
 
-        assert_eq!(state.status(&hash, b1), TxStatus::Safe);
-        assert_eq!(state.status(&hash, genesis), TxStatus::Pending);
+        // Tx is still pending (not finalized yet, lib hasn't advanced)
+        assert_eq!(state.pending_count(), 1);
+
+        // But pending_txs at b1 excludes it (it's in the safe set)
+        assert!(state.pending_txs(b1).next().is_none());
```
davidrusu, line 228: this is confusing, how about changing the `pending_count()` to `unfinalized_count()`

zone-sdk/src/state.rs
davidrusu: So this on it's own does not remove the state growth, we would have to iterate through all the block states and remove any finalized txs from all of the block states.

This is only purging the LIB block states, but the tip block states (which cloned from an earlier version of this LIB state that included the now finalized txs) still contains the finalized txs

zone-sdk/src/sequencer.rs
```diff
@@ -246,14 +177,19 @@ impl ZoneSequencer {
...
         let signed_tx = reply_rx.await.map_err(|_| Error::Unavailable {
-            reason: "actor dropped reply",
+            reason: "sequencer dropped reply",
         })??;
 
         let tx_hash = signed_tx.mantle_tx.hash();
+
+        // Subscribe to events BEFORE posting to avoid a race where the tx
+        // finalizes between posting and subscribing.
+        let mut event_rx = self.event_tx.subscribe();
```
davidrusu: By this logic, shouldn't we subscribe even before the `self.request_tx.send(request)` above?

zone-sdk/src/sequencer.rs
```diff
@@ -193,50 +149,25 @@ impl ZoneSequencer {
...
+    /// waits for finalization:
+    ///
+    /// ```ignore
+    /// let finalized = handle.set_keys(vec![admin_pk]).await?;
+    /// finalized.await?; // wait for finalization
+    /// ```
+    pub async fn set_keys(
+        &self,
+        keys: Vec<Ed25519PublicKey>,
+    ) -> Result<impl Future<Output = Result<(), Error>>, Error> {
```
davidrusu: Can we unify this`impl Future<Output = Result<(), Error>>` with the `PublishResult` somehow? We need checkpoints after setting keys as well so that pending set key transactions can be restored from checkpoint as well.

davidrusu (commented):
The approach splitting the handle from the event loop is nice, makes it easy to pass the handle around cheaply to where we need it in the sequencer.

The approach looks good!

zone-sdk/src/sequencer.rs
```diff
@@ -265,235 +200,333 @@ impl ZoneSequencer {
...
+            checkpoint,
+        );
+        tokio::spawn(async move {
+            loop {
+                sequencer.next_event().await;
             }
-        }
+        });
+        handle
     }
```
davidrusu: can we decouple init from spawn? that way you can init with either constructor and spawn later

```rust
let (sequencer, handle) = ZoneSequencer::{init|init_with_config}(...);

sequencer.spawn(); // consumes self and spawns the event loop

// use handle later
```

zone-sdk/src/sequencer.rs
```diff
@@ -584,6 +620,21 @@ fn handle_inflight(event: InFlight, resubmit_active: &mut bool) {
...
+        Ok(Some(block)) => block.header().slot(),
+        Ok(None) => {
+            // Genesis case - slot 0
+            Slot::genesis()
+        }
+        Err(e) => {
+            warn!("Failed to get lib block slot: {e}, assuming slot 0");
+            Slot::genesis()
+        }
+    }
```
davidrusu: Lets return error here rather than assuming 0

zone-sdk/src/state.rs
```diff
@@ -110,9 +94,26 @@ impl TxState {
...
+            // structural sharing, causing unbounded memory growth.
+            for safe_set in self.block_states.values_mut() {
+                let mut fresh = HashTrieSetSync::new_sync();
+                for hash in safe_set.iter() {
+                    if self.pending.contains_key(hash) {
+                        fresh = fresh.insert(*hash);
+                    }
+                }
+                *safe_set = fresh;
+            }
```
davidrusu, line 105: Rather than rebuilding the set each time, you can use the the `newly_finalized` set you built above:
```
for each block_state safe_set
    for each newly_finalized finalized_tx
        *safe_set = safe_set.remove(finalized_tx)
```

This allows memory sharing between current safe blocks

zone-sdk/src/sequencer.rs
pradovic: good catch!

zone-sdk/src/state.rs
pradovic: Nice, you are right!

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -97,6 +97,13 @@ async fn test_sequencer_publish_and_indexer_read() {
         None, // Fresh start, no checkpoint
     );
 
+    // Spawn a task to drive the sequencer
+    let poll_task = tokio::spawn(async move {
+        loop {
+            sequencer.next_event().await;
+        }
+    });
+
```
davidrusu: can we use the `sequencer.spawn()` api here?

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -263,11 +257,18 @@ async fn test_sequencer_checkpoint_resume() {
         None, // Fresh start
     );
 
+    // Spawn polling task
+    let poll_task = tokio::spawn(async move {
+        loop {
+            sequencer.next_event().await;
+        }
+    });
+
```
davidrusu: Same here, `sequencer.spawn()`?

zone-sdk/src/sequencer.rs
```diff
@@ -265,235 +200,322 @@ impl ZoneSequencer {
...
+    /// handle.publish(b"hello".to_vec()).await?;
+    /// ```
+    pub fn spawn(mut self) {
+        tokio::spawn(async move {
+            loop {
+                self.next_event().await;
             }
-        }
+        });
     }
```
davidrusu: I noticed in your tests, you maintained a reference to the poll task, do you want to return the result of `tokio::spawn(..)` here in case you the caller wants to abort early?

pradovic: Yes, I have modified the call to return the handle of spawn

pradovic:
Hey @hansieodendaal, thanks for the review — really helpful points.

To address both:

1. Driving loop
I think exposing the event-driven handle is quite useful when integrating the sequencer into an already async environment. It allows the user to drive the system instead of the SDK spawning its own task. This seems like a fairly common pattern in async systems, but happy to revisit if it turns out to be inconvenient in practice. There is also a convenience `spawn()` route so this is only for advanced users.

2. Checkpoint resume scenario
I thought through your scenario, and I believe the current behavior handles it correctly:
	• On resume, the sequencer backfills from the checkpoint’s LIB to the current network LIB.
	• Any pending txs from the checkpoint that already landed on-chain get finalized during backfill (and removed from pending).
	• The sequencer then discovers the current channel tip and continues from there.

So whether we start fresh or resume from a checkpoint with the same key, the sequencer reconciles with the current chain state during startup. The only difference is that checkpoint resume restores pending txs, but backfill cleans up the ones that already landed.

In the competing-sequencers PR (#2446), this is handled more explicitly as well — detect_channel_update fires after startup and invalidates any pending inscriptions whose parents were consumed while the sequencer was away. The user is notified and can decide whether to re-publish.

For the edge case of stale pending txs (e.g. parent already consumed while away):
	• In this PR, they would be harmlessly resubmitted and rejected by the node.
	• In #2446, they get properly invalidated.

Either way, we shouldn’t end up in an incorrect state.

Does this align with what you had in mind? My understanding is that this behavior (if problematic) would not be introduced by this PR, but rather already exists in the current checkpoint/resume logic.  Would it make sense to tackle this in a separate PR?

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -368,3 +320,218 @@ async fn test_sequencer_checkpoint_resume() {
...
+            .expect("indexer error");
+        for msg in &result.messages {
+            if expected.contains(&msg.data) {
+                seen.insert(msg.data.clone());
+            }
+        }
+        cursor = Some(result.cursor);
+        if seen == expected {
+            break;
+        }
```
hansieodendaal, line 415: You may want to sort before checking

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -368,3 +320,218 @@ async fn test_sequencer_checkpoint_resume() {
...
+            .expect("indexer error");
+        for msg in &result.messages {
+            if expected_all.contains(&msg.data) {
+                seen.insert(msg.data.clone());
+            }
+        }
+        cursor = Some(result.cursor);
+        if seen == expected_all {
+            break;
+        }
```
hansieodendaal, line 460: ditto sort - to have the test stable

zone-sdk/src/sequencer.rs
```diff
@@ -447,16 +425,27 @@ impl ZoneSequencer {
                     )
                     .await;
 
-                    if newly_finalized.is_empty() {
+                    // Update channel tip from backfill/block inscriptions
+                    if let Some(tip) = result.channel_tip {
+                                                self.last_msg_id = tip;
+                    }
```
hansieodendaal: formatting?

zone-sdk/src/sequencer.rs
```diff
@@ -662,7 +735,17 @@ async fn backfill_to_lib(
...
+                for tx in &block.transactions {
+                    for op in &tx.mantle_tx.ops {
+                        if let Op::ChannelInscribe(inscribe) = op
+                            && inscribe.channel_id == channel_id
+                        {
+                            latest_msg_id = Some(inscribe.id());
+                        }
+                    }
+                }
+
```
hansieodendaal: We could have many inscriptions in the block, and we are not guaranteed that the node will provide it in our expected order.

Here we have to check the parent-child ordering - get all transactions then sort them, `parent_id` not referenced as `this_id` is the 1st transaction, `this_id` not referenced as `parent_id` is the last transaction.

zone-sdk/src/sequencer.rs
```diff
@@ -591,12 +650,23 @@ async fn handle_block_event(
...
-    // advanced)
+    for tx in &event.block.transactions {
+        for op in &tx.mantle_tx.ops {
+            if let Op::ChannelInscribe(inscribe) = op
+                && inscribe.channel_id == channel_id
+            {
+                channel_tip = Some(inscribe.id());
+            }
+        }
+    }
```
hansieodendaal: ditto transactions are not guaranteed to be ordered

hansieodendaal (commented):
Loos good, I like the new proving test case. Just a couple of comments to consider for a next PR.

tests/src/tests/zone_sdk/e2e.rs
pradovic: seen == expected is already order-independent I think

tests/src/tests/zone_sdk/e2e.rs
pradovic: seen == expected is already order-independent I think

zone-sdk/src/sequencer.rs
davidrusu: hold on, this is not true, the transactions in the block are guaranteed to be in channel order.

They will never land in the block out of order and the node block api is guaranteed to return the block transactions in order.

cc @pradovic

zone-sdk/src/sequencer.rs
davidrusu: again, not true, channel ops are guaranteed to be in order in a block, block validation would fail if the channel ops are not in order

PR #2446, opened 2026-03-30
feat(zone-sdk): competing sequencers 

pradovic:

The slightly modified implementation of: https://www.notion.so/nomos-tech/Decentralized-Sequencing-Behaviour-32e261aa09df804c9017c3e61b4a83c4?source=copy_link.

The goal is to allow users to spawn multiple sequencers on the same channel and resolve conflicts when they arise. The API is intentionally flexible and offloads most decisions to the user.

Conflicts are bounded by block production and finalization latency, so they resolve relatively slowly. For high-throughput or high-collision applications, the preferred approach is round-robin sequencing, which is being developed separately.

The TUI and e2e test changes were reviewed in separate PRs.

Core API change

`Event::ChannelUpdate` now carries `{ orphaned, adopted, pending, invalidated }`. See the event doc comment for the expected consumer pattern.

TODO (follow-up PRs)

* Refactor sequencer and state
* Reassess tests once testing on master stabilizes
* Potentially move some tests to a single-node setup
* Improve return types and their relationships; provide a more user-friendly API with a clearer lifecycle (inscriptions, user tx, node tx, etc.)

Yes

@pradovic 
spec: @davidrusu 

Yes

MessageUpdated event now has additional fields {orphaned, adopted, invalidated, pending } to various all the use-cases and offer flexibility to the user

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -97,39 +113,36 @@ async fn test_sequencer_publish_and_indexer_read() {
...
+            }) = sequencer.next_event().await
+            {
+                let adopted_payloads: HashSet<Vec<u8>> =
+                    adopted.into_iter().map(|a| a.payload).collect();
+                for inv in invalidated {
+                    if !adopted_payloads.contains(&inv.payload) {
+                        drop(reorg_handle.publish(inv.payload).await);
+                    }
+                }
+            }
```
hansieodendaal: **Observation**
For these driver loops we need to use more information from the channel update event if ordering is important, then the parent-child relationships should be managed.

Just a note, as this is client implementation dependent.

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -382,3 +397,491 @@ async fn test_sequencer_checkpoint_resume() {
...
+                    adopted.into_iter().map(|a| a.payload).collect();
+                for inv in invalidated {
+                    if !adopted_payloads.contains(&inv.payload) {
+                        debug!(
+                            "Re-publishing orphaned: {:?}",
+                            String::from_utf8_lossy(&inv.payload)
+                        );
+                        if let Err(e) = handle.publish(inv.payload).await {
+                            debug!("Failed to re-publish: {e}");
+                        }
```
hansieodendaal: ditto parent-child ordering comment

zone-sdk/src/sequencer.rs
```diff
@@ -304,9 +344,26 @@ impl ZoneSequencer {
...
+                            inscribe.id(),
+                            inscribe.inscription.clone(),
+                        );
+                        is_inscription = true;
+                        break;
+                    }
+                }
+                if !is_inscription {
+                    tx_state.submit_other(hash, tx);
+                }
```
hansieodendaal: In this logic when `init_with_config` is called, one cannot assume that a checkpoint's pending transactions are still pending. These should be matched to what is the blockchain state, and only inscription's truly pending must be re-submitted. The checkpoint could have 1,000 pending, but be 1 year old.

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -382,3 +397,491 @@ async fn test_sequencer_checkpoint_resume() {
...
+            .publish(data.clone())
+            .await
+            .expect("publish should succeed after wait_ready");
+    }
+}
+
+/// Helper: wait for all expected payloads to appear in the indexer.
+async fn wait_for_indexer(
+    indexer: &ZoneIndexer,
+    expected: &HashSet<Vec<u8>>,
```
hansieodendaal, line 358: This should be for msg_id, not the inscription data, as the latter can be duplicated.

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -382,3 +397,491 @@ async fn test_sequencer_checkpoint_resume() {
...
+        None,
+    );
+    let poll_a = spawn_sequencer_poll(sequencer_a, handle_a.clone());
+
+    let phase1_data: Vec<Vec<u8>> = vec![tag_payload("a1"), tag_payload("a2"), tag_payload("a3")];
+    publish_all(&mut handle_a, &phase1_data).await;
+
+    let indexer = ZoneIndexer::new(channel_id, node_url.clone(), None);
+    let expected_phase1: HashSet<Vec<u8>> = phase1_data.iter().cloned().collect();
+    wait_for_indexer(&indexer, &expected_phase1, Duration::from_secs(360)).await;
```
hansieodendaal: I feel POC's #2375 and #2379 method to subscribe and be notified of inclusions is a much better pattern, also using much less resources. Clients subscribe to inscriptions they need to be notified of, the sequencer has all the data, then notifies the client via a one-shot channel when the inscription is mined. No need to fire up an indexer.

There could be two subscriptions - one when mined and one when include in the LIB chain.

hansieodendaal (commented):
This is working nicely already and when running the test cases most of them succeed the 1st time round, suggesting most issues for this design has been discovered and addressed. I like the test cases, by the way, as it demonstrates the distributed nature nicely.

I have three major comments:
- The first pertain to the concept of a client subscribing to inclusion events, as commented below.
- The second is that maybe we should rethink what the sequencer is supposed and/or able to do in comparison to the client. As the sequencer is aware of all the data, a lot of code in the demonstration tests can be moved into the sequencer efficiently. One thing that comes to mind is where the responsibility for the _driver_ loop lies - in the client or in the sequencer?
- Sequencer resume from checkpoint should be resilient to state changes while the sequencer was away. This is pertinent where sequencers would share the same signing key. `SeqA_key123` makes inscriptions and exits, `SeqB_key123` starts up, discover the last `msg_id`, makes inscriptions and exit. Then when `SeqA_key123` starts up again, its saved checkpoint must be sanitized before use.

zone-sdk/src/sequencer.rs
pradovic: Good catch! 
Should be handled now (after merge from master) -  seeding from the checkpoint here is just in-memory state. Before any resubmit, we :
1. Backfill from checkpoint.lib_slot to network LIB, dropping txs that finalized in the gap, 
2. on the first block event treat it as a reorg from finalized_msg and emit ChannelUpdate with invalidated stale inscriptions, and (3) gate the resubmit timer on ready_tx, which only flips after backfill completes. So a 1-year-old checkpoint gets fully reconciled against chain state before anything is re-sent.

tests/src/tests/zone_sdk/e2e.rs
pradovic: Agreed in general, but this is test-only and the tests pick distinct payloads on purpose, so payload-equality is a stronger assertion ("the exact bytes I published landed") than msg_id matching. Happy to switch to msg_id if you'd prefer consistency.

tests/src/tests/zone_sdk/e2e.rs
pradovic: I think the point of this test is to exercise both components — sequencer and indexer — end-to-end. The sequencer already exposes a `TxsFinalized` event stream, so the indexer isn't strictly necessary here, but I wanted to use it as the final source of truth to mimic a real client scenario where the indexer is the read path. 

That being said we can/should expose better API on top of the low-level API, but right now I think everything can be achieved by following events.

Indexer in tests is only to test a different  component (indexer) as a source of truth that is independent of what sequencer is doing. The indexer only checks final inclusion as the source of truth.

tests/src/tests/zone_sdk/e2e.rs
pradovic: Could you elaborate more on this pls? I am not following 🙈

testnet/tui-zone/src/lib.rs
```diff
@@ -83,6 +83,38 @@ pub async fn run(args: InscribeArgs) {
...
+                    ..
+                }) => {
+                    let adopted_payloads: HashSet<Vec<u8>> =
+                        adopted.into_iter().map(|a| a.payload).collect();
+                    for inv in invalidated {
+                        if !adopted_payloads.contains(&inv.payload)
+                            && let Err(e) = reorg_handle.publish(inv.payload).await
+                        {
+                            eprintln!("  Failed to re-publish after reorg: {e}");
+                        }
```
davidrusu: What is this branch trying to prevent?

pradovic: This guards against double-publishing during a reorg. When `ChannelUpdate` fires, some of our pending inscriptions are invalidated (their parent was taken on the new branch), but the reorg may have also adopted a branch that already contains an inscription with the same payload. Does that makes sense?

zone-sdk/src/state.rs
```diff
@@ -2,113 +2,214 @@ use std::collections::{BTreeMap, HashMap};
...
+#[derive(Debug, Clone)]
+pub struct InscriptionInfo {
+    /// The transaction hash containing this inscription.
+    pub tx_hash: TxHash,
+    /// The parent message ID this inscription chains from.
+    pub parent_msg: MsgId,
+    /// The message ID of this inscription.
+    pub this_msg: MsgId,
+    /// The opaque inscription payload.
+    pub payload: Vec<u8>,
```
davidrusu, line 19: any reason not to provide the full `InscriptionOp` here? it could be meaningful to have the signer (e.g. suppose you only want to replay msgs that you sent, not what the other sequencer sent)

pradovic: You are correct, we can do that. 
Let's try to identify how would the user actually use the API. 
For example if we wanted to make tui full supporting competing sequencers and handling duplicates, the easiest way would probably be to prepend each message with a random prefix, and match it exactly when comparing adopted/invalidated. This way it knows:
1. It's their message
2. It's exact message so "A" would be different from different "A" because it would "pref1:A", "pref2:A" etc...

Signer can also help here I agree.

davidrusu: I guess my thought here is that we are limiting the api when it's not necessary. We can replace all three of these fields with just the inscription op (`this_msg` can be derived from the inscription op's `.id()`)

So rather than trying to enumerate all the use-cases and only providing what we think is needed for that, why not just give the user all the information we have and let them pick what they need.

pradovic: Ah, I see now, makes sense!

testnet/tui-zone/src/lib.rs
davidrusu: yes! makes sense thanks, there is potential for lost messages, but that's ok

testnet/tui-zone/src/lib.rs
```diff
@@ -109,7 +141,12 @@ pub async fn run(args: InscribeArgs) {
             break;
         }
 
-        match handle.publish(msg.as_bytes().to_vec()).await {
+        // Tag payload with a random ID so we can identify our inscriptions
+        // during reorg handling (avoids duplicate re-publishing).
+        let id: u64 = rand::random();
+        let tagged_payload = format!("{id:016x}:{msg}");
+
+        match handle.publish(tagged_payload.into_bytes()).await {
```
davidrusu: I think we can just expose the signer, no need for this in this example zone

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -321,6 +347,829 @@ async fn test_sequencer_checkpoint_resume() {
...
+                            debug!(
+                                "Re-publishing orphaned: {:?}",
+                                String::from_utf8_lossy(&inv.payload)
+                            );
+                            let h = handle.clone(); // TODO: remove spawn when publish is fire-and-forget
+                            tokio::spawn(async move {
+                                if let Err(e) = h.publish(inv.payload).await {
+                                    debug!("Failed to re-publish: {e}");
+                                }
+                            });
```
davidrusu: why spawn? it seems we could get some unintended re-ording if tokio schedules the publish tasks out of order. Can we await directly in the loop?

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -321,6 +347,829 @@ async fn test_sequencer_checkpoint_resume() {
...
+                Err(_) => break,
+            }
+        }
+    })
+}
+
+/// Drive the sequencer event loop in the background.
+fn spawn_sequencer_poll(sequencer: ZoneSequencer) -> tokio::task::JoinHandle<()> {
+    sequencer.spawn()
+}
```
davidrusu, line 383: this doesn't seem worthwhile to put in a function, any reason we shouldn't just call `.spawn()` directly?

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -321,6 +347,829 @@ async fn test_sequencer_checkpoint_resume() {
     poll_task.abort();
 }
 
+/// Subscribe to a sequencer's events and re-publish orphaned inscriptions.
+///
+/// Uses `handle.subscribe()` — the intended usage pattern for client apps.
+/// The event loop must be driven separately (e.g. via `spawn_sequencer_poll`
+/// or `ZoneSequencer::spawn`).
+fn spawn_republish_handler(
```
davidrusu: Can we rename this to `spawn_rebase_on_conflict_handler`?

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -321,6 +347,829 @@ async fn test_sequencer_checkpoint_resume() {
...
+    let mut events = handle.subscribe();
+    let handle = handle.clone();
+    tokio::spawn(async move {
+        loop {
+            match events.recv().await {
+                Ok(Event::ChannelUpdate {
+                    invalidated,
+                    adopted,
+                    ..
+                }) => {
```
davidrusu, line 358: what do you think of renaming `invalidated` to `orphaned`?

tests/src/tests/zone_sdk/e2e.rs
davidrusu: I think `msg_id` is actually a stronger assertion, but it I think it's too strong for our tests.

Your tests rely on rebasing on conflicts, that will cause the msg ids to change. msg ids are derived from (among other things) both payload and parent msg ids. 

You can't make assertions on the exact msg ids if you are rebasing

tests/src/tests/zone_sdk/e2e.rs
davidrusu: yes, just to re-iterate. The channel is the SoT, clients should be able to re-derive the Zone state from the channel.

They may use the sequencer as a shortcut, but it should be verified against the channel to be sure the sequencer is not making things up.

tests/src/tests/zone_sdk/e2e.rs
davidrusu, line 1: For these tests, we don't need 2 nodes to test the ZoneSDK e2e. We can get away with just 1 (this will remove a lot of flakeyness)

Also, drop the `slot_activation_coeff` to 1/2 to get even faster blocks (it defaults to 1/10 i.e. 1 block every 10 slots, or 10s blocks). With one node network, you don't have reorgs since the node self-applies blocks as soon as it creates it so it's safe to have very fast block times.

```rust
config.deployment.cryptarchia.slot_activation_coeff =
      NonNegativeRatio::new(1, 2.try_into().unwrap());
```

davidrusu: Is this branch not rebased on master? or this test was missed?

we don't need to check against the seen hashset, only seen_ordered is necessary

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -97,18 +113,19 @@ async fn test_sequencer_publish_and_indexer_read() {
         None, // Fresh start, no checkpoint
     );
 
+    // Spawn a task to drive the sequencer event loop
     let poll_task = sequencer.spawn();
-    handle.wait_ready().await;
+    spawn_republish_handler(&handle);
```
davidrusu, line 126: this is a single-sequencer case, so we don't want republishing. If there is a conflict we want to know about it because something is wrong in the SDK.

perhaps its better here to write a new behaviour: `spawn_fail_on_channel_conflict_handler(&handle)`

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -236,18 +254,27 @@ async fn test_sequencer_checkpoint_resume() {
         None, // Fresh start
     );
 
+    // Spawn polling task with reorg handling
     let poll_task = sequencer.spawn();
-    handle.wait_ready().await;
+    spawn_republish_handler(&handle);
```
davidrusu: again, single sequencer case, we don't want to paper over conflicts

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -236,18 +254,27 @@ async fn test_sequencer_checkpoint_resume() {
...
 
-    let test_data_phase1: Vec<Vec<u8>> = vec![b"Message 1".to_vec(), b"Message 2".to_vec()];
+    let test_data_phase1: Vec<Vec<u8>> = vec![tag_payload("Message 1"), tag_payload("Message 2")];
```
davidrusu, line 250: no need to tag, this is a single sequencer case

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -264,12 +291,11 @@ async fn test_sequencer_checkpoint_resume() {
     );
 
     let poll_task = sequencer.spawn();
-    handle.wait_ready().await;
+    spawn_republish_handler(&handle);
```
davidrusu, line 286: ditto, single sequencer, use a `spawn_fail_on_channel_conflict_handler(&handle)` behavior.

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -264,12 +291,11 @@ async fn test_sequencer_checkpoint_resume() {
...
 
-    let test_data_phase2: Vec<Vec<u8>> = vec![b"Message 3".to_vec(), b"Message 4".to_vec()];
-    for data in &test_data_phase2 {
-        handle.publish(data.clone()).await.expect("publish failed");
-    }
+    let test_data_phase2: Vec<Vec<u8>> = vec![tag_payload("Message 3"), tag_payload("Message 4")];
```
davidrusu, line 288: no need to tag

tests/src/tests/zone_sdk/e2e.rs
davidrusu: given that this is expected to be used with the `rebase_on_conflict` conflict resolution behaviour, consider renaming the function to `wait_for_indexer_unordered()`

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -321,6 +347,829 @@ async fn test_sequencer_checkpoint_resume() {
...
+    // Create two signing keys — SeqA is the channel creator/admin
+    let mut key_bytes_a = [0u8; 32];
+    thread_rng().fill(&mut key_bytes_a);
+    let signing_key_a = Ed25519Key::from_bytes(&key_bytes_a);
+    let admin_pk = signing_key_a.public_key();
+    let channel_id = channel_id_from_key(&signing_key_a);
+
+    let mut key_bytes_b = [0u8; 32];
+    thread_rng().fill(&mut key_bytes_b);
+    let signing_key_b = Ed25519Key::from_bytes(&key_bytes_b);
```
davidrusu, line 492: this is common enough that it's worth turning into a helper `keygen() -> Ed25519Key`

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -321,6 +347,829 @@ async fn test_sequencer_checkpoint_resume() {
...
+
+    // --- Phase 2: SeqB publishes b1, b2, b3 ---
+    let (sequencer_b, mut handle_b) = ZoneSequencer::init_with_config(
+        channel_id,
+        signing_key_b,
+        node_url.clone(),
+        None,
+        sequencer_config.clone(),
+        None, // Fresh start — SeqB discovers channel tip from chain
+    );
```
davidrusu, line 536: discovering channel tip from chain could take some time? does init_with_config block until we are in sync?

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -321,6 +347,829 @@ async fn test_sequencer_checkpoint_resume() {
...
+    let mut on_chain_order = Vec::new();
+    loop {
+        let result = indexer
+            .next_messages(cursor, 100)
+            .await
+            .expect("next_messages should succeed");
+        for msg in &result.messages {
+            if expected_all.contains(&msg.data) {
+                on_chain_order.push(msg.data.clone());
+            }
```
davidrusu: why gate? if there is an unexpeted message, that should be a fail, no?

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -321,6 +347,829 @@ async fn test_sequencer_checkpoint_resume() {
...
+}
+
+/// Helper: tag a message with a random ID for reorg deduplication.
+fn tag_payload(msg: &str) -> Vec<u8> {
+    format!("{:016x}:{msg}", rand::random::<u64>()).into_bytes()
+}
+
+#[tokio::test]
+#[serial]
+async fn test_sequential_multi_sequencer() {
```
davidrusu, line 444: for this test, I would add the `fail_on_conflict` conflict resolution handler for each sequencer that we spawn here. We don't expect any conflicts because the sequencers are taking turns, so we should check that.

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -321,6 +347,829 @@ async fn test_sequencer_checkpoint_resume() {
...
+    handle_c.wait_ready().await;
+    debug!("Phase 2: All 3 sequencers ready");
+
+    // Phase 3: Publish initial inscriptions concurrently.
+    debug!("Phase 3: Publishing 9 inscriptions concurrently");
+    tokio::join!(
+        async {
+            for d in &data_a {
+                handle_a.publish(d.clone()).await.expect("publish failed");
+            }
```
davidrusu, line 757: `publish_all()` helper?

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -321,6 +347,829 @@ async fn test_sequencer_checkpoint_resume() {
...
+
+    let unique: HashSet<&Vec<u8>> = all_payloads.iter().collect();
+    assert_eq!(
+        unique.len(),
+        all_payloads.len(),
+        "Duplicate inscriptions detected on chain: expected {} unique, got {} total",
+        unique.len(),
+        all_payloads.len(),
+    );
+    assert_eq!(unique.len(), 9, "Expected exactly 9 inscriptions on chain");
```
davidrusu, line 828: Isn't this what `wait_for_indexer()` is doing?

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -321,6 +347,829 @@ async fn test_sequencer_checkpoint_resume() {
...
+    tokio::join!(
+        async {
+            for d in &data_a {
+                handle_a.publish(d.clone()).await.expect("publish failed");
+            }
+        },
+        async {
+            for d in &data_b {
+                handle_b.publish(d.clone()).await.expect("publish failed");
+            }
```
davidrusu, line 1056: publish_all?

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -321,6 +347,829 @@ async fn test_sequencer_checkpoint_resume() {
...
+    // The key invariant: whatever survived on chain must be sorted.
+    let is_sorted = on_chain.windows(2).all(|w| w[0] <= w[1]);
+    assert!(
+        is_sorted,
+        "On-chain payloads must be sorted, got: {:?}",
+        on_chain
+            .iter()
+            .map(|p| String::from_utf8_lossy(p).to_string())
+            .collect::<Vec<_>>()
+    );
```
davidrusu, line 1129: fun example :)

zone-sdk/src/sequencer.rs
```diff
@@ -127,6 +164,9 @@ impl SequencerHandle {
     /// Publish an inscription to the zone's channel.
     ///
     /// Returns the inscription ID and a checkpoint for persistence.
+    ///
+    /// TODO: make fire-and-forget so clients can call from event handlers
+    /// without spawning a task. Currently goes through the actor loop.
```
davidrusu: Which part do you want to do off-thread? the tx signing? I think we cannot return until we are sure that the actor has updated it's internal state so that subsequent publishes parent off of this one.

Else we risk something like
```
publish(a) // fire and forget
publish(b) // fire and forget

// <tokio re-orders publish tasks>
```
and the b get's handled first by the actor and we end up with [b, a] in the channel.

zone-sdk/src/sequencer.rs
```diff
@@ -318,9 +364,25 @@ impl ZoneSequencer {
...
+                // Try to extract inscription metadata for lineage tracking
+                let mut is_inscription = false;
+                for op in &tx.mantle_tx.ops {
+                    if let Op::ChannelInscribe(inscribe) = op {
+                        tx_state.submit_inscription(
+                            tx.clone(),
+                            inscribe.parent,
+                            inscribe.id(),
+                            inscribe.inscription.clone(),
+                        );
```
davidrusu, line 482: why not pass the full `inscribe` op to `submit_inscription()`?

zone-sdk/src/sequencer.rs
```diff
@@ -318,9 +364,25 @@ impl ZoneSequencer {
...
                 cp.lib_slot
             );
-            let mut tx_state = TxState::new(cp.lib);
+            let mut tx_state = TxState::new(cp.lib, cp.last_msg_id);
             for (_hash, tx) in cp.pending_txs {
-                tx_state.submit(tx);
+                // Try to extract inscription metadata for lineage tracking
+                let mut is_inscription = false;
+                for op in &tx.mantle_tx.ops {
+                    if let Op::ChannelInscribe(inscribe) = op {
```
davidrusu, line 476: This is probably paranoid, but should we filter these op's by channel id, i.e. add a `&& inscribe.channel == channel_id`.

zone-sdk/src/sequencer.rs
```diff
@@ -476,31 +538,72 @@ impl ZoneSequencer {
...
+        };
+
+        let from_u64: u64 = from.into();
+        let to_u64: u64 = to.into();
+
+        if from_u64 > to_u64 {
+            self.backfill_from = None;
+            self.backfill_to = None;
+            return None;
+        }
```
davidrusu, line 687: this seems like it should be an assert rather than returning None

zone-sdk/src/sequencer.rs
```diff
@@ -850,39 +1118,58 @@ fn enqueue_resubmit(
...
+                }
+            })
         })
         .collect();
 
-    if inscriptions.is_empty() {
-        return None;
+    if items.len() <= 1 {
+        return items;
     }
```
davidrusu: the following code works for the len() == 1 case right? no need for this special case

zone-sdk/src/sequencer.rs
```diff
@@ -850,39 +1118,58 @@ fn enqueue_resubmit(
...
+        .expect("inscriptions for a channel in a block must form a chain (no root found)");
+
+    let mut sorted = Vec::with_capacity(items.len());
+    sorted.push(root.clone());
+    let mut current = root.this_msg;
+    while let Some(next) = by_parent.get(&current).copied() {
+        sorted.push(next.clone());
+        current = next.this_msg;
+    }
+    sorted
```
davidrusu, line 1348: So, inscriptions are guaranteed to be in sorted order, blocks are monotonically ordered by slots, and within blocks, transactions are ordered, and within transactions, ops are ordered.

So we don't need this topological sort here.

To be safe, what we can do is to replace all this sorting logic with a scan through `items` and `assert_eq!(item.parent, previous_item.id())`

zone-sdk/src/sequencer.rs
```diff
@@ -696,63 +967,60 @@ async fn get_lib_slot(
...
-    }
-
-    debug!(
-        "Backfilling finalized blocks from slot {} to {}",
-        from + 1,
-        to
-    );
-
-    let mut latest_msg_id = None;
+) -> FetchedBatch {
```
davidrusu: so, how does this deal with forks? we linearize everything onto the slot timeline, but not all the those txs and inscriptions in the returned `FetchedBatch` are on the canonical chain right?

zone-sdk/src/sequencer.rs
```diff
@@ -659,16 +849,97 @@ async fn handle_block_event(
...
+                .iter()
+                .find_map(|op| {
+                    if let Op::ChannelInscribe(i) = op {
+                        Some(String::from_utf8_lossy(&i.inscription).to_string())
+                    } else {
+                        None
+                    }
+                })
+                .unwrap_or_else(|| "non-inscription".to_owned());
+            debug!(" Backfill-finalized: payload={payload_str:?}, tx={tx_hash:?}");
```
davidrusu: don't forget to remove this before we merge, don't want to be logging all inscriptions

zone-sdk/src/sequencer.rs
```diff
@@ -659,16 +849,97 @@ async fn handle_block_event(
...
+            .invalidated
+            .extend(stale.into_iter().filter(|i| !existing.contains(&i.tx_hash)));
+    } else {
+        *channel_update = Some(crate::state::ChannelUpdateInfo {
+            invalidated: stale,
+            adopted: Vec::new(),
+            new_channel_tip: s.channel_tip_at(tip),
+        });
     }
 }
```
davidrusu, line 1127: With our logic, there should not be any stale pending right? on conflict, we purge all transactions that were invalidated from pending and return to user?

So I'm wondering if we should reduce this to an `assert!(stale.is_empty())`

tests/src/tests/zone_sdk/e2e.rs
hansieodendaal: It is important for a client to know if the actual inscription made it through, so the report back from the indexer loop should use the stronger match. It would be a problem if the inscription became warped and changed `msg_id` somehow.

tests/src/tests/zone_sdk/e2e.rs
hansieodendaal: I am happy that the indexer is also used in the tests as confirmation, but my comment stands: our current design is not efficient as pointed out. That being said, it is good enough for now for a POC as it gets the job done, and we can always decide what to tweak later on.

tests/src/tests/zone_sdk/e2e.rs
hansieodendaal: This was fixed - payloads could be duplicated

tests/src/tests/zone_sdk/e2e.rs
hansieodendaal: I think the async handling here nicely demonstrates the resiliency we need.

tests/src/tests/zone_sdk/e2e.rs
hansieodendaal: I would stick with `invalidated`, as the only way to resubmit the inscription is the recreate it.

tests/src/tests/zone_sdk/e2e.rs
hansieodendaal: I believe the `spawn_republish_handler` step is done here because we do not currently have a contract with the node's mempool. The mempool can misbehave and throw away or forget about inscription transactions. This is currently the reliable way to ensure the mempool will continue to try include the transaction in a block proposal until it is mined or we know the inscription is invalidated.

testnet/tui-zone/src/lib.rs
```diff
@@ -109,7 +141,12 @@ pub async fn run(args: InscribeArgs) {
             break;
         }
 
-        match handle.publish(msg.as_bytes().to_vec()).await {
+        // Tag payload with a random ID so we can identify our inscriptions
+        // during reorg handling (avoids duplicate re-publishing).
+        let id: u64 = rand::random();
+        let tagged_payload = format!("{id:016x}:{msg}");
```
hansieodendaal: I do not think this responsibility should lie with the client, and it is warping what the inscription data actually should be.

zone-sdk/src/sequencer.rs
hansieodendaal: I would keep the `None`, as the client can then deal with it and not crash.

zone-sdk/src/sequencer.rs
hansieodendaal: We could make it a trace maybe?

zone-sdk/src/sequencer.rs
hansieodendaal: I think it is a good idea to keep this functionality around until proven we never have any stale inscription. I would not add an assert here. If in the end this would be an error condition, I would rather return an error so the client can handle it than panic.

zone-sdk/src/sequencer.rs
hansieodendaal: I would rather keep this check and log an error if the information happens to be not sorted as expected and always return valid information to the client. No need to panic if we can fix it for the client.

tests/src/tests/zone_sdk/e2e.rs
davidrusu: The behaviour of the `republish_handler` is that if there is a conflicting write to the channel that causes your inscription to be abandoned, it republishes the message onto the new tip of the channel, this means that the same inscription will have a new message id.

tests/src/tests/zone_sdk/e2e.rs
davidrusu: So, I think we're mixing things, what you are suggesting is called 'preconfirmations' and this something that is specific to the design of a zone, each zone will have it's own type of preconfirmation that is deeply connected to the internal model of that zone. i.e. EVM preconfirmations will look very different from LEZ preconfirms. 

We don't want to provide that at the zone sdk level, it's up to the sequencer to provide this to its users.

What the zone sdk is providing is a way to interact with the L1, you can build the preconfs on top of this.

tests/src/tests/zone_sdk/e2e.rs
davidrusu: @hansieodendaal are you advocating for the current code? or suggesting that we should not spawn and simply await the publish() in the loop?

tests/src/tests/zone_sdk/e2e.rs
davidrusu: I like `orphaned` as it has the symmetry with `adopted` but `invalidated` is good as well

tests/src/tests/zone_sdk/e2e.rs
davidrusu: No, this is not related to the mempool dropping transactions. This is different from the republishing of pending txs, this is rebasing the invalidated inscriptions on top of the new tip.

zone-sdk/src/sequencer.rs
davidrusu: This is an internal call, it doesn't get exposed to the client.

These `backfill_from` and `backfill_to` fields are only set internally by the SDK, never by the client.

We should crash early if there is any invariant broken in our system rather than attempting to paper over the failures (This is the pattern we use in our node)

zone-sdk/src/sequencer.rs
davidrusu: If we follow that approach, we pollute our code with all sorts of arbitrary dead code "just in case" we missed something. That's not how we make simple systems.

And, we should be asserting our invariants, you can report an error to the user if its within your expected state space, but if we invalidate an invariant, that's outside our state space so we should panic.

e.g. rust's stdlib HashMap uses asserts internally to check invariants.

zone-sdk/src/sequencer.rs
davidrusu: This is a core assumption in our API, if it's broken than something bigger is wrong. We should not be papering over this bug by trying to fix it here

tests/src/tests/zone_sdk/e2e.rs
hansieodendaal: I am happy with the current async handling implementation in the test, voting for it.

tests/src/tests/zone_sdk/e2e.rs
pradovic: I agree. Should we leave 2 nodes for competing sequencers test (2 tests)?

tests/src/tests/zone_sdk/e2e.rs
pradovic: Returns immediately. Channel tip discovery and backfill happen lazily during `next_event()` calls via `ensure_connected()`

zone-sdk/src/sequencer.rs
pradovic: Good idea, let me refactor this in a separate PR. I will refactor everything in separate files, and unify the types and lifecycle objects better. It would be hard to review now I think

hansieodendaal (approved):
LGTM

_(All new zone sdk e2e tests pass)_

PR #2449, opened 2026-03-30
feat: Use non spawned task in buffered stream

danielSanchezQ:

Avoid spawning a background task to create a buffered stream. Instead, poll the stream already on creation, and keep a large enough buffer.

Yes.

@danielSanchezQ with the idea and most of the implementation, @ntn-x2 with the polishing and some ergonomics improvements.

N/A

No.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5259,7 +5259,6 @@ dependencies = [
  "serde_with",
  "time",
  "tokio",
- "tokio-stream",
 ]
 
 [[package]]
--- a/utils/Cargo.toml
+++ b/utils/Cargo.toml
@@ -13,25 +13,24 @@ version     = { workspace = true }
 workspace = true
 
 [dependencies]
-async-trait  = "0.1"
-blake2       = { default-features = false, optional = true, version = "0.10" }
-cipher       = { version = "0.4" }
-const-hex    = { default-features = false, version = "1" }
-futures      = { optional = true, workspace = true }
-humantime    = { optional = true, version = "2.1" }
-overwatch    = { workspace = true }
-rand         = { features = ["getrandom"], workspace = true }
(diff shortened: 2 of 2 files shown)


utils/src/tokio/stream.rs
```diff
@@ -3,64 +3,65 @@ use core::{
...
+        let peeked = poll_fn(|cx| {
+            Poll::Ready(match stream.as_mut().poll_next(cx) {
+                Poll::Ready(item) => item,
+                Poll::Pending => None,
+            })
+        })
+        .await;
+
+        Self { stream, peeked }
     }
```
danielSanchezQ: The idea is to pre-poll to trigger everything. This means that we have to run it in an async context. Then I had to do some type massaging (mostly because of the use of the `buffered` method, `Buffered` type).

utils/src/tokio/stream.rs
```diff
@@ -3,64 +3,65 @@ use core::{
...
-    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        self.stream.poll_next_unpin(cx)
+    type Item = <WrappedStream::Item as Future>::Output;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        // SAFETY: We do not move `Buffered` out of the pin.
+        // - `peeked` is a plain `Option<Output>` with no self-referential state.
+        // - `stream` is `Pin<Box<...>>` which carries its own pin guarantee; we only
+        //   call `as_mut()` on it, never move it.
+        let this = unsafe { self.get_unchecked_mut() };
```
danielSanchezQ: I hate this, I had to consult the AI for a workaround, but couldn't find it 🤔

PR #2450, opened 2026-03-30
feat(ledger): Gas Context (2/3)

strinnityk:
Adds Gas Context in advance for ChannelWithdraw gas calculation requirements.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/core/src/mantle/ops/channel/mod.rs
+++ b/core/src/mantle/ops/channel/mod.rs
@@ -14,10 +14,8 @@ serde_bytes_newtype!(ChannelId, 32);
 
 impl Display for ChannelId {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        for byte in &self.0 {
-            write!(f, "{byte:02x}")?;
-        }
-        Ok(())
+        let hex_string = hex::encode(self.0);
+        write!(f, "{hex_string}")
     }
 }
 


core/src/mantle/tx.rs
```diff
@@ -151,21 +174,22 @@ impl<'de> Deserialize<'de> for MantleTx {
...
+        let storage_gas = self.signed_serialized_size(context);
         execution_gas * self.execution_gas_price + storage_gas * self.storage_gas_price
     }
 }
 
 impl MantleTx {
     #[must_use]
-    pub fn signed_serialized_size(&self) -> u64 {
+    pub fn signed_serialized_size(&self, _context: &MantleTxGasContext) -> u64 {
         super::encoding::predict_signed_mantle_tx_size(self) as u64
```
youngjoon-lee, line 193: If we don't use `_context` in this function, wdyt about just removing it?

Maybe, I didn't understand the purpose of the gas context. Could you please explain what it is and how it should be used?

core/src/mantle/ops/channel/mod.rs
```diff
@@ -2,12 +2,25 @@ pub mod deposit;
...
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
 pub struct ChannelId([u8; 32]);
 serde_bytes_newtype!(ChannelId, 32);
 
+impl Display for ChannelId {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        for byte in &self.0 {
+            write!(f, "{byte:02x}")?;
+        }
+        Ok(())
```
danielSanchezQ: I think we use `hex` for this kind of stuff usually (?)

core/src/mantle/tx.rs
danielSanchezQ: Before the new channel operations we didnt need any external information related to the tx to compute the gas. Now for some operations we need a context which is information that lives outside the tx itself (in this case the channel thresholds).

youngjoon-lee: Yeah, but I don't find where the gas context is used in this PR.

youngjoon-lee: Ahhhhh, okay, it is used in a next PR.. I got it.

PR #2451, opened 2026-03-30
feat(ledger): Implement Channel Withdraw op (3/3)

strinnityk:
Implement Channel Withdraw operation

#### Design Choices
Proof verification has been split into two paths:
- `verify_ops_proofs`: covers checks that do not require external context.
- `verify_ops_proofs_with_helper`: covers checks that depend on additional state (e.g. `ChannelWithdraw`, which requires channel withdrawal thresholds).

Because full verification depends on external data (channel state), it is not possible to construct a fully verified transaction at deserialization time. Deserialization occurs without access to this context.

As a result, transactions are deserialized into a partially verified state. Full verification is deferred until the required context is available, typically just before applying the transaction.

This introduces a trust-based assumption: any component that requires a fully verified Tx must ensure that verification has already been completed.

This approach was chosen to keep the implementation simple given current constraints. A follow-up refactor is planned to replace this with a trait-based/type-state model that enforces verification correctness at compile time.

Yes

@strinnityk 

Yes
https://www.notion.so/nomos-tech/v1-3-Mantle-Specification-330261aa09df80a899a6efd74f12a7c4

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/ledger/src/lib.rs
+++ b/ledger/src/lib.rs
@@ -399,11 +399,34 @@ impl LedgerState {
         self.mantle_ledger.active_sessions()
     }
 
-    // Some of the ops' proofs are verified in `SignedMantleTx::verify_ops_proofs`,
-    // either fully or partially.
-    // This probably benefits from some refactoring.
-    // For now, though, you should manually ensure that everything is correctly
-    // verified.
+    /// Applies a transaction to the ledger state, returning the updated state
+    /// and the net balance change.
+    ///
+    /// # Prerequisites
+    ///
+    /// A transaction must not be applied unless all required proofs have been
+    /// fully verified.
+    ///
+    /// Proof verification is currently split across multiple paths depending on
+    /// the operation:
+    /// - `SignedMantleTx::verify_ops_proofs`: Invoked during construction
+    ///   (`SignedMantleTx::new`, e.g. on deserialization). Handles:
+    ///   `ChannelInscribe`, `LeaderClaim`.
+    /// - `SignedMantleTx::verify_ops_proofs_with_helper`: Invoked here before
+    ///   applying the transaction. Handles: `ChannelWithdraw`.
+    /// - Additional validation: Performed by the ledger or implicitly satisfied
+    ///   by certain operations.
+    ///
+    /// This fragmented design means verification may be:
+    /// - Distributed across different stages, and
+    /// - Potentially duplicated or missed if assumptions about prior
+    ///   verification are incorrect.
+    ///
+    /// Callers are responsible for ensuring that all required proofs have been
+    /// verified before applying the transaction.
+    ///
+    /// TODO: A refactor into a typed state model to enforce verification at
+    /// compile is planned.
     fn try_apply_tx<Id, Constants: GasConstants>(
         mut self,
         config: &Config,


ledger/src/lib.rs
```diff
@@ -443,17 +469,18 @@ impl LedgerState {
...
                 }
-                (Op::SDPWithdraw(op), Some(OpProof::ZkSig(sig))) => {
+                (Op::SDPWithdraw(op), OpProof::ZkSig(sig)) => {
                     self.mantle_ledger = self
                         .mantle_ledger
                         .try_apply_sdp_withdraw(op, sig, tx_hash, config)?;
                 }
-                (Op::LeaderClaim(op), None) => {
+                // TODO: Double check the proof for LeaderClaim is correct
+                (Op::LeaderClaim(op), OpProof::PoC(_)) => {
```
strinnityk, line 506: @youngjoon-lee @thomaslavaur @danielSanchezQ 
Please, don't miss this one. Is it okay as it is now?

danielSanchezQ: Just seeing it. We were not checking claims as that `None` would have never reached anyway. We should definitely be checking the proof, I just don't know now if it was disabled because we were missing some machinery. cc @youngjoon-lee @thomaslavaur

thomaslavaur: Yes I'm on it

core/src/mantle/tx.rs
```diff
@@ -333,7 +383,53 @@ impl SignedMantleTx {
                     }
                 }
                 // Other operations are checked by the ledger or don't require verification here
-                _ => {}
+                _ => {
+                    // TODO: If the op and proof don't match, we are silently
+                    // delaying the error
+                    //  until tx execution.
+                }
```
youngjoon-lee, line 390: So, does this TODO mean that we should fail early?

ledger/src/lib.rs
```diff
@@ -394,28 +399,37 @@ impl LedgerState {
         self.mantle_ledger.active_sessions()
     }
 
+    // Some of the ops' proofs are verified in `SignedMantleTx::verify_ops_proofs`,
+    // either fully or partially.
+    // This probably benefits from some refactoring.
+    // For now, though, you should manually ensure that everything is correctly
+    // verified.
     fn try_apply_tx<Id, Constants: GasConstants>(
```
youngjoon-lee: I don't clearly understand what this comment means. Does this mean that this function verifies all the proofs regardless of whether they were already verified before (iirc, this is what we agreed somewhere in Discord)?

ledger/src/lib.rs
youngjoon-lee: ah, i think this is a leftover. thank you.

ledger/src/lib.rs
youngjoon-lee: If my understanding is correct, I think it's better to write a proper doc comment with a clear TODO.

For example,
```rust
/// Applies a transaction with verifying proofs of operations, regardless of whether they have been verified already by the caller.
//
// TODO: refactor: introduce a type for a Tx with verified proofs.
```

core/src/mantle/tx.rs
strinnityk: Hm, technically it's probably more of a clarification rather than a TODO.
Tbf, I would expect this to be refactored at soom point in the near future. I think there's no reason to let the TX pass if it's malformed to begin with.

ledger/src/lib.rs
strinnityk: Hmmm, let me try to rephrase, see if we can make it more understandable.
```rust
/// A transaction must not be applied unless all required proofs have been fully verified.
///
/// Proof verification is currently split across multiple paths depending on the operation:
/// - `SignedMantleTx::verify_ops_proofs`:
///   Invoked during construction (`SignedMantleTx::new`, e.g. on deserialization).
///   Handles: ChannelInscribe, LeaderClaim.
/// - `SignedMantleTx::verify_ops_proofs_with_helper`:
///   Invoked here before applying the transaction.
///   Handles: ChannelWithdraw.
/// - Additional validation:
///   Performed by the ledger or implicitly satisfied by certain operations.
///
/// This fragmented design means verification may be:
/// - Distributed across different stages, and
/// - Potentially duplicated or missed if assumptions about prior verification are incorrect.
///
/// Callers are responsible for ensuring that all required proofs have been verified
/// before applying the transaction.
///
/// TODO: A refactor into a typed state model to enforce verification at compile is planned.
```

youngjoon-lee: Thank you for clarification. I understand the intention. I think it's good to have this comment in the code, though it is quite long, because someone else might have the same confusions we had initially. Anyway, we will simplify the code and the comment after resolving the TODO.

PR #2454, opened 2026-03-31
chore: panic on unexpected circuit version

ntn-x2:

Until we migrate circuits to be a library, we need to enforce that the circuits match the version the node expects. @andrussal and I spent the whole day trying to debug an issue which was caused by an outdated version of the circuits, with the latest version only updating leadership proofs.

My suggestion is to enforce the expected version in the zk utility library, which is called both at build time in the various `build.rs` modules as well as runtime. So now it's not possible to build a node binary referencing an unexpected circuits version, and if using a pre-compiled binary, the right circuits must be set up before the node can correctly produce a block.

This should avoid issues related to running outdated circuit binaries going forward, until we swap the binaries with bundled static libraries.

Yes.

@ntn-x2 

N/A

No.

### DIFF
--- a/zk/circuits/utils/src/lib.rs
+++ b/zk/circuits/utils/src/lib.rs
@@ -1,7 +1,8 @@
-use std::path::PathBuf;
+use std::{fs::read_to_string, path::PathBuf};
 
 const LOGOS_BLOCKCHAIN_CIRCUITS_ENV_VAR: &str = "LOGOS_BLOCKCHAIN_CIRCUITS";
 const LOGOS_BLOCKCHAIN_CIRCUITS_DEFAULT_DIR: &str = ".logos-blockchain-circuits";
+const EXPECTED_CIRCUITS_VERSION: &str = "v0.4.2";
 
 #[cfg(target_os = "windows")]
 const BINARY_NAME: &str = "witness_generator.exe";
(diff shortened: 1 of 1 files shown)


zk/circuits/utils/src/lib.rs
```diff
@@ -1,7 +1,8 @@
-use std::path::PathBuf;
+use std::{fs::read_to_string, path::PathBuf};
 
 const LOGOS_BLOCKCHAIN_CIRCUITS_ENV_VAR: &str = "LOGOS_BLOCKCHAIN_CIRCUITS";
 const LOGOS_BLOCKCHAIN_CIRCUITS_DEFAULT_DIR: &str = ".logos-blockchain-circuits";
+const EXPECTED_CIRCUITS_VERSION: &str = "v0.4.2";
```
hansieodendaal, line 5: I think we need a unit test that will bind this value to the expected version supported by the code, as this value can get outdated. So if someone updates the circuits and related code, the gate-keeper unit test must fail if the constant is not updated.

The installed circuits and this constant can still be the same, but outdated.

danielSanchezQ: But we have tests using circuits already, if the circuits gets updated the tests will fail. If this value is not in sync something will fail already (?)

hansieodendaal: Yeah, agreed, it will fail eventually, even if only in CI.

youngjoon-lee (approved):
👍

zk/circuits/utils/src/lib.rs
ntn-x2: Yes agree with @danielSanchezQ. If any of them mismatches, the code will fail to build and/or generate any proofs by panicking. If the rest of the codebase changes without changing the circuits, it will also fail in some tests. If the circuits are updated without updating the rest of the codebase, the tests elsewhere will also fail.

PR #2457, opened 2026-04-01
feat(zone-sdk): atomic ops+inscribe

youngjoon-lee:

This PR allows sequencers/users to perform Deposit and Inscribe atomically.

### Background

So far, a sequencer had to [subscribe to **finalized** deposits](https://github.com/logos-blockchain/logos-blockchain/pull/2399) submitted by users directly to the Bedrock via [HTTP API](https://github.com/logos-blockchain/logos-blockchain/pull/2443). It is actually not so bad, and is what's described in the [spec](https://www.notion.so/nomos-tech/v1-3-Mantle-Specification-330261aa09df80a899a6efd74f12a7c4?source=copy_link#330261aa09df81b8adb7cbfa9f58f801):
> Note that the Zone may wait for the deposit to be finalized before interpreting the deposit in order to guarantee that the deposit will occur on-chain and won't be removed due to reorganization of the chain.

However, in real world, sequencers may want to interpret deposits and perform state transitions before they are finalized. For those use cases, this PR introduces an atomic way: A sequencer gathers deposit operations from users, builds a tx, gathers signatures from users, and submits it to Bedrock.

NOTE: A deposit op must always be accompanied by a transfer op because the depositer must burn his note to deposit it to the channel. It is described in the [spec](https://www.notion.so/nomos-tech/v1-3-Mantle-Specification-330261aa09df80a899a6efd74f12a7c4?source=copy_link#330261aa09df811c9a4ccf2ba17c6c45) and an [API example](https://github.com/logos-blockchain/logos-blockchain/pull/2443).

### Changes
Two new functions have been added:
- `ZoneSequencer::prepare_tx(ops, msg)`
  - A sequencer can gather operations (deposits, transfers, ...) and can build a `MantleTx` with them, including an additional inscription that represents the zone state transition corresponding the those operations. Please note that this function returns the `MantleTx` that is not signed yet.
  - Then, the sequencer can ask the user (who wrote the deposit op and the transfer op) to sign the `MantleTx` using his key. By doing so, the sequencer can finally build a `SignedMantleTx`. This communication is out of scope of `zone-sdk`.
- `ZoneSequencer::submit_signed_tx(tx, msg_id)`
  - The sequencer can finally submit the signed tx to the Bedrock. The function shares the internal implementation of `ZoneSequencer::publish` which handle re-submissions. 

If this PR looks good, I will add this feature to the `tui-zone`.

Yes

@youngjoon-lee 
spec: @thomaslavaur 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/zone-sdk/src/sequencer.rs
+++ b/zone-sdk/src/sequencer.rs
@@ -91,6 +91,11 @@ enum ActorRequest {
         reply: tokio::sync::oneshot::Sender<Result<(SignedMantleTx, PublishResult), Error>>,
     },
     /// Build an unsigned tx for the given ops and an inscription
+    ///
+    /// Calling this multiple times without submitting the prepared txs via
+    /// `SubmitSignedTx` can cause parent msg ID conflicts, so ensure
+    /// prepared txs are submitted promptly. If additional prepares are
+    /// unavoidable, handle potential conflicts carefully.
     PrepareTx {
         ops: Vec<Op>,
         msg: Vec<u8>,


zone-sdk/src/sequencer.rs
```diff
@@ -517,21 +608,33 @@ where
...
             }
+            ActorRequest::PrepareTx { ops, msg, reply } => {
+                let result = prepare_tx(
+                    ops,
+                    self.channel_id,
+                    &self.signing_key,
+                    msg,
+                    self.last_msg_id,
+                );
+                // do not update last_msg_id since tx is not submitted yet
```
pradovic, line 636: May be worth noting in the doc comment that `prepare_tx` should be called one at a time. Two concurrent `prepare_tx` calls will use the same parent and conflict. Or maybe return an error if there's already a prepared-but-not-submitted tx?

youngjoon-lee: Yeah, good point. I thought about managing a prepared-but-not-submitted tx, but chose to not do it, because a sequencer may want to give up `tx1` (already prepared) and prepare `tx2` instead in some cases. For example, the sequencer prepared `tx1` with `op1` (from user1) and `op2` (from user2). The sequencer requests users to sign on the `tx1` but only user2 replies. User1 is suddenly gone. In this case, the sequencer may want to build a new `tx2` only with `op2` (from user2). 
We may be able to introduce some functions to cover this prepare/discard scenario, but I think we can keep it simple for now unless it becomes truly necessary.

So, in this PR, I just improved the doc comment 9b86e4cbdfa668e465384622fc00dab1f4bad9b5, as you suggested.

PR #2473, opened 2026-04-04
fix: validate MAX_MSG_LEN before allocation in unpack_from_reader

### LINKED ISSUE #2460: MAX_MSG_LEN not checked in unpack_from_reader
MAX_MSG_LEN (16 MiB) is never checked before the allocation. This constant is defined but only used for an error. A malicious peer sends a 4-byte header with data_length = u32::MAX and causes a ~4 GiB heap allocation attempt per stream.

giwaov:
## Summary

Validates the length prefix against \MAX_MSG_LEN\ (16 MiB) **before** allocating the receive buffer in \unpack_from_reader\. Without this check, a malicious peer can send a 4-byte header with \data_length = u32::MAX\ and force a ~4 GiB heap allocation per stream.

## Changes

- **\unpack_from_reader\**: Added an early \data_length > MAX_MSG_LEN\ guard that returns \PackingError::MessageTooLarge\ before the \ec!\ allocation.
- **\pack_to_writer\**: Added an explicit \MAX_MSG_LEN\ check on the serialized payload. Previously it only rejected messages that overflowed \u32\, not the 16 MiB protocol limit.
- Added a unit test (\
eject_oversized_message_on_read\) that confirms oversized length prefixes are rejected without allocating.

## Related Issue

Closes #2460

### DIFF
--- a/c-bindings/src/api/lifecycle.rs
+++ b/c-bindings/src/api/lifecycle.rs
@@ -164,7 +164,7 @@ pub unsafe extern "C" fn stop_node(node: *mut LogosBlockchainNode) -> OperationS
 
 #[cfg(test)]
 mod test {
-    use std::{path::PathBuf, sync::LazyLock};
+    use std::{ffi::CString, path::PathBuf, sync::LazyLock};
 
     use crate::api::lifecycle::{start_lb_node, stop_node};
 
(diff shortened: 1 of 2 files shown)


danielSanchezQ:
@giwaov This is nice thanks!. I approved CI, once it is green I'll merge this. PR will probably need to be updated tho (rebase/merge).

davidrusu:
@giwaov can you rebase on master and fix the formatting Ci failures?

giwaov:
> @giwaov can you rebase on master and fix the formatting Ci failures?

I'm on it

ntn-x2:
@hansieodendaal can you confirm this is a duplicate of one of the existing PRs?

hansieodendaal:
> @hansieodendaal can you confirm this is a duplicate of one of the existing PRs?

@ntn-x2 

This PR is obsolete now. The receive-side fix was independently implemented and merged in #3008, so master already rejects oversized peer-provided length prefixes before allocating. The send-side part is now superseded by #3423 (when merged), which enforces the bound during serialization rather than serializing unrestricted first and checking the result afterward.

The only part worth carrying over may be the explicit receive-side regression test from this PR (`async fn reject_oversized_message_on_read() {`).

ntn-x2:
Cc @giwaov

PR #2495, opened 2026-04-08
feat(mmr): change ACCEPTABLE_MAX_HEIGHT: 32 -> 33, as required by circuits

youngjoon-lee:

Increased `ACCEPTABLE_MAX_HEIGHT` from 32 to 33 in the MMR crate to match the PoC circuit's [`proof_of_membership(32)`](https://github.com/logos-blockchain/logos-blockchain-circuits/blob/1ce5a77e4420fa761486f42d4a3d28ebbedff329/mantle/poc.circom#L51) template, which requires 32 siblings, except root (i.e., height 33 tree).

This behavior is identical to the [`DynamicMerkleTree`](https://github.com/logos-blockchain/logos-blockchain/blob/725806bd05b93ee49db14aecfb8b376222379411/utxotree/src/merkle.rs#L14) that we are currently using and that will be replaced with MMR soon in next PRs.

I believe so.

@youngjoon-lee 

We don't have spec for MMR. Instead, I followed the circuit [implementation](https://github.com/logos-blockchain/logos-blockchain-circuits/blob/1ce5a77e4420fa761486f42d4a3d28ebbedff329/mantle/poc.circom#L51).

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/mmr/src/lib.rs
+++ b/mmr/src/lib.rs
@@ -12,7 +12,7 @@ pub use path::MerklePath;
 use path::{update_paths_above_merge, update_paths_at_merge};
 
 const EMPTY_VALUE: Fr = Fr::ZERO;
-const ACCEPTABLE_MAX_HEIGHT: u8 = 32;
+const ACCEPTABLE_MAX_HEIGHT: u8 = 33;
 
 /// An append-only persistent Merkle Mountain Range (MMR), which can accept up
 /// to 2^(`MAX_HEIGHT`-1) elements (leaves).
@@ -283,7 +283,7 @@ mod test {
     #[expect(clippy::clone_on_copy, reason = "for the sake of the test")]
     fn test_empty_roots() {
         let mut root = Fr::ZERO;
-        for i in 1..=32 {
+        for i in 1..=ACCEPTABLE_MAX_HEIGHT {
             assert_eq!(root, empty_subtree_root::<ZkHasher>(i));
             root = <ZkHasher as Digest>::compress(&[root.clone(), root]);
         }


thomaslavaur:
I think there is something wrong here. At least in the naming that may not match the cryptographic wording. The depth is equal to the number of sibling excluding the root. If you have one leaf and one sibling the depth is 1.

thomaslavaur:
So I think that's not the right change (the constant). The MMR implementation instead may be wrong

mmr/src/lib.rs
```diff
@@ -12,7 +12,7 @@ pub use path::MerklePath;
 use path::{update_paths_above_merge, update_paths_at_merge};
 
 const EMPTY_VALUE: Fr = Fr::ZERO;
-const ACCEPTABLE_MAX_HEIGHT: u8 = 32;
```
thomaslavaur, line 15: it should be 32

thomaslavaur (changes requested):

mmr/src/lib.rs
thomaslavaur, line 87: a root alone is of height 0

thomaslavaur, line 136: here too the root alone is of height 0

thomaslavaur, line 130: I think it should start at 0 and stop at max_height-1

thomaslavaur, line 232: it should start at 0 and stop at max_height-1

thomaslavaur, line 237: 1..acceptable_max - 1 because the first one is the EMPTY_VALUE

mmr/src/lib.rs
youngjoon-lee: I understand your point. I also had the same confusion when I read this code first. 
In this MMR implementation, "height" starts from 1, and includes the root. I don't think we should modify it right now. In a separate PR, I will try to use the same convention that `DynamicMerklePath` uses (0-based).

All of your other comments are related to this topic.

thomaslavaur: Yes it's just about vocabulary. We could merge this one temporarily

thomaslavaur (approved):
As long as we add it to the todo list to come back on later it's good

mmr/src/lib.rs
youngjoon-lee: Yeah, thanks. I opened an issue: https://github.com/logos-blockchain/logos-blockchain/issues/2539

PR #2496, opened 2026-04-08
fix(mmr): use raw element as leaf instead of hashing it

youngjoon-lee:
The MMR was applying `Hash::digest()` to leaves on insertion, but the PoC circuit expects raw voucher commitments as leaves (matching the [`DynamicMerkleTree`](https://github.com/logos-blockchain/logos-blockchain/blob/a62c09a7f8ad04b574931eca5c1d3b26f61c9306/utxotree/src/merkle.rs#L53) behavior that will be replaced by MMR in nexst PRs).

So I removed the double-hashing in MMR, so the merkle root and paths match what the circuit computes. It's possible because the type of both internal nodes and leaves is `Fr`.

I believe so. 

@youngjoon-lee 
spec: @thomaslavaur 

AFAIK, the [PoC](https://www.notion.so/nomos-tech/v1-3-Mantle-Specification-330261aa09df80a899a6efd74f12a7c4?source=copy_link#330261aa09df81d48e26e141e8eed17b) spec or any other specs don't cover this topic clearly. But I checked that this fix works with the [PoC](https://github.com/logos-blockchain/logos-blockchain-circuits/blob/1ce5a77e4420fa761486f42d4a3d28ebbedff329/hash_bn/merkle.circom#L34) circuit.

It'd be great if this topic is clearly covered by specs, but I think it's not urgent. Maybe I failed to find specs that are already covering the topic.

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/mmr/src/lib.rs
+++ b/mmr/src/lib.rs
@@ -83,7 +83,7 @@ where
             return Err(MmrFull);
         }
 
-        let root = Hash::digest(&[*elem.as_ref()]);
+        let root = *elem.as_ref();
         let mut last_root = Root { root, height: 1 };
         let mut roots = self.roots.clone();
 
@@ -132,7 +132,7 @@ where
                 .collect(),
         };
 
-        let root = Hash::digest(&[*elem.as_ref()]);
+        let root = *elem.as_ref();
         let mut last_root = Root { root, height: 1 };
         let mut roots = self.roots.clone();
 
@@ -276,7 +276,7 @@ mod test {
     }
 
     pub fn leaf(data: &[u8]) -> Fr {
-        ZkHasher::digest(&[b2p(data)])
+        b2p(data)
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


thomaslavaur:
LGTM. Just one question: you modified the MMR lib. Is it use only for voucher? Or for now it's not used (in this PR) ? Don't forget that nodes maintain the MMR but they still need to compute the root of a Merkle tree of depth 32 for vouchers. (using the MMR peaks as internal node and hashing the rest where we can use precomputed roots of empty trees of various depth)

thomaslavaur (approved):

youngjoon-lee:
> LGTM. Just one question: you modified the MMR lib. Is it use only for voucher? Or for now it's not used (in this PR) ? Don't forget that nodes maintain the MMR but they still need to compute the root of a Merkle tree of depth 32 for vouchers. (using the MMR peaks as internal node and hashing the rest where we can use precomputed roots of empty trees of various depth)

Yeah, MMR is not used yet anywhere. We will begin using it since the PR #2504.

PR #2502, opened 2026-04-09
fix(ledger): read storage gas value from settings

pradovic:

The deployment config value for storage gas is overriden by `from_utxos` fn. This PR reads the value from config and updated `from_utxos` to demand providing a value in the signature. This should also make unit tests pass.

Yes

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


davidrusu (approved):

ledger/src/cryptarchia/mod.rs
```diff
@@ -544,15 +544,26 @@ impl LedgerState {
         config: &Config,
         epoch_nonce: Fr,
     ) -> Result<Self, LedgerError<Id>> {
+        let initial_storage_gas_price = tx.mantle_tx().storage_gas_price;
```
youngjoon-lee, line 546: Do we also need to read `initial_execution_gas_price`?

ledger/src/cryptarchia/mod.rs
```diff
@@ -1286,7 +1297,7 @@ pub mod tests {
         let output_note2 = Note::new(3000, output_note2_sk.to_public_key());
 
         let locked_notes = LockedNotes::new();
-        let ledger_state = LedgerState::from_utxos([input_utxo], &config(), Fr::ZERO);
+        let ledger_state = LedgerState::from_utxos([input_utxo], &config(), Fr::ZERO, 1.into());
```
youngjoon-lee: I'm curious why we use `1` in the tests in this PR.

youngjoon-lee (commented):
thank you. I left two questions.

ledger/src/cryptarchia/mod.rs
pradovic: I left it as is, at 0 still hard-coded. Should I wire it up as well?

ledger/src/cryptarchia/mod.rs
pradovic: The unit test were implemented together when the 1 was set to the gas price. So they still pass with value 1 and test desired functionality. We can also change it to 0, and modify the tests as you did in your PR?

pradovic: Ah sorry, I did not see that PR was merged, I will update from master and use the same values as your PR.

ledger/src/cryptarchia/mod.rs
youngjoon-lee: I think we should read `initial_execution_gas_price` as well. Otherwise, it's overriden by the constant (0). I was confused when I wrote #2486, sorry. 

So, it's clear that now the constants are only used when validating genesis tx: https://github.com/logos-blockchain/logos-blockchain/blob/f18a40699120b3f6c334490eac7438246b45da6b/core/src/mantle/genesis_tx.rs#L57-L67

Thank you.

ledger/src/cryptarchia/mod.rs
youngjoon-lee: I'm fine with using any value that makes tests green. TBH, I'm not sure which value we have to use for each test. I will go through tests tomorrow. You can merge this without waiting for me :)

pradovic: Fixed. I merged the master (including your PR), and I used zero same as on master. I think that should pass as you modified the tests to pass with the constant. Let's see. I needed to update from master anyway or I would break the test.

ledger/src/cryptarchia/mod.rs
pradovic: Make sense, I will add this as well, thanks!

pradovic: I have added execution price as well in the same fashion: https://github.com/logos-blockchain/logos-blockchain/pull/2502/changes/78ae410964ba4563da3f6404cc7b3aad9c63e680. I think there are more params there and we could abstract it nicely in a separate PR. Does that makes sense?

pradovic: @youngjoon-lee But actually if they are validated against constant why do we even set them as config values?

youngjoon-lee: I’m also thinking about it. I feel that the implementation needs to be substantially simplified. For example, we could remove those gas prices from the deployment config. We need to discuss with Thomas. 
I think we can merge this if it is urgent for Devnet. If not, i think is’s better to jump to the ultimate way.

pradovic: I think the current devent would have hard-coded values form your PR. So your PR needs to be on devnet, and we could drop this one for the ultimate way.

danielSanchezQ: They are taken from genesis now right @pradovic @youngjoon-lee ? If they are taken from genesis then yes we should remove from config. Otherwise we have to keep it (?)

pradovic: In any way, they are now 0 at devnet, as 1490 is merged. I think I can close this one then, and we can decide and fix on configuration the separately?

pradovic:
Master is already using constants. Genesis tx gas prices need to match the constants at the moment, that is expected behavior.

PR #2503, opened 2026-04-09
test(cucumber): use fee account

andrussal:

This PR makes the e2e transaction tests handle gas properly.

It adds an optional scenario-level sponsored fee account, so the amounts written in the scenarios stay the “real” amounts being sent around, while fees can be covered separately when a scenario opts into sponsorship. It also adds explicit settlement waits and exact final balance checks in the cases where we really care about the final wallet state, plus a consistency check that tracked wallet fees match what the sponsored fee account actually spent.

The point is to keep these tests passing even if gas costs change, without having to keep retuning the scenario amounts or making the assertions weaker.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


hansieodendaal (commented):
Nice idea!

As discussed, let's make sponsored fees opt-in, using a central account, e.g.
```
And we have a sponsored genesis fee account with <number_of_tokens> tokens of <token_value> value each
```

We should still keep track of all fees paid for each wallet, being it from the wallet itself or the sponsored fee wallet.

Additional change into wallets due to sponsored fee tokens being broken up when used for fee payment should be tracked as well per wallet.

tests/src/cucumber/steps/manual_nodes/steps.rs
```diff
@@ -114,6 +114,30 @@ fn step_cluster_has_wallet_resources(world: &mut CucumberWorld, step: &Step) ->
...
+fn step_sponsored_genesis_fee_account(
+    world: &mut CucumberWorld,
+    token_count: usize,
+    token_value: u64,
+) -> StepResult {
+    if token_count == 0 {
+        return Err(StepError::InvalidArgument {
+            message: "Sponsored genesis fee account token_count must be greater than 0".to_owned(),
+        });
+    }
```
hansieodendaal:
```suggestion
    let token_count = non_zero!("genesis fee token count", token_count)?.get()
```

tests/src/cucumber/steps/manual_nodes/steps.rs
```diff
@@ -114,6 +114,30 @@ fn step_cluster_has_wallet_resources(world: &mut CucumberWorld, step: &Step) ->
...
+    if token_value == 0 {
+        return Err(StepError::InvalidArgument {
+            message: "Sponsored genesis fee account token_value must be greater than 0".to_owned(),
+        });
+    }
```
hansieodendaal:
```suggestion
    let token_value = non_zero!("genesis fee token value", token_value )?.get()
```

tests/src/cucumber/world.rs
```diff
@@ -300,13 +313,22 @@ impl Debug for CucumberWorld {
                 &format!("{}", self.faucet_task_handles.as_ref().map_or(0, Vec::len)),
             )
             .field("wallet_accounts", &self.wallet_accounts.len())
+            .field(
```
hansieodendaal: These `self.fee_state` members could be printed together in a single condensed line (the formatter will print it in json type and waste screen real estate). We could have one special print for `&self.fee_state` to include all three debug prints in a single line.

tests/src/cucumber/steps/manual_transactions/utils.rs
```diff
@@ -166,6 +134,481 @@ pub async fn create_and_submit_transaction(
...
+
+    Ok(())
+}
+
+/// Waits for specified transactions to be included in the blockchain.
+pub async fn wait_for_transactions_inclusion(
+    client: &NodeHttpClient,
+    tx_hashes: &[TxHash],
+    timeout: Duration,
+) -> Result<(), StepError> {
```
hansieodendaal: Suggestion: It might be useful to specify an optional blocks deep parameter, like a user specified security parameter (much less than the LIB value); only acknowledge if that amount of blocks deep. It makes for flexible scenarios.

Wdyt?

tests/src/cucumber/steps/manual_transactions/utils.rs
```diff
@@ -166,6 +134,481 @@ pub async fn create_and_submit_transaction(
...
+    }
+}
+
+/// Waits for wallet balance to match the exact settled value.
+pub async fn wait_for_exact_settled_wallet_balance(
+    world: &mut CucumberWorld,
+    step_value: &str,
+    wallet_name: &str,
+    nominal_token_value: u64,
+    time_out_seconds: u64,
```
hansieodendaal, line 448: ditto _optional blocks deep parameter_

hansieodendaal (commented):
Really nice! Some questions and a couple of small comments/suggestions.

I could not pick up if we account for sponsored fee change added to wallet balances?

To which node is the sponsored fee wallet anchored?

tests/src/cucumber/steps/manual_transactions/utils.rs
andrussal: > Suggestion: It might be useful to specify an optional blocks deep parameter, like a user specified security parameter (much less than the LIB value); only acknowledge if that amount of blocks deep. It makes for flexible scenarios.
> 
> Wdyt?

I think the idea itself is good, but I wouldn’t put it into wait_for_transactions_inclusion itself. To me that helper should stay narrowly scoped to waiting for the transaction’s first appearance on chain. If we want a “blocks deep” condition, I think that belongs at a higher caller/policy level.

tests/src/cucumber/steps/manual_nodes/steps.rs
andrussal: Good point, but I make the field itself `NonZero`.

hansieodendaal (approved):
Looking real good!

Just two remarks to address before merging please:
- please prohibit fee and fork steps to be used together in the same scenario, as fork scenarios are very specific and will result in double spends when the forks join;
- please unify UTXO scanning in `scenario_fee_account_state` with the previous step `update_wallet_balance_all_user_wallets`.

PR #2504, opened 2026-04-09
feat(ledger): DRAFT: replace DynamicMerkleTree with MMR for voucher management

youngjoon-lee:
### Summary

Replaced `DynamicMerkleTree` with `MerkleMountainRange` (MMR) for voucher commitments.

Unlike `DynamicMerkleTree` which holds all leaves and internal nodes in memory, MMR doesn't keep leaves. It holds only peaks, achieving at most `33 * Fr = 1KiB` memory (at most 33 peaks).
Since each block has a voucher, it's much better to use MMR to minimize memory usage.

### Changes

I made changes in `chain-service`, `chain-leader-service`, `wallet-service`, and `ledger`. 

I highly recommend reviewing 3 commits separately. I grouped them into a single PR because they’re most meaningful together. Instead of writing explanations here, I put detailed descriptions in each commit.
You can also review all of them at once, if you want. I tried my best to make diffs easy to read.

These changes are tested by unit tests added in this PR. Also, I confirmed that the e2e test for LeaderClaim (added in #2500) passes.

Before diving into code, please read the design choice below.

### Design choice: Chain service owns Merkle paths

MMR doesn't keep leaves (voucher commitments). Instead it only keeps peaks. In other words, it's **not always possible** to compute a Merkle path of a certain leaf. A Merkle path can be computed initially when the leaf is added to MMR. After that, the Merkle path must be incrementally updated as new leaves are added to MMR, as implemented in #2469.

Therefore, we have to maintain Merkle paths that we're interested in, **somewhere** outside of MMR, and incrementally update them.

At the beginning, I tried to put this responsibility to the wallet service, because it is already tracking voucher secrets / commitments / nullifiers that the node owns. 
However, the issue was that MMR insertion is done by chain service (on epoch transition). I tried to make chain service fetch the tracked paths from wallet service, update paths via MMR insertions, and return the updated paths back to the wallet service. But it causes cyclic dependency because wallet service is already dependent on chain service.

Therefore, I decided to maintain Merkle paths in the chain service. Voucher secrets are still maintained by wallet service. When wallet service wants to build LeaderClaim operations, it fetches Merkle paths from chain service.

Yes, the responsibilities are split between two services. But, I believe that this is the cleanest flow among other architectures I've considered. Wallet service manages voucher secrets that require KMS. Chain service manages only Merkle paths of voucher commitments. 
Please see the following sequence diagram. 

```mermaid
sequenceDiagram
    participant Leader as Chain Leader Service
    participant Chain as Chain Service
    participant Ledger as Ledger
    participant Wallet as Wallet Service

    Note over Leader: Node wins slot
    Leader->>Chain: ApplyBlock(block, locally_proposed=true)
    Chain->>Chain: Extract voucher_cm from block and Add it to tracked_voucher_paths
    Chain->>Ledger: try_apply_header (updates MMR & paths)

    Note over Wallet: Wallet needs to claim voucher
    Wallet->>Chain: GetVoucherMerklePath(voucher_cm)
    Chain-->>Wallet: MerklePath
    Wallet->>Wallet: Build PoC

    Note over Chain: LIB advances (blocks become immutable)
    Chain-->>Wallet: LibUpdate
    Wallet->>Wallet: Prune claimed vouchers
    Wallet->>Chain: PruneTrackedVoucherPaths(voucher_cms)
```

### Note

Inserting new elements to MMR with updating existing Merkle paths (implemented in #2469) requires much computation than `DynamicMerkleTree` (which doesn't compute Merkle paths incrementally). It is a tradeoff between memory and CPU. Tomorrow, I'm going to benchmark both.

I believe so.

@youngjoon-lee 
spec: @thomaslavaur 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.


thomaslavaur:
from the PR description:
> 33 * Fr = 1KiB memory (at most 33 peaks)
I think it's 32 Fr elements because max is when the tree is almost full (1 less element) that is 31 to 0 depth

From:
> Instead it only keeps peaks. In other words, it's not always possible to compute a Merkle path of a certain leaf.
I would add "only using the MMR peaks" to make it clearer

I'm not sure to understand why chain-service is maintaining the paths. In this case, is the chain-service aware of what voucher the wallet has? Because a node shouldn't maintain the Merkle path of every voucher but only the one where the wallet knows the secret voucher. I think I'll understand when reading the code.

core/src/proofs/merkle.rs
```diff
@@ -1,6 +1,21 @@
+use lb_groth16::Fr;
 use lb_utxotree::{MerkleNode, MerklePath};
 
-/// Converts a [`MerklePath`] to the witness format expected by the circuit.
+/// Converts a [`lb_mmr::MerklePath`] to the witness format expected by
+/// the circuit: siblings in bottom-top order, selectors reversed.
+pub fn mmr_path_to_witness(path: &lb_mmr::MerklePath) -> (Vec<Fr>, Vec<bool>) {
+    let items = path.siblings.clone();
+    // Selectors: true if sibling is on the left (i.e. leaf is a right child).
+    // Circuit expects reversed order.
```
thomaslavaur, line 9: I think this can be done in a better way: the selectors are just the list of leaf's indice in big endian bits

thomaslavaur (approved):
It looks pretty clean to me. I added a small comment that can make code and review slightly better but feel free to ignore it.

services/chain/chain-service/src/lib.rs
```diff
@@ -214,6 +226,7 @@ pub struct Cryptarchia {
     pub ledger: lb_ledger::Ledger<HeaderId>,
     pub consensus: lb_cryptarchia_engine::Cryptarchia<HeaderId>,
     pub genesis_id: HeaderId,
+    pub tracked_voucher_paths: lb_ledger::mantle::leader::TrackedVoucherPaths,
```
youngjoon-lee, line 229: @thomaslavaur @danielSanchezQ 

SELF-COMMENT: This is not a right place to put tracked voucher paths. It must be maintained inside `lb_ledger::Ledger`. Otherwise, we cannot handle chain reorgs at epoch boundary.

~~Also, I have an optimization idea for MMR. Instead of tracking paths outside of MMR, I think we can have a hybrid of MMR and MerkleTree. In short, MMR can keep **part** of leaves, which a caller wants to keep track of. Then, we don't need to track paths outside and update them incrementally.~~

I will make this PR as draft, and open other PRs to make this better.

PR #2522, opened 2026-04-10
fix(blend): address self audit findings

ntn-x2:

PR on top of https://github.com/logos-blockchain/logos-blockchain/pull/2514.

It introduces the following fixes, ordered by importance:

* Verify that the received message has at least one encapsulation layer, without panicking otherwise (with test). This is a fix that does not change the current code structure. Ideally we enforce this at the type level after verifying the minimum number of headers is matched. But for now, this will do it.
* Close connections with spammy peers belonging to the old session (with tests). The swarm is not notified about it, so it won't try to dial new peers, nor the peer will be banned. Since old session won't be accepting/establishing new connections, the node is safe going forward until the STP elapses.
* Verify that the version of a received public header matches the expected one (with test)
* Do not panic in core-core connection handler in case of unexpected shutdowns, but attempt to simply close the substreams
* Fix the return value when updating the state of a negotiated connection in case it belongs to the old session (now returning `None` instead of `Some`). Minor issue, but could create confusion in the future.
* Change the error kind captured when self-decapsulating

Yes.

@ntn-x2 

N/A

No.

### DIFF
--- a/blend/message/src/encap/encapsulated.rs
+++ b/blend/message/src/encap/encapsulated.rs
@@ -124,17 +124,32 @@ pub struct EncapsulatedPart {
 }
 
 impl EncapsulatedPart {
-    /// Initializes the encapsulated part as preparation for actual
-    /// encapsulations.
-    pub(super) fn initialize(
+    #[cfg(test)]
+    #[must_use]
+    pub fn new_unchecked(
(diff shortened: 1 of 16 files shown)


youngjoon-lee (approved):
The changes you listed look reasonable and necessary. Thank you.

PR #2523, opened 2026-04-11
feat(api): add raw signing REST API and e2e test for atomic deposit

youngjoon-lee:

- Added `POST /sign/tx/ed25519` and `POST /sign/tx/zk`.
  - For signing `MantleTx` using keys in KMS. These don't involve any funding mechanism, only sign the tx hash.
- Added an e2e test in zone-sdk for atomic deposit+inscription.

yes

[Atomic deposit suggestion in the spec](https://www.notion.so/nomos-tech/RFC-Make-Ledger-Transaction-an-Operation-31e261aa09df80bc9e02ea4e9affc082?source=copy_link#333261aa09df80b1baaddc41a08107ec)

@youngjoon-lee 
spec: @thomaslavaur 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nodes/api-common/src/bodies/wallet.rs
+++ b/nodes/api-common/src/bodies/wallet.rs
@@ -85,3 +85,33 @@ pub mod transfer_funds {
         }
     }
 }
+
+pub mod sign {
+    use lb_core::mantle::TxHash;
+    use lb_key_management_system_keys::keys::{
+        Ed25519Key, ZkPublicKey, ZkSignature, secured_key::SecuredKey,
+    };
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Serialize, Deserialize)]
+    pub struct WalletSignTxEd25519RequestBody {
+        pub tx_hash: TxHash,
+        pub pk: <Ed25519Key as SecuredKey>::PublicKey,
+    }
+
+    #[derive(Serialize, Deserialize)]
+    pub struct WalletSignTxEd25519ResponseBody {
+        pub sig: <Ed25519Key as SecuredKey>::Signature,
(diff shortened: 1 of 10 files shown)


pradovic (approved):
LGTM! Just one question: is it safe to expose this kind of public API? I assume the answer is yes in our setup, but wanted to double-check the intended trust/security model here.

youngjoon-lee:
> LGTM! Just one question: is it safe to expose this kind of public API? I assume the answer is yes in our setup, but wanted to double-check the intended trust/security model here.

Yeah, good point. Users must set authentication to the API. Also, users anyway need to their nodes since KMS and note tracking (that wallet service does) are built in the node. But, it is not so secure in general. 

I think the ultimate way would be extracting KMS and note tracking from node to zone-sdk (or a separate library), as a complete client-side wallet. But, I'm also fine with the current architecture for now, assuming that we will encourage end users to run their nodes, of course, with security configs. 

Actually we already have the [`POST /wallet/transactions/transfer-funds`](https://github.com/logos-blockchain/logos-blockchain/blob/f04c8496ae962fd10ad1f73d65ae67eaac54616f/nodes/node/binary/src/api/handlers.rs#L753) which funds and **signs** the tx built by an end user. But I agree with your concern. I will open a discussion soon to hear opinions from the team.

youngjoon-lee:
> I will open a discussion soon to hear opinions from the team.

Oh, damn. I merged this by mistake before opening a discussion https://discord.com/channels/973324189794697286/1494211933333622835. Let’s talk there. I can revert this if necessary.

PR #2528, opened 2026-04-13
test(mmr): benchmark MMR vs DynamicMerkleTree for voucher mgmt

youngjoon-lee:

Added benchmarks to compare MMR vs `DynamicMerkleTree` (the "Tree").

### Background
Before talking about benchmark, let me give you some background.

First of all, please keep in mind that voucher insertion (to MMR or Tree) happens only at each epoch transition.

Tree keeps all leaves (vouchers) and internal nodes. Even if users claim rewards, vouchers are not removed from Tree because claiming comes with voucher nullifiers which cannot be directly connected to voucher commitments. That is, Tree grows to the max capacity (2^32 vouchers). Then, the max size is 756GB. 

Unlike Tree, MMR doesn't keep leaves. It only holds peaks. Of course, it doesn't have "remove" operation, like Tree. Therefore, its max memory size is 1.6KB which is way smaller than Tree. 

But, MMR is not a silver bullet. Since it doesn't keep leaves, we cannot compute Merkle path of a certain leaf. So, I implemented incremental path maintenance #2469 and integrated it to the system #2504. 

Each Merkle path takes 1KB of memory. And, updating them incrementally is not for free.

So, in these benchmarks, I measured the overhead of incremental path maintenance in MMR, comparing the result of Tree.

### Benchmark Results

You can see what these tests are from the code. Here, I will give you only summary. 
```
voucher                        fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ mmr_1st_epoch               17.46 s       │ 17.47 s       │ 17.46 s       │ 17.47 s       │ 3       │ 3
├─ mmr_11th_epoch              20.96 s       │ 20.98 s       │ 20.96 s       │ 20.97 s       │ 3       │ 3
├─ tree_1st_epoch              17.37 s       │ 17.37 s       │ 17.37 s       │ 17.37 s       │ 3       │ 3
├─ tree_11th_epoch             17.37 s       │ 17.38 s       │ 17.37 s       │ 17.37 s       │ 3       │ 3
```
- On Raspberry Pi 5. (14s slower than MacBook Pro M1 Max, in general)
- k = 2160, f = 1/30, slot = 1s
- epoch length = 648,000 secs = 21,600 blocks
- 21,600 vouchers are inserted to MMR and Tree at epoch transition.
- 50 nodes in the network. So, 21600/50 = 432 vouchers are owned by a single node, which means that 432 voucher paths must be added to the tracked path set after epoch transition.

First, compare `mmr_1st_epoch` vs `tree_1st_epoch`. The result is almost the same because insertion on MMR has almost the same complexity as Tree. But, we have to track/update paths of newly added vouchers if they are owned by us. That makes the 0.1s difference.

Compare `mmr_11th_epoch` vs `tree_11th_epoch`. 3s of difference caused by path updates is significant in this case. During 10 epoch, we gathered 4,320 voucher paths that we own. We never claimed them. When transitioning to the epoch 11, we had to update those paths (+ 432 newly added our voucher paths) during new 21,600 voucher insertions. 

### Suggestions

It is clear that `DynamicMerkleTree` should be replaced with something else, because it grows up to 756GB and no elements are removed. MMR is a good alternative in terms of memory usage (1.6KB max).

But MMR causes additional computation/memory overhead for incremental Merkle path maintenance.
- 1KB for a Merkle path
- A non-negligible delay if we don't claim rewards periodically.

The delay might not be so critical in production because there would be more nodes in the network (which means that each node tracks less number of voucher paths). And, the delay happens only at epoch transition. But still, the overhead is not negligible. 

For now, MMR is the only one option we found to replace `DynamicMerkleTree`. Until we have any better idea, I think we should encourage users to claim rewards periodically. It would be the best if they claim all rewards before a new epoch starts.

### Another Suggestion

In both MMR and Tree, bulk insertions at epoch transition is heavy (17~ secs). 

Instead of keeping pending vouchers in `Vec` and flushing them to MMR or Tree at once at epoch transition, we may be able to maintain a clone of MMR or Tree and insert a new voucher as soon as a new block is processed. I think it works because they are `rpds`s and because the order of voucher insertions is the same of the order of blocks which is deterministic across all nodes. Then, each block processing time will increase, especially for MMR if users are lazy to claim rewards. But this would allow us to distribute the overhead of the epoch transition across all block processings. It may be worth trying because block interval is anyway long. I'm not sure yet but we can think of it.

yes

@youngjoon-lee 
spec: @thomaslavaur 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.


danielSanchezQ:
I had the feeling this would happen when we had to start tracking the paths. For now as you mentioned we have no other option, at least is not that much slower while we keep the savings on the memory side (which are more important imo).

thomaslavaur:
We have other ways to improve. The path can be updated once every X block and batch update the path. This is reducing the computation as you have less hash with empty nodes to do

thomaslavaur (approved):

PR #2538, opened 2026-04-14
fix(chain-leader): disable self-applying blocks to avoid early session transition (temp)

youngjoon-lee:

We've seen that a block proposing node transitions to a new Blend session much earlier than other nodes who receive the proposal after a few seconds, even though we are already bypassing Blend if a block is the first block in a new Blend session.

So, this PR disables self-applying blocks in the chain-leader service to completely eliminate possibilities of the early session transition. Of course, this may be not a right direction for the chain-leader service because it can cause some self-forking situations, but they are not so critical in terms of the consensus correctness. Anyway, this change is a temporary workaround until we replace sessions with epochs in Blend.

yes

@youngjoon-lee 
spec: @davidrusu @madxor 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.


danielSanchezQ (approved):

ntn-x2:
Before merging this, I'd like to compare it with the current baseline on DST nodes before deciding how to go ahead. Thanks @youngjoon-lee for opening this. I can take over from here :)

ntn-x2:
Quick update: we deployed the current HEAD of this branch (despite some warnings due to unused fields) on the DST nodes this morning. We will let it run for 24h, but from some local tests it seems all nodes change sessions at the same time, which is what we want, and as far as I know this is the only way to go about it without introducing a lot of complicated logic that will be thrown away soon. So if the run looks good tomorrow, I package this up, ask for a re-review, then we are good to go with this.

youngjoon-lee:
> Quick update: ...

Thank you! @ntn-x2

ntn-x2:
@danielSanchezQ I re-requested your review. If you give green light again, I will go ahead and merge this.

danielSanchezQ (approved):

PR #2541, opened 2026-04-14
feat(NoteId): Implement NoteId Uniqueness RFC

thomaslavaur:

This PR implements the [NoteId Uniqueness RFC](<https://www.notion.so/nomos-tech/RFC-Enforce-NoteId-uniqueness-335261aa09df807b9fe3c9bb9bd2c6db>). 

This makes each operation that consume or output value to create or consumed their notes directly without impacting the Mantle Transaction balance. This change also require to modify the NoteId derivation that was linked to the transfer hash. The NoteId is now derived from an `op_id` that must be unique for every Operation that outputs note. We uniformized the derivation of `op_id` from the operation payload which guarantee that the `op_id` is unique as long as the payload is unique. For this reason, the `TRANSFER` inputs is required to be at least of len 1 and the `CHANNEL_WITHDRAW` includes a withdraw_nonce that is increased by one for the channel after each withdraw.

I think so

specs and code: @thomaslavaur 

Yes

Minor change in the Service Reward Distribution Protocol that wasn't updated to reflect the change of NoteId derivation

## Checklist

* [x] 2. Description added.


core/src/mantle/ops/channel/withdraw.rs
```diff
@@ -1,9 +1,36 @@
...
+        let withdraw_id = self.id();
+        self.outputs
+            .iter()
+            .enumerate()
+            .map(move |(index, note)| Utxo {
+                op_id: withdraw_id,
+                output_index: index,
+                note: *note,
+            })
+    }
```
danielSanchezQ: I've seen this pattern. I think at this point we should abstract outputs in to a proper `Outputs` (`pub struct Outputs(Vec<Note>)`) wrapper whic takes the op_id as input and build the Utxos. So we can reuse it everywhere.

core/src/mantle/ops/channel/withdraw.rs
```diff
@@ -1,9 +1,36 @@
...
+
+impl ChannelWithdrawOp {
+    #[must_use]
+    pub fn id(&self) -> Hash {
+        let encoded_bytes = encode_channel_withdraw(self);
+        let mut hasher = Hasher::new();
+        hasher.update(b"OPERATION_ID_V1");
+        hasher.update(encoded_bytes);
+        hasher.finalize().into()
+    }
```
danielSanchezQ: If this is included in ops now it maybe be better to take it out to some trait that the ops implement (?)

core/src/mantle/gas.rs
```diff
@@ -170,13 +170,13 @@ pub trait GasConstants {
...
-    const LEADER_CLAIM: Gas = Gas(1150);
+    const TRANSFER: Gas = Gas(590);
+    const CHANNEL_INSCRIBE: Gas = Gas(56);
+    const CHANNEL_SET_KEYS: Gas = Gas(56);
+    const CHANNEL_DEPOSIT: Gas = Gas(590);
+    const CHANNEL_WITHDRAW: Gas = Gas(56);
+    const SDP_DECLARE: Gas = Gas(646);
+    const SDP_WITHDRAW: Gas = Gas(590);
+    const SDP_ACTIVE: Gas = Gas(590);
+    const LEADER_CLAIM: Gas = Gas(580);
```
danielSanchezQ, line 181: This should have been an independant PR. It is not related to the changes in this PR (?)

danielSanchezQ (commented):
Overall is a much better design imo. I think we should somehow abstract it a bit better behind some types and interfaces (left some comments). Otherwise is good to go, thanks!

core/src/mantle/gas.rs
thomaslavaur: this RFC was modifying the Gas cost of channel deposit and I didn't know what to do (either we keep old stuff but the RFC isn't fully implemented) or we go with RFC for now and the gas needs an update.

I think it's a good opportunity to update it (it's a small update of the constant price)

danielSanchezQ: Ok, since its attached to the rfc itself then its fine. Just looking at code it didn't seem related :+1: thanks!

core/src/block/genesis.rs
```diff
@@ -624,18 +625,19 @@ impl GenesisBlockBuilder<WithAll> {
...
+                .collect();
         let n = ops.len();
         let signed_tx = SignedMantleTx::new_unverified(
             MantleTx {
                 ops,
                 execution_gas_price: GasPrice::new(0),
                 storage_gas_price: GasPrice::new(0),
             },
-            vec![OpProof::NoProof; n],
+            vec![OpProof::Ed25519Sig(Ed25519Signature::from_bytes(&[0u8; 64])); n],
```
danielSanchezQ, line 640: Ok, so we build an empty signature. This is a bug from my previous PR, thanks for the fix.

core/src/mantle/ops/channel/withdraw.rs
```diff
@@ -1,9 +1,25 @@
...
 pub struct ChannelWithdrawOp {
     pub channel_id: ChannelId,
-    pub amount: Value,
+    pub outputs: Outputs,
+    pub withdraw_nonce: u32,
+}
+
+impl OpId for ChannelWithdrawOp {
+    fn op_id(&self) -> Hash {
+        let mut encoded_bytes: Vec<u8> = b"OPERATION_ID_V1".into();
```
danielSanchezQ: This should be a const value reused for everyone.

core/src/mantle/ops/transfer.rs
```diff
@@ -38,26 +39,13 @@ impl TransferOp {
...
-            .iter()
-            .enumerate()
-            .map(move |(index, note)| Utxo {
-                transfer_hash,
-                output_index: index,
-                note: *note,
-            })
+impl OpId for TransferOp {
+    fn op_id(&self) -> Hash {
+        let mut encoded_bytes: Vec<u8> = b"OPERATION_ID_V1".into();
```
danielSanchezQ: yes, we should reuse a variable otherwise we may accidentally miss a change in one of them in the future.

danielSanchezQ (approved):
A nitpick, otherwise good to go imo. @youngjoon-lee should check this anyway if possible.

core/src/block/genesis.rs
thomaslavaur: Yes this PR remove NoProof since no Operation as None anymore

core/src/block/genesis.rs
```diff
@@ -624,18 +625,19 @@ impl GenesisBlockBuilder<WithAll> {
                 },
         } = self;
         // Order is important to keep here
-        let ops: Vec<Op> = std::iter::once(Op::Transfer(TransferOp::new(vec![], notes)))
-            .chain(std::iter::once(Op::ChannelInscribe(inscription)))
-            .chain(sdp_declarations.into_iter().map(Op::SDPDeclare))
-            .collect();
+        let ops: Vec<Op> =
+            std::iter::once(Op::Transfer(TransferOp::new(vec![], Outputs::new(notes))))
```
youngjoon-lee, line 629: So, is this the only one exception where we accept a `TransferOp` with empty `inputs`, even though the RFC changes `TransferOp` to have at least one input? I think it's fine because we build `SignedMantleTx` with `new_unverified` function, IIRC.

core/src/mantle/ops/channel/deposit.rs
```diff
@@ -1,10 +1,10 @@
 use serde::{Deserialize, Serialize};
 
-use crate::mantle::{Value, ops::channel::ChannelId};
+use crate::mantle::{NoteId, ops::channel::ChannelId};
 
 #[derive(Clone, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
 pub struct DepositOp {
     pub channel_id: ChannelId,
-    pub amount: Value,
+    pub inputs: Vec<NoteId>,
```
youngjoon-lee, line 8: I left a [question](https://www.notion.so/RFC-Enforce-NoteId-uniqueness-335261aa09df807b9fe3c9bb9bd2c6db?d=345261aa09df80a0aa91001cd66a3cbb&source=copy_link#335261aa09df80869e75d3952564cb48) about this in the RFC.

core/src/mantle/ops/channel/withdraw.rs
```diff
@@ -1,9 +1,25 @@
...
+    pub outputs: Outputs,
+    pub withdraw_nonce: u32,
+}
+
+impl OpId for ChannelWithdrawOp {
+    fn op_id(&self) -> Hash {
+        let mut encoded_bytes = OPERATION_ID_V1.clone();
+        encoded_bytes.extend(encode_channel_withdraw(self));
+        Hasher::digest(&encoded_bytes).into()
+    }
```
youngjoon-lee, line 19: If all operations (that produce notes) use the same derivation logic (with the same `OPERATION_ID_V1` and an encoded byte array of operation), I think we can polish the `OpId` trait like:
```rust
trait OpId {
  fn op_id(&self) -> Hash {
    let mut encoded_bytes = OPERATION_ID_V1.clone();
    encoded_bytes.extend(self.op_bytes());
    Hasher::digest(&encoded_bytes).into()
  }

  fn op_bytes(&self) -> Vec<u8>;
}
```
Then, each operation can implement `OpId` like:
```rust
impl OpId for ChannelWithdrawOp {
  fn op_bytes(&self) -> Vec<u8> {
    encode_channel_withdraw(self)
  }
}
```
Then, we can avoid any mistake when implementing `op_id` function for each operation.

core/src/mantle/tx_builder.rs
```diff
@@ -156,24 +157,7 @@ impl MantleTxBuilder {
...
-            .map(|op| match op {
-                Op::ChannelDeposit(deposit) => -i128::from(deposit.amount),
-                Op::ChannelWithdraw(withdraw) => i128::from(withdraw.amount),
-                Op::LeaderClaim(_) => i128::from(self.context.leader_reward_amount),
-                // `Op::Transfer` is not handled here since `self.ledger_inputs` and
-                // `self.pending_transfer.outputs` already account for the balance changes from
-                // `Op::Transfer`s.
-                _ => 0,
-            })
-            .sum();
```
youngjoon-lee, line 174: Very good that we don't need this anymore.

ledger/src/mantle/channel.rs
```diff
@@ -175,13 +181,18 @@ impl Channels {
...
-                .balance
-                .checked_sub(op.amount)
-                .ok_or(Error::InsufficientFunds)?;
-            Ok(self)
+            if channel.withdraw_nonce == op.withdraw_nonce {
+                channel.balance = channel
+                    .balance
+                    .checked_sub(amount)
+                    .ok_or(Error::InsufficientFunds)?;
+                channel.withdraw_threshold += 1;
```
youngjoon-lee:
```suggestion
                channel.withdraw_nonce += 1;
```
I think, we have to update `withdraw_nonce` instead of `withdraw_threshold`. 

Also, it'd be great if we use `checked_add` because this update in Ledger is critical for consensus. If we intent the cyclic arithmetic (such as `u32::MAX + 1 == 0` which works only in `release` mode), we would refactor this in a different way, but I don't think that's the way we intent.

```rust
channel.withdraw_nonce = channel.withdraw_nonce.checked_add(1).ok_or(Error::WithdrawNonceOverflow);
```

tests/src/tests/mantle/channel.rs
```diff
@@ -1,3 +1,5 @@
+// TODO: this needs a refactoring with the new deposit
```
youngjoon-lee, line 1: I will do it soon.

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -686,24 +691,33 @@ async fn test_atomic_deposit_inscription() {
...
     let transfer = TransferOp {
         inputs: vec![note_id],
         outputs: if change > 0 {
-            vec![Note::new(change, pk)]
+            Outputs::new(vec![deposit_note, Note::new(change, pk)])
         } else {
-            vec![]
+            Outputs::new(vec![deposit_note])
         },
     };
```
youngjoon-lee, line 709: I'm trying to understand why this `transfer` operation is necessary. Is it for splitting the existing note, as I mentioned in the [comment](https://www.notion.so/RFC-Enforce-NoteId-uniqueness-335261aa09df807b9fe3c9bb9bd2c6db?d=345261aa09df80a0aa91001cd66a3cbb&source=copy_link#335261aa09df80869e75d3952564cb48) in RFC?

youngjoon-lee (commented):
Thank you so much for the changes across many files. I left some suggestions and questions before approving this.

services/wallet/src/lib.rs
```diff
@@ -574,6 +574,23 @@ where
...
+        note_ids: Vec<NoteId>,
+        kms: &KmsServiceApi<Kms, RuntimeServiceId>,
+        tx_builder: MantleTxBuilder,
+    ) -> Result<OpProof, WalletServiceError> {
+        let input_pks: Vec<ZkPublicKey> = tx_builder
+            .ledger_inputs()
+            .iter()
+            .filter(|&utxo| note_ids.contains(&utxo.id()))
+            .map(|utxo| utxo.note.pk)
+            .collect();
```
youngjoon-lee, line 588: This code uses input notes for Transfer ops in the tx builder, not inputs for the deposit operation. Actually, the tx builder itself needs to be refactored. But, let me think how to resolve this now without refactoring.

core/src/block/genesis.rs
thomaslavaur: Yes I had to modify it somewhere (to build with new_unverified). That's something we discussed with David and Daniel, everything in the genesis block can skip validation

core/src/mantle/ops/channel/withdraw.rs
thomaslavaur: Good idea! thanks

ledger/src/mantle/channel.rs
thomaslavaur: omg good spot you saved us

tests/src/tests/zone_sdk/e2e.rs
thomaslavaur: yes it's to get the change by splitting the note first. If there is no change needed we could build the tx without a transfer but the change was easier this way

PR #2542, opened 2026-04-14
fix(blend): use exponential back-off for retrials

ntn-x2:

Use exponential backoff after each failed attempt for both edge swarm when sending a message and core swarm when connecting to a peer.

Yes.

@ntn-x2 

Yes.

No.


services/blend/src/core/backends/libp2p/tests/redials.rs
```diff
@@ -23,14 +30,110 @@ async fn core_redial_same_peer() {
...
+    for _ in 0..3 {
+        dialing_swarm
+            .poll_next_until(|event| {
+                let SwarmEvent::OutgoingConnectionError { peer_id, .. } = event else {
+                    return false;
+                };
+                *peer_id == Some(random_peer_id)
+            })
+            .await;
+    }
```
youngjoon-lee, line 53: Don't we need to check if `ongoing_dials().is_empty()` in the loop, to check if the next try is delayed?

youngjoon-lee (approved):
Great! Well organized. Thank you. I left a simple question in the test.

services/blend/src/core/backends/libp2p/tests/redials.rs
ntn-x2: Yes good point. I added an assertion that the peer should not be in the map of ongoing dials before retrying.

PR #2543, opened 2026-04-14
feat: Genesis builder

danielSanchezQ:

This includes a small library to build Genesis blocks. It's the first step for a set of utilities that will be used to setup offline deployment settings.

It heavily uses the type-state pattern so its a bit cumbersome to ready. But overall its a simple approach.

@danielSanchezQ 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/core/src/block/genesis.rs
+++ b/core/src/block/genesis.rs
@@ -8,7 +8,7 @@ use crate::{
     mantle::{
         Op, OpProof, SignedMantleTx,
         genesis_tx::{self, GenesisTx},
-        ops::sdp::SDPDeclareOp,
+        ops::{sdp::SDPDeclareOp, transfer::TransferOp},
         tx::VerificationError,
         tx_builder::MantleTxBuilder,
     },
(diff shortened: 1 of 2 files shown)


bacv (approved):
This actually looks good. I think next step for applying this would be to update the chain services inital state to use `Block<GenesisTx>` instead of `StartingState::Genesis`.

core/src/block/genesis.rs
```diff
@@ -0,0 +1,965 @@
...
+
+/// Errors that can occur when building a genesis block via
+/// [`GenesisBlockBuilder`].
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    /// The op proofs supplied to [`SignedMantleTx`] failed verification.
+    #[error("Transaction verification failed: {0}")]
+    Verification(#[from] VerificationError),
+    /// The constructed transaction does not satisfy genesis transaction
+    /// invariants (e.g. non-zero gas price, missing transfer/inscription,
```
thomaslavaur, line 25: In fact genesis gas price should be 1 and are set to 0 for devnet only. Can we add a todo here?

thomaslavaur (approved):

core/src/block/genesis.rs
danielSanchezQ: I'll open an issue with this instead. But yes, good catch, thanks!

danielSanchezQ:
Unrelated CI errors. Merging.

PR #2544, opened 2026-04-14
fix(leader): avoid slot loss on oversized block proposals

### LINKED ISSUE #2505: Leader may lose slot when creating oversized block
Currently we have size limits inforced on block creation. But the leader does not handle these failures gracefully. The result is that the leader loses his slot.

This could be a problem if a large portion of leaders have the same mempool view and are all creating oversized blocks, this may lead to a stall in the network until the large tx's are handled somehow.

Block creation validates block size: https://github.com/logos-blockchain/logos-blockchain/blob/786a060e0a504cf08cf49425a58b46f0aaf126c3/core/src/block/mod.rs#L104-L110

But chain-leader ignores the Block::create errors and returns early leading to a lost slot: https://github.com/logos-blockchain/logos-blockchain/blob/786a060e0a504cf08cf49425a58b46f0aaf126c3/services/chain/chain-leader/src/lib.rs#L688-L695

PoulavBhowmick03:

Fixes slot loss on oversized block proposals for the leader
Closes #2505 

Yes

> [!IMPORTANT]  
> explained.  

Author: @davidrusu 
PR: @PoulavBhowmick03

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


services/chain/chain-leader/src/lib.rs
```diff
@@ -655,6 +684,19 @@ where
...
+                tracing::debug!(
+                    target: LOG_TARGET,
+                    ?tx_hash,
+                    tx_size,
+                    max = MAX_BLOCK_SIZE,
+                    "skipping individually oversized tx during block assembly"
+                );
+                invalid_tx_hashes.push(tx_hash);
+                continue;
+            }
```
danielSanchezQ, line 669: Uhm, I'm not sure this is the way of doing this. Tx validation has their own rules regarding size and it is not related to the block itself. cc @youngjoon-lee

core/src/mantle/select.rs
```diff
@@ -31,7 +31,58 @@ impl<const SIZE: usize, Tx: Transaction + Send> TxSelect for FillSize<SIZE, Tx>
...
         S: Stream<Item = Self::Tx> + Send + 'i,
     {
-        let stream = utils::select::select_from_till_fill_size_stream::<SIZE, Self::Tx>(|_| 1, txs);
+        let stream = utils::select::select_from_till_fill_size_stream::<SIZE, Self::Tx>(
+            |tx: &Self::Tx| tx.storage_size(),
+            txs,
+        );
         Box::pin(stream)
     }
 }
```
danielSanchezQ, line 40: Nice, this was needed. There is one thing that is missing. And it is that atm we use a const `M16` (16Megabytes) block size, which is what should be replace by the `MAX_BLOCK_SIZE` (it is when we compose the final node binary).

services/chain/chain-leader/src/lib.rs
```diff
@@ -68,6 +68,35 @@ const SERVICE_ID: &str = "ChainLeader";
...
+            "selected transactions exceeded block size, trimming tail"
+        );
+        let removed_tx = txs
+            .pop()
+            .expect("txs is not empty while trimming oversized block");
+        total_size -= removed_tx.storage_size();
+    }
+
+    Block::create(parent, slot, proof, txs, signing_key)
+}
```
danielSanchezQ: I think this is not necessary at all. Block will fill to the proper size if we use the `MAX_BLOCK_SIZE` as the size for `FillSize` (?)

core/src/mantle/select.rs
PoulavBhowmick03: yeah i did change that here
https://github.com/logos-blockchain/logos-blockchain/pull/2544/changes#diff-be8ecaf52b133cf0c2735d8e4ebb93aadd4370d2229493041e3700f90ff4e409L13
There is a redundant `MB16` at `binary/src/lib.rs` line 64 which i am getting rid of.

services/chain/chain-leader/src/lib.rs
PoulavBhowmick03: yeah i introduced this before working on `FillSize` as `FillSize` was measuring |_| 1 instead of bytes. this is redundant with the size validation in place already

core/src/mantle/select.rs
danielSanchezQ: Ah nice! Somehow missed that in the changelog! Thanks!

services/chain/chain-leader/src/lib.rs
PoulavBhowmick03: now that `build_block_with_tail_trim` is removed, tx validation via `try_apply_contents` only checks fees, not the storage against MAX_BLOCK_SIZE. Without this guard, a tx greater than max size in the mempool would cause FillSize's `take_while` stream to terminate immediately on every proposal, producing an empty block. would love to know your thoughts about this

danielSanchezQ: Yes. We have a smarter pool solution in the pipeline which will mean that this should not really happen. It will impact block construction as well. I would say for now it is enough :thinking:

PoulavBhowmick03: alright. made the changes. do take a look :)

danielSanchezQ: I think we could split the incoming stream into valid and invalid ones. Also we should work over stream directly if possible. I know we are collecting things already, but maybe we could take the opportunity to do it properly. wdyt?

PoulavBhowmick03: the incoming stream currently is broken into two vecs of valid and invalid. do u mean to change that to stream? i think that makes sense as we are converting it into stream later anyways. on the full stream pipeline, I am not sure because I could see that the ledger validation is stateful at line 671. so the loop seems to be the correct abstraction

danielSanchezQ: yeah, I think it is fine now.

danielSanchezQ (approved):

danielSanchezQ:
Branch needs rebasing/merging @PoulavBhowmick03, then we wait for greener CI and I'll merge. Thanks!

PoulavBhowmick03:
@danielSanchezQ fixed the merge conflict, can you please run the CI check. thanks!

danielSanchezQ:
@PoulavBhowmick03 the `build_block_with_tail_trim` tests is failing (?)

core/src/block/mod.rs
```diff
@@ -103,12 +103,28 @@ impl<Tx> Block<Tx> {
...
-            });
+        for tx in &transactions {
+            let tx_size = tx.storage_size();
+            if tx_size > MAX_BLOCK_SIZE {
+                return Err(Error::ContentTooBig {
+                    count: tx_size,
+                    max: MAX_BLOCK_SIZE,
+                });
+            }
+        }
```
davidrusu: This is checking if any one tx on it's own is bigger than the max block size? Why do we need this if we are checking the accumulated size below?

core/src/block/mod.rs
```diff
@@ -103,12 +103,28 @@ impl<Tx> Block<Tx> {
...
+            let size = tx.storage_size();
+            if acc_size + size <= MAX_BLOCK_SIZE {
+                acc_size += size;
+                false
+            } else {
+                true
+            }
+        });
+        if let Some(point) = trim_point {
+            transactions.truncate(point);
```
davidrusu: Silently truncating the transactions to fit inside the block is not a clear API, the caller would expect a block formed with all the transactions that were passed in.

services/chain/chain-leader/src/lib.rs
```diff
@@ -594,6 +594,19 @@ where
 
         while let Some(tx) = tx_stream.next().await {
             let tx_hash = tx.hash();
+            let tx_size = tx.storage_size();
+            if tx_size > MAX_BLOCK_SIZE {
```
davidrusu: Don't you want to maintain an accumulated size of valid txs here?

If the `accum_block_size + tx_size > MAX_BLOCK_SIZE` then skip the tx?

davidrusu (changes requested):
`Block::create()` should not try to be too smart (i.e. trimming the txs silently) It should validate that the block size does not exceed max block size and return error if that's not the case.

Then in the leader, you should have the logic to accumulate the total block size and skip transactions that pushes us over the limit.

Also, no need to check individual txs against the max block size, only check the accumulated size against the max block size.

PoulavBhowmick03:
> `Block::create()` should not try to be too smart (i.e. trimming the txs silently) It should validate that the block size does not exceed max block size and return error if that's not the case.
> 
> Then in the leader, you should have the logic to accumulate the total block size and skip transactions that pushes us over the limit.
> 
> Also, no need to check individual txs against the max block size, only check the accumulated size against the max block size.

yeah it makes sense. the silent truncation was a bad API and passing the txs to the block. Will remove both the per-tx check and the truncation from `Block::create` and replace them with a simple aggregate error. In the leader loop I will track accumulated size and skip any tx that would push us over the limit

services/chain/chain-leader/src/lib.rs
```diff
@@ -739,3 +743,124 @@ where
...
+            TestTx { id: 2, size: 32 },
+            TestTx { id: 3, size: 32 },
+        ];
+
+        let result = Block::create(parent, slot, proof, txs, &signing_key);
+
+        assert!(
+            matches!(result, Err(BlockError::ContentTooBig { .. })),
+            "Block::create must error when aggregate tx size exceeds MAX_BLOCK_SIZE"
+        );
```
youngjoon-lee: Yeah, for now it's hard to test the `propose_block` function in this module. So, I understand why you chose to test only `Block::create`. I think it's better to move this test to `core/src/block/mod.rs` which implements `Block::create`.

testnet/cfgsync/deployment-settings.yaml
```diff
@@ -0,0 +1,1122 @@
+blend:
```
youngjoon-lee: This file has been moved to the other location on master. So, please remove this file from your branch.

youngjoon-lee (approved):
Looks good and simpler than your initially version. Thank you so much. 

(I'd suggest waiting for @davidrusu's approval since he opened the issue.)

PoulavBhowmick03:
@youngjoon-lee ty for the review, i have made the change!

danielSanchezQ:
@PoulavBhowmick03 could you merge master or rebase? to see if CI turns green.

services/chain/chain-leader/src/lib.rs
```diff
@@ -562,14 +562,14 @@ where
...
         level = "debug",
-        skip(tx_selector, relays, ledger_state, ledger_config, proof, signing_key)
+        skip(_tx_selector, relays, ledger_state, ledger_config, proof, signing_key)
     )]
     async fn propose_block(
         parent: HeaderId,
         slot: Slot,
         proof: Groth16LeaderProof,
         signing_key: &Ed25519Key,
-        tx_selector: TxS,
```
davidrusu, line 584: @PoulavBhowmick03 is TxSelector completely dead in this new implementation? If so lets completely remove it.

cc @danielSanchezQ

danielSanchezQ: I would rather encapsulate the computation in the selector and make selector return the set of valid/invalid txs. Then use it as it was instead of having the logic hardcoded in the leader. Feel free to change the selector trait signature to fit this changes.

danielSanchezQ (approved):

davidrusu (changes requested):
I think this is no longer needed with #2842 #2868

danielSanchezQ:
Closing because of https://github.com/logos-blockchain/logos-blockchain/pull/2842 https://github.com/logos-blockchain/logos-blockchain/pull/2868. We should clean the selector as it is not used anymore.

PR #2554, opened 2026-04-16
fix(chain-network): filter by topic for block proposal stream

ntn-x2:

With trace logs enabled, I saw a bunch of `TRACE ChainNetwork: logos_blockchain_chain_network_service::network::adapters::libp2p: unrecognized gossipsub message` log entries. That is because in the block proposal stream we get all messages, including the mempool ones, as we did not filter by topic. I added the topic filter and changed the log to `debug` since we are now not meant to get any invalid messages anymore after applying the filter.

Yes.

@ntn-x2 

N/A

No.


hansieodendaal (approved):
LGTM

(Good catch!)

danielSanchezQ (approved):
Nice catch!!

PR #2555, opened 2026-04-16
feat(zone-sdk): zone sdk competing sequencers tui

pradovic:

Tui implementation that uses drive loop directly. If this is OK I will merge it to `zone-sdk-competing-sequencers` and continue from there.

Note: there is also a bug fix that was found while testing tui which is fixed here in sequencer.rs. This will be reviewed in [pr/zone-sdk-competing-sequencers](https://github.com/logos-blockchain/logos-blockchain/tree/pr/zone-sdk-competing-sequencers).


testnet/tui-zone/src/message.rs
```diff
@@ -0,0 +1,31 @@
...
+
+/// A structured application message with a unique ID for deduplication.
+///
+/// Real sequencers need to distinguish "same content published twice" from
+/// "same logical message re-published after a reorg". The `tx_id` field
+/// provides this: each user action gets a unique ID, and conflict resolution
+/// checks whether that ID is already on the canonical branch.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct AppMessage {
+    pub tx_id: Uuid,
```
hansieodendaal: I think `tx_id` is too close to transaction id, and could be misunderstood upon first glance.
```suggestion
    pub tx_uuid: Uuid,
```

testnet/tui-zone/src/state.rs
```diff
@@ -0,0 +1,132 @@
...
+
+/// Trait for zone state management.
+///
+/// The sequencer surfaces chain events (reorgs, finalization); the application
+/// maintains its own view of the world by implementing this trait.
+///
+/// A production implementation might use a database. This demo uses in-memory
+/// vecs.
+pub trait ZoneState {
+    /// Apply a message to the canonical (unfinalized) state.
```
hansieodendaal, line 14: Why would we need a trait if we only impl it once?

hansieodendaal (commented):
Looking good! Two small comments.

### Logging
Could we add some more logging? When `pub async fn run(...)` starts
```rust
let file_appender = tracing_appender::rolling::daily("logs", "tui-zone.log");
    let (log_writer, _log_guard) = tracing_appender::non_blocking(file_appender);

    let console_layer = tracing_subscriber::fmt::layer()
        .with_writer(std::io::stdout)
        .with_ansi(true)
        .with_filter(LevelFilter::WARN);

    let file_layer = tracing_subscriber::fmt::layer()
        .with_writer(log_writer)
        .with_ansi(false)
        .with_filter(LevelFilter::DEBUG);

    tracing_subscriber::registry()
        .with(console_layer)
        .with(file_layer)
        .init();
```
then add logs entries for TUI events as well.

testnet/tui-zone/src/state.rs
pradovic: Currently checkpoint is illustrated but does not help with in memory state. But this leaves the option to just implement trait over any db, and it should just start working. I thought it might be useful if we decide to introduce the db

davidrusu (approved):

PR #2557, opened 2026-04-16
fix(blend): keep memory of failed peers for new attempts in both edge and core swarms

ntn-x2:

PR on top of https://github.com/logos-blockchain/logos-blockchain/pull/2542.

Previously, we were only excluding the latest failed peer from a re-attempt. This means that if attempt 1 with peer A fails, attempt 2 will try any peer but peer A, and if attempt 2 fails, attempt 3 could pick peer A again. This optimization avoids this, enforcing that we try to iterate over all peers before resetting the list and retry with peers with which there was a previous failure. This is a rather edge case, but it's an optimization nonetheless.

This mechanism is used in the edge swarm to pick peers to send a message to, and in the core swarm to pick peers to connect to.

Yes.

@ntn-x2 

N/A

No.

### DIFF
--- a/services/blend/src/core/backends/libp2p/swarm.rs
+++ b/services/blend/src/core/backends/libp2p/swarm.rs
@@ -60,6 +60,10 @@ pub struct DialAttempt {
     address: Multiaddr,
     /// The latest (ongoing) attempt number.
     attempt_number: NonZeroU64,
+    /// Peers that have already been tried and failed for this dial cycle.
+    /// When all available peers have been tried, this set is cleared to allow
+    /// retrying from scratch.
+    failed_peers: HashSet<PeerId>,
 }
 
 /// [`DialAttempt`] with session information, i.e., whether the attempt was made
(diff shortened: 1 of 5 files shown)


youngjoon-lee (approved):
This looks a reasonable optimization. Thank you. 

Just one question: Do you think that we don't need to manage the failed peers *globally*? In this PR, we manage failed peers in the context of a dial attempt, which should work well. But if we are trying multiple dials, it might be useful to have a global set of failed peers and exclude them from all dial retries. However, it might not be so important, I think. Just asking for your opinion.

ntn-x2:
@youngjoon-lee yes I considered that as an alternative approach, but I did not choose it for the following reasons:

* Per-dial set of failed peers is closer to the current behaviour, which has been tested to work fine
* A global set would need to be pruned at regular intervals since a failure could occur because a peer was temporarily unreachable.
* In general, I found the per-attempt set to be easier to implement and to follow (closely related to point 1), so I chose to go with that.

So I'd go ahead with this, which should already improve things. If we find this does not work well, we can revise. What do you think?

PR #2560, opened 2026-04-16
chore(tests): use test name as prefix instead of suffix

ntn-x2:

I find it easier if failing test folders are grouped by test name instead of being sorted by the random string used for each node in each test. So I made the test name a prefix instead of a suffix.

Yes.

@ntn-x2 

N/A

No.


hansieodendaal (approved):
LGTM!

danielSanchezQ (approved):

PR #2564, opened 2026-04-17
chore: rust 1.95

### LINKED ISSUE #2561: [CI GENERATED] Update Rust to version 1.95.0
Update Rust to version 1.95.0. Also, please check the output of the feature manager job to see if we can avoid importing unnecessary features and update the `Cargo.lock` file with `cargo update`.

ntn-x2:

Fixes https://github.com/logos-blockchain/logos-blockchain/issues/2561.

Update codebase to [Rust 1.95](https://blog.rust-lang.org/2026/04/16/Rust-1.95.0/). Most lints were about using `Duration::from_secs` when `Duration::from_mins` was a better choice, and I think it's fair.

I also noticed the nix file was using the latest stable version, which is wrong since we want it pinned to a specific version, so I tried to do that myself.

Yes.

@ntn-x2 

N/A

No.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -5,7 +5,7 @@
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
 
     rust-overlay = {
-      url = "github:oxalica/rust-overlay/629bbb7f9d02787a54e28398b411da849246253b";
+      url = "github:oxalica/rust-overlay/8087ff1f47fff983a1fba70fa88b759f2fd8ae97";
       inputs.nixpkgs.follows = "nixpkgs";
     };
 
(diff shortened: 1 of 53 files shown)


flake.nix
```diff
@@ -40,13 +40,15 @@
           inherit system;
           overlays = [ rust-overlay.overlays.default ];
         };
+
+      rustVersion = "1.95.0";
```
strinnityk, line 44: Good idea!
I'd suggest going a step further, and tying it to the the source of truth, `rust-toolchain.toml`:
```suggestion
      rustVersion = (builtins.fromTOML (builtins.readFile ./rust-toolchain.toml)).toolchain.channel;
```

strinnityk (approved):

flake.nix
ntn-x2: Does this still require us to update the rev to one that includes the version we want to use?

ntn-x2: As discussed, since we need to update the ref hash anyway on toolchain updates, we decided to maintain an explicit version in this file.

danielSanchezQ (approved):

bacv (approved):

PR #2551, opened 2026-04-15
feat: Use genesis block

danielSanchezQ:
## 1. What does this PR implement?

* Expands usage of `GenesisBlock` 
* Add and fixes some ergonomics on `Inputs` and `Outputs` derived from different use in tests related to genesis blocks

## 2. Does the code have enough context to be clearly understood?

Should be fairly easy to follow by checking changes in methods signatures.

## 3. Who are the specification authors and who is accountable for this PR?

@danielSanchezQ 

## 4. Is the specification accurate and complete?

N/A

## 5. Does the implementation introduce changes in the specification?

N/A

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [x] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [x] 2. Description added.
* [x] 3. Context and links to Specification document(s) added.
* [x] 4. Main contact(s) (developers and specification authors) added
* [x] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 6. Link PR to a specific milestone.

PR #2565, opened 2026-04-17
feat: Cli for deployment configs

danielSanchezQ:

This adds a base to migrate to use cli tooling for devnet purposes. It adds a cli tool to deal with deployment settings and makes the first migration to GenesisBlock in config. This is needed for #2551.
Next steps will be upgrading the tool so it can generated configurations for specific keys for example. The idea is that at some point we can generate everything needed to run the dev/testnet fleet nodes with a few hits to this tool.

I really hope so.

@danielSanchezQ @bacv 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.


tools/blockchain-tools/src/bin/genesis.rs
bacv, line 1: Could the `tools` directory be not a cargo project, but a directory that could house multiple of them? I'm planning of moving cfgsync to this dir. Also I think the cargo project could be called `logos-blockchain-genesis-tools`, we add a prefix so that it's easier to identify logos related processes.

tools/blockchain-tools/src/bin/genesis.rs
```diff
@@ -0,0 +1,291 @@
...
+struct Cli {
+    #[command(subcommand)]
+    command: Commands,
+}
+
+#[derive(Subcommand, Debug)]
+enum Commands {
+    /// Generate a deployment config YAML from a well-known deployment or file,
+    /// with optional field overrides.
+    Config(ConfigArgs),
```
bacv, line 38: Do I understand correctly that for our release process we'll first generate the genesis using `run_block` command and then use `run_config` to apply the genesis to the deployment settings? If that's the case, maybe these two steps could be combined into one?

bacv (approved):
Looks exactly like what we need. If we're planning to use this mainly in our deployment process, commands could be combined to one. Minor comments about directory and cargo project title.

tools/blockchain-tools/src/bin/genesis.rs
danielSanchezQ: Ah, yes, I will do that and rename this crate inside!

tools/blockchain-tools/src/bin/genesis.rs
danielSanchezQ: Yes, I wanted to keep them separated so they compose.

nodes/node/binary/src/config/deployment/devnet.rs
```diff
@@ -42,895 +42,297 @@ cryptarchia:
...
+          payload:
+            service_type: BN
+            locators:
+            - /ip4/65.109.51.37/udp/3403/quic-v1
+            provider_id: '0e78cd0bf8cff5faa364776ac024c4c632c48f8a9d35789541a1b71009c3537e'
+            zk_id: ea7c784a570634bb4afb67553336984a9ce40eec02ed3d9249bd6d093b281f30
+            locked_note_id: f31739eb4397e7e117a6422d84eb89388765c86d4d2a912625a6d821a4bbe701
+        execution_gas_price: 0
+        storage_gas_price: 0
+      ops_proofs:
```
danielSanchezQ, line 329: @thomaslavaur this may collide later with the changes removing noproof. But can be fixed in a further pr I think. Pinging to confirm.

thomaslavaur: Yes I saw it we will have to replace it by the proof corresponding to the operation but it can be a default proof full of 0s

thomaslavaur: Maybe something we can remove from the deployment file and when we charge it we fill automatically with dumbproofs ?

danielSanchezQ: > Maybe something we can remove from the deployment file and when we charge it we fill automatically with dumbproofs ?

No, we will have to change it in the genesis builder and then it will be generated properly. I'll do in further PRs.

tools/blockchain-tools/Cargo.toml
ntn-x2, line 1: Is this a single project with multiple binaries? Or a directory meant to include multiple projects? In the former case, we should probably add the binary in this TOML file, to be able to add more in the future. But since we have a workspace already, I would rather have the `tools` folder be a regular folder with multiple crates inside.

ntn-x2: Then whatever utils crate could be re-used in tests, if needed.

tools/blockchain-tools/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+            }
+            Value::Mapping(input_map)
+        }
+        (_, overwrite) => overwrite,
+    }
+}
+
+#[expect(
+    clippy::too_long_first_doc_paragraph,
+    reason = "Necessary documentation"
```
ntn-x2, line 25: We could simply split the sentence leaving an empty line for this warning to go.

nodes/node/binary/src/config/cryptarchia/mod.rs
```diff
@@ -105,7 +105,15 @@ impl ServiceConfig {
                     .with_extension("json")
                     .as_path(),
             ),
-            starting_state: self.deployment.genesis_state.into(),
+            // TODO: Change starting state to be genesis block.
+            starting_state: self
```
ntn-x2, line 109: Ideally we want all these errors to be caught as soon as possible. Can't we modify the definition of a genesis block or change the deserialization logic to enforce this at that time instead? I think it would be more self-contained. This step is only making transformations and calculations, it should not panic for invalid values.

nodes/node/binary/src/config/deployment/devnet.rs
```diff
@@ -42,895 +42,297 @@ cryptarchia:
...
+      - NoProof
```
ntn-x2, line 330: Can I say this is ugly? If we need a special logic genesis block, why don't we just override whatever value the genesis block can take? Because we are then transforming into whatever the services need, we can be creative here since we have good decoupling in place. I would simply remove the need to have these proofs in the deployment value directly, and do any conversions internally.

ntn-x2 (commented):
As far as the deployment structure goes, I think we can do better than this. The whole purpose of having that was so that we can decouple the values specified in there from the values the services need. And in this case if we don't want to require proofs for the genesis block, we should just not require them at all at deserialization time. It will also make the config easier to follow, in my opinion. The CLI tool looks good tho and is useful to have regardless what we decide to do with genesis tx proofs.

nodes/node/binary/src/config/deployment/devnet.rs
danielSanchezQ: This is ugly indeed. We can actually just decide and say that genesis block does not contain any op_proofs at all maybe? cc @thomaslavaur

nodes/node/binary/src/config/cryptarchia/mod.rs
danielSanchezQ: This is refactored in a different PR and it will take a genesis block. It will be fixed in the other PR after this one is merged.

tools/blockchain-tools/src/lib.rs
danielSanchezQ: I'll try that, but I dont really understand this lint. I guess is because the first phrase is used for the doc entries when listing content?

tools/blockchain-tools/Cargo.toml
danielSanchezQ: There is a tools folder for tooling, then a blockchain-tools for rust specific tooling. I forgot to add the binary to the toml file tho, will do.

danielSanchezQ:
> As far as the deployment structure goes, I think we can do better than this. The whole purpose of having that was so that we can decouple the values specified in there from the values the services need. And in this case if we don't want to require proofs for the genesis block, we should just not require them at all at deserialization time. It will also make the config easier to follow, in my opinion. The CLI tool looks good tho and is useful to have regardless what we decide to do with genesis tx proofs.

I don't know what do you mean by "deployment structure goes". This changes nothing in the deployment structure but to use a genesis block instead of the transaction. If it is because of the block serialization is a different issue. I think we can actually move the genesis block to its own file instead of it be embedded in the deployment config. Then we can use the tooling to generate the genesis block when needed. There is an anex PR in where you can check the related changes too, in #2551 (which is waiting for this pr).  
wdyt?

nodes/node/binary/src/config/deployment/devnet.rs
ntn-x2: No what I am suggesting is that we specify a deserializable type that lives entirely in the node binary, that takes no proofs, and then we pass it to the chain service by adding a bunch of `NoProof`s, if we don't want to change the serialization format of the genesis block.

tools/blockchain-tools/src/lib.rs
ntn-x2: Yes, the first sentence is what is shown directly. The rest is expanded only if needed indeed.

ntn-x2:
Meaning that my main comment was about the structure of the deployment file. Everything else looked fine. So my main grip is about whether we would need to always use the tool to generate genesis txs, or if we can define a simpler format expected by the deployment type that makes it possible to spin up a chain without necessarily using the tooling.  But if we remove the need for signatures from the genesis block, that would already make things much easier.

danielSanchezQ:
> Meaning that my main comment was about the structure of the deployment file. Everything else looked fine. So my main grip is about whether we would need to always use the tool to generate genesis txs, or if we can define a simpler format expected by the deployment type that makes it possible to spin up a chain without necessarily using the tooling. But if we remove the need for signatures from the genesis block, that would already make things much easier.

Yes I agree, ideally we get rid of the proofs. cc @thomaslavaur I dont think we need to change the spec for this as it is an implementation detail where and how to handle the proofs, and in this case they are discarded anyway, wdyt?

ntn-x2:
Yes I'm not sure if the spec needs changing, but we can come up with whatever format for the deployment file as that is specific to our binary. So dis-intermediating between that and what the service expects (which is probably spec compliant) is possible, and I would argue in this case desirable.

danielSanchezQ:
> Yes I'm not sure if the spec needs changing, but we can come up with whatever format for the deployment file as that is specific to our binary. So dis-intermediating between that and what the service expects (which is probably spec compliant) is possible, and I would argue in this case desirable.

I removed the proofs from the genesis block building. Default proofs are injected when needed. Block is still a bit ugly, we should use hex_bytes for signatures and ids. Will do in a separate pr.

ntn-x2:
@danielSanchezQ do we need stuff like `parent_block`, `slot`, `proof_of_leadership` etc in the deployment-level block? I don't think so: that's what I was suggesting. Stuff that we can inject into the actual block AFTER deserializing the deployment setting should be done afterwards. Unless we see value in having a genesis block with a slot different than `0` of a parent block different than `0` (I don't see one), then we can get rid of them completely. We just need to define a different struct for the deployment config that will then be converted into the lower-level struct.

danielSanchezQ:
> @danielSanchezQ do we need stuff like `parent_block`, `slot`, `proof_of_leadership` etc in the deployment-level block? I don't think so: that's what I was suggesting. Stuff that we can inject into the actual block AFTER deserializing the deployment setting should be done afterwards. Unless we see value in having a genesis block with a slot different than `0` of a parent block different than `0` (I don't see one), then we can get rid of them completely. We just need to define a different struct for the deployment config that will then be converted into the lower-level struct.

Yes, we can do that. Or we can remove it from the configuration itself. It should have its own file anyway. Imo it should be a fully formed block. We do not really need to do the types massaging maybe?
Let me think about it.

youngjoon-lee (approved):
👍 Please merge this PR first. Then, I will rebase my PR #2174

danielSanchezQ:
I reverted a few changes. We will tackle the genesis block separation in a later PR @ntn-x2 as per discussions.

nodes/node/genesis/genesis_block.yaml
danielSanchezQ, line 1: I'll make a PR in where this file is embedded in the stand-alone-config.yaml. I'll also work on the serialization grooming a bit.

danielSanchezQ: I did it in this PR, as it was needed so tests would work.

thomaslavaur:
> > Meaning that my main comment was about the structure of the deployment file. Everything else looked fine. So my main grip is about whether we would need to always use the tool to generate genesis txs, or if we can define a simpler format expected by the deployment type that makes it possible to spin up a chain without necessarily using the tooling. But if we remove the need for signatures from the genesis block, that would already make things much easier.
> 
> Yes I agree, ideally we get rid of the proofs. cc @thomaslavaur I dont think we need to change the spec for this as it is an implementation detail where and how to handle the proofs, and in this case they are discarded anyway, wdyt?

Yes, as long as the genesis block is the same for everyone (and it contains the proof because for now the specs don't specify that the genesis block shouldn't contain proofs but just that their verification is skipped).

PR #2568, opened 2026-04-17
fix(tests): change tests artifacts folder

ntn-x2:

Changing the naming of the test artifacts messed up with our `.gitignore` as well as with our test results upload step on GH runners. To maintain the test name as the prefix, we need to introduce some other scheme. I propose we pre-pend test artifacts with `.run_` so that they remain hidden files, all have the same prefix and are hence easy to ignore. At the same time, I did not feel just uploading everything starting with `.` was the right way as we might have other hidden files in that folder in the future.

Yes.

@ntn-x2 

N/A

No.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,8 +19,7 @@ sim_config.json
 null
 
 # Integration test temp dirs
-tests/.tmp*
-/.tmp*
+tests/.run_*
 
 # Wildcard for any file that contains ignore
 *ignore*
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -19,8 +19,7 @@ sim_config.json
 null
 
 # Integration test temp dirs
-tests/.tmp*
-/.tmp*
+tests/.run_*
```
hansieodendaal, line 22: We might as well keep on ignoring .tmp
```suggestion
tests/.run_*
tests/.tmp*
/.tmp*
```

hansieodendaal (approved):
LGTM

hansieodendaal (approved):
LGTM

PR #2570, opened 2026-04-19
fix: failing tests align block fetch client with api block shape

hansieodendaal:

Return `ApiBlock` from block-by-id instead of deserializing into domain `Block<SignedMantleTx>`, which failed because the HTTP API omits internal header fields like `version`.

- update node/test clients and zone-sdk to consume `ApiBlock`
- switch cucumber/testing-framework block walkers to field-based access
- treat block-by-id 404 as `Ok(None)` for parent-chain traversal
- preserve existing block range/stream behavior

This fixes the failing e2e transaction and cucumber transaction tests.
Also adapted `Scenario: XXX staggered start` to be more robust, as both were failing, although their root cause failing is still not fixed. with this PR

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


andrussal (approved):
LGTM!

PR #2573, opened 2026-04-20
feat(blend): add HTTP endpoint for currently connected peers

ntn-x2:

This PR adds an HTTP endpoint at `/blend/info` that returns the following:
* If the node is a core node, a list of `(peer ID, is_healthy flag)` for each peer with which a connection is negotiated. If we are within the session transition period, also a list of IDs for those peers still connected from the old session.
* If the core is an edge node or if Blend is disabled (minimum network size is higher than the current number of core nodes), the endpoint returns `null`, to distinguish from the case where a core node has no peers at all.

Given the current Blend structure, it was extremely hard to come up with a design that was somehow acceptable. Once we finalize the spec to remove sessions, I will take the chance to do a big refactoring of the Blend codebase to have a single service and an internal state machine, which will make this logic also easier to implement. For the time being, also because I do not want to mess around too much with code that seems to work quite well now, I will let it be as it is.

So, this PR introduces a very basic HTTP endpoint to query the state of a running node.

It was hard to implement a somewhat easy-to-read solution, but I tried my best given the Blend huge size and large amount of connected pieces.

@ntn-x2 

N/A

No.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4189,6 +4189,7 @@ dependencies = [
  "bytes",
  "futures",
  "hyper",
+ "logos-blockchain-blend-service",
  "logos-blockchain-chain-broadcast-service",
  "logos-blockchain-chain-leader-service",
  "logos-blockchain-chain-service",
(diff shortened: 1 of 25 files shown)


youngjoon-lee (approved):
Looks good!

PR #2574, opened 2026-04-21
chore(api): Add testing connect endpoint

andrussal:

Add a testing endpoint that lets a node connect to a peer after it is already running. This is useful for a range of test scenarios and replaces the more cumbersome approach of editing initial peers and restarting nodes.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/nodes/api-common/src/paths.rs
+++ b/nodes/api-common/src/paths.rs
@@ -29,3 +29,4 @@ pub mod wallet {
 
 // testing paths
 pub const UPDATE_MEMBERSHIP: &str = "/test/membership/update";
+pub const TEST_NETWORK_CONNECT: &str = "/test/network/connect";
--- a/nodes/api-common/src/paths.rs
+++ b/nodes/api-common/src/paths.rs
@@ -29,4 +29,4 @@ pub mod wallet {
 
 // testing paths
 pub const UPDATE_MEMBERSHIP: &str = "/test/membership/update";
-pub const TEST_NETWORK_CONNECT: &str = "/test/network/connect";
+pub const DIAL_PEER: &str = "/test/network/dial_peer";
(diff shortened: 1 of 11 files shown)


danielSanchezQ (commented):
What does this solve? This is not gonna simulate any real situation right?
My point is that it feels a bit synthetic and may mislead results in tests (?)

andrussal:
> What does this solve? This is not gonna simulate any real situation right? My point is that it feels a bit synthetic and may mislead results in tests (?)

We have some tests where we start with isolated set of nodes and then let them join with each other to test convergence in some way.

andrussal:
> > What does this solve? This is not gonna simulate any real situation right? My point is that it feels a bit synthetic and may mislead results in tests (?)
> 
> We have some tests where we start with isolated set of nodes and then let them join with each other to test convergence in some way.

I think it is realistic for use cases like network partitions. Or simulating parallel proposals, which is hard to achieve directly in short time frame(which I tried to do actually).

nodes/api-common/src/paths.rs
```diff
@@ -29,3 +29,4 @@ pub mod wallet {
 
 // testing paths
 pub const UPDATE_MEMBERSHIP: &str = "/test/membership/update";
+pub const TEST_NETWORK_CONNECT: &str = "/test/network/connect";
```
hansieodendaal:
```suggestion
pub const DIAL_PEER: &str = "/test/network/dial_peer";
```

hansieodendaal (approved):
I strongly support this, but the API should rather be called DIAL_PEER. This is common in blockchains, for nodes to be able to dial peers, and an important test vector.

PR #2578, opened 2026-04-21
feat(zone-sdk):  competing sequencers e2e

pradovic:

This is the e2e tests part of https://github.com/logos-blockchain/logos-blockchain/pull/2446 for easier review.

The review suggested to switch tests to single node setup but this is not working, the node is not producing blocks. On master, the latest commit fix was to use 2 node setup in all tests and I did the same. We can investigate this issue in a separate PR as it is not related to zone-sdk.


hansieodendaal:
> The review suggested to switch tests to single node setup but this is not working

I think this has to do with blend changes.

**Root Cause:** 
A single cannot produce blocks when running alone

**The bug:**
The chain leader proposes a block but intentionally skips self-applying it and instead publishes it to the blend network:
```rust
// TODO: enable this once we eliminate sessions from Blend and so on
// Now we're disabling this to avoid a case which a proposing node
// transitions to a new session much earlier than other nodes.
debug!("skipping self-applying block and just publishing it");
blend_adapter.publish_proposal(block.to_proposal()).await;
```
The expectation is that blend routes the block back to the proposer through the p2p network — where it is then applied like any other received block. But with no blend peers, blend drops the block:
```
05:35:58  INFO  proposed block with id HeaderId(fab8fb6b...) [slot 100]
05:35:58  DEBUG skipping self-applying block and just publishing it
05:35:58  ERROR blend::backend::libp2p: Failed to publish message to blend network: NoPeers
```
**Consequence:**
The proposed block is lost. The chain tip stays at genesis (height 0). Every subsequent proposal still parents off genesis.

hansieodendaal (approved):
LGTM

_**Note:** Please just verify that this test also passes locally, as all tests except this one passed  in at least one of the two nextest e2e runs_
```
TRY 3 FAIL [  23.036s] logos-blockchain-tests::test_zone_sdk_e2e test_subscribe_to_finalized_deposit
```

PR #2581, opened 2026-04-21
feat: add retry logic to block_not_found error in block sync

### LINKED ISSUE #2530: Cryptarchia block sync should be deterministic and converge easily in system-level tests
These scenarios should easily succeed, but they fail almost all the time due to nodes not being able to sync:
- `Two fork chains join later and preserve persisted state wallet balances`
- `Orphan staggered fork start 1`
- `Orphan staggered fork start 2`

Run with
```
cargo test -p logos-blockchain-tests --features cucumber --test cucumber -- --name "<scenario_name>"
```

Typical failures:
```
2026-04-13T06:48:55.125254Z  INFO cucumber_steps: Waiting for consensus majority tip before selecting best node for 
  group 'FORK_B' - elapsed: 177.64s, reachable: 2/2, tips: ea2114fa5434e5bd..(1,h=7) | baeab17c029ff49b..(1,h=5)
2026-04-13T06:48:57.558778Z  WARN cucumber_steps: Step `I send 1 transactions of 700 LGO each from wallet 
  "WALLET_1A" to wallet "WALLET_BURN"` error: Logical error: No stable majority tip across candidate nodes for 
  group 'FORK_B' after 180.07s. Reachable nodes: 2/2. Tip groups: ea2114fa5434e5bd..(1,h=7) | baeab17c029ff49b..(1,h=5)
2026-04-13T06:48:57.559024Z  WARN cucumber_steps: Step `I send 1 transactions of 700 LGO each from wallet 
  "WALLET_1A" to wallet "WALLET_BURN"` error: Logical error: No stable majority tip across candidate nodes for group 
  'FORK_B' after 180.07s. Reachable nodes: 2/2. Tip groups: ea2114fa5434e5bd..(1,h=7) | baeab17c029ff49b..(1,h=5)

Finished - Scenario: Two fork chains join later and preserve persisted state wallet balances (Feature: Transactions)

   ✘  And I send 1 transactions of 700 LGO each from wallet "WALLET_1A" to wallet "WALLET_BURN"
```
```
2026-04-13T11:29:47.371503Z  INFO cucumber_steps: 3 fork chains detected!!! Elapsed: 298.95s, diff: 5, 
  heights: [8, 6, 3], fork hashes at height 3: {"6113daa1c55420fd5cc98763b4bf0bdb8334d737eab5c968be2fd9c145d6493b", 
  "a14da8cacf46d9a9c9a5bfe473d1d9b61abc4216e69ce2ce4adee650ed051ed8", 
  "32aea0868a96eb5fc153ed287ce3730c974fa056eaade34c2eba5709ef5bebe0"}

Finished - Scenario: Orphan staggered fork start 1 (Feature: Cryptarchia)

   ✘  When all nodes have at least 3 blocks and converged to within 1 blocks in 300 seconds
```

### Logs
- Example scenario console log from a run that passed: [scenario.log](https://github.com/user-attachments/files/26669266/scenario.log)
- Example scenario console log from a run that failed: [scenario_fail.log](https://github.com/user-attachments/files/26669286/scenario_fail.log)

hansieodendaal:

  This branch hardens orphan block recovery in chain sync by making block-provider failures explicit, retrying transient provider-side not-found responses during orphan parent download, and adding a dedicated future-
  block retry path so valid blocks are not dropped immediately when local slot time briefly lags behind block time.

  _What Changed_

  - Added structured BlocksUnavailableReason errors across the cryptarchia sync stack instead of passing opaque failure strings.
  - Propagated typed block-provider unavailability through libp2p messages, provider responses, chain-sync errors, and chain-service block-provider handling.
  - Updated orphan download startup to retry when peers return BlockNotFound or StartBlockNotFound, rather than failing orphan recovery immediately.
  - Hardened libp2p peer selection so a peer is only accepted after its first block-stream item is validated.
  - Rejected “poisoned” peers whose stream fails on the first item, instead of letting an erroring stream win peer selection and bypass the retry path.
  - Added a new bounded future-block retry path in chain-network:
      - surfaces FutureBlock { block_slot, current_slot } from chain-service through the API
      - retries block application with a short delay instead of treating the first FutureBlock as terminal
      - logs deferred retries so the behavior is visible in diagnostics
  - Applied that future-block retry path to both reconstructed proposal blocks and orphan-downloaded blocks.
  - Added regression tests for:
      - retryable typed not-found errors
      - non-retryable provider errors
      - first-item stream validation in the libp2p adapter
      - future-block retry behavior
  - Removed the `@undefined_behaviour` tag form the two `Orphan staggered fork start` cucumber scenarios that previously failed.

  _Why_

  The staggered-fork failures exposed two separate recovery gaps:

  - orphan parent download could stop after transient provider-side not-found responses
  - valid blocks could be dropped as FutureBlock when slot jitter caused the local node’s notion of time to lag briefly behind the block’s slot

  In these runs, slot timing was unstable enough that local time could be corrected backward and then clamped from actually moving backward. That meant local slot progression could temporarily trail the incoming
  block’s slot, making otherwise valid blocks look future-dated and causing them to be rejected too aggressively. This branch addresses that by retrying FutureBlock instead of dropping the block immediately.

  _Validation_

  - `cargo test -p logos-blockchain-chain-network-service --features libp2p`
  - `cargo test -p logos-blockchain-tests --features cucumber --test cucumber -- --name "Orphan staggered fork start 1"`
  - `cargo test -p logos-blockchain-tests --features cucumber --test cucumber -- --name "Orphan staggered fork start 2"`
  - `cargo test -p logos-blockchain-tests --features cucumber --test cucumber -- --name "Two fork chains join later and preserve persisted state wallet balances"`

Fixes #2530.

Yes

@hansieodendaal and @youngjoon-lee 

**TBD**

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.
* [X] 7. New or changed logs were reviewed against the [logging guidelines](../../CONTRIBUTING.md#logging-guidelines).


consensus/cryptarchia-sync/src/lib.rs
```diff
@@ -7,15 +7,36 @@ mod messages;
...
+
+impl std::fmt::Display for BlocksUnavailableReason {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::BlockNotFound(id) => write!(f, "BlockNotFound({id:?})"),
+            Self::StartBlockNotFound => write!(f, "StartBlockNotFound"),
+            Self::Unknown(reason) => f.write_str(reason),
+        }
+    }
+}
```
youngjoon-lee:
```suggestion
#[derive(thiserror::Error, Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
pub enum BlocksUnavailableReason {
    #[error("Block not found: {0}"]
    BlockNotFound(HeaderId),
    #[error("Start block not found"]
    StartBlockNotFound,
    #[error("Unknown error: {0}]  // or #[transparent]
    Unknown(String),
}
```
You can use [thiserror](https://docs.rs/thiserror/latest/thiserror/).

services/chain/chain-network/src/network/adapters/libp2p.rs
```diff
@@ -61,6 +64,63 @@ impl<Tx, RuntimeServiceId> LibP2pAdapter<Tx, RuntimeServiceId>
...
+    // before this peer is considered a successful candidate by `select_ok`.
+    //
+    // Behavior:
+    // - If the first item is any error, returns `Err` so this peer is excluded from
+    //   winner selection.
+    // - Otherwise, returns a reconstructed stream where the first item is put back
+    //   (`iter([first_item]).chain(stream)`) so downstream consumers see the full
+    //   original stream.
+    // - If the stream is immediately exhausted (`None`), returns it unchanged.
+    fn validate_first_block_response(
```
youngjoon-lee: What do you think about renaming this function to `check_first_block_response_ready`, and renaming `request_validated_blocks_stream_from_peer` to `request_available_blocks_stream_from_peer` (or something like that)? I'm suggesting this because the `validate` could be read as "validating the block contents" which cannot be done in this layer.

services/chain/chain-network/src/sync/orphan_handler.rs
```diff
@@ -179,26 +180,64 @@ where
...
+                        download_started_at,
+                    ));
+                }
+                Err(err) if Self::is_retryable_block_not_found(&err) => {
+                    warn!(
+                        target: LOG_TARGET,
+                        ?err,
+                        orphan_id = ?orphan_info.orphan_id,
+                        attempts_remaining,
+                        "Orphan fetch hit transient provider not-found; retrying with a reshuffled peer selection"
```
youngjoon-lee, line 227: Okay, now I see the flow in detail. This error happens only if all selected peers returned a `*BlockNotFound` error, and we're retrying by selecting another set of peers, which may be overlapped with the previous set. I think it is fine probabilistically if we have enough number of connected peers. 

Let's see how much this retry helps with the issue. If we often see the issue again, we can include the "source peer" to the download, or any other ways (adding delays between retries, or so).

services/chain/chain-network/src/lib.rs
```diff
@@ -344,11 +346,9 @@ where
 
                         Self::log_received_block(&block);
 
-                        match apply_block_and_reconcile_mempool::<_, Mempool, _>(
-                            block.clone(),
-                            relays.cryptarchia(),
-                            relays.mempool_adapter(),
-                        ).await {
+                        match Self::apply_reconstructed_block_with_future_block_retry(block, &relays)
```
youngjoon-lee:
```suggestion
                        match Self::apply_block_with_future_block_retry(block, &relays)
```
I think it's better to rename this function because it is used for both reconstructed blocks (received from gossipsub) and downloaded blocks (i.e., chainsync).

services/chain/chain-network/src/lib.rs
```diff
@@ -602,6 +624,48 @@ where
     }
 }
 
+async fn retry_future_block_apply_with_delay<F, Fut>(
```
youngjoon-lee, line 639: The reason why we have this function in chain-network service (not in chain service) is that we want to control the retry depending on the source of blocks (e.g., gossipsub, chainsync, etc.)? Then, it would be good to leave a doc comment in this function to explain the reason. 

I shortly thought about having this retry in the chain service regardless of source of blocks, but I'm not sure which way is better yet. So, let's go with this.

Another concern is that this function blocks at most 1.5s (= 3 * 500ms), which causes the main event loop blocked. But I think it's inevitable because:
- In chainsync, we cannot process subsequent blocks if we don't process this block right now.
- If this block was from gossipsub, it's likely that we cannot process subsequent new blocks if we don't process this block right now. 
  - If we put the block into the delay queue in this case, a subsequent block will trigger orphan handling, and it is cascaded. 

So, waiting 1.5s for a block looks fine, and I guess this is one of the reasons why we have this retry mechanism in the chain-network service (also because the main event loop of chain service handles more kinds of events).

I hope that 1.5s delays are not accumulated. In other words, I guess, just one 1.5s delay may mitigate the slot skew and subsequent blocks may not cause `FutureBlock` error. If not, this delay will make chain-network service too slow.

youngjoon-lee (commented):
Thank you so much for investigating the issues and opening this PR 👍 I left some suggestions and concerns before approving.

- I really like to have the explicit error types! The previous error types in chainsync were too opaque.
- I'd prefer splitting the deferred future-block processing into another PR, if possible next time, because it is not directly required for the chainsync retry. I know we need it for making CI green, but we can stack the 2nd PR on top of the 1st PR to see if CI passes in the 2nd PR. However, this PR is not hard to review. So, it's fine. Thank you!

services/chain/chain-network/src/lib.rs
youngjoon-lee: Hm, I think we can optimize this:
- If `FutureBlock` occurs during orphan handling, we can delay the entire orphan handling job by putting it into a queue because subsequent blocks will be rejected as well. By putting it into the queue, we don't block the main event loop and process other events which may succeed. 
- If `FutureBlock` occurs during processing new blocks from gossipsub, we put it in another delay queue to not block the main event loop. If a subsequent block from gossipsub is rejected by chain service with `ParentMissing`, we can check if the block is a child of one of the blocks in the delay queue. If so, we can put it into the queue as well.

In this way, we don't block the main event loop at all, and handle all the delays necessary. 
What do you think? If this looks good, I can prepare a PR on top of this PR, so that we can merge this first. 

---

Also, I think we can calculate how many seconds we have to delay blocks because we know `block_slot` and `current_slot`. (Of course, we need the `slot_duration` config to calculate the seconds).

hansieodendaal: Please go ahead and prepare that PR, I fully agree with your assessment!

youngjoon-lee (approved):
Thank you. I think you can merge this PR. 

For https://github.com/logos-blockchain/logos-blockchain/pull/2581#discussion_r3127625833, I will open a separate PR next week.

hansieodendaal:
> Thank you. I think you can merge this PR.
> 
> For [#2581 (comment)](https://github.com/logos-blockchain/logos-blockchain/pull/2581#discussion_r3127625833), I will open a separate PR next week.

**TODO**:  The retry should be optimized to not block the main event loop.

hansieodendaal:
**Note:** Merging, failing e2e tests are due to a regression elsewhere, not this PR.

PR #2586, opened 2026-04-22
feat(wallet): MMR(2/4): voucher path tracking

youngjoon-lee:
## 1. What does this PR implement?

This is PR 2/4 in the series. For detailed background, please see the 1st PR #2585.
- Now wallet tracks Merkle paths of vouchers owned by the user.
- Paths are incrementally updated on each block and snapshotted at epoch transitions.
  - Chain service gives wallet service two pieces: MMR and `new_voucher_cm` (both are lightweight).
  - Wallet service updates the existing tracked paths by inserting `new_voucher_cm` into MMR. The same insertion is also done by chain service, but wallet service does it again to update tracked paths. I tried many other approaches, but I believe that this one is the best.

## 2. Does the code have enough context to be clearly understood?

yes

## 3. Who are the specification authors and who is accountable for this PR?

@youngjoon-lee 
spec: @thomaslavaur 

## 4. Is the specification accurate and complete?

yes

## 5. Does the implementation introduce changes in the specification?

no

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [x] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [x] 2. Description added.
* [x] 3. Context and links to Specification document(s) added.
* [x] 4. Main contact(s) (developers and specification authors) added
* [x] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 6. Link PR to a specific milestone.
* [x] 7. New or changed logs were reviewed against the [logging guidelines](https://github.com/logos-blockchain/logos-blockchain/blob/master/CONTRIBUTING.md#logging-guidelines).

PR #2585, opened 2026-04-22
feat(ledger): MMR(1/4): replace DynamicMerkleTree with MMR for voucher management

youngjoon-lee:
### Background

Replaced `DynamicMerkleTree` with `MerkleMountainRange` (MMR) for voucher commitments.

Unlike `DynamicMerkleTree` which holds all leaves and internal nodes in memory, MMR doesn't keep leaves. It holds only peaks, achieving at most `33 * Fr = 1KiB` memory (at most 33 peaks).
Since each block has a voucher, it's much better to use MMR to minimize memory usage.

### Changes

This is PR 1/4 in the series.
- Insert vouchers into MMR immediately on each block instead of accumulating pending vouchers and flushing at epoch boundary.
- The MMR root/size are snapshotted at epoch transitions.
- Remove voucher_merkle_path and has_claimable_voucher from LeaderState since path maintenance will move to the wallet service.

## 2. Does the code have enough context to be clearly understood?

I hope so.

## 3. Who are the specification authors and who is accountable for this PR?

@youngjoon-lee 
spec: @thomaslavaur 

## 4. Is the specification accurate and complete?

yes

## 5. Does the implementation introduce changes in the specification?

no

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [x] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [x] 2. Description added.
* [x] 3. Context and links to Specification document(s) added.
* [x] 4. Main contact(s) (developers and specification authors) added
* [x] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 6. Link PR to a specific milestone.
* [ ] 7. New or changed logs were reviewed against the [logging guidelines](https://github.com/logos-blockchain/logos-blockchain/blob/master/CONTRIBUTING.md#logging-guidelines).

PR #2587, opened 2026-04-22
feat(wallet-service): MMR(3/4): use wallet MMR paths and persist LIB WalletState

youngjoon-lee:

This is PR 3/4 in the series. For detailed background, please see the 1st PR #2585.
- Use voucher merkle paths tracked in wallet for LeaderClaim
- Persist LIB WalletState to rebuild voucher paths when restart
  - As implemented in the 2nd PR #2586, wallet service must update tracked voucher paths **incrementally** block by block. Therefore, wallet service must be able to track all blocks without any missing one.
  - Previously, wallet service was not persisting any WalletState. Instead, on restart, it fetched the current chain's LIB and its LedgerState from the chain service. And then, it fetched/applied blocks in range `[lib, tip]`. It was good enough because wallet service was tracking only UTXOs, all of which are contained in a single LedgerState. However, voucher paths are different. MMR doesn't contain leaves (vouchers). Wallet service must incrementally update voucher paths for each block. Therefore, wallet service must persist the last LIB WalletState that it knew, and must be able to fetch/apply blocks from the last LIB to the tip.
    - To support it, chain service API must be changed. A next PR 4/4 will cover it. 

yes

@youngjoon-lee 
spec: @thomaslavaur 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.


services/wallet/src/states.rs
```diff
@@ -46,22 +51,41 @@ impl<'u> ServiceState<'u> {
...
-                settings
-                    .known_keys
-                    .clone()
-                    .into_iter()
-                    .map(|(key_id, pk)| (pk, key_id)),
-                state.vouchers,
+        let known_keys = settings
+            .known_keys
+            .iter()
+            .map(|(key_id, pk)| (*pk, key_id.clone()));
```
danielSanchezQ, line 57: What happend if the recovery file gets corrupted? 🤔

danielSanchezQ (approved):

services/wallet/src/states.rs
youngjoon-lee: If the recovery file is corrupted unexpectedly on disk? Then, in the worst case, this `ServiceState::new` function will load incorrect states **silently** unless JSON parsing is failed, and it will cause the backfilling (outside of this function) failed. But there may be other issues that are hard to predict now. I think this is a common issue for all services that load recovery files.

Or, do you mean a case which the recovery file (which contains `(pk, key_id)`s) is leaked? Yes, that can happen. But I think it is fine as long as secret keys are secured by KMS. Wdyt?
Actually, `known_keys` have been already in the recovery file before this PR. If we have a security concern, we can discuss/improve it as a next PR.

core/src/proofs/merkle.rs
```diff
@@ -40,4 +54,27 @@ mod tests {
...
+
+    #[test]
+    fn test_mmr_path_to_witness() {
+        let path = lb_mmr::MerklePath {
+            leaf_index: 5,
+            siblings: vec![Fr::from(1u64), Fr::from(2u64), Fr::from(3u64)],
+        };
+        let (items, selectors) = mmr_path_to_witness(&path);
+        assert_eq!(items, vec![Fr::from(1u64), Fr::from(2u64), Fr::from(3u64)]);
+        // For leaf index 5, the sibling positions are right, left, right respectively.
```
thomaslavaur: can we take an asymetric example here instead ? (because 5 is 101 so we don't care if the order is reversed or not) Maybe we can try with 11? (which is 1011 and so the order is important here)

thomaslavaur (approved):

core/src/proofs/merkle.rs
youngjoon-lee: Good point. 639c3de94

PR #2588, opened 2026-04-22
feat(chain): MMR(4/4): GetHeaders with storage fallback

youngjoon-lee:

This is PR 4/4 in the series. For detailed background, please see the 1st PR #2585.

### Background
This PR fixes chain service to support wallet service fetching blocks from its last LIB, which is not guaranteed to always match with the chain's LIB.
Chain service already has `GetHeaders` API, but it provides blocks only up to (down to?) the current chain's LIB, which may be newer than the last LIB that wallet service remembers. So, with this API, wallet service cannot catch up the chain.

### Solution
- GetHeaders now walks in-memory branches first, then falls back to storage for blocks pruned below LIB.
  - This allows wallet svc to backfill from its persisted LIB, which is not guaranteed to always match the chain's LIB.
- Also, removed LIMIT=512 in chain service. Instead, defined LIMIT only for REST API.

yes

@youngjoon-lee 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/services/chain/chain-service/src/lib.rs
+++ b/services/chain/chain-service/src/lib.rs
@@ -103,6 +103,8 @@ pub enum Error {
     Mempool(String),
     #[error("Block header id not found: {0}")]
     HeaderIdNotFound(HeaderId),
+    #[error("Parent header ID not found for child={0}")]
+    ParentIdNotFound(HeaderId),
     #[error("Service session not found: {0:?}")]
     ServiceSessionNotFound(ServiceType),
 }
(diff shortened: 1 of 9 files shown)


services/chain/chain-service/src/lib.rs
```diff
@@ -1064,55 +1068,76 @@ where
...
+
+        let mut in_memory = Vec::new();
+        let mut current = from_descendant;
+        while let Some(branch) = branches.get(&current) {
+            in_memory.push(Ok(branch.id()));
+            if branch.id() == to_ancestor {
+                // All blocks are found in memory. Return immediately
+                return Box::pin(stream::iter(in_memory));
+            }
+            current = branch.parent();
```
danielSanchezQ, line 1103: I wonder if we could do this lazily. Probably really difficult without using using sync structures or spawning up a channel and a task 🤔 . Would be nice to open an issue just to check it.

danielSanchezQ (approved):

services/chain/chain-service/src/lib.rs
youngjoon-lee: Yeah, I tried it initially but was not easy as you mentioned. I opened an issue https://github.com/logos-blockchain/logos-blockchain/issues/2601. I think it's good to go with the current approach for now, assuming that `k` is not very large (2160 blocks in mainnet = 67KiB).

PR #2595, opened 2026-04-22
chore(cfgsync): Load entropy from file or generate random in cfgsync

bacv:

Allows cfgsync to generate random entropy for genesis creation.

Yes

@bacv @hansieodendaal 

N/A

N/A

## Checklist

* [x] 2. Description added.


bacv:
@davidrusu do we still need a predefined entropy file when generating genesis, or could we use randomly generated one to simplify the deployments?

davidrusu:
@bacv entropy file was to have a secure but deterministic genesis block, if we don't need to have a deterministic genesis block, then we don't need the entropy file.

Previously we were deriving the initial node keys using insecure key-material

hansieodendaal (approved):
LGTM

bacv:
> Previously we were deriving the initial node keys using insecure key-material

At this point it seems that we don't need deterministic genesis block, as we have no use case for it.

davidrusu:
> > Previously we were deriving the initial node keys using insecure key-material
> 
> At this point it seems that we don't need deterministic genesis block, as we have no use case for it.

OK, than we don't need an entropy file at all, we can just generate the keys directly without an entropy file.

danielSanchezQ:
> > > Previously we were deriving the initial node keys using insecure key-material
> > 
> > 
> > At this point it seems that we don't need deterministic genesis block, as we have no use case for it.
> 
> OK, than we don't need an entropy file at all, we can just generate the keys directly without an entropy file.

Also this is due to migrate to a generation tool in where either we reuse some keys or we generate some indeed.

PR #2604, opened 2026-04-23
fix: handle cases of 1-node networks for both blend and gossipsub

ntn-x2:

With a libp2p backend, a network size of 1 is problematic, because the peer will try to blend a message but it won't since there are no peers. So we enforce in the deployment deserialization that the value is at least `2`, so that single-node deployments will always bypass Blend and use gossipsub directly.

As for gossipsub, even with Blend disabled, sending a message to peers will fail if there is no peers. So only on the first failure we notify the node about its own block before scheduling it for retry. On success, we only notify the node about its own message if it never failed, otherwise it means the node was already notified and we avoid double-notifying it.

Yes.

@ntn-x2 

N/A

No.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1793,7 +1793,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7ab67060fc6b8ef687992d439ca0fa36e7ed17e9a0b16b25b601e8757df720de"
 dependencies = [
  "data-encoding",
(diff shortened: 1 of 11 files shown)


danielSanchezQ (approved):
Nice, minimal changes. Thanks!

andrussal (approved):

davidrusu (approved):
🚢!

PR #2610, opened 2026-04-23
feat: add flexible bounded blocks streaming api

hansieodendaal:

This PR introduces a **new ranged block streaming API** (`/cryptarchia/blocks_range`), without modifying the existing streaming endpoint (`/cryptarchia/events/blocks/stream`).

---

##### New endpoint

- `GET /cryptarchia/blocks_range`

This endpoint allows clients to stream processed blocks within a **slot-bounded window**, with:

- block-count limiting (`blocks_limit`)
  - If `blocks_limit` is omitted:
    - explicit bounded range → stream the full matching range (bounded by server max)
    - default/open request → return up to 100 blocks
- direction control (ascending / descending)
- server-side batching
- optional immutable-only filtering
- optional slot window `[slot_from, slot_to]` (inclusive)
  - defines the search range; only actual blocks in this range are returned

Streaming is **incremental (NDJSON)**, allowing clients to process results as they arrive.

---

##### Query parameters

All parameters are optional:

```rust
#[derive(IntoParams, Deserialize, Validate)]
#[into_params(parameter_in = Query)]
pub struct BlocksStreamQuery {
    /// If omitted, the server chooses a default lower bound.
    /// For descending streams this is `slot 0` (bounded by `blocks_limit`).
    /// For ascending streams, `slot_from` is estimated from the average
    /// slots-per-block and `blocks_limit`, biased so the stream ends near
    /// `slot_to`. This may return fewer than `blocks_limit` blocks; callers
    /// can refine by specifying `slot_from` explicitly.
    #[serde(default)]
    #[param(minimum = 0)]
    pub slot_from: Option<u64>,
    /// Upper bound slot (inclusive). Defaults to tip slot, or LIB slot when
    /// `immutable_only=true`.
    #[serde(default)]
    #[param(minimum = 0)]
    pub slot_to: Option<u64>,
    /// Sort direction. Defaults to descending (`true`).
    #[serde(default)]
    pub descending: Option<bool>,
    /// The maximum number of actual blocks to return. If omitted:
    /// - explicit bounded slot range (`slot_from` and `slot_to`) defaults to
    ///   the server maximum (`630_720_000`);
    /// - otherwise defaults to `100`.
    #[serde(default)]
    #[validate(custom(function = "validate_blocks_limit"))]
    #[param(minimum = 1, maximum = 630_720_000, default = 100, example = 100)]
    pub blocks_limit: Option<usize>,
    /// Server chunk size hint for streamed delivery. Defaults to `100` ,
    /// maximum `1000`.
    #[serde(default)]
    #[validate(custom(function = "validate_server_batch_size"))]
    #[param(minimum = 1, maximum = 1_000, default = 100, example = 100)]
    pub server_batch_size: Option<usize>,
    /// When true, include only immutable blocks.
    /// If `slot_to` is omitted, the default anchor is LIB slot.
    #[serde(default)]
    pub immutable_only: Option<bool>,
}
```

---

##### Default behavior

```
GET /cryptarchia/blocks_range
```

- streams up to 100 blocks
- descending (newest → oldest)
- anchored at tip

---

##### Examples

- `/cryptarchia/blocks_range`  
  → latest 100 blocks from tip (_descending order_)

- `/cryptarchia/blocks_range?immutable_only=true`  
  → latest 100 immutable blocks from LIB (_descending order_)

- `/cryptarchia/blocks_range?blocks_limit=26`  
  → latest 26 blocks (_descending order_)

- `/cryptarchia/blocks_range?slot_to=<slot>`  
  → stream up to `blocks_limit` blocks ending at the given slot (_descending order_)

- `/cryptarchia/blocks_range?slot_from=<a>&slot_to=<b>`  
  → stream within `[a, b]` (_inclusive, descending order_)

- `/cryptarchia/blocks_range?slot_from=1&slot_to=<tip>`  
  → scan the full chain `[1, tip]` and return matching blocks  
    (bounded by `blocks_limit` or server maximum) (_descending order_)

- `/cryptarchia/blocks_range?slot_to=<b>&descending=false`
→ stream blocks within `[<server_estimate_lower_bound>, b]` (_inclusive, ascending order, may return fewer than `blocks_limit` blocks_)

- `/cryptarchia/blocks_range?slot_from=<a>&slot_to=<b>&descending=false&immutable_only=true`
→ stream immutable blocks within `[a, b]` (_inclusive, ascending order_)

---

##### Execution model

The endpoint uses a **hybrid strategy depending on the requested slot window**:

- **Entire range at or below LIB (`slot_to <= LIB`)**
  - fetched using the immutable slot index
  - efficient storage-native iteration

- **Entire range above LIB (`slot_from > LIB`)**
  - fetched by walking the current canonical chain from tip
  - reorg-aware (does not rely on persisted mutable indexing)

- **Range crosses LIB (`slot_from <= LIB < slot_to`)**
  - immutable portion (`<= LIB`) fetched via slot index
  - mutable portion (`> LIB`) fetched via canonical chain walk
  - both segments are merged in requested order

`blocks_limit` always applies to the number of **actual blocks returned**, not the number of slots scanned.

---

##### Streaming semantics

- NDJSON stream
- incremental delivery (no full buffering)
- batched using `server_batch_size`
- supports large/full-chain downloads safely

---

##### Empty results

An empty stream is returned when:

- no blocks exist in the resolved slot range
- the first chunk resolves to no events

---

##### Validation / errors (400)

Request is rejected if:

- `blocks_limit` or `server_batch_size` out of bounds
- `slot_from > slot_to`
- `slot_to` exceeds allowed anchor:
  - > tip (normal mode)
  - > LIB (`immutable_only=true`)

---

#### Testing

- end-to-end coverage for ranged streaming
- LIB boundary scenarios
- chunking behavior (`server_batch_size`)
- immutable-only streaming
- validation and error handling

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


nodes/node/binary/src/api/handlers.rs
```diff
@@ -51,18 +50,245 @@ use overwatch::{
...
+    number_of_blocks: usize,
+    blocks_to: Option<HeaderId>,
+    chunk_size: usize,
+    immutable_only: bool,
+}
+
+fn parse_blocks_stream_request(
+    query: &BlocksStreamQuery,
+) -> Result<BlocksStreamRequest, &'static str> {
+    if query.number_of_blocks == Some(0) {
```
andrussal: Should we also validate an upper bound here? Right now very large number_of_blocks values seem possible, and later we fall back to usize::MAX.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -51,18 +50,245 @@ use overwatch::{
...
     },
 };
 
+const BLOCKS_STREAM_CHUNK_SIZE: usize = 1_000;
+const DEFAULT_NUMBER_OF_BLOCKS_STREAM_BLOCKS: u64 = 100;
+
+struct BlocksStreamRequest {
+    number_of_blocks: usize,
+    blocks_to: Option<HeaderId>,
+    chunk_size: usize,
```
andrussal: chunk_size feels more like an execution/batching hint than part of the API semantics. I’m a bit worried this will be confusing to users, since it doesn’t really change what data is requested, only how it is produced.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -678,35 +905,98 @@ where
...
 {
-    let stream = mantle::get_new_blocks_stream::<_, _, ConsensusService, _>(&handle)
-        .await
-        .map(|stream| stream.map(ApiProcessedBlockEvent::from));
-    match stream {
-        Ok(stream) => responses::ndjson::from_stream(stream),
-        Err(error) => (StatusCode::INTERNAL_SERVER_ERROR, error.to_string()).into_response(),
+    let request = match parse_blocks_stream_request(&query) {
+        Ok(request) => request,
+        Err(message) => return (StatusCode::INTERNAL_SERVER_ERROR, message).into_response(),
```
andrussal: Should we distinguish client vs server errors here? For example, invalid query parameters seem like 400 Bad Request, while this currently maps all parse failures to 500 Internal Server Error.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -51,18 +50,245 @@ use overwatch::{
...
 };
 
+const BLOCKS_STREAM_CHUNK_SIZE: usize = 1_000;
+const DEFAULT_NUMBER_OF_BLOCKS_STREAM_BLOCKS: u64 = 100;
+
+struct BlocksStreamRequest {
+    number_of_blocks: usize,
+    blocks_to: Option<HeaderId>,
+    chunk_size: usize,
+    immutable_only: bool,
```
andrussal: I’m not sure immutable_only is justified in this API. We already have a dedicated immutable-blocks endpoint, and having it here also makes blocks_to semantics less clear. For callers who only want immutable history, the existing endpoint seems like the better path anyway — and probably the more efficient one too.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -51,18 +50,245 @@ use overwatch::{
...
+
+fn parse_blocks_stream_request(
+    query: &BlocksStreamQuery,
+) -> Result<BlocksStreamRequest, &'static str> {
+    if query.number_of_blocks == Some(0) {
+        return Err("number_of_blocks must be greater than or equal to 1");
+    }
+
+    Ok(BlocksStreamRequest {
+        number_of_blocks: usize::try_from(
```
andrussal: `(usize::try_from(...).unwrap_or(usize::MAX)` means an out-of-range value gets coerced into usize::MAX, which feels pretty risky here. Should we reject invalid/oversized values instead, or at least enforce a reasonable upper bound?)

andrussal (commented):
Nice work! Left some comments.

nodes/node/binary/src/api/handlers.rs
hansieodendaal: Maybe we can rename it `server_batch_size`? It is bounded to `1_000` maximum, which should give the best overall performance, but uses the most memory.

andrussal: > Maybe we can rename it `server_batch_size`? It is bounded to `1_000` maximum, which should give the best overall performance, but uses the most memory.

That sounds better. 

I am not sure if it possible to optimise it tho, we don't know if next 1000 blocks each are 2Mb or 100Kb. 

I think it is fine to have it if we think it is used only by the same users who run the node. Not some kind of public API.

nodes/node/binary/src/api/handlers.rs
hansieodendaal: Fair point about the overlap.
- The reason immutable_only exists on blocks_stream is that the consumption model is different. The dedicated immutable-blocks endpoint (`get_immutable_blocks`) currently does not stream blocks, it returns all requested blocks in a single fully buffered JSON response, while `blocks_stream` returns NDJSON incrementally, so clients can process results as they arrive instead of waiting for everything.
- So even if filtering semantics are similar, they serve different use cases:
  - `immutable_blocks:` batch snapshot in a single payload
  - `blocks_stream?immutable_only=true:` progressive delivery with better backpressure behavior
- Also, both endpoints already treat out-of-range requests as empty results, so behavior is consistent there.

nodes/node/binary/src/api/handlers.rs
hansieodendaal: I made the default 100 and the max 1_000.

nodes/api-common/src/paths.rs
```diff
@@ -14,7 +14,7 @@ pub const SDP_POST_ACTIVITY: &str = "/sdp/activity";
 pub const SDP_POST_WITHDRAWAL: &str = "/sdp/withdrawal";
 pub const LEADER_CLAIM: &str = "/leader/claim";
 
-pub const BLOCKS: &str = "/cryptarchia/blocks";
```
pradovic, line 17: This might break existing clients. Do we know what needs to be updated after merging?

From the top of my head LEZ and block explorer. Although, LEZ should be using pinned node and client version so it should be fine

hansieodendaal: Thanks, I will update the PR description

nodes/node/binary/src/api/handlers.rs
andrussal: I’m surprised the existing immutable_blocks endpoint is not streaming, because it already has an immutable-block index available. That seems like a much more natural place to expose a real stream that reads blocks from storage one at a time, which is why I said it is probably the more efficient path too.

In contrast, this new path does not have an index for direct range access. It resolves the range by walking backward from the canonical tip via parent links, and then buffers each matching chunk in memory before yielding it. So the HTTP response is streamed, but the storage access is still chunk-buffered traversal rather than a true streaming read.

But unless we implement index for all blocks, we have to live with it. The reason we didn't add index for non-immutable blocks was that immutable blocks are a simple stable chain, whereas non-immutable blocks can still belong to competing branches, so it's harder to build the same kind of index.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -45,24 +48,188 @@ use lb_tx_service::{
...
+                let Ok(next_chunk) = fetch_blocks_stream_chunk::<StorageBackend, RuntimeServiceId>(
+                    &handle,
+                    &chain_info,
+                    next_from,
+                    chunk_to,
+                    immutable_only,
+                )
+                .await
+                else {
+                    return None;
```
pradovic: Will this swallow error? Would it make sense to return `Stream<Item = Result<ProcessedBlockEvent, Error>>`?

services/api/src/http/mantle.rs
```diff
@@ -257,7 +257,210 @@ where
...
+
+        current_id = parent_id;
+        current_height = current_height.saturating_sub(1);
+    }
+
+    blocks.reverse();
+    Ok(blocks)
+}
+
+pub async fn find_canonical_height_by_header_with_snapshot<
```
pradovic: This might be unbounded chain walk, basically a  DOS vector. Would it make sense to bound it, maybe to number_of_blocks + slack?

Probably best way would be height index, would it make sense to implement it? Just an idea, maybe for different PR.

services/api/src/http/mantle.rs
```diff
@@ -257,7 +257,210 @@ where
...
+    <StorageBackend as StorageChainApi>::Tx: From<Bytes> + AsRef<[u8]>,
+    RuntimeServiceId: Debug
+        + Send
+        + Sync
+        + Display
+        + 'static
+        + AsServiceId<StorageService<StorageBackend, RuntimeServiceId>>
+        + AsServiceId<Cryptarchia<RuntimeServiceId>>,
+{
+    let tip_height = usize::try_from(chain_info.height).unwrap_or(usize::MAX);
```
pradovic: nitpick: I think this is silently ignoring usize overflow, and using MAX in this case. Maybe we should return an error in this case.  As if using max, subsequent `.saturating_sub / .min` will give weird result.

pradovic (commented):
Nice work! Looks good to me, I left some comments

services/api/src/http/mantle.rs
```diff
@@ -257,7 +257,210 @@ where
...
+        handle,
+        blocks_from,
+        blocks_to,
+        immutable_only,
+        &chain_info,
+    )
+    .await
+}
+
+pub async fn get_blocks_in_range_with_snapshot<Transaction, StorageBackend, RuntimeServiceId>(
```
andrussal: I tried to wrap my head around this because something still feels a bit off here.

Do we always start walking from `chain_info.tip` even when fetching a later chunk from the middle of the requested range? If so, large requests seem to repeatedly walk the same parent links from tip for each chunk.

With `number_of_blocks` capped at `630_720_000`, a request can effectively ask for the whole chain, so this could become very expensive on a long-running node.

danielSanchezQ: Exactly @andrussal, I agree, that is why I said we should use pagination instead, with a fixed limit. The add a rate-limiter middleware so people cannot abuse the system.

services/api/src/http/mantle.rs
hansieodendaal: Yes, a height index would make sense, for a different PR, @andrussal also metntioned it.

nodes/api-common/src/paths.rs
strinnityk: For future reference: A way to achive the name update without breaking anything is to enable both URLs simultaneously, merge, update dependentes and finally remove the old name.

nodes/api-common/src/paths.rs
```diff
@@ -30,3 +30,9 @@ pub mod wallet {
 // testing paths
 pub const UPDATE_MEMBERSHIP: &str = "/test/membership/update";
 pub const DIAL_PEER: &str = "/test/network/dial_peer";
+
+/// This maximum blocks stream chunk size is a happy medium between performance
+/// and memory use
+pub const MAX_BLOCKS_STREAM_CHUNK_SIZE: usize = 1_000;
+/// 200 years worth of blocks if 1 is produced every 10s
+pub const MAX_BLOCKS_STREAM_BLOCKS: usize = 630_720_000;
```
strinnityk: I'd suggest using an inner mod for clarity's sake, given these are constants and not paths.
On another note: Why does block streaming have maximum streaming? As for the chunk size I could understand it, although I don't think it serves an actual purpose here.

nodes/api-common/src/paths.rs
```diff
@@ -14,7 +14,7 @@ pub const SDP_POST_ACTIVITY: &str = "/sdp/activity";
 pub const SDP_POST_WITHDRAWAL: &str = "/sdp/withdrawal";
 pub const LEADER_CLAIM: &str = "/leader/claim";
 
-pub const BLOCKS: &str = "/cryptarchia/blocks";
+pub const IMMUTABLE_BLOCKS: &str = "/cryptarchia/immutable_blocks";
```
strinnityk: If we want to rename it, a more appropriate name would be `/cryptarchia/blocks/immutable`

nodes/node/binary/src/api/handlers.rs
```diff
@@ -45,24 +48,188 @@ use lb_tx_service::{
...
     serializers::{
         blocks::{ApiBlock, ApiProcessedBlockEvent},
         transactions::ApiSignedTransactionRef,
     },
 };
 
+/// This is a safe default number of blocks to present the canonical chain
+/// at the tip but not too much to overburden a client.
+pub const DEFAULT_NUMBER_OF_BLOCKS_TO_STREAM: usize = 100;
+const DEFAULT_BLOCKS_STREAM_CHUNK_SIZE: usize = 100;
```
strinnityk: I think these constants shouldn't be split from the ones in `paths.rs`, as they are related.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -45,24 +48,188 @@ use lb_tx_service::{
...
+
+    let server_batch_size = non_zero!("server_batch_size", server_batch_size_raw)?;
+
+    Ok(BlocksStreamRequest {
+        number_of_blocks,
+        blocks_to: query.blocks_to,
+        server_batch_size,
+        immutable_only: query.immutable_only.unwrap_or_default(),
+    })
+}
```
strinnityk: I'm feeling all of this can be achieved with the Query entity macros. If so, there's no need to reinvent the wheel here.
If not, then this should be living next to the source Query entity, and probably exposed as a `TryFrom`.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -45,24 +48,188 @@ use lb_tx_service::{
...
+
+async fn fetch_blocks_stream_chunk<StorageBackend, RuntimeServiceId>(
+    handle: &OverwatchHandle<RuntimeServiceId>,
+    chain_info: &lb_chain_service::CryptarchiaInfo,
+    from: usize,
+    to: usize,
+    immutable_only: bool,
+) -> Result<Vec<ApiProcessedBlockEvent>, DynError>
```
strinnityk, line 168: This function takes `chain_info` as parameter and uses it to filter the blocks. It's very redudant to send `ApiProcesedBlockEvent` if one of the params is the `chain_info`, we're effectively sending duplicated information times n.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -678,35 +846,107 @@ where
...
     responses(
-        (status = 200, description = "Stream of processed blocks with chain state"),
+        (status = 200, description = "Stream of processed blocks with chain state. When immutable_only=true and blocks_to is omitted, the stream anchors at LIB by default."),
+        (status = 400, description = "Invalid request parameters", body = String),
+        (status = 404, description = "Requested header not found on canonical chain", body = String),
         (status = 500, description = "Internal server error", body = String),
     )
 )]
-pub async fn blocks_stream<StorageBackend, ConsensusService, RuntimeServiceId>(
+pub async fn blocks_stream<StorageBackend, RuntimeServiceId>(
```
strinnityk: What happens if, while streaming, a reorg happens? We're consuming CPU time serving blocks from a dead fork.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -45,24 +48,188 @@ use lb_tx_service::{
...
+/// at the tip but not too much to overburden a client.
+pub const DEFAULT_NUMBER_OF_BLOCKS_TO_STREAM: usize = 100;
+const DEFAULT_BLOCKS_STREAM_CHUNK_SIZE: usize = 100;
+
+struct BlocksStreamRequest {
+    number_of_blocks: NonZero<usize>,
+    blocks_to: Option<HeaderId>,
+    server_batch_size: NonZero<usize>,
+    immutable_only: bool,
+}
```
danielSanchezQ: What is the difference between this and the `BlocksStreamQuery`?
I have the feeling we are not using the proper tooling here, limits and defaults could go in the query directly?
Also `String` is not a proper error format.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -45,24 +48,188 @@ use lb_tx_service::{
...
+{
+    futures::stream::unfold(
+        (
+            first_chunk.into_iter(),
+            next_from,
+            blocks_to,
+            chunk_size,
+            immutable_only,
+            chain_info,
+            handle,
```
danielSanchezQ: Take it out to a variable.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -45,24 +48,188 @@ use lb_tx_service::{
...
+                else {
+                    return None;
+                };
+
+                next_from = chunk_to.saturating_add(1);
+                buffered = next_chunk.into_iter();
+            }
+        },
+    )
+}
```
danielSanchezQ: Maybe extract a method and make the state a custom struct with named fields.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -678,35 +846,107 @@ where
...
-        .await
-        .map(|stream| stream.map(ApiProcessedBlockEvent::from));
-    match stream {
-        Ok(stream) => responses::ndjson::from_stream(stream),
-        Err(error) => (StatusCode::INTERNAL_SERVER_ERROR, error.to_string()).into_response(),
-    }
+    let request = match parse_blocks_stream_request(&query) {
+        Ok(request) => request,
+        Err(message) => return (StatusCode::BAD_REQUEST, message).into_response(),
+    };
```
danielSanchezQ: There are tooling for this kind of stuff in axum. Use them better.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -678,35 +846,107 @@ where
...
+
+    let chain_info = match consensus::cryptarchia_info::<RuntimeServiceId>(&handle).await {
+        Ok(info) => info,
+        Err(error) => {
+            return (StatusCode::INTERNAL_SERVER_ERROR, error.to_string()).into_response();
```
danielSanchezQ: We should have an error that implements into response instead.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -678,35 +846,107 @@ where
...
+        chain_info.lib
+    } else {
+        chain_info.tip
+    };
+    let blocks_to_header = request.blocks_to.unwrap_or(default_blocks_to_header);
+    let blocks_to_height = match mantle::find_canonical_height_by_header_with_snapshot::<
+        _,
+        _,
+        RuntimeServiceId,
+    >(&handle, blocks_to_header, &chain_info)
```
danielSanchezQ: take off the `find_canonical...` call into a variable, await the in the match later.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -678,35 +846,107 @@ where
...
+        first_chunk_to,
+        request.immutable_only,
+    )
+    .await
+    {
+        Ok(events) => events,
+        Err(error) => {
+            return (StatusCode::INTERNAL_SERVER_ERROR, error.to_string()).into_response();
+        }
+    };
```
danielSanchezQ: Again, better to read if you separate the future call outside in a variable. Rule of thumb is that it should fit in one or two lines tops.

nodes/node/http-client/src/lib.rs
```diff
@@ -255,9 +255,83 @@ impl CommonHttpClient {
...
+        blocks_to: Option<HeaderId>,
+    ) -> Result<impl Stream<Item = ProcessedBlockEvent> + use<>, Error> {
+        self.get_blocks_stream_in_range_with_server_batch_size(
+            base_url,
+            number_of_blocks,
+            blocks_to,
+            None,
+            None,
+        )
+        .await
```
danielSanchezQ: This is probably not necessary, wrapping a call for adding 2 defaults just increase an indirection to understand what is going on.

nodes/node/http-client/src/lib.rs
```diff
@@ -255,9 +255,83 @@ impl CommonHttpClient {
...
+            if let Some(blocks_to) = blocks_to {
+                let blocks_to_hex = hex::encode(blocks_to.as_ref());
+                query.append_pair("blocks_to", &blocks_to_hex);
+            }
+            if let Some(server_batch_size) = server_batch_size {
+                query.append_pair("server_batch_size", &server_batch_size.to_string());
+            }
+            if immutable_only == Some(true) {
+                query.append_pair("immutable_only", "true");
+            }
```
danielSanchezQ: Cant this be a type that then transform into the query params?

nodes/node/http-client/src/lib.rs
```diff
@@ -266,16 +340,64 @@ impl CommonHttpClient {
...
+                        }
+
+                        if let Some(event) = pending.pop_front() {
+                            return Some((event, (byte_stream, Vec::new(), pending)));
+                        }
+                        return None;
+                    }
+                }
+            },
+        );
```
danielSanchezQ: This should be its own stuff.

nodes/node/http-client/src/lib.rs
danielSanchezQ: In fact, isn't this the QueryParams type you already have?

services/api/src/http/mantle.rs
```diff
@@ -257,7 +257,210 @@ where
...
+    <StorageBackend as StorageChainApi>::Tx: From<Bytes> + AsRef<[u8]>,
+    RuntimeServiceId: Debug
+        + Send
+        + Sync
+        + Display
+        + 'static
+        + AsServiceId<StorageService<StorageBackend, RuntimeServiceId>>
+        + AsServiceId<Cryptarchia<RuntimeServiceId>>,
+{
+    if blocks_from == 0 {
```
danielSanchezQ: Why is it nonzero typed then?

tests/testing_framework/src/unique_persistent.rs
```diff
@@ -130,41 +204,55 @@ impl TestPortAllocator {
...
     /// Returns an available UDP port from this allocator's reserved block.
-    fn next_udp_port(&mut self) -> Option<u16> {
+    fn next_udp_port(&mut self) -> Result<u16, ReservedPortError> {
         while self.udp_next <= self.udp_end {
             let port = self.udp_next;
             self.udp_next += 1;
 
             if UdpSocket::bind(("127.0.0.1", port)).is_ok() {
-                return Some(port);
+                return Ok(port);
```
danielSanchezQ: Why is this in this PR?

tests/testing_framework/src/unique_persistent.rs
```diff
@@ -293,12 +383,22 @@ fn try_reap_stale_port_claim_file(path: &Path) {
...
 }
 
 /// Returns an available UDP port from this process's reserved port block.
 #[must_use]
 pub fn get_reserved_available_udp_port() -> Option<u16> {
+    try_get_reserved_available_udp_port().ok()
+}
+
+/// Returns an available UDP port from this process's reserved port block.
+pub fn try_get_reserved_available_udp_port() -> Result<u16, ReservedPortError> {
```
danielSanchezQ: I don't understand why we have so many changes not related to the scope of the PR

utils/src/types.rs
```diff
@@ -68,3 +68,10 @@ pub mod enumerated {
         }
     }
 }
+
+#[macro_export]
+macro_rules! non_zero {
+    ($field:expr, $value:expr) => {
+        std::num::NonZero::new($value).ok_or_else(|| format!("'{}' must be > 0", $field))
+    };
+}
```
danielSanchezQ: This is not doing what it should do, the name is misguiding, `non_zero` says nothing on what this do.

nodes/node/http-client/src/lib.rs
```diff
@@ -255,9 +255,83 @@ impl CommonHttpClient {
...
+        }
+
+        Ok(())
+    }
+
+    /// Stream processed blocks ending at a specific header id.
+    ///
+    /// `server_batch_size` lets callers request smaller chunks; the server
+    /// still enforces its own upper bound.
+    pub async fn get_blocks_stream_in_range_with_server_batch_size(
```
strinnityk: We can't be having 90-lines-functions unless there's a very specific reason as to why it can't be done any other way.

danielSanchezQ (changes requested):
This lacks a lot of quality and clarity. Its really difficult to follow and some decisions code-wise were not in the proper direction. I feel it could have been spitted in at least 2 PRs.
Feels like a lot AI slop. No personal grudges, its just not enough.

nodes/node/http-client/src/lib.rs
```diff
@@ -266,16 +340,64 @@ impl CommonHttpClient {
...
+            };
         }
+
+        let blocks_stream = futures::stream::unfold(
+            (
+                response.bytes_stream(),
+                Vec::<u8>::new(),
+                VecDeque::<ProcessedBlockEvent>::new(),
+            ),
+            async |(mut byte_stream, mut buffer, mut pending)| {
```
strinnityk: I think there might be a crate that handles this already. Not implying this is wrong, just trying to avoid manually reinventing the wheel.

services/api/src/http/mantle.rs
```diff
@@ -257,7 +257,210 @@ where
     Ok(new_blocks_stream)
 }
 
-/// Fetch block header ids in range.
+/// Fetch canonical blocks in the requested height range and annotate each block
+/// with a consistent chain snapshot (tip/LIB) taken at request time.
+pub async fn get_blocks_in_range<Transaction, StorageBackend, RuntimeServiceId>(
```
strinnityk: YAGNI: This is not used.

services/api/src/http/mantle.rs
```diff
@@ -257,7 +257,210 @@ where
...
+        + 'static
+        + AsServiceId<StorageService<StorageBackend, RuntimeServiceId>>
+        + AsServiceId<Cryptarchia<RuntimeServiceId>>,
+{
+    if blocks_from == 0 {
+        return Err("blocks_from must be greater than or equal to 1".into());
+    }
+    if blocks_to < blocks_from {
+        return Err("blocks_to must be greater than or equal to blocks_from".into());
+    }
```
strinnityk: Duplicated, validation is done in `get_blocks_in_range_with_snapshot`.

services/api/src/http/mantle.rs
```diff
@@ -257,7 +257,210 @@ where
...
+            });
+        }
+
+        if parent_id == current_id {
+            break;
+        }
+
+        current_id = parent_id;
+        current_height = current_height.saturating_sub(1);
+    }
```
strinnityk, line 459: I think this would benefit from being implemented as a Stream.

services/api/src/http/mantle.rs
strinnityk: Is there a use case where this `try_from` legitimately fails?

services/api/src/http/mantle.rs
hansieodendaal: Fn removed, now slot based

nodes/node/binary/src/api/handlers.rs
strinnityk: IIUC, we have two endpoints:
- `immutable_blocks`: Fetch blocks older than `lib`.
- `blocks_stream`: Stream blocks, but we can parametrise the range of the streamed blocks.

Can you ELI5 `blocks_stream` behaviour? 

Also, the design intention of both and how do they compliment each other?

hansieodendaal: - As discussed, this request is changing to be slot based.
- Currently when we retrieve blocks above LIB, it also comes from storage - see `pub async fn block<HttpStorageAdapter, RuntimeServiceId>(`.
- `immutable_only=true` must behave the same way as `pub async fn immutable_blocks<StorageBackend, RuntimeServiceId>(`.
- We could consider to deprecate the legacy immutable-blocks endpoint, as the new streaming impl will do all and more, but his has not been the purpose of the current PR, just a spin-off.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -45,24 +48,188 @@ use lb_tx_service::{
...
+    )
+    .await?;
+
+    Ok(chunk
+        .into_iter()
+        .map(ApiProcessedBlockEvent::from)
+        .collect())
+}
+
+fn build_blocks_stream<StorageBackend, RuntimeServiceId>(
```
strinnityk, line 214: The way this is getting called results in an O(N^2).
Whenever we call this, we begin traversing back from `tip` until we find the height.
This means that if we're going back 1_000_000 blocks to get the first chunk, for the second chunk we'll need tro traverse (almost) all the way back, and so on and so forth.

services/api/src/http/mantle.rs
strinnityk: If we're fetching `immutable_only` blocks we don't need to begin traversing at tip, or even doing conditional insertion. We can just get `chain_info.lib` or `chain_info.tip` conditionally and thus avoid further checks and unnecessary traversing.

strinnityk (commented):
I have separate concerns here:
First, I don't think it's up to the code quality that we should be putting out. I believe there's some pieces of code that are quite convoluted, even quite confusing. I've flagged those with inline comments.
Second, while I'm in totally in favour of the boy-scout rule, I think there's a line between tidying up minor things and effectively smuggling a second PR in. To me, the port handling changes cross that line.

hansieodendaal:
> I think there's a line between tidying up minor things and effectively smuggling a second PR in. To me, the port handling changes cross that line.

Agreed - these came about hunting down errors when we had an unstable code base, I will remove it.

nodes/api-common/src/paths.rs
hansieodendaal: > Why does block streaming have maximum streaming?

That was a prior comment, to not have `usize:MAX`. Ah.. this is not a constant stream, just an efficient way to get many blocks within a start slot and an end slot. Blocks are streamed, not bundled and delivered in a single JSON body.

nodes/api-common/src/paths.rs
hansieodendaal: Reverting this to keep the interface the same, as per topic discussion

nodes/node/binary/src/api/handlers.rs
hansieodendaal: Agreed, this has been changed in the new slot-based apprpoach

nodes/node/binary/src/api/handlers.rs
hansieodendaal: Yes, a lot of clean-up has been done here

utils/src/types.rs
hansieodendaal: Removed

tests/testing_framework/src/unique_persistent.rs
hansieodendaal: Agreed - these came about hunting down errors when we had an unstable code base, I will remove it.

tests/testing_framework/src/unique_persistent.rs
hansieodendaal: Agreed - these came about hunting down errors when we had an unstable code base, I will remove it.

nodes/node/binary/src/api/handlers.rs
hansieodendaal: Refactored.

`BlocksStreamQuery` is the HTTP/serde model and mirrors incoming query params, while `BlocksStreamRequest` is the fully validated/normalized internal model used by handler logic, intentionally kept separate so handler code never re-checks parse-time invariants.

bacv (commented):
I think the complexity of query makes the scope of PR bigger than necessary. The previous block streaming solution was based on storage service returning the stream, api service applying a filter. I think this was a pretty robust and simple solution. To achieve this for ranged block streams we need to implement a query mechanism in the storage layer, and in the api layer only pass the stream through the ndjson crate for serialization.

I think a query could be simplified by removing the `server_batch_size` and `immutable_only`. 
- `server_batch_size` should be replaced with server side config, or constant, this is a stream so lets just set it to 1?
- `immutable_only` should have a dedicated endpoint.

If we can open a PR with all the suggested comments focusing only on the latest block streaming, it will be easier to adjust it according to the suggestions and later do the same for the immutable blocks with minor changes.
I also suggest to write our own ndjson crate that the same writer as in `rojul/ndjson`, but also contains reader implementation that could be used inside the http client, current solution seems pretty verbose and not modular, our crate would have all required parts in one place.

nodes/node/binary/src/api/handlers.rs
hansieodendaal: Not following you, the chain state information embedded in `ApiProcessedBlockEvent` change with every block, not so?

nodes/node/binary/src/api/handlers.rs
hansieodendaal: Any blocks up to LIB will have no impact, any blocks above LIB that are reorged out will cause the stream construction to error in the backend when those blocks cannot be found, if the canonical blockchain is boken.

nodes/node/http-client/src/lib.rs
hansieodendaal: Refactored

nodes/node/http-client/src/lib.rs
hansieodendaal: Found it

services/api/src/http/mantle.rs
hansieodendaal: Removed in new slot-base version

services/api/src/http/mantle.rs
hansieodendaal: Refactored

services/api/src/http/mantle.rs
hansieodendaal: Refactored - removed in slot-based version

services/api/src/http/mantle.rs
hansieodendaal: Yes, the new approach is slot-based, when fetching `immutable_only` blocks (any/all blocks up to LIB) a direct lookup is done to get the header_ids.

services/api/src/http/mantle.rs
hansieodendaal: Refactored out

services/api/src/http/mantle.rs
hansieodendaal: Refactored out in the new slot-based approach

services/api/src/http/mantle.rs
hansieodendaal: Refactored out in the new slot-based approach

nodes/node/binary/src/api/handlers.rs
hansieodendaal: Refacto added explicit streamed error propagation.

nodes/node/binary/src/api/handlers.rs
hansieodendaal: Refactored

nodes/node/binary/src/api/handlers.rs
hansieodendaal: Refactored

nodes/node/binary/src/api/handlers.rs
hansieodendaal: Refactored

nodes/node/binary/src/api/handlers.rs
hansieodendaal: Refactored

nodes/node/binary/src/api/handlers.rs
hansieodendaal: Refactored

nodes/node/binary/src/api/handlers.rs
hansieodendaal: Refactored

nodes/node/http-client/src/lib.rs
hansieodendaal: Refactored

nodes/node/http-client/src/lib.rs
hansieodendaal: Refactored

nodes/node/http-client/src/lib.rs
hansieodendaal: Refactored

hansieodendaal:
@andrussal @pradovic @danielSanchezQ @strinnityk please have a re-look at the PR, I have addressed all the comments.

**Note:** @bacv  I have not implemented any of your comments yet, as I have reworked the PR considerably.

We could change direction, adapt the "_flexible_" streaming API requirement or close this and start something else. For the current implementation, the extensive block streaming test cases all pass.

tests/src/tests/cryptarchia/blocks_streaming.rs
```diff
@@ -0,0 +1,815 @@
...
+                .cryptarchia
+                .service
+                .bootstrap
+                .prolonged_bootstrap_period = Duration::ZERO;
+            config.deployment.cryptarchia.security_param = NonZero::new(7).unwrap();
+            config
+        })
+        .collect::<Vec<_>>();
+
+    let nodes = futures_util::future::join_all(configs.into_iter().map(Validator::spawn))
```
andrussal: We should spawn with tf instead as we remov old spawn logic soon.

nodes/api-common/src/lib.rs
```diff
@@ -9,3 +9,15 @@ pub mod settings;
...
+/// and memory use
+pub const MAX_BLOCKS_STREAM_CHUNK_SIZE: usize = 1_000;
+/// This is a safe default chunk size for streaming blocks, allowing for
+/// efficient delivery without overburdening the server or client.
+pub const DEFAULT_BLOCKS_STREAM_CHUNK_SIZE: usize = 100;
+/// 200 years worth of blocks if 1 is produced every 10s
+pub const MAX_BLOCKS_STREAM_BLOCKS: usize = 630_720_000;
+/// This is a safe default number of blocks to present the canonical chain
+/// at the tip but not too much to overburden a client.
+pub const DEFAULT_NUMBER_OF_BLOCKS_TO_STREAM: usize = 100;
```
danielSanchezQ, line 23: At first glance this feels unnecesary.

nodes/node/binary/src/api/handlers.rs
```diff
@@ -55,14 +58,253 @@ use tokio_stream::StreamExt as _;
...
+                // For ascending streams without an explicit slot_from, estimate a narrow
+                // window ending near slot_to. This prioritizes ending close to slot_to over
+                // guaranteeing blocks_limit returned blocks.
+                let estimated_slot_span = (request.blocks_limit.get() as u64)
+                    .saturating_mul(average_slots_per_block.max(1))
+                    .saturating_mul(ASCENDING_ESTIMATED_WINDOW_NOM)
+                    / ASCENDING_ESTIMATED_WINDOW_DEM;
+
+                slot_to.saturating_sub(Slot::new(estimated_slot_span))
+            }
```
danielSanchezQ: Better as `let / else`

nodes/node/binary/src/api/handlers.rs
```diff
@@ -55,14 +58,253 @@ use tokio_stream::StreamExt as _;
...
+        },
+        Slot::new,
+    );
+    if slot_from > slot_to {
+        return Err(format!(
+            "'slot_from' must be <= 'slot_to', got slot_from={}, slot_to={}",
+            slot_from.into_inner(),
+            slot_to.into_inner()
```
danielSanchezQ: Can we have proper errors with `thiserror` instead of returning a formatted string here (and everywhere)

nodes/node/binary/src/api/handlers.rs
```diff
@@ -55,14 +58,253 @@ use tokio_stream::StreamExt as _;
...
+            .await
+            {
+                Ok(next_chunk) => next_chunk,
+                Err(error) => {
+                    // Terminal error: avoid re-emitting the same error forever.
+                    state.remaining = 0;
+                    state.next_cursor = None;
+                    return Some((Err(error), state));
+                }
+            };
```
danielSanchezQ: Again (from my previous review). Take the call outside of the match:
```suggestion
            let fetched_blocks = etch_blocks_stream_chunk::<StorageBackend, RuntimeServiceId>(
                &state.handle,
                &state.chain_info,
                chunk_from,
                chunk_to,
                state.descending,
                request_limit,
                state.immutable_only,
            )
            .await;
            let next_chunk = match fetched blocks {
                Ok(next_chunk) => next_chunk,
                Err(error) => {
                    // Terminal error: avoid re-emitting the same error forever.
                    state.remaining = 0;
                    state.next_cursor = None;
                    return Some((Err(error), state));
                }
            };
```

nodes/node/binary/src/api/queries.rs
```diff
@@ -10,3 +19,261 @@ pub struct BlockRangeQuery {
...
+    /// Sort direction. Defaults to descending (`true`).
+    #[serde(default)]
+    pub descending: Option<bool>,
+    /// The maximum number of actual blocks to return. If omitted:
+    /// - explicit bounded slot range (`slot_from` and `slot_to`) defaults to
+    ///   the server maximum (`630_720_000`);
+    /// - otherwise defaults to `100`.
+    #[serde(default)]
+    #[validate(custom(function = "validate_blocks_limit"))]
+    #[param(minimum = 1, maximum = 630_720_000, default = 100, example = 100)]
```
danielSanchezQ, line 52: Can't we use the defined constants here? We have values already defined in constants but here use hardcoded ones.

nodes/node/binary/src/api/queries.rs
```diff
@@ -10,3 +19,261 @@ pub struct BlockRangeQuery {
...
+    pub blocks_limit: Option<usize>,
+    /// Server chunk size hint for streamed delivery. Defaults to `100` ,
+    /// maximum `1000`.
+    #[serde(default)]
+    #[validate(custom(function = "validate_server_batch_size"))]
+    #[param(minimum = 1, maximum = 1_000, default = 100, example = 100)]
```
danielSanchezQ, line 58: Same here?

nodes/node/binary/src/api/queries.rs
danielSanchezQ, line 52: Doesnt the values get validated from the param section already?

services/api/src/http/mantle.rs
```diff
@@ -257,7 +258,372 @@ where
...
+        TryFrom<Block<Transaction>> + TryInto<Block<Transaction>>,
+    <StorageBackend as StorageChainApi>::Tx: From<Bytes> + AsRef<[u8]>,
+    RuntimeServiceId: Debug
+        + Send
+        + Sync
+        + Display
+        + 'static
+        + AsServiceId<StorageService<StorageBackend, RuntimeServiceId>>,
+{
+    let mut blocks = Vec::with_capacity(header_ids.len().min(blocks_limit));
```
danielSanchezQ, line 330: Why not a stream instead?

services/api/src/http/mantle.rs
```diff
@@ -257,7 +258,372 @@ where
...
+}
+
+fn validate_blocks_slot_range(
+    slot_from: Slot,
+    slot_to: Slot,
+    immutable_only: bool,
+    chain_info: &CryptarchiaInfo,
+) -> Result<(), super::DynError> {
+    if slot_to < slot_from {
+        return Err("slot_to must be greater than or equal to slot_from".into());
```
danielSanchezQ: Proper errors would be better imo.

services/api/src/http/mantle.rs
```diff
@@ -257,7 +258,372 @@ where
...
+
+        if parent_id == current_id {
+            break;
+        }
+        current_id = parent_id;
+    }
+
+    if !descending {
+        blocks.reverse();
+        blocks.truncate(limit);
```
danielSanchezQ, line 463: truncating should just not happen, instead break the loop when reached the limit (?)

services/api/src/http/mantle.rs
```diff
@@ -257,7 +258,372 @@ where
...
+{
+    validate_blocks_slot_range(slot_from, slot_to, immutable_only, chain_info)?;
+
+    let relay = handle
+        .relay::<StorageService<StorageBackend, RuntimeServiceId>>()
+        .await?;
+    let storage_adapter =
+        StorageAdapter::<StorageBackend, Transaction, RuntimeServiceId>::new(relay).await;
+
+    let mut blocks = Vec::with_capacity(blocks_limit.get().min(1024));
```
danielSanchezQ, line 551: Why not a stream too?

services/api/src/http/mantle.rs
```diff
@@ -257,7 +258,372 @@ where
...
+                    chain_info,
+                    mutable_slot_from,
+                    slot_to,
+                    remaining,
+                    false,
+                )
+                .await?;
+            blocks.extend(mutable_blocks);
+        }
+    }
```
danielSanchezQ, line 615: Can we use match here instead of nested blocks?

```rust
match (descending, remaining, has_mutable_range) {
...
}
```

services/storage/src/api/backend/rocksdb/chain.rs
```diff
@@ -146,6 +146,28 @@ impl StorageChainApi for RocksBackend {
...
+                Some(&start_key),
+                Some(&end_key),
+                Some(limit),
+            )
+            .await?;
+
+        result
+            .into_iter()
+            .map(|bytes| bytes.as_ref().try_into().map_err(Into::into))
+            .collect::<Result<Vec<HeaderId>, Error>>()
```
danielSanchezQ: You could probably return an iterator and avoid one extra collect+allocation

services/storage/src/backends/rocksdb.rs
```diff
@@ -210,6 +210,52 @@ impl StorageBackend for RocksBackend {
...
+        prefix: &[u8],
+        start_key: Option<&[u8]>,
+        end_key: Option<&[u8]>,
+        limit: Option<NonZeroUsize>,
+    ) -> Result<Vec<Bytes>, <Self as StorageBackend>::Error> {
+        let mut values = Vec::new();
+
+        let start_key =
+            start_key.map(|k| prefix.iter().chain(k.iter()).copied().collect::<Vec<_>>());
+        let end_key = end_key.map(|k| prefix.iter().chain(k.iter()).copied().collect::<Vec<_>>());
```
danielSanchezQ: Don't we have methods to build the keys already? 🤔

services/storage/src/backends/rocksdb.rs
```diff
@@ -210,6 +210,52 @@ impl StorageBackend for RocksBackend {
...
+
+                    if let Some(limit) = limit
+                        && values.len() == limit.get()
+                    {
+                        break;
+                    }
+                }
+                Err(e) => return Err(e),
+            }
+        }
```
danielSanchezQ, line 253: Nitpick: This could be a `take_while` probably, then just collect.

nodes/node/binary/src/api/queries.rs
hansieodendaal: Constants are not allowed as it breaks the `utopia::IntoParams` derive

hansieodendaal: `#[param]` is for OpenAPI schema generation/docs only, and needs to be duplicated due to utoipa attribute limitations.
Runtime request validation is done by custom functions (`validate_blocks_limit`, `validate_server_batch_size`) and `query.validate()?`.

nodes/node/binary/src/api/queries.rs
hansieodendaal: See above

services/api/src/http/mantle.rs
hansieodendaal: I think we should keep `load_blocks_with_chain_state_by_ids` as a batch loader, not an internal stream.

The stream belongs at the API boundary (HTTP NDJSON), while the storage path should stay batch-oriented for throughput. In this code path, we fetch multiple blocks into a `Vec` per cycle and return that batch; if more data is needed, the next cycle fetches the next batch. This avoids per-item stream overhead in the storage layer and aligns with the current limit/pagination flow (`remaining`, chunked fetches).

```
Internal (batch):
  get_immutable_block_ids_in_slot_range()
    → load_blocks_with_chain_state_by_ids() [Vec batch]
    → fetch_and_load_immutable_blocks() [Vec batch]

API boundary (stream):
  blocks_range_stream() handler
    → Stream<Item = Result<ApiProcessedBlockEvent, DynError>>
```

services/api/src/http/mantle.rs
hansieodendaal: This is for ascending ordering, and ascending is always anchored at `slot_from` if specified. We cannot know the number of blocks returned before we get to `slot_from`, as for this mutable block range we have to walk newest to oldest and start counting from `slot_from` up to `slot_to`.

services/api/src/http/mantle.rs
hansieodendaal: This is the same reason as earlier; we fetch a chunked number of blocks from storage then stream it to the client, repeating until the contract is fulfilled.

services/api/src/http/mantle.rs
hansieodendaal: Good suggestion. I considered different tuple-matching approaches, but this flow has two sequential phases and `remaining` is mutated after the first one. In practice, the tuple-match versions either duplicated logic or made the state transition harder to follow.

I simplified the code a bit for readability, but I kept the explicit `if` flow because it matches the execution order directly:
- descending: mutable -> immutable
- ascending: immutable -> mutable

services/storage/src/api/backend/rocksdb/chain.rs
hansieodendaal: Both `scan_immutable_block_ids` and `scan_immutable_block_ids_reverse` return `Result<Vec<HeaderId>, _>`, so the final `collect` is required by the current API shape.

Given that, we can’t remove that allocation here unless we change either:
- the lower-level loading API (`load_prefix` / `load_prefix_reverse`) to be lazy, or
- the trait methods to return something other than `Vec`.

I would keep this as-is for consistency with the forward scan path, and handle any API-level optimization in a separate PR.

Wdyt?

services/storage/src/backends/rocksdb.rs
hansieodendaal: I added a new helper for this: `pub fn key_bytes_raw(prefix: &[u8], suffix: impl AsRef<[u8]>) -> Vec<u8> {`

services/storage/src/backends/rocksdb.rs
hansieodendaal: I’m not sure we’d keep identical boundary semantics with `take_while` here.  
Given that, we could keep the explicit loop for now and handle a unified streaming refactor of both `load_prefix` and `load_prefix_reverse` in the follow-up PR mentioned above^ ?

nodes/api-common/src/lib.rs
hansieodendaal: This enables short period burst backend reads with follow-up client streaming.

services/storage/src/api/backend/rocksdb/chain.rs
danielSanchezQ: This is just a method, result is alredy allocated and have ownership, you can just return 
```rust
result
            .into_iter()
            .map(|bytes| bytes.as_ref().try_into().map_err(Into::into))
```
and avoid the extra allocation, then if the caller decides it needs a vec can collet it himself.

services/storage/src/backends/rocksdb.rs
danielSanchezQ: Something like:
```rust
iter.take_while(|(key, value)| !key.starts_with(prefis) || matches!(start_key, Some(start_key) if key.as_ref() < start_key.as_slice())).take(limit.unwrap_or(usize::MAX)).collect

```

danielSanchezQ (approved):
I left some more comments.... I'm pre-approving this but imo its an overly complicated PR to follow.
Image
I wish it would have been easier to review.

services/storage/src/api/backend/rocksdb/chain.rs
hansieodendaal: Ok, refactored

services/storage/src/backends/rocksdb.rs
hansieodendaal: For now, I’d keep the explicit loop here. It is clearer and less error-prone, especially around the prefix/start-key boundary and RocksDB error propagation.

For real async streaming, we would probably want to run the RocksDB iterator on `spawn_blocking` and send items through a channel. That is more code and only really worth it if we need to stream very large ranges without collecting a `Vec`.

For this PR, we already mitigated the memory concern with the limited range/chunking implementation, so I’d prefer to keep this explicit for now and revisit a unified streaming refactor for `load_prefix` / `load_prefix_reverse` separately.

danielSanchezQ: Ok. For streaming ideally we do not need to send things to a channel nor spawn_blocking, we just encapsulate the stream pull. Just as a note for future implementation.

PR #2630, opened 2026-04-27
fix(test): avoid deleting fresh port claims

andrussal:

This fixes a race in `TestPortAllocator`.

The lock file creation itself is atomic, but there was still a small window after creating the file and before writing metadata into it. During that window another test process could see an empty/partial claim file, treat it as malformed, delete it, and then claim the same port block.

This PR makes malformed claim cleanup more conservative: fresh malformed claim files are left alone and only old malformed files are deleted. That way another process skips the block instead of accidentally stealing it while the original owner is still writing metadata.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/tools/config/src/unique.rs
+++ b/tools/config/src/unique.rs
@@ -7,7 +7,7 @@ use std::{
     path::{Path, PathBuf},
     process::Command,
     sync::{Mutex, OnceLock},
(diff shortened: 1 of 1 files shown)


pradovic (approved):
LGTM!

PR #2632, opened 2026-04-27
fix(tracing): respect env filter when sending logs to loki, gelf, or otlp

ntn-x2:

I noticed we were pushing really spammy log entries, and realized we did not respect the env filter set when sending logs to remote logging services.

Yes.

@ntn-x2 

N/A

No.


bacv (approved):
Good catch!

PR #2430, opened 2026-03-26
chore: reduce and standardize node logging

andrussal:
## 1. What does this PR implement?

This PR standardises log levels across the node codebase.

The main goal of this PR is to make log levels more consistent across the codebase. In practice, that mostly means being stricter about what belongs in `info`, `debug`, and `trace`, so `debug` is more useful during normal investigation and less dominated by lower-level detail.

One result of that cleanup is smaller log files, mainly because a lot of high-frequency logging moved down from `info`/`debug` into `debug`/`trace`.

It also makes future logging easier to maintain. By making the levels more consistent, it becomes easier to decide where new logs belong and to spot places where useful logs are still missing.

Most of the judgment in this PR is around `debug` vs `trace`. In a few places that line is naturally debatable, especially where low-level logs can still be useful during investigation. If any of those look off, please call them out.

Log levels in this PR are intended to mean:

- `error`: real failures that need attention
- `warn`: something is off, but the node is still running
- `info`: important lifecycle and high-level state changes
- `debug`: useful diagnostics for understanding node behavior
- `trace`: detailed internal flow, protocol, and transport logging

**What changed**
- moved high-frequency internal logs from `debug` to `trace`
- replaced large struct dumps with shorter, field-based log messages
- added a default filter policy for verbose logging levels
- kept compact aggregate flow logs at `debug` where they are still useful for normal investigation

It also reduces log file size significantly in the verification workload:
- before: ~10.5 MB / 10 min / node
- after: ~167 KB / 10 min / node
- reduction: ~98.4%

**Largest removed categories**

The biggest reductions in emitted log volume in the strict 10-minute `DEBUG` comparison were:
- `blend::*`: ~29.6k fewer log lines (~25.6% of all removed log lines)
- `libp2p` swarm internals: ~23.6k fewer log lines (~20.4%)
- Blend connection lifecycle logs: ~21.4k fewer log lines (~18.6%)
- `logos_blockchain_blend_message::*`: ~10.4k fewer log lines (~9.0%)
- `logos_blockchain_blend_service::*`: ~6.5k fewer log lines (~5.6%)
- `overwatch::*`: ~3.8k fewer log lines (~3.3%)
- `logos_blockchain_blend_scheduling::*`: ~1.5k fewer log lines (~1.3%)
- `logos_blockchain_network_service::*`: ~1.2k fewer log lines (~1.0%)
- `logos_blockchain_chain_network_service::*`: ~0.6k fewer log lines (~0.5%)

**Dependency logging**

[...]

PR #2651, opened 2026-04-29
feat(logging): introduce log targets crate

andrussal:

This PR adds a dedicated crate for log targets.

It follows the earlier discussion here:
- [chore: reduce and standardize node logging #2430 (comment)](https://github.com/logos-blockchain/logos-blockchain/pull/2430#discussion_r2993130299)

So far targets have mostly been plain string literals spread around the codebase. That makes them easy to drift, and it also means config cannot be checked against any known set of valid targets. This change moves target definitions into one place so the same definitions can be reused both from code and when validating configured log filters.

Target declarations look like this:

```rust
log_targets! {
    blend::{
        service::{CORE, core::KMS_POQ_GENERATOR},
        network::core::handler::{CORE_EDGE},
    }
}
```

Call sites then use the generated constants directly:

```rust
const LOG_TARGET: &str = lb_log_targets::blend::service::CORE;

tracing::debug!(target: LOG_TARGET, "...");
```

A proc macro is used here so target definitions only need to be written once, while still generating both the constants used by code and the collected target list used for validation. A normal macro could do this too, but the declaration format and expansion logic became harder to read.

This PR only switches a couple of Blend call sites as examples. The rest can be moved over in follow-up PRs.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,8 @@ members = [
   "tools/blockchain-tools",
   "tools/config",
   "tracing",
+  "tracing/targets",
+  "tracing/targets/macros",
   "utils",
   "utxotree",
   "wallet",
@@ -129,6 +131,7 @@ lb-key-management-system-operators = { default-features = false, package = "logo
 lb-key-management-system-service   = { default-features = false, package = "logos-blockchain-key-management-system-service", path = "./services/key-management-system" }
 lb-ledger                          = { default-features = false, package = "logos-blockchain-ledger", path = "./ledger" }
 lb-libp2p                          = { default-features = false, package = "logos-blockchain-libp2p", path = "./libp2p" }
+lb-log-targets                     = { default-features = false, package = "logos-blockchain-log-targets", path = "./tracing/targets" }
 lb-mmr                             = { default-features = false, package = "logos-blockchain-mmr", path = "./mmr" }
 lb-network-service                 = { default-features = false, package = "logos-blockchain-network-service", path = "./services/network" }
 lb-node                            = { default-features = false, package = "logos-blockchain-node", path = "./nodes/node/binary" }
@@ -168,6 +171,7 @@ blake2            = { default-features = false, version = "0.10" }
 bytes             = { default-features = false, version = "1.3" }
 cached            = { default-features = false, version = "0.55.1" }
 cfg-if            = { default-features = false, version = "1.0.4" }
+const-str         = { default-features = false, version = "0.4.3" }
 divan             = { default-features = false, version = "0.1" }
 ed25519-dalek     = { default-features = false, version = "2" }
 fork_stream       = { default-features = false, version = "0.1.0" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -132,6 +132,7 @@ lb-key-management-system-service   = { default-features = false, package = "logo
 lb-ledger                          = { default-features = false, package = "logos-blockchain-ledger", path = "./ledger" }
 lb-libp2p                          = { default-features = false, package = "logos-blockchain-libp2p", path = "./libp2p" }
 lb-log-targets                     = { default-features = false, package = "logos-blockchain-log-targets", path = "./tracing/targets" }
+lb-log-targets-macros              = { default-features = false, package = "logos-blockchain-log-targets-macros", path = "./tracing/targets/macros" }
 lb-mmr                             = { default-features = false, package = "logos-blockchain-mmr", path = "./mmr" }
 lb-network-service                 = { default-features = false, package = "logos-blockchain-network-service", path = "./services/network" }
 lb-node                            = { default-features = false, package = "logos-blockchain-node", path = "./nodes/node/binary" }
@@ -185,6 +186,8 @@ log               = { default-features = false, version = "0.4" }
 nutype            = { default-features = false, version = "0.6.2" }
 overwatch         = { default-features = false, git = "https://github.com/logos-co/Overwatch", rev = "448c192" }
 overwatch-derive  = { default-features = false, git = "https://github.com/logos-co/Overwatch", rev = "448c192" }
+proc-macro2       = { default-features = false, version = "1" }
+quote             = { default-features = false, version = "1" }
 rand              = { default-features = false, version = "0.8" }
 reqwest           = { default-features = false, version = "0.12" }
 serde             = { default-features = false, version = "1.0" }
@@ -194,6 +197,7 @@ serde_json        = { default-features = false, version = "1.0" }
 serde_with        = { default-features = false, version = "3.14.0" }
 serde_yaml        = { default-features = false, version = "0.9.33" }
 subtle            = { default-features = false, version = "2.6.1" }
+syn               = { features = ["full"], version = "2" }
 tempfile          = { default-features = false, version = "3" }
 thiserror         = { default-features = false, version = "2.0" }
 time              = { default-features = false, version = "0.3" }
--- a/services/blend/Cargo.toml
+++ b/services/blend/Cargo.toml
@@ -25,7 +25,7 @@ lb-cryptarchia-engine            = { workspace = true }
 lb-groth16                       = { default-features = false, workspace = true }
 lb-key-management-system-service = { features = ["unsafe"], workspace = true }
 lb-ledger                        = { workspace = true }
-lb-libp2p                        = { optional = true, workspace = true }
+lb-libp2p                        = { workspace = true }
 lb-log-targets                   = { workspace = true }
 lb-network-service               = { workspace = true }
 lb-poq                           = { workspace = true }
--- a/services/blend/Cargo.toml
+++ b/services/blend/Cargo.toml
@@ -25,7 +25,8 @@ lb-cryptarchia-engine            = { workspace = true }
 lb-groth16                       = { default-features = false, workspace = true }
 lb-key-management-system-service = { features = ["unsafe"], workspace = true }
 lb-ledger                        = { workspace = true }
-lb-libp2p                        = { workspace = true }
+lb-libp2p                        = { optional = true, workspace = true }
+lb-log-targets                   = { workspace = true }
 lb-network-service               = { workspace = true }
 lb-poq                           = { workspace = true }
 lb-sdp-service                   = { workspace = true }
--- a/tracing/src/filter/envfilter.rs
+++ b/tracing/src/filter/envfilter.rs
@@ -13,6 +13,7 @@ const DEFAULT_DEBUG_TARGETS: &[&str] = &[
     "cryptarchia",
     "ledger",
 ];
+const ENVFILTER_GLOBAL_TARGET: &str = "*";
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct EnvFilterConfig {
@@ -20,7 +21,7 @@ pub struct EnvFilterConfig {
     ///
     /// The global default directive is represented internally with the `*`
     /// key and converted back into native `EnvFilter` syntax at the boundary.
-    #[serde(with = "serde_filters")]
+    #[serde(with = "serde_validated_filters")]
     pub filters: HashMap<String, Level>,
 }
 
@@ -42,7 +43,7 @@ pub fn default_envfilter_config(level: Level) -> Option<EnvFilterConfig> {
 #[must_use]
 /// Builds the default verbose filter policy as a typed map.
 pub fn default_debug_log_filter(level: Level) -> HashMap<String, Level> {
-    let mut filters = HashMap::from([("*".to_owned(), Level::WARN)]);
+    let mut filters = HashMap::from([(ENVFILTER_GLOBAL_TARGET.to_owned(), Level::WARN)]);
     filters.extend(
         DEFAULT_DEBUG_TARGETS
             .iter()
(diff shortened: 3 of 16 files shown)


Cargo.toml
```diff
@@ -129,6 +131,7 @@ lb-key-management-system-operators = { default-features = false, package = "logo
 lb-key-management-system-service   = { default-features = false, package = "logos-blockchain-key-management-system-service", path = "./services/key-management-system" }
 lb-ledger                          = { default-features = false, package = "logos-blockchain-ledger", path = "./ledger" }
 lb-libp2p                          = { default-features = false, package = "logos-blockchain-libp2p", path = "./libp2p" }
+lb-log-targets                     = { default-features = false, package = "logos-blockchain-log-targets", path = "./tracing/targets" }
```
ntn-x2, line 134: Why not also the macros crate?

tracing/targets/macros/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+name        = "logos-blockchain-log-targets-macros"
+readme      = { workspace = true }
+repository  = { workspace = true }
+version     = { workspace = true }
+
+[lib]
+proc-macro = true
+
+[dependencies]
+proc-macro2 = { default-features = false, version = "1" }
```
ntn-x2: Since I'm working on a related PR: can we move these to workspace deps as well? So we keep everything in a single place.

tracing/targets/src/lib.rs
```diff
@@ -0,0 +1,78 @@
+#![forbid(unsafe_code)]
```
ntn-x2: Why? Ideally this will be a workspace-wide prefix with a few allows like for the C bindings.

tracing/targets/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+
+pub mod blend;
+
+#[must_use]
+pub(crate) fn matches_target_prefix(target: &str, candidate: &str) -> bool {
+    target == candidate
+        || candidate
+            .strip_prefix(target)
+            .is_some_and(|suffix| suffix.starts_with("::"))
```
ntn-x2: This assumes namespaces are divided by `::`. Is this documented anywhere?

ntn-x2: Also I think it would make sense to define the delimiter as a const instead of repeating it in a few places.

tracing/targets/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+#[must_use]
+fn target_root(target: &str) -> &str {
+    target.split("::").next().unwrap_or(target)
+}
+
+#[must_use]
+pub fn all_targets() -> Vec<&'static str> {
```
ntn-x2: I think this should rather return a `HashSet`, since targets cannot be duplicated. That would also make searching faster (marginal, yes, but still nice)

tracing/targets/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+    blend::all_targets()
+}
+
+#[must_use]
+pub fn is_valid_target(target: &str) -> bool {
```
ntn-x2: As far as I've seen this is only used in tests, why not adding it in the tests module directly?

tracing/targets/src/blend.rs
```diff
@@ -0,0 +1,25 @@
+use logos_blockchain_log_targets_macros::log_targets;
+
+log_targets! {
```
ntn-x2, line 3: I am not sure I like this format. It is prone to errors and introduces a lot of duplications for nested structures. They might be easier to read (although I don't have a full example of how it would look like with the other solution), but, again, introduces a lot of duplication. I would rather see things descending from each other. This is my opinion, but if I add a new service type, I would want to define it in the service module and not copy-paste the whole thing, delete the last part, and add the new service name.

tracing/targets/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+name        = "logos-blockchain-log-targets"
+readme      = { workspace = true }
+repository  = { workspace = true }
+version     = { workspace = true }
+
+[lints]
+workspace = true
+
+[dependencies]
+logos-blockchain-log-targets-macros = { path = "./macros" }
```
ntn-x2: I also mentioned it elsewhere: why not making this a workspace dependency?

tracing/src/filter/envfilter.rs
```diff
@@ -51,6 +51,45 @@ pub fn default_debug_log_filter(level: Level) -> HashMap<String, Level> {
     filters
 }
 
+/// Validates a configured log-filter target against the known Logos target
+/// catalog.
+///
+/// Targets outside the Logos catalog are currently accepted so that
+/// external targets and not-yet-catalogued internal targets continue to work.
+pub fn validate_log_filter_target(target: &str) -> Result<(), String> {
+    if target == "*" {
```
ntn-x2: `"*"` is also repeated in a few places here. Might be best to make it a constant instead.

tracing/src/filter/envfilter.rs
```diff
@@ -51,6 +51,45 @@ pub fn default_debug_log_filter(level: Level) -> HashMap<String, Level> {
...
+        return Ok(());
+    }
+
+    if lb_log_targets::is_logos_target_root(target)
```
ntn-x2: Why not exposing one more function that combines these two, which would then not be public? Something like `is_valid_logos_target`?

tracing/src/filter/envfilter.rs
```diff
@@ -68,6 +107,29 @@ fn envfilter_directives(filters: &HashMap<String, Level>) -> String {
     directives.join(",")
 }
 
+fn parse_filter_directive(directive: &str) -> Result<(String, Level), String> {
+    if let Some((target, level)) = directive.split_once('=') {
+        let target = target.trim();
+        let level = level.trim();
+
+        if target.is_empty() || level.is_empty() {
```
ntn-x2, line 118: Should these checks be part of `validate_log_filter_target` and `parse_filter_level` respectively?

tracing/src/filter/envfilter.rs
```diff
@@ -68,6 +107,29 @@ fn envfilter_directives(filters: &HashMap<String, Level>) -> String {
...
+}
+
+fn parse_filter_level(level: &str) -> Result<Level, String> {
+    level
+        .trim()
+        .parse()
+        .map_err(|_| format!("Invalid log filter level provided: {level}"))
+}
+
 pub mod serde_filters {
```
ntn-x2: Is this module now only used in `serde_validated_filters`? Does it make sense for it to exist?

ntn-x2 (commented):
Looks a step in the right direction, and I'd like to see this used sooner rather than later with the larger and larger amount of logs we will have to deal with. Nevertheless, I am still not sold on the flat structure and would like to briefly discuss why this would be better than a nested approach. Thank you!

andrussal:
> Looks a step in the right direction, and I'd like to see this used sooner rather than later with the larger and larger amount of logs we will have to deal with. Nevertheless, I am still not sold on the flat structure and would like to briefly discuss why this would be better than a nested approach. Thank you!

Thanks for the review!

I kept the flat structure because it felt easier to read and review than the tree-style versions tried before. With the flat form, every declared target is visible directly, while the nested versions made it harder to scan the full set at once.

It also keeps the declaration side simpler when adding targets at different depths, since there is no extra nesting structure to maintain around them. The downside is some repetition in the shared prefixes, but in practice that did not seem too bad because related targets still sit together visually.

That said, this is not something I feel strongly about. If you think the nested form would be clearer long term, we can change.

andrussal:
> > Looks a step in the right direction, and I'd like to see this used sooner rather than later with the larger and larger amount of logs we will have to deal with. Nevertheless, I am still not sold on the flat structure and would like to briefly discuss why this would be better than a nested approach. Thank you!
> 
> Thanks for the review!
> 
> I kept the flat structure because it felt easier to read and review than the tree-style versions tried before. With the flat form, every declared target is visible directly, while the nested versions made it harder to scan the full set at once.
> 
> It also keeps the declaration side simpler when adding targets at different depths, since there is no extra nesting structure to maintain around them. The downside is some repetition in the shared prefixes, but in practice that did not seem too bad because related targets still sit together visually.
> 
> That said, this is not something I feel strongly about. If you think the nested form would be clearer long term, we can change.

Let me try also nested structure, so we can compare.

andrussal:
@ntn-x2 I switched it to a tree structure. This version is actually not bad. It is different from what I tried before, and it feels easier to follow and edit.

https://github.com/logos-blockchain/logos-blockchain/blob/34d2b9013678157c139a008e687a257b83edb95f/tracing/targets/src/blend.rs

tracing/targets/src/lib.rs
andrussal: No strong reason, I think I just got a bit carried away with the new crate. I agree it does not make sense as a one-off local rule here, removed it.

tracing/targets/macros/Cargo.toml
andrussal: > Since I'm working on a related PR: can we move these to workspace deps as well? So we keep everything in a single place.

Sounds good!

tracing/targets/src/lib.rs
andrussal: > This assumes namespaces are divided by `::`. Is this documented anywhere?

I think it will be easy this to see from existing use cases, not sure if people will find docs more easily.

tracing/src/filter/envfilter.rs
andrussal: I would keep them here because this is where the whole directive is parsed, so it can report an error against the full directive. Moving the empty checks into the helpers would lose some of that context.

tracing/targets/src/lib.rs
ntn-x2: A function doc so that it pops up in IDEs would be sufficient I think.

services/blend/Cargo.toml
```diff
@@ -25,7 +25,7 @@ lb-cryptarchia-engine            = { workspace = true }
 lb-groth16                       = { default-features = false, workspace = true }
 lb-key-management-system-service = { features = ["unsafe"], workspace = true }
 lb-ledger                        = { workspace = true }
-lb-libp2p                        = { optional = true, workspace = true }
+lb-libp2p                        = { workspace = true }
```
ntn-x2, line 28: Was this necessary? Or is it a typo?

tracing/src/filter/envfilter.rs
```diff
@@ -57,17 +58,19 @@ pub fn default_debug_log_filter(level: Level) -> HashMap<String, Level> {
...
-    if target == "*" {
+    if target == ENVFILTER_GLOBAL_TARGET {
         return Ok(());
     }
 
-    if lb_log_targets::is_logos_target_root(target)
-        && !lb_log_targets::is_valid_logos_target_prefix(target)
-    {
-        return Err(format!("unknown log filter target `{target}`"));
+    if !lb_log_targets::is_logos_target_root(target) {
```
ntn-x2, line 65: So now we return `Ok` if the target is not a valid logs root target?

ntn-x2 (approved):
I like the new structure much more! Thank you!

tracing/src/filter/envfilter.rs
andrussal: Here we also consider it valid if it is not part of logos, it passes without any validation.

services/blend/Cargo.toml
andrussal: > Was this necessary? Or is it a typo?

Just a typo when adding it back after accidental removal.

tracing/targets/src/blend.rs
```diff
@@ -1,25 +1,33 @@
...
-    blend::scheduling::proofs::CORE,
-    blend::scheduling::proofs::CORE_AND_LEADER,
-    blend::scheduling::proofs::LEADER,
-    blend::service::CORE,
-    blend::service::EDGE,
-    blend::service::EPOCH,
-    blend::service::MODES,
-    blend::service::core::KMS_POQ_GENERATOR,
-    blend::service::edge::backend::LIBP2P,
+    blend::{
```
ntn-x2: Just a last question: since this is in the `blend` module, can we get rid of the `blend` prefix? Or do we need a common prefix for the macro to work? In that case, can we use the same name for all modules, like `root`?

andrussal: > Just a last question: since this is in the `blend` module, can we get rid of the `blend` prefix? Or do we need a common prefix for the macro to work? In that case, can we use the same name for all modules, like `root`?

Good point, what about https://github.com/logos-blockchain/logos-blockchain/commit/77af04dbf74e80027d17390caa2b5b2521b6fb80 ?

ntn-x2: But this is the same, no? We are still defining `root = blend` within the `blend` module.

andrussal: Now it is not part of the tree and reduces nesting, but yes, we still need it. I don't think Rust can find out from the module name(?)

ntn-x2: I mean the tracing crate uses the path of the component that is logging, so I think it should be possible.

andrussal: Thanks for pushing, it was yes doable - https://github.com/logos-blockchain/logos-blockchain/compare/320371dc34a3...802bf5108e08

I also moved `log_targets` to top and added more docs to it.

PR #2656, opened 2026-04-30
chore(genesis): Skip genesis SDP Op proof validation

strinnityk:
- Add a workaround to skip genesis SDP Op proof validation. The implemented if-based method is temporary until the verified/unverified/genesis TX is done.
- Add lifecycle test to debug a panic that's currently happening on `start_lb_node`.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.


core/src/mantle/ops/sdp/declare.rs
```diff
@@ -98,22 +99,25 @@ impl Operation for SDPDeclareOp {
...
+            ctx.locked_notes = ctx
+                .locked_notes
+                .lock(
+                    &ctx.min_stake,
+                    self.service_type,
+                    utxo.note,
+                    &self.locked_note_id,
+                )
+                .map_err(|_| SdpError::UnexpectedError)?;
+        }
```
youngjoon-lee: I think we still have to lock the note even if the operation is in the genesis tx. Does the test fail if we enable this code block?

strinnityk: I don't think the note exists, though, does it? Otherwise the validation wouldn't fail in the first place.
The `lock` method takes two parameters: `utxo.note` and `self.locked_note_id` that don't exist, it's the same reason the above operation fails.

core/src/mantle/ops/sdp/declare.rs
```diff
@@ -25,6 +25,7 @@ pub struct SDPDeclareExecutionContext {
     pub declarations: Declarations,
     pub locked_notes: LockedNotes,
     pub min_stake: MinStake,
+    pub is_genesis: bool,
```
danielSanchezQ: Uhm, this is not that bad tbh, glad that we got the context now.

nodes/node/binary/src/config/deployment/devnet.rs
```diff
@@ -87,7 +87,12 @@ cryptarchia:
       ops_proofs:
       - !Ed25519Sig '00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000'
       - !Ed25519Sig '00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000'
-      - !Ed25519Sig '00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000'
+      - !ZkAndEd25519Sigs
+        zk_sig:
+          pi_a: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
+          pi_b: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
+          pi_c: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```
danielSanchezQ, line 92: We should move this to use hex in human readable serializers

danielSanchezQ (approved):
Look ok to me for now. How was this working before with the genesis tx? cc @youngjoon-lee

core/src/mantle/ops/sdp/declare.rs
strinnityk: It's definitely not horrible, but I think we can make it cleaner after the TX refactor. Probably even remove the conditional.

nodes/node/binary/src/config/deployment/devnet.rs
strinnityk: Definitely. I have this written down somewhere since it'd also help unify bytes format for APIs (e.g. REST).

core/src/mantle/ops/sdp/declare.rs
youngjoon-lee: Hm, maybe I misunderstand the logic or this PR. I thought that the genesis tx must contain output notes and SDP declarations that lock some of those output notes.

youngjoon-lee: I just checked the cfgsync generates real output notes and uses them for SDP declarations. Also it generates real proofs. So, I think we don't need we skip proof validation for the genesis.

bacv (approved):
This looks great, ~~could you also update the `run_block` command in the `logos-blockchain-tools` to generate empty ZkAndEd25519 proof instead of ed255?~~

core/src/mantle/ops/sdp/declare.rs
bacv: Missed this, we need to lock the note.

bacv (changes requested):

core/src/mantle/ops/sdp/declare.rs
bacv: Added https://github.com/logos-blockchain/logos-blockchain/pull/2731 on top of this PR, it resolves the node locking and skips proofs for genesis block, once the provider sig question is resolved and it's approved, will merge both.

bacv (approved):

youngjoon-lee (approved):
👍

PR #2666, opened 2026-05-04
refactor(sdp): remove dead code: `SessionState::update`

youngjoon-lee:

We had a dead code.

`ServiceState` has two `SessionState`s:
- `active`: Declarations that are active now in the current session.
- `forming` Declarations that will become active from the next session.

We were calling `forming.update()` every new block processing, but it does its job **only** when the new block is **older** than its parent. It is impossible because we always build a new ledger state by cloning its parent ledger state.

So, in this PR, I removed the `update` function entirely, because we actually don't need to update the forming session state. It is actually a snapshot taken at the start block of the current session.
The name `forming` is misleading, ~~but this PR doesn't change it now~~. So, I renamed it to `next` because it's a snapshot that will be used from the next session.

I didn't add any new test for this change because it is quite clear that this is a dead code. I checked that the existing tests are sufficient.

I believe so.

@youngjoon-lee 
spec: @madxor 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/ledger/src/mantle/sdp/mod.rs
+++ b/ledger/src/mantle/sdp/mod.rs
@@ -66,9 +66,9 @@ impl Service {
     }
 
     #[cfg(test)]
-    const fn forming_session(&self) -> &SessionState {
+    const fn next_session(&self) -> &SessionState {
         match self {
-            Self::BlendNetwork(state) => &state.forming,
+            Self::BlendNetwork(state) => &state.next,
         }
     }
 
@@ -144,8 +144,8 @@ pub enum Error {
     DuplicateDeclaration(DeclarationId),
     #[error("Active session for service {0:?} not found")]
     ActiveSessionNotFound(ServiceType),
-    #[error("Forming session for service {0:?} not found")]
-    FormingSessionNotFound(ServiceType),
+    #[error("Next session for service {0:?} not found")]
+    NextSessionNotFound(ServiceType),
     #[error("Session parameters for {0:?} not found")]
     SessionParamsNotFound(ServiceType),
     #[error("Service parameters are missing for {0:?}")]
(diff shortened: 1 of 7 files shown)


pradovic (approved):
This looks to me like it was meant to "form" the session until the session boundary is crossed and the session is formed. But the condition does not look good in that case.

ledger/src/mantle/sdp/mod.rs
```diff
@@ -255,11 +238,11 @@ impl<R: Rewards> ServiceState<R> {
...
         } else {
             assert!(
-                current_session < self.active.session_n + 1,
-                "Logos blockchain isn't ready for time travel yet"
+                current_session == self.active.session_n,
+                "Logos blockchain isn't ready for time travel yet: session_of_block={current_session}, active_session={}",
+                self.active.session_n
             );
             self.rewards = self.rewards.update_epoch(epoch_state);
-            self.forming = self.forming.update(&self, block_number, config);
```
ntn-x2: If we take it from the snapshot of the right block (and I agree that should be the case) we should completely remove `forming` as part of this PR then. It has no reason to exist if we don't update it here anymore.

youngjoon-lee: Yeah, the `forming` is never mutated. The name is misleading as I mentioned in the PR description.

a507c7d9e9d86ce4dde0610a7a51cc6c5c24413a I renamed it to `next`. We still need to keep it because it is snapshotted when the current session starts.

@pradovic Could you pls do a review once again? I think I understood the logic correctly, but I want to get confirmation from someone who knows better about SDP.

pradovic: The changset looks good to me! Having said that, I must say I was working on membership in DA and sdp service, but this part was done by Giacomo and @bacv also migrated membership to ledger. So @bacv might be a good person to take a look as well.

bacv (approved):

PR #2667, opened 2026-05-04
chore: enforce dependency declaration structure

ntn-x2:

Introduce a script that checks the following conditions:

* Crates must use only dependencies defined in the workspace root
* Crates cannot override the `default-features` property of the imported workspace dependencies
* Workspace root cannot activate `features` itself on dependencies
* Workspace root must use `default-features = false` on dependencies

I added a AI-generated script, which works on my local machine, to check those cases and that we would run in CI to make sure we all follow the same process, to keep dependencies in order and avoid using multiple version of the same dependency.

We cannot yet disable the `multiple_crate_versions` lint our dependencies still use different versions of sub-dependencies, but we will tackle this at some point, definitely not a high priority task. But this PR better defines the process of introducing new dependencies into the workspace. 

Yes. The script is to be treated as a black box, I'd say.

@ntn-x2 

N/A

No.

### DIFF
--- a/scripts/check-workspace-dependency-policies.sh
+++ b/scripts/check-workspace-dependency-policies.sh
@@ -1,0 +1,329 @@
+#!/usr/bin/env sh
+set -eu
+
+# Enforces workspace dependency policy across Cargo manifests.
+#
+# Checks:
+# - Workspace members must declare dependencies using `workspace = true` only
+#   (no direct version/path/git dependency declarations).
+# - Workspace members must not override `default-features` on workspace deps.
+# - Root `[workspace.dependencies]` entries must set `default-features = false`.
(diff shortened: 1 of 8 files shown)


scripts/check-workspace-dependency-policies.sh
danielSanchezQ, line 1: Pray claude, we trust.

danielSanchezQ (approved):
Pre-approving this. Nice that we can enforce it, I was a bit bugged about it to be honest.

scripts/check-workspace-dependency-policies.sh
ntn-x2: I tested it and refined it quite a bit locally :D It's a good boy after all ❤️

PR #2674, opened 2026-05-04
feat(block-tracker): introduce block-tracker to be re-used to track chain state where applicable

pradovic:

Adds a new `block-tracker` crate that provides a generic per-branch state tracker (`BlockView` strategy + `BlockTree<Id, V>`), and
  refactors zone-sdk's `state.rs` to use it.

**Not for merge yet** — please review the trait shape and flag anything that wouldn't fit the use case. I have tried to integrate it to zone-sdk right away to show the usage. Let's see if it makes sense for other consumers as well.

 ## The abstraction

  ```rust
  trait BlockView {
      type Input;          // generic — Block, (Block, Payouts), or anything
      type State: Clone;
      fn apply(&self, parent: &Self::State, input: &Self::Input) -> Self::State;
      fn default_state() -> Self::State;
  }

  struct BlockTree<Id, V: BlockView> { ... }
  // process, state_at, find_lcm, inputs_between, advance_lib, iter_states_mut
  
  ```
`BlockView` is closer to a strategy than a pure FP lens. — apply is functional, but the tree exposes `iter_states_mut` for cross-branch maintenance that doesn't fit a per-block fold (e.g. shrinking states when finalized externally).

@danielSanchezQ, @youngjoon-lee  — I also took a look at the wallet side to sanity check the fit.

My understanding is that wallet currently handles reorgs implicitly — it relies on chain-service to feed the canonical view and signal pruned blocks, rather than computing branch diffs itself.

Adopting BlockTree wouldn’t change that model, but it would expose primitives like find_lcm / inputs_between in case wallet ever wants to reason about reorgs locally. And the block aware state is shared.

For rewards specifically, it seems the current approach could be extended without BlockTree (e.g. passing additional data alongside blocks). The generic Input shape might still be useful for structuring that (e.g. Input = (Block, Payouts)), but it doesn’t look strictly required.

So from my side it looks like:

* zone-sdk: clear fit 
* wallet: optional — current model works, this would mainly add structure / optional capabilities
* mempool: likely similar to wallet (chain-service-driven), but unclear if the abstraction adds value there without more reorg-aware logic

**Bottom line**: the shared piece is the multi-branch state-map pattern. The reorg-detection primitives (find_lcm, inputs_between) are external-consumer-only — they exist because the public API doesn't expose what chain-service gives internally. If a future event-stream-style public API ever ships, those primitives become vestigial; until then, zone-sdk needs them.

Curious to hear your thoughts.

Yes

@pradovic @danielSanchezQ @youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/block-tracker/src/lib.rs
+++ b/block-tracker/src/lib.rs
@@ -1,0 +1,46 @@
+//! Generic per-branch state tracker over a block tree.
+//!
+//! A [`BlockView`] is a strategy that describes how to derive per-block
+//! state from an input. A [`BlockTree`] maintains those states across the
+//! chain's branching structure, supports reorgs via lowest-common-ancestor
+//! lookup, and prunes finalized history as LIB advances.
+//!
+//! The `Input` type is generic so consumers can feed not just blocks but
+//! richer per-position data (e.g. `(Block, Payouts)`) — useful when the
+//! domain-of-interest depends on data that the protocol computes alongside
+//! a block but that doesn't appear in transactions.
+//!
+//! By convention, the driver is responsible for summarizing raw chain data
+//! into the lens's `Input` type before calling [`BlockTree::process`]. The
+//! lens's `apply` method is itself a pure derivation `(parent, input) ->
+//! child`, but the tree exposes mutation primitives (e.g.
+//! [`BlockTree::iter_states_mut`]) for cross-branch maintenance that
+//! doesn't fit a per-block fold.
+
+mod tree;
+
+pub use tree::{BlockTree, PrunedBlock};
+
+/// View into a chain: how to fold per-block input into accumulated branch
+/// state.
+///
+/// Implementations are pure functions of their arguments. The driver feeds
+/// `Input` for each block; the tracker takes care of branching and reorgs.
+pub trait BlockView {
+    /// Per-position input. May be a block summary, or a tuple bundling a
+    /// block with additional protocol-derived data the lens needs.
+    type Input;
+
+    /// Accumulated state along a branch (e.g. UTXO set, safe-tx set).
+    ///
+    /// Required to be `Clone` because the tree stores one state per block;
+    /// implementations should use a persistent data structure (e.g. `rpds`)
+    /// to keep cloning cheap across deep histories.
+    type State: Clone;
+
+    /// Derive child state from parent state plus this block's input.
+    fn apply(&self, parent: &Self::State, input: &Self::Input) -> Self::State;
+
+    /// Initial state at genesis (before any block has been applied).
+    fn default_state() -> Self::State;
+}
--- a/block-tracker/src/tree.rs
+++ b/block-tracker/src/tree.rs
@@ -1,0 +1,406 @@
+use std::{
+    collections::{HashMap, HashSet},
+    hash::Hash,
+};
+
+use crate::BlockView;
+
+/// A block reported by [`BlockTree::advance_lib`] as newly finalized.
+///
+/// The `input` is moved out of the tree at finalization time — once a
+/// block is below LIB, the tree no longer needs its per-block data.
+/// Consumers that want to materialize "finalized" effects (e.g. mark
+/// rewards as spendable) read these.
+pub struct PrunedBlock<Id, Input> {
+    pub id: Id,
+    pub input: Input,
+}
+
+/// Per-branch state tracker over a block tree.
+///
+/// Stores per-block accumulated state and per-block input along every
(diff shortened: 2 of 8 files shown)


block-tracker/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+mod tree;
+
+pub use tree::{BlockTree, PrunedBlock};
+
+/// View into a chain: how to fold per-block input into accumulated branch
+/// state.
+///
+/// Implementations are pure functions of their arguments. The driver feeds
+/// `Input` for each block; the tracker takes care of branching and reorgs.
+pub trait BlockView {
```
danielSanchezQ, line 29: We could call it View to be generic, as the input may not be a block.

block-tracker/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+    /// Required to be `Clone` because the tree stores one state per block;
+    /// implementations should use a persistent data structure (e.g. `rpds`)
+    /// to keep cloning cheap across deep histories.
+    type State: Clone;
+
+    /// Derive child state from parent state plus this block's input.
+    fn apply(&self, parent: &Self::State, input: &Self::Input) -> Self::State;
+
+    /// Initial state at genesis (before any block has been applied).
+    fn default_state() -> Self::State;
```
danielSanchezQ, line 45: This may not be needed, its better to allow any other constructions as some may require a default other loading from settings, etc.

block-tracker/src/tree.rs
```diff
@@ -0,0 +1,406 @@
...
+
+/// Per-branch state tracker over a block tree.
+///
+/// Stores per-block accumulated state and per-block input along every
+/// branch. Reorgs are handled by switching to a different tip;
+/// [`Self::find_lcm`] gives the common ancestor for diffing.
+///
+/// As LIB advances, the chain segment below the new LIB is finalized and
+/// returned to the caller; sibling branches of that segment are pruned.
+pub struct BlockTree<Id, V: BlockView> {
```
danielSanchezQ, line 27: ok, so we keep a state per id so in case of re-org its just a matter of taking back all states. There should be an automatic way of promoting those states to inputs or views agains so for example:
* In the case of sdk the transactions can be resubmitted in order.
* Wallet recover the utxos+transactions and can resubmit them
* Mempool make pending txs ready to be selected for the leader.

block-tracker/src/tree.rs
```diff
@@ -0,0 +1,406 @@
...
+    /// Mutable iterator over all branch states in the tree.
+    ///
+    /// Useful for tree-wide mutations that aren't expressible as a per-block
+    /// fold — e.g. shrinking states in response to information learned
+    /// outside the lens (such as which entries became finalized in an
+    /// external system). The lens itself remains a pure function; this
+    /// escape hatch lets the driver perform cross-branch maintenance.
+    pub fn iter_states_mut(&mut self) -> impl Iterator<Item = &mut V::State> + '_ {
+        self.states.values_mut()
+    }
```
danielSanchezQ, line 97: I think this method is just AI slop, its not useful imo. It doesnt keep the order of insertion or parenting, so worthless.

block-tracker/src/tree.rs
```diff
@@ -0,0 +1,406 @@
...
+        loop {
+            if a_chain.contains(&cur) {
+                return Some(cur);
+            }
+            match self.parent_map.get(&cur) {
+                Some(p) => cur = p.clone(),
+                None => return None,
+            }
+        }
+    }
```
danielSanchezQ, line 127: Not sure this is the proper implementation (doesn't look so). Also I don't think would be used neither.

block-tracker/src/tree.rs
```diff
@@ -0,0 +1,406 @@
...
+                for k in kids {
+                    if keep.insert((*k).clone()) {
+                        frontier.push((*k).clone());
+                    }
+                }
+            }
+        }
+        keep
+    }
+}
```
danielSanchezQ, line 236: Not sure we need the pathing neither.

danielSanchezQ (commented):
Looks like a good start. I'm missing a few stuff. To be discussed.

danielSanchezQ (commented):
Looks like a good start. I'm missing a few stuff. To be discussed.

pradovic:
We are closing this PR for now until we have a more full picture from the different consumers (wallet, mempool, zone-sdk), and then we will revisit.

https://github.com/logos-blockchain/logos-blockchain/issues/2687.

PR #2675, opened 2026-05-04
fix(logging): reduce noisy node logs

### LINKED ISSUE #2662: gossipsub: duplicate mempool broadcast generates ~120k ERROR/WARN log lines per hour
## Summary

The node emits two paired log lines in a tight loop whenever it tries to re-broadcast a mempool message that gossipsub has already seen:

```
WARN  libp2p_gossipsub::behaviour: Not publishing a message that has already been published <hash>
ERROR logos_blockchain_network_service::backends::libp2p::swarm::gossipsub: failed to broadcast message to topic: /logos-blockchain-testnet-v0.1.2/mempool/1.0.0 Duplicate
```

## Rate observed

Measured on a single testnet node (`live-v0.1.2`, `tracing.level: DEBUG`):

| Line | Rate |
|------|------|
| `WARN … Not publishing … already been published` | ~57,000 / hr |
| `ERROR … failed to broadcast … Duplicate` | ~63,000 / hr |
| **Total** | **~120,000 / hr** |

These are ERROR and WARN level — they appear regardless of tracing level and are the dominant driver of disk usage (~3 GB/hr even alongside DEBUG logging).

## Impact

- Logs fill a 295 GB partition in ~3 days at this rate
- Switching `tracing.level` from `DEBUG` to `INFO` does **not** help — these lines are emitted above DEBUG threshold
- The only current mitigation is aggressive log rotation (24h retention)

## Expected behaviour

Gossipsub already deduplicates at the protocol level. The node layer should either:
1. Not attempt to re-broadcast a message gossipsub has already published, or
2. Suppress / rate-limit the resulting WARN/ERROR to avoid per-message log spam

## Related

#2332 touches the mempool broadcast code path (dead broadcast channel in `tx-service`) — may be connected.

## Environment

- Node: `logos-blockchain-node` (testnet `live-v0.1.2`)
- Topic: `/logos-blockchain-testnet-v0.1.2/mempool/1.0.0`
- Config: `tracing.level: DEBUG` (reproduces at INFO too)

---

*Drafted with AI assistance (Claude Code)*

andrussal:

This PR adjusts several noisy logging paths to keep routine node activity out of higher log levels.

This mainly quiets repeated block-processing logs, routine LIB updates, NTP update ticks, no-op mempool removals, and duplicate gossipsub publish noise. 

The gossipsub part is specifically for the duplicate publish loop reported in #2662

Closes #2662

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/tracing/src/filter/envfilter.rs
+++ b/tracing/src/filter/envfilter.rs
@@ -13,6 +13,7 @@ const DEFAULT_DEBUG_TARGETS: &[&str] = &[
     "cryptarchia",
     "ledger",
 ];
+const DEFAULT_QUIET_TARGETS: &[(&str, Level)] = &[("libp2p_gossipsub", Level::ERROR)];
 const ENVFILTER_GLOBAL_TARGET: &str = "*";
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
@@ -48,6 +49,11 @@ pub fn default_debug_log_filter(level: Level) -> HashMap<String, Level> {
         DEFAULT_DEBUG_TARGETS
             .iter()
             .map(|target| ((*target).to_owned(), level)),
+    );
+    filters.extend(
+        DEFAULT_QUIET_TARGETS
+            .iter()
+            .map(|(target, level)| ((*target).to_owned(), *level)),
     );
     filters
 }
@@ -183,8 +189,8 @@ mod tests {
     use tracing::Level;
 
     use super::{
-        ENVFILTER_GLOBAL_TARGET, EnvFilterConfig, create_envfilter_layer, parse_filter_directives,
-        validate_log_filter_target,
+        ENVFILTER_GLOBAL_TARGET, EnvFilterConfig, create_envfilter_layer, default_debug_log_filter,
+        parse_filter_directives, validate_log_filter_target,
     };
 
     #[test]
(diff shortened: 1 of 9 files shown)


tracing/src/filter/envfilter.rs
```diff
@@ -49,6 +50,11 @@ pub fn default_debug_log_filter(level: Level) -> HashMap<String, Level> {
             .iter()
             .map(|target| ((*target).to_owned(), level)),
     );
+    filters.extend(
+        DEFAULT_QUIET_TARGETS
+            .iter()
+            .map(|(target, level)| ((*target).to_owned(), *level)),
```
danielSanchezQ, line 56: Can we activate this in runtime somehow?

danielSanchezQ (approved):

tracing/src/filter/envfilter.rs
bacv: Very good suggestion, would allow us to debug remote users nodes without restart, should be possible with https://docs.rs/tracing-subscriber/0.3.23/tracing_subscriber/reload/index.html

andrussal: Just to clarify, I think there are two separate things here.

For this PR, I can avoid keeping this only as a hidden default and make `init` write the default filter into the generated config instead. Then the generated config would contain something like:

```yaml
    tracing:
      filter:
        Env:
          filters:
            "*": WARN
            logos_blockchain: DEBUG
            libp2p_gossipsub::behaviour: ERROR
```

Runtime reload with `tracing_subscriber::reload` would be useful for debugging remote nodes without restart, but I think that is a separate change and probably better as a follow-up PR, wdyt?

bacv: > but I think that is a separate change and probably better as a follow-up PR, wdyt

Definitely a separate PR, maybe not even immediate one if you have other in progress work, this PR is ready to go.

PR #2676, opened 2026-05-04
fix(sdp): do not store Blend reward parameters in recovery state

### LINKED ISSUE #2673: Do not store Blend reward parameters in state in chain service
Stuff that is passed in settings should not really be stored in state, else we would have a mismatch if we restart the node with a different set of values (e.g., `minimum_blend_size`). At least we should throw a big warning, potentially a panic, but even better if we don't store them at all.

youngjoon-lee:

Closes #2673 

I removed `RewardsParameters` from the SDP ledger state, so that it is not stored in the recovery state.
It is part of the deployment config. It should be loaded from the deployment config whenever the node is (re-)started. It shouldn't be loaded from the recovery state.
For example, if users agreed to change some values in the deployment config, rebuild the binary, and restart the node, the reward module should run with the new deployment config, not with something recovered from the state. 

One important thing is that, the SDP Ledger has 2 modes for blend rewards.
- `WithoutTargetSession`: Reject all activity proofs because network is small or because we're still in session 0 or 1.
- `WithTargetSession`: Accept activity proofs because network is large enough.

SDP Ledger stores this mode and related data into the recovery state. When the node is restarted, it loads the previous mode from the state, regardless of the new deployment config. 
Therefore, the new deployment config becomes effective from the next session, when it comes to Blend rewards.
Also, 
- If activity proofs are generated/received due to the decreased minimum network size in the new deployment config, and if the loaded mode is `WithoutTargetSession`, the SDP ledger will return `Error::TargetSessionNotSet` to reject activity proofs, resulting that the block that contains the activity message is rejected. 
- If activity proofs are no longer generated due to the increased minimum network size in the new deployment config, and if the loaded mode is `WithTargetSession`, the SDP ledger won't see any new activity proofs until the next session transition.
These behaviors are correct, and have no issue in terms of consensus.

This PR doesn't have any new tests because the existing tests cover all cases.

yes

@youngjoon-lee 
spec: @madxor 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/ledger/src/mantle/mod.rs
+++ b/ledger/src/mantle/mod.rs
@@ -66,10 +66,8 @@ impl LedgerState {
     pub fn new(config: &Config, epoch_state: &EpochState) -> Self {
         Self {
             channels: channel::Channels::new(),
-            sdp: sdp::SdpLedger::new().with_blend_service(
-                config.sdp_config.service_rewards_params.blend.clone(),
-                epoch_state,
-            ),
+            sdp: sdp::SdpLedger::new()
+                .with_blend_service(&config.sdp_config.service_rewards_params.blend, epoch_state),
             leaders: leader::LeaderState::new(),
         }
     }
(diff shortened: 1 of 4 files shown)


ntn-x2 (approved):
Nice, thank you!

danielSanchezQ (approved):

PR #2678, opened 2026-05-04
fix(cryptarchia): add panics to `Branches::walk_back_*` to avoid infinite loop

youngjoon-lee:

`Branches::walk_back_before` and `Branches::walk_back_to_block` never end if they meet the genesis block before reaching the target. It's better to panic in that case.

I chose to panic instead of returning an error because all functions of `Branches` are already panicking if they cannot finish their job with the provided arguments. I think it's reasonable because `Branches` is a sub-component of `Cryptarchia` and is not used anywhere else. 

yes

@youngjoon-lee 
spec: @davidrusu 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/consensus/cryptarchia-engine/src/lib.rs
+++ b/consensus/cryptarchia-engine/src/lib.rs
@@ -268,6 +268,10 @@ where
     fn walk_back_before(&self, branch: &Branch<Id>, slot: Slot) -> Branch<Id> {
         let mut current = branch;
         while current.slot > slot {
+            assert!(
+                current.id != current.parent,
+                "shouldn't meet genesis block before reaching target slot: {slot:?}"
+            );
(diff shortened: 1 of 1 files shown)


danielSanchezQ (approved):
What are the cases of the infinite loop? Is there a way to avoid it logically without panicing?

youngjoon-lee:
> What are the cases of the infinite loop? Is there a way to avoid it logically without panicing?

I found that there are so many infinite-loop points in `Branches`. It shouldn't happen in the current code base, because the upper layer always uses `Branches` correctly, but I feel I can refactor `Branches` safer. I opened an issue https://github.com/logos-blockchain/logos-blockchain/issues/2702 and I'm closing this PR.

PR #2682, opened 2026-05-05
fix(chain): verify immutable block id

andrussal:

Fixes `load_immutable_block` so a block is only treated as immutable when the immutable index for that slot points to the same block ID.

Before this, any block present in storage could be misclassified as immutable if the immutable index had some block for the same slot. That later made path computation fail with `Couldn't reach start_block`.

Added a regression test for that case.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


danielSanchezQ (approved):
🚀

PR #2685, opened 2026-05-05
feat(chain-service): Await chain start time

bacv:

Added an `AwaitingStart` mode to the chian service to allow users observe this state in the api and to prevent remote blocks being processed while chain didn't start yet. Wrapped consensus info and cryptarchia state into `ChainServiceInfo`. 

`ChainServiceMode::AwaitingStart` is only used in the chain service `run` method to prevent applying blocks that are received too early, or switch chain to online after ibd is complete. Ibd in case of `AwaitingStart` is not relevant and it's response is ignored.

Yes

@bacv 

N/A

N/A

## Checklist

* [x] 2. Description added.


services/chain/chain-service/src/lib.rs
```diff
@@ -640,14 +686,18 @@ where
...
+                            ConsensusMsg::IbdCompleted if chain_start_timer.is_none() => {
                                 info!("Received IBD completion notification. Starting prolonged bootstrap timer.");
                                 // Start the prolonged bootstrap timer now that IBD is complete
                                 prolonged_bootstrap_timer = Some(Box::pin(tokio::time::sleep_until(
                                     Instant::now() + bootstrap_config.prolonged_bootstrap_period,
                                 )));
                             }
-                            ConsensusMsg::ApplyBlock { block, tx } => {
+                            // Blocks will be applied if chain start time didn't begin yet.
+                            ConsensusMsg::ApplyBlock { block, tx } if chain_start_timer.is_none() => {
```
danielSanchezQ: Nitpick: lets use `reply_channel` or something like that instead of `tx` its really confusing (even if it is kinda rust standard for channels).

services/chain/chain-service/src/lib.rs
```diff
@@ -681,6 +731,16 @@ where
                                     Self::reject_chain_sync_event(event).await;
                                 }
                             }
+                            ConsensusMsg::Info { tx } => {
```
danielSanchezQ: Likewise

danielSanchezQ (approved):

services/chain/chain-service/src/lib.rs
bacv: Renamed all channels named `tx` to `reply_channel` in chain service https://github.com/logos-blockchain/logos-blockchain/pull/2685/changes/ceba68b7fe9309183cffd3266c4fbc564120cfdc

services/chain/chain-service/src/lib.rs
```diff
@@ -601,6 +617,25 @@ where
...
+                .genesis_tx()
+                .cryptarchia_parameter()
+                .genesis_time;
+            let now = OffsetDateTime::now_utc();
+
+            if genesis_time > now {
+                let delay = (genesis_time - now).try_into().unwrap_or_default();
+                info!("Chain configured to start in the future: {genesis_time}");
+                chain_start_timer = Some(Box::pin(tokio::time::sleep(delay)));
+            }
```
hansieodendaal, line 637: I think the UX/logs can benefit from the logarithmic countdown implemented here: https://github.com/logos-blockchain/logos-blockchain/pull/2553/changes#diff-47cb1fc523693404c39c7c72bcb31465cce5921a0341d0b3b2d9fdd79aa17257R110

hansieodendaal (commented):
Looking good, two comments to consider please. 

The one is below, the other has to do with system-level testing. I did not see any added system-level testing in this PR? We should expand cucumber test `Scenario: Nodes with delayed genesis start joins network` to verify the wait mode. Most of the plumbing for that test has already been done.

bacv:
> The one is below, the other has to do with system-level testing

Added a testcase in https://github.com/logos-blockchain/logos-blockchain/pull/2685/commits/a77d721308091aa14e6b6fc80689ce98ec024c12

services/chain/chain-service/src/lib.rs
bacv: I don't think this countdown should be part of chain service, but as you've implemented in the linked change, done on demand by who ever needs such functionality by parsing the genesis time, I consider it out of scope for this PR.

hansieodendaal (commented):
Belated comments:
- LGTM, thanks

services/chain/chain-service/src/lib.rs
```diff
@@ -622,7 +658,18 @@ where
...
-                    () = async { prolonged_bootstrap_timer.as_mut().unwrap().as_mut().await }, if prolonged_bootstrap_timer.is_some() && cryptarchia.is_boostrapping() => {
+                    () = async { if let Some(timer) = chain_start_timer.as_mut() { timer.await; } }, if chain_start_timer.is_some() => {
+                        info!("Genesis time reached. Chain is now starting...");
+                        chain_start_timer = None;
+
+                        // Just like in the Ibd case, the bootstrap timer is started after the chain
+                        // start time began.
+                        prolonged_bootstrap_timer = Some(Box::pin(tokio::time::sleep_until(
+                            Instant::now() + bootstrap_config.prolonged_bootstrap_period,
+                        )));
```
youngjoon-lee, line 669: @bacv I think we need to remove this because PBP should start only after IBD is done. If we start PBP here right after the chain start time began, PBP would pass before IBD is completed if PBP is short.

youngjoon-lee: Um, I just understood why IBD is not relevant with chain start time. The node should be able to download historic blocks even before chain start time begins. But, we should delay starting PBP even if IBD is done, if chain start time hasn't begun yet, because it is meaningless to start PBP if the node won't process any new blocks. 

I'd propose adding two boolean variables: `is_ibd_done: bool` and `is_chain_started: bool`. We should enable PBP only after both variables become true. 

Or, I think it's time to introduce a typestate system.

I just opened an issue https://github.com/logos-blockchain/logos-blockchain/issues/2699. I can take this over after finishing other tasks. Or, please feel free to take it.

PR #2686, opened 2026-05-05
feat(tracing): Add compressed appender

bacv:

Add option to compress old log files. A simple CompressedRollingAppender was added which wraps `tracing_appender::rolling` appender.

Yes

@bacv

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/tracing/src/compressed_appender.rs
+++ b/tracing/src/compressed_appender.rs
@@ -1,0 +1,92 @@
+use std::{
+    fs,
+    io::{self, Write},
+    path::{Path, PathBuf},
+    time::{Duration, Instant},
+};
+
+use flate2::{Compression, write::GzEncoder};
+
+pub struct CompressedRollingAppender {
+    inner: tracing_appender::rolling::RollingFileAppender,
+    directory: PathBuf,
+    prefix: String,
+    last_compression: Instant,
+    compression_interval: Duration,
+}
+
+impl CompressedRollingAppender {
+    pub fn new(directory: PathBuf, prefix: &Path, compression_interval: Duration) -> Self {
+        let prefix_str = prefix.to_string_lossy().to_string();
+        let inner = tracing_appender::rolling::hourly(&directory, &prefix_str);
+
+        Self {
+            inner,
+            directory,
+            prefix: prefix_str,
+            last_compression: Instant::now() - compression_interval,
+            compression_interval,
+        }
+    }
+
+    fn try_spawn_compression(&mut self) {
+        if self.last_compression.elapsed() >= self.compression_interval {
+            self.last_compression = Instant::now();
+            self.spawn_compression_task();
+        }
+    }
+
+    fn spawn_compression_task(&self) {
+        let dir = self.directory.clone();
+        let prefix = self.prefix.clone();
+
+        std::thread::spawn(move || {
+            let Ok(read_dir) = fs::read_dir(&dir) else {
+                return;
+            };
+
+            for entry in read_dir.flatten() {
+                let path = entry.path();
+
+                if path.is_file()
+                    && path
+                        .file_name()
+                        .is_some_and(|n| n.to_string_lossy().starts_with(&prefix))
+                    && path.extension().and_then(|s| s.to_str()) != Some("log")
+                    && path.extension().and_then(|s| s.to_str()) != Some("gz")
+                {
+                    if let Err(e) = compress_file_gzip(&path) {
+                        eprintln!("failed to compress logs {}: {e}", path.display());
+                    } else {
+                        drop(fs::remove_file(path));
+                    }
+                }
+            }
+        });
+    }
+}
+
+impl Write for CompressedRollingAppender {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        let n = self.inner.write(buf)?;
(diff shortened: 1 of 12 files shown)


danielSanchezQ (approved):
Just a question is not clear to me. Does this append files to the same compressed one?

bacv:
> Does this append files to the same compressed one?

No, current rolling appender creates files every hour, the compress appender compresses each file and removes the uncompressed one.

hansieodendaal (commented):
Nice addition. I have two main comments:

1. The main risk here looks like file lifecycle/locking, especially on Windows where open-file deletion semantics are stricter. We spawn a background thread that scans matching files, compresses them, and then deletes the original, but I do not see clear evidence that the selected file is guaranteed to no longer be open by the rolling appender at that point. Please confirm how that is ensured — e.g. whether rollover drops the handle before this path can run. Otherwise this may race with rollover/current writes and behave differently across platforms.

1. As a separate enhancement, it may be useful to support an optional retention bound by hourly file count in addition to compression. For example, keeping only the most recent 24 hourly log files would give operators a simple way to cap disk usage while retaining about a day of logs.

bacv:
> Nice addition. I have two main comments:

Great points, 

> 1. The main risk here looks like file lifecycle/locking

Will update this PR to check that the compression interval is greater than the rolling appender. Unfortunatelly `tracing_appender::rolling` doesn't provide any way to hook into it's events, but I think this solution is enough for our needs. Proper deployments shouldn't rely on such functionality, but use tools like `logrotate`.

> 2. As a separate enhancement, it may be useful to support an optional retention bound

Great suggestion and `tracing_appender::rolling` already has this functionality, I think the next step would be to expand `AppenderType` to have configurable rolling appender with `roll_interval` and `retention_bound: RetentionBound{None, Bounded(Duration)}`, and if compression is toggled on, the rolling appender is passed to the compression appender as an `inner_writer`.

Will rerequest review when first point is fixed.

tracing/src/compressed_appender.rs
```diff
@@ -0,0 +1,92 @@
...
+
+        std::thread::spawn(move || {
+            let Ok(read_dir) = fs::read_dir(&dir) else {
+                return;
+            };
+
+            for entry in read_dir.flatten() {
+                let path = entry.path();
+
+                if path.is_file()
```
andrussal: I wonder if this filter should explicitly exclude the current hourly file?

tracing/src/compressed_appender.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    fn flush(&mut self) -> io::Result<()> {
+        self.inner.flush()?;
+        Ok(())
+    }
+}
+
+fn compress_file_gzip(path: &Path) -> io::Result<()> {
+    let input = fs::File::open(path)?;
+    let output_path = path.with_extension("gz");
```
andrussal: Maybe I am mistaken, but doesn't  `path.with_extension("gz")` collapse all rotated files to the same compressed output name? For example, if `path` is `prefix.log.2026-05-06-13` or `prefix.log.2026-05-06-14`, both become `prefix.log.gz`.`

danielSanchezQ:
> > Does this append files to the same compressed one?
> 
> No, current rolling appender creates files every hour, the compress appender compresses each file and removes the uncompressed one.



> > Does this append files to the same compressed one?
> 
> No, current rolling appender creates files every hour, the compress appender compresses each file and removes the uncompressed one.

I think it would be better to compress into a single file where all of them are stored. wdyt?

bacv:
> I think it would be better to compress into a single file where all of them are stored. wdyt?

How would I prevent corruption when downloading such file, while it's being updated? I think it's better to compress chunks as separate file. Also it's easier to remove old compressed logs when they are in different files.

tracing/src/compressed_appender.rs
bacv: > I wonder if this filter should explicitly exclude the current hourly file?

Updated the code to use `tracing_subscriber::rolling` symlink feature, to always have a link to the latest log file. This link is used by compressor to filter out currently used logfile.

tracing/src/compressed_appender.rs
bacv: Fixed, now it looks like this:
```
 $ ls -la target/debug/1777917934.*
 
 target/debug/1777917934.2026-05-06-09-32.gz
 target/debug/1777917934.2026-05-06-09-33.gz
 target/debug/1777917934.2026-05-06-09-34.gz
 target/debug/1777917934.2026-05-06-09-35.gz
 target/debug/1777917934.2026-05-06-09-36.gz
 target/debug/1777917934.2026-05-06-09-37.gz
 target/debug/1777917934.2026-05-06-09-38.gz
 target/debug/1777917934.2026-05-06-09-39
 target/debug/1777917934.latest -> ./1777917934.2026-05-06-09-39
```

andrussal (approved):

hansieodendaal (commented):
My belated comments:
- I think this looks good.
- We could make the _all file handles have been dropped_ argument a bit stronger by also skipping files with recent modification times, e.g. `skip if mtime >= now - rollover_interval - margin`. Not a proof, but a safer heuristic alongside the `.latest` check, especially for cross-platform concerns.

PR #2695, opened 2026-05-06
feat: add HTTP CLI wrapper with a first command to join Blend as a core node

ntn-x2:

Based on top of the locator validation PR as the new binary uses the locator validation function: https://github.com/logos-blockchain/logos-blockchain/pull/2712.

Add a new `api` utility binary that will introduce QoL improvements without having to touch the node binary. In this PR, we introduce a subcommand to post a Blend SDP declaration starting from the user config file and a bunch of other info.

We can start bundling this binary in our releases, and it would be the only other binary we would probably release for now.

I tested the feature with an e2e test that uses cucumber and the TF.

The actual code, yes. The testing code might be a bit messy, but I am open to feedback on how to improve it.

@ntn-x2 

N/A

No.

### DIFF
--- a/.github/workflows/end-to-end-with-cucumber.yml
+++ b/.github/workflows/end-to-end-with-cucumber.yml
@@ -187,6 +187,16 @@ jobs:
           artifact_subdir: Transactions
           verbose_console: "true"
 
+      # Feature: CLI utility binary (@cli_ci)
+      - name: Run cucumber suite - CLI utilities
+        if: ${{ !cancelled() && steps.verify_local_ntp.conclusion == 'success' }}
+        timeout-minutes: 60
+        uses: ./.github/actions/run-cucumber-suite
+        with:
+          suite_tag: "@cli_ci"
+          suite_name: CLI
+          artifact_subdir: Cli
+
       # Always upload JUnit test results for reporting, even if tests fail
       - name: Upload cucumber JUnit results
         if: ${{ !cancelled() && steps.verify_local_ntp.conclusion == 'success' }}
--- a/tests/cucumber_tests/features/api-cli.feature
+++ b/tests/cucumber_tests/features/api-cli.feature
@@ -12,6 +12,6 @@ Feature: API CLI
       | NODE_1    | 1             | WALLET_1A   |              |
     When all nodes have at least 2 blocks and converged to within 1 blocks in 180 seconds
     And I send 1 transactions of 1000 LGO each from wallet "WALLET_1A" to blend core zk key of node "NODE_1"
-    Then I run blend core SDP declaration CLI for node "NODE_1"
+    Then I declare node "NODE_1" as blend core node via the CLI binary
     And blend core SDP declaration for node "NODE_1" is included on node "NODE_1"
     And I stop all nodes
--- a/tests/cucumber_tests/features/api-cli.feature
+++ b/tests/cucumber_tests/features/api-cli.feature
@@ -1,17 +1,17 @@
 Feature: API CLI
 
   @api_cli_ci
-  Scenario: Join Blend core via CLI declaration
+  Scenario: Join Blend via CLI declaration
     Given the genesis block has the following wallet resources:
       | account_index | token_count | token_amount |
       | 1             | 1           | 2000         |
-    And I have a cluster with capacity of 2 nodes
+    And I have a cluster with capacity of 1 nodes
     And no nodes are declared as blend providers
     And I start nodes with wallet resources:
       | node_name | account_index | wallet_name | connected_to |
       | NODE_1    | 1             | WALLET_1A   |              |
-    And I start peer node "NODE_2" connected to node "NODE_1"
     When all nodes have at least 2 blocks and converged to within 1 blocks in 180 seconds
-    And I send 1 transactions of 1000 LGO each from wallet "WALLET_1A" to blend core zk key of node "NODE_2"
-    Then I run blend core SDP declaration CLI for node "NODE_2" against node "NODE_2"
+    And I send 1 transactions of 1000 LGO each from wallet "WALLET_1A" to blend core zk key of node "NODE_1"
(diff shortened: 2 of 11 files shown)


tools/blockchain-tools/src/bin/api.rs
```diff
@@ -0,0 +1,257 @@
...
+    })
+}
+
+// Validate and return the Blend listening address from the provided config.
+fn extract_blend_locator(config: &UserConfig) -> Result<Locator> {
+    config
+        .blend
+        .core
+        .backend
+        .listening_address
```
bacv, line 223: Maybe this should be passed by the user via cli args. If node is behind proxy, it will most likely listen on 0.0.0.0, other nodes won't be able to connect.

Edit: resolved, this case is handled by the cli args

bacv:
Looks good

bacv (approved):

tests/cucumber_tests/features/api-cli.feature
```diff
@@ -0,0 +1,17 @@
...
+    And I have a cluster with capacity of 1 nodes
+    And no nodes are declared as blend providers
+    And I start nodes with wallet resources:
+      | node_name | account_index | wallet_name | connected_to |
+      | NODE_1    | 1             | WALLET_1A   |              |
+    When all nodes have at least 2 blocks and converged to within 1 blocks in 180 seconds
+    And I send 1 transactions of 1000 LGO each from wallet "WALLET_1A" to blend core zk key of node "NODE_1"
+    Then I declare node "NODE_1" as blend core node via the CLI binary
+    And blend core SDP declaration for node "NODE_1" is included on node "NODE_1"
+    And I stop all nodes
```
danielSanchezQ: Nice to see cucumber tests for this!

tools/blockchain-tools/src/bin/api.rs
```diff
@@ -0,0 +1,257 @@
...
+    /// the one present in the config file. This is useful for the case in which
+    /// a node is listening on the `0.0.0.0` address, and the declaration needs
+    /// to be posted with the externally reachable address, since `0.0.0.0` is
+    /// not a valid `Locator` value.
+    #[arg(long, value_name = "BLEND_ADDR")]
+    blend_addr: Option<Locator>,
+
+    /// Note ID to lock for the Blend declaration (HEX-encoded field element).
+    #[arg(long, value_name = "NOTE_ID_HEX", value_parser = parse_hex_serde::<NoteId>)]
+    locked_note_id: NoteId,
```
danielSanchezQ, line 97: Can't we derive this from somewhere? It can stay as optional, otherwise it will just lock any note that fits?
**Not to be done now, just if we simplify the API at some point**.

tools/blockchain-tools/src/bin/api.rs
```diff
@@ -0,0 +1,257 @@
...
+
+    /// Base node URL, for example `http://localhost:8080`.
+    #[arg(long, value_name = "NODE_URL")]
+    node_address: Url,
```
danielSanchezQ, line 101: Can we use localhost by default?

tools/blockchain-tools/src/bin/api.rs
ntn-x2: Yeah I did not spend too much time thinking about this. I am not sure if there would specific reasons why someone would want to choose a specific note ID, maybe they want the smallest that fits the minimum stake requirement.

tools/blockchain-tools/src/bin/api.rs
ntn-x2: https://github.com/logos-blockchain/logos-blockchain/pull/2695/commits/a7965ce79cea6e4fe222162baf091b879b8480f3

tests/src/cucumber/steps/manual_nodes/steps.rs
```diff
@@ -1002,3 +1009,271 @@ fn step_stop_all_nodes(world: &mut CucumberWorld) -> StepResult {
...
+            message: format!("Node '{declarer_node_name}' not found in world state"),
+        })?
+        .started_node
+        .client
+        .base_url()
+        .clone();
+
+    // Query notes from the declarer node API so wallet ownership lookups use
+    // the same node config as the key source.
+    let locked_note_id = loop {
```
andrussal, line 1110: I think it would be nice to have a timeout around the polling loop.

.github/workflows/end-to-end-with-cucumber.yml
```diff
@@ -187,6 +187,16 @@ jobs:
           artifact_subdir: Transactions
           verbose_console: "true"
 
+      # Feature: CLI utility binary (@cli_ci)
+      - name: Run cucumber suite - CLI utilities
+        if: ${{ !cancelled() && steps.verify_local_ntp.conclusion == 'success' }}
+        timeout-minutes: 60
+        uses: ./.github/actions/run-cucumber-suite
+        with:
+          suite_tag: "@cli_ci"
```
andrussal, line 196: I think this should match the scenario tag, otherwise this new scenario will not be picked up by the CI suite.

ntn-x2: Ops. A renaming issue.

tests/src/cucumber/steps/manual_nodes/steps.rs
ntn-x2: https://github.com/logos-blockchain/logos-blockchain/pull/2695/commits/43cadfa04a0fb243c996ee6e0bb6b1e3f6d9eb48

.github/workflows/end-to-end-with-cucumber.yml
ntn-x2: https://github.com/logos-blockchain/logos-blockchain/pull/2695/commits/43cadfa04a0fb243c996ee6e0bb6b1e3f6d9eb48

PR #2697, opened 2026-05-06
chore(zone-sdk): simplify reorg event

pradovic:

This simplifies the `ChannelUpdate` event, reducing to only { orphaned , adopted }, deduped by pending tx. 

It also simplifies the re-submission policies and adds another e2e test for concurrent sequencers. Also adds more e2e tests to cover new scenarios.

Also, now multiple sequencers with same signing keys are supported, so the orphaned and adopted are filtered per pending tx set and not per signing key.

Yes

@pradovic @davidrusu 

Yes

No, the spec for FFW is still in progress.

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -1045,6 +1037,252 @@ async fn test_sorted_conflict_resolution() {
...
+    publish_concurrently(vec![
+        (handle_a, payloads_a.clone()),
+        (handle_b, payloads_b.clone()),
+        (handle_c, payloads_c.clone()),
+    ])
+    .await;
+
+    // Wait for things to settle. With k=10 and 1s slots, finality is ~10s
+    // per block. We wait long enough for retry cycles + finality.
+    sleep(Duration::from_secs(180)).await;
```
hansieodendaal: I think "Wait for things to settle" can be more deterministic, as each sequencer will report events detailing state.

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -769,14 +755,14 @@ async fn test_concurrent_multi_sequencer() {
...
 
-    wait_for_indexer_unordered(&indexer, &expected_all, Duration::from_mins(20)).await;
+    wait_for_indexer_unordered(&indexer, &expected_all, Duration::from_mins(30)).await;
 
     // Wait for enough blocks so any late re-published duplicates would have
-    // landed. With k=3 and 1s slots, finality is ~3 blocks. We wait 30s
+    // landed. With k=10 and 1s slots, finality is ~10 blocks. We wait 60s
     // to be safe — enough for resubmit cycles and in-flight txs to settle.
-    sleep(Duration::from_secs(30)).await;
+    sleep(Duration::from_secs(60)).await;
```
hansieodendaal: ditto - could this not be deterministic?

testnet/tui-zone/src/state.rs
```diff
@@ -105,39 +90,25 @@ impl ZoneState for InMemoryZoneState {
...
+/// track current canonical membership. Once a message is added it stays;
+/// there is no revert. Reorgs and bouncing don't change what was published,
+/// only where it lives on chain.
 ///
-/// Authorship is read from `state.is_ours`, which is durable across reorgs —
-/// so the order is the natural one: mutate state first, then decide based on
-/// the new chain.
+/// 1. Apply adopted to state (others' new inscriptions).
+/// 2. Iterate orphaned and return entries not in `pending` — those are
+///    republish candidates the user must decide on.
```
hansieodendaal: I think for this TUI demo we can assert that every orhaned entry carries our signature

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -291,48 +291,26 @@ fn spawn_drive(
...
-                    && !state.contains(&inv.payload)
-                    && !pending_payloads.contains(&inv.payload)
-                {
-                    debug!(
-                        "Re-publishing invalidated: {:?}",
-                        String::from_utf8_lossy(&inv.payload)
-                    );
+            for inv in &orphaned {
+                if !pending_payloads.contains(&inv.payload) {
+                    debug!("Re-publishing: {:?}", String::from_utf8_lossy(&inv.payload));
```
hansieodendaal: We should assert all orphaned carries our signature

hansieodendaal (commented):
Nice simplification and looking good overall. Possible areas for improvement are:

1. Assert ownership of orphaned explicitly
   Since the new contract is that orphaned only contains entries that belong to this sequencer, I think both the E2E helpers and the TUI demo should assert that invariant explicitly by checking `InscriptionInfo.signer == our_public_key`. That makes the new API contract executable and protects us against future regressions.

1. Replace magic waits with event-driven settling
   I think all the “wait for things to settle” type sleeps should be made more deterministic. Each sequencer already emits enough state-transition events to drive convergence checks, so we should prefer waiting on explicit conditions over fixed sleep(...) calls.

1. Add identical-payload concurrent conflict test
   We should add a concurrent competing-sequencer test where all payloads are identical. Right now most concurrent tests rely on unique payloads, but identical payloads are a useful edge case for dedup/conflict/republish behavior and would better exercise the simplified orphaned semantics.

tests/src/tests/zone_sdk/e2e.rs
```diff
@@ -1045,6 +1141,383 @@ async fn test_sorted_conflict_resolution() {
...
+    handle_a.wait_ready().await;
+    handle_b.wait_ready().await;
+    handle_c.wait_ready().await;
+
+    publish_concurrently(vec![
+        (handle_a, vec![payload.clone()]),
+        (handle_b, vec![payload.clone()]),
+        (handle_c, vec![payload.clone()]),
+    ])
+    .await;
```
hansieodendaal, line 1415: For this test to be effective, each sequencer client should publish multiple of the same, say 10 each, otherwise conflict resolution will not be exercised adequately.

All else looking good.

pradovic: Ah ok I see. I thought this is a know limitation of this approach, as the user cannot really distinguish the tx if they are all the same and make a decision weather to resubmit, but let me try, it is a good use case to document a justify more complex lineage tracking if needed. This is why we have tx_uuud in tui.

pradovic: Edit: @hansieodendaal proposed to use inscription info to make a policy, I will try

pradovic:
@hansieodendaal @davidrusu 

Regarding adopted filter — I think this is a design call.

Today in `build_channel_event`:

- `adopted` = block-delta inscriptions on the canonical branch, filtered to exclude signer == my_signer because our own publishes are already delivered via Event::Published.
- `pending / orphaned` = derived from our local pending tx state, not from signer filtering, so these semantics would stay the same.

The current filter gives us a simple “is this mine?” signal under the one-key-per-sequencer assumption.

To support shared signing keys across multiple instances, we would need to drop the adopted signer filter. Then applications would need to maintain a local `HashSet<MsgId>` from `Event::Published` and dedup adopted against it to avoid double-counting their own messages.

I lean toward keeping the filter and explicitly not supporting shared keys for now. Multiple sequencers feels like a decentralization pattern — different signers per instance — not horizontal scaling under one identity. I can add a doc note making that assumption explicit.

WDYT?

tests/src/tests/zone_sdk/e2e.rs
pradovic: Done

hansieodendaal:
> @hansieodendaal @davidrusu
> 
> Regarding adopted filter — I think this is a design call.
> 
> Today in `build_channel_event`:
> 
> WDYT?

Does a client not know the tx_id of their own submitted inscription? Then a shared signing key should not be a problem, and the adopted filter can be kept.

pradovic:
> > @hansieodendaal @davidrusu
> 
> > 
> 
> > Regarding adopted filter — I think this is a design call.
> 
> > 
> 
> > Today in `build_channel_event`:
> 
> > 
> 
> > WDYT?
> 
> 
> 
> Does a client not know the tx_id of their own submitted inscription? Then a shared signing key should not be a problem, and the adopted filter can be kept.

It can, the info is there but it complicates the tui implementation to cover this case. Also, I see several problems there:
1. Adopted would filter other inscriptions (as they have same signing key), so we will not see adopted at all.
2. With same gas prices + signing key + payload + same parent_id even tx_id can be the same in the mempool, meaning same tx, deduped in mempool. With different gas prices, different behavior. Different signing keys unify the API in this case better at the moment (until tke is implemented).

I think this is confusing for the user regarding the API, but also I am not against it fully, so wanted feedback from both of you guys.
To me the decentralized seq -> different signing keys (must be supported),
while distributed seq -> the same signing keys -> horizontally scaled "same" sequencer.

I just noticed round-robin implementation so pls let's align here: https://discord.com/channels/973324189794697286/1502951288932012042/1502951292639641763.

hansieodendaal (approved):
LGTM

davidrusu (approved):
Thanks for going through so many versions on this, it's looking much better

PR #2700, opened 2026-05-06
fix: orphan recovery node stalling

hansieodendaal:

This PR hardens orphan recovery and inbound sync behavior under load. 
- Orphan recovery fix: new orphan fetches now start with meaningful known anchors (tip + lib) instead of an empty set, improving provider path resolution and reducing recovery stalls. 
- Backpressure redesign: inbound sync no longer drops requests immediately when active slots fill; it now uses bounded queueing, timeout-based expiry, and fair promotion/per-peer caps so drops happen only as a last resort. 
- Test coverage update: tests were updated/added to lock in the new semantics (queue-before-drop, promotion, timeout expiry, fairness, and expire-before-promote ordering).
- Net effect: recovery is more resilient during orphan storms, and overload handling degrades gracefully instead of amplifying failures.

-----------

Observed system-level failures in testnet (Evidence from node_02 logs):

- At 12:40:06, node processes 28a926... at slot 1372835.
- By 12:47:57, it reaches the stalled state:
  - lib=28a926...
  - tip=5f42f3...
  - slot=1373304
  - height=70480
- Immediately after, blocks arrive whose parents are missing, causing repeated:
  - Failed to process block: ParentMissing
  - Orphan block enqueued for sync
  - queue grows rapidly: 57 → 66+
- Recovery then fails because block downloads from peers fail:
  - Failed to create a block stream: BlockNotFound(...)
  - StartBlockNotFound
  - repeated QUIC handshake timeouts
  - Rejected excess pending incoming request

Likely reason: the node accepted/advanced to a tip, but later received descendants/forks whose ancestry was not locally available; orphan sync then hammered peers, but peers either did not have the requested blocks or could not serve them due to connection/request saturation.

-----------

Yes

@hansieodendaal and **???**

**???**

**???**

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/consensus/cryptarchia-sync/src/libp2p/behaviour.rs
+++ b/consensus/cryptarchia-sync/src/libp2p/behaviour.rs
@@ -1,7 +1,8 @@
 use core::future::ready;
 use std::{
-    collections::HashSet,
+    collections::{HashMap, HashSet, VecDeque},
     task::{Context, Poll},
+    time::Instant,
 };
 
 use futures::{AsyncWriteExt as _, FutureExt as _, StreamExt as _, future::BoxFuture};
@@ -17,7 +18,7 @@ use libp2p::{
 };
 use libp2p_stream::{Behaviour as StreamBehaviour, Control, IncomingStreams};
 use tokio::sync::{mpsc, mpsc::Sender, oneshot};
-use tracing::{debug, error};
+use tracing::{debug, error, warn};
 
 use crate::{
     BlocksResponse, DownloadBlocksRequest, TipResponse,
@@ -31,7 +32,9 @@ use crate::{
     messages::{GetTipResponse, SerialisedBlock},
 };
 
-const MAX_INCOMING_REQUESTS: usize = 4;
+pub const MAX_ACTIVE_INCOMING_REQUESTS: usize = 4;
+pub const MAX_QUEUED_INCOMING_REQUESTS: usize = 32;
+pub const MAX_QUEUED_INCOMING_REQUESTS_PER_PEER: usize = 8;
 
 type SendingBlocksRequestsFuture = BoxFuture<'static, Result<BlocksRequestStream, ChainSyncError>>;
 
@@ -46,6 +49,12 @@ type SendingBlocksResponsesFuture = BoxFuture<'static, Result<(), ChainSyncError
 type SendingTipResponsesFuture = BoxFuture<'static, Result<(), ChainSyncError>>;
 
 type ReceivingRequestsFuture = BoxFuture<'static, Result<ReceivingRequestStream, ChainSyncError>>;
+
+struct PendingIncomingStream {
+    peer_id: PeerId,
+    stream: Libp2pStream,
+    enqueued_at: Instant,
+}
 
 pub type BoxedStream<T> = Box<dyn futures::Stream<Item = T> + Send + Unpin>;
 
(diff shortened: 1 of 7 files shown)


consensus/cryptarchia-sync/src/libp2p/behaviour.rs
```diff
@@ -31,7 +32,9 @@ use crate::{
     messages::{GetTipResponse, SerialisedBlock},
 };
 
-const MAX_INCOMING_REQUESTS: usize = 4;
+pub const MAX_ACTIVE_INCOMING_REQUESTS: usize = 4;
+pub const MAX_QUEUED_INCOMING_REQUESTS: usize = 32;
+pub const MAX_QUEUED_INCOMING_REQUESTS_PER_PEER: usize = 8;
```
youngjoon-lee, line 38: I was working on factor this out as a parameter. But, I will resolve conflicts anyway.

consensus/cryptarchia-sync/src/libp2p/behaviour.rs
```diff
@@ -145,6 +154,11 @@ pub struct Behaviour {
     /// Futures for closing incoming streams that were rejected due to excess
     /// requests.
     incoming_streams_to_close: FuturesUnordered<BoxFuture<'static, ()>>,
+    /// Bounded FIFO queue for inbound streams waiting for active capacity.
+    pending_incoming_streams: VecDeque<PendingIncomingStream>,
```
youngjoon-lee, line 158: Do we really need this queue? IMO, we are making this network behaviour complicated. We already discussed spreading IBD requests across many (untrusted) nodes, and factoring the constant `MAX_INCOMING_REQUESTS` as a config parameter. I believe it is sufficient for tuning the node capacity. 

Of course, queuing requests would be helpful to some extents, but I don't think it's a real solution. It just delays hitting the limit. In traditional client-service models, it's common to configure only the max concurrent connections, afaik. If we hit the limit, we scale out servers. I think we can follow the same practice.

services/chain/chain-network/src/bootstrap/ibd.rs
```diff
@@ -109,6 +109,88 @@ where
     BlockProcessor: IbdBlockProcessor<NetAdapter::Block> + Sync,
     RuntimeServiceId: Sync,
 {
+    async fn resolve_peer_target(&self, peer: NetAdapter::PeerId) -> Result<HeaderId, Error> {
```
youngjoon-lee, line 112: What are these changes for? PR description doesn't explain about these, I think.

services/chain/chain-network/src/sync/orphan_handler.rs
```diff
@@ -283,8 +287,163 @@ where
...
 
-        self.dequeue_next_orphan()
-            .map(|orphan_info| (orphan_info, HashSet::new()))
+        self.dequeue_next_orphan().map(|orphan_info| {
+            let known_blocks = HashSet::from([orphan_info.tip, orphan_info.lib]);
+            (orphan_info, known_blocks)
+        })
+    }
+
+    fn poll_idle_state(&mut self, cx: &Context<'_>) -> Poll<Option<NetAdapter::Block>> {
```
youngjoon-lee, line 296: Also, what are these changes for? Just refactoring? If so, please leave PR comments for easier review.

youngjoon-lee: > Orphan recovery fix: new orphan fetches now start with meaningful known anchors (tip + lib) instead of an empty set, improving provider path resolution and reducing recovery stalls.

If this was not implemented before, it is a bug. Could you please point where I can find changes for this fix?

consensus/cryptarchia-sync/src/libp2p/behaviour.rs
hansieodendaal: Ok, I will retire this PR (keep it in my _library_), and just make a small PR for the primary culprit.

services/chain/chain-network/src/bootstrap/ibd.rs
hansieodendaal: Those were clippy instigated

hansieodendaal:
Closing this in favour of a small PR for the primary culprit.

New replacement PR in #2709.

PR #2706, opened 2026-05-07
feat(logging): add runtime log reload

andrussal:

This PR adds runtime logging reload support to the node through a new admin API endpoint. This is mainly useful when debugging remote nodes, where changing the logging filter without asking node operators to restart the process can make investigation easier.

There are also a couple of related changes:

- Introduce a separate admin API endpoint and port. This keeps system-level controls separate from the regular node API, since operators may want to expose and protect them differently. By default it listens only on the local interface, but operators can still use something like SSH port forwarding when they need remote access. Strictly speaking this is not required; if the regular API is considered secure enough for this, the endpoint can be moved there.
- Remove the node-level `tracing` feature flag and always include tracing support. Keeping tracing conditional made the admin API wiring more awkward, and there does not seem to be a useful build mode where the node runs without tracing.

Example:

```bash
curl -X PUT http://127.0.0.1:8082/admin/tracing/filter \
  -H 'content-type: application/json' \
  -d '{
    "filters": {
      "*": "WARN",
      "logos_blockchain": "DEBUG",
      "libp2p_gossipsub::behaviour": "ERROR"
    }
  }'
```

Tracing and admin endpoint changes could be a separate PR but I kept these here because right now they are only meaningful in the context of runtime reload.

Yes

@bacv @andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/services/tracing/src/lib.rs
+++ b/services/tracing/src/lib.rs
@@ -28,27 +28,74 @@ use serde::{Deserialize, Serialize};
 use tracing::{Level, warn};
 use tracing_appender::non_blocking::WorkerGuard;
 use tracing_subscriber::{
-    Layer as _, filter::LevelFilter, layer::SubscriberExt as _, util::SubscriberInitExt as _,
+    EnvFilter, filter::LevelFilter, layer::SubscriberExt as _, util::SubscriberInitExt as _,
 };
 
 #[cfg(feature = "profiling")]
 mod console;
 
-macro_rules! push_logger_layer {
-    ($layers:expr, $env_filter:expr, $layer:expr) => {{
-        let layer = $layer;
-        if let Some(filter) = $env_filter.clone() {
-            $layers.push(Box::new(layer.with_filter(filter)));
-        } else {
-            $layers.push(Box::new(layer));
-        }
-    }};
-}
+type LoggerSubscriber =
+    tracing_subscriber::layer::Layered<LevelFilter, tracing_subscriber::Registry>;
+type FilterReloadHandle = tracing_subscriber::reload::Handle<EnvFilter, LoggerSubscriber>;
 
 pub struct Tracing<RuntimeServiceId> {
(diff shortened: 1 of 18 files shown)


services/tracing/src/lib.rs
```diff
@@ -28,29 +28,76 @@ use serde::{Deserialize, Serialize};
...
+
+        self.layers.push(Box::new(layer.with_filter(filter)));
+        self.reload_handles.push(reload_handle);
+    }
+
+    fn add_guarded_layer<L>(&mut self, layer: L, guard: WorkerGuard)
+    where
+        L: tracing_subscriber::Layer<LoggerSubscriber> + Send + Sync + 'static,
+    {
+        self.add_layer(layer);
```
bacv, line 96: Just curious, why guarded layers doesn't get the filter and reload_handle?

bacv (approved):
Looks good, the "admin" http server works for now, but I think eventually it might get too vague, like shouldn't only admins use the wallet api? Maybe mempool info should also go into admin. 

But in the scope of this PR it makes sense, I think it's good to go, just that we need to have that discussion later.

andrussal:
> Looks good, the "admin" http server works for now, but I think eventually it might get too vague, like shouldn't only admins use the wallet api? Maybe mempool info should also go into admin.
> 
> But in the scope of this PR it makes sense, I think it's good to go, just that we need to have that discussion later.

Yeah, agree, I’d treat wallet endpoints as sensitive, especially transfer/signing since they can spend funds or sign arbitrary tx hashes with keys loaded in the node.

services/tracing/src/lib.rs
andrussal: Actually they do get the filter and reload handle. `add_guarded_layer` delegates to `add_layer`, and `add_layer` wraps the layer with the reloadable filter and stores the handle.

bacv: > Actually they do get the filter and reload handle.

I epically misread the function body, thank you for patience.

PR #2709, opened 2026-05-07
fix: orphan recovery node stalling

hansieodendaal:

This PR tries to solves issue #2738 (Orphan recovery results in node stalling)

So far, after the first rounds of comments, only two unit tests remains:
- `test_empty_known_blocks_returns_start_block_not_found()`
- `test_known_blocks_with_unknown_tip_and_known_lib_succeeds()`

-----------

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


services/chain/chain-network/src/sync/orphan_handler.rs
```diff
@@ -283,8 +283,10 @@ where
             return Some((orphan_info, HashSet::from([last_block_id])));
         }
 
-        self.dequeue_next_orphan()
-            .map(|orphan_info| (orphan_info, HashSet::new()))
+        self.dequeue_next_orphan().map(|orphan_info| {
+            let known_blocks = HashSet::from([orphan_info.tip, orphan_info.lib]);
+            (orphan_info, known_blocks)
```
youngjoon-lee: I just took a look at this module in detail, and I think we don't need this fix. 

I agree that this module is not easy to read. I will refactor this module substantially after June. Anyway, I just found that we are actually passing tip and lib correctly to the provider, through the following code path.



https://github.com/logos-blockchain/logos-blockchain/blob/17d584371f54686a1514fec51c867b0cd0a82636/services/chain/chain-network/src/sync/orphan_handler.rs#L317-L323

https://github.com/logos-blockchain/logos-blockchain/blob/17d584371f54686a1514fec51c867b0cd0a82636/services/chain/chain-network/src/sync/orphan_handler.rs#L205-L212

https://github.com/logos-blockchain/logos-blockchain/blob/17d584371f54686a1514fec51c867b0cd0a82636/services/chain/chain-network/src/network/adapters/libp2p.rs#L360-L368

https://github.com/logos-blockchain/logos-blockchain/blob/17d584371f54686a1514fec51c867b0cd0a82636/services/chain/chain-network/src/network/adapters/libp2p.rs#L106-L114

https://github.com/logos-blockchain/logos-blockchain/blob/17d584371f54686a1514fec51c867b0cd0a82636/services/chain/chain-network/src/network/adapters/libp2p.rs#L293-L301

Actually, I think it's better to include the genesis block in the known blocks.
I think the reason why you got `StartBlockNotFound` is that, the LIB and tip of the downloader node are not found in the DB of the provider node. It means that those two nodes have been permanently forked, for some reason.
If we add the genesis block to the known blocks, the provider node will return a stream starting from the genesis block instead of throwing `StartBlockNotFound`.
https://github.com/logos-blockchain/logos-blockchain/blob/17d584371f54686a1514fec51c867b0cd0a82636/services/chain/chain-service/src/sync/block_provider.rs#L295-L304

hansieodendaal: So what will be the fix for this issue? Two out of 7 nodes encountered exactly the same issue, and inspecting the logs those two nodes were in sync up to events described above.

hansieodendaal: And yes agreed, the change does not fix the issue. I will revert the code change, but keep the applicable tests. Making this PR a draft until we find a solution, creating an issue with applicable log files and repo commit.

hansieodendaal:
Need a new plan for this

PR #2712, opened 2026-05-07
chore: validate format of a locator

ntn-x2:

By definition, a locator is a multiaddress that does not contain a peer ID. Also, because of its ambiguity, an undefined IPv4 or IPv6 should not be used as a locator address, so we filter those out as well. 

Yes.

@ntn-x2 

N/A

No.

### DIFF
--- a/core/src/sdp/mod.rs
+++ b/core/src/sdp/mod.rs
@@ -1,11 +1,12 @@
 pub mod blend;
 pub mod locked_notes;
 
+use core::str::FromStr;
 use std::hash::Hash;
 
 use blake2::{Blake2b, Digest as _};
 use lb_key_management_system_keys::keys::ZkPublicKey;
-use multiaddr::Multiaddr;
+use multiaddr::{Multiaddr, Protocol};
 use nom::{IResult, Parser as _, bytes::complete::take};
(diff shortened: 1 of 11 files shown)


core/src/sdp/mod.rs
```diff
@@ -60,6 +66,47 @@ impl AsRef<Multiaddr> for Locator {
...
+                Protocol::Ip4(ip) if ip.is_unspecified() => {
+                    return Err(format!(
+                        "Locator multiaddr must not contain an unspecified IPv4 address: {value}"
+                    ));
+                }
+                Protocol::Ip6(ip) if ip.is_unspecified() => {
+                    return Err(format!(
+                        "Locator multiaddr must not contain an unspecified IPv6 address: {value}"
+                    ));
+                }
```
danielSanchezQ, line 84: We should include dns at least. Although im not sure right now if we have dns resolve supported in blend atm.

danielSanchezQ (approved):

core/src/sdp/mod.rs
ntn-x2: We have! We hit the issue when running in DST, but we added support for DNS, so it works.

PR #2719, opened 2026-05-08
feat(logging): normalise default node logs

andrussal:

As discussed in Discord, `INFO` should be the default logging level because most node runners are interested in high-level node progress, not low-level implementation details. (Internally we should keep DEBUG level though).

This PR changes the default level to `INFO` and makes the final logging-level adjustments around that change, including a few wording updates so the main startup and chain-progress logs are easier to follow.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/core/src/mantle/ops/sdp/active.rs
+++ b/core/src/mantle/ops/sdp/active.rs
@@ -1,7 +1,7 @@
 use lb_cryptarchia_engine::Epoch;
 use lb_key_management_system_keys::keys::{ZkPublicKey, ZkSignature};
 use lb_log_targets::mantle;
-use tracing::debug;
+use tracing::info;
 
 use super::{SDPActiveOp, SdpError};
 use crate::{
@@ -78,7 +78,7 @@ impl Operation<SDPActiveValidationContext<'_>> for SDPActiveOp {
 
         declaration.active = ctx.epoch;
         declaration.nonce = self.nonce;
-        debug!(
+        info!(
             target: LOG_TARGET,
             provider_id = ?declaration.provider_id,
             active = ?declaration.active,
--- a/services/chain/chain-service/src/lib.rs
+++ b/services/chain/chain-service/src/lib.rs
@@ -691,7 +691,9 @@ where
                     }
 
                     () = async { prolonged_bootstrap_timer.as_mut().unwrap().as_mut().await }, if prolonged_bootstrap_timer.is_some() && cryptarchia.is_bootstrapping() => {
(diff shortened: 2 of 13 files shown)


core/src/mantle/ops/sdp/active.rs
```diff
@@ -66,7 +66,7 @@ impl Operation for SDPActiveOp {
 
         declaration.active = ctx.block_number;
         declaration.nonce = self.nonce;
-        info!(
+        debug!(
```
youngjoon-lee: IMO, it's good to keep `info` for this log because active messages are submitted once per epoch. Also, this log is useful for debugging because it's very hard to reproduce this operation. But, if there are so many declarations in the network, too many `info` logs may be printed. So, I'm fine with this change.

youngjoon-lee (approved):

davidrusu (approved):

danielSanchezQ (approved):

core/src/mantle/ops/sdp/active.rs
andrussal: > IMO, it's good to keep `info` for this log because active messages are submitted once per epoch. Also, this log is useful for debugging because it's very hard to reproduce this operation. But, if there are so many declarations in the network, too many `info` logs may be printed. So, I'm fine with this change.

Good point, changed this back to `info`.

services/chain/chain-service/src/lib.rs
```diff
@@ -1403,7 +1406,11 @@ where
             relays.storage_adapter().clone(),
         )
         .await;
-        info!(target: LOG_TARGET, "loaded {} blocks from storage: (lib, tip]", blocks.len());
+        info!(
+            target: LOG_TARGET,
+            "found {} stored blocks to replay during chain recovery",
```
ntn-x2, line 1411: This sounds more like a `debug` directive to me.

services/chain/chain-service/src/lib.rs
```diff
@@ -1392,7 +1395,7 @@ where
         // Phase 1: Collect and load blocks in (LIB, tip].
         info!(
             target: LOG_TARGET, lib = ?lib_id, tip = ?self.state.tip,
-            "loading blocks from storage: (lib, tip]",
+            "loading stored blocks for chain recovery",
```
ntn-x2, line 1398: Same here.

ntn-x2 (commented):
Just left comments about something we might want to make DEBUG instead of INFO.

PR #2722, opened 2026-05-08
fix(blend): handle better core->edge transitions

ntn-x2:

We had code in place to gracefully handle the core->edge transition, but the stream adapter was assuming core node stuff, that is not held when transitioning into edge mode. I fixed that and added some more tests to cover such case.

Yes.

@ntn-x2 @youngjoon-lee 

Yes.

No.

### DIFF
--- a/services/blend/src/core/mod.rs
+++ b/services/blend/src/core/mod.rs
@@ -597,6 +597,14 @@ where
                         session: session_number,
                     };
                 };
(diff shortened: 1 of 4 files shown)


danielSanchezQ (approved):

youngjoon-lee (commented):
look good. good catch!

PR #2728, opened 2026-05-09
feat(blockchain-tools): Tools genesis block ceremony command

bacv:

Unified ceremony command to generate `deployment-settings` with genesis block from chain params, stakeholders and providers. 

Yes

@danielSanchezQ @bacv 

N/A

N/A

## Checklist

* [x] 2. Description added.


tools/blockchain-tools/src/bin/genesis.rs
```diff
@@ -169,13 +203,53 @@ struct InscribeArgs {
...
+
+fn run_ceremony(args: &CeremonyArgs) -> Result<()> {
+    let inscribe_params: InscribeParams = load_yaml_file(&args.inscription_params)?;
+    let inscription_op = inscription::inscribe::<ZkHasher>(
+        inscribe_params.chain_id,
+        inscribe_params.genesis_time,
+        inscribe_params.entropy_sources,
+    );
+
+    let stakeholders: Vec<StakeHolderInfo> = load_yaml_file(&args.stake_holders)?;
```
hansieodendaal, line 233: I can appreciate this PR, but this feels wrong, that `StakeHolderInfo` can be read from a file and not be compiled into the binary with prior PR to approve it.

bacv: > not be compiled into the binary with

I think I don't understand this, but its important to have shared agreement here. Without this PR, we would still have a way to generate a new genesis block by using seperate commands:
- inscribe - generates `inscription_op.yaml`
- distribute - generates `notes.yaml` and `declarations.yaml`
- block - genrates `genesis_block.yaml` from all the prior yaml outputs.
- config - generates `deployment-settings.yaml` from the template and the `genesis_block.yaml`.

This `ceremony` command combines all of these command at code level, to remove the need for scripts. Do you think I missed something when creating this combined command? It takes inputs for `inscribe` `distribute` and `config` commands.

danielSanchezQ (approved):

tools/blockchain-tools/src/bin/genesis.rs
hansieodendaal: No, nothing is amiss, and as you explained, stake can be verified in plain text in the genesis block.

hansieodendaal (approved):
LGTM

hansieodendaal (commented):
Ah wait, what about the faucet? I think it must be visible in the stakeholder's yaml file and get assigned a value explicitly if at all.

bacv:
> Ah wait, what about the faucet? I think it must be visible in the stakeholder's yaml file and get assigned a value explicitly if at all.

I really didn't think about this when creating a PR. What is you opinion if we encode the "faucet" as a stakeholder into the `stakeholders.yaml` file with the stake amount of max u64 value? But this is problematic as it might skew the block proposal stake distribution.

@davidrusu maybe you have opinion how we should encode faucet when generating  a new genesis block?

davidrusu:
I did not look at the code, but we need to encode the faucet pk in the deployment config explicitly  so that the faucet node knows to exclude it from PoL.

So if the faucet pk is in deployment config, the initial total stake should exclude this faucet stake from the sum (filter by pk != faucet_pk).

danielSanchezQ:
> I did not look at the code, but we need to encode the faucet pk in the deployment config explicitly so that the faucet node knows to exclude it from PoL.
> 
> So if the faucet pk is in deployment config, the initial total stake should exclude this faucet stake from the sum (filter by pk != faucet_pk).

Is there not anyway else of doing it? This seems a completely ad-hoc fix for the faucet which imo should not really go in the deployment config. I think its better to use the stakeholders.yaml file later for the deployment itself and do the calculations in a script or something?

Maybe im missing something too tho.

bacv:
>  I think its better to use the stakeholders.yaml file later for the deployment itself and do the calculations in a script or something?

This is the problem. We need a faucet to have funds somehow, we need the faucet not to skew the PoL stake. 

The previous solution was to explicitly embed faucet key into a `deployment-settings.yaml` so that node could exclude that from the local PoL calculation.

We can still use this, if the current PR stack is updated to explicitly take the faucet_key as a pram and generate `deployment-settings.yaml` by taking it into account.

davidrusu:
We don't need to mix the faucet distribution with the stakeholder distribution, we can introduce a separate output just for the faucet that is not part of the stakeholder.yaml. (just as we have been doing for the last couple testnets).

I do think it would be better not to have the faucet distribution in the stakeholder.yaml that way the initial total stake estimate is just a straight sum over the stakeholder.yaml

bacv:
> I do think it would be better not to have the faucet distribution in the stakeholder.yaml

It can be excluded, but it will stil be in the genesis block. Semantically it would be more correct if we included it into the stakeholders list.

Or, then we need a separate Operation for the faucet in genesis block.

davidrusu:
Well, it could just be one last output that is tacked onto the TransferOp, does not need to be a separate operation.

> Semantically it would be more correct if we included it into the stakeholders list.

But the faucet is not a stakeholder, (we want to exclude it from PoS) so I do think it makes sense to special case the faucet distribution in the genesis builder

bacv:
> But the faucet is not a stakeholder, (we want to exclude it from PoS) so I do think it makes sense to special case the faucet distribution in the genesis builder

Okay, if we don't include it into the stakeholders file, then it can be added as a last note after all notes for the PoL is generated. 

It would be a special case and would have a separate configuration:
- stakeholders distrubution
- providers information
- faucet amount (configurable, by default u64::MAX)

If that sounds right, I'll add another PR on top of this.

hansieodendaal:
Sounds good, but maybe not give the faucet a default value so it can be reasoned about in its separate config.

PR #2732, opened 2026-05-10
chore(genesis): Genesis ceremony in CI

bacv:

Workflow to create a new `nodes/node/binary/src/config/deployment/devnet/deployment.yaml` file by using `logos-blockchain-tools-genesis` and parameters defined in `testnet/ceremony/genesis/{env}/`.

Workflow needs to be triggered manually with two params: environemnte (e.g. devnet) and protocol version number (e.g. 0.1.3-rc.6), workflow succeeds by committing to a new branch and giving a link to create a PR with new `deployment.yaml`.

Example workflow run: https://github.com/logos-blockchain/logos-blockchain/actions/runs/25632093587

@ntn-x2 @danielSanchezQ @bacv 

N/A

N/A

N/A

## Checklist

* [x] 2. Description added.


tools/blockchain-tools/src/bin/genesis.rs
```diff
@@ -423,15 +423,15 @@ fn run_inscribe(args: &InscribeArgs) -> Result<()> {
...
 ///    unconditionally; `serde_yml::to_string` rejects those with an error.
 ///
-/// Using `serde_json` as an intermediate format avoids both problems: JSON is
-/// a human-readable format (fixing pitfall 1), and its `serialize_bytes`
-/// implementation emits a JSON array of integers (fixing pitfall 2). JSON is
-/// a strict subset of YAML, so `serde_yml::from_str` can parse the resulting
-/// JSON string transparently, and YAML's human-readable deserializer correctly
-/// interprets all field formats.
+/// Using `serde_yaml::to_string` as an intermediate format avoids both
+/// problems: is a human-readable format (fixing pitfall 1).
```
ntn-x2:
```suggestion
/// problems: YAML is a human-readable format (fixing pitfall 1).
```

nodes/node/binary/src/config/deployment/devnet/mod.rs
```diff
@@ -0,0 +1,3 @@
+pub const NAME: &str = "devnet";
+
+pub const SERIALIZED_DEPLOYMENT: &[u8] = include_bytes!("deployment.yaml");
```
ntn-x2, line 3: `include_bytes!` was giving issues with Nix, that's why we chose the raw string approach. Cc @danielSanchezQ

.github/workflows/genesis-ceremony.yml
```diff
@@ -0,0 +1,72 @@
...
+  workflow_dispatch:
+    inputs:
+      deployment:
+        description: "Target deployment environment (e.g., devnet, testnet)"
+        required: true
+        default: "devnet"
+      version:
+        description: "Protocol version string"
+        required: true
+        default: "X.Y.Z-rc.0"
```
ntn-x2: I don't think we need a default here if this is the default we would use 😂

.github/workflows/genesis-ceremony.yml
```diff
@@ -0,0 +1,72 @@
...
+        uses: actions/checkout@85e6279cec87321a52edac9c87bce653a07cf6c2  # Version 4.2.2
+
+      - name: Setup Circuits
+        uses: ./.github/actions/setup-circuits
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Inject Variables into Template
+        run: |
+          cp testnet/ceremony/genesis/$LB_TARGET_ENV/deployment-template.yaml \
```
ntn-x2: If we're using the target env to access a directory, do you think it makes sense to make it an enum value in the input description, so we can only use values for which we have directories in place?

.github/workflows/genesis-ceremony.yml
```diff
@@ -0,0 +1,72 @@
...
+            /tmp/deployment-generated.yaml
+
+          # We use '|' as a delimiter so that slashes in paths don't break the command
+          sed -i "s|ENV_PLACEHOLDER|$LB_TARGET_ENV|g" /tmp/deployment-generated.yaml
```
ntn-x2: Curiosity, why did you not define placeholders with some sort of prefix, like `${}`?

.github/workflows/genesis-ceremony.yml
ntn-x2: We could add a test value if we want to test the workflow before merging a PR, if we need to.

.github/workflows/genesis-ceremony.yml
```diff
@@ -0,0 +1,72 @@
...
+          if git diff --staged --quiet; then
+            echo "Error: No changes detected in deployment.yaml. Genesis ceremony likely produced identical output."
+            exit 1
+          else
+            git commit -m "chore: genesis ceremony for $LB_TARGET_ENV version $LB_VERSION"
+            git push -u origin "$NEW_BRANCH"
+
+            PR_URL="${{ github.server_url }}/${{ github.repository }}/pull/new/$NEW_BRANCH"
+            echo "### Branch Created Successfully" >> $GITHUB_STEP_SUMMARY
+            echo "[Click here to create your Pull Request]($PR_URL)" >> $GITHUB_STEP_SUMMARY
```
ntn-x2: What would we need the PR for? I guess we would merge this into the release branch?

testnet/ceremony/genesis/devnet/inscribe.yaml
```diff
@@ -0,0 +1,4 @@
+chain_id: "devnet-0.1.3-rc6"
+genesis_time: "2026-05-08T22:54:09Z"
+entropy_sources:
+  - [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1]
```
ntn-x2: Do you think now it's a good time to change these to be HEX-encoded instead?

testnet/ceremony/genesis/devnet/deployment-template.yaml
```diff
@@ -0,0 +1,91 @@
...
+      BN:
+        lock_period: 10
+        inactivity_period: 1
+        retention_period: 1
+        timestamp: 0
+    min_stake:
+      threshold: 1
+      timestamp: 0
+  gossipsub_protocol: /logos-blockchain-ENV_PLACEHOLDER-VERSION_PLACEHOLDER/cryptarchia/1.0.0
+  genesis_block:
```
ntn-x2, line 42: Do we need stuff like genesis block in the template? I am not sure how it would play out with version control in the long term. We could remove stuff that are meant to be re-generated anyway by the ceremony tool, and we would just commit the final template to the repo, as a single source of truth.

ntn-x2 (commented):
Looks nice! I left some comments for some ideas we could implement.

nodes/node/binary/src/config/deployment/devnet/mod.rs
danielSanchezQ: Yes, but we decided to deal with it. Let's see how it goes.

.github/workflows/genesis-ceremony.yml
bacv: > I don't think we need a default here if this is the default we would use

Agree :D We need to validate that value somehow, and I think we could change cargo toml and flake.nix versions, trigger build. Cargo would fail if the version string is in incorrect format.

.github/workflows/genesis-ceremony.yml
bacv: > What would we need the PR for? I guess we would merge this into the release branch?

I am now thinking that if we trigger this from e.g. `pre-release-0.1.3` branch, the commit could be added the branch it was triggered from.

testnet/ceremony/genesis/devnet/deployment-template.yaml
bacv: > Do we need stuff like genesis block in the template?

Good point, pasted this from our previous deployment-settings.yaml, this can be cleaned up significantly.
Will do in a moment and ping you.

The main purpose of this template is to break the loop: WellKnown::Devnet -> Genesis -> WellKnown::Devnet.

Now it's: Template -> Genesis -> WellKnown::Devnet.

.github/workflows/genesis-ceremony.yml
ntn-x2: The branch name, according to our current release process, would always be `release/0.1.3`. But I am not sure I understand your second sentence.

testnet/ceremony/genesis/devnet/deployment-template.yaml
ntn-x2: Yes, I think we have the building block in place to fail if the YAML does not match the structure we expect, and I think the template should be much leaner than the actual deployment file and only contain stuff we want to end up in the final deployment. Other stuff should not be part of this `struct DeploymentTemplate` type, so we could commit the template to `master` while keeping the deployment local to a release branch until we release a full version and merge back into `master`.

ntn-x2: Similar to what we do now by requiring all changes to be on `master` before forking off for the releases. Those changes will now be to the template and not the deployment itself.

.github/workflows/genesis-ceremony.yml
bacv: > But I am not sure I understand your second sentence.

At the moment the proposed GH workflow creates a new branch e.g. `genesis-ceremony-devnet-0.1.3-rc.6` and commits newly generated `nodes/node/binary/src/config/deployment/devnet/deployment.yaml`. 

We could change this to the flow, where GH workflow doesn't create a new branch, but commits to the branch that the workflow was triggered from. In our case it would be `release/0.1.3`. I will make this change if you are not against it.

ntn-x2: I don't have any preference. Maybe I like more the idea of the action creating a PR against the release branch, so we can review and change anything before merging into the release branch. But then I think the PR could be opened automatically.

testnet/ceremony/genesis/devnet/deployment-template.yaml
bacv: > Other stuff should not be part of this struct DeploymentTemplate type, so we could commit the template to master while keeping the deployment local to a release branch until we release a full version and merge back into master.

Agree, will ping when it's cleaned up.

.github/workflows/genesis-ceremony.yml
bacv: That would just make the search replace query contain `${}`, kept it simple as I didn't anticipate these strings to collude with param names of our deployment config.

.github/workflows/genesis-ceremony.yml
bacv: Added options in https://github.com/logos-blockchain/logos-blockchain/pull/2732/commits/e323e7de97dd0605337ede12dcd87d96196a163a

.github/workflows/genesis-ceremony.yml
bacv: Updated to commit to the branch from which it was triggered. Creating PR complicates stuff, because we need thirdparty action, and token that has permission to create PR on our repo.

Committing is simpler, we don't create PRs our selves for the `release/0.1.3` branches, right?

Worst case, the commit could be removed manually from that branch by resetting HEAD, happy path would be automatic without the requirement to open a PR. Change in https://github.com/logos-blockchain/logos-blockchain/pull/2732/commits/0205a9c3d88b405e45e3326e5143693522991f88

testnet/ceremony/genesis/devnet/inscribe.yaml
bacv: Fixed in https://github.com/logos-blockchain/logos-blockchain/pull/2732/commits/ef9e2eabb9555c9119d63b681c95da4a62e14aa9

testnet/ceremony/genesis/devnet/deployment-template.yaml
bacv: Cleaned up in https://github.com/logos-blockchain/logos-blockchain/pull/2732/commits/942f4c6a81e65dfe71bc6191f45d9cf92d551a85

.github/workflows/genesis-ceremony.yml
bacv: > We could add a test value if we want to test the workflow before merging a PR

Not sure how to solve this, the version change should be part of a release. So the versions should not be changed by the genesis CI workflow.

I think our release steps will need to explain this:
- before any release we change versions;
- optional genesis workflow is performed after the version is bumped.

danielSanchezQ (approved):

.github/workflows/genesis-ceremony.yml
ntn-x2: Can we then remove this?

testnet/ceremony/genesis/devnet/deployment-template.yaml
ntn-x2: That's good, but it's not entirely what I meant. There's a few other things we don't need in the genesis template, but those should first be enforced in the Rust type. I.e., we don't need the whole `cryptarchia.genesis_block` in the template, since the genesis header itself depends on the transactions in the block, e.g., for the block root. Everything else is set to `0` anyway in the deployment file directly, so we could avoid adding it to the template. But perhaps this is a larger topic that is not part of this PR, but I think we should still discuss what goes in the template and what goes in the final deployment file directly.

.github/workflows/genesis-ceremony.yml
bacv: Then it will be up to the user to prepare template before release, updated: https://github.com/logos-blockchain/logos-blockchain/pull/2732/commits/5ff980c734bc4deb2a115970de0c022be377ce7f

testnet/ceremony/genesis/devnet/deployment-template.yaml
bacv: > That's good, but it's not entirely what I meant. There's a few other things we don't need in the genesis template, but those should first be enforced in the Rust type. I.e., we don't need the whole cryptarchia.genesis_block in the template

This is to be discussed, atm genesis generator is based on YAML, which I think is not what we want, because we have all these concrete types and we shouldn't rely on one specific serializer to do generation and overwrites. Since we already have genesis builder merged in master, this PR is mainly for CI part. 

I'll sync with @danielSanchezQ to see how we could remove template requirement from genesisi block generation, decouple everything from yaml serializer and generate it from a specific genesis config.

danielSanchezQ: > That's good, but it's not entirely what I meant. There's a few other things we don't need in the genesis template, but those should first be enforced in the Rust type. I.e., we don't need the whole `cryptarchia.genesis_block` in the template, since the genesis header itself depends on the transactions in the block, e.g., for the block root. Everything else is set to `0` anyway in the deployment file directly, so we could avoid adding it to the template. But perhaps this is a larger topic that is not part of this PR, but I think we should still discuss what goes in the template and what goes in the final deployment file directly.

We already discussed this and agree to go for explicit values as this is not gonna face users at all? Also its automated creation now so it should be easy to deal. 
Imo thats the best way to go. I see your point. But if values are not there then its difficult to follow up on what was there, a default, were was it filled, etc. That way we have a "single failing point"?

testnet/ceremony/genesis/devnet/deployment-template.yaml
```diff
@@ -15,9 +15,9 @@ blend:
     normalization_constant: 1.03
     activity_threshold_sensitivity: 1
 network:
-  kademlia_protocol_name: /logos-blockchain-ENV_PLACEHOLDER-VERSION_PLACEHOLDER/kad/1.0.0
-  identify_protocol_name: /logos-blockchain-ENV_PLACEHOLDER-VERSION_PLACEHOLDER/identify/1.0.0
-  chain_sync_protocol_name: /logos-blockchain-ENV_PLACEHOLDER-VERSION_PLACEHOLDER/chainsync/1.0.0
+  kademlia_protocol_name: /logos-blockchain-devnet-0.1.3-rc.5/kad/1.0.0
+  identify_protocol_name: /logos-blockchain-devnet-0.1.3-rc.5/identify/1.0.0
```
danielSanchezQ, line 19: Is the fixed version a regresion?

bacv: > Is the fixed version a regresion?

This will be part of the release steps for the release ~~wizard~~ genie to update.

testnet/ceremony/genesis/devnet/deployment-template.yaml
ntn-x2: Ok sorry I was not part of those discussions so I don't have full context. I jst thknk that having information that will be ignored in the template could be misleading, e.g. we would expect the final deployment to have a specific block root (for any reason) but it will have a different one. We will commit the template to the repo, so as a sanity check we might want to make sure the generated config has all the fields from the template. Having stuff in the template that won't end up in the final file could be confusing.
This said, this flow is already much better, so if you guys agree this is good enough to move forward, by all means I don't have anything against this.

danielSanchezQ: Actually I think I was mixing it with other convo (related). One thing that we can do maybe is to detach the genesis block in the same fashion we detached the keys from user_config.yaml files. 
That way we do not keep the block in the template at all, we just point to a different file maybe?
wdyt @ntn-x2 @bacv?

ntn-x2: That sounds like a better idea indeed.

testnet/ceremony/genesis/devnet/deployment-template.yaml
danielSanchezQ: I think it should be automated, like we were using before a selector from a tag ?

PR #2736, opened 2026-05-11
test(e2e): migrate zone sdk tests to cucumber

andrussal:

Migrates Zone SDK e2e tests from nextest to Cucumber.

This adds the migrated Zone Cucumber scenarios and removes the old nextest Zone e2e file.

Yes

@pradovic, @andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


tests/src/cucumber/steps/manual_zone/support.rs
```diff
@@ -0,0 +1,1154 @@
...
+) -> bool {
+    on_chain.contains(&payload.payload)
+        || pending
+            .iter()
+            .any(|pending_payload| pending_payload.payload == payload.payload)
+}
+
+fn preserves_sorted_order(payload: &InscriptionInfo, max_seen_on_chain: Option<&[u8]>) -> bool {
+    max_seen_on_chain.is_some_and(|seen| payload.payload.as_slice() >= seen)
+}
```
hansieodendaal: `preserves_sorted_order()` returns `false` when `max_seen_on_chain` is `None`.
That means if the sequencer gets an invalidation before any adoption has been recorded, the payload is immediately discarded rather than retried.

That seems inconsistent with the previous test’s intent. In the removed suite, the comment says the policy is effectively “smallest wins”, but this implementation becomes “nothing wins until something is already on chain.”

```suggestion
fn preserves_sorted_order(payload: &InscriptionInfo, max_seen_on_chain: Option<&[u8]>) -> bool {
    max_seen_on_chain.is_none_or(|seen| payload.payload.as_slice() >= seen)
}
```

hansieodendaal (commented):
This is looking good. One questionable test condition below, and two high-level comments.
- Could we please add the original removed test name to the new cucumber test as a comment, to assist reviewers, and later on we can remove it, for example
```
@zone_ci
# [tests/zone_sdk/e2e.rs] test_sequencer_publish_and_indexer_read
Scenario: Publish messages and read them from the zone indexer
  Given I have a zone cluster
  ...
```
- **As a note for a next PR:** 
Pretty soon we would need a ledger transaction to support an inscription transaction, to pay for the fees. I was wondering if it would make sense to rather add a sequencer and indexer pointer to a wallet, so that a normal wallet can do ledger and inscription type transactions. Later when the sequencer evolves, we either give it the ledger bit of the transaction or inputs and outputs it needs to construct the funded inscription transaction. A normal user or funding wallet will be able to, via the sequencer pointer, create a zone channel, add keys, publish, republish. It can then also via the indexer pointer verify inscriptions. This may fit in nicely with the node-wallet ecosystem we already have. The cucumber zone step language can stay largely the same, just a bit differently behind the scenes. What do you think?

tests/src/cucumber/steps/manual_zone/support.rs
```diff
@@ -0,0 +1,1339 @@
...
+
+    async fn already_discarded(&self, payload: &[u8]) -> bool {
+        self.discarded.lock().await.contains(payload)
+    }
+
+    fn preserves_order(&self, inscription: &InscriptionInfo) -> bool {
+        self.max_seen_on_chain
+            .as_deref()
+            .is_some_and(|seen| inscription.payload.as_slice() >= seen)
+    }
```
hansieodendaal, line 516: This still returns false if `max_seen_on_chain` is `None` - is that intentional?

Maybe a better intuitive return would be `Option<bool>`, to support `max_seen_on_chain` when `None`, which feels more like “ordering baseline not established yet” than a definite failure?

hansieodendaal (commented):
This is looking good - same comments as last time.
- Could we please add the original removed test name to the new cucumber test as a comment, to assist reviewers, and later on we can remove it, for example
```
@zone_ci
# [tests/zone_sdk/e2e.rs] test_sequencer_publish_and_indexer_read
Scenario: Publish messages and read them from the zone indexer
  Given I have a zone cluster
  ...
```
- **As a note for a next PR:** 
Pretty soon we would need a ledger transaction to support an inscription transaction, to pay for the fees. I was wondering if it would make sense to rather add a sequencer and indexer pointer to a wallet, so that a normal wallet can do ledger and inscription type transactions. Later when the sequencer evolves, we either give it the ledger bit of the transaction or inputs and outputs it needs to construct the funded inscription transaction. A normal user or funding wallet will be able to, via the sequencer pointer, create a zone channel, add keys, publish, republish. It can then also via the indexer pointer verify inscriptions. This may fit in nicely with the node-wallet ecosystem we already have. The cucumber zone step language can stay largely the same, just a bit differently behind the scenes. What do you think?

tests/src/cucumber/steps/manual_zone/support.rs
pradovic: `ChannelConfig` event can also cause tip to change and invalidate txs. In this case the `max_seen_on_chain` can still be `None` imo, so in this case we want to republish.

 I agree, I think we should republish if `max_seen_on_chain` is `None`, so in this case `preservers_order` should return true, imo. 

I lean towards `is_none_or(|seen| inscription.payload.as_slice() >= seen)`, as suggested earlier, unless I am missing something?

Returning Option also makes sense, but user policy needs to check it so probably user will mostly do `is_non_or(true)`

pradovic:
LGTM mod `preserve_order` comment.

Also, we increased the number of inscriptions in multi sequencer case to test it better. The time of execution did not change drastically. Could we preserve the same number of inscriptions as on `master`?

andrussal:
> LGTM mod `preserve_order` comment.
> 
> Also, we increased the number of inscriptions in multi sequencer case to test it better. The time of execution did not change drastically. Could we preserve the same number of inscriptions as on `master`?

Actually, I think this is already covered in `Concurrent identical payloads`: it publishes 10 messages per sequencer, so with 3 sequencers that gives 30 total.

pradovic:
> > LGTM mod `preserve_order` comment.
> > Also, we increased the number of inscriptions in multi sequencer case to test it better. The time of execution did not change drastically. Could we preserve the same number of inscriptions as on `master`?
> 
> Actually, I think this is already covered in `Concurrent identical payloads`: it publishes 10 messages per sequencer, so with 3 sequencers that gives 30 total.

I agree, but multi sequencer is better test to have, imho. We don't even have to support identical payloads, it is up to the user to supply unique payloads because of the possible mempool de-duplication. If we want to save test times, I would suggest to drop `Concurrent identical payloads` and just leave multi-sequencer tests with more inscriptions. But no worries, approved, I can also do it in a separate PR later. I wanted to refactor the tests a bit differently anyway. Thanks!

pradovic (approved):

hansieodendaal (approved):
LGTM! (Nice work)

PR #2742, opened 2026-05-12
fix: build.rs git rerun watchers

hansieodendaal:

Fix `build.rs` Git rerun watchers to resolve repository Git metadata from the workspace checkout instead of crate-local `.git` paths, reducing unnecessary rebuilds in workspace builds.

Validation:
Run consecutive `nextest` commands with fingerprint check, and verify that the 2nd run does not trigger a rebuild.
```
CARGO_LOG=cargo::core::compiler::fingerprint=info cargo nextest run --workspace --lib --bins --examples --tests --retries 2 --all-features --no-fail-fast --exclude logos-blockchain-tests
```

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


nodes/node/binary/build.rs
```diff
@@ -28,9 +34,63 @@ fn get_rustc_version() -> String {
...
+        for tag in tag_names.lines().map(str::trim).filter(|s| !s.is_empty()) {
+            emit_rerun_if_changed(&git_dir.join("refs").join("tags").join(tag));
+        }
+    }
+}
+
 fn main() {
-    println!("cargo:rerun-if-changed=.git/HEAD");
-    println!("cargo:rerun-if-changed=.git/refs");
+    if let Some(git_dir) = find_repo_git_dir() {
```
ntn-x2: We already assume git is present below, so I think we can always call `emit_git_watchers`.

nodes/node/binary/build.rs
```diff
@@ -28,9 +34,63 @@ fn get_rustc_version() -> String {
...
+                dir.join(gitdir)
+            });
+        }
+    }
+
+    None
+}
+
+fn emit_rerun_if_changed(path: &Path) {
+    println!("cargo:rerun-if-changed={}", path.display());
```
ntn-x2: I think we want `as_os_str()` here?

nodes/node/binary/build.rs
```diff
@@ -28,9 +34,63 @@ fn get_rustc_version() -> String {
...
+}
+
+fn emit_git_watchers(git_dir: &Path) {
```
ntn-x2, line 67: Do we need all this logic? Can't we just fix the existing `build.rs` script to point to the workspace root git folder instead? Or did you find that is not enough? The reason why we introduced this is to re-compute the version information that goes into the binary. Other than that, the default behaviour is to scan the folder to identify changes, so I would simply add a `cargo:rerun-if-changed=.`, and that would catch the default case, no?

ntn-x2 (commented):
I see why the current solution is wrong, but I think we don't need to be this drastic with the changes. I think we could fix the current state by pointing to the workspace root, but also we can add the default Cargo behaviour of monitoring the crate nested folders for any changes to then trigger the rebuild. What do you think?

nodes/node/binary/build.rs
hansieodendaal: I kept this conditional because “git binary exists” and “a repository `.git` dir are separate conditions.

nodes/node/binary/build.rs
hansieodendaal: I refactored this a bit to make it local-dev friendly and release-process resilient. I don’t think `cargo:rerun-if-changed=.` is right here, because it would rerun the build script for normal source changes that Cargo already tracks.

The build script only needs to rerun when the version metadata can change, so this now watches the relevant repo Git metadata directly: `HEAD`, the current branch ref, `refs/tags`, and `packed-refs`. This keeps tag changes covered without watching all of `.git/refs`.

nodes/node/binary/build.rs
hansieodendaal: `cargo:rerun-if-changed` is emitted as a textual build-script instruction, and `OsStr` does not implement `Display`, so `as_os_str()` would not be directly printable here. I used `path.display()` intentionally for the Cargo output line as I understand it is more idiomatic. Happy to switch to `to_string_lossy()` if you prefer that to make the lossy conversion explicit.

nodes/node/binary/build.rs
ntn-x2: True, but the whole `build.rs` file would fail if the binary exists but this is not a git folder since we use git commands. So at the very least we can gate-keep the whole logic? So if this is not a git repo we can return `None`s, or then panic as we would do now?

nodes/node/binary/build.rs
ntn-x2: I did not look into the details of how `display()` works, but if you say that it prints the right format cargo expect in all cases, on all platforms, then we can leave `display()`.

ntn-x2 (approved):

PR #2743, opened 2026-05-13
feat(ledger): DRAFT: eliminate session and use epoch for SDP/SRP

youngjoon-lee:
## 1. What does this PR implement?

This replaces https://github.com/logos-blockchain/logos-blockchain/pull/2692. This will be merged to the `session-removal` branch, not `master`.

This PR eliminates the concept of session from the `lb-ledger` crate which includes SDP/SRP.

Specs:
- [[RFC] Remove Concept of a Session](https://www.notion.so/nomos-tech/RFC-Remove-Concept-of-a-Session-f39261aa09df826db83e81613bb454dc?source=copy_link)
  - [[1.0.0] Service Declaration Protocol](https://www.notion.so/nomos-tech/1-0-0-Service-Declaration-Protocol-33d261aa09df808f834bc728ae0c5e0f?source=copy_link)
  - [[1.2.1] Service Reward Distribution Protocol](https://www.notion.so/nomos-tech/1-2-1-Service-Reward-Distribution-Protocol-4e5261aa09df831cab35816cc0f1f7f1?source=copy_link)
  - [[1.0.0] Blend Protocol - Rewarding](https://www.notion.so/nomos-tech/1-0-0-Blend-Protocol-12c261aa09df82998a810130b32f94ae?source=copy_link#adc261aa09df82f89fe781e55ef959a6)

I initially tried to split this PR into each of SDP and SRP, but it wasn't easy. I think it's fine to put them into one PR because diffs are not so difficult to read.

This PR doesn't implement the time-based ledger update yet, which is being discussed in https://discord.com/channels/973324189794697286/1503638327083925544. We still update ledger only when a new block arrives (not immediately after the new epoch starts). Actually, the time-based approach is the whole point of the work (removing session), it will be a big change in the current architecture. We need to discuss more.

For this reason, this PR doesn't implement handling epoch jumps in SDP/SRP, which couldn't happen in the session-based model. We need to either switch to the time-based ledger update or define how to handle epoch jumps.

In a next PR, I will update chain-service. CI fails currently, but I checked that all tests in `lb-ledger` pass.

## 2. Does the code have enough context to be clearly understood?

yes

## 3. Who are the specification authors and who is accountable for this PR?

@youngjoon-lee 
spec: @madxor 

## 4. Is the specification accurate and complete?

yes

## 5. Does the implementation introduce changes in the specification?

no

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

[...]

PR #2744, opened 2026-05-13
feat(chain-service): broadcast active providers when epoch of LIB changes

youngjoon-lee:

On top of PR #2743 

Previously, chain service was broadcasting new active providers when all of the following conditions are true:
- A new block is added to the honest chain, which means that LIB is changed.
- The session number of the new LIB's SDP active snapshot is greater than the one of the previous LIB's.

Now that we're removing the concept of session, this PR changes the 2nd condition as below:
- The **epoch** of the new LIB's SDP active snapshot is greater than the one of the previous LIB's.  

This PR changes chain-service, broadcast-service, and membership adapter of the blend service. But it doesn't change the blend service internals because I want to get a confirmation whether this approach aligns well with the purpose of removing the concept of session.

yes

@youngjoon-lee 
spec: @madxor 

yes

no

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/services/blend/src/instance.rs
+++ b/services/blend/src/instance.rs
@@ -627,7 +627,7 @@ mod tests {
             let instance = instance
                 .handle_session_event(
                     // With an empty membership smaller than the minimal size.
-                    SessionEvent::NewSession(MembershipInfo::from_membership_and_session_number(
+                    SessionEvent::NewSession(MembershipInfo::from_membership_and_epoch(
                         membership(&[], local_node),
                         1,
                     )),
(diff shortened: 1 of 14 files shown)


services/blend/src/instance.rs
```diff
@@ -627,7 +627,7 @@ mod tests {
             let instance = instance
                 .handle_session_event(
                     // With an empty membership smaller than the minimal size.
-                    SessionEvent::NewSession(MembershipInfo::from_membership_and_session_number(
+                    SessionEvent::NewSession(MembershipInfo::from_membership_and_epoch(
```
ntn-x2, line 630: `EpochEvent`?

ntn-x2 (commented):
This is the bit I am not sure about. If we do all the effort of replacing sessions with epochs, but then maintain the current behaviour of yielding a new update only when the LIB changes, we are back at square 1 of either letting Blend nodes temporarily diverge because one will race ahead of the others or then not being able to self-apply a block for the same reason.

PR #2745, opened 2026-05-13
feat(ledger): create/store events resulted by block execution

youngjoon-lee:

In [Discord](https://discord.com/channels/973324189794697286/1503299047480496219), we discussed to introduce the block event model, so that users can see what happened during block executions. 
Subscribing to `Block`s is not enough in some cases. For example, `Deposit` operation contains `Inputs` which has only `NoteId`s, not `Value`s. So, users cannot know how much fund has been deposited to the channel, only by looking into transactions in the `Block`. 

Now, users can query the events resulted by a certain block. The `Deposit` event contains `amount` which is the sum of all `Inputs` specified in the `Deposit` operation executed.
```
GET /cryptarchia/blocks/:id/events
```
```json
[
  {
    "Tx": {
      "tx_hash": [1, ...],
      "op_id": [2, ...],
      "payload": {
        "Deposit": {
          "channel_id": "abababababababababababababababababababababababababababababababab",
          "amount": 16,
          "metadata": [104, ...]
        }
      }
    }
  },
  {
    "Ledger": {
      -- to be implemented
      "SdpReward": {
        "amount": 100,
        "pk": "..."
      }
    }
  }
]
```

Using this model, we can emit any event resulted by operation, transaction, or even block.
For example, SDP rewards are distributed automatically when a certain block is added. These rewards are not associated with any transaction. This event model covers this case.
In a next PR, I will add more events (for withdraw, leader claim, sdp rewards, ...). It will be also useful for wallet.

P.S.
You can review each commit separately if it's easier.

yes

@youngjoon-lee 

i think we need to write a spec for Event.

n/a

## Checklist

> [!WARNING]  

* [x] 2. Description added.


core/src/events/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+
+/// An event generated by the execution of block/transaction/operation
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Event {
+    /// `None` if the event is not associated with any transaction
+    pub tx_hash: Option<TxHash>,
+    /// `None` if the event is not associated with any operation
+    pub op_id: Option<Hash>,
+    pub payload: EventPayload,
+}
```
danielSanchezQ, line 54: Is ti possible to have both at the same time? Otherwise I think it is better to have an enumeration?:

```rust
pub enum Event {
    Tx {tx_hash, op_id, payload}
    Payload(EventPayload)
}
```

core/src/mantle/ops/channel/deposit.rs
```diff
@@ -18,6 +20,12 @@ pub struct DepositOp {
     pub metadata: Vec<u8>,
 }
 
+impl OpId for DepositOp {
+    fn op_bytes(&self) -> Vec<u8> {
+        encode_channel_deposit(self)
```
danielSanchezQ, line 25: Is it missing the hash?

core/src/mantle/ops/channel/deposit.rs
```diff
@@ -82,6 +91,16 @@ impl Operation<DepositValidationContext<'_>> for DepositOp {
...
-        Ok(ctx)
+        let events = Events::from_iter([Event::new(
+            ctx.tx_hash,
+            self.op_id(),
+            EventPayload::Deposit {
+                channel_id: self.channel_id,
+                amount: amount_deposited,
+                metadata: self.metadata.clone(),
+            },
+        )]);
```
danielSanchezQ: Better std:iter::once than allocating an array.

core/src/mantle/ops/channel/deposit.rs
youngjoon-lee: I followed other implementations: https://github.com/logos-blockchain/logos-blockchain/blob/78a6041e8c6d1a933af577fa8307b7e3e23d452b/core/src/mantle/ops/channel/withdraw.rs#L24

When we call `op_id()`, it calls `op_bytes()` internally and concatenate it with the prefix.
https://github.com/logos-blockchain/logos-blockchain/blob/78a6041e8c6d1a933af577fa8307b7e3e23d452b/core/src/mantle/ops/mod.rs#L45-L53

danielSanchezQ: Ah true true. Thanks. I didn't realize it was op_bytes not op_id method itself.

core/src/mantle/ops/channel/deposit.rs
youngjoon-lee: Yes. Done 92b753d4082ab17143afacd98e4fdf216136d870

core/src/events/mod.rs
youngjoon-lee: Yes. Done ba61cd28fd8d2c0b8c7d9ecd619d6888388ac696. Enum is better. I updated the PR description with the new JSON example.

danielSanchezQ (approved):

pradovic (approved):

PR #2746, opened 2026-05-13
feat(mempool): Mempool txtracker

danielSanchezQ:

* Adds transaction tracking across forks to mempool. It is intended to be a memory efficient implementation in where we only have a single Tx shared across different forks. Also txs are loaded and shared the same way during recovery.
* Txs can be selected across different forks and API uses the ancestor hint to query for the txs ready for the specified block.

It's a really big PR which went a bit out of hand, but most of the things are cointained. I'll prepare a small guide for reviwers so it can be follow easy.

@danielSanchezQ @davidrusu 

There is no specification for this yet. But we will prepare a proper document once proved that the design is correct.

N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.


services/tx-service/src/backend/tracker.rs
```diff
@@ -0,0 +1,298 @@
...
+pub struct TxTrackerState<Tx, TxId>
+where
+    TxId: Eq + Hash,
+{
+    ready_txs: HashTrieMap<TxId, Tx>,
+    orphan_txs: HashTrieMap<TxId, Tx>,
+    dep_to_tx: HashTrieMap<DependencyId, HashTrieSet<TxId>>,
+    tx_pending_count: HashTrieMap<TxId, usize>,
+    // TODO: Replace later for the current ledger state
+    processed_deps: HashSet<DependencyId>,
```
danielSanchezQ: This is mock. @davidrusu this should be a set of non-consumed deps instead (?)

core/src/mantle/ops/mod.rs
```diff
@@ -174,4 +175,30 @@ impl Op {
...
+            }
+            _ => Box::new(std::iter::empty()),
+        }
+    }
+
+    pub fn produces(&self) -> impl Iterator<Item = DependencyId> {
+        match self {
+            Self::ChannelInscribe(op) => Box::new(std::iter::once(DependencyId::copy_from_slice(
+                Hasher::digest(&op.inscription).as_slice(),
+            ))) as Box<dyn Iterator<Item = DependencyId>>,
```
davidrusu: This needs to be the inscription message id. It needs to match the `parent` exactly.

```suggestion
            Self::ChannelInscribe(op) => Box::new(std::iter::once(DependencyId::copy_from_slice(op.id()))) as Box<dyn Iterator<Item = DependencyId>>,
```

core/src/mantle/tx.rs
```diff
@@ -297,6 +285,16 @@ impl Transaction for MantleTx {
...
 
+impl TransactionDependencies for MantleTx {
+    fn consumes(&self) -> impl Iterator<Item = DependencyId> {
+        self.ops().iter().flat_map(Op::consumes)
+    }
+
+    fn produces(&self) -> impl Iterator<Item = DependencyId> {
+        self.ops().iter().flat_map(Op::produces)
+    }
+}
```
davidrusu: We need to remove the self-dependencies here.

In fact, I think you need to process this one op at a time because order of operations is important.

```suggestion
impl TransactionDependencies for MantleTx {
    fn consumes(&self) -> impl Iterator<Item = DependencyId> {
        let mut external_produces = HashSet<DependencyId>::new();
        let mut external_consumes = HashSet<DependencyId>::new();
        
        for op in self.ops().iter() {
            for consumed in op.consumes() {
                if external_produces.contains(consumed) {
                    // this is an internally resolved dependency
                    external_produces.remove(consumed);
                } else {
                    // this is an external dependecy
                    external_consumes.add(consumed)
                }
            }
            external_produces.extend(op.produces();
        }
        
        external_consumes.into_iter()
    }

    fn produces(&self) -> impl Iterator<Item = DependencyId> {
        // <something very similar to consume
    }
}
```

core/src/mantle/genesis_tx.rs
```diff
@@ -163,6 +165,16 @@ impl GasCalculator for GenesisTx {
...
 
+impl TransactionDependencies for GenesisTx {
+    fn consumes(&self) -> impl Iterator<Item = DependencyId> {
+        self.tx.mantle_tx.0.iter().flat_map(Op::consumes)
+    }
+
+    fn produces(&self) -> impl Iterator<Item = DependencyId> {
+        self.tx.mantle_tx.0.iter().flat_map(Op::produces)
+    }
+}
```
davidrusu: the `mantle_tx` implements `TransactionDependencies`, can we delegate to that implementation here?

services/tx-service/src/backend/tracker.rs
davidrusu: It should be a ledger state actually.

this is making me doubt the `DependencyId` abstraction since DependencyId is opaque so, we won't know where to search in the ledger to see if that dependency is satisfied..

What if we have a `Consumes` struct instead that is returned from `op.consumes()-> Consumes`?

```rust
struct Consumes {
    channel: HashMap<ChannelId, MsgId>,
    notes: HashSet<NoteId>,
}
```

Then, given a `LedgerState` and a `Consumes` instance, we can trivially loop up in the channel state in the ledger for the `consumes.channel` and in the utxos set for `consumes.notes`.

services/tx-service/src/backend/tracker.rs
```diff
@@ -0,0 +1,298 @@
...
+                    self.dep_to_tx.insert_mut(dep, set);
+                }
+            }
+            self.tx_pending_count
+                .insert_mut(tx.hash(), pending_deps_count);
+            self.orphan_txs.insert_mut(tx.hash(), tx);
+        }
+    }
+
+    pub fn tx_in_block(&mut self, tx_id: &Tx::Hash) {
```
youngjoon-lee, line 160: Will this function be called when a block, which includes `tx_id`, is added to the chain (regardless of whether on the honest chain or on any fork)?

services/tx-service/src/backend/forks.rs
```diff
@@ -0,0 +1,477 @@
...
+        // Check current_tips first, then states: a fork sibling may have already
+        // moved the shared parent out of current_tips into states.
+        let parent_state = self
+            .current_tips
+            .get(&parent)
+            .or_else(|| self.states.get(&parent))
+            .ok_or(ForksTrackerError::ParentNotFound(parent))?;
+        let mut block_state = parent_state.clone();
+        for tx in transactions {
+            block_state.process_tx(tx);
```
youngjoon-lee: Hm, I don't understand when `process_tx` should be used and when `tx_in_block` should be used.

In this function, we're passing all processed txs to the `TxTrackerState` via `process_tx`. But, it just inserts those txs into `ready_txs` or `orphan_txs`. Shouldn't we call `tx_in_block` here?

services/tx-service/src/backend/tracker.rs
danielSanchezQ: Yes, because we are keeping a frontier of all tips here (basically tracking all forks). Or that is the intention at least.

services/tx-service/src/backend/forks.rs
danielSanchezQ: Actualy this is a bug, we should call `tx_in_block` indeed!

core/src/mantle/ops/mod.rs
danielSanchezQ: Ah yes, nice catch thanks!

core/src/mantle/genesis_tx.rs
danielSanchezQ: Yes! thanks!

core/src/mantle/tx.rs
danielSanchezQ: Should the order be part of tx validation itself? Like checking that the tx was properly built?

danielSanchezQ: Ok, I've been checking. This actually does not solve anything imo. By the time the tx is check either is properly build or not. 
If its properly built (order of ops are ok) then just filtering out the difference in produces/consumes from ops is enough (we take out non internal resolves, the tx is properly built so by the time is able to be selected it will go through). 
If the tx is not properly built (order of ops are wrong) it is still expected that inner deps produces/consumes will be process inside the tx. So, by the time it reaches the leader, even if it was a candidate it will be thrown out because it was not properly built.

So imo we have two options
1. we validate construction with consideration of what operations are produced consumed inside and effectively evict wrong ones from the mempool directly. 
2. Or we just filter broadly inner tx dependencies and let the leader discard them.

I think 2nd one is easier right now. 1st one is a bit better but implies changing/expanding the spec maybe?
cc @thomaslavaur

davidrusu: sure, this is a bit looser but should be fine

core/src/mantle/mod.rs
```diff
@@ -191,3 +143,27 @@ pub enum Error {
...
+    #[error(transparent)]
+    TransferError(#[from] TransferError),
+}
+pub trait TxRewardsRatio: AuthenticatedMantleTx {
+    fn rewards_ratio<Constants: GasConstants>(
+        &self,
+        gas_price: &GasPrices,
+        utxos: &Utxos,
+    ) -> Result<Value, TxRewardsRatioError>;
+}
```
danielSanchezQ: This are 2 traits added to transsacitions that are now needed for the mempool.

services/tx-service/src/backend/adapter.rs
danielSanchezQ, line 1: Mempool now needs more ways to check different things, all of them are grouped into a single adapter:
* Communicates with storage
* Queries ledger state
* Queries blocks

services/tx-service/src/backend/tracker.rs
danielSanchezQ, line 1: The `TxTrackerState` holds the txs state per block basis.

services/tx-service/src/backend/inspector.rs
danielSanchezQ, line 1: Convenient extractor of ledger state wanted data in the mempool

services/tx-service/src/backend/history.rs
danielSanchezQ, line 1: This holds the txs history as they arrive to the mempool. It is used to replay txs to parents block states to keep consistent in reorgs situations.

services/tx-service/src/backend/forks.rs
danielSanchezQ, line 1: Wrapper around states and history which basically process events and move information around.

services/tx-service/src/backend/pool.rs
```diff
@@ -1,252 +1,220 @@
...
+        let fetch_adapter = adapter.clone();
+        // not beautiful but we need to collect the txs from storage and the interface
+        // is async
+        let txs: HashMap<Tx::Hash, Arc<Tx>> =
+            tokio::runtime::Handle::current().block_on(async move {
+                let stream = fetch_adapter
+                    .get_txs(&recover_txs)
+                    .await
+                    .unwrap_or_else(|e| panic!("Could not recover txs from storage: {e:?}"));
+                stream.map(|tx| (tx.hash(), Arc::new(tx))).collect().await
```
danielSanchezQ, line 233: Recovery works in the following:
* We dump a map of the kept historical txs (txs seen in the current pending states)
* All other structs are save with pointers to those txs.
* Txs are loaded and wrapped in `Arc` 
* We "de-reference" the pointers (txids really) and load the `Arc<Tx>` in the proper rpds structures to minimize memory usage.

danielSanchezQ (commented):
The PR is quite big, but there is a lot of types changes that bubbled up through the layer. Do not check those too much. I;m adding comments to the main files that should be properly check. 

Also as a late point. There is 2 pending changes that will be tackled separetely later which are to optimize `DependencyIds` to use internal `Bytes` reprensentation to avoid copying bytes constantly.

services/tx-service/src/backend/forks.rs
```diff
@@ -0,0 +1,981 @@
...
+        }
+        for tx in txs {
+            if self.mempool_log.forget_tx(tx) {
+                removed += 1;
+            }
+        }
+        removed
+    }
+
+    pub fn get_txs(&self) -> HashMap<Tx::Hash, Tx> {
```
bacv: This is only used in `services/tx-service/src/backend/pool.rs` in `get_items_by_keys` method, where we convert this back to iter to filter, then collect again to make another stream. Would it be possible to return here an iter, and avoid another collect where filtering happen to create a stream?

services/tx-service/src/tx/service.rs
```diff
@@ -443,30 +520,36 @@ where
 
     async fn partition_transactions_by_availability(
         pool: &Pool,
-        hashes: Vec<Pool::Key>,
-    ) -> Result<TransactionsByHashesResponse<Pool::Item, Pool::Key>, MempoolError> {
-        let keys_set: BTreeSet<Pool::Key> = hashes.into_iter().collect();
-
+        hashes: Vec<Pool::TxHash>,
+    ) -> Result<TransactionsByHashesResponse<Pool::Tx, Pool::TxHash>, MempoolError> {
```
bacv, line 529: If this returned only "found_hashes", then the full `TransactionsByHashesResponse` could be constructed by the caller (line 416), this way we would collect `items_stream` only once, this method would return just once?

services/tx-service/src/backend/forks.rs
danielSanchezQ: Yeah, sounds good, thanks!

services/tx-service/src/tx/service.rs
danielSanchezQ: Good call, but I will raise an issue to be fixed outside this changes.

hansieodendaal:
I have been doing system-level tests on this branch, just noting some of my findings.

**1, Node does not restart cleanly (but it does on master)**
It seems that we have an interlock due to the panic below (`error sending request for url (http://127.0.0.1:20264/cryptarchia/info)`).
```rust
2026-05-29T14:10:22.575266Z DEBUG chain::service: 14/14 blocks applied during initialization
2026-05-29T14:10:22.575270Z  INFO chain::service: 14 blocks recovered. finishing initialization tip_height=14 lib_height=0
2026-05-29T14:10:22.575282Z  INFO chain::service: Initializing chain online notifier with Bootstrapping
2026-05-29T14:10:22.575304Z  INFO logos_blockchain_chain_service: Service 'Cryptarchia' is ready.
2026-05-29T14:10:22.575396Z DEBUG logos_blockchain_sdp_service: Fetching declaration state for DeclarationId([131, 52, 24, 129, 44, 135, 116, 22, 228, 253, 151, 247, 66, 170, 17, 161, 115, 117, 31, 205, 54, 69, 207, 170, 188, 38, 225, 237, 75, 210, 201, 121]) from ledger tip HeaderId(24a442795fb4a71e9f9679019e5101e8db689fe91512cb1bd2e7015155f84275)
2026-05-29T14:10:22.575444Z ERROR logos_blockchain_node::panic: A panic occurred panic.payload="Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks." panic.location="services/tx-service/src/backend/pool.rs:206:47" panic.backtrace="disabled backtrace" panic.note="run with RUST_BACKTRACE=1 environment variable to display a backtrace"
```

core/src/mantle/ops/mod.rs
```diff
@@ -220,6 +228,38 @@ impl Op {
         }
     }
 
+    pub fn consumes(&self) -> impl Iterator<Item = DependencyId> {
+        match self {
+            // The root MsgId is the "no prior message" sentinel for the first
```
hansieodendaal: I see the zone SDK assumes that, within a block, channel tip-bearing ops already appear in valid chain order. But `ChannelConfigOp` has no explicit parent lineage and directly resets `channel.tip_message = config.id()`. What enforces that a block producer can’t order `ChannelConfig` before inscriptions that were authored against the old tip?

Related concern: if a sequencer replays the same config op later, `config.id()` is the same, so this appears to reset `channel.tip_message` back to the old config id and invalidate anything authored against the newer tip lineage.

**Example:**
- client post order where `i4` takes new parent lineage into account: `i1`, `i2`, `i3`, `config`, `i4`, `i5`, `i6`
- block proposal/mempool selection order: `i1`, `i2`, `i3`, `config`, => `i4` -> `i6` will fail
- block proposal/mempool selection order: `config` => `i1` to `i6` will fail

Should `ChannelConfigOp` also be rooted into `DependencyId`? It seems like config is also a tip-bearing op, but unlike `ChannelInscribe`, it has no explicit parent lineage and doesn’t appear to participate in the dependency graph. If config consumed the previous channel tip and produced `config.id()`, that would make the ordering constraint explicit and avoid this special-case behavior.

If that guarantee exists elsewhere, could we document or test it here?

danielSanchezQ: Yes, if that the case it should added as dependency. Can you post a link to the spec so I can asses the changes tomorrow? Thanks

hansieodendaal: You mean this? [CHANNEL_INSCRIBE](https://app.notion.com/p/nomos-tech/1-5-0-Mantle-358261aa09df802faa45d49011bc134a?source=copy_link#358261aa09df8103836eef3c80f4c8fa)

core/src/mantle/ops/mod.rs
```diff
@@ -220,6 +228,38 @@ impl Op {
...
+    }
+
+    pub fn produces(&self) -> impl Iterator<Item = DependencyId> {
+        match self {
+            Self::ChannelInscribe(op) => Box::new(std::iter::once(DependencyId::copy_from_slice(
+                op.id().as_ref(),
+            ))) as Box<dyn Iterator<Item = DependencyId>>,
+            Self::Transfer(op) => Box::new(
+                op.outputs
+                    .utxos(op)
```
davidrusu, line 261: nit: this `op.outputs.utxos(op)` API is ugly, should just be `op.utxos()`

core/src/mantle/tx.rs
```diff
@@ -256,15 +258,11 @@ impl MantleTx {
...
 
-    #[must_use]
-    pub fn transfers(&self) -> Vec<TransferOp> {
-        let mut transfers: Vec<TransferOp> = vec![];
-        for op in self.ops() {
-            if let Op::Transfer(transfer_op) = op {
-                transfers.push(transfer_op.clone());
-            }
-        }
-        transfers
```
davidrusu, line 267: no need to do these micro refactors in this PR, could have just logged an issue to move to iterators and continue using the vec api

it just blows up the PR size and obscures the main thrust of the PR

core/src/mantle/tx.rs
```diff
@@ -637,6 +633,34 @@ impl GasCalculator for SignedMantleTx {
...
+            GasCalculator::execution_gas_consumption::<Constants>(self, gas_prices)?,
+            gas_prices.execution_base_gas_price,
+        )?
+        .checked_add(GasCalculator::storage_gas_cost(self, gas_prices)?)?;
+
+        Ok(GasCost::from(balance? as Value)
+            .checked_sub(tx_fee_burned)?
+            .into_inner())
+    }
+}
```
davidrusu: Why is this called a Ratio? ratio implies a division of some sort e.g. X/Y

Lets call the trait `TxPriorityTip`

core/src/mantle/tx.rs
```diff
@@ -1074,4 +1098,185 @@ mod tests {
             })
         );
     }
+
+    // ── TransactionDependencies tests ────────────────────────────────────────
+    //
+    // These tests verify that MantleTx correctly filters out dependencies that
+    // are both produced and consumed within the same transaction (internal
+    // deps), exposing only the truly external consume/produce surface.
+    mod transaction_dependencies {
```
davidrusu, line 1111: missing a `#[cfg(test)]` here

services/chain/chain-network/src/lib.rs
```diff
@@ -104,85 +105,106 @@ pub struct ChainNetwork<
     NetAdapter,
     Mempool,
     MempoolNetAdapter,
+    MempoolAdapter,
     TimeBackend,
     RuntimeServiceId,
 > where
-    Cryptarchia: CryptarchiaServiceData<Tx = Mempool::Item>,
+    Cryptarchia: CryptarchiaServiceData<Tx = Mempool::Tx>,
```
davidrusu, line 109: tbh, we did not need to rename this `::Item`and `::Key` type params in this PR, I'd bet ~100LOC in this diff are from that alone.

We need to get more disciplined about focusing our changes to make it easier for reviewers, when you notice something like this we can open an issue and tackle it later

tests/testing_framework/src/diagnostics.rs
```diff
@@ -518,14 +515,12 @@ impl From<BlendNetworkInfo<lb_network_service::backends::libp2p::PeerId>> for Bl
 
 #[derive(Clone)]
 struct MempoolSnapshot {
-    pending_items: usize,
     last_item_timestamp: u64,
 }
```
davidrusu, line 516: Is there a strong reason why we are touching this struct? Can't we populate the pending_items struct here?

services/tx-service/src/metrics.rs
```diff
@@ -6,6 +6,10 @@ pub fn mempool_transactions_removed(removed_count: usize) {
     lb_tracing::increase_counter_u64!(mempool_transactions_removed_total, removed_count as u64);
 }
 
+#[expect(
+    dead_code,
+    reason = "Will be plugged when we decide what to query pending transactions from"
+)]
 pub fn mempool_transactions_pending(pending_count: usize) {
     lb_tracing::metric_gauge_u64!(mempool_transactions_pending, pending_count as u64);
 }
```
davidrusu: lets fix this before merging

services/tx-service/src/lib.rs
```diff
@@ -108,6 +107,5 @@ where
 #[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
 #[derive(serde::Serialize, serde::Deserialize)]
 pub struct MempoolMetrics {
-    pub pending_items: usize,
     pub last_item_timestamp: u64,
 }
```
davidrusu, line 113: to resolve before we merge

services/tx-service/src/tx/service.rs
```diff
@@ -531,14 +613,6 @@ where
...
         }
 
-        tracing::trace!(
-            target: LOG_TARGET,
-            {
-                counter.tx_mempool_pending_items = pool.pending_item_count(),
-            },
-            "mempool pending items updated"
-        );
-
```
davidrusu, line 541: this was a useful log, no? can we add it back?

services/tx-service/src/backend/pool.rs
```diff
@@ -258,16 +226,16 @@ fn current_timestamp_millis() -> u64 {
...
+            "Removed {removed_count} items from mempool"
         );
     } else {
         tracing::debug!(
             target: LOG_TARGET,
-            "Removed {removed_count} items from mempool; pending_items={pending_items}"
+            "Removed {removed_count} items from mempool"
         );
     }
 }
```
davidrusu, line 273: TODO: revert this before merging

services/tx-service/src/backend/inspector.rs
```diff
@@ -0,0 +1,25 @@
...
+            .utxos()
+            .keys()
+            .map(|note_id: &NoteId| DependencyId::copy_from_slice(&note_id.as_bytes()));
+        let inscriptions_tips = state
+            .mantle_ledger()
+            .inscriptions_tips()
+            .map(|id: MsgId| DependencyId::copy_from_slice(id.as_ref()));
+        utxos.chain(inscriptions_tips).collect()
+    }
+}
```
davidrusu: This is very inefficient, we discussed this already and I thought we said we were going to have smarter dependency design rather than a full duplicate of every entity in the ledger.

i.e. replace tx dependencies from `Set<DependencyId>` with `TxDependency { channel_tips, utxos }`.

I don't see how this approach can scale

services/tx-service/src/tx/service.rs
danielSanchezQ: Yes, but what is pending now? avg per branch? we enforce a query to get the current tip and we return the current tip pending ones?

core/src/mantle/ops/mod.rs
danielSanchezQ: Ill create an issue for this thanks.

core/src/mantle/tx.rs
danielSanchezQ: True, but at some point it was annoying (which i dont remember the reason for now) and just went ahead.

core/src/mantle/tx.rs
danielSanchezQ: Leftover, had a convo with @thomaslavaur about making this a bit smarter and sketched some of it. Then resolved into going the original idea and missed the name.

core/src/mantle/tx.rs
danielSanchezQ: good catch, thanks!

services/chain/chain-network/src/lib.rs
danielSanchezQ: I think we had to. `Item/Key` was a legacy leftover in the mempool and imo we have to be explicit with what we do. We had to go through generics hell one more time and just decided to do it. I see your pov too, but I dont think is that harmful in this case.

services/tx-service/src/backend/inspector.rs
danielSanchezQ: > I thought we said we were going to have smarter dependency design rather than a full duplicate of every entity in the ledger.

Yes, my thoughts were to implement this and iterate if necessary later. But I think you are right and even with just pointers is memory inneficient. 
The other option is to traverse all txs everytime, but for 30s blocks it may be much better. I'll do it shortly, thanks!

services/tx-service/src/backend/forks.rs
danielSanchezQ: sadly because of trait bounds on the mempool trait collects are necessary for now, so not changing it atm.

core/src/mantle/tx.rs
danielSanchezQ: actually this mod was inside a cfg flagged mod already

danielSanchezQ:
> I have been doing system-level tests on this branch, just noting some of my findings.
> 
> **1, Node does not restart cleanly (but it does on master)** It seems that we have an interlock due to the panic below (`error sending request for url (http://127.0.0.1:20264/cryptarchia/info)`).
> 
> ```rust
> 2026-05-29T14:10:22.575266Z DEBUG chain::service: 14/14 blocks applied during initialization
> 2026-05-29T14:10:22.575270Z  INFO chain::service: 14 blocks recovered. finishing initialization tip_height=14 lib_height=0
> 2026-05-29T14:10:22.575282Z  INFO chain::service: Initializing chain online notifier with Bootstrapping
> 2026-05-29T14:10:22.575304Z  INFO logos_blockchain_chain_service: Service 'Cryptarchia' is ready.
> 2026-05-29T14:10:22.575396Z DEBUG logos_blockchain_sdp_service: Fetching declaration state for DeclarationId([131, 52, 24, 129, 44, 135, 116, 22, 228, 253, 151, 247, 66, 170, 17, 161, 115, 117, 31, 205, 54, 69, 207, 170, 188, 38, 225, 237, 75, 210, 201, 121]) from ledger tip HeaderId(24a442795fb4a71e9f9679019e5101e8db689fe91512cb1bd2e7015155f84275)
> 2026-05-29T14:10:22.575444Z ERROR logos_blockchain_node::panic: A panic occurred panic.payload="Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks." panic.location="services/tx-service/src/backend/pool.rs:206:47" panic.backtrace="disabled backtrace" panic.note="run with RUST_BACKTRACE=1 environment variable to display a backtrace"
> ```

I opened an [issue](https://github.com/logos-blockchain/logos-blockchain/issues/2853) for this. I'll do the changes in a different PR, I have to check what solution is better and simpler.

core/src/mantle/tx.rs
davidrusu: yes, I understand the temptation, but it just makes reviews take longer and obscures the main points of your change

core/src/mantle/tx.rs
```diff
@@ -637,6 +637,34 @@ impl GasCalculator for SignedMantleTx {
     }
 }
 
+impl TxPriorityTip for SignedMantleTx {
+    fn tip<Constants: GasConstants>(
+        &self,
+        gas_prices: &GasPrices,
+        utxos: &Utxos,
+    ) -> Result<Value, TxRewardsRatioError> {
```
davidrusu: can we call this function `priority_tip`? currently it collides with the chain `tip` name

core/src/mantle/mod.rs
```diff
@@ -191,3 +149,57 @@ pub enum Error {
...
+    }
+}
+
+pub trait TxDependencies: Transaction {
+    fn consumes(&self) -> TxDependency;
+    fn produces(&self) -> TxDependency;
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum TxRewardsRatioError {
```
davidrusu:
```suggestion
pub enum TxPriorityTipError {
```

core/src/header/mod.rs
```diff
@@ -25,6 +25,13 @@ impl Debug for HeaderId {
     }
 }
 
+impl HeaderId {
+    #[must_use]
+    pub const fn genesis() -> Self {
+        Self([0; 32])
+    }
+}
```
davidrusu, line 33: This [0;32] header is *not* the genesis block header id. It is the *parent* of the genesis block. We need to change this name to make that clear.

Why do we need this? do we use it elsewhere?

core/src/mantle/tx.rs
```diff
@@ -637,6 +637,34 @@ impl GasCalculator for SignedMantleTx {
...
+        let tx_fee_burned = GasCost::calculate(
+            GasCalculator::execution_gas_consumption::<Constants>(self, gas_prices)?,
+            gas_prices.execution_base_gas_price,
+        )?
+        .checked_add(GasCalculator::storage_gas_cost(self, gas_prices)?)?;
+
+        Ok(GasCost::from(balance? as Value)
+            .checked_sub(tx_fee_burned)?
+            .into_inner())
+    }
```
davidrusu, line 665: I worry a bit about the duplication of this code with the ledger code path. Can we log a tech debt issue to deal with the duplication?

ledger/src/mantle/mod.rs
```diff
@@ -123,6 +125,13 @@ impl LedgerState {
         Self { channels, ..self }
     }
 
+    pub fn inscriptions_tips(&self) -> impl Iterator<Item = MsgId> {
+        self.channels
+            .channels
+            .values()
+            .map(|state: &ChannelState| state.tip_message)
+    }
```
davidrusu: I believe this is dead code now, lets remove?

services/tx-service/src/backend/pool.rs
```diff
@@ -1,252 +1,244 @@
...
-        }
-
-        for key in expired_keys {
-            self.removed_items.remove(&key);
+            adapter: adapter.clone(),
+            forks_tracker: ForksTracker::from_state_and_adapter(state.forks_state, &txs, adapter),
+            _phantom: PhantomData,
         }
     }
 }
```
davidrusu, line 242: If recovery is not working, lets just put a dummy implementation here that initializes an empty mempool on restart. We can tackle it in future PR

services/tx-service/src/backend/pool.rs
```diff
@@ -1,252 +1,244 @@
...
+    async fn remove(&mut self, keys: &[Self::TxHash]) {
+        let removed_count = self.forks_tracker.force_remove_txs(keys);
+        let pending_items = self.pending_item_count().await.unwrap_or_else(|e| {
+            tracing::error!(target: LOG_TARGET, "Failed to get pending item count: {}", e);
+            usize::MAX
+        });
+        log_removed_items(removed_count, pending_items);
+        mempool_transactions_removed(removed_count);
+        mempool_transactions_pending(pending_items);
     }
```
davidrusu, line 153: Can we not delete this "remove" path now? the mempool itself tracks inclusion so we don't need the chain-service to notify the mempool to remove txs

core/src/header/mod.rs
danielSanchezQ: Ill call it zeroes, yes, i used it to initialize the mempool. it was used raw in other places too.

core/src/mantle/tx.rs
danielSanchezQ: Yes, but it wasnt exactly the same (just parts). Ill open an issue.

services/tx-service/src/backend/pool.rs
danielSanchezQ: I would prefer to let the bug, the end result is the same, we keep failing instead of silently going ahead on a wrong recovery and it would be easier to tackle later (?)

services/tx-service/src/backend/pool.rs
danielSanchezQ: This is not used for the chain service, but for operators to manually remove txs. We had a conversation about this in discord a couple weeks ago.

ledger/src/mantle/mod.rs
danielSanchezQ: yep, it was a leftover from a merge, thanks!

danielSanchezQ:
Not merging this, @davidrusu will take over.

PR #2748, opened 2026-05-14
chore: rework nom encoding/decoding logic for better type cohesion

ntn-x2:

The current mantle encoding module is a big file that has lots of functions and also a lot of duplicated logic with the actual types used in the NOM encoding/decoding.

This PR applies the structure I envision and applies it to a single op, the `Inscribe` op. The result is that the requirements that should hold true for any `Inscribe` op are now part of the type itself, which is a `BoundedVec` (newly introduced type) with a statically-defined maximum length, which is enforced both at nom decoding and serde deserialize time. Then, we don't need to `assert!` anything anymore at encoding time because we are sure the length of the inscription is within the defined limit. The added benefit is that everything that interacts with inscriptions, e.g., the zone SDK, now benefits from a rich type that won't compile if users of the SDK try to build an inscription that is bigger than the maximum allowed limit, so we also reduce load on the chain as those inscriptions won't go through in the first place.

If we agree this is a good way to go forward, I will open follow-up PRs updating the remaining ops, until we have strong types everywhere.

There are several changes in this PR, but they are all the result of enforcing strict typing when creating inscriptions. The key files to review are:
* everything under `core/src/mantle`, particularly the `nom.rs` module with the new trait and blanket impls and `encoding.rs` with many removals of logic that now lives inside the `InscribeOp` type.
* `utils/src/bounded_vec.rs` for the new `BoundedVec` type which a (hopefully) nice ergonomics
* `zone-sdk/src` for an example of the changes required when using inscriptions. Now if we know an array is within the maximum limit, we can call `[u8; SIZE].into()` directly. If we deal with dynamically-sized `Vec`s, we still need to `try_into()` and deal with the `Result`.

As I said in the PR, with this approach, we can eventually have a `MaxEncodedLen` trait that can represent the maximum statically-enforce length of an operation, or a set of operations, ensuring we never go above that, at compile time.

I think so. As I said, there are a lot of changes but the core of the changes are on the paths above.

@ntn-x2 

N/A

No.


core/src/mantle/nom/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+            matches!(N_BYTES, 1 | 2 | 4 | 8),
+            "N_BYTES must be 1, 2, 4, or 8",
+        );
+        let max_repr: u64 = if N_BYTES == 8 {
+            u64::MAX
+        } else {
+            (1u64 << (N_BYTES * 8)) - 1
+        };
+        assert!(N as u64 <= max_repr, "N exceeds what N_BYTES can encode");
+    };
```
danielSanchezQ: Nice stuff

core/src/mantle/ops/channel/inscribe.rs
```diff
@@ -33,13 +36,35 @@ impl InscriptionOp {
...
-    fn payload_bytes(&self) -> Bytes {
-        encode_channel_inscribe(self).into()
+// ChannelInscribe = ChannelId Inscription Parent Signer
+impl NomEncode for InscriptionOp {
+    fn encode(&self) -> Vec<u8> {
+        let mut bytes = self.channel_id.encode();
+        bytes.extend(self.inscription.encode());
+        bytes.extend(self.parent.encode());
+        bytes.extend(self.signer.encode());
+        bytes
```
danielSanchezQ, line 53: Can we pre-allocate if we know the sizes? This is a general question. Some places could be applied some not I guess.

danielSanchezQ (approved):

utils/src/bounded_vec.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /// Construct without checking the cap.
+    ///
+    /// Reserved for callers that have already validated the length. Prefer
+    /// [`Self::try_from<Vec<T>>`] at trust boundaries.
+    #[must_use]
+    pub const fn new_unchecked(items: Vec<T>) -> Self {
+        Self(items)
+    }
```
hansieodendaal, line 42: `pub const fn new_unchecked(items: Vec<T>) -> Self {` is out of place here as a public API as it bypasses the very invariant `BoundedVec` is supposed to enforce.

utils/src/bounded_vec.rs
```diff
@@ -0,0 +1,158 @@
...
+
+/// `Vec<T>` whose length is statically capped at `MAX`. The cap is enforced at
+/// every construction site (`new`, `TryFrom<Vec<T>>`, deserialization), so an
+/// instance can never hold more than `MAX` elements.
+#[derive(Clone, Debug, Eq, PartialEq, Hash, PartialOrd, Ord, Default, Serialize, Deserialize)]
+#[serde(
+    into = "Vec<T>",
+    try_from = "Vec<T>",
+    bound(serialize = "T: Clone + Serialize")
+)]
```
hansieodendaal, line 24: Is `Clone` required here? This may unnecessarily reject otherwise valid element types where `T: Serialize` but `T: !Clone`.

utils/src/bounded_vec.rs
```diff
@@ -0,0 +1,158 @@
...
+pub struct BoundedVec<T, const MAX: usize>(Vec<T>);
+
+impl<T, const MAX: usize> BoundedVec<T, MAX> {
```
hansieodendaal, line 27: 1. Missing explicit truncation helper - If bounded containers are used at trust boundaries or input normalization points, truncation is sometimes intentional and useful. Example `from_items_truncate`
1. Missing `FromIterator`

pradovic (approved):

core/src/mantle/nom/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    #[must_use]
+    pub const fn new() -> Self {
+        Self(BoundedVec::new())
+    }
+
+    #[must_use]
+    pub const fn new_unchecked(items: Vec<T>) -> Self {
+        Self(BoundedVec::new_unchecked(items))
+    }
```
hansieodendaal: ditto - `pub const fn new_unchecked(items: Vec<T>) -> Self` is out of place here as a public API as it bypasses the very invariant `NomBoundedVec` is supposed to enforce.

core/src/mantle/nom/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+        let Ok(items) = items.try_into() else {
+            panic!("Decoded `N` elements.");
+        };
+        Ok((input, items))
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash, PartialOrd, Ord, Default, Serialize, Deserialize)]
+#[serde(bound(serialize = "T: Clone + Serialize"))]
+pub struct NomBoundedVec<T, const N: usize, const N_BYTES: usize>(BoundedVec<T, N>);
```
hansieodendaal: ditto - Is `Clone` required here? This may unnecessarily reject valid types where `T: Serialize` but `T: !Clone`.

core/src/mantle/ops/channel/inscribe.rs
```diff
@@ -1,29 +1,32 @@
...
         TxHash,
         channel::{ChannelState, Channels, Error},
-        encoding::encode_channel_inscribe,
         ledger::Operation,
+        nom::{NomBoundedVec, NomEncode},
     },
 };
 
+pub const MAX_BYTES: usize = MAX_BLOCK_SIZE * 7 / 8;
+pub type Inscription = NomBoundedVec<u8, MAX_BYTES, 4>;
```
hansieodendaal: This would make the domain model more readable; otherwise the transport concern leaks into it.
```suggestion
pub type InscriptionData = BoundedVec<u8, inscribe::MAX_BYTES>;
```
NOM encoding can be implemented for `InscriptionData` separately

core/src/mantle/ops/channel/inscribe.rs
```diff
@@ -1,29 +1,32 @@
...
+
 #[derive(Clone, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
 pub struct InscriptionOp {
     pub channel_id: ChannelId,
     /// Message to be written in the blockchain
-    #[serde(with = "serde_bytes_vec")]
-    pub inscription: Vec<u8>,
+    #[serde(with = "lb_utils::serde::serde_bytes_slice")]
+    pub inscription: Inscription,
```
hansieodendaal, line 31:
```suggestion
    pub inscription: InscriptionData,
```
The current type mixes the domain constraint with wire-format detail (NOM / length-prefix width), while the struct field really just wants to say “bounded inscription bytes”.

core/src/mantle/encoding.rs
```diff
@@ -41,16 +39,15 @@ use crate::{
...
-/// Protects against unbounded allocation in `decode_channel_inscribe`
-pub const MAX_ENCODE_DECODE_INSCRIPTION_SIZE: u32 = (MAX_BLOCK_SIZE * 7 / 8) as u32;
 // Maximum memory allocation size allowed for SDP activity metadata.
 // Protects against unbounded allocation in `decode_sdp_active`
 const MAX_ENCODE_DECODE_METADATA_SIZE: u32 = 234; // `ActiveMessage` has a fixed size of 234 bytes
 
 // Maximum byte size allowed for a locator in SDPDeclare operations.
 const LOCATOR_BYTES_SIZE_LIMIT: usize = 329usize;
 
+pub type Ops = NomBoundedVec<Op, { u8::MAX as usize }, 1>;
```
hansieodendaal: This could be made clearer too. The raw u8::MAX / 1 makes the alias a bit transport-shaped and forces the reader to decode the meaning.
```suggestion
pub const MAX_OPS_PER_TX: usize = u8::MAX as usize;
pub const OPS_LEN_BYTES: usize = 1;

pub type Ops = NomBoundedVec<Op, MAX_OPS_PER_TX, OPS_LEN_BYTES>;
```

core/benches/mantle_tx_components.rs
```diff
@@ -46,13 +49,14 @@ const SIZES: &[usize] = &[
...
-    })]
-    .into()
+    MantleTx(Ops::new_unchecked(vec![Op::ChannelInscribe(
+        InscriptionOp {
+            channel_id: ChannelId::from([0xAA; 32]),
+            inscription: Inscription::new_unchecked(vec![0xAB; payload_size]),
+            parent: MsgId::from([0xBB; 32]),
+            signer: signing_key.public_key(),
+        },
+    )]))
```
hansieodendaal: We can easily do this without needing a public-facing `new_unchecked`
```suggestion
let inscription = Inscription::try_from(vec![0xAB; payload_size])
    .expect("payload_size should fit in Inscription");

MantleTx([Op::ChannelInscribe(InscriptionOp {
    channel_id: ChannelId::from([0xAA; 32]),
    inscription,
    parent: MsgId::from([0xBB; 32]),
    signer: signing_key.public_key(),
})].into())
```

hansieodendaal (commented):
This is a needed addition to the code base, very nice. I have some specific comments and a couple of concerns/suggestions - detail below.

Other powerful uses would be when we declare a bounded vec and try_push to it, for example ledger transaction inputs and outputs.

Furthermore, we could consider adding `BoundedBytes` as something like
```rust
use core::{fmt, ops::Deref};

use thiserror::Error;

use crate::bounded_vec::BoundedError;

/// Errors returned by [`BoundedBytes`].
#[derive(Debug, Error, Eq, PartialEq)]
pub enum BoundedBytesError {
    #[error(transparent)]
    Bounded(#[from] BoundedError),
    #[error("Invalid hex: {0}")]
    Hex(String),
}

/// Byte buffer whose length is statically capped at `MAX`.
#[derive(Clone, Debug, Eq, PartialEq, Hash, PartialOrd, Ord, Default, Serialize, Deserialize)]
#[serde(try_from = "Vec<u8>", into = "Vec<u8>")]
pub struct BoundedBytes<const MAX: usize>(Vec<u8>);

impl<const MAX: usize> BoundedBytes<MAX> {
    pub const MAX: usize = MAX;

    #[must_use]
    pub const fn new() -> Self {
        Self(Vec::new())
    }

    #[must_use]
    pub const fn len(&self) -> usize {
        self.0.len()
    }

    #[must_use]
    pub const fn is_empty(&self) -> bool {
        self.0.is_empty()
    }

    #[must_use]
    pub fn as_slice(&self) -> &[u8] {
        &self.0
    }

    #[must_use]
    pub fn into_inner(self) -> Vec<u8> {
        self.0
    }

    pub fn try_push(&mut self, byte: u8) -> Result<(), BoundedError> {
        if self.0.len() >= MAX {
            return Err(BoundedError::TooLong {
                actual: self.0.len() + 1,
                max: MAX,
            });
        }
        self.0.push(byte);
        Ok(())
    }

    pub fn try_extend_from_slice(&mut self, bytes: &[u8]) -> Result<(), BoundedError> {
        let new_len = self.0.len() + bytes.len();
        if new_len > MAX {
            return Err(BoundedError::TooLong {
                actual: new_len,
                max: MAX,
            });
        }
        self.0.extend_from_slice(bytes);
        Ok(())
    }
}

impl<const MAX: usize> TryFrom<Vec<u8>> for BoundedBytes<MAX> {
    type Error = BoundedError;

    fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
        if value.len() > MAX {
            return Err(BoundedError::TooLong {
                actual: value.len(),
                max: MAX,
            });
        }
        Ok(Self(value))
    }
}

impl<const MAX: usize> TryFrom<&[u8]> for BoundedBytes<MAX> {
    type Error = BoundedError;

    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
        Self::try_from(value.to_vec())
    }
}

impl<const MAX: usize> TryFrom<&str> for BoundedBytes<MAX> {
    type Error = BoundedBytesError;

    fn try_from(value: &str) -> Result<Self, Self::Error> {
        let bytes = const_hex::decode(value).map_err(|e| BoundedBytesError::Hex(e.to_string()))?;
        Self::try_from(bytes).map_err(BoundedBytesError::from)
    }
}

impl<const MAX: usize> TryFrom<String> for BoundedBytes<MAX> {
    type Error = BoundedBytesError;

    fn try_from(value: String) -> Result<Self, Self::Error> {
        Self::try_from(value.as_str())
    }
}

impl<const MAX: usize, const N: usize> From<[u8; N]> for BoundedBytes<MAX> {
    fn from(value: [u8; N]) -> Self {
        const { assert!(N <= MAX, "Array length exceeds BoundedBytes MAX") }
        Self(value.into())
    }
}

impl<const MAX: usize, const N: usize> From<&[u8; N]> for BoundedBytes<MAX> {
    fn from(value: &[u8; N]) -> Self {
        (*value).into()
    }
}

impl<const MAX: usize> From<BoundedBytes<MAX>> for Vec<u8> {
    fn from(value: BoundedBytes<MAX>) -> Self {
        value.0
    }
}

impl<const MAX: usize> AsRef<[u8]> for BoundedBytes<MAX> {
    fn as_ref(&self) -> &[u8] {
        &self.0
    }
}

impl<const MAX: usize> Deref for BoundedBytes<MAX> {
    type Target = [u8];

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

impl<const MAX: usize> fmt::Display for BoundedBytes<MAX> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.write_str(&const_hex::encode(self.as_slice()))
    }
}
```
Then we could have 
```rust
pub type InscriptionData = BoundedBytes<{ inscribe::MAX_BYTES }>;

#[derive(Clone, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
pub struct InscriptionOp {
    pub channel_id: ChannelId,
    #[serde(with = "lb_utils::serde::serde_bytes_slice")]
    pub inscription: InscriptionData,
    pub parent: MsgId,
    pub signer: Ed25519PublicKey,
}
```

core/src/mantle/ops/channel/inscribe.rs
```diff
@@ -1,29 +1,32 @@
...
     mantle::{
         TxHash,
         channel::{ChannelState, Channels, Error},
-        encoding::encode_channel_inscribe,
         ledger::Operation,
+        nom::{NomBoundedVec, NomEncode},
     },
 };
 
+pub const MAX_BYTES: usize = MAX_BLOCK_SIZE * 7 / 8;
```
youngjoon-lee, line 22: Could you leave a comment explaining where these 7 and 8 come from?

youngjoon-lee (approved):

core/src/mantle/ops/channel/inscribe.rs
ntn-x2: Yeah at the moment we don't, because `encode()` returns `Vec<u8>`. I think we can include a `max_encoded_len` or make it return a `BoundedVec` and at that point we will know the maximum possible length of the encoded payload. But at the moment it's not possible, yet.

utils/src/bounded_vec.rs
ntn-x2: Not really. We have a bunch of `_unchecked` functions throughout our codebase, and they are meant to be used if the caller knows the input is within the set limits. Std functions also have a bunch of unchecked, e.g., for non-zero stuff.

utils/src/bounded_vec.rs
ntn-x2: Yes. It's required by `Vec`'s `Serialize` implementation, and since we are relying on it, we need to have the some bound.

utils/src/bounded_vec.rs
ntn-x2: Do you think truncation provides actual values to callers? What would be the use cases in which it's useful compared to making sure the input fits into whatever limit is configured for the vec?

As for `FromIterator`, it's not possible because it's not fallible, while we need to fail if the iterator contains more elements than the set limit.

core/src/mantle/ops/channel/inscribe.rs
ntn-x2: Mhh ok I see your point. I will think about it. The thing is that we do use the `InscriptionOp` model directly when encoding, but I think we can convert from a `BoundedVec` into a `NomBoundedVec` at encoding time since the last generic is a const so it's a no-op conversion. I am afraid it would introduce one more layer of conversion, but maybe it's a good thing.

core/benches/mantle_tx_components.rs
ntn-x2: Yes, but it's ugly to call `expect`. Also, `try_from` cannot be used in `const` environments, while `new_unchecked` can.

core/src/mantle/ops/channel/inscribe.rs
ntn-x2: This values were there before, I just encapsulated them in the inscribe op. @hansieodendaal maybe can explain those?

ntn-x2:
@hansieodendaal could you explain to me the difference between the existing `BoundedVec` and your suggested `BoundedBytes`? Can't all features be added to the `BoundedVec` type directly?

core/src/mantle/ops/channel/inscribe.rs
hansieodendaal: @youngjoon-lee, it is just a way to make sure a single maxed size inscription transaction can fit into a block. Restricting it to something tied to max block size is easier to reason about than just having a number, from there the fractional calculation, and in this case `7 / 8` is a safe bet.

hansieodendaal:
> @hansieodendaal could you explain to me the difference between the existing `BoundedVec` and your suggested `BoundedBytes`? Can't all features be added to the `BoundedVec` type directly?

`BoundedVec<T, MAX>` is a generic bounded container, while `BoundedBytes<MAX>` as suggested would be a byte-specific type/API with byte-specific semantics like `AsRef<[u8]>`, `try_extend_from_slice`, hex display, hex/string parsing, and byte-oriented serde helpers. Some of that could technically be added for BoundedVec<u8>, but I don’t think `BoundedVec` itself should become byte-specific.

core/benches/mantle_tx_components.rs
hansieodendaal: With the same bad input, the checked construction path fails immediately at the call site, while `new_unchecked` silently creates an invalid value and defers the failure. With a public-facing `new_unchecked`, every call site has to manually uphold the size invariant; with the bounded constructors, that invariant is enforced by construction.

ntn-x2:
> BoundedVec<T, MAX> is a generic bounded container, while BoundedBytes<MAX> as suggested would be a byte-specific type/API with byte-specific semantics like AsRef<[u8]>, try_extend_from_slice, hex display, hex/string parsing, and byte-oriented serde helpers. Some of that could technically be added for BoundedVec, but I don’t think BoundedVec itself should become byte-specific.

Mhh if we design `BoundedVec` properly, `BoundedBytes` can simply become `BoundedVec<u8, MAX>` instead. All those methods that you mention can be abstracted away over the type `T`.

core/benches/mantle_tx_components.rs
ntn-x2: But we have a `TryFrom<Vec<T>>` for that case.

ntn-x2: We are talking about benches here. I purposedly decided to use the unchecked variant for benches and tests. Everything else uses the checked `TryFrom` implementation, or even better the `From<[T; N]>` for arrays.

ntn-x2:
@hansieodendaal your suggestion to split the actual vec type from nom encoding/decoding logic resulted in the following commit: https://github.com/logos-blockchain/logos-blockchain/pull/2748/commits/d0010ef7bb2ca24d78c044b5e865f66329a78c71. Here we use `BoundedVec` for the type itself, and create a `NomBoundedVec` only when encoding/decoding in nom. I like this split up so thanks for pushing back on that. It did mean that we need to split encoding and decoding into two traits, and have an associated type `Output` on decode, else `NomBoundedVec`, would not be able to decode into a `BoundedVec` since it only holds a reference to it internally. @danielSanchezQ let me know what you think of this new approach.

core/benches/mantle_tx_components.rs
hansieodendaal: My comment was not directed at benches, but at the use of `new_unchecked`, which I think we should drop altogether. I will move my comment to the appropriate section.

utils/src/bounded_vec.rs
hansieodendaal: With the same bad input, the checked construction path fails immediately at the call site, while `new_unchecked` silently creates an invalid value and defers the failure. With a public-facing `new_unchecked`, every call site has to manually uphold the size invariant; with the bounded constructors, that invariant is enforced by construction.

ntn-x2: We have `TryFrom<Vec<u8>>` for that. This unchecked is meant to be used in those contexts where we know the payload is of a fixed length. E.g., when we decode using `NomBoundedVec` and we know we take `N` elements, we create the vec with `BoundedVec::new_unchecked` as that is simply more efficient and we know it will never be any other case.

hansieodendaal: I still do not like it, but not going to push back any further on this.

hansieodendaal (approved):
LGTM, nice work!

**Note:** Remaining comments are not deal breakers and could be re-evaluated in future when necessary.

PR #2749, opened 2026-05-14
feat(zone-sdk): atomic withdraw

pradovic:

This PR improves Withdraw API in zone-sdk to support atomic withdraw requests for a single signer case.

API:
`sequencer.publish_atomic_withdraw(inscribe, vec![WithdrawArg])`

Related issue: https://github.com/logos-blockchain/logos-blockchain/issues/2681.

TODO:
Sequencer is past it's time to be refactored (2000 files without unit tests) and types to be more unified. To be done in the separate PR.

Yes

@pradovic @davidrusu 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


zone-sdk/src/state.rs
```diff
@@ -47,13 +103,18 @@ impl ChannelUpdateInfo {
 }
 
 /// Local pending inscription with lineage metadata.
+///
```
pradovic, line 106: Naming of `PendingInscription` and the whole struct should be refactored (probably PendingTx with enum), to be done in separate PR to keep the scope smaller.

For now if withdraws.is_none() -> standard inscription, otherwise atomic_withdraw

tests/src/cucumber/steps/manual_zone/support.rs
```diff
@@ -1150,6 +1173,71 @@ pub async fn submit_zone_withdraw(
...
+pub async fn publish_atomic_zone_withdraw(
+    sequencer: &SequencerHandle<ZoneNodeHttpClient>,
+    sequencer_events: &mut tokio::sync::mpsc::Receiver<Event>,
+    funding_public_key: ZkPublicKey,
+    amount: Value,
+    inscription_data: Vec<u8>,
+    deadline: PublishDeadline,
+) -> Result<ZoneWithdrawSubmission, ZoneTestError> {
+    let withdraw_args = vec![WithdrawArg {
+        outputs: Outputs::new(vec![Note::new(amount, funding_public_key)]),
```
hansieodendaal: Q; Should we maybe also test a withdraw with multiple args? Yeah, multiple args with multiple outputs for each.

tests/src/cucumber/steps/manual_zone/support.rs
```diff
@@ -1150,6 +1173,71 @@ pub async fn submit_zone_withdraw(
...
+            if tx.inscription().payload != inscription_data {
+                continue;
+            }
+            let PublishedTx::AtomicWithdraw(info) = *tx else {
+                return Err(ZoneTestError::SubmitWithdraw {
+                    message:
+                        "publish_atomic_withdraw produced Inscription variant instead of AtomicWithdraw"
+                            .to_owned(),
+                });
+            };
```
hansieodendaal, line 1252: Myabe this 
```
            let PublishedTx::AtomicWithdraw(info) = *tx else {
                return Err(ZoneTestError::SubmitWithdraw {
                    message:
                        "publish_atomic_withdraw produced Inscription variant instead of AtomicWithdraw"
                            .to_owned(),
                });
            };
```
could be a warning and continue instead as a pattern as it is not strictly an error if the sequencer emits other events? Then the pattern could be used in concurrent scenarios.

Or is `if tx.inscription().payload != inscription_data {` above a strong enough guarantee of uniqueness?

zone-sdk/src/sequencer.rs
```diff
@@ -1067,10 +1312,51 @@ fn build_checkpoint(state: &TxState, last_msg_id: MsgId, lib_slot: Slot) -> Sequ
...
+/// Txs for other channels (checkpoint reused across channels) hit the
+/// `submit_other` fallback.
+fn restore_pending_tx(state: &mut TxState, tx: SignedMantleTx, channel_id: ChannelId) {
+    let mut inscribe_meta: Option<(MsgId, MsgId, Vec<u8>)> = None;
+    let mut withdraws: Vec<WithdrawInfo> = Vec::new();
+    for op in tx.mantle_tx.ops() {
+        match op {
+            Op::ChannelInscribe(i) if i.channel_id == channel_id => {
+                inscribe_meta = Some((i.parent, i.id(), i.inscription.clone()));
+            }
```
hansieodendaal, line 1360: Q: Just wondering, is it guaranteed that we can only ever have one `Op::ChannelInscribe` per channel in one transaction?

zone-sdk/src/sequencer.rs
```diff
@@ -1013,24 +1109,173 @@ where
...
+    async fn handle_publish_atomic_withdraw(
+        &mut self,
+        inscribe: Vec<u8>,
+        withdraws: Vec<WithdrawArg>,
+    ) -> Result<Event, Error> {
+        if withdraws.is_empty() {
+            return Err(Error::Network(
+                "publish_atomic_withdraw requires at least one withdraw".into(),
+            ));
+        }
```
hansieodendaal: Let's error early for this implementation if `channel_state.withdraw_threshold > 1`

hansieodendaal (commented):
Looking good, just one or two small things to consider please.

zone-sdk/src/sequencer.rs
pradovic: Our api supports only one grouped with multiple withdrawOp for now, per feature scoper request. But we will support more flexible API in the future. Tx could have more unrelated inscriptions.

hansieodendaal: Maybe in that case we should not overwrite the `inscribe_meta ` but use a vec to collect, even if it is just sized 1 for now?

pradovic: I rewrote it to not include it in the atomic bundle, but still treat it as regular tx and not overwrite, as this can be actually submitted with YJ prepare_tx API, but it is not tracked across reorgs and resubmittion API. 

With current zone-sdk it cannot be done with atomic_withdraw API. Also added the comment.
So in the supported case, we track it through reorgs and everything and return it with the regular flow. 

For prepare_tx it is as it used to be, up to the user to do it custom. Later, when we expand the API to be more flexible (atomic API), than we will cover all the cases. Does this makes sense?

hansieodendaal (approved):
LGTM

PR #2753, opened 2026-05-15
refactor(logging): migrate more log targets

andrussal:

Migrate `libp2p` and `network-service` logs to the shared log target crate, so these crates now use declared targets instead of implicit module-path targets.

This also makes `logos_blockchain` the common prefix for generated project targets, so filters can match all migrated project logs through one root.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


tracing/targets/src/network_service.rs
```diff
@@ -0,0 +1,15 @@
+use lb_log_targets_macros::log_targets;
+
+log_targets! {
+    root = network_service;
+
+    backends::{
+        MOCK,
+        libp2p::swarm::{
```
ntn-x2: Should we remove `swarm`?

tracing/targets/src/libp2p.rs
```diff
@@ -0,0 +1,31 @@
+use lb_log_targets_macros::log_targets;
+
+log_targets! {
+    root = libp2p;
+
+    behaviour::{
+        kademlia::BEHAVIOUR,
```
ntn-x2: Also here I think `kademlia` (or `KADEMLIA`) would be enough. What do you think?

tracing/targets/src/libp2p.rs
```diff
@@ -0,0 +1,31 @@
...
+                UPNP,
+                pcp_core::{
+                    CLIENT,
+                    wire::{
+                        ANNOUNCE,
+                        MAP,
+                    },
+                },
+            },
+            state_machine::transitions::{
```
ntn-x2: Also here maybe no `transitions`?

ntn-x2 (commented):
Nice chnange! I think, personal opinion, we could group better the new log targets, since they currently pretty much follow the module structure. I don't have yet a better suggestion tho.

andrussal:
> Nice chnange! I think, personal opinion, we could group better the new log targets, since they currently pretty much follow the module structure. I don't have yet a better suggestion tho.

Good points. I simplified the hierarchy based on your suggestions, so it does not follow the Rust module tree so closely.

Maybe something like this: Drop a segment if the parent + child already says the same useful thing.

ntn-x2 (approved):
Looks better nice, thanks!

andrussal:
> > Nice chnange! I think, personal opinion, we could group better the new log targets, since they currently pretty much follow the module structure. I don't have yet a better suggestion tho.
> 
> Good points. I simplified the hierarchy based on your suggestions, so it does not follow the Rust module tree so closely.
> 
> Maybe something like this: Drop a segment if the parent + child already says the same useful thing.

Another thought for future: some modules are just there to group code, and I don’t think every such group needs to show up in the log target. If the surrounding path already makes the source clear, we can skip that extra segment.

For example, based on your suggestions, `network_service::backends::libp2p::gossipsub` is clear enough without `swarm`, and `libp2p::behaviour::nat::state_machine::public` is clear enough without `transitions`.

PR #2755, opened 2026-05-15
feat(cli): Node cli utilities 

danielSanchezQ:
This adds 2 commands to the node cli
* participate
  
    Reads a node's user_config.yaml and generates two files ready to be used in genesis distribution:
    - stakeholder.yaml — the node's ZK identity and stake amount
    - provider.yaml — the node's blend service registration (provider key, ZK identity, network locator, and service type)
  
    The locator address is resolved from the blend listening address, substituting the host from the network NAT static config if the listen address is unspecified, or from
    --external-address as an explicit override.

 
*  get-peer-id
  
    Reads a node's user_config.yaml and prints the libp2p PeerId derived from the node's network key. Useful for sharing the node's identity when setting up initial peers or IBD
    config on other nodes.

*It also refactors the commands into a cli module*

Commands are pretty simple

@bacv @danielSanchezQ 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.


core/src/sdp/genesis.rs
```diff
@@ -0,0 +1,25 @@
...
+/// Used to register a service provider at genesis.
+/// The `zk_id` must match the `StakeHolderInfo` entry so the locked note can
+/// be identified.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct ProviderInfo {
+    pub provider_id: Ed25519PublicKey,
+    pub zk_id: ZkPublicKey,
+    pub locators: Locators,
+    pub service_type: ServiceType,
+}
```
danielSanchezQ: Had to move them here to avoid circular deps

nodes/node/binary/Cargo.toml
```diff
@@ -84,8 +84,7 @@ tikv-jemallocator = { optional = true, workspace = true }
 rand = { workspace = true }
 
 [features]
-config-gen = ["lb-key-management-system-service/unsafe"]
-default    = ["config-gen", "tracing"]
```
danielSanchezQ: I removed the config-gen feature as we decided it was needed always anyway

nodes/node/binary/src/cli/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+}
+
+#[derive(Parser, Debug)]
+pub struct ParticipateArgs {
+    /// Path to the user config YAML file
+    #[arg(long, default_value = "user_config.yaml")]
+    pub config: PathBuf,
+    /// Stake amount in base units
+    #[arg(long)]
+    pub stake: u64,
```
bacv: Just a note, I think once we have a mechanism other than faucet for the testnet, users shouldn't be required to enter their stake, only ids.

bacv (approved):
Looks good, will update my PRs once this is merged.

nodes/node/binary/src/cli/mod.rs
danielSanchezQ: I can remove it now. In fact we can generate a single file with the data for the peer instead of 2 and use that data file to generate the stakeholder and provider files. wdyt?

bacv: > single file with the data for the peer

That would be ideal. And that would match the future version, where we collect `OperatorInfo` to generate genesis block. Could in this case `StakeholderInfo` and `ProviderInfo` go back to blockchain-tools crate?

danielSanchezQ: @bacv I merged everything into a single file `participation_data.yaml`:

```yaml
stakeholder_identity: a725f24c57ba5d685fa6a9ad01bbe29aba03c12b5ba3b4312b39909b28a6c12b
provider_id: 7716827c099b4d0c5bcb8ee0aa323ebb12b4a612780c7f8f8507c8f839f2ab33
locators:
- /dns4/foo.bar.com/udp/3400/quic-v1
service_type: BN
```

bacv: This looks better now, and requires no changes if we assume that blend core settings will always be present in the config. But is "core" section in blend settings always present?

bacv: I mean this shouldn't fail to get participant data if for some reason blend is not set.

danielSanchezQ: True, I made it optional and added a blend tag to nest blend related information:

```yaml
stakeholder_identity: a725f24c57ba5d685fa6a9ad01bbe29aba03c12b5ba3b4312b39909b28a6c12b
blend:
  provider_id: 7716827c099b4d0c5bcb8ee0aa323ebb12b4a612780c7f8f8507c8f839f2ab33
  locators:
  - /dns4/foo.bar.com/udp/3400/quic-v1
  service_type: BN
```

bacv (approved):
Looks great.

PR #2756, opened 2026-05-17
feat: add zone sdk round robin

hansieodendaal:
### Summary

  This PR adds turn-aware decentralized sequencing to the Zone SDK so a sequencer only posts `CHANNEL_INSCRIBE` transactions when its local signing key is the currently authorized round-robin sequencer for the channel.

  The SDK now reads and tracks `ChannelState`, predicts the current slot, checks `ChannelState::round_robin(current_slot)` before posting pending publish txs, and keeps publish requests in the existing pending outbox while it
  is not our turn. Pending publish txs survive checkpoint/restart and are posted once the local sequencer becomes authorized.

  The PR also exposes live sequencing state to `tui-zone` and adds e2e coverage for wait-for-turn, checkpoint restoration, bounded pending submit depth, unbounded pending submit depth, and immediate-publish paths.

  ### Problem

  Protocol validation already enforces that only the currently authorized channel sequencer may post `CHANNEL_INSCRIBE` messages. Before this change, the Zone SDK signed and submitted inscriptions eagerly, without checking
  whether the local key was authorized for the current turn.

  In multi-sequencer deployments, that could produce predictable validator rejections and poor operator visibility.

  ### What Changed

  Added client-side access to existing node routes:

  - `GET /cryptarchia/info`
  - `GET /channel/:id`

  Mapped `channel not found` responses to `Ok(None)` in the SDK HTTP client, without changing node routes.

  Extended `SequencerConfig` with:

  - `slot_duration`
  - `chain_start_time`
  - `min_slots_remaining_in_turn`
  - `max_pending_publish_depth`
  - `auto_requeue_orphaned`

  Added a sequencer-owned `SlotClock` to predict the current slot and correct it from processed block events.

  Updated sequencer startup to:

  - fetch chain info
  - initialize/sanitize the slot clock
  - fetch initial channel state
  - compute the local accredited key index

  Refactored normal publish flow so it:

  - accepts `publish_message` requests as fire-and-forget work
  - builds signed pending publish txs in the existing outbox
  - checks `channel.round_robin(current_slot)` before posting pending txs to the node
  - waits when it is not our turn
  - posts pending txs once the local sequencer is authorized
  - bounds first-time active posts with `max_pending_publish_depth`
  - allows already-posted pending txs to be reposted for mempool recovery
  - removes pending publish txs only after finalization

  Refreshed channel state on every processed block and attempted pending submit after refresh.

  Added checkpoint support for restored pending publish txs so a restarted sequencer can continue posting its pending outbox on a later authorized turn.

  Added shared sequencing state via watch:

  - `SequencerChannelView`
  - `SequencerHandle::subscribe_channel_view()`

  Updated `tui-zone` to display:

  - current slot
  - own key index
  - authorized key index
  - turn status
  - posting timeframe / timeout
  - accredited key count
  - pending publish txs / queued messages
  - `tip_message`

  Updated republish-policy test support to treat `ChannelUpdate.orphaned` as authoritative for local orphaned outbox entries. This avoids dropping orphaned payloads that were invalidated before the policy observed a
  `Published` event.

  ### E2E Coverage

  Added or updated cucumber coverage for:

  - waiting until our turn before posting pending messages
  - publishing immediately when already authorized
  - checkpointing and restoring pending publish txs
  - bounded pending submit depth
  - unbounded pending submit depth
  - dynamically adding sequencers to a round-robin channel
  - concurrent republish convergence without duplicates
  - concurrent identical payload convergence

  The round-robin scenarios now assert the lifecycle explicitly:

  1. messages are accepted while not our turn
  2. pending publish txs survive checkpoint/restart
  3. only the configured submit depth emits `Published` events on turn
  4. posted txs remain pending until finalized
  5. pending count drops after indexer/finality

  ### Behavioral Notes

  The SDK now behaves conservatively: if it cannot confidently determine authorization, it waits instead of submitting an obviously invalid inscription.

  This implementation does not introduce a separate round-robin queue. It reuses the existing pending tx outbox and gates only the node submission step.

  `pending_publish_txs` means unfinalized publish txs. It includes both not-yet-posted txs and posted-but-not-yet-finalized txs.

  `max_pending_publish_depth` limits first-time active publish posts. Already-posted pending txs may still be reposted for mempool recovery.

  Channel sequencing state uses `tip_message` terminology consistently, matching `ChannelState`.

  Channel config handling remains separate from normal turn-gated publish flow.

  Validator-side validation remains authoritative; the SDK just avoids predictable invalid submissions.
 
Yes

@hansieodendaal @thomaslavaur 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


zone-sdk/src/sequencer.rs
```diff
@@ -665,19 +831,141 @@ where
...
+                .channel_state
+                .as_ref()
+                .map(|channel| u32::from(channel.posting_timeout.clone())),
+            accredited_key_count: self
+                .channel_state
+                .as_ref()
+                .map(|channel| channel.accredited_keys.len()),
+        };
+        drop(self.channel_view_tx.send(view));
+    }
```
danielSanchezQ: Nitpick: This is doing more stuff than recomputing the channel view right?
Also this feels like it would be better as a state transition rather than mutating self.
My 2cts on the structure.

zone-sdk/src/sequencer.rs
```diff
@@ -976,41 +1315,66 @@ where
...
-    async fn handle_publish(&mut self, data: Vec<u8>) -> Event {
-        // Safe to unwrap — handle_request checks is_ready() first
-        let s = self.state.as_mut().unwrap();
-
-        // Derive publish parent from state instead of trusting
-        // last_msg_id blindly — handles branch switches correctly.
-        let parent = if let Some(tip) = self.current_tip {
-            s.publish_parent(tip)
+    async fn handle_publish(&mut self, data: Vec<u8>) -> Option<Event> {
+        if self.can_publish_now() {
```
danielSanchezQ: I know it probably way too big a refactor. But wouldn't you say that a publish event to trigger this things would have been much better?

zone-sdk/src/sequencer.rs
pradovic: Hmm, this is actually a good input for our whole state, to keep it that way. Will take a note to myself for the next refactor after we stabilize the features, thanks!

zone-sdk/src/sequencer.rs
pradovic: Just my 2c, maybe you can fill the pending, and publish when publish event is fired.  It would maybe make sense for the whole pending set now instead of the resubmit interval. Or maybe combination (publish all from pending that are on round robin channel).
Should be possible to reuse existing tokio select with additional channel or something similar.
Just an idea to consider.

hansieodendaal: Please see updated

zone-sdk/src/sequencer.rs
hansieodendaal: Please see updated

nodes/node/http-client/src/lib.rs
```diff
@@ -279,26 +283,47 @@ impl CommonHttpClient {
         self.post(request_url, declaration).await
     }
 
-    /// Get the on-chain state for a channel (`accredited_keys`,
-    /// `withdrawal_nonce`, balance, etc.). Returns the tip-of-chain view.
-    pub async fn get_channel_state(
+    /// Get consensus info (tip, height, etc.)
```
pradovic: nitpick: duplicated comment

pradovic (commented):
This looks solid overall! 

I have two concerns that I think are important to consider here — they’re tied together and could actually simplify the change rather than expand it.

1. Drain the full queue while can_publish_now() holds.
You already mention this as a future optimization in the doc comment at try_publish_queued. Given TxState handles parent chaining via publish_parent(tip) and off-branch recovery via shed_off_branch_pending, off-turn rejections would flow through the same orphan path as reorgs. Could we drain via the existing in_flight: FuturesUnordered?

2. Collapse pending_turn_queue into TxState.pending_*.
The new queue looks volatile across checkpoints — SequencerCheckpoint only restores pending_txs, not pending_turn_queue. So publish_message no longer has the pre-round-robin durability semantics: caller gets Ok(()), but if the sequencer crashes before its turn arrives, the request is silently lost.

Proposed flow:
- publish_message signs immediately using state.publish_parent(tip) as today
- registers the signed tx into TxState.pending_*  instead of a pending_turn_queue→ durable via existing checkpoint
- the turn gate only controls the post step
- drain via in_flight: FuturesUnordered while can_publish_now() holds

This subsumes point 1 cleanly, removes the volatile pending_turn_queue state, reuses the parent-chaining and orphan-handling we already pay for, and restores the original durability contract of publish_message.

Net effect: less code and fewer lifecycle stages, unless I am missing something.

hansieodendaal:
> I have two concerns that I think are important to consider here — they’re tied together and could actually simplify the change rather than expand it.

Hi @pradovic, thank you for those comments, they were on the dot, and I implemented changes to accommodate both.

pradovic:
Hi @hansieodendaal,

Thanks for addressing both points! Sorry, I still have some concerns — not trying to be nit-picky, I just feel we might be able to reduce complexity and reuse more of the existing flows, unless there’s a reason not to.

I think we’re still carrying more lifecycle complexity than needed: pending_turn_queue + TxState.pending as partially overlapping mechanisms, with SequencerCheckpoint now serializing both independently (two fields for what is conceptually the same intent), separate restore paths, a new drain-limit config knob, and sequential drain rather than the existing in_flight pipeline. Is there a reason not to re-use existing FuturesUnordered and make it concurrent?

Less about elegance, more about API/runtime behavior over time. The “collapse into TxState.pending + gate only the post step” approach still seems like it would reduce moving parts:

- sign immediately using publish_parent(tip)
- persist only through existing TxState.pending
- drain via existing in_flight (FuturesUnordered) while can_publish_now() — pipelined instead of serial, avoiding blocking the drive loop
- reuse existing orphan / re-sign flow

If this reasoning is correct, I think this direction would actually simplify the implementation overall and reduce the total code surface. Let me know what you think 🙂

hansieodendaal:
Hi @pradovic, thank you for your comments. There are specific reasons I believe the proposed collapse is not safe. We are dealing with two separate lifecycle phases here:
- `pending_turn_queue`: accepted publish intents that have not been built, signed, or submitted because the sequencer is not authorized yet;
- `TxState.pending`: concrete signed transactions with fixed `parent_msg`, `this_msg`, `tx_hash`, and chain-facing lifecycle.

The main theme of this PR is to make sure we only commit pending publish intents once we know this sequencer is authorized to write and can choose the correct parent lineage.

Once something enters `TxState.pending`, the rest of the SDK treats it as a concrete pending transaction: it participates in `publish_parent(...)`, `pending_txs(...)`, orphan detection, resubmission, finalization, checkpoint restore, and event correlation. An off-turn publish intent is not that yet.

The riskiest part of the suggestion is “sign immediately using `publish_parent(tip)`.” Parent selection is exactly what should be deferred until the sequencer is allowed to write. Between enqueue time and turn to write time, the canonical channel tip can change, pending suffixes can become ambiguous, other sequencers can publish, and earlier queued messages from this sequencer may need to advance local lineage first. Signing early freezes a parent before we know it is still valid.

I also do not think `in_flight: FuturesUnordered<BoxFuture<'static, InFlight>>` is the right abstraction for newly built queued publishes. In this code, `in_flight` is used for resubmitting already-built signed transactions. Queued publishes are different: each successful publish mutates `TxState` and advances `last_msg_id`, and the next queued publish should chain from that updated local state. Running those concurrently risks producing siblings from the same parent or transactions built from stale lineage.

Similarly, the two checkpoint fields represent two lifecycle phases:
  - `pending_txs`: signed transactions already in the transaction lifecycle;
  - `pending_turn_queue`: unsigned accepted intents waiting for authorization.

pradovic:
> Hi @pradovic, thank you for your comments. There are specific reasons I believe the proposed collapse is not safe. We are dealing with two separate lifecycle phases here:
> 
>     * `pending_turn_queue`: accepted publish intents that have not been built, signed, or submitted because the sequencer is not authorized yet;
> 
>     * `TxState.pending`: concrete signed transactions with fixed `parent_msg`, `this_msg`, `tx_hash`, and chain-facing lifecycle.
> 
> 
> The main theme of this PR is to make sure we only commit pending publish intents once we know this sequencer is authorized to write and can choose the correct parent lineage.
> 
> Once something enters `TxState.pending`, the rest of the SDK treats it as a concrete pending transaction: it participates in `publish_parent(...)`, `pending_txs(...)`, orphan detection, resubmission, finalization, checkpoint restore, and event correlation. An off-turn publish intent is not that yet.
> 
> The riskiest part of the suggestion is “sign immediately using `publish_parent(tip)`.” Parent selection is exactly what should be deferred until the sequencer is allowed to write. Between enqueue time and turn to write time, the canonical channel tip can change, pending suffixes can become ambiguous, other sequencers can publish, and earlier queued messages from this sequencer may need to advance local lineage first. Signing early freezes a parent before we know it is still valid.
> 
> I also do not think `in_flight: FuturesUnordered<BoxFuture<'static, InFlight>>` is the right abstraction for newly built queued publishes. In this code, `in_flight` is used for resubmitting already-built signed transactions. Queued publishes are different: each successful publish mutates `TxState` and advances `last_msg_id`, and the next queued publish should chain from that updated local state. Running those concurrently risks producing siblings from the same parent or transactions built from stale lineage.
> 
> Similarly, the two checkpoint fields represent two lifecycle phases:
> 
>     * `pending_txs`: signed transactions already in the transaction lifecycle;
> 
>     * `pending_turn_queue`: unsigned accepted intents waiting for authorization.

Thanks, this clarifies the disagreement.

I think my main concern is that gating signing/building on round-robin authorization does not really match the existing zone-sdk lifecycle.

TxState.pending is already speculative. A tx can enter pending, then later become invalid because the canonical tip changed, another sequencer published first, or our local suffix moved off-branch. That is exactly what shed_off_branch_pending, OrphanedTx, and the re-sign/re-publish flow are for.

So I think round-robin should gate post_transaction, not signing/building.

In the single-sequencer case, we are the only writer moving the tip. In the multi-sequencer case, if another sequencer moves the tip before our tx lands, the existing orphan/re-sign path should handle that the same way it handles reorg/conflict today.

There is also a concrete inconsistency in the current model: the existing resubmission interval already takes txs from TxState.pending and re-posts them via post_transaction without checking round-robin authorization (sequencer.rs:864, gated only on current_tip.is_some() && !resubmit_active). So if “only post while authorized” is the invariant, we need to enforce it at the posting boundary anyway, including resubmission.

On in_flight: I also do not think it is resubmit-only. It is already the async posting pipeline (sequencer.rs:2190), and resubmit is just the current producer. Queued publish drains could use the same pipeline. Draining serially seems unnecessarily slow and can block the drive loop, while FuturesUnordered lets us pipeline the post_transaction calls once can_publish_now() holds.

So the model I’m arguing for is:

- TxState.pending owns the speculative tx lifecycle
- round-robin gates every post_transaction attempt, including first publish and re-submission
- in_flight pipelines posting while the gate is open
- if lineage becomes stale, the existing orphan/re-sign flow handles it (whereas in the current PR orphaned queued items are not surfaced back to the user for re-sign/re-submit decisions)

That keeps the authorization boundary in one place and avoids introducing a second pending lifecycle. The current integration tests were already passing with round-robin, it should just makes it more efficient and aim for less resubmittion, but I feel likle the most of the existing logic is already covering it.

tests/cucumber_tests/features/zone.feature
```diff
@@ -201,7 +202,159 @@ Feature: Zone SDK
...
+
+  @zone_ci
+  Scenario: Round-robin waits for turn and submits pending messages
+    Given I have a zone cluster
+    And the following zone sequencers exist:
+      | alias |
+      | SEQ_A |
+      | SEQ_B |
+    When the zone node is at height 1 in 120 seconds
+    And I start zone sequencer "SEQ_A" with pending submit depth "2" with indexer
```
andrussal: Maybe it would be nice to declare sequencer runtime startup in a table instead of adding more step variants?

For example, instead of separate steps like:

```gherkin
And I start zone sequencer "SEQ_A" with indexer
And I start zone sequencer "SEQ_B"
And I start zone sequencer "SEQ_A" with pending submit depth "2"
```

we could have one step with options:

```gherkin
And I start zone sequencers:
  | alias | indexer | pending_submit_depth |
  | SEQ_A | true    | 2                    |
  | SEQ_B | false   | 2                    |
```

But I don't mind doing it later in my opened PR after this gets merged.

tests/src/cucumber/steps/manual_zone/support.rs
```diff
@@ -305,39 +318,18 @@ pub fn start_republish_policy(
...
+                    let OrphanedTx::Inscription(inscription) = entry else {
+                        // Republish-by-payload helper doesn't handle bundles.
+                        continue;
+                    };
 
-                        if let Err(error) = handle.publish_message(inscription.payload).await {
-                            warn!(%error, "Failed to re-publish orphaned zone payload");
-                        }
+                    if let Err(error) = handle.publish_message(inscription.payload).await {
+                        warn!(%error, "Failed to re-publish orphaned zone payload");
```
andrussal: Would it be better to propagate this error back to the scenario somehow instead of only logging it? There are a few similar cases below in background policy tasks as well. But maybe it's not worth of trouble, just a thought.

tests/src/cucumber/steps/manual_zone/support.rs
```diff
@@ -374,6 +369,8 @@ pub fn start_balance_aware_policy(
                         orphaned_inscriptions,
                     )
                     .await;
+                    publish_planned_balance_updates(&handle, &view_rx, &mut balances, &mut planned)
```
andrussal: Should this be called after the `match` instead of only inside the `ChannelUpdate` branch? Otherwise the initial planned balance updates seem to depend on a channel update happening first, and this scenario could pass without publishing any of them.

tests/cucumber_tests/features/zone.feature
```diff
@@ -201,7 +202,159 @@ Feature: Zone SDK
...
+
+  @zone_ci
+  Scenario: Round-robin submits all pending messages with no active depth limit
+    Given I have a zone cluster
+    And the following zone sequencers exist:
+      | alias |
+      | SEQ_A |
+      | SEQ_B |
+    When the zone node is at height 1 in 120 seconds
+    And I start zone sequencer "SEQ_A" with pending submit depth "None" with indexer
```
andrussal: nit: We could use something less Rust specific in place of None, maybe `unlimited`.

nodes/node/http-client/src/lib.rs
```diff
@@ -245,26 +249,46 @@ impl CommonHttpClient {
...
+        let response = request.send().await.map_err(Error::Request)?;
+        let status = response.status();
+        let body = response.text().await.map_err(Error::Request)?;
+
+        match status {
+            StatusCode::OK => serde_json::from_str::<ChannelState>(&body)
+                .map(Some)
+                .map_err(|e| Error::Server(format!("Failed to parse response: {e}"))),
+            StatusCode::NOT_FOUND => Ok(None),
+            StatusCode::INTERNAL_SERVER_ERROR if Self::is_channel_not_found(&body) => Ok(None),
```
andrussal, line 318: nit: looks like maybe better to return 404 as status at some point?

tests/cucumber_tests/features/zone.feature
hansieodendaal: Good suggestion

tests/cucumber_tests/features/zone.feature
hansieodendaal: Good suggestion

nodes/node/http-client/src/lib.rs
hansieodendaal: Agreed - issue #2794 created

tests/src/cucumber/steps/manual_zone/support.rs
hansieodendaal: I would say at this stage it is not a test failure yet, so maybe worth just logging it to the console. If it cannot be resolved dynamically, the scenario will error out.

zone-sdk/src/sequencer.rs
```diff
@@ -725,39 +885,289 @@ where
...
+    }
+
+    fn can_publish_now(&self) -> bool {
+        let Some(slot_clock) = &self.slot_clock else {
+            return false;
+        };
+        let current_slot = slot_clock.current_slot();
+
+        let Some(channel) = &self.channel_state else {
+            return true;
```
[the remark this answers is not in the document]
hansieodendaal: No, see below:
```rust
        let Some(channel) = &self.channel_state else {
            // A missing channel is the normal pre-genesis-inscription state.
            // Network/query failures are surfaced before this point, so this
            // still only publishes when the absence is known.
            return true;
        };
```

zone-sdk/src/sequencer.rs
```diff
@@ -756,39 +925,296 @@ where
             }
         };
 
+        if let Some(slot_clock) = self.slot_clock.as_mut() {
+            slot_clock.observe_slot(block_event.tip_slot);
+        }
+
+        if let Err(err) = self.refresh_channel_state().await {
```
pradovic, line 930: Why are we querying node on each block? I think we are following the channel update events and could save one http call if we follow them.
Not super expensive, but it could add up during back-fill maybe or in some situations. 

Also error is swollen. Similar to your comment on my PR I think correct behavior should be to break the stream and let it reconnect and retry instead of living with invalid state.

zone-sdk/src/sequencer.rs
```diff
@@ -106,10 +116,141 @@ impl Default for SequencerConfig {
...
+            min_slots_remaining_in_turn: 1,
+            auto_requeue_orphaned: false,
+            max_pending_publish_depth: 10,
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct SlotClock {
+    slot_duration: Duration,
```
pradovic, line 128: I am not sure it is the best to rely inertly on the system time. Would it make sense to improve self correct to query cryptarchia info instead of Instant::now()?

zone-sdk/src/sequencer.rs
```diff
@@ -98,6 +103,11 @@ pub struct SequencerConfig {
     pub resubmit_interval: Duration,
     pub reconnect_delay: Duration,
     pub publish_channel_capacity: usize,
+    pub slot_duration: Duration,
+    pub chain_start_time: Option<SystemTime>,
+    pub min_slots_remaining_in_turn: u64,
+    pub auto_requeue_orphaned: bool,
```
pradovic: Why do we need auto_requeue_orphaned? Orphans are already surfaced via Event::ChannelUpdate { orphaned, .. } for the consumer to handle — that's the documented contract sequnecer.

There's also a possible  bug, unless I am missing something: at zone-sdk/src/sequencer.rs:1524-1549, build_channel_event calls rebuild_orphaned_inscriptions and then emits the same orphans in Event::ChannelUpdate. So with auto_requeue_orphaned: true plus any consumer that handles OrphanedTx::Inscription (the TUI does, the test policy does), you double-republish — the SDK rebuilds it and the consumer also rebuilds it from the event.

I'd rather drop this flag entirely. Keep the SDK's job single-sourced: orphans always go out via ChannelUpdate, the consumer always handles them. As it is up to the consumer to provide the unique payloads so this makes the contract more obvious for the user, and makes them handle it.

pradovic (changes requested):
Looking good!

I left some comments to consider, especially the one about the auto resubmitting orphans.

Could we also add an event to the user that it is his turn to write, so the user can also wait for their turn to publish as well as sequencers should not be publishing when it is not their turn to publish. We could also add e2e test (or modify existing one) that uses this event and published only when it is it's turn.

Edit:
Agreed with @hansieodendaal to tackle suggested event and e2e test in a separate PR.

zone-sdk/src/sequencer.rs
hansieodendaal: As discussed, it is auto corrected with block stream events

zone-sdk/src/sequencer.rs
hansieodendaal: Will break stream on error, good call.

Thinking about the `self.refresh_channel_state().await` on every block event...
Not sure that if `accredited_keys` changes it will be an event we can track. For now, the call is intentionally blocking because publish gating depends on fresh ChannelState; maybe if block stream carries channel state, we can remove the extra HTTP call.

andrussal (approved):
LGTM!

(Reviewed e2e testing code)

pradovic (approved):
Looks great! Thank you!

PR #2764, opened 2026-05-18
feat(chain): remove concept of session from ledger and chain-service

youngjoon-lee:
### RFCs
- [Remove Concept of a Session](https://www.notion.so/nomos-tech/RFC-Remove-Concept-of-a-Session-f39261aa09df826db83e81613bb454dc?source=copy_link)
  - [SDP](https://www.notion.so/nomos-tech/1-0-0-Service-Declaration-Protocol-33d261aa09df808f834bc728ae0c5e0f?source=copy_link)

### Summary

(Unchecked tasks will be done in next PRs.)
- [x] Removed concept of session from `lb-ledger`
- [ ] Removed concept of session from `lb-ledger::mantle::sdp::rewards`.
- [x] Chain service should yield a new SDP snapshot on every epoch. 
  - [x] Chain service should take a SDP snapshot at the most recent block that is older than both LIB and the last block of `current_epoch - 2`.
- [x] Tests (More tests will be added in next PRs.)

### Details

- Now, `lb-ledger` doesn't manage any snapshot. It just accumulates/removes declarations when it builds a new `SdpLedger` when a new block is added.
  - I didn't change SRP yet. For now, no activity tokens are accumulated, and no rewards are distributed. But, activity messages are accepted as long as they have the correct nonce, and so on. 
- Chan service persists the SDP declarations of each block to the storage, along with the block. Of course, it removes them when removing corresponding blocks from storage (e.g. removing stale forks). 
- When a new epoch starts (i.e. triggered by time service), the chain service finds the SDP snapshot block on demand, by walking back blocks in storage, starting from LIB. As defined in the [RFC](https://www.notion.so/nomos-tech/1-0-0-Service-Declaration-Protocol-33d261aa09df808f834bc728ae0c5e0f?source=copy_link#33d261aa09df8093b614c27fb112358c), it uses the most recent block that is older than both LIB and the last block of `current_epoch - 2`. 

> [!IMPORTANT]  
> explained.  

> [!IMPORTANT]  
> advancing  
> the review.

> [!IMPORTANT]

> [!IMPORTANT]
    modification/solutions.

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


core/src/sdp/mod.rs
```diff
@@ -30,17 +37,59 @@ pub struct MinStake {
...
+impl NumberOfEpochs {
     #[must_use]
-    pub const fn session_for_block(&self, block_number: BlockNumber) -> SessionNumber {
-        block_number / self.session_duration
+    pub const fn new(epoch: Epoch) -> Self {
+        Self(epoch)
+    }
+
+    #[must_use]
+    pub const fn into_inner(self) -> Epoch {
```
ntn-x2: Why would we want to convert a number of epochs into an epoch? I think adding to an epoch would be ok, which is done below.

core/src/sdp/mod.rs
```diff
@@ -191,21 +240,62 @@ pub struct ProviderInfo {
...
+
+impl FromIterator<(ServiceType, HashMap<DeclarationId, Declaration>)> for Declarations {
+    fn from_iter<I: IntoIterator<Item = (ServiceType, HashMap<DeclarationId, Declaration>)>>(
+        iter: I,
+    ) -> Self {
+        Self(iter.into_iter().collect())
+    }
+}
+
+impl TryFrom<Bytes> for Declarations {
```
ntn-x2, line 349: Isn't `Deserialize` enough for this?

core/src/sdp/mod.rs
```diff
@@ -191,21 +240,62 @@ pub struct ProviderInfo {
...
+    type Error = codec::Error;
+
+    fn try_from(bytes: Bytes) -> Result<Self, Self::Error> {
+        Self::from_bytes(&bytes)
+    }
+}
+
+impl TryFrom<Declarations> for Bytes {
```
ntn-x2, line 357: Same here. Isn't `Serialize` enough? More flexible also, I'd say.

ledger/src/mantle/sdp/mod.rs
```diff
@@ -172,93 +155,66 @@ pub enum Error {
...
-            assert!(
-                current_session == self.active.session_n,
-                "Logos blockchain isn't ready for time travel yet: session_of_block={current_session}, active_session={}",
-                self.active.session_n
-            );
-            self.rewards = self.rewards.update_epoch(epoch_state, rewards_params);
-            reward_utxos = Vec::new();
-        }
+        // Remove expired declarations based on retention_period
+        self.declarations = self
```
ntn-x2, line 195: Just a note that we also briefly discussed in private: here, we remove declarations from the ledger state of tip, based on the current epoch. Nevertheless, we know that before a node can submit any active message 2 epochs have to pass until the declaration is included in the finalized snapshot, so the earliest the node can submit an active message will be during epoch N+3. Doesn't this logic here conflict with that? Since we will for sure apply a block during epoch N+3 before the node gets a chance to submit an active message, which could lead to the `DeclarationNotFound` issue, no?

services/storage/src/api/chain/mod.rs
```diff
@@ -20,6 +20,7 @@ use crate::api::backend::rocksdb::chain::HeaderIdStream;
 pub trait StorageChainApi {
     type Error: Error + Send + Sync + Debug + 'static;
     type Block: Send + Sync;
+    type SdpDeclarations: Send + Sync;
```
ntn-x2, line 23: As far as I see declarations are always stored and retrieved together, so it's probably better to leave this trait generic over `Block` and define a concrete `Block` type that also includes SDP declarations, but could include more in the future. I don't think SDP declarations should bubble up to this point. What do you think?

services/storage/src/api/backend/rocksdb/chain.rs
```diff
@@ -32,7 +33,9 @@ pub type HeaderIdStream = Pin<Box<dyn Stream<Item = Result<HeaderId, Error>> + S
 impl StorageChainApi for RocksBackend {
     type Error = Error;
     type Block = Bytes;
+    type SdpDeclarations = Bytes;
```
ntn-x2, line 35: Ok I see why you introduced a new associated type here: using `Bytes` for everything is not the best.

services/chain/chain-service/src/lib.rs
```diff
@@ -1535,72 +1533,41 @@ where
...
-
-                let broadcast_relay = relays.broadcast_relay();
-
-                let broadcast_future = match service {
-                    ServiceType::BlendNetwork => {
-                        broadcast_blend_session(broadcast_relay, update).boxed()
-                    }
+        chain_start_time_passed: bool,
+    ) -> Option<Declarations> {
+        if chain_start_time_passed {
```
ntn-x2: Could be a reverse condition with an early exit to avoid one level of indentation.

services/chain/chain-service/src/sdp/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+use crate::{LOG_TARGET, relays::BroadcastRelay};
+
+/// Take/broadcast a SDP snapshot at the most recent block that is older
+/// than both the LIB and the last block of epoch `epoch -
+/// SNAPSHOT_FINALIZATION_DELAY`.
+///
+/// The caller supplies LIB's id, slot, and the in-memory declarations at LIB
+/// (used on the fast path when LIB is already old enough). Storage is only
+/// queried when the walk needs to go past LIB.
+pub async fn take_and_broadcast_sdp_snapshot<Storage>(
```
ntn-x2, line 19: I was thinking that when we apply a block, we know if the new block belongs to a new epoch, so we could store the block ID of its parent indexed by the epoch, so that when a new epoch `N` starts, we can fetch in `O(1)` the last block of epoch `N-2` directly. This would require us to write once per epoch-crossing for each fork (so not that many times) + removal of the previous value stored. It's probably an optimization we don't need to introduce now, but it would make this lookup process effectively `O(1)` instead of `O(N)`, with a few minor things to clean up on the way. When we look it up, we can then easily compare the height of that block with LIB to draw our conclusions.

services/chain/chain-service/src/sdp/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+    storage: &Storage,
+    broadcast_relay: &BroadcastRelay,
+) -> Result<Declarations, DynError>
+where
+    Storage: storage::Storage + Sync,
+{
+    let (snapshot, snapshot_slot) =
+        take_sdp_snapshot(epoch, lib_id, genesis_declarations, config, storage).await?;
+    broadcast_sdp_snapshot(epoch, &snapshot, broadcast_relay).await;
+    info!(target: LOG_TARGET, ?epoch, ?snapshot_slot, "took/broadcasted SDP snapshot");
```
ntn-x2: This is more of a `debug` entry not relevant for end users?

services/chain/chain-service/src/sdp/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+                                    locators: declaration.locators.clone(),
+                                    zk_id: declaration.zk_id,
+                                },
+                            )
+                        })
+                        .collect(),
+                };
+
+                if let Err(e) = broadcast_blend_providers(relay, providers).await {
+                    error!(target: LOG_TARGET, ?epoch, err = ?e, "Failed to broadcast a new blend SDP snapshot");
```
ntn-x2: We know these are not really errors since it happens when there's no receivers, so I think `debug` or even `trace` would be better, in my opinion.

ntn-x2 (changes requested):
I know it's an early stage PR, but I left some comments that will also be valid once the PR has been packaged up better. Overall I don't think the changeset is that large, all things considered. Nevertheless, after checking the SDP core myself a bit deeper, I now understand some of your concerns regarding this approach. Nevertheless, I think this is a step in the right direction.

ntn-x2 (commented):
Sorry did not mean to request changes! I was still finishing to write my review and I pressed by mistake!

What I was saying is that we can now zip the epoch info with the SDP info and yield once per epoch in Blend, so that is nice. I don't think this PR solves the problem of epoch finality if we react to slots instead of blocks (see https://discord.com/channels/973324189794697286/1504841478768300243/1506254340736942090), so we still need to think how that would work.

ledger/src/mantle/sdp/mod.rs
youngjoon-lee: I agree that this logic is problematic, but I didn't fix it in this PR because it's a different topic. I'd say, let's fix this in a separate PR. I can prepare it with proper testing.

services/storage/src/api/backend/rocksdb/chain.rs
youngjoon-lee: I defined this associated type because we don't always fetch declarations along with blocks. We fetch only blocks, and we fetch declarations for a certain block only when we need it. So, I think it's better to separate them. 
I also don't like that we use `Bytes` everywhere, but I didn't try to fix it in this PR. Probably, on the master, I'll open a PR to use a proper type for `Block`. 
(also related to https://github.com/logos-blockchain/logos-blockchain/pull/2764#discussion_r3266677101)

core/src/sdp/mod.rs
youngjoon-lee: This is because of the storage adapter. See below https://github.com/logos-blockchain/logos-blockchain/pull/2764#discussion_r3270348033. I also don't like this approach, but in this PR, I followed the same practice as what we do for blocks.

services/chain/chain-service/src/sdp/mod.rs
youngjoon-lee: Yes. I'm working on it. This also tells me that I made a bug in this version, which will be fixed with this optimization.

youngjoon-lee: I realized that optimizing this is not very simple. So, I'd like to do it separately after merging this. Anyway, I fixed the bug that I mentioned: 995da292b8e74dd73f6ddab318f7c6f132b38831. You can see the doc comments in the code to understand what the bug was.

The summary is:

In this case which LIB is older than the last block of `cur_epoch - 2`:
```
  epoch 0     epoch 1    epoch 2    epoch 3   epoch 4
G---b1-----|---b2--b3--|--b4-----|--b5-b6---|----
                  (LIB)                       (cur_slot)

```
the snapshot for the epoch 4 should be taken at `b1` which is the last block of epoch 0.

But, my previous implementation was taking a snapshot at `b2` which is the most recent block that is older than LIB and `cur_epoch - 2`.

I fixed this bug.

core/src/sdp/mod.rs
youngjoon-lee: You're right. Done 7d361840b

services/chain/chain-service/src/lib.rs
youngjoon-lee: Thanks. Done e8d533f31

services/chain/chain-service/src/sdp/mod.rs
youngjoon-lee: I think we can use `info` here because it shows an useful info: where the snapshot was taken from. But, I'm not sure if end users may be interested in. I changed this to `debug`, anyway. a093ae038

services/chain/chain-service/src/sdp/mod.rs
youngjoon-lee: Right. Done baa18d8a9

services/chain/chain-service/src/sdp/mod.rs
ntn-x2: No, in this case, according to your diagram, the snapshot for epoch 4 becomes the state at block `b3`, which is the same snapshot used for epoch 3. The last block of epoch `N-2` is LIB, so we can take that when yielding the snapshot for epoch 3. And for epoch 4, since LIB has not changed, we re-use the same. Unless LIB was advanced during epoch 3, in which case the snapshot we yield for epoch 3 will be different, but for epoch 4 the snapshot should still be taken from `b3` because epoch 1 is "epoch-finalized" and `b3` is block-finalized.

ntn-x2 (approved):
Fair to implement other changes in a different PR. Since I've not had a chance to review this code in the past, I wanted to give some comments for parts I have not reviewed before.

ntn-x2:
@youngjoon-lee I merged `master` into the `session-removal` branch to try to keep conflicts limited to a few days worth of commits. Sorry about that. Once we merge this onto the base branch I will start working on Blend integration.

services/chain/chain-service/src/sdp/mod.rs
youngjoon-lee: @ntn-x2 
Oh, damn, yes. We're on the same page, but I made a mistake when drawing the diagram because Github editor doesn't show characters in monospace font. Sorry for confusion.

This is the correct diagram that I wanted to post. LIB is `b2` (not `b3`). In this case, `b1` should be the snapshot for epoch 4. 
```
  epoch 0     epoch 1    epoch 2    epoch 3   epoch 4
G---b1-----|---b2--b3--|--b4-----|--b5-b6---|----
             (LIB)                            (cur_slot)
```
The corresponding test is https://github.com/logos-blockchain/logos-blockchain/blob/1cb3d29afe381de54f9339fbc8ff12def79680e7/services/chain/chain-service/src/sdp/tests.rs#L166

Of course, if LIB is `b3`, it should be the snapshot for epoch 4, as you said. 
This PR implements it correctly.

youngjoon-lee:
I resolved the conflicts. Will merge this once unit tests in CI pass.

PR #2771, opened 2026-05-20
fix(test): continuous round robin fail fast

andrussal:

CI hit the suite timeout because the transaction scenario only logged per-wallet wait/send errors and kept running. This changes it to fail immediately on those errors, so CI reports the actual failure instead of timing out.

Failed run - https://github.com/logos-blockchain/logos-blockchain/actions/runs/26111224503/job/76788699327?pr=2766

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/tests/cucumber_tests/features/transactions.feature
+++ b/tests/cucumber_tests/features/transactions.feature
@@ -132,7 +132,7 @@ Feature: Transactions
       | NODE_1    | 1             | WALLET_1A   |              |
       | NODE_2    | 2             | WALLET_2A   | NODE_1       |
     When node "NODE_1" is at height 2 in 300 seconds
-    When I perform continuous transactions on user wallets with 5 coin split outputs of 100 LGO, 3 transactions of 50 LGO each for 3 cycles
+    When I perform continuous transactions on user wallets with 5 coin split outputs of 100 LGO, 3 transactions of 50 LGO each for 3 cycles and timeout of 300 seconds
     Then I stop all nodes
 
   @local_transactions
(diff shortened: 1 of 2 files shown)


hansieodendaal (commented):
Thanks for picking this up.

This test usually complete very quickly. It is intended to continue with warnings and not return an error, for bigger stress tests outside of CI. Maybe for CI we can do this
```
When I perform continuous transactions on user wallets with 5 coin split outputs of 100 LGO, 3 transactions of 50 LGO each for 2 cycles and timeout of 300 seconds
```
instead of
```
When I perform continuous transactions on user wallets with 5 coin split outputs of 100 LGO, 3 transactions of 50 LGO each for 3 cycles
```
so we add a step that time out and make the cycles less (but keep the original step for the actual stress tests)?

andrussal:
> Thanks for picking this up.
> 
> This test usually complete very quickly. It is intended to continue with warnings and not return an error, for bigger stress tests outside of CI. Maybe for CI we can do this
> 
> ```
> When I perform continuous transactions on user wallets with 5 coin split outputs of 100 LGO, 3 transactions of 50 LGO each for 2 cycles and timeout of 300 seconds
> ```
> 
> instead of
> 
> ```
> When I perform continuous transactions on user wallets with 5 coin split outputs of 100 LGO, 3 transactions of 50 LGO each for 3 cycles
> ```
> 
> so we add a step that time out and make the cycles less (but keep the original step for the actual stress tests)?

I think this is fine too, then we will get logs if it fails for some reason and we can investigate. I'll update.

hansieodendaal (approved):
LGTM

PR #2775, opened 2026-05-20
chore: silence more targets when on level >= DEBUG

ntn-x2:

OTLP logs are quite chatty because of otlp components themselves. Let's silence those if we are logging at `DEBUG` or lower, so that log entries [like this](https://vlselect.lab.vac.dev/select/vmui/#/?query=telemetry.sdk.name%3A+%22opentelemetry%22+%7C+service.name%3A+%22i-2%22&g0.range_input=25m37s367ms&g0.end_input=2026-04-29T16%3A40%3A01&g0.relative_time=none&limit=5000&displayFields=_msg%2Cscope.name) won't show up anymore.

Yes.

@ntn-x2 

N/A

No.

### DIFF
--- a/tracing/src/filter/envfilter.rs
+++ b/tracing/src/filter/envfilter.rs
@@ -12,7 +12,18 @@ const DEFAULT_DEBUG_TARGETS: &[&str] = &[
     "cryptarchia",
     "ledger",
 ];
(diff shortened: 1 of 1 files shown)


bacv:
Looks good, but we need to fix that cargo issue.

bacv (approved):

PR #2776, opened 2026-05-20
feat(zone-sdk): deposit amount + LEZ discovered bugs

pradovic:

This PR extends deposit with required amount on both indexer and sequencer. To accomplish this it consumes the new block events API. Required for LEZ bridging integration.

It also unifies `TxFinalized` and `TxsFinalized` event into one to make the API more simple, as they are treating equally by the user.

Also addresses 2 bugs found during LEZ implementation:
- back-fill should be genesis aware
- Indexer `next_messages` should not anchor over msgID, but only slot, as user could also have `deposit` and `withdraw` events and using the msgID might be ambiguous.

>Yes

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


davidrusu (approved):

zone-sdk/src/adapter.rs
```diff
@@ -186,10 +270,21 @@ fn op_to_zone_message(op: &Op, channel_id: ChannelId) -> Option<ZoneMessage> {
...
+                    metadata: deposit.metadata.clone(),
+                }))
+            } else {
+                warn!(
+                    ?tx_hash,
+                    ?op_id,
+                    "Deposit op has no matching event in block; skipping"
+                );
+                None
+            }
```
hansieodendaal, line 287: Maybe we need a different resilient strategy over here - a deposit without an amount may warrant retrieving the block and sift through it manually, or maybe we could retry event retrieval, derive the amount from another authoritative source if available, or give feedback on the deposit in a degraded/error state rather than silently dropping it.

zone-sdk/src/indexer.rs
```diff
@@ -58,23 +58,28 @@ where
...
-            .map_or_else(Slot::genesis, |(_, slot)| *slot);
-        let mut skip_until = last_zone_block;
+        let start_slot = last_slot.map_or_else(Slot::genesis, |s| s + 1);
 
         #[expect(
             closure_returning_async_block,
             reason = "Signature expected by `unfold`"
         )]
-        let stream = futures::stream::unfold(current_slot, move |current_slot| async move {
+        let stream = futures::stream::unfold(start_slot, move |current_slot| async move {
```
hansieodendaal, line 82: I think this TODO should be considered for the PR:
```
// TODO: return error to stream
```

zone-sdk/src/state.rs
```diff
@@ -42,8 +44,29 @@ pub struct AtomicWithdrawInfo {
...
 }
 
-/// A tx tracked by the SDK — either a plain inscription or an atomic
-/// inscription+withdraw bundle. Used in event payloads for
+/// A channel deposit observed in a finalized L1 block. Sequencers do not
+/// publish deposits — these are pure observations enriched with the deposit
+/// `amount` from the chain events API.
+#[derive(Debug, Clone)]
+pub struct DepositInfo {
+    /// The transaction hash containing this deposit op.
```
hansieodendaal, line 57: Would it make sense to add the running balance here?

zone-sdk/src/sequencer.rs
```diff
@@ -531,6 +538,11 @@ pub struct ZoneSequencer<Node> {
     // Incremental backfill state — processes one batch per next_event() call
     backfill_from: Option<Slot>,
     backfill_to: Option<Slot>,
+    // True until the cold-start backfill has been scheduled. Distinguishes
+    // `lib_slot == genesis()` because we haven't started yet from
+    // `lib_slot == genesis()` from a checkpoint, so the genesis slot itself
+    // gets backfilled exactly once.
```
hansieodendaal: This comment confused me:
- "because we haven't started yet from" what does this mean?
- two instances of `lib_slot == genesis()` - is this intentional?

zone-sdk/src/sequencer.rs
```diff
@@ -531,6 +538,11 @@ pub struct ZoneSequencer<Node> {
...
+    cold_start: bool,
```
hansieodendaal:
```suggestion
    backfil_from_genesis: bool,
```

zone-sdk/src/sequencer.rs
```diff
@@ -872,6 +897,11 @@ where
     /// Check whether an incremental backfill range is needed (checkpoint lib
     /// behind current network lib). Returns `false` if a backfill was set up
     /// (caller defers readiness until backfill completes).
+    ///
+    /// On cold start the genesis slot is included (`backfill_from = 0`); a
+    /// channel could be inscribed at genesis, and without this it would be
+    /// silently dropped. On warm restart the checkpoint slot has already been
+    /// processed, so we start at `from + 1`.
```
hansieodendaal: See above suggstion^ `backfil_from_genesis` - then change description here, just _backfil from genesis_ or not.

zone-sdk/src/sequencer.rs
```diff
@@ -1606,6 +1643,144 @@ where
...
+/// Inscriptions are emitted in tx order. The channel protocol guarantees a
+/// linear parent-child chain per channel within a block, so tx order already
+/// equals parent-chain order — do NOT add a topological sort here, it would
+/// mask any real protocol violation rather than fix it.
+///
+/// `bundle_withdraws` is consumed: entries are removed when an inscription's
+/// `tx_hash` matches, producing a [`PublishedTx::AtomicWithdraw`]; otherwise
+/// the inscription surfaces as a plain [`PublishedTx::Inscription`]. Deposits
+/// without a matching event entry are skipped with a warning.
+fn extract_finalized_items(
```
hansieodendaal: Why are `Op::ChannelWithdraw`not covered in this function? `fn extract_finalized_items(` ?

hansieodendaal (commented):
Looking good, and good catches on the two bugs that were fixed. 

- I found indexer coverage for cold-start including genesis, but do we have a sequencer test for the new cold_start path with channel-relevant trasnactions in slot 0?

Some furter comments/questions/suggestions below.

zone-sdk/src/adapter.rs
pradovic: A deposit op in a block "must" have a matching event by node semantics, imo (events and blocks share
  storage). A missing event means either (a) transient HTTP partial response or (b) node bug. Options:
  - Retry the events fetch once before giving up (cheap)
  - Surface as PublishedTx::Deposit with amount: None so consumer can see it in degraded form
  - Hard error / propagate (loud but breaks the stream)
  
  What do you think?

zone-sdk/src/indexer.rs
pradovic: I would do this in a separate PR as this is how the indexer was now. 
The LEZ will try to use only sequencer for everything. 
But good point, and let's address it right after as it changes the API slightly.
Does that makes sense?

zone-sdk/src/sequencer.rs
pradovic: Thanks, I missed this point I actually wanted to redefine TxsFinalized event as well to be per tx with a vec of ops inside. Will apply this

pradovic: I will also update tests to assert sequencer + indexer finalized as this is the main path now as well.

zone-sdk/src/adapter.rs
hansieodendaal: Although events and blocks share storage, they are not the same. If we have a db or data retrieval error we should continue to run if possible, after we tried to fetch the block and manually sift (preferred) or retry events (not so good), then error hard if no joy.

`PublishedTx::Deposit with amount: None` does not feel right as it may hide some other problem.

hansieodendaal: Is it possible to get the db in a state where we have a deposit with no amount? Maybe that is what needs to be fixed then.

pradovic: Two things worth confirming here:

1. The deposit amount is not encoded in the op itself — it only exists in EventPayload::Deposit.

   The op carries inputs (Vec<NoteId>) and metadata, but the actual value is derived during proof verification on the chain side and only surfaced through events. So "fetch the block and recover the amount manually" is not really possible here; the events API is effectively the source of truth.

2. @youngjoon-lee  maybe you can advise here— are blocks and block events guaranteed to become visible atomically?

   In other words: can GET /block succeed while GET /block/:id/events still returns None (or partial data) for a legitimate transient window? Or are they committed behind the same DB transaction / write barrier?

Depending on the answer to (2), the handling changes quite a bit:

- If visibility is atomic, then a deposit op without a matching event is effectively a node-side invariant violation. In that case I think the correct behavior is to log an error and fail/stop the stream — amount: None (or silently skipping) would hide a real consistency bug.

- If there is a legitimate transient race, then a retry on the events fetch would make sense before treating it as fatal.

Either way, I agree that amount: None feels like the wrong abstraction boundary here — it pushes a "did this deposit actually happen?" consistency decision onto every downstream consumer.

pradovic: I will treat them as both need to be success now or we treat it as failed. If either call fails, we return error, error is propagated, breaks the stream and already existing stream reconnect will try again (nothing advances so it will be processed again). As, for the user, I think this is the correct behavior, unless we get info that amount can be non-existent in some cases, but I think it should be present.

zone-sdk/src/state.rs
pradovic: This was my reasoning:
 - Tracking running balance in the SDK means new mutable state per channel, mutate on every deposit AND withdraw, persist it in
  SequencerCheckpoint, and handle the cold-start-from-non-genesis case where the persisted value can drift from chain truth.
  - With both deposits (with amount) and standalone withdraws now in FinalizedTx.ops, the consumer can compute the running balance from a. deterministic ordered stream themselves.
  - For an authoritative point-in-time read, channel_state is the right source.
  
  Does that makes sense?

hansieodendaal: Ah... what would be a helpful addition in future would be to get the accounting state per block - as a single deposit cannot know the totals as a block can have multiple deposits.

pradovic: I agree

pradovic:
> Looking good, and good catches on the two bugs that were fixed.
> 
>     * I found indexer coverage for cold-start including genesis, but do we have a sequencer test for the new cold_start path with channel-relevant trasnactions in slot 0?
> 
> 
> Some furter comments/questions/suggestions below.

Thanks for the review, this is the summary of change set for easier review:

  - Added FinalizedTx { tx_hash, ops: Vec<FinalizedOp> } and FinalizedOp { Inscription | Deposit | Withdraw } in state.rs.
  - Added tx_hash: TxHash to WithdrawInfo.
  - Changed Event::TxsFinalized from Vec<PublishedTx> → Vec<FinalizedTx>. Per-tx grouping with ops in execution order; atomicity is
  structural.
  - Rewrote extract_finalized_items to walk all four op types (Inscribe / ChannelConfig / Deposit / Withdraw — new), symmetric for our txs
  and others', no bundle-detection logic. Dropped bundle_withdraws_for.

  Cold-start naming + comment cleanup

  - Renamed cold_start → backfill_from_genesis.
  - Rewrote the field comment and setup_backfill_range doc to disambiguate the two lib_slot == 0 cases without the fragmented "haven't
  started yet from" phrasing.

  New unit tests (zone-sdk)

  - cold_start_backfills_genesis_slot — drives a ColdStartMockNode that serves a ChannelInscribe at slot 0; asserts it lands in
  TxsFinalized. Regression guard for the genesis-skip fix.
  - extract_finalized_items_surfaces_standalone_withdraw — non-bundled ChannelWithdraw appears as FinalizedOp::Withdraw.
  - extract_finalized_items_interleaves_deposit_then_inscription_in_same_tx rewritten to assert the new FinalizedTx/FinalizedOp shape.
  - extract_finalized_items_panics_if_deposit_amounts_incomplete — replaces the old silent-skip test, documents the new invariant.

  Error-handling refactor (per your push: drop stream, let reconnect retry)

  - fetch_block_deposit_amounts → Result<HashMap, Error>. Computes expected (tx_hash, op_id) upfront, validates events response is complete.
   Single boundary that encodes the events-atomic-with-block invariant.
  - fetch_and_process_blocks → Result<FetchedBatch, Error>. State mutation moved after the per-block fetch succeeds.
  - handle_block_event → Result<BlockEventResult, Error>. LIB backfill happens before any state advance.
  - handle_stream_item → on Err, error! + drop block stream + clear ready signal. Existing reconnect picks it up.
  - process_incremental_backfill → on Err, error! + don't advance backfill_from + sleep reconnect_delay. Next tick retries same range.
  - extract_finalized_items deposit branch → expect("fetch_block_deposit_amounts invariant") instead of if-let-else paranoia.

  E2e (cucumber) sequencer-side parity

  - New helpers in support.rs: wait_for_finalized_deposit_via_sequencer, wait_for_finalized_withdraw_via_sequencer, shared
  poll_sequencer_finalized_until. Drain Event::TxsFinalized.items, walk FinalizedTx.ops.
  - New ZoneTestError::SequencerStopped variant.
  - New cucumber steps in steps.rs:
    - sequencer "{}" finalizes deposit "{}" in {int} seconds
    - sequencer "{}" finalizes withdraw "{}" in {int} seconds
  - zone.feature — "Finalized withdraws…" scenario (the one with inscription + deposit + withdraw) now asserts both indexer AND sequencer
  paths, verifying parity.

  Downstream cascade

  - TUI: walks FinalizedTx.ops for inscriptions.
  - Cucumber start_republish_policy: walks FinalizedTx.ops for FinalizedOp::Inscription.
  - Match-exhaustiveness handled for the new variants.

  Declined / deferred (per discussion)

  - Running balance on DepositInfo — consumer can compute from the granular stream; concrete proposal evolved to per-block accounting as a
  future enhancement (mutual agreement with hansie).
  - Indexer error-to-stream TODO — deferred to a separate PR (stream item type change).
  - PublishedTx::AtomicWithdraw kept (still used by Event::Published for our own publishes) — only TxsFinalized went granular.

tests/src/cucumber/steps/manual_zone/support.rs
```diff
@@ -823,6 +829,7 @@ pub async fn wait_for_deposit(
         |message| match message {
             ZoneMessage::Deposit(deposit)
                 if deposit.inputs == expected.inputs
+                    && deposit.amount == expected_amount
```
hansieodendaal, line 832: **Nit:** 
Is this adding any value? `&& deposit.amount == expected_amount`
`if deposit.inputs == expected.inputs` should be definitive on its own otherwise we have double spends.
Not saying we should remove any of these assertions.

hansieodendaal (approved):
LGTM

tests/src/cucumber/steps/manual_zone/support.rs
pradovic: Agreed inputs is definitive. I kept the amount check because the amount comes from a separate API path (events). Felt useful as extra coverage, not as a selector.

PR #2777, opened 2026-05-21
test(e2e): use TF binary provider

andrussal:

This PR makes local e2e tests build the node binary automatically when LOGOS_BLOCKCHAIN_NODE_BIN is not set.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -53,14 +53,23 @@ local setup.
 
 ## Setup and Usage (using `cargo test`)
 
-Where tests involve spawning node binaries locally, the testing framework runs:
+Where tests involve spawning node binaries locally, the testing framework uses
+`LOGOS_BLOCKCHAIN_NODE_BIN` when it is set:
+
+```bash
+LOGOS_BLOCKCHAIN_NODE_BIN=/path/to/logos-blockchain-node cargo test ...
+```
+
+If `LOGOS_BLOCKCHAIN_NODE_BIN` is not set, local runs build the testing-featured
+release binary automatically:
 
 ```bash
 cargo build --locked --release -p logos-blockchain-node --features testing
 ```
 
-CI expects `target/release/logos-blockchain-node` to already exist, or `LOGOS_BLOCKCHAIN_NODE_BIN` to point at the
-prebuilt binary.
+CI expects `target/release/logos-blockchain-node` to already exist, or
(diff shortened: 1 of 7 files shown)


tests/testing_framework/src/framework/local/provisioning.rs
```diff
@@ -283,17 +289,49 @@ fn launch_file(relative_path: &str, contents: Vec<u8>) -> LaunchFile {
...
+                "--locked",
+                "--release",
+                "-p",
+                "logos-blockchain-node",
+                "--features",
+                "testing",
+            ]),
+            output_path: release_node_binary_path(),
+            working_dir: Some(workspace_root()),
+            lock_dir: Some(workspace_root().join("target").join(".tf-binaries")),
```
hansieodendaal, line 315: I think it should be possible to override this behaviour, for example via env var `DO_NOT_AUTO_BUILD_NOT_BINARY`

tests/README.md
```diff
@@ -55,9 +53,14 @@ local setup.
 
 ## Setup and Usage (using `cargo test`)
 
-Where tests involve spawning node binaries, preference will be given to binaries corresponding to `USE_DEBUG_BINARIES` 
-and `USE_RELEASE_BINARIES` environment variables in the in `target/debug` and in `target/debug` paths respectively. If 
-neither are defined, preference will be given to debug binaries.
+Where tests involve spawning node binaries locally, the testing framework runs:
+
+```bash
+cargo build --locked --release -p logos-blockchain-node --features testing
```
hansieodendaal: and we could explain the override option here in the docs

hansieodendaal (commented):
I think this is a move in the right direction, but it should not be an automatic build, we need an _opt out_ behaviour path as well please.

tests/testing_framework/src/framework/local/provisioning.rs
andrussal: Actually default behaviour is to check first `LOGOS_BLOCKCHAIN_NODE_BIN` env var. I think this is enough for most workflows because you can always set this to point to a binary in target, and whenever needed, run cargo build to update it.

tests/README.md
andrussal: > and we could explain the override option here in the docs

Good point.

hansieodendaal (approved):
LGTM

PR #2778, opened 2026-05-21
refactor(logging): migrate log targets

andrussal:

Migrate `api`, `storage`, `system_sig`, `time`, and `tx_service` to the log targets crate.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


tracing/targets/src/api.rs
```diff
@@ -0,0 +1,7 @@
+use lb_log_targets_macros::log_targets;
+
+log_targets! {
+    root = api;
+
+    http::{MANTLE},
```
ntn-x2, line 6: Do we need the `HTTP` prefix? Just wondering if we expose anything else under the `API` namespace?

tracing/targets/src/time.rs
```diff
@@ -0,0 +1,7 @@
+use lb_log_targets_macros::log_targets;
+
+log_targets! {
+    root = time;
+
+    ntp::{TESTUTILS},
```
ntn-x2: We should probably not bubble up test-only targets at this level, what do you think?

ntn-x2: Especially if only used for unit tests. For e2e tests, we could probably discuss case by case.

tracing/targets/src/tx_service.rs
```diff
@@ -0,0 +1,16 @@
+use lb_log_targets_macros::log_targets;
+
+log_targets! {
```
ntn-x2: I think this naming could be restructured. This is all mempool related, so I would personally use a `mempool` top level definition, then derive `service`, `storage`, `network::libp2p`, `network::mock`.

andrussal: At some point we renamed this service to `tx_service` from mempool because we added storage and more functionality. But I think you are right, it's all mempool related anyways.

tracing/targets/src/time.rs
andrussal: > We should probably not bubble up test-only targets at this level, what do you think?

Makes sense, it should have only public modules.

tracing/targets/src/api.rs
andrussal: > Do we need the `HTTP` prefix? Just wondering if we expose anything else under the `API` namespace?

We potentially could add other protocols in future, but I don't have strong preference. Maybe it is better to keep http just for more clarity.

ntn-x2 (approved):
Nice!

PR #2779, opened 2026-05-21
chore(release): Deployment stateless user configs

bacv:

Sets predefined shareholder and provider values for the init nodes.
When deploying files in `testnet/ceremony/genesis/devnet` should stay the same between releases, even when a new genesis is generated.

Updated CI workflow to automatically overwrite protocol string, otherwise when releasing a manual commit would be required.

Fixed blockchain-tools `run_ceremony` command to update the `faucet_pk`, otherwise nodes doesn't exclude faucet from stake value and ledger overflows.

Genesis block generation during release. 

@danielSanchezQ @bacv 

N/A

N/A

## Checklist

* [x] 2. Description added.


danielSanchezQ (approved):
Black magic

tools/blockchain-tools/src/bin/genesis.rs
```diff
@@ -382,6 +386,21 @@ fn wrap_as_cryptarchia_genesis_block(block_value: Value) -> Value {
     Value::Mapping(outer)
 }
 
+/// Wrap a serialised faucet public key in the mapping that corresponds to
+/// `cryptarchia.faucet_pk` in a deployment config.
+fn wrap_as_cryptarchia_faucet_pk(faucet_pk: Value) -> Value {
+    let mut inner = serde_yml::Mapping::new();
+    inner.insert(Value::String("faucet_pk".to_owned()), faucet_pk);
```
ntn-x2, line 393: Not sure how easy it would be to make sure this is always in sync with the expected key? What if we change `faucet_pk` to be `faucet_public_key` tomorrow? Not saying it should be fixed here, but this will be hard to catch if we change something in the deployment config.

.github/workflows/genesis-ceremony.yml
```diff
@@ -32,14 +32,23 @@ jobs:
...
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Inject Variables into Template
+        run: |
+          cp testnet/ceremony/genesis/$LB_TARGET_ENV/deployment-template.yaml \
+            /tmp/deployment-generated.yaml
+
+          # We use '|' as a delimiter so that slashes in paths don't break the command
+          sed -i "s|ENV_PLACEHOLDER|$LB_TARGET_ENV|g" /tmp/deployment-generated.yaml
+          sed -i "s|VERSION_PLACEHOLDER|$LB_VERSION|g" /tmp/deployment-generated.yaml
```
ntn-x2, line 42: I guess the only case this does not include is if we release a new version but we don't want to change the protocol number. I agree this probably deserves a wider discussion as to whether a protocol number is the right thing to put here, or if we should rather include the genesis tx in there, so we can release multiple versions for the same network without forking the overlay. Also, if we want to support forks, we should have a fork ID or a chain ID of some sort. Just speculating, but it's something we should think about.

ntn-x2 (approved):

tools/blockchain-tools/src/bin/genesis.rs
bacv: > Not sure how easy it would be to make sure this is always in sync

We already discussed this with @danielSanchezQ, once the flow is working, we'll refactor this part not to depend on serializer when manipulating the config. All required parts should be deserialized to structs, and operations should manipulate the structs not yaml values.

.github/workflows/genesis-ceremony.yml
bacv: > I guess the only case this does not include is if we release a new version but we don't want to change the protocol number.

This part could grow and be updated once we have such functionality.

nodes/node/binary/src/config/deployment/devnet/deployment.yaml
```diff
@@ -54,42 +54,67 @@ cryptarchia:
...
+            - value: 1000000
+              pk: '7dfab2c05ee11439587776a9704a86b166d7802606c01505719fc27af1bc7b1e'
+            - value: 1000000
+              pk: dbfde040040915600148de9ef99aca409b0707b012f23ced36a361d32ec59b17
+            - value: 1000000
+              pk: bd6dd65543fdabb10e155c1ae982ecf6e9ed02fa58140a0e5249f52aac66cb17
+            - value: 1000000
+              pk: '3bd893ccdf4f4276d9eeeb0b65ba5817082a001adc84206b9705c10fe293b001'
+            - value: 18446744073709551615
+              pk: '073d876844fe35923cc0a3e43699b96fc04a0389d383c329e215f628a1bffa09'
```
hansieodendaal, line 70: What is the maximum token value in the blockchain? If it is `u64::MAX` we cannot have all these notes...

hansieodendaal (commented):
Looks good, just one question I was wondering about.

nodes/node/binary/src/config/deployment/devnet/deployment.yaml
bacv: >  If it is u64::MAX we cannot have all these notes.

This is configurable amount, not sure about the max value in spec but the type we use in the node indeed is limited to u64.

hansieodendaal: So the thought experiment here is, how many tokens can we allocate in the genesis block and would it invalidate any token supply validation check? When we are only bounded by the type max, we could have many `u64::MAX` token allocations in the genesis block.

bacv: > When we are only bounded by the type max

This is something to discuss with token economics. We might already have a decision written somewhere, in that case we'll update this value accordingly. Will leave u64::MAX for devnet environment.

hansieodendaal (commented):
LGTM

bacv:
The stakeholder and provider list is tested on the devnet, genesis ceremony generates working deployment settings with a correct genesis block. The keystore files for init nodes will be uploaded to internal notion page.

PR #2781, opened 2026-05-21
chore: implement static bound checks for the ChannelConfig op

ntn-x2:

Continuing with adding SDP encoding/decoding requirements into the types, this time we migrate the `ChannelConfig` op. As per the spec, the number of keys in a channel config has to be included in the range `[1, u16::MAX]`. We can model that with a `BoundedVec` which for the occasion has been updated to support a minimum number of elements as well. In the code the max number of keys used to be `u8::MAX`, but we found out it was an inconsistency as the spec defines it as `u16::MAX` instead.

The same module also exposes auxiliary type aliases like `UpperBoundedVec<T, MAX>`, `LowerBoundedVec<T, MIN>` and `NonEmptyBoundedVec<T, MAX>` which enforce, respectively, that a vector length is in the range `[0, MAX]`, `[MIN, usize::MAX]`, and `[1, MAX]`. The latter type is currently used for the list of locators in an SDP declare, although once we get to that we need to check if the spec enforces an upper limit, and if it does not, enforce it, and then change the type of `Locators` to be a `NonEmptyBoundedVec<Locator, <max_as_per_spec>>`.

The only downside, at the moment, is that we cannot provide an API specifically for non-empty bounded vectors (with a `MIN` length of at least `1`) due to current Rust generics limitations. Once that is unlocked, we can do much more with this type.

Yes.

@ntn-x2 @thomaslavaur 

Yes, it was fixed.

There were minor inconsistencies which were promptly fixed.


core/src/mantle/channel.rs
```diff
@@ -236,10 +271,10 @@ mod tests {
...
             balance: 0,
             withdrawal_nonce: 0,
-            accredited_keys: (0..num_keys as u8)
-                .map(test_public_key)
-                .collect::<Vec<_>>()
-                .into(),
+            accredited_keys: Keys::new_unchecked(
+                (0..num_keys).map(test_public_key).collect::<Vec<_>>(),
+            )
+            .into(),
```
hansieodendaal: I think we should use `try_from` here - if a call site specifies `num_keys = 0` it should panic.

ledger/src/lib.rs
```diff
@@ -1370,45 +1370,6 @@ mod tests {
         ));
     }
 
-    #[test]
```
hansieodendaal, line 1373: You rmoved ` fn test_empty_keys_error() {`, but we could still test 
- Keys::try_from(vec![]) fails
- serde deserialization of empty Keys fails
- ChannelConfigOp::decode rejects zero key count

core/src/sdp/mod.rs
```diff
@@ -48,65 +45,9 @@ impl ServiceParameters {
...
-    type Item = &'a Locator;
-    type IntoIter = Chain<Once<&'a Locator>, slice::Iter<'a, Locator>>;
-
-    fn into_iter(self) -> Self::IntoIter {
-        once(&self.0).chain(self.1.iter())
-    }
-}
+// TODO: Check spec for max limit once we migrate the SDP Declare op to use the
+// `NomEncode` and `NomDecode` traits.
+pub type Locators = LowerBoundedVec<Locator, 1>;
```
hansieodendaal, line 50: Why not this? Then we retain the semantics, and rather put a TODO here to come back and change it when `LowerBoundedVec::first()` is sorted out
```suggestion
pub struct Locators(Locator, UpperBoundedVec<Locator, { MAX - 1 }>);
```
**Note:** I am sure there must be an upper limit to locators already in the spec?

hansieodendaal (commented):
I like this demonstration of applying bounded vecs, nicely done. Some comments to consider below please.

core/src/mantle/channel.rs
ntn-x2: This is a test, so we know exactly the input. Also, if `0` is passed, there will be no loop. But yeah, in this case I agree we can `expect`.

ledger/src/lib.rs
ntn-x2: These tests all semantically belong to the `BoundedVec` type. And I agree, we need to test all those cases, but I will do it in a separate PR, if that's ok.

ntn-x2: Minus the last one, that I could add here.

ntn-x2: Added the test here: https://github.com/logos-blockchain/logos-blockchain/pull/2781/commits/35ce07996be8c59035fcc67168a32ff6af45e8ef. Will add tests to the `BoundedVec` type in a follow-up PR.

core/src/sdp/mod.rs
ntn-x2: Yes there is, but I have not migrated the SDP declare op to the new format yet, so I did not want to touch that bit. As for the new type, I'd rather have that under `BoundedVec` itself, so once Rust has full support for const generics, we can expose a unified API there. Otherwise we would need to define this type everywhere we have a non-empty collection of items. Initially I defined a `NonEmptyBoundedVec` in this PR as something similar to this, but more generic, but realized that we had a lot of code duplication, so I opted for having a single `BoundedVec` instance and a bunch of const assertions in code that are triggered on `cargo build` or `cargo test`. So I think that is the right place to define types such as this one.

hansieodendaal (approved):
LGTM

danielSanchezQ (approved):

PR #2785, opened 2026-05-22
fix(sdp-ledger): advance session only when new LIB crosses border.

youngjoon-lee:

This PR fixes a bug that @ntn-x2 [found](https://discord.com/channels/973324189794697286/1506558022481346590/1506950711899066409).

### Bug
Chain service is yielding a new session only when the new LIB enters the new session. This behavior is correct because we want the blend service to always have the irreversible view.

However, SDP ledger was transitioning sessions as soon as a session boundary block (1st block of the new session) is applied. That is, it causes a time gap.

This time gap caused an issue. Here is the scenario:
- The **SRP** ledger in the SDP ledger built an immutable `TargetSessionState` as soon as the session boundary block is applied (session 26->27).
- But, the blend service didn't know that yet because the chain service didn't yield the session 27 yet (since LIB is still in session 26).
- Unfortunately, before LIB enters the session 27, the new **epoch** 27 was started. This new epoch was yielded immediately from the chain leader service to the blend service. 
- So, blend service began to create blend messages with PoQs generated against the epoch state 27.
- After that, the session info 27 arrived in the blend service, in the end. So, blend service submitted an activity proof to the network.
- But, unfortunately, the activity proof was from the message generated in the epoch 27 (not 26). This proof was rejected by SDP ledger because its `TargetSessionState` didn't have the proof verifier for epoch 27. It had only the epoch 26 verifier because it built the `TargetSessionState` earlier because of the time gap.

### Solution

The solution is removing the time gap.

The behavior of chain service (yielding sessions) is correct. So, this PR changes the way how SDP ledger transitions sessions. In short, now SDP ledger transitions the session only when the new LIB crosses the session border.

Previously, SDP ledger had the following algorithm:
```rust
let current_session = service_params.session_for_block(new_block_number);
if current_session == self.active.session_n + 1 {
  // transition to the new session
```
This PR changes the way how `current_session` is calculated.
```rust
let current_session = service_params.current_session(new_lib);
if current_session == self.active.session_n + 1 {
  // transition to the new session
```
The current session must be always calculated based on the new LIB.

For example, if the new LIB didn't cross the border yet, SDP ledger shouldn't transition the session.
```
// session_duration = 10 blocks
-------b9--|--b10---b11---
        ^            ^
        |            |
       LIB        new_block
// current_session = 0
```
Only when the new LIB crosses the border, SDP ledger should transition the session.
```
// session_duration = 10 blocks
------b9--|--b10---b11---b12---
              ^           ^
              |           |
             LIB       new_block
// current_session = 1
```

I believe so.

@youngjoon-lee 
spec: @madxor 

we should mention this behavior if there's nothing mentioned about this.

we should mention this behavior if there's nothing mentioned about this.

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/ledger/src/mantle/sdp/mod.rs
+++ b/ledger/src/mantle/sdp/mod.rs
@@ -24,9 +24,11 @@ use lb_core::{
 };
 use lb_key_management_system_keys::keys::{Ed25519Signature, ZkSignature};
 use rewards::{Error as RewardsError, Rewards};
-use tracing::warn;
+use tracing::{debug, warn};
 
 use crate::{EpochState, UtxoTree, mantle::sdp::rewards::blend};
+
+const LOG_TARGET: &str = "ledger::mantle::sdp";
 
 type Declarations = rpds::RedBlackTreeMapSync<DeclarationId, Declaration>;
 
@@ -40,6 +42,8 @@ impl Service {
         self,
         block_number: BlockNumber,
         epoch_state: &EpochState,
+        // LIB after the given block is applied to Cryptarchia
+        lib: BlockNumber,
         config: &ServiceParameters,
         rewards_params: &ServiceRewardsParameters,
     ) -> (Self, Vec<Utxo>) {
@@ -48,6 +52,7 @@ impl Service {
                 let (new_state, utxos) = state.try_apply_header(
                     block_number,
                     epoch_state,
+                    lib,
                     config,
                     &rewards_params.blend,
                 );
@@ -212,14 +217,23 @@ impl<R: Rewards> ServiceState<R> {
         mut self,
         block_number: u64,
         epoch_state: &EpochState,
+        // LIB after the given block is applied to Cryptarchia
+        lib: BlockNumber,
         service_params: &ServiceParameters,
         rewards_params: &R::Params,
     ) -> (Self, Vec<Utxo>) {
-        let current_session = service_params.session_for_block(block_number);
+        let current_session = service_params.current_session(lib);
         let reward_utxos;
 
-        // shift all session!
         if current_session == self.active.session_n + 1 {
+            debug!(
+                target: LOG_TARGET,
+                current_session,
+                previous_sessoin = self.active.session_n,
+                lib_height = lib,
+                "Session transition detected. Removing inactive declarations, commiting snapshot, and updating rewards for the new session",
+            );
+
             // Remove expired declarations based on retention_period
             // This essentially duplicates the declaration set so it's only triggered at
             // session boundaries
@@ -359,6 +373,8 @@ impl SdpLedger {
         &self,
         config: &Config,
         epoch_state: &EpochState,
+        // LIB after the given block is applied to Cryptarchia
+        lib: BlockNumber,
     ) -> Result<(Self, Vec<Utxo>), Error> {
         let block_number = self.block_number + 1; // overflow?
         let mut all_reward_utxos = Vec::new();
@@ -374,6 +390,7 @@ impl SdpLedger {
                 let (new_state, reward_utxos) = service_state.clone().try_apply_header(
                     block_number,
                     epoch_state,
+                    lib,
                     service_params,
                     &config.service_rewards_params,
                 );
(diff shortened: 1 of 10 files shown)


ledger/src/mantle/sdp/mod.rs
```diff
@@ -24,10 +24,12 @@ use lb_core::{
 };
 use lb_key_management_system_keys::keys::{Ed25519Signature, ZkSignature};
 use rewards::{Error as RewardsError, Rewards};
-use tracing::warn;
+use tracing::{debug, warn};
 
 use crate::{EpochState, UtxoTree, mantle::sdp::rewards::blend};
 
+const LOG_TARGET: &str = "ledger::mantle::sdp";
```
ntn-x2, line 31: We have a crate for log target now, although we have not migrated ledger stuff yet, so I guess it's fine to have it here for now.

services/chain/chain-leader/src/lib.rs
```diff
@@ -445,6 +445,9 @@ where
                         winning_pol_slot_notifier.process_epoch(&eligible, latest_tree, &epoch_state, &kms_api).await;
 
                        if let Some((proof, signing_key)) = build_proof_for(&eligible, latest_tree, &epoch_state, slot, &winning_pol_slot_notifier, &wallet_api, &kms_api).await {
+                            // Simulate the new LIB height by +1, because we already propose a block on top of the tip,
```
ntn-x2, line 448: We don't apply the block automatically, so we don't know if by the time we receive this block, we will append it to the current tip tho. But I guess this is the best we can do when trying to apply the block locally before proposing?

ntn-x2: Also, this is not necessarily true for the early stages of the network, where the LIB stays at `0` until we have produces `k` blocks.

ledger/src/mantle/sdp/mod.rs
```diff
@@ -212,14 +217,23 @@ impl<R: Rewards> ServiceState<R> {
...
+        lib: BlockNumber,
         service_params: &ServiceParameters,
         rewards_params: &R::Params,
     ) -> (Self, Vec<Utxo>) {
-        let current_session = service_params.session_for_block(block_number);
+        let current_session = service_params.current_session(lib);
         let reward_utxos;
 
-        // shift all session!
         if current_session == self.active.session_n + 1 {
```
ntn-x2, line 228: I think this deserves more attention. We do not know if we are on the canonical tip or not, and we are comparing the session number of (potentially) a fork with the session number of the LIB state. If we apply the block on a fork after the LIB has rotated session (`current_session`), we would trigger the session transition logic again, and that logic is dependent on the timing `k` was advanced. In short, the consensus is not entirely based on current state + parent tip, but it's now dependent on LIB, which can have different timings for different nodes, yielding to non-deterministic consensus rules.

ntn-x2 (commented):
I am quite sure the bit of logic where we compare the local state with LIB needs more attention, because building on a fork while the LIB changes could have unintended consequences, once and if that forks becomes canonical.

services/chain/chain-leader/src/lib.rs
youngjoon-lee: > Also, this is not necessarily true for the early stages of the network, where the LIB stays at `0` until we have produces `k` blocks.

Hm, yes, my code is wrong. I just thought about the basic case where LIB is higher than the genesis, because the chain leader service starts its work only after the chain switches from Bootstrapping to Online mode. But, if PBP is very short, and if there were not many blocks added when the network is bootstrapped from scratch, yes, you're right. LIB may stay at genesis. So, adding a new block to the honest chain doesn't advance LIB by 1. It may stay at the genesis.

So, yeah, I'll think about any better way...

> I agree that this is the simplest model we can consider. I personally want to make this approach feasible, since I feel that some logics have became complicated because of the finality + slot-vs-block issue.

Actually, this logic is not for applying a new block automatically. When the chain leader service assembles a new block, it tries to pick only "valid" transactions from the mempool. The way to check their validity is applying them to the cloned (parent) ledger. So, we still need this logic anyway.

ledger/src/mantle/sdp/mod.rs
youngjoon-lee: Hm, yes, LIB is an external information that is not managed by the consensus+ledger. So, it's not safe to rely on it when updating the ledger. Let's factor out the snapshot-related (LIB-related) parts from the ledger as we discussed in the call.

PR #2786, opened 2026-05-22
fix(blend): SDP snapshot in EpochState

davidrusu:

Previously we would get our epoch info from two different sources:
1. SDP membership information was pushed by chain-service through chain-broadcast-service to Blend when a block finalizes
2. PoL information was pulled from blend on epoch tick boundaries from chain-services `get_epoch_state()` api.

These two sources of information happen on their own timeline will (and has) lead to cases where these two bits of information are out of sync (POL may update but SDP membership still has not been pushed to blend)

This change moves the SDP snapshot into the EpochState struct so that we fetch everything together at once.

To minimize the diff, I have not removed the old code-path where ActiveProviders is pushed from chain-service -> broadcast-service -> blend. We can do this cleanup in a follow up PR.

But, critically, all active code paths in blend are now using the SDP state from EpochState. and hence guaranteed to be consistent.

yes

> [!IMPORTANT]
    modification/solutions.

## Checklist

* [x] 2. Description added.

### DIFF
--- a/blend/scheduling/src/session.rs
+++ b/blend/scheduling/src/session.rs
@@ -20,13 +20,9 @@ pub struct UninitializedSessionEventStream<Stream> {
 
 impl<Stream> UninitializedSessionEventStream<Stream> {
     #[must_use]
-    pub const fn new(
-        session_stream: Stream,
-        first_ready_timeout: Duration,
-        transition_period: Duration,
-    ) -> Self {
+    pub const fn new(session_stream: Stream, transition_period: Duration) -> Self {
         Self {
-            stream: UninitializedFirstReadyStream::new(session_stream, first_ready_timeout),
+            stream: UninitializedFirstReadyStream::new(session_stream),
             transition_period,
         }
     }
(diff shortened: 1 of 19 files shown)


ntn-x2 (commented):
~As I said in the Discord thread, I personally don't think this is solving an actual issue, and also this PR is based on `master` which at the moment has sessions baked in, and this solution completely bypasses sessions. The session-removal PR will have a similar behaviour that you are suggesting in this PR, but in an "epoch-native" context.~

Sorry I commented before having my coffee! I noticed the PR **is** based on the session removal branch, so that's good. The problem of non finality still stands. While cryptarchia is resilient to forks, Blend is not, so if we want to go ahead with this, we must be aware that there are edge cases in which Blend will break.

ntn-x2 (approved):
Based on our conversation, this looks like the right approach, so let's go!

youngjoon-lee (approved):
thank you for confirming this approach!

youngjoon-lee:
@davidrusu I can also take over the TODO you left: Moving the sdp ledger to the same level where cryptarchia ledger and mantle ledger live.

davidrusu:
@youngjoon-lee that would be great, thank you

PR #2787, opened 2026-05-22
ci: nuke codecov tests

siddarthkay:
## Summary

As discussed with @bacv for now we nuke these tests and they will discuss/decide about integrating codecov back into CI later.


bacv (approved):

bacv:
Good to go without waiting for CI

PR #2790, opened 2026-05-22
chore: static bounds for ChannelDeposit op

ntn-x2:

As per title.

Yes.

@ntn-x2 

Yes.

No.


core/src/mantle/ledger.rs
```diff
@@ -150,18 +155,55 @@ impl<'output> IntoIterator for &'output Outputs {
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl DerefMut for Inputs {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
```
danielSanchezQ: Uhm, Deref is usually better (and widely used) for kinda pointers. I would say its better to to use explicit ref and ref_mut calls, so we know there was a change of type.

core/src/mantle/tx_builder.rs
```diff
@@ -425,8 +428,8 @@ mod tests {
         let builder = MantleTxBuilder::new(context)
             .push_op(Op::ChannelDeposit(DepositOp {
                 channel_id: [0; 32].into(),
-                inputs: Inputs::new(vec![deposit_input]),
-                metadata: vec![],
+                inputs: deposit_input.into(),
+                metadata: [].into(),
```
danielSanchezQ: One of the drawbacks is that is more difficult to know what goes there because we use `into` everywhere. Its true that its also more flexible to change stuff if we dont break the from api.

ledger/src/lib.rs
```diff
@@ -870,15 +870,15 @@ mod tests {
         let sk = ZkKey::from(BigUint::from(0u8));
         // determine fees
         let tx = create_tx(
-            vec![utxo.id()],
+            utxo.id().into(),
```
danielSanchezQ: This is highly confusing too. I think its easier to see that we are using a vector.

danielSanchezQ (changes requested):
The into's may be ok. But the `Deref` I think we should change it.

core/src/mantle/ledger.rs
ntn-x2: Yeah I kind of agree with this in general. But I was surprised in the first place to see that is was a newtype. Anyway, `AsRef` and `AsMut` introduce an additional `.as_ref()` and `.as_mut()` which make things even more obscure, in my opinion. We already implement 'AsRef<[T]>`, so I think in this case it would only add to the confusion. Also, I can't think of a case in which we don't want to use methods of the underlying bounded. This was my reasoning. If you think we should still revert this impl, fine by me.

core/src/mantle/tx_builder.rs
ntn-x2: Yeah here also I don't have a strong opinion.

ledger/src/lib.rs
ntn-x2: It's a collection that can be initialized with a single element as well, as long as its minimum size is not `0`. But I can add a surrounding pair of `[]` to it, no problem.
We can use an array instead of a vector tho, since we know it fits for sure within the range bounds.

core/src/mantle/ledger.rs
danielSanchezQ: For me its better to have an explicit call, with as_ref and as_mut at least we know there is something below that is what we need, otherwise with deref/mut it just goes through from one type to the other.

ledger/src/lib.rs
danielSanchezQ: Ah, yes, `[utxo.id()]` is much better thanks!

hansieodendaal (commented):
Please see alternative implementation for `Ledger::Inputs` in #2791.

core/src/mantle/ledger.rs
ntn-x2: Removed `Deref` here: https://github.com/logos-blockchain/logos-blockchain/pull/2790/commits/0724f0aa3afc64ffbf094c190054f16438938b4e. I though we were using more of the underlying type methods, but we weren't, so the removal was not painful.

ledger/src/lib.rs
ntn-x2: https://github.com/logos-blockchain/logos-blockchain/pull/2790/commits/403867b35dabd90ea529bfb93df4a2bbb6ab6519

core/src/mantle/tx_builder.rs
ntn-x2: I added an `empty()` constructor, with a const assertion that this cannot be called when minimum length is greater than `1`: https://github.com/logos-blockchain/logos-blockchain/pull/2790/commits/ae2ed86c06de66c5b5e57477426c81ce9ff436f9.

hansieodendaal (approved):
We have to read this PR in conjunction with #2791.

I like the ChannelDeposit-specific direction here, especially the dedicated Metadata type and moving deposit encode/decode onto DepositOp.

My main concern is that the PR now also introduces a broader bounded ledger::Inputs refactor, which overlaps heavily with #2791. I think the final shape should probably use the shared TransferInputs / TransferOutputs aliases from #2791 and keep builder paths fallible rather than expecting on bound violations. #2791’s version is cleaner because it gives the bound a semantic name and reuses it everywhere:

If we rebase #2791 on top of this, I’d suggest retaining the deposit-specific pieces from this PR, but preferring #2791’s central bounded type aliases and error propagation for the generalized input/output plumbing.

Happy to rebase.

ntn-x2:
@hansieodendaal this PR is only concerned with the `ChannelDeposit` op itself, hence I simply changed all occurrences of `Vec` into its respective `BoundedVec` according to the spec. I did not even remove the new type, since this is meant to be as little breaking of a change as possible. Nevertheless, of course we can have a wider discussion around whether we want to factor into a common type inputs/outputs for multiple ops, but as far as this PR is concerned, it only changes `Vec` to `BoundedVec`. So the PR you opened is a better place for this discussion to happen.

danielSanchezQ (approved):

PR #2791, opened 2026-05-24
chore: add bounded_vec to ledger tx builder inputs and outputs

hansieodendaal:

Added `bounded_vec` to ledger tx builder inputs and outputs.

The main change in this PR was to define the types
```rust
const MAX_TRANSACTION_INPUTS: usize = u8::MAX as usize;
const MAX_TRANSACTION_OUTPUTS: usize = u8::MAX as usize;
pub type BoundedUtxos = UpperBoundedVec<Utxo, MAX_TRANSACTION_INPUTS>;
pub type BoundedInputs = UpperBoundedVec<NoteId, MAX_TRANSACTION_INPUTS>;
pub type BoundedOutputs = UpperBoundedVec<Note, MAX_TRANSACTION_OUTPUTS>;
```
and use them throughout.

Yes

@hansieodendaal @ntn-x2 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


core/src/block/genesis.rs
```diff
@@ -197,7 +248,9 @@ impl GenesisBlockBuilder<Empty> {
     #[must_use]
     pub fn add_note(self, note: Note) -> GenesisBlockBuilder<WithNotes> {
         GenesisBlockBuilder {
-            state: WithNotes { notes: vec![note] },
+            state: WithNotes {
+                notes: TransferOutputs::new_unchecked(vec![note]),
```
ntn-x2: No need to do this. `BoundedVec` implements a non-fallible `From` for both a single element and a statically-sized array. So this can be `[note].into()` or even `note.into()`. If you want to make it explicit, you can use `TransferOutputs::from()`. In short, no need to use a `vec!` input if we know the length of the input. Applies to every other instance.

core/src/block/genesis.rs
```diff
@@ -1339,7 +1352,9 @@ mod tests {
         let block = GenesisBlockBuilder::new()
             .add_note(make_note(100))
```
ntn-x2: I think this can honestly be replaced with an infallible alternative: `[make_note(100), make_note(200), make_note(300)].into()`.

core/src/mantle/ops/transfer.rs
```diff
@@ -116,12 +116,16 @@ mod test {
         let pk1 = ZkPublicKey::from(Fr::from(BigUint::from(1u8)));
         let pk2 = ZkPublicKey::from(Fr::from(BigUint::from(2u8)));
         let transfer = TransferOp {
-            inputs: Inputs::new(vec![NoteId(BigUint::from(0u8).into())]),
-            outputs: Outputs::new(vec![
-                Note::new(100, pk0),
-                Note::new(200, pk1),
-                Note::new(300, pk2),
-            ]),
+            inputs: Inputs::new(vec![NoteId(BigUint::from(0u8).into())].try_into().unwrap()),
```
ntn-x2: Same as elsewhere. This does not need to be a failing op with a vec, it can be an infallible one with an array or even just a single element. `[NoteId(BigUint::from(0u8).into())].into()` or the more explicit `Inputs::from([...])` if you wish.

core/src/mantle/channel.rs
```diff
@@ -383,7 +383,7 @@ mod tests {
 
         let deposit_op = DepositOp {
             channel_id,
-            inputs: Inputs::new(vec![utxo.id()]),
+            inputs: Inputs::new(vec![utxo.id()].try_into().unwrap()),
```
ntn-x2: Same as above, and same for all next occurrences.

core/src/mantle/encoding.rs
```diff
@@ -244,16 +249,27 @@ fn decode_note(input: &[u8]) -> IResult<&[u8], Note> {
 fn decode_inputs(input: &[u8]) -> IResult<&[u8], Inputs> {
     // Inputs = InputCount *NoteId
     let (input, input_count) = decode_byte(input)?;
+    if input_count as usize > TransferInputs::MAX {
```
ntn-x2: This is gone in my PR, since this logic is already part of `NomBoundedVec` implementation. So I'd suggest you either wait for my PR to be merged, or you rebase this on top of that.

utils/src/bounded_vec.rs
```diff
@@ -85,6 +85,12 @@ impl<T, const MIN: usize, const MAX: usize> BoundedVec<T, MIN, MAX> {
     }
 }
 
+impl<T, const MIN: usize, const MAX: usize> Default for BoundedVec<T, MIN, MAX> {
+    fn default() -> Self {
```
ntn-x2, line 95: This does not work. If `MIN` is greater than `0`, we can't have an empty inner vec, which is why I removed the `Default` impl for `BoundedVec`. We could add it by having a const assertion that checks that `MIN == 0`, perhaps.

ntn-x2 (changes requested):
I think we can rely on the type semantics more. Almost all the places which try to fallibly create inputs from a vec can be changed to use infallible arrays instead, which is one of the major benefits of this new approach. As the `Default` implementation for `BoundedVec`, that also needs to change or needs to be gated by a const assertion. In general, I think this PR should be rebased on top of the ChannelTransfer PR, and perhaps suggest improvements on that, where applicable.

core/src/mantle/encoding.rs
hansieodendaal: Yeah, we can just follow the normal merge pattern here, everwhich way it is merged.

ntn-x2:
@hansieodendaal not sure what was the reason to have a newtype for inputs and outputs. I would be in favor of removing it and use the underlying `BoundedVec`, directly, but maybe the reason for the newtype was to avoid confusing inputs with outputs. @danielSanchezQ any insights/feedback on this? I would not mind if this PR rebases on mine and updates it with removing the newtype wrapper.

danielSanchezQ:
> @hansieodendaal not sure what was the reason to have a newtype for inputs and outputs. I would be in favor of removing it and use the underlying `BoundedVec`, directly, but maybe the reason for the newtype was to avoid confusing inputs with outputs. @danielSanchezQ any insights/feedback on this? I would not mind if this PR rebases on mine and updates it with removing the newtype wrapper.

We created Inputs/Outputs because we had duplicated code for calculation the same stuff that now is encapsulated in those types. Why remove them? You can change the underliying type to the Bounded versions of the vector right? I think that would be the way to go unless I have missed something.

ntn-x2:
> We created Inputs/Outputs because we had duplicated code for calculation the same stuff that now is encapsulated in those types. Why remove them? You can change the underliying type to the Bounded versions of the vector right? I think that would be the way to go unless I have missed something.

Yes, the underlying type in my PR is already a bounded vec, so we reap the benefits of that. So yeah let's keep the newtype alive @hansieodendaal. Maybe you can revert that change in your PR, unless you have a good point for getting rid of them?

hansieodendaal:
Are you referring to these? That was the crux pf the PR.
```rust
pub type LedgerInputs = UpperBoundedVec<Utxo, MAX_TRANSACTION_INPUTS>;
pub type TransferInputs = UpperBoundedVec<NoteId, MAX_TRANSACTION_INPUTS>;
pub type TransferOutputs = UpperBoundedVec<Note, MAX_TRANSACTION_OUTPUTS>;
```

There is a slight change in approach between this PR and #2790 about how to treat the inner types for `Inputs`. I can rebase this one on top of that one when merged then we will be able to assess.

core/src/mantle/ledger.rs
```diff
@@ -63,14 +66,19 @@ pub enum LedgerError {
 }
 
 #[derive(Clone, Eq, Debug, PartialEq, Serialize, Deserialize)]
-pub struct Outputs(Vec<Note>);
+pub struct Outputs(TransferOutputs);
```
danielSanchezQ: I think it would be better to use the proper bounded vec here instead of the alias, precisly because is obscuring the layer on top (`Outputs`);

core/src/mantle/ledger.rs
```diff
@@ -124,44 +132,44 @@ impl Outputs {
...
 
     fn into_iter(self) -> Self::IntoIter {
-        self.0.iter()
+        (&self.0).into_iter()
     }
 }
 
 #[derive(Clone, Eq, Debug, PartialEq, Hash, Serialize, Deserialize)]
-pub struct Inputs(Vec<NoteId>);
+pub struct Inputs(TransferInputs);
```
danielSanchezQ: Same here, imo

core/src/mantle/tx_builder.rs
```diff
@@ -14,10 +17,35 @@ use crate::{
...
+        BoundedError::TooLong { actual, max } => (*actual, *max),
+        BoundedError::EmptyInput => (0, 0),
+    };
+
+    match kind {
+        "inputs" => TxBuilderError::TooManyInputs { actual, max },
+        "outputs" => TxBuilderError::TooManyOutputs { actual, max },
+        _ => TxBuilderError::TooManyOps { actual, max },
+    }
+}
```
danielSanchezQ: If the kind is static this is probably better off as a const tag instead?

core/src/mantle/tx_builder.rs
```diff
@@ -14,10 +17,35 @@ use crate::{
...
+
 #[derive(Debug, Clone)]
 pub struct MantleTxBuilder {
     mantle_tx: MantleTx,
-    ledger_inputs: Vec<Utxo>,
+    ledger_inputs: LedgerInputs,
```
danielSanchezQ: What is ledger inputs? It is confusing

core/src/mantle/tx_builder.rs
```diff
@@ -42,61 +70,80 @@ impl MantleTxBuilder {
...
     // block.
-    #[must_use]
-    pub fn extend_ops(mut self, ops: impl IntoIterator<Item = Op>) -> Self {
+    pub fn extend_ops(mut self, ops: impl IntoIterator<Item = Op>) -> Result<Self, TxBuilderError> {
         for op in ops {
-            self.mantle_tx.0.try_push(op).expect("Too many ops.");
+            self.mantle_tx
+                .0
+                .try_push(op)
+                .map_err(|err| too_many_from_bounded(&err, "ops"))?;
```
danielSanchezQ: This is better if you implement From in the errors instead of manually having to map the error.

ledger/src/lib.rs
```diff
@@ -761,7 +761,10 @@ mod tests {
     type HeaderId = [u8; 32];
 
     fn create_tx(inputs: Vec<NoteId>, outputs: Vec<Note>, sks: &[ZkKey]) -> SignedMantleTx {
-        let transfer_op = TransferOp::new(Inputs::new(inputs), Outputs::new(outputs));
+        let transfer_op = TransferOp::new(
+            Inputs::new(TransferInputs::try_from(inputs).expect("Invalid inputs size")),
+            Outputs::new(TransferOutputs::try_from(outputs).expect("Invalid outputs size")),
```
danielSanchezQ: this is the stuff, we need to simplify this somehow, there is an extra type that we should not have. We have a wrapper type which should be enough to handle all this without manually injecting the other type.

services/chain/chain-leader/src/leadership.rs
```diff
@@ -404,8 +405,8 @@ mod pol_tests {
 
         // Create a UTXO
         let transfer = TransferOp::new(
-            Inputs::new(vec![]),
-            Outputs::new(vec![Note::new(1000u64, pk)]),
+            Inputs::empty(),
```
danielSanchezQ: good

services/chain/chain-leader/src/leadership.rs
```diff
@@ -404,8 +405,8 @@ mod pol_tests {
...
+            Outputs::new(TransferOutputs::from([Note::new(1000u64, pk)])),
```
danielSanchezQ: not good

danielSanchezQ (changes requested):
We need to work on the ergonomics of the types here. Nice to have interesting types in the underlaying structures, but we should not add complexity on the overall ones. We have a wrapper, lets wrap whats necessary so its transparent.

core/src/mantle/ledger.rs
hansieodendaal: Type names changed to these as discussed:
```rust
pub type BoundedUtxos = UpperBoundedVec<Utxo, MAX_TRANSACTION_INPUTS>;
pub type BoundedInputs = UpperBoundedVec<NoteId, MAX_TRANSACTION_INPUTS>;
pub type BoundedOutputs = UpperBoundedVec<Note, MAX_TRANSACTION_OUTPUTS>;
```

core/src/mantle/ledger.rs
hansieodendaal: Type names changed to these as discussed:
```rust
pub type BoundedUtxos = UpperBoundedVec<Utxo, MAX_TRANSACTION_INPUTS>;
pub type BoundedInputs = UpperBoundedVec<NoteId, MAX_TRANSACTION_INPUTS>;
pub type BoundedOutputs = UpperBoundedVec<Note, MAX_TRANSACTION_OUTPUTS>;
```

hansieodendaal:
@ntn-x2 and @danielSanchezQ - rebased on top of #2790 after it was merged

danielSanchezQ (approved):
Much cleaner, thanks!

ntn-x2 (approved):

PR #2795, opened 2026-05-25
fix(blend): fix e2e tests

davidrusu:

Two failining e2e tests are addresed in this PR:
- **sdp_blend_activity**: 2 nodes start with genesis blend declarations; the test waits past the inactivity window and asserts declarations survive (proving activity proofs flow).
- **sdp_ops_e2e**: Declare -> wait lock period -> Withdraw -> verify declaration removed flow.

Both had the same root cause: PoQ/PoL proof failures on epoch boundaries due to mismatched between PoQ public inputs and private witness (PoQ using some proof inputs from previous epoch and some from the new epoch)

The fix was to introduce a "barrier" pattern where we don't immediately try to start PoQ/PoL proving when we receive the new SDP Membership, instead we transition the epoch and then when the PoL secrets arrive, we start PoQ generation.

TBD

@ntn-x2 

N/A

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.


services/api/src/http/mantle.rs
```diff
@@ -931,21 +905,10 @@ where
...
-        .map(|(_, declaration)| declaration)
-        .collect();
-
-    Ok(declarations)
+    let ledger_state = cryptarchia_ledger_state(handle).await?;
+    Ok(ledger_state
+        .sdp_declarations()
+        .iter()
+        .flat_map(|(_, inner)| inner.values().cloned())
+        .collect())
```
davidrusu, line 913: Prior code path was reading finalized sdp declarations, for our tests we just needed to see the declarations land on chain (i.e. they are valid declarations).

This removes the `GetSdpDeclarations` message from chain-service.

services/blend/src/core/tests/mod.rs
```diff
@@ -1503,84 +1497,6 @@ async fn test_handle_clock_event_new_epoch() {
...
-
-    // PoL info for an older epoch: should return None.
-    let future_epoch = Epoch::new(5);
-    let result = handle_new_secret_epoch_info(&settings, &pol_info, &mut processor, future_epoch);
-    assert!(
-        result.is_none(),
-        "Should return None when PoL epoch < current epoch"
-    );
-}
-
```
davidrusu: The PoL secrets code-path is now processed through a different code path, this test no longer applies

services/blend/src/core/mod.rs
```diff
@@ -929,21 +927,27 @@ where
...
-                current_secret_pol_info = Some(pol_info.poq_private_inputs);
+                tracing::debug!(
+                    target: LOG_TARGET,
+                    current_epoch = ?epoch,
+                    pol_epoch = ?pol_info.epoch,
+                    slot = ?pol_info.poq_public_inputs.slot,
+                    "Received new secret PoL info",
+                );
+                current_pol_info = Some(pol_info);
+                try_rebuild_leader_gen(blend_config, &mut crypto_processor, current_pol_info.as_ref(), current_membership_epoch);
```
davidrusu, line 938: on new PoL secrets, we buffer the PoL info and then try to trigger the leader PoQ generator.

*try* here means checking that the buffered SDP membership info is for the same epoch as the PoL secrets info.

services/blend/src/core/mod.rs
```diff
@@ -929,21 +927,27 @@ where
...
+                match handle_session_event(session_event, blend_config, crypto_processor, message_scheduler, public_info, recovery_checkpoint, backend, sdp_relay, epoch).await {
                     HandleSessionEventOutput::Transitioning { new_crypto_processor, old_crypto_processor, new_scheduler, old_scheduler, new_public_info, new_recovery_checkpoint } => {
                         crypto_processor = new_crypto_processor;
                         old_session_crypto_processor = Some(old_crypto_processor);
                         message_scheduler = new_scheduler;
                         old_session_message_scheduler = Some(old_scheduler);
                         public_info = new_public_info;
                         recovery_checkpoint = new_recovery_checkpoint;
+                        current_membership_epoch = Some(epoch);
+                        try_rebuild_leader_gen(blend_config, &mut crypto_processor, current_pol_info.as_ref(), current_membership_epoch);
```
davidrusu, line 950: Buffer the SDP membership epoch here so that the PoL secrets stream can know whether the SDP membership has transitioned.

We again *try* to rebuild the leader PoQ generator in-case the PoL secrets stream event came before the SDP membership event.

services/blend/src/core/mod.rs
```diff
@@ -1176,16 +1181,7 @@ where
...
-                    if let Some(current_secret_info) = current_secret_info {
-                        new_processor.set_epoch_private(
-                            current_secret_info.clone(),
-                            current_public_info.epoch,
-                            current_epoch,
-                        );
-                    }
-                    new_processor
-                }
+                Ok(new_processor) => new_processor,
```
davidrusu, line 1184: `set_epoch_private` is only called through `try_rebuild_leader_gen` which is called each time the inputs change (either we get another SDP membership, or we get fresh PoL secrets, both attempt to instantiate the PoQ generators)

tests/src/tests/mantle/sdp/ops.rs
```diff
@@ -462,7 +451,14 @@ fn patch_sdp_manual_cluster_config(mut config: RunConfig) -> RunConfig {
         .service
         .bootstrap
         .prolonged_bootstrap_period = Duration::ZERO;
-    config.deployment.cryptarchia.security_param = NonZero::new(5).unwrap();
+    config.deployment.cryptarchia.security_param = NonZero::new(2).unwrap();
+    config.deployment.cryptarchia.epoch_config = EpochConfig {
```
ntn-x2, line 455: Are these changes necessary?

services/blend/src/core/mod.rs
```diff
@@ -2107,10 +2092,6 @@ where
                 ..current_public_info
             };
```
ntn-x2, line 2094: We need this logic here. Yes, we can assume the secret PoL info arrives almost immediately after the public epoch one, but we anyway need to rotate the verifier to be able to receive cover messages using the new epoch information. We cannot wait until the core node receives secret info, because the node might not win a single lottery for the new epoch, yet it must be able to generate cover messages which rely solely on session (soon to be epoch) information. So we must not couple full epoch rotation to the reception of secret PoL info.

ntn-x2 (changes requested):
We cannot wait until the node receives the secret info before updating the cryptographic processor. A node might not have enough stake to win a single lottery slot, yet they will want to generate cover messages and collect activity proofs. Also, this was not an issue before, so I am wondering how did you reach to the conclusion that this has become an issue now? In any case, I think this PR should wait until we remove sessions completely from Blend code, as now we still have them and this PR just relies more on that monitoring the epoch at which we receive a session update. I'd suggest to wait until we're done with the changes that will yield membership info for new epochs together with the rest of the public epoch info.

PR #2799, opened 2026-05-26
test(e2e): add mempool coverage

andrussal:

This PR adds a mempool-specific cucumber feature covering basic functionality. It also introduces a test-only mempool view endpoint.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


tests/cucumber_tests/features/mempool.feature
```diff
@@ -0,0 +1,142 @@
...
+    And transaction "TX_INCLUDED" remains not pending in mempool of all nodes for 2 blocks in 180 seconds
+    Then I stop all nodes
+
+  @mempool_ci @mempool_manual
+  Scenario: Pending transaction survives restart before inclusion
+    Given the genesis block has the following wallet resources:
+      | account_index | token_count | token_amount |
+      | 1             | 2           | 1000         |
+      | 2             | 0           | 0            |
+    And I have deployment config override "time.slot_duration" as "seconds(60)"
```
hansieodendaal, line 51: Could we add short a comment explaining why we override the config?

tests/cucumber_tests/features/mempool.feature
```diff
@@ -0,0 +1,142 @@
...
+    When node "NODE_1" is at height 2 in 360 seconds
+    And I prepare transfer transaction "TX_PENDING_RESTART" of 100 LGO from wallet "WALLET_A" to wallet "WALLET_B"
+    And I submit prepared transaction "TX_PENDING_RESTART" to nodes:
+      | node_name |
+      | NODE_1    |
+    Then transaction "TX_PENDING_RESTART" is pending in mempool of nodes in 30 seconds:
+      | node_name |
+      | NODE_1    |
+    When I restart node "NODE_1"
+    Then transaction "TX_PENDING_RESTART" is pending in mempool of nodes in 30 seconds:
```
hansieodendaal, line 70: We have only one node and the in-memory mempool is lost at a restart. Are we testing the node's ability to recover the mempool state or the test framework being persistent about a submitted transaction?

tests/cucumber_tests/features/mempool.feature
```diff
@@ -0,0 +1,142 @@
...
+    And I start nodes with wallet resources:
+      | node_name | account_index | wallet_name | connected_to |
+      | NODE_1    | 1             | WALLET_A    |              |
+      | NODE_1    | 2             | WALLET_B    |              |
+    When node "NODE_1" is at height 2 in 240 seconds
+    And I submit funded transfer transaction "TX_RESTART" of 100 LGO from wallet "WALLET_A" to wallet "WALLET_B"
+    Then transaction "TX_RESTART" is included on node "NODE_1" in 240 seconds
+    When I restart node "NODE_1"
+    And node "NODE_1" is at height 3 in 240 seconds
+    Then transaction "TX_RESTART" is not pending in mempool of all nodes in 30 seconds
```
hansieodendaal, line 93: ditto - please explain what si being tested here?

tests/src/cucumber/world.rs
```diff
@@ -634,6 +634,8 @@ pub struct CucumberWorld {
     pub wallets: TrackedWallets,
     /// Manual: Mapping of scenario transaction aliases to submitted hashes.
     pub submitted_transactions: HashMap<String, TxHash>,
+    /// Manual: Exact signed transactions prepared for later submission.
+    pub prepared_transactions: HashMap<String, SignedMantleTx>,
```
hansieodendaal, line 694: **Not for this PR:** 

- Would it make sense for a wallet to remember all its transactions?
- If so, transaction types could be tagged, for example tag transactions we want to query on in a scenario.
- This will be closer to a real-world wallet.
- We could have scenario's where we replay past transactions.

hansieodendaal (commented):
Looks really good! I wish we did not need this as a test API only. Some questions:

- Does it make sense to add `mempool_changed` event in the testing API and wire it through to the server? Then the cucumber scenarios can subscribe to this event and be notified of any state change, thereby pinpointing mempool view requests.
- Maybe we should add a test where we start 10 nodes daisy chained with 1 wallet, then submit a user wallet transaction and a funding wallet transaction in rapid succession, then we assert the two transactions landed in the all mempools, and we record the time each mempool state changed for all mempool state changes. Then we assert all mempools received both transactions and all mempools were cleared. This will need the `mempool_changed` event suggestin above.

tests/cucumber_tests/features/mempool.feature
andrussal: > Could we add short a comment explaining why we override the config?

The overrides are there to keep the transaction pending across restart instead of racing block inclusion. I’ll add a short comment for that.

tests/cucumber_tests/features/mempool.feature
andrussal: > We have only one node and the in-memory mempool is lost at a restart. Are we testing the node's ability to recover the mempool state or the test framework being persistent about a submitted transaction?

This is meant to test node recovery. TF only remembers the tx hash for later assertions; it does not re-submit the transaction after restart. If the node loses the recovered mempool state, this check should fail.

tests/cucumber_tests/features/mempool.feature
andrussal: > ditto - please explain what si being tested here?

This checks the inverse case: after inclusion, the tx must not come back as pending after restart. So it guards against recovering stale mempool state.

andrussal:
> * Does it make sense to add `mempool_changed` event in the testing API and wire it through to the server? Then the cucumber scenarios can subscribe to this event and be notified of any state change, thereby pinpointing mempool view requests.

I’m working on a block feed approach for this kind of tracking, so I think it makes sense to align this with that work .

tests/src/cucumber/world.rs
andrussal: > **Not for this PR:**
> 
> * Would it make sense for a wallet to remember all its transactions?
> * If so, transaction types could be tagged, for example tag transactions we want to query on in a scenario.
> * This will be closer to a real-world wallet.
> * We could have scenario's where we replay past transactions.

Yes, I think that would be useful, but I agree it is separate from this PR. Here I only added prepared transaction tracking as scenario state so we can submit the exact same signed tx to selected nodes.

hansieodendaal (approved):
LGTM!

PR #2800, opened 2026-05-26
chore: remove sessions for Blend proxy service

ntn-x2:

Minimal set of changes so that the Blend proxy service can subscribe to epoch-triggered membership changes, replacing the previous session-based logic. Based on top of the Blend core changes: https://github.com/logos-blockchain/logos-blockchain/pull/2798.

I removed the unused `MembershipAdapter` to avoid any confusion, and introduced a `BlendEpochState` type, which includes the relevant bits of information from Blend taken from the epoch state. Also, it paves the way to remove the `EpochHandler` type in favor of the existing `UninitializedEpochEventStream`, formerly known as `UninitializedSessionEventStream`, which already has the logic to yield items from a stream interleaved with transition periods, so we will re-use that for epochs and get rid of the `EpochHandler`.

Yes, I think so. I tried to keep the changes as minimal as possible while maintaining readability of the changeset.

@ntn-x2 

Yes, so far.

No.


nodes/node/binary/src/config/blend/mod.rs
```diff
@@ -65,9 +67,14 @@ impl ServiceConfig {
...
-                        .deployment
-                        .slots_per_epoch_transition_period(slots_per_block, &slot_duration),
+                    epoch_transition_period: Duration::from_secs(
+                        self.deployment
+                            // TODO: No need to divide by `slot_duration` to then multiply by the
+                            // same right after.
+                            .slots_per_epoch_transition_period(slots_per_block, &slot_duration)
+                            .get()
+                            * slot_duration.as_secs(),
+                    ),
```
davidrusu, line 77: why not stay in slots?

services/blend/src/membership/chain.rs
```diff
@@ -80,27 +91,47 @@ where
...
+                                lottery_1: epoch_state.lottery_1,
+                                membership_info,
+                            }))
+                        }
+                        Ok(Err(e)) => {
+                            tracing::warn!(target: LOG_TARGET, "Chain service returned error for epoch state at slot {slot:?}: {e:?}");
+                            Some(None)
+                        }
+                        Err(e) => {
+                            tracing::warn!(target: LOG_TARGET, "Failed to query epoch state at slot {slot:?}: {e:?}");
```
davidrusu, line 129: both should be ::error level logs

davidrusu (approved):
ship it!

nodes/node/binary/src/config/blend/mod.rs
ntn-x2: Yeah we can come back to this later, but the idea is that I'd like to get rid of the `EpochHandler` type and just use the `UninitializedEpochEventStream` which did the job for sessions and session transitions. But that takes a transition period as a `Duration`, and I think Blend does not need to know how long a slot is per se, just be aware of slot-based events (like epochs). But as I said, we can come back to this once the whole picture is clear(er).

PR #2804, opened 2026-05-27
chore: remove sessions from Blend edge service

ntn-x2:

Changed the logic in the Blend edge service to remove sessions. Also, we now buffer epoch info slightly differently, since we don't have to deal with intra-session epoch transitions. On each new public/private epoch info, we check if the other half for the current epoch is set, and proceed to create the message handler. Else, update/clear the buffer depending on the condition. I also cleaned up mentions of `session` from the edge folder wherever possible.

Yes.

@ntn-x2 

Yes.

No.

### DIFF
--- a/services/blend/src/edge/mod.rs
+++ b/services/blend/src/edge/mod.rs
@@ -336,7 +336,7 @@ where
         .await
         .expect("Should not fail to subscribe to secret PoL info stream.");
 
-    let mut pending_epoch_info: Option<PendingEpochInfo<NodeId>> = Some(PendingEpochInfo {
+    let mut pending_epoch_info = Some(PendingEpochInfo {
         epoch: current_epoch_info.epoch,
         info_type: PendingEpochInfoType::Public(Box::new(current_epoch_info.membership_info)),
     });
@@ -379,7 +379,7 @@ where
 
 fn handle_new_epoch_info<Backend, NodeId, ProofsGenerator, RuntimeServiceId>(
     BlendEpochState {
-        epoch,
+        epoch: new_epoch,
         membership_info,
         lottery_0,
         lottery_1,
@@ -417,12 +417,12 @@ where
     }
 
     match pending_epoch_info.take() {
-        // Previous epoch handler is running, so we stop it and buffer membership info.
+        // Previous epoch handler is running, so we stop it and buffer new membership info.
         None => {
             debug!(target: LOG_TARGET, "New epoch public info received. Stopping message handler until secret PoL info is received.");
             *current_epoch_message_handler = None;
             *pending_epoch_info = Some(PendingEpochInfo {
-                epoch,
+                epoch: new_epoch,
                 info_type: PendingEpochInfoType::Public(Box::new(membership_info.clone())),
             });
         }
@@ -434,20 +434,23 @@ where
             ..
         }) => {
             warn!(target: LOG_TARGET, "New epoch public info received without the previous epoch info being consumed.");
-            *current_epoch_message_handler = None;
+            assert!(
+                current_epoch_message_handler.is_none(),
+                "If public epoch info is buffered, the message handler should not be running."
+            );
             *pending_epoch_info = Some(PendingEpochInfo {
-                epoch,
+                epoch: new_epoch,
                 info_type: PendingEpochInfoType::Public(Box::new(membership_info.clone())),
             });
         }
         Some(PendingEpochInfo {
-            epoch: new_epoch,
+            epoch,
             info_type: PendingEpochInfoType::Private(new_private_pol_info),
         }) => {
             // Let's make sure we always clean up the pending info whenever a public or
             // private epoch event is processed.
             assert!(
-                epoch == new_epoch,
+                new_epoch == epoch,
                 "Old pending epoch info was found, and this should never happen."
             );
             debug!(target: LOG_TARGET, "New epoch public info received with a buffered secret PoL info for the same epoch {new_epoch:?}. Trying to create a new epoch-bound message handler.");
@@ -475,7 +478,7 @@ where
                 new_public_inputs,
                 *new_private_pol_info,
                 overwatch_handle,
-                epoch,
+                new_epoch,
             )?;
 
             *current_epoch_message_handler = Some(new_handler);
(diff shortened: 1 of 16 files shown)


services/blend/src/membership/chain.rs
```diff
@@ -1,16 +1,15 @@
-//! Chain-derived membership.
+//! Chain-derived per-epoch state.
```
davidrusu, line 1: feels like this module should be renamed `epoch`

services/blend/src/edge/mod.rs
```diff
@@ -413,38 +373,27 @@ where
...
-    new_membership_info: &MembershipInfo<NodeId>,
+fn handle_new_epoch_info<Backend, NodeId, ProofsGenerator, RuntimeServiceId>(
+    BlendEpochState {
+        epoch: new_epoch,
+        membership_info,
+        lottery_0,
+        lottery_1,
+        nonce,
+        aged,
+    }: BlendEpochState<NodeId>,
```
davidrusu: personally, I find this sort of destructuring in arguments really ugly, you don't like doing `epoch_state.membership_info`?

services/blend/src/edge/mod.rs
```diff
@@ -453,158 +402,199 @@ where
...
+            // A buffered private must be for the same epoch as the incoming
+            // public: it can't be from the future (private is only produced
+            // once its epoch is active, by which time the public has already
+            // been emitted) and it can't be from the past (the previous
+            // public for that epoch would have consumed it). Either inequality
+            // is a real bug upstream.
+            assert!(
+                new_epoch == epoch,
+                "Buffered secret PoL info for epoch {epoch:?} does not match incoming public epoch info for epoch {new_epoch:?}."
+            );
```
davidrusu: this statement is too strong. there are many interleaving of events and network parameter settings where this assert can fail.

e.g. as an extreme case:
1. 1 slot epochs
2. blend epoch stream enqueues BlendEpochState{epoch:1}
3. system/os/tokio delays delivering blend epoch state for a few slots
4. concurrently, chain-leader transitions to Epoch 2
5. chain-leader publishes PoL secrets to blend for Epoch 2
6. PoL secrets and BlendEpochState are on different channels so tokio can schedule delivery however it wishes
7. PoL secrets are delivered first and pending state is set to `{epoch: 2 Private(..)}`
8. Finally BlendEpochState{epoch:1} is delivered.
9. this assert panics.

We chose to use streams for this information so we need to deal with every interleaving, even if they are unlikely.

In this case, 
- if pending private does not match epoch state, just start the epoch handler with no private data
- if pending private matches epoch state, then start the epoch handler with private data

services/blend/src/edge/mod.rs
```diff
@@ -453,158 +402,199 @@ where
...
+            });
+        }
+        Some(PendingEpochInfo {
+            info_type: PendingEpochInfoType::Private(_),
+            epoch,
+        }) => {
+            panic!(
+                "New secret PoL info received while there is already buffered secret PoL info for a epoch {epoch:?}. This should never happen."
+            );
+        }
```
davidrusu: This condition is again too strong, BlendEpochState and PoL info are on separate streams, they can be interleaved in any order (including BlendEpochState stream being starved for more than an epoch).

We need to deal with any interleaving.

In this case,

Just replace the pending state with the new Private info.

services/blend/src/membership/chain.rs
ntn-x2: Yeah we need to rename a move few things around, but I would like the changeset to remain as small as it can be while removing sessions.

services/blend/src/edge/mod.rs
ntn-x2: Personally, I like this, because it gives a chance to check whether you are not using some fields of a type or if you are continuously ignoring some others with `..`. So unless the whole thing is required in some function call, I prefer to destructure, especially in cases where many fields have the same type, because not destructuring would not help with typos, while with destructuring if you use the same field twice instead of two different fields, one of them will remain unused and you can spot that.

services/blend/src/edge/mod.rs
ntn-x2: I mean, 1 slot epoch is per se not viable in this case. You are talking about a case I personally would not even consider edge, but simply impractical. And in that case the node is not able to progress at all, so a panic in my opinion is the right thing there. And that's because

> if pending private does not match epoch state, just start the epoch handler with no private data

This is not possible for an edge node. This node does not produce any cover traffic, so a message handler for edge node is designed to be instantiated only when both public and private info for an epoch are available.

> if pending private matches epoch state, then start the epoch handler with private data

This is why the assert here, to rule out cases which are not practical and would lead to undefined behaviour regardless of whether we panic or not, like in the case of 1-slot epochs, since we then have epoch transitions and other stuff (in the core node) which would simply break.

I personally like to embed these assumptions in the codebase for stuff that cannot be verified at compile time. So a node that is running fine will have always public info for an epoch, and optionally private one. If the node has no private info, it means it will never propose a slot, which means the edge node will never receive a message to blend (based on current assumptions), so this is fine. If a node receives a secret info but not public info for an epoch, it will again prevent the node from working because public info should ALWAYS be available for all epochs, so I thought a panic would help signal that something is definitely off for the node. Even just performance-wise, but that means the node will not be able to blend blocks if it's not able to provide public epoch info on time for the edge service to process and generate the required proofs.

services/blend/src/edge/mod.rs
ntn-x2: > including BlendEpochState stream being starved for more than an epoch

This means the whole node is not working, as it will never be able to propose blocks. I think a panic would be the right behaviour here, because at least a process restart might help clean up some of the stuff that led to the node being that slow (hypothetically).

But in general, we changed panic to abort the process because we wanted a way to be notified when an assumption we made was broken either due to a bug (which we would catch right away) or because of something else weird going on that would bring the node to an invalid or degraded state, at which point it would be pointless to let it run anyway (e.g., in the past, it would not blend any blocks if Blend panicked but the rest of the node was still running, which was misleading).

ntn-x2:
@davidrusu refactored the code based on your suggestion: https://github.com/logos-blockchain/logos-blockchain/pull/2804. It does look cleaner indeed, so thank you. I removed the asserts, although I think we need to enforce some more conditions instead of silently swallow weird edge cases, but I did not do it for now.

danielSanchezQ (approved):
Letting @davidrusu leading the deeper review, so wait for his approval. Otherwise looks clean to me. Looking forward to the remaining changes.

davidrusu (approved):
beautiful, 🚢 it!

PR #2808, opened 2026-05-27
feat: add turn-to-write event for round-robin sequencers

hansieodendaal:

- Added turn-to-write event via watch channel for round-robin sequencers.
- Expanded a round robin cucumber scenario to act on the submitted events.

Yes

@hansieodendaal @pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


pradovic:
Hi Hansie, thanks for getting this together!

One design thought: have we considered making this an Event::TurnChanged variant returned from the drive loop instead of a separate watch channel?

The reasoning:

1. The drive loop is already our canonical state machine — Ready, Published, TxsFinalized, ChannelUpdate all flow through Event. A turn transition feels like a natural fit there.
2. Events naturally carry rich payloads, so we could include the slot info the consumer actually needs: something like TurnChanged {is_ours, starting_slot, ends_at_slot, posting_timeframe }. With a bool we lose the time-budget info and the consumer has to do a second hop into channel_view to learn it.
3. Emit on both edges (true→false and false→true), so consumers don't need a separate watch.
4. For passive "am I in turn right now" queries, subscribe_channel_view() already has is_our_turn plus the slot timeframe fields, so we keep that capability for free.

That would let us drop turn_to_write_tx / subscribe_turn_to_write / publish_turn_to_write and avoid a third parallel notification surface alongside channel_view_tx and ready_tx. The cucumber test then asserts on Event::TurnChanged { is_ours: true } consistent with other event assertions.

What do you think?

hansieodendaal:
> One design thought: have we considered making this an Event::TurnChanged variant returned from the drive loop instead of a separate watch channel?
> 
> The reasoning:
> 
> 1. The drive loop is already our canonical state machine — Ready, Published, TxsFinalized, ChannelUpdate all flow through Event. A turn transition feels like a natural fit there.
> 2. Events naturally carry rich payloads, so we could include the slot info the consumer actually needs: something like TurnChanged {is_ours, starting_slot, ends_at_slot, posting_timeframe }. With a bool we lose the time-budget info and the consumer has to do a second hop into channel_view to learn it.
> 3. Emit on both edges (true→false and false→true), so consumers don't need a separate watch.
> 4. For passive "am I in turn right now" queries, subscribe_channel_view() already has is_our_turn plus the slot timeframe fields, so we keep that capability for free.
> 
> That would let us drop turn_to_write_tx / subscribe_turn_to_write / publish_turn_to_write and avoid a third parallel notification surface alongside channel_view_tx and ready_tx. The cucumber test then asserts on Event::TurnChanged { is_ours: true } consistent with other event assertions.
> 
> What do you think?

Thanks for the suggestions!

1. I agree that `Event::TurnChanged` would be great for event-driven consumers. However, I think the watch channel should stay for reliability - if a consumer isn't actively polling `next_event()`, a turn transition could be missed. A watch channel is genuinely a fire-and-forget broadcast. These are fundamentally different guarantees. I would keep the watch for notification reliability. 
2. Agree with the rich notification - I changed the watch to return
```rust
pub struct TurnNotification {
    pub our_turn_to_write: bool,
    pub starting_slot: Option<u64>,
    pub ends_at_slot: Option<u64>,
    pub turn_to_write_slots: Option<u32>,
    pub current_slot: Option<u64>,
}
```
3. Edge triggers are already implemeted - the watch fires on `true→false` and `false→true` (just not observed in a test)

pradovic (approved):

PR #2813, opened 2026-05-28
fix(chain): apply block after storage succeeds

andrussal:

This PR prevents chain state from being committed before the required block storage writes succeed.

Previously process_block applied the block to in-memory Cryptarchia state before storing the block and immutable block indexes. If a storage write failed after that point, runtime state could advance while persisted/recoverable state was incomplete, which could lead to restart/recovery inconsistency. This now applies the block against a candidate state, stores the required data first, and only commits the candidate state after storage succeeds.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/services/chain/chain-service/src/lib.rs
+++ b/services/chain/chain-service/src/lib.rs
@@ -1045,12 +1045,12 @@ where
         let header = block.header();
         let prev_lib = cryptarchia.lib();
 
+        let mut candidate = cryptarchia.clone();
         let (pruned_blocks, reorged_blocks, events) =
-            cryptarchia.try_apply_block(&block, current_slot)?;
-        let new_lib = cryptarchia.lib();
+            candidate.try_apply_block(&block, current_slot)?;
+        let new_lib = candidate.lib();
 
         let tx_count = block.transactions().count();
-        metrics::emit_block_transactions_metric(tx_count);
 
         relays
             .storage_adapter()
@@ -1062,10 +1062,13 @@ where
             &pruned_blocks,
             Some(prev_lib),
             new_lib,
-            cryptarchia.consensus.lib_branch().slot(),
+            candidate.consensus.lib_branch().slot(),
             relays.storage_adapter(),
         )
         .await?;
+
+        *cryptarchia = candidate;
+        metrics::emit_block_transactions_metric(tx_count);
 
         let processed_block_event = {
             let tip = cryptarchia.tip_branch();
--- a/services/chain/chain-service/src/tests/mod.rs
+++ b/services/chain/chain-service/src/tests/mod.rs
@@ -25,6 +25,7 @@ use lb_ledger::{
 };
 use lb_storage_service::{
     StorageMsg, StorageService,
+    api::{StorageApiRequest, chain::requests::ChainApiRequest},
     backends::{
         StorageBackend as _,
         rocksdb::{RocksBackend, RocksBackendSettings},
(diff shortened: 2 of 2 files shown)


services/chain/chain-service/src/tests/mod.rs
```diff
@@ -239,6 +240,135 @@ async fn get_block_ids() {
     ));
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn process_block_does_not_mutate_state_when_storage_send_fails() {
+    let (broadcast_tx, _broadcast_rx) = mpsc::channel(10);
+    let (storage_tx, storage_rx) = mpsc::channel(10);
+    drop(storage_rx);
```
youngjoon-lee, line 249: Is dropping `storage_rx` for triggering storage errors? Then, please leave a comment to explain it.

services/chain/chain-service/src/tests/mod.rs
```diff
@@ -239,6 +240,135 @@ async fn get_block_ids() {
...
+        .await;
+    let (new_block_tx, mut new_block_rx) = broadcast::channel(10);
+    let (lib_tx, mut lib_rx) = broadcast::channel(10);
+
+    let storage_task = tokio::spawn(async move {
+        let Some(StorageMsg::Api {
+            request: StorageApiRequest::Chain(ChainApiRequest::StoreBlock { response_tx, .. }),
+        }) = storage_rx.recv().await
+        else {
+            panic!("expected store block request");
```
youngjoon-lee: Is this panic for triggering an error if the request is `StoreImmutableBlockIds`? If so, please leave a comment. 
Or, IMO, it's better to explicitly return error if the request is `StoreImmutableBlockIds`.

services/chain/chain-service/src/lib.rs
```diff
@@ -1079,12 +1079,12 @@ where
             }
         };
 
+        let mut candidate = cryptarchia.clone();
```
youngjoon-lee, line 1048: Cloning `Cryptarchia` is not always cheap because it holds `HashMap<HeaderId, LedgerState>` which may be large especially when bootstrapping. Inside `LedgerState`, we use `rpds`s because we designed `LedgerState` to be cloned frequently. In contrast, we didn't design `Cryptarchia` to be cloned many times. (Yeah, we should have not make it `Clone`).

But now, I understand that it's not easy to find any better way to not clone `Cryptarchia` in this PR. So, I think it's fine to replace all `HashMap`s in the `Cryptarchia` with `rpds`.

One concern is the insertion cost of rpds. However, the additional rpds structures we’re considering are likely negligible compared to the existing ones in LedgerState, which are updated on a per-tx (or per-UTXO) basis rather than per block.

I will open a PR to introduce rpds, on top of your PR, soon.

youngjoon-lee: https://github.com/logos-blockchain/logos-blockchain/pull/2896 I just opened a draft PR. I will mark it ready once CI passes.

services/chain/chain-service/src/tests/mod.rs
andrussal: > Is this panic for triggering an error if the request is `StoreImmutableBlockIds`? If so, please leave a comment.
> Or, IMO, it's better to explicitly return error if the request is `StoreImmutableBlockIds`.

Changed it to return an error instead of panicking. Also added a short comment there, since the failure is a bit indirect: we ack StoreBlock, then drop the receiver so the immutable index write fails on send.

danielSanchezQ:
Good catch @andrussal and good fix @youngjoon-lee , thanks!

services/chain/chain-service/src/tests/mod.rs
youngjoon-lee: Thanks, but I still think it's not easy to understand what the `storage_task` intends. What do you think about my suggestion below? In this ways, the `storage_task` doesn't need to know that `StoreBlock` is always requested first before `StoreImmutableBlockIds`. Also, the task now returns an error explicitly without dropping the channel, which I think much easier to read. In other words, the `storage_task` doesn't need to know the order of requests made by the chain service.
```diff
diff --git a/services/chain/chain-service/src/tests/mod.rs b/services/chain/chain-service/src/tests/mod.rs
index 870fbc671..960997bab 100644
--- a/services/chain/chain-service/src/tests/mod.rs
+++ b/services/chain/chain-service/src/tests/mod.rs
@@ -302,24 +302,31 @@ async fn process_block_does_not_mutate_state_when_immutable_index_write_fails()
     let (new_block_tx, mut new_block_rx) = broadcast::channel(10);
     let (lib_tx, mut lib_rx) = broadcast::channel(10);
 
-    let storage_task = tokio::spawn(async move {
-        let Some(msg) = storage_rx.recv().await else {
-            return Err("expected store block request, storage channel closed");
-        };
-
-        let StorageMsg::Api {
-            request: StorageApiRequest::Chain(ChainApiRequest::StoreBlock { response_tx, .. }),
-        } = msg
-        else {
-            return Err("expected store block request");
-        };
-
-        response_tx
-            .send(Ok(()))
-            .map_err(|_| "store block reply receiver dropped")?;
-
-        // Drop the receiver so the following immutable-index write fails at send time.
-        Ok(())
+    tokio::spawn(async move {
+        while let Some(msg) = storage_rx.recv().await {
+            match msg {
+                StorageMsg::Api {
+                    request:
+                        StorageApiRequest::Chain(ChainApiRequest::StoreBlock { response_tx, .. }),
+                } => {
+                    response_tx.send(Ok(())).unwrap();
+                }
+                StorageMsg::Api {
+                    request:
+                        StorageApiRequest::Chain(ChainApiRequest::StoreImmutableBlockIds {
+                            response_tx,
+                            ..
+                        }),
+                } => {
+                    response_tx
+                        .send(Err(
+                            "StoreImmutableBlockIds is always rejected in this test".to_owned(),
+                        ))
+                        .unwrap();
+                }
+                _ => panic!("unexpected msg: {msg:?}"),
+            }
+        }
     });
 
     let (mut cryptarchia, block) = test_chain_with_next_block();
@@ -340,10 +347,6 @@ async fn process_block_does_not_mutate_state_when_immutable_index_write_fails()
         &lib_tx,
     )
     .await;
-    storage_task
-        .await
-        .expect("storage task should not panic")
-        .expect("storage task should handle initial StoreBlock request");
 
     match &result {
         Err(Error::Storage(_)) => {}

```

andrussal: Yes, I agree. I updated it to explicitly reject `StoreImmutableBlockIds` in the storage task instead of relying on the channel closing.

youngjoon-lee (approved):
Thank you. It looks good to merge after rebasing the branch and CI passes. I will merge my PR #2896 right after you merge this.

danielSanchezQ (approved):

PR #2819, opened 2026-05-28
test: improves e2e test artifact handling

hansieodendaal:

This PR improves E2E artifact handling for manual-cluster tests so successful runs clean up automatically while failed runs remain inspectable, and makes CI artifact upload target the new artifact root.

- Added configurable E2E scenario base directory support in tests/src/common/manual_cluster.rs via `E2E_TESTS_BASE_DIR_OVERRIDE`.
- Switched manual-cluster scenario directory creation to temp-managed directories under the chosen base root (`tempfile::Builder::tempdir_in`), restoring auto-cleanup semantics for successful runs. 
- Added guard-based cleanup in `manual_cluster`:
  - unregisters system monitor output on drop
  - preserves directories on panic 
  - preserves directories when `E2E_KEEP_LOGS=true` (truthy env handling) 
  - preserves directories when real artifacts exist (beyond `system_stats`)
- Updated manual-cluster helpers/signatures to accept optional scenario directory override and propagated this to callers.
- Updated E2E test call sites to route artifacts into `tests/.e2e_artefacts` using `E2E_ARTIFACTS_DIR`.
- Fixed stats sink flush behavior in `tests/testing_framework/src/diagnostics/system_monitor/sink.rs` by calling `sync_all()` for NDJSON/CSV writes.
- Updated workflow artifact upload in `.github/workflows/end-to-end-with-cucumber.yml`
  - upload path changed to `tests/.e2e_artefacts` 
  - `if-no-files-found: ignore` added
- Updated docs in `tests/README.md` with
  - `E2E_TESTS_BASE_DIR_OVERRIDE 
  - `E2E_KEEP_LOGS=true` behavior for manual-cluster runs 
  - clarification of `TF_KEEP_LOGS=1`

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


tests/src/common/manual_cluster.rs
```diff
@@ -9,17 +9,68 @@ use lb_libp2p::Multiaddr;
...
 use testing_framework_core::scenario::{DynError, PeerSelection, StartNodeOptions, StartedNode};
 use tokio::time::error::Elapsed;
 
+use crate::cucumber::defaults::{E2E_KEEP_LOGS, E2E_TESTS_BASE_DIR_OVERRIDE};
+
 pub struct LocalManualClusterHarnessBase {
     pub scenario_base_dir: PathBuf,
     pub deployment: DeploymentPlan,
     pub cluster: LbcManualCluster,
+    _scenario_base_dir_guard: ScenarioBaseDirGuard,
```
andrussal, line 26: This looks risky. `LocalManualClusterHarnessBase` owns `_scenario_base_dir_guard`, so once `base` is dropped, `ScenarioBaseDirGuard::drop` runs. 

I noticed this in the SDP ops test helper where we return `cluster` and `scenario_base_dir` separately, but leave the guard behind. After the helper returns, the guard is dropped while the caller still uses the returned cluster/path. I think this should return the harness itself if possible.

I think longer term, it would be better if `LocalManualClusterHarnessBase` made this harder to misuse, e.g. by not exposing movable fields directly and forcing callers to borrow the cluster/path from the harness or move the harness as a whole.

tests/src/common/manual_cluster.rs
```diff
@@ -9,17 +9,68 @@ use lb_libp2p::Multiaddr;
...
+            let _kept_path = tempdir.keep();
+        }
+    }
+}
+
+fn e2e_keep_logs_enabled() -> bool {
+    is_truthy_env(E2E_KEEP_LOGS)
+}
+
+fn should_preserve_scenario_base_dir(path: &Path) -> bool {
```
andrussal: What semantics are intended for `should_preserve_scenario_base_dir`? Panic and `E2E_KEEP_LOGS` are already handled above, so this branch seems to preserve the directory based only on whether it contains entries other than `system_stats.ndjson` / `system_stats.csv`.

andrussal (approved):
Nice! Left couple of comments, I think at last one would be good to handle somehow.

tests/src/common/manual_cluster.rs
hansieodendaal: Good catch

tests/src/common/manual_cluster.rs
hansieodendaal: Renamed the fn and added a comments to why it is there

PR #2824, opened 2026-05-29
refactor(logging): migrate more crates to log targets

andrussal:

Migrate more crates to the log targets crate.

This PR covers smaller crates that only have one or a few logging call sites, and adds the corresponding targets to the shared target registry.

Changed crates include:

- `core`
- `kms/keys`
- `kms/operators`
- `nodes/api-common`
- `nodes/node/binary`
- `nodes/node/http-client`
- `services/key-management-system`
- `services/tracing`
- `services/utils`
- `services/wallet`
- `wallet`
- `zk/proofs/poc`
- `zk/proofs/pol`
- `zk/proofs/poq`
- `zk/proofs/zksign`
- `zone-sdk`

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


tracing/targets/src/api.rs
```diff
@@ -3,5 +3,10 @@ use lb_log_targets_macros::log_targets;
 log_targets! {
     root = api;
 
+    PPROF,
+    wallet::{
```
ntn-x2: Should these live under `http` as well then?

tracing/targets/src/lib.rs
ntn-x2: Do you think we could automate this with a macro? What happens if we add new targets and forget to add them here?

tracing/targets/src/mantle.rs
```diff
@@ -0,0 +1,10 @@
+use lb_log_targets_macros::log_targets;
+
+log_targets! {
+    root = mantle;
+
+    sdp::{
```
ntn-x2, line 6: Maybe a `message` layer?

ntn-x2 (approved):
Nice!

tracing/targets/src/lib.rs
andrussal: > Do you think we could automate this with a macro? What happens if we add new targets and forget to add them here?

Good point, I thought about this as well. If a target module was added here but not included in `all_targets()`, those targets would not be used for validation.

I wrapped the module declarations in a macro now, so the same list exports the modules and adds them to `all_targets()`. That removes the extra manual step.

PR #2826, opened 2026-05-29
chore: remove sessions from Blend core service

ntn-x2:

By far the biggest changeset. I did not focus on removing all unused components in this PR, but I updated the core service to change what used to happen on session changes to now do it on epoch changes. For secret info, if it arrives before the public info (for the same epoch), we buffer it until the public part arrives. If it arrives afterwards, we call `set_private_info` on the underlying proof generator to let it start pre-generating leadership proofs. 

As before, as soon as a new epoch starts, we start pre-generating core proofs regardless of whether the node will get private PoL info and start generating leadership proofs, this is the main difference between how the edge and core modes work.

I kept the same tests as before, and added a couple more to test the interaction between public and private epoch info. I also tried to minimize changes to the logic which is today quite well tested. The only additional bit is to check if a secret epoch info has arrived by the time we rotate epoch, which was not the case for session rotations. The rest should remain quite the same.

Very large changeset since Blend core touches on a lot of components that did not make it into previous PRs, but the core of the changes is the core Blend service itself.

Yes, so far.

No.


davidrusu (approved):
🔥

PR #2834, opened 2026-06-01
chore(zone-sdk): split refactor

pradovic:

This is unfortunately a large PR because it includes a structural refactor of the sequencer into multiple files, and most of the diff comes from that, the actual additions set is small. The goal was not to change the logic, but keep the changes "minimal" but still split the code, and hopefully make it both easier for maintainers and users to follow.

We could probably split it better in the future, but would need to introduce new types and make a bigger refactor. I opted for more minimal change for now. We can refactor it more in different iterations when things are more stable and less time critical. The main API remains the same. Subscription API is unified to use the same already existing pattern for turn events.

For review, I think the most important is general  project layout and usage example in lib.rs. The logic should be same/similar to what it was before.

The PR summary:

### Commits 1 & 2 — Project structure split + cleanup

Reorganized zone-sdk to match its two-subsystem architecture (sequencer / indexer + shared types). The 3475-line sequencer.rs is gone; everything sequencer-specific now lives under sequencer/, indexer-specific code remains at the crate root, and lib.rs contains only the shared Zone* types.

```
zone-sdk/src/
├── lib.rs              shared types (ZoneMessage, ZoneBlock, Deposit, Withdraw)
├── adapter.rs          HTTP adapter (used by both subsystems)
├── indexer.rs          indexer subsystem
└── sequencer/
    ├── mod.rs          public re-exports
    ├── types.rs        public event/wire types
    ├── handle.rs       SequencerHandle
    ├── slot_clock.rs   internal
    ├── tx_builder.rs   internal pure helpers
    ├── block_fetch.rs  internal block IO + extract tests
    ├── state.rs        internal TxState (used to be leaked at lb_zone_sdk::state)
    ├── zone_sequencer.rs  struct + init + API
    ├── actor.rs        impl: request/event handling + actor tests
    └── backfill.rs     impl: connection + backfill loop
```

Alongside the split:

- Removed dead code (PublishedTx::Deposit, three unused TxState methods, ChannelUpdateInfo::is_conflict)
- Tightened PublishedTx::inscription() from Option<&InscriptionInfo> to &InscriptionInfo
- Fixed stale doc links
- Added module-level //! docs to lib.rs and sequencer/mod.rs with the canonical usage pattern

#![allow(clippy::multiple_inherent_impl)] on actor.rs / backfill.rs covers the split impl ZoneSequencer blocks.

### Commits 3 & 4 — Sync / stream API split + subscription API unification

Split the sequencer API by semantics; type signatures alone now tell the reader which side a method lives on.

- **`ZoneSequencer`** — sync surface: `init`, `events()` (stream-based drive), `is_ready`, `checkpoint`, `subscribe_ready`, `subscribe_channel_view`, `subscribe_turn_to_write`, `subscribe_checkpoint`.

    `next_event` is now private; users drive the sequencer via `events()`, which returns `impl Stream<Item = Event> + Send + Unpin + '_`.

    Each `subscribe_*` calls is now unified with turn event (watch), fires current state on the first call.

 - **`SequencerHandle`** — cloneable async commands only: `publish_message`, `publish_atomic_withdraw`, `prepare_tx`, `sign_tx`, `submit_signed_tx`, `channel_config`.

Also fixed a watch-channel bug surfaced by the receiver-side cleanup: send_replace instead of send on checkpoint_tx / channel_view_tx, ensuring the stored value updates even when no subscriber is currently alive.

For usage examples, see the module-level docs in lib.rs.

Yes

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


zone-sdk/src/sequencer/block_fetch.rs
```diff
@@ -0,0 +1,798 @@
...
+                let parent_msg = last_in_block.unwrap_or_else(MsgId::root);
+                let info = InscriptionInfo {
+                    tx_hash,
+                    parent_msg,
+                    this_msg: config.id(),
+                    payload: [].into(),
+                };
+                last_in_block = Some(info.this_msg);
+                items.push(info);
+            }
```
hansieodendaal, line 548: _This may be a question outside the scope of this PR, maybe related to #2746:_

I see the SDK assumes that, within a block, channel tip-bearing ops already appear in valid chain order. But `ChannelConfigOp` has no explicit parent lineage and directly resets `channel.tip_message = config.id()`.

What happens if the same `ChannelConfigOp` is replayed later on? Since `config.id()` is content-derived, replaying the same op appears to reset the parent lineage back to that old config id. That seems like it would invalidate any later inscriptions that were originally built on the newer tip lineage, unless the chain is effectively rebuilt from that old config tip again.

zone-sdk/src/sequencer/block_fetch.rs
```diff
@@ -0,0 +1,798 @@
...
+
+/// Extract channel inscription info from a block's transactions, in
+/// parent→child chain order. Transactions in a block are not guaranteed
+/// to be in chain order, so we topologically sort by inscription lineage.
+/// Callers (e.g. `channel_tip_at`) rely on `last()` being the chain tail.
+///
+/// Panics if the inscriptions for the channel in a single block do not
+/// form a single linear chain — that would be a protocol-level invariant
+/// violation.
+fn extract_inscriptions(txs: &[SignedMantleTx], channel_id: ChannelId) -> Vec<InscriptionInfo> {
```
hansieodendaal: The name of `fn extract_inscriptions` could possibly better describe what is happening inside?
Something like this (or better):
- extract_channel_tip_ops
- extract_channel_lineage_entries
- extract_channel_message_lineage
- extract_tip_tracking_entries

zone-sdk/src/sequencer/block_fetch.rs
```diff
@@ -0,0 +1,798 @@
...
+    let mut current = root.this_msg;
+    while let Some(next) = by_parent.get(&current).copied() {
+        sorted.push(next.clone());
+        current = next.this_msg;
+    }
+    sorted
+}
+
+fn matches_channel(tx: &SignedMantleTx, channel_id: ChannelId) -> bool {
+    tx.mantle_tx.ops().iter().any(|op| match op {
```
hansieodendaal: `fn matches_channel` gives the idea that all `Op` variants will be matched. Maybe it could be renamed to something like this (or better):
- matches_channel_lineage_op
- touches_channel_tip
- matches_tracked_channel_op
- matches_channel_message_flow

testnet/tui-zone/src/lib.rs
```diff
@@ -64,20 +65,13 @@ pub async fn run(args: InscribeArgs) {
 
     loop {
         tokio::select! {
-            event = sequencer.next_event() => {
+            event = events.next() => {
                 if let Some(event) = event {
+                    state.set_channel_view(view_rx.borrow().clone());
```
hansieodendaal, line 70: Since the watch event was dropped here, the TUI appears to no longer re-renders on pure channel-view changes - is this intentional? Maybe not so important for the UI, as all meaningful channel view changes are covered by events?

hansieodendaal (commented):
This looks really good! Just a couple of questions to answer please.

zone-sdk/src/sequencer/block_fetch.rs
pradovic: Good question, this is how I understood the spec and the code:
- Identical ChannelConfigOp replayed: config.id() is content-derived, so replaying the exact same op produces the same tip_message value.
  The SDK observes "tip set to X" then later "tip set to X again" — same value, no actual lineage break.
- Different ChannelConfigOp on a reorg branch: handled the same way the SDK handles any reorg — channel_tip_at(tip) walks the canonical chain on every block event, so if a config op is no longer on canonical, the tip rolls back; if a new config op is adopted, the tip moves to it. Pending inscriptions whose parent is now off-canonical surface as OrphanedTx::Inscription in Event::ChannelUpdate.

So the SDK should be OK either way, unless I am missing something.

zone-sdk/src/sequencer/block_fetch.rs
pradovic: Nice, will apply!

testnet/tui-zone/src/lib.rs
pradovic: I was thinking about the same thing. 

All meaningful channel-view changes coincide with an event (block processed → ChannelUpdate / TxsFinalized; user publish → eventually Event::Published; reconnect → Event::Readiness). The view is picked up via view_rx.borrow().clone() right before each handle_event, so the render uses the latest value at the moment of any event.

Just thought it would make the tui more simple, less extra handling, but happy to revert to the watch if you think it is better?

zone-sdk/src/sequencer/block_fetch.rs
pradovic: Hmm, I see your point now, maybe this is a bug.

pradovic: You are correct, good find! The SDK does not handle this properly. Let's fix it in a separate, dedicated PR?

pradovic: I think that the SDK does actually rebuilds the chain forward from the restored config tip, unless I am missing something.

When the replayed ChannelConfig lands, channel_tip_at returns the restored `config.id().shed_off_branch_pending` then walks `pending_by_parent` from that new tip and orphans any pending whose parent isn't reachable -> those surface as `OrphanedTx` events with the original payload preserved. The republish path (`publish_parent`→ `create_inscribe_tx`) builds new ops with `parent = restored_config_id`, producing fresh op_ids carrying the same payload bytes, which validate cleanly against the restored tip.

  The literal old inscriptions can't be reused (their wire parent references the obsolete lineage). What gets rebuilt is the forward chain from the restored config tip, carrying the same user-visible payloads attached to new msg_ids. 
  
But it uncovers a bug that the topological sort cycled on same block replay. Do we actually need to topologically sort here? We are kinda trusting the node enough already by design. What do you think? Did I miss something?

pradovic: pls see test cases here and let me know if they make sense, solution can be also to improve topological sort instead of removing it (if we need to preserve the topsort): https://github.com/logos-blockchain/logos-blockchain/pull/2840.

pradovic: If we decide to remove the channel update resetting the tip (per discord discussion), it should be easy to apply it to SDK as well.

hansieodendaal: Yes, I think we should wait for direction there.

**Current block validation code/spec:**
```
chain_tip: | a    | b | c | cfg_1 | d     | e | f | g | cfg_1 | h     | i | j |
parent:    | root | a | b | ----- | cfg_1 | d | e | f | ----- | cfg_1 | h | i |
```
`h` will be accepted because when it is posted, `cfg_1` is the tip
We then get a _logical fork_, `cfg_1 -> d -> ...`  and , `cfg_1 -> h -> ...` and `cfg_1` is a duplicate `MsgId` on the chain.
If it is on the chain, it must be the truth from the sequencer perspective, so maybe when lineage is built, we just need to allow duplicate `MsgId` if the type was config op.

**New direction:** TBD.

testnet/tui-zone/src/lib.rs
hansieodendaal: I think the previous UI update on all channel view changes (which include tip changes) were over eager - what will be meaningful will be to see config changes. I know events now cover that as well, so I think we are good for now.

zone-sdk/src/sequencer/block_fetch.rs
```diff
@@ -0,0 +1,807 @@
...
+        }
+    }
+
+    if items.len() <= 1 {
+        return items;
+    }
+
+    let this_msgs: std::collections::HashSet<MsgId> = items.iter().map(|i| i.this_msg).collect();
+    let by_parent: HashMap<MsgId, &InscriptionInfo> =
+        items.iter().map(|i| (i.parent_msg, i)).collect();
```
hansieodendaal, line 559: This `let by_parent: HashMap<MsgId, &InscriptionInfo>` will break if duplicate `cfg` message id's are present.

hansieodendaal: Maybe we need a cucumber test that replays the same config message a couple of times with some inscriptions in between? After the matter has been resolved in the spec.

zone-sdk/src/sequencer/block_fetch.rs
pradovic: Yes, I think this is covered: https://github.com/logos-blockchain/logos-blockchain/pull/2834#discussion_r3343937885.  Only problem was cycle in toposort and potential change of spec. So either we remove the toposort or improve it to handle cycles.

In any way, I would apply fix from the PR because it fixes cycle problem. If/when we update the spec then we should also update the SDK. 

In any case let's tackle this in a separate PR, as it is not related to refactor.

zone-sdk/src/sequencer/block_fetch.rs
pradovic: Pls see the PR i provided, this should be solved by the PR: https://github.com/logos-blockchain/logos-blockchain/pull/2840/. Let's comment there if you think we need to keep the toposort to check the order of transactions I can rewrite it to cover the cycle as well.

We can apply the fix from the PR, or improve the toposort to be cycle aware if we need to keep the toposort. The question is do we trust the node that tx are in order.

Does that makes sense?

pradovic: > Maybe we need a cucumber test that replays the same config message a couple of times with some inscriptions in between? After the matter has been resolved in the spec.

Yes, let's add it after the spec is settled. In the meantime I think I have similar unit test in the fix PR.

hansieodendaal (approved):
LGTM

PR #2835, opened 2026-06-02
test(e2e): track wallets from block feed

andrussal:

This PR moves Cucumber wallet tracking from repeated point-in-time chain syncs to a block-feed-backed wallet projection.

Instead of each step rebuilding wallet state by scanning the chain again, wallet state is now advanced from observed blocks and kept as scenario state. This is a better general model for E2E tests because the chain itself is the source of truth: wallet balances, reserved inputs, submitted transactions, and future chain-derived state can all be updated from the same stream of block events.

This also keeps the old sync-facing step APIs as compatibility wrappers, so existing scenarios can continue to call the same helpers while the underlying state comes from the block feed.

Why this helps:
- Reduces repeated chain scans during scenarios.
- Keeps wallet state closer to how tests actually observe node progress.
- Makes wallet checks less dependent on ad-hoc “sync before this step” calls.
- Gives us a natural place to plug in future chain-derived collectors.
- Keeps a direct backfill path for cases where the feed history is not enough to safely expand a projection.

```mermaid
flowchart TD
    A["Node HTTP clients"] --> B["DynamicWalletBlockFeedSources"]
    B --> C["ObservationRuntime&lt;BlockFeedObserver&gt;"]
    C --> D["BlockFeed"]

    D --> E["WalletBlockFeedStateCollector"]
    E --> F["WalletBlockFeedTracker.apply_feed"]
    F --> G["WalletSourceTracker"]
    G --> H["WalletChainState"]

    H --> I["Observed wallet outputs"]
    H --> J["Observed wallet spends"]
    H --> K["Current wallet UTXOs"]

    I --> L["TrackedWallets"]
    J --> L
    K --> L

    L --> M["WalletChainStateCache"]
    L --> N["WalletStateView / balances / available UTXOs"]

    O["Cucumber wallet step"] --> P["observe_* wallet function"]
    P --> Q["best node / source selection"]
    Q --> R["TrackedWalletKeysBySource"]
    R --> S["TrackedWalletKeysForSource"]
    S --> T["WalletFeedTrackingBatch"]
    T --> F
    P --> N

    F --> U{"Feed has needed blocks?"}
    U -- yes --> N
    U -- no --> V["Direct HTTP backfill"]
    V --> W["replace_source_state"]
    W --> L

    Bn["Mutable source registry"]
    Dn["Shared latest heads + retained blocks"]
    Hn["Chain-derived wallet state for tracked keys"]
    Rn["Wallet IDs + public keys grouped by logical source"]
    Ln["Main cucumber wallet state"]
    Vn["Fallback when feed history is missing"]

    B -.-> Bn
    D -.-> Dn
    H -.-> Hn
    R -.-> Rn
    L -.-> Ln
    V -.-> Vn

    classDef chain fill:#e8f1ff,stroke:#3b6ea8,stroke-width:2px,color:#0b2545;
    classDef wallet fill:#eaf7ea,stroke:#3f8f46,stroke-width:2px,color:#103b16;
    classDef step fill:#fff0e3,stroke:#c7772f,stroke-width:2px,color:#4a2505;
    classDef fallback fill:#fdecec,stroke:#b84a4a,stroke-width:2px,color:#4a0707;
    classDef note fill:#fff9d8,stroke:#b7a13a,stroke-dasharray:4 3,color:#3a3200;

    class A,B,C,D chain;
    class E,F,G,H,I,J,K,L,M,N,R,S,T wallet;
    class O,P,Q step;
    class U,V,W fallback;
    class Bn,Dn,Hn,Rn,Ln,Vn note;
```

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


tests/src/cucumber/wallet/sync.rs
```diff
@@ -179,59 +236,327 @@ fn build_wallet_sync_requests(
...
+            ),
+        }
+    }
+
+    fn add_wallet_keys(
+        &mut self,
+        min_height: u64,
+        source_wallet_keys: &TrackedWalletKeysForSource,
+    ) {
+        self.min_height = self.min_height.max(min_height);
```
hansieodendaal: Perhaps `min` here?
```suggestion
        self.min_height = self.min_height.min(min_height);
```

tests/src/common/wallet/chain/state.rs
```diff
@@ -0,0 +1,360 @@
...
+                        utxo,
+                    });
+                }
+            }
+        }
+
+        observed_outputs
+    }
+
+    fn apply_spent_notes(&mut self, tx: &SignedMantleTx) -> Vec<WalletObservedSpend> {
```
hansieodendaal: Should we maybe have placeholders at least for the other op types? I think we already use channel withdraw and channel config op in the tests.

hansieodendaal (approved):
LGTM, nice work!

I will rebase #2854 on top of this when merged and re-assess large stress test performance. We may need to keep some form of cache-based available UTXO set to speed up the large stress tests. The follow-up intent after #2854 is to be able to send many concurrent transactions when the wallet state is _almost/hopefully always up to date_ as implemented by this PR.In #2854 I could manage ~6 txs/s, still way too little to produce may full blocks.

tests/src/cucumber/wallet/sync.rs
andrussal: Good catch, fixed

PR #2868, opened 2026-06-04
fix(chain-leader): respect tx topological order under block size limit

davidrusu:

`txs_for_block` enforced the block size/count limit with `continue`, skipping a too-big transaction and scanning on. Since candidates are in dependency order, this could drop a dependency while keeping its dependent — an invalid block. It now truncates at the first transaction that trips the limit (longest fitting prefix), preserving order.

Yes — change is local to `chain-leader/src/lib.rs`, with a doc comment and tests covering the prefix invariant. No external spec applies.

@davidrusu.

N/A.

No.

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/services/chain/chain-leader/src/lib.rs
+++ b/services/chain/chain-leader/src/lib.rs
@@ -601,7 +601,21 @@ where
 
         // Collect all candidate transactions up front so the ones that fail can
         // be retried across multiple rounds.
-        let mut pending: Vec<_> = tx_stream.collect().await;
+        let candidate_txs: Vec<_> = tx_stream.collect().await;
+
+        // A transaction larger than the maximum block size can never be included
+        // in any block. Evict such transactions up front so they cannot stall
+        // block assembly; any of their dependents will fail to apply below and be
+        // evicted in turn.
+        let mut pending = Vec::with_capacity(candidate_txs.len());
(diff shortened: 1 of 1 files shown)


services/chain/chain-leader/src/lib.rs
```diff
@@ -785,11 +787,11 @@ where
 
         let tx_size = tx.storage_size();
         let Some(next_block_size) = block_size.checked_add(tx_size) else {
-            continue;
+            break;
```
danielSanchezQ, line 790: Actually i think this should still be continue, as any other smaller tx may fit ?

davidrusu: see the PR description, the valid tx stream is ordered and potentially contains dependencies.

skipping one tx could invalidate txs in the rest of the stream. But validation happened before we enforce the block size / tx count limit.

youngjoon-lee (approved):
Oh, sorry, I missed this point when reviewing the previous PR. Thank you.

andrussal (approved):
I think `break` is right here. But it can break if the first tx is oversized. Made a separate fix in https://github.com/logos-blockchain/logos-blockchain/pull/2870

PR #2872, opened 2026-06-05
feat(zone-sdk): refactor events/checkpoints + publish

pradovic:

The PR refactors the events API to return a checkpoint alongside every state-mutating event, so consumers can persist the contained fields plus the checkpoint atomically. Requested by LEZ to simplify their atomic state-save flow. The PR changset is on the large side as the API changed so it propagated to consumers (tui & tests).

  The events that carries checkpoint: `BlockProcessed` fires per live block and carries a `Vec<FinalizedTx>` and a `ChannelUpdate`. Either may be empty. Fires during back fill as well.

  Public handle methods (`publish`, `submit_signed_tx`, `channel_config`) are now sync and return the resulting `PublishResult` plus checkpoint inline.

  Posting lifecycle tightened: `publish` and `publish_atomic_withdraw` only post when it's our turn. Turn-change to ours triggers a full pending resubmit. The block-cadence resubmit was dropped — the 30s self-heal tick covers eviction recovery.

  Tests refactored to use the new SDK API idiomatically

Yes

@pradovic 

Yes

No
## Checklist

> [!WARNING]  

* [ ] 2. Description added.


tests/src/cucumber/steps/manual_zone/runner.rs
```diff
@@ -0,0 +1,353 @@
...
+/// the policy inline. Events reach observers via the SDK's own
+/// broadcast channel (subscribed in [`spawn`]) — the drive loop does
+/// not forward them, so there's no backpressure on the runner from
+/// idle subscribers.
+///
+/// Pulled out as a free async function so callers can spawn it
+/// themselves (e.g. onto a custom runtime, a `LocalSet`, or interleaved
+/// with other state in the same task). [`spawn`] is the convenience
+/// wrapper that creates channels + a tokio task.
+pub(super) async fn run<Node, P>(
```
pradovic, line 248: @hansieodendaal  We could expose this kind of run() method and Policy trait to our async API (sequencer). This would enable spawning the drive loop task for user while user supply the policy for updating it's state on events without losing the state machine correctness.  

This could be a orthogonal API for user that don't want to poll the loop themselves and don't mind additional task. We would also expose needed shutdown hooks, etc... Maybe more idiomatic for some users. Even without the run() method, just encapsulating commands and supporting them in SDK would allow t he users to call it outside of the drive loop, as long as it is driven.

Not for this PR, but just leaving it out for the future maybe. Let me know what you think.

hansieodendaal: We should add an issue for this for consideration

pradovic: Good point, I will open a issue

pradovic: https://github.com/logos-blockchain/logos-blockchain/issues/2875

zone-sdk/src/sequencer/handle.rs
```diff
@@ -261,34 +316,149 @@ where
...
+            .node
+            .channel_state(self.sequencer.channel_id)
+            .await
+            .map_err(|e| Error::Network(format!("channel_state query failed: {e}")))?
+            .ok_or_else(|| {
+                Error::Network(format!(
+                    "publish_atomic_withdraw requires channel state for {:?}",
+                    self.sequencer.channel_id
+                ))
+            })?;
```
hansieodendaal: I think we should use the cached channel state here, which is already processed with every new block, then this function can become synchronous `pub fn publish_atomic_withdraw(`. 

`self.sequencer.channel_state` must have become available already.

If `channel_state` is slow or blocked, `sequencer.next_event()` is not being polled.

zone-sdk/src/sequencer/zone_sequencer.rs
```diff
@@ -347,10 +310,10 @@ where
...
     /// Drive the sequencer one step. Returns `None` when the step completed
     /// without emitting a public event (e.g. an empty backfill batch, a
-    /// failed reconnect attempt that already slept, a `PrepareTx` /
-    /// `SignTx` / `SubmitSignedTx` request handled via oneshot). Internal —
-    /// public callers use [`Self::events`], which loops on `None`.
-    pub(super) async fn next_event(&mut self) -> Option<Event> {
+    /// failed reconnect attempt that already slept, a periodic resubmit
+    /// tick). Designed for `tokio::select!` so the drive task can interleave
+    /// publish calls via [`Self::handle`] between turns.
+    pub async fn next_event(&mut self) -> Option<Event> {
```
hansieodendaal, line 335: Although the bahaviour is documented clearly: `publish`, `submit_signed_tx`, and `channel_config` queue a future into `in_flight`, and the actual `post_transaction` runs only when the drive loop later polls `ZoneSequencer::next_event()`, a direct SDK consumer can still do this:
1. Call `sequencer.handle().publish(...)`.
2. Persist the returned result/checkpoint.
3. Wait for inclusion/finalization before returning to the drive loop.
4. Stall forever because the queued post future has never been polled.

We should decouple post-driving from `next_event()` polling for user-initiated publishes. Right now `Ok((result, checkpoint))` can be interpreted as “submitted enough to await inclusion”, while in reality the tx may not even be posted yet unless the caller promptly resumes driving `next_event()`. That coupling creates an easy self-deadlock/forward-progress footgun for SDK consumers.

Preferred fix: run publish posts on an independent poster worker/task (or equivalent background executor), and report completion back into sequencer state.

zone-sdk/src/sequencer/zone_sequencer.rs
```diff
@@ -403,6 +359,55 @@ where
...
+
+    /// Push a single-tx publish post into `in_flight`. Used by
+    /// `handle.publish` / `submit_signed_tx` / `channel_config` —
+    /// independent user-initiated actions that can run concurrently.
+    pub(super) fn queue_publish_post(&self, tx_hash: TxHash, signed_tx: SignedMantleTx) {
+        self.in_flight.push(Box::pin(post_batch(
+            self.node.clone(),
+            vec![(tx_hash, signed_tx)],
+        )));
+    }
```
hansieodendaal: `queue_publish_post` is invoked from handle paths (`publish`, `submit_signed_tx`, `channel_config`), but the transaction is not marked as “queued/in-flight” in state before being pushed.

`mark_pending_inscription_posted(&mut self, tx_hash: &TxHash)` only flips `pending.posted = true` after a successful post result. `PendingInscription` has no separate `queued`/`in_flight` flag, so state is effectively binary (`posted` vs `not posted`).

This leaves a duplication window:
1. We enqueue a single-transaction post future.
2. Before that future completes, the transaction is still `posted == false`.
3. `resubmit_interval` tick or turn-change triggers `resubmit_pending()`.
4. The sweep sees the same transaction as unposted/eligible and queues it again.

Result: duplicate `post_transaction` attempts can occur while the first attempt is still in flight.

Suggested fix:
- Track three states instead of two:
  - pending (not queued),
  - queued/in-flight,
  - posted/accepted.
- Or maintain a separate `in_flight_hashes: HashSet<TxHash>` and have resubmit skip hashes already in flight.

zone-sdk/src/sequencer/actor.rs
```diff
@@ -44,77 +25,102 @@ where
...
+    /// optional `ChannelUpdate`, and the block's finalized txs.
     pub(super) async fn handle_stream_item(
         &mut self,
         maybe_event: Option<ProcessedBlockEvent>,
     ) -> Option<Event> {
         let Some(block_event) = maybe_event else {
             warn!(target: TARGET, "Blocks stream disconnected, will reconnect on next call");
             self.blocks_stream = None;
-            return self.signal_not_ready();
+            self.handle_stream_drop();
```
hansieodendaal, line 38: On stream drop, `handle_stream_drop()` only calls `publish_turn_to_write(false)`; it does not invalidate `slot_clock`, `channel_state`, or `own_key_index`.

As a result, `can_publish_inscription_now()` can still evaluate using stale data during reconnect, and `publish()` can queue an immediate post based on that stale turn decision.

We could add an explicit connected/stream-valid flag: set it `false` on stream drop, set it `true` only after reconnect/catch-up, and check it in `can_publish_inscription_now()` so all callers inherit the guard.

tests/src/cucumber/steps/manual_zone/runner.rs
```diff
@@ -0,0 +1,353 @@
...
+use lb_zone_sdk::{adapter, sequencer::ZoneSequencer};
+use tokio::{
+    sync::{broadcast, mpsc, oneshot, watch},
+    task::JoinHandle,
+};
+
+/// Inline policy executed on the drive task for each SDK event before the
+/// event is forwarded to test observers.
+///
+/// Implementations may call `sequencer.handle().publish(...).await` to issue
```
hansieodendaal:
```suggestion
/// Implementations may call `sequencer.handle().publish(...)` to issue
```

zone-sdk/src/sequencer/handle.rs
```diff
@@ -2,181 +2,239 @@ use lb_core::mantle::{
...
+    /// in progress (the sequencer hasn't emitted [`super::Event::Ready`]
+    /// yet). After the first `Ready`, publishes are always accepted:
+    /// during a mid-life reconnect the tx is queued locally and posted
+    /// when the stream resumes (or when our turn comes back). To wait for
+    /// readiness asynchronously, subscribe via
+    /// [`ZoneSequencer::subscribe_ready`].
+    pub fn publish(
+        &mut self,
+        data: Inscription,
+    ) -> Result<(PublishResult, SequencerCheckpoint), Error> {
```
hansieodendaal, line 90: I’m concerned about the return contract/meaning of handle methods like `publish()`.

`publish()` constructs `InscriptionInfo { tx_hash, parent_msg, this_msg, payload }` and returns it immediately in `PublishResult` together with a checkpoint, regardless of whether `post_transaction` has succeeded (or even run) yet.

At return time this is a local acceptance/outbox state transition, not a network submission guarantee. With the current naming (`publish`, `PublishResult`, `PublishedTx`), this is easy to interpret as “submitted enough to wait for inclusion,” which can be misleading.

Could we make this state explicit in the API shape/naming (e.g., queued/enqueued terminology) or return an explicit status phase so callers don’t treat it as post success? Example:
- `PublishResult` -> `QueuedPublish` or `EnqueueResult`
- `PublishedTx` -> `QueuedTx` or `PendingTx`

hansieodendaal (commented):
As all tests are currently passing, I reviewed this from a potential blocking/interlock/stall perspective. Some comments below for consideration, otherwise looking good.

zone-sdk/src/sequencer/zone_sequencer.rs
pradovic: This is exactly what https://github.com/logos-blockchain/logos-blockchain/pull/2872#discussion_r3362220596 was trying to show.

I think for now the good-enough contract is: the event loop must be driven for publish-related work to make progress. There are two ways a user can call publish:

1. Inside the drive loop — publish mutates state in place and the post is queued onto the same state-machine cycle that the loop is already advancing.
2. Outside the drive loop — the call would need to feed an event back into the loop (e.g. via a channel into a tokio::select! arm) so the loop picks it up on its next iteration.

We can support both more cleanly in the future (one with the run() loop and another with manual driving), but the current behaviour matches what was already true before this PR — the drive loop has always needed to be polled for publish-related work to make progress.

This PR doesn't introduce a new failure mode on this front; it just makes the queue-into-in_flight step more visible because there are now multiple call sites doing it.

zone-sdk/src/sequencer/zone_sequencer.rs
pradovic: The overlap between publish and resubmit_pending is intentional. Until a transaction is observed a successfully posted, it remains eligible for resubmission. This can result in a short window where the initial post attempt and a resubmit overlap, but the node deduplicates by transaction hash so the effect is limited to an extra RPC call. We could introduce an explicit InFlight state to eliminate that overlap, but that would significantly complicate the state machine for what is currently a benign retry behavior.

Does that make sense?

zone-sdk/src/sequencer/handle.rs
pradovic: Good point, renaming PublishedTx -> PendingTx.

I think publish here means accepted by the zone-sdk, so I would keep PublishResult for now as the result of the publish call.

Longer term, we could make the phases explicit, e.g. Accepted/Queued -> Posted/Pending -> Safe -> Finalized. But I’d prefer to tackle that in a dedicated PR, since this PR does not change the existing behaviour: we also did not previously guarantee that the tx had been successfully posted to the mempool before returning from publish.

zone-sdk/src/sequencer/handle.rs
pradovic: Good point, addressed in the latest commit!

zone-sdk/src/sequencer/zone_sequencer.rs
hansieodendaal: Let's make an issue for this then?

zone-sdk/src/sequencer/zone_sequencer.rs
hansieodendaal: Maybe _maintain a separate in_flight_hashes: HashSet<TxHash> and have resubmit skip hashes already in flight_ then?

zone-sdk/src/sequencer/handle.rs
hansieodendaal: Yeah, let's make an issue and tackle it in a next PR.

zone-sdk/src/sequencer/zone_sequencer.rs
pradovic: I have already created an issue, as suggested, in the original discussion?
https://github.com/logos-blockchain/logos-blockchain/pull/2872#discussion_r3362220596

I think publish calls being inside the drive loop have nice advantage:
- the user call publish exactly on this state transitions so the flow with ChannelUpdate works as it should


When the user starts to work outside of the loop, for example if we had a run method and just exposed events, user could have weird republishing scenarios where the state is advancing while it is trying to pusblish, and it could publish on the wrong state and have double publishing. The intent of the publish is to publish on the current state of state machine. If the future states invalidates it it will come through channel update.

hansieodendaal (approved):
LGTM

zone-sdk/src/sequencer/handle.rs
pradovic: https://github.com/logos-blockchain/logos-blockchain/issues/2900.

I think this is not critical, submitted -> pending is a clean win. Safe definition we should probably leave for later and plan it better.

hansieodendaal (approved):
LGTM

PR #2893, opened 2026-06-08
feat(blend): do not pick new peers for failed message attempts

ntn-x2:

Based on internal discussions with @madxor, and also given we saw his node spamming our devnet with infinite retries for a single message attempt, we decided it's enough to stop trying to send a message if the message cannot be sent to the picked peer after the configured number of attempts. Since the redundancy is also configurable, we can increase that value to have more peers being randomly picked for each message instead. This simplifies the edge swarm code quite a bit and prevent infinite retry loops.

Yes.

@ntn-x2 @madxor 

Yes.

No.


youngjoon-lee (approved):
Yeah, makes sense 👍

PR #2896, opened 2026-06-09
fix(chain): use rpds in Cryptarchia for prepare-commit pattern

youngjoon-lee:

The parent PR #2813 is changing the way how chain service mutates `Cryptarchia`. It clones `Cryptarchia`, mutates it, persists artifacts to the storage, and replaces the original `Cryptarchia` with the cloned/mutated one only if the storage returned `Ok`. 

However, cloning `Cryptarchia` is not always cheap because it holds `HashMap<HeaderId, LedgerState>` which may be large especially when bootstrapping. Inside `LedgerState`, we use `rpds`s because we designed `LedgerState` to be cloned frequently. In contrast, we didn't design `Cryptarchia` to be cloned many times. (Yeah, we should have not make it `Clone`).

Now, because it's unavoidable to clone `Cryptarchia` in the PR #2813, I am replacing all `HashMap`s with `rpds` in this PR.

One concern is the insertion cost of rpds. However, the additional rpds structures we’re considering are likely negligible compared to the existing ones in LedgerState, which are updated on a per-tx (or per-UTXO) basis rather than per block.

Here are some simple benchmarks. Insertion/removal are definitely heavier. But, in absolute terms, the cose is still 200~800ns.
- Insert into a map of size N
  |N|HashMap|HashTrieMapSync|ratio|
  |-|-|-|-|
  |100|70 ns|87 ns|1.2x|
  |1000|153 ns|316 ns|2x|
  |10000|145 ns|479 ns|3x|
- Remove + insert into a set of size N
  |N|HashSet|HashTrieSetSync|ratio|
  |-|-|-|-|
  |100|110 ns|180 ns|1.6x|
  |1000|106 ns|234 ns|2.2x|
  |10000|125 ns|1.0 **us**|8x|
  

Yes

@youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4451,6 +4451,7 @@ version = "0.1.2"
 dependencies = [
  "logos-blockchain-pol",
  "logos-blockchain-utils",
+ "rpds",
  "serde",
  "serde_with",
  "thiserror 2.0.18",
--- a/consensus/cryptarchia-engine/Cargo.toml
+++ b/consensus/cryptarchia-engine/Cargo.toml
@@ -19,6 +19,7 @@ workspace = true
 [dependencies]
 lb-pol     = { workspace = true }
 lb-utils   = { features = ["time"], workspace = true }
+rpds       = { workspace = true }
 serde      = { workspace = true }
 serde_with = { features = ["macros"], workspace = true }
 thiserror  = { workspace = true }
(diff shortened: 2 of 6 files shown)


danielSanchezQ (commented):
This looks good @youngjoon-lee. I dont think insertion times are a problem, so we should be fine with it 😉

youngjoon-lee:
> This looks good @youngjoon-lee. I dont think insertion times are a problem, so we should be fine with it 😉

@danielSanchezQ I just marked this PR ready. Could you please approve this again? :)

danielSanchezQ (approved):

PR #2904, opened 2026-06-09
test(blend): reproduce duplicate-replica decapsulation panic

davidrusu:

Adds a regression test reproducing a liveness bug found during a blend-service audit: two replicas of a single local data message panic the core blend service task, halting blending for the whole node.

The core service emits each local data message as `data_replication_factor + 1` independently-encapsulated copies, each with fresh random layers (so distinct encapsulated IDs that the swarm dedup cache does not collapse). When two replicas are fully decapsulated at the same final node, they yield the same inner `NetworkMessage`; `ProcessedMessage` derives `Eq`/`Hash` on that content, so the second insert into `unsent_processed_messages` returns `Err` and the `.expect("Swarm should bubble up unique messages only.")` panics.

```text
  replica A ─encap(rand)→ ID_a ─┐ swarm dedups on ID, so both pass
  replica B ─encap(rand)→ ID_b ─┘
                                 │  decapsulate at same final node
                                 ▼
       both → ProcessedMessage::Network(same bytes)   (Eq/Hash on content)
                                 │
                                 ▼  add_unsent_processed_message(..)
           A: Ok ──► inserted        B: Err ──► .expect(..) panics ✗
```

The test (`test_duplicate_decapsulated_replica_panics_core_service`) uses a size-1 membership to force local full decapsulation, encapsulates one payload twice (asserting the two encapsulations differ so the swarm would not dedup them), and feeds both through `handle_incoming_blend_message`; the second panics. It is marked `#[should_panic(expected = "Swarm should bubble up unique messages only.")]`, so it currently passes as confirmed evidence of the crash. When the bug is fixed (treat a duplicate as already-known rather than an invariant violation), this should be flipped into a positive assertion that the second replica is handled gracefully.

This PR contains **only the test** — no production behavior change.

The test's doc comment carries the full failure mechanism, including the diagram above. Relevant production sites:
- replica emission: `services/blend/src/core/mod.rs` (`handle_serialized_local_data_message`, ~`:802-805`)
- panic site: `services/blend/src/core/mod.rs:1516-1519` (`handle_decapsulated_incoming_message_from_current_epoch`)
- content-based `Eq`/`Hash`: `services/blend/src/message.rs`
- swarm dedup keyed on encapsulated ID: `blend/network/src/core/with_core/behaviour/message_cache.rs`

_Specification link: TODO — please add the relevant blend spec section._

@davidrusu accountable. Reviewers / blend spec authors: _TODO — add._

N/A for this PR — it only adds a failing-scenario regression test and asserts no spec change. The follow-up fix PR should reconcile the intended dedup/replication behavior with the spec.

No. Test-only; no production code or spec changes.

## Checklist

* [x] 2. Description added.
* [ ] 3. Context and links to Specification document(s) added. _(spec link TODO)_
* [ ] 4. Main contact(s) (developers and specification authors) added _(reviewers TODO)_
* [x] 5. Implementation and Specification are 100% in sync including changes. _(test-only, no spec impact)_
* [x] 7. New or changed logs were reviewed against the logging guidelines. _(no log changes)_

## Test plan

- `cargo test -p logos-blockchain-blend-service --lib test_duplicate_decapsulated_replica_panics_core_service` → `should panic ... ok`

### DIFF
--- a/services/blend/src/core/mod.rs
+++ b/services/blend/src/core/mod.rs
@@ -1304,14 +1304,18 @@ where
     };
     state_updater.collect_current_epoch_tokens(blending_tokens.into_iter());
 
+    scheduler.schedule_processed_message(processed_message.clone());
     // We treat a partially or fully decapsulated message as a processed message,
     // and we schedule for its release at the next release round.
-    scheduler.schedule_processed_message(processed_message.clone());
-    assert_eq!(
-        state_updater.add_unsent_processed_message(processed_message.clone()),
-        Ok(()),
-        "There should not be another copy of the same locally-generated processed message: {processed_message:?}."
-    );
+    if state_updater
+        .add_unsent_processed_message(processed_message.clone())
+        .is_err()
+    {
+        tracing::warn!(
+            target: LOG_TARGET,
+            "There should not be another copy of the same locally-generated processed message: {processed_message:?}."
+        );
+    }
     state_updater.commit_changes()
 }
 
--- a/services/blend/src/core/mod.rs
+++ b/services/blend/src/core/mod.rs
@@ -1513,10 +1513,15 @@ where
         cryptographic_processor,
     );
 
-    if let Some(processed_message) = maybe_processed_message {
-        state_updater
+    if let Some(processed_message) = maybe_processed_message
+        && state_updater
             .add_unsent_processed_message(processed_message)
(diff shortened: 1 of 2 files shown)


youngjoon-lee (approved):
👍

danielSanchezQ (approved):

PR #2908, opened 2026-06-10
test(blend-network): characterize edge connection leak in Starting state

davidrusu:

Adds a characterization test reproducing a liveness bug found during a blend-service audit (finding #9): a core→edge connection that establishes but never opens its inbound substream is **never closed**, leaking the connection.

The handler starts in `Starting`, whose `poll` only stores the waker and returns `Pending` — it **never arms a timer**. The `connection_timeout` is consumed only when transitioning to `ReadyToReceive` on `FullyNegotiatedInbound` (which *does* poll a `futures_timer::Delay`). Meanwhile `connection_keep_alive()` returns `true` for every state except `Dropped`, so the swarm keeps the connection open. A peer that completes the connection but never opens the inbound substream therefore pins the handler in `Starting` forever.

The behaviour's `max_incoming_connections` cap only counts **upgraded** peers (`upgraded_edge_peers`, populated on `SubstreamOpened` at `with_edge/behaviour/mod.rs:132`; gate at `:191`), so these stuck connections bypass it entirely — an unbounded, remote-triggered connection leak (FD/memory exhaustion).

### Test style — please weigh in

Unlike the panic repros (#2904 finding #1, #2907 finding #2) which use `#[should_panic]`, this bug is a **silent leak**, so there's nothing to catch. The test is a **characterization test**: it asserts the *current, buggy* behavior — long after the connection timeout elapses, `poll` is still `Pending` and `connection_keep_alive()` is still `true`. The doc comment states plainly that this documents a bug. The valuable property is the same as `#[should_panic]`: it **flips when fixed** (once `Starting` arms the timeout, `poll` yields `SubstreamClosed(Timeout)` and keep-alive drops), at which point the assertions should be updated to assert the correct behavior.

Timing note: it uses a real 50ms timeout + 500ms `tokio::time::sleep` (the state machine uses `futures_timer::Delay`, real wall-clock, not tokio's mock clock). On current code timing is irrelevant (no timer exists → always `Pending`), so it's deterministic; the 10× margin only matters for the flip-on-fix direction. Runs in ~0.5s.

This PR contains **only the test** — no production behavior change. The fix direction is to arm the timeout in `Starting` as well (or otherwise bound the un-negotiated lifetime), and have the cap account for not-yet-upgraded connections.

The test's doc comment carries the full mechanism. Relevant production sites:
- no timer armed: `blend/network/src/core/with_edge/behaviour/handler/starting.rs` (`poll`)
- timeout only on transition: same file, `on_connection_event` → `ReadyToReceiveState::new(Box::pin(Delay::new(connection_timeout)), ..)`
- keep-alive: `handler/mod.rs:157-159`
- cap counts only upgraded peers: `with_edge/behaviour/mod.rs:107-111, 132, 191`

_Specification link: TODO._

@davidrusu accountable. Reviewers / blend spec authors: _TODO._

N/A — test-only; asserts no spec change. The follow-up fix PR should reconcile the edge connection lifecycle with the spec.

No. Test-only.

## Checklist

* [x] 2. Description added.
* [ ] 3. Context and links to Specification document(s) added. _(spec link TODO)_
* [ ] 4. Main contact(s) (developers and specification authors) added _(reviewers TODO)_
* [x] 5. Implementation and Specification are 100% in sync including changes. _(test-only)_
* [x] 7. New or changed logs were reviewed against the logging guidelines. _(no log changes)_

## Test plan

- `cargo test -p logos-blockchain-blend-network --lib starting_state_without_inbound_substream_leaks_connection` → `ok`

### DIFF
--- a/blend/network/src/core/with_edge/behaviour/handler/mod.rs
+++ b/blend/network/src/core/with_edge/behaviour/handler/mod.rs
@@ -22,6 +22,9 @@ mod ready_to_receive;
 mod receiving;
 mod starting;
 
+#[cfg(test)]
+mod tests;
+
 const LOG_TARGET: &str = blend::network::core::handler::CORE_EDGE;
 
 type TimerFuture = Pin<Box<dyn Future<Output = ()> + Send>>;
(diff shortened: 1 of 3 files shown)


danielSanchezQ (approved):

youngjoon-lee (approved):
👍

PR #2883, opened 2026-06-08
fix(wallet): recover from stale wallet snapshots

andrussal:
## 1. What does this PR implement?

This PR fixes wallet startup recovery when the persisted wallet snapshot belongs to a stale LIB.

On startup, the wallet service now compares the recovered wallet LIB with the current chain LIB. If they differ, it ignores the stale wallet snapshot, rebuilds wallet state from the current chain LIB ledger, and immediately persists the corrected recovery state so the same stale snapshot is not reused on the next restart.

Related to #2810.

## 2. Does the code have enough context to be clearly understood?
Yes

## 3. Who are the specification authors and who is accountable for this PR?
@andrussal 

## 4. Is the specification accurate and complete?
N/A

## 5. Does the implementation introduce changes in the specification?
N/A

## Checklist

* [x] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [x] 2. Description added.
* [x] 3. Context and links to Specification document(s) added.
* [x] 4. Main contact(s) (developers and specification authors) added
* [x] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 6. Link PR to a specific milestone.
* [x] 7. New or changed logs were reviewed against the [logging guidelines](https://github.com/logos-blockchain/logos-blockchain/blob/master/CONTRIBUTING.md#logging-guidelines).

PR #2909, opened 2026-06-10
fix(wallet): revert "fix(wallet): recover from stale wallet snapshots (#2883)"

youngjoon-lee:

I'm reverting the PR #2883 which is wrong. I'm sorry that I missed an important point when reviewing the PR.

The PR tried to fix the issue https://github.com/logos-blockchain/logos-blockchain/issues/2810, but it actually doesn't solve the issue. I'm investigating the issue.

Anyway, the PR #2883 should be reverted because it clears the previously-accumulated state (utxos, vouchers, etc.) and starts with the empty state from the current LIB. That is wrong. The wallet service must always track all blocks from genesis to the current tip (even if chain reorged) to gradually update the state. Even if the wallet service's recovered state is outdated compared to the current chain, it must catch up the chain by running backfilling (which is already implemented). The wallet service shouldn't give up the state that it has maintained so far.

So, I'm reverting the change. And, I will investigate the issue reported in detail, and open fixes soon.

Yes

@youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/services/wallet/src/states.rs
+++ b/services/wallet/src/states.rs
@@ -6,7 +6,6 @@ use lb_ledger::LedgerState;
 use lb_wallet::{Vouchers, WalletBlock, WalletError, WalletState};
 use overwatch::services::state::StateUpdater;
 use serde::{Deserialize, Serialize};
-use tracing::warn;
 
 use crate::{KeyId, WalletServiceError, WalletServiceSettings};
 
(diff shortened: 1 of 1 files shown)


andrussal (approved):
Ah I see. We keep the known vouchers, but rebuilding from the LIB ledger clears the voucher paths/snapshots,

youngjoon-lee:
> Ah I see. We keep the known vouchers, but rebuilding from the LIB ledger clears the voucher paths/snapshots,

Yes. In the PR #2883, the wallet service **jumps** to the **current** LIB which is newer than the **persisted/recovered** LIB (aka. **stale LIB**, in your word), with giving up the state that it has accumulated so far. Instead, the wallet service must backfill missing blocks starting from the stale block to the current tip. The current backfilling implementation should work because it requests the chain service to fetch missing block IDs from storage (not from memory), and because the wallet service loads corresponding blocks from storage directly.

danielSanchezQ (approved):

PR #2911, opened 2026-06-10
chore: add curl to node image

vpavlin:
Adds `curl` to the node Docker image so it's available for health checks and debugging inside the container (e.g. querying the local HTTP API on port 8080).

## Change

```dockerfile
RUN apt-get update && apt-get install -y --no-install-recommends curl && rm -rf /var/lib/apt/lists/*
```

Uses `--no-install-recommends` and cleans up the apt cache to keep the image lean.


Dockerfile
```diff
@@ -31,6 +31,11 @@ LABEL maintainer="augustinas@status.im" \
     source="https://github.com/logos-blockchain/logos-blockchain" \
     description="Logos blockchain node image"
 
+RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates curl jq && \
+    curl -sL https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 -o /usr/local/bin/yq && \
```
bacv: Why not https://packages.debian.org/trixie/yq? If you used the package from the debian repo then it would also work on arm.

vpavlin: good question, let me punish Claude properly!

bacv:
Now that the PR looks good, could explain how these tools are necessary in the Akash network case?

The reason I'm asking is because what if we target some different platform, and need to add even more tools?
I personally would like to keep the node image as simple as possible, if we didn't depend on libc, it could even be a scratch image.

Do you think it would be possible to create a new image specific to Akash Network? We build node docker image after the release happens, the same could be done for a new Akash network image, this way users could choose. The Akash network image could be layered on top of the logos blockchain image?

vpavlin:
Yeah, sorry for not explainign properly - the reason is that if you run the image locally, you'd use `curl CONTAINER_IP:8080...`, but we do not want to expose port 8080 to the public on public cloud - so for anyone to "debug" and check the state - we need some way - and the simplest would be to have curl inside. THe other tools are for "config" - I am not sure/convinced `jq` is needed - we can get rid of that. `yq` is there to update `user_config.yaml` with actualy container public IP - for it to announce the right addr.

That said, yes, we could have a new image building on top of this one, but that also means chaining the rebuilds etc. Not sure if it is worth it - I can try to get rid of yq and use sed to modify the user_config? Which would mean just curl as a debug/ops tool- that should be relatively harmless?

Another option is we run 2 containers - LB node + reverse proxy with basic auth which guards the 8080 - we did this for Codex when I added it to the Akash repo

Whatever is your preference:)

bacv:
> That said, yes, we could have a new image building on top of this one, but that also means chaining the rebuilds etc

Agree, in this case it's better to just add these tools. If jq is required in the Akash case, lets leave it, otherwise this looks good. If in future we will need to add more tools, I'll look into chaining the docker images.

bacv (approved):

vpavlin:
That is the question - I think it is not required:D Claude dug into K8S API thinking it can derive the public IP automatically, but it cannot  for most compute providers, so I think it is pointless:) I'll remove jq and only keep curl (for state checks) and yq (for easier user_config editing)

bacv (approved):

PR #2916, opened 2026-06-10
test(e2e): use feed-backed wallet state

andrussal:

This PR removes the remaining explicit wallet balance sync steps from Cucumber scenarios and uses the existing block-feed-backed wallet state directly.

Wallet reads now wait until the feed has seen the relevant node height before checking in-memory state. This keeps assertions deterministic without calling balance refresh commands from scenarios. The fork/join test also now waits for the actual condition we care about: submitted transactions are settled and wallet state is visible before the forks are bridged. It no longer waits for extra random blocks on each fork.

The manual `BALANCE*` commands were removed because they only refreshed wallet state as a side effect. Scenarios should use assertion steps instead.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/tests/cucumber_tests/features/manual_control.feature
+++ b/tests/cucumber_tests/features/manual_control.feature
@@ -7,6 +7,10 @@ Feature: Manual control of transactions
   # Supported commands (one per line):
   #   COIN_SPLIT, wallet '<wallet_name>', outputs <count>, value <amount>
   #   VERIFY, wallet '<wallet_name>', outputs <count>, time_out <duration_seconds>
+  #   BALANCE, wallet '<wallet_name>'
+  #   BALANCE_ALL_WALLETS
+  #   BALANCE_ALL_USER_WALLETS
+  #   BALANCE_ALL_FUNDING_WALLETS
   #   CLEAR_ENCUMBRANCES, wallet '<wallet_name>'
   #   CLEAR_ENCUMBRANCES_ALL_WALLETS
   #   SEND, transactions <count>, value <amount>, from '<wallet_name>', to '<wallet_name>'
@@ -32,6 +36,7 @@ Feature: Manual control of transactions
   #   VERIFY_MAX, wallet 'WALLET_1A', wallet_state_type 'encumbered', outputs 0, time_out 60
   #   VERIFY_MAX, wallet 'WALLET_2A', wallet_state_type 'encumbered', outputs 0, time_out 60
   #   SEND, transactions 5, value 100, from 'WALLET_1A', to 'WALLET_2A'
+  #   BALANCE, wallet 'WALLET_1A'
   #   SEND, transactions 5, value 100, from 'WALLET_2A', to 'WALLET_1A'
   #   VERIFY_MAX, wallet 'WALLET_1A', wallet_state_type 'encumbered', outputs 0, time_out 60
   #   VERIFY_MAX, wallet 'WALLET_2A', wallet_state_type 'encumbered', outputs 0, time_out 60
--- a/tests/cucumber_tests/features/manual_control.feature
+++ b/tests/cucumber_tests/features/manual_control.feature
@@ -7,10 +7,6 @@ Feature: Manual control of transactions
   # Supported commands (one per line):
   #   COIN_SPLIT, wallet '<wallet_name>', outputs <count>, value <amount>
   #   VERIFY, wallet '<wallet_name>', outputs <count>, time_out <duration_seconds>
-  #   BALANCE, wallet '<wallet_name>'
-  #   BALANCE_ALL_WALLETS
-  #   BALANCE_ALL_USER_WALLETS
-  #   BALANCE_ALL_FUNDING_WALLETS
   #   CLEAR_ENCUMBRANCES, wallet '<wallet_name>'
   #   CLEAR_ENCUMBRANCES_ALL_WALLETS
   #   SEND, transactions <count>, value <amount>, from '<wallet_name>', to '<wallet_name>'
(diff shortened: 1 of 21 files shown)


tests/cucumber_tests/features/transactions.feature
```diff
@@ -284,20 +284,20 @@ Feature: Transactions
...
+    And wallet "WALLET_4A" has all submitted transactions settled in 240 seconds
+    And wallet "WALLET_5A" has all submitted transactions settled in 240 seconds
     # Each wallet should now have 3 outputs: one received + two change (allow for fees)
     When wallet "WALLET_1A" has 3 or more outputs in 120 seconds
     And wallet "WALLET_2A" has 3 or more outputs in 60 seconds
     And wallet "WALLET_4A" has 3 or more outputs in 60 seconds
     And wallet "WALLET_5A" has 3 or more outputs in 60 seconds
-    # Wait for more blocks to be mined on each fork to ensure the chains are well established
-    When node "NODE_1" is at height 5 in 180 seconds
-    And node "NODE_4" is at height 5 in 180 seconds
```
hansieodendaal, line 294: These lines
```
    # Wait for more blocks to be mined on each fork to ensure the chains are well established
    When node "NODE_1" is at height 5 in 180 seconds
    And node "NODE_4" is at height 5 in 180 seconds
```
are an integral part of the test and should not be removed. The objective is to try and merge two well established chains.

tests/cucumber_tests/features/transactions.feature
```diff
@@ -284,20 +284,20 @@ Feature: Transactions
...
     And I send 1 transactions of 700 LGO each from wallet "WALLET_5A" to wallet "WALLET_4A"
+    When wallet "WALLET_1A" has all submitted transactions settled in 240 seconds
+    And wallet "WALLET_2A" has all submitted transactions settled in 240 seconds
+    And wallet "WALLET_4A" has all submitted transactions settled in 240 seconds
+    And wallet "WALLET_5A" has all submitted transactions settled in 240 seconds
     # Each wallet should now have 3 outputs: one received + two change (allow for fees)
     When wallet "WALLET_1A" has 3 or more outputs in 120 seconds
     And wallet "WALLET_2A" has 3 or more outputs in 60 seconds
     And wallet "WALLET_4A" has 3 or more outputs in 60 seconds
     And wallet "WALLET_5A" has 3 or more outputs in 60 seconds
```
hansieodendaal, line 295: Are there any reason why we need the above `has all submitted transactions settled in 240 seconds` and not just extend the timeout on these lines? Are those required to drive wallet scanning on forks?

tests/cucumber_tests/features/manual_control.feature
```diff
@@ -7,10 +7,6 @@ Feature: Manual control of transactions
   # Supported commands (one per line):
   #   COIN_SPLIT, wallet '<wallet_name>', outputs <count>, value <amount>
   #   VERIFY, wallet '<wallet_name>', outputs <count>, time_out <duration_seconds>
-  #   BALANCE, wallet '<wallet_name>'
-  #   BALANCE_ALL_WALLETS
-  #   BALANCE_ALL_USER_WALLETS
-  #   BALANCE_ALL_FUNDING_WALLETS
```
hansieodendaal, line 13: The objective with these manual commands were to drive wallet scanning and to display wallet balances, so although they are not driving wallet scanning anymore, I think we should keep it for the display benefit.

tests/src/cucumber/steps/manual_transactions/command_file_parsing.rs
```diff
@@ -28,12 +28,6 @@ pub enum ManualCommand {
         wallet_state_type: WalletOutputState,
         verify_max: bool,
     },
-    WalletBalance {
-        wallet_name: String,
-    },
-    WalletBalanceAllUserWallets,
-    WalletBalanceAllFundingWallets,
```
hansieodendaal, line 35: As per previous comment, please keep these for the console log/display purposes

tests/src/cucumber/wallet/sync.rs
```diff
@@ -219,506 +117,338 @@ pub async fn sync_wallet_available_state(
...
+        wallet.wallet_name.as_str(),
+        &wallet_keys,
+        feed_requirements,
+    )
+    .await
 }
 
-async fn build_wallet_feed_source_plan(
-    world: &CucumberWorld,
+pub async fn current_wallet_state_for_key(
```
hansieodendaal, line 179: I think this fn is unused. It also passes empty feed requirements?

hansieodendaal (commented):
Looks good overall, nice reduction in complexity and coding!

**Backfill**
I could not trace what happens to backfill batches beyond the log message in `pub fn track_known_wallets_with_block_feed(&self) -> StepResult {` - how is this handled?
If a wallet is introduced/tracked after the feed has already advanced, `track_wallets` can only say “needs backfill”; it does not add that new wallet to the existing source tracker?
Maybe we should preserve the old behavior by performing direct backfill when `NeedsBackfill` / `requires_direct_backfill()` is encountered?

**Other comments**
Then also a couple of other minor comments below to consider please.

andrussal:
> Looks good overall, nice reduction in complexity and coding!
> 
> **Backfill** I could not trace what happens to backfill batches beyond the log message in `pub fn track_known_wallets_with_block_feed(&self) -> StepResult {` - how is this handled? If a wallet is introduced/tracked after the feed has already advanced, `track_wallets` can only say “needs backfill”; it does not add that new wallet to the existing source tracker? Maybe we should preserve the old behavior by performing direct backfill when `NeedsBackfill` / `requires_direct_backfill()` is encountered?
> 
> **Other comments** Then also a couple of other minor comments below to consider please.

Good catch. I changed this so `track_known_wallets_with_block_feed()` no longer only logs `NeedsBackfill`. It now reuses the direct backfill path immediately, so when known wallets are registered after a feed source has already advanced, the source tracker is rebuilt and transaction hashes observed during that backfill are recorded too.

tests/cucumber_tests/features/transactions.feature
andrussal: > Are there any reason why we need the above `has all submitted transactions settled in 240 seconds` and not just extend the timeout on these lines? Are those required to drive wallet scanning on forks?

I kept the settled-transaction checks. They are not there to drive wallet scanning; they assert that the submitted tx hashes were actually observed before the scenario moves on. A longer height timeout only proves blocks were produced, not that the expected txs landed.

tests/cucumber_tests/features/manual_control.feature
andrussal: > The objective with these manual commands were to drive wallet scanning and to display wallet balances, so although they are not driving wallet scanning anymore, I think we should keep it for the display benefit.

Agreed, restored these for display/console use. They no longer drive wallet scanning, but keeping them as manual inspection commands makes sense.

tests/src/cucumber/wallet/sync.rs
andrussal: > I think this fn is unused. It also passes empty feed requirements?

This is still used by the scenario-fee wallet check path. I also changed it so it no longer passes empty feed requirements; it now tracks the key against active feed sources and waits for those source heights before reading state.

hansieodendaal (approved):
LGTM

PR #2921, opened 2026-06-10
chore: add bounded size and count transaction vec for blocks

hansieodendaal:

This PR introduces bounded block transactions as a first-class invariant across block creation, serialization/deserialization (_including serde deserialization_), and network ingestion.

At the end state, block transactions are enforced by a bounded container that validates both:
- max transaction count, and
- max total serialized size

This enforcement is applied consistently through core and service paths, so oversized/invalid transaction sets are rejected at boundaries rather than being accepted and failing later.

Why this matters
- Prevents blocks from exceeding protocol transaction/count/size constraints.
- Makes block validity checks deterministic and centralized.
- Aligns mempool/block assembly and network block intake with the same bounded-transaction rules.
- Preserves real transaction-size semantics (serialized byte size), so limits are meaningful, not placeholder checks.

Network block intake warrants further discussion: previously, the libp2p adapter at `async fn request_blocks_from_peer` accepted a raw wire block if it passed deserialized without any validation - the same checks are now implemented as when reconstructing a block from a proposal. 
In practice, this turns the libp2p adapter into a hard validation gate:
- peers cannot inject blocks with transaction vectors that violate count/size bounds;
- the same canonical block constraints used in core logic are enforced at network ingress;
- downstream services can rely on already-bounded block transaction data.

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


core/src/mantle/ops/channel/inscribe.rs
```diff
@@ -199,7 +199,7 @@ mod tests {
         let oversized = vec![0u8; MAX_BYTES + 1];
         let err = Inscription::try_from(oversized).unwrap_err();
         assert!(
-            matches!(err, BoundedError::TooLong { actual, max } if actual == MAX_BYTES + 1 && max == MAX_BYTES)
+            matches!(err, BoundedError::TooManyItems { count, max } if count == MAX_BYTES + 1 && max == MAX_BYTES)
```
youngjoon-lee, line 170: I think the previous `TooLong` is more appropriate than `TooManyItems` because the `oversized` is too long. Wdyt?

core/src/utils/storage_bounded_vec.rs
```diff
@@ -0,0 +1,766 @@
...
+    )
+)]
+pub struct StorageBoundedVec<
+    T,
+    const MIN_COUNT: usize,
+    const MAX_COUNT: usize,
+    const MAX_SIZE: usize,
+> {
+    items: BoundedVec<T, MIN_COUNT, MAX_COUNT>,
+    total_cached_size: usize,
```
youngjoon-lee: What does `cached` mean here? I think `storage_size` is enough?

core/src/utils/storage_bounded_vec.rs
```diff
@@ -0,0 +1,766 @@
...
+}
+
+impl<T, const MIN_COUNT: usize, const MAX_COUNT: usize, const MAX_SIZE: usize>
+    StorageBoundedVec<T, MIN_COUNT, MAX_COUNT, MAX_SIZE>
+{
+    pub const MIN_COUNT: usize = MIN_COUNT;
+    pub const MAX_COUNT: usize = MAX_COUNT;
+    pub const MAX_SIZE: usize = MAX_SIZE;
```
youngjoon-lee: I think we can remove these constants and the test `constants_reflect_generic_parameters` below.

youngjoon-lee: I see that these constants are used by only the tests in `services/chain/chain-leader/src/lib.rs`. IMO, duplicating these constants is redundant, but I'll follow your decision if you think this is necessary.

core/src/block/mod.rs
```diff
@@ -121,50 +115,65 @@ impl<Tx> Block<Tx> {
...
 
-        Ok(Self {
+        Ok(())
+    }
+
+    /// Convert a raw/deserialized block into a validated non-genesis block.
+    ///
+    /// This enforces transaction count, storage-size, block-root, and signature
+    /// validation before returning the block.
+    pub fn try_into_bounded_block(self) -> Result<Self, Error>
```
youngjoon-lee: It seems that this function should be always used after `Blocks::from_bytes`. If so, what do you think about changing this function to accept `&[u8]` and call `Self::from_bytes` internally? In this way, we can prevent callers from accidentally using it incorrectly.

Also, I think we can (should) remove `impl TryFrom<Bytes> for Block<Tx> {` below.

core/src/utils/storage_bounded_vec.rs
```diff
@@ -0,0 +1,766 @@
...
+
+    pub fn try_pop(&mut self) -> Result<Option<T>, StorageBoundedError>
+    where
+        T: StorageSize,
+    {
+        let Some(item) = self.items.try_pop()? else {
+            return Ok(None);
+        };
+
+        self.total_cached_size -= item.storage_size();
```
youngjoon-lee: I think we can use `checked_sub` and `unwrap`, similar as `checked_add` above.

Also, in the `try_remove` below.

youngjoon-lee (commented):
Great 👍  I left some suggestions.

core/src/mantle/ops/channel/inscribe.rs
youngjoon-lee: Ah, I just saw that you renamed the error in the `bounded_vec.rs`. I understand what you intended.

core/src/utils/storage_bounded_vec.rs
hansieodendaal: I added this member function instead:
```rust
    pub const fn bounds() -> StorageBoundedVecBounds {
        StorageBoundedVecBounds {
            min_count: MIN_COUNT,
            max_count: MAX_COUNT,
            max_size: MAX_SIZE,
        }
    }
```
Those constants are the type declaration bounds.

core/src/block/mod.rs
hansieodendaal: Good observation, I also thought about it, obviously not long and hard enough -:).

hansieodendaal: Ok, I opted for the reverse... 
- Added the `try_into_bounded_block` call into the `impl TryFrom<Bytes> for Block<Tx>` as final step.
- Made a custom serde deserialize for the `Block<Tx>` that also calls `try_into_bounded_block`
All deserialize paths are now covered, and `Blocks::from_bytes(...)` are not exposed anymore - callers either use the custom serde or `TryFrom` to deserialize blocks.

hansieodendaal:
@bacv @ntn-x2 @youngjoon-lee the stricter serde deserialization resulted in these unit test failures, 5x due to `Devnet deployment config is valid.: SerdeError(Error("cryptarchia: Signature error.", line: 21, column: 3))` and 1x due to `ERROR [get_deployment_config] Could not parse deployment file: Signature error.`
```
     Summary [ 285.223s] 902 tests run: 895 passed, 6 failed, 1 timed out, 15 skipped
  TRY 3 FAIL [   0.035s] logos-blockchain-c api::lifecycle::test::test_basic_lifecycle
   TRY 3 TMT [  90.008s] logos-blockchain-cfgsync::server_client smoke_test_four_clients
  TRY 3 FAIL [   0.022s] logos-blockchain-node config::blend::deployment::tests::blend_devnet
  TRY 3 FAIL [   0.020s] logos-blockchain-node config::deployment::tests::devnet_initialization
  TRY 3 FAIL [   0.024s] logos-blockchain-node config::deployment::tests::serialize_deserialize_yaml
  TRY 3 FAIL [   0.050s] logos-blockchain-node config::tests::common_recovery_folder
  TRY 3 FAIL [   0.019s] logos-blockchain-node config::tests::standalone_deployment_config_deserializes
  ```
  
  Any ideas? How is that signature generated, as the now implicit block validation throws it out?

File: core/src/block/mod.rs [hunk 6]
```diff
@@ -51,13 +50,41 @@ pub struct References {
...
+    {
+        let raw = RawBlock::<Tx>::deserialize(deserializer)?;
+        let block = Self {
+            header: raw.header,
+            signature: raw.signature,
+            transactions: raw.transactions,
+        };
+
+        block
+            .try_into_bounded_block()
```
bacv, line 84: Because of this custom Deserialize impl, the genesis block is now being validated when deserializing, where the leader key is set to all zeroes. We shouldn't use this deserialize method for genesis block

danielSanchezQ:
image

I think this is adding more complexity than what we already had. Not blocking it, but delaying its inclusion till we have a stable 0.2.0 version deployed and working.

danielSanchezQ (changes requested):
Delaying inclusion till next iterations after 0.2.0 release.

bacv:
> Any ideas? How is that signature generated, as the now implicit block validation throws it out?

I wasn't able to fetch your branch, so didn't check precisely where the issue is, but I suspect custom deserialization.

To answer the question how the standalone genesis block is generated: by using a script in `scripts/standalone-genesis-ceremony.sh`. We don't generate it on every rebuild, it's supposed to be ran only when the format changes.

hansieodendaal:
Ok thanks, so this can just bypass the genesis block.

ntn-x2:
In general I am in favor of this, but before making the block as such bounded, we should "boundify" all the lower-level components, and everything that goes into a block. Also, we need a strategy on how to deal with serde and nom encoding/decoding.

davidrusu:
Is the block the only consumer of this StorageBoundedVec? this feels over-engineered/over-generalized for just one use case.

Is the Block API unsafe? how about instead we focus on a BlockBuilder that tracks the tx-count, size-so-far, current-ledger-state and give a simple block building interface rather than this bounded vec?

hansieodendaal:
> Is the block the only consumer of this StorageBoundedVec? this feels over-engineered/over-generalized for just one use case.
> 
> Is the Block API unsafe? how about instead we focus on a BlockBuilder that tracks the tx-count, size-so-far, current-ledger-state and give a simple block building interface rather than this bounded vec?

Ok, it feels heavy, I agree, let me park this PR and only make a simple PR using only `BoundedVec` with size aware helpers where needed, also retaining the improvements regarding block deserialization on the wire and serde boundaries.

`StorageBoundedVec` as a type can be parked and pulled out again when we identify more consumers.

PR #2923, opened 2026-06-11
feat(wallet): add pending voucher tracking to prevent race conditions

andrussal:

This PR moves leader-claim transaction construction into the wallet service because the wallet is the component with the local context needed to build these transactions correctly.

Previously, chain-leader selected a voucher and then asked the wallet to fund/sign the transaction, which split claim construction across services. That becomes brittle once a node can have multiple claimable vouchers, because voucher selection needs to account for claim transactions that are already being built.

With this change, chain-leader only asks the wallet for a ready claim transaction. The wallet handles voucher selection, funding, signing, and short-lived pending-claim tracking in one place, so concurrent claim requests do not reuse the same voucher.

Funding-note reservation is intentionally not included in this PR. For the current testnet leader-claim flow we expect 0 gas fees, so claim transactions should not need funding notes in the normal path. The broader funding-note concurrency issue still exists for other wallet-funded operations and should be handled separately.

The focused e2e test currently fails because concurrent claim txs can still race on funding notes. There is already a fix for the zero-fee path in https://github.com/logos-blockchain/logos-blockchain/pull/2970, so once that lands this should be pass.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/services/wallet/src/lib.rs
+++ b/services/wallet/src/lib.rs
@@ -200,7 +200,7 @@ pub struct UtxoWithKeyId {
     pub key_id: KeyId,
 }
 
-struct BuiltLeaderClaimTx {
+struct LeaderClaimTx {
     signed_tx: SignedMantleTx,
     voucher_nullifier: VoucherNullifier,
 }
@@ -1087,8 +1087,8 @@ where
         ledger: LedgerState,
         state: &mut ServiceState<'_>,
         kms: &KmsServiceApi<Kms, RuntimeServiceId>,
-    ) -> Result<BuiltLeaderClaimTx, WalletServiceError> {
-        let voucher_nullifier = Self::reserve_claimable_voucher(state, request.tip)
+    ) -> Result<LeaderClaimTx, WalletServiceError> {
+        let voucher_nullifier = Self::reserve_claimable_voucher(state, request.tip)?
             .ok_or(WalletServiceError::NoClaimableVoucher)?;
 
         let result =
@@ -1099,7 +1099,7 @@ where
             state.release_claim_reservation(voucher_nullifier);
         }
 
-        result.map(|signed_tx| BuiltLeaderClaimTx {
+        result.map(|signed_tx| LeaderClaimTx {
             signed_tx,
             voucher_nullifier,
         })
@@ -1108,8 +1108,13 @@ where
     fn reserve_claimable_voucher(
         state: &mut ServiceState<'_>,
         tip: HeaderId,
-    ) -> Option<VoucherNullifier> {
-        match state.find_available_claimable_voucher(tip) {
+    ) -> Result<Option<VoucherNullifier>, WalletServiceError> {
+        let claimable_vouchers = state.claimable_vouchers(tip)?;
+
+        match (
+            claimable_vouchers.available.into_iter().next(),
+            claimable_vouchers.pending.len(),
+        ) {
             (Some(voucher), _) => {
                 state.reserve_claim(voucher.nullifier);
 
@@ -1120,16 +1125,16 @@ where
                     "Found and reserved claimable voucher"
                 );
 
-                Some(voucher.nullifier)
+                Ok(Some(voucher.nullifier))
             }
             (None, pending_count) if pending_count > 0 => {
                 debug!(
                     target: LOG_TARGET,
                     "No available vouchers: {pending_count} are pending"
                 );
-                None
+                Ok(None)
             }
-            (None, _) => None,
+            (None, _) => Ok(None),
         }
     }
 
@@ -1146,38 +1151,21 @@ where
                 return;
             }
         };
-        let response = Self::find_claimable_vouchers(state, tip).map(|vouchers| TipResponse {
+        let response = state.claimable_vouchers(tip).map(|vouchers| TipResponse {
             tip,
-            response: vouchers,
+            response: vouchers
+                .available
+                .into_iter()
+                .map(|voucher| ClaimableVoucherInfo {
+                    commitment: voucher.commitment,
+                    nullifier: voucher.nullifier,
+                })
+                .collect(),
         });
 
         if resp_tx.send(response).is_err() {
             debug!(target: LOG_TARGET, "Failed to respond to GetClaimableVouchers");
         }
-    }
-
-    fn find_claimable_vouchers(
-        state: &ServiceState<'_>,
-        tip: HeaderId,
-    ) -> Result<Vec<ClaimableVoucherInfo>, WalletServiceError> {
-        let wallet = state.wallet();
-
-        wallet
-            .voucher_commitments_and_nullifiers()
-            .filter(|voucher| !state.is_claim_reserved(&voucher.nullifier))
-            .filter_map(|voucher| {
-                wallet
-                    .voucher_path_snapshot(tip, &voucher.commitment)
-                    .map_err(WalletServiceError::from)
-                    .transpose()
-                    .map(|path| {
-                        path.map(|_| ClaimableVoucherInfo {
-                            commitment: voucher.commitment,
-                            nullifier: voucher.nullifier,
-                        })
-                    })
-            })
-            .collect()
     }
 
     async fn build_reserved_leader_claim_tx(
@@ -1341,11 +1329,11 @@ where
             state.release_claim_reservation(*nullifier);
         }
 
-        let lib_blocks_count = lib_update.pruned_blocks.immutable_blocks.len() as u64;
+        let new_immutable_blocks_count = lib_update.pruned_blocks.immutable_blocks.len() as u64;
         state.advance_lib(
             lib_update.new_lib,
             lib_update.pruned_blocks.all(),
-            lib_blocks_count,
+            new_immutable_blocks_count,
             claimed_nullifiers,
         );
     }
--- a/services/wallet/src/lib.rs
+++ b/services/wallet/src/lib.rs
@@ -16,8 +16,8 @@ use lb_core::{
     events::Events,
     header::HeaderId,
     mantle::{
-        AuthenticatedMantleTx, Note, NoteId, Op, OpProof, SignedMantleTx, Transaction as _, TxHash,
-        Utxo, Value,
+        AuthenticatedMantleTx, NoteId, Op, OpProof, SignedMantleTx, Transaction as _, TxHash, Utxo,
+        Value,
         gas::{GasCost, GasOverflow, MainnetGasConstants},
         ledger::Inputs,
         ops::{
@@ -1175,13 +1175,12 @@ where
         state: &ServiceState<'_>,
         kms: &KmsServiceApi<Kms, RuntimeServiceId>,
     ) -> Result<SignedMantleTx, WalletServiceError> {
-        let tx_builder = MantleTxBuilder::new(ledger.tx_context())
-            .push_op(Op::LeaderClaim(LeaderClaimOp {
+        let tx_builder =
+            MantleTxBuilder::new(ledger.tx_context()).push_op(Op::LeaderClaim(LeaderClaimOp {
                 rewards_root: request.rewards_root,
                 voucher_nullifier,
                 pk: request.funding_pk,
-            }))?
-            .add_ledger_output(Note::new(request.reward_amount, request.funding_pk))?;
+            }))?;
 
         let funded_tx_builder = state.wallet().fund_tx::<MainnetGasConstants>(
             request.tip,
--- a/services/wallet/src/lib.rs
+++ b/services/wallet/src/lib.rs
@@ -1,7 +1,7 @@
 pub mod api;
 mod states;
 
-use std::{collections::HashMap, path::PathBuf, time::Duration};
+use std::{collections::HashMap, num::NonZeroU64, path::PathBuf, time::Duration};
 
 use async_trait::async_trait;
 use bytes::Bytes;
@@ -382,6 +382,7 @@ where
         let mut lib_receiver = cryptarchia_api.subscribe_lib_updates().await?;
 
         let (epoch_config, consensus_config) = cryptarchia_api.get_epoch_config().await?;
+        let security_param = NonZeroU64::from(consensus_config.security_param()).get();
         let epoch_config = EpochConfig {
             epoch_config,
             consensus_config,
@@ -402,6 +403,7 @@ where
             lib,
             &lib_ledger,
             &service_resources_handle.state_updater,
+            security_param,
         );
         let voucher_master_key_id = settings.voucher_master_key_id;
 
@@ -1358,9 +1360,11 @@ where
             state.release_claim_reservation(*nullifier);
         }
 
+        let lib_blocks_count = lib_update.pruned_blocks.immutable_blocks.len() as u64;
         state.advance_lib(
             lib_update.new_lib,
             lib_update.pruned_blocks.all(),
+            lib_blocks_count,
             claimed_nullifiers,
         );
     }
--- a/services/wallet/src/lib.rs
+++ b/services/wallet/src/lib.rs
@@ -640,7 +640,7 @@ where
                 .await;
             }
             WalletMsg::GetClaimableVouchers { tip, resp_tx } => {
-                Self::get_claimable_vouchers(tip, resp_tx, state.wallet(), cryptarchia).await;
+                Self::get_claimable_vouchers(tip, resp_tx, state, cryptarchia).await;
             }
             WalletMsg::GetKnownAddresses { resp_tx } => {
                 Self::get_known_addresses(state.wallet(), resp_tx);
@@ -1136,7 +1136,7 @@ where
     async fn get_claimable_vouchers(
         tip: Option<HeaderId>,
         resp_tx: Sender<Result<TipResponse<Vec<ClaimableVoucherInfo>>, WalletServiceError>>,
-        wallet: &Wallet,
+        state: &ServiceState<'_>,
         cryptarchia: &CryptarchiaServiceApi<Cryptarchia, RuntimeServiceId>,
     ) {
         let tip = match Self::msg_tip_or_latest(tip, cryptarchia).await {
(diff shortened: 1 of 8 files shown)


services/wallet/src/lib.rs
```diff
@@ -996,49 +1069,144 @@ where
...
-                return Some(VoucherCommitmentAndNullifier {
-                    commitment: *cm,
-                    nullifier: *nf,
-                });
-            }
+    ) -> Option<VoucherNullifier> {
+        let vouchers: Vec<_> = state
+            .wallet()
+            .voucher_commitments_and_nullifiers()
+            .map(|(nf, cm)| (*nf, *cm))
```
danielSanchezQ: `copied`?

services/wallet/src/lib.rs
```diff
@@ -996,49 +1069,144 @@ where
...
+
+            debug!(
+                target: LOG_TARGET,
+                ?nf,
+                ?cm,
+                "Found and reserved claimable voucher"
+            );
+
+            return Some(nf);
+        }
```
danielSanchezQ: this can be collapsed
```suggestion
       match Self::find_available_voucher(&vouchers, state, tip) {
       (Some((nf, cm)), pending_count) => {
           state.reserve_claim(nf);

            debug!(
                target: LOG_TARGET,
                ?nf,
                ?cm,
                "Found and reserved claimable voucher"
            );

            Some(nf)
       }
       (None, pending_count) if pending_count > 0 {
           debug!(
                target: LOG_TARGET,
                "No available vouchers: {}/{} are pending",
                pending_count,
                vouchers.len()
            );
            None
       }
       _ => None
```

services/wallet/src/lib.rs
```diff
@@ -996,49 +1069,144 @@ where
...
+                );
+                continue;
+            }
+
+            if let Ok(Some(_)) = state.wallet().voucher_path_snapshot(tip, &cm) {
+                return (Some((nf, cm)), pending_count);
+            }
+        }
+
+        (None, pending_count)
```
danielSanchezQ:
```suggestion
        vouchers.iter().zip(0..).find(|((nf, cm), _)| !state.is_claim_pending(nf) && matches!(state.wallet().voucher_path_snapshot(tip, cm), Ok(Some(_)))).unwrap_or((None, vouchers.len()))
```
Sorry for the formatting, but you get the idea.

danielSanchezQ: You can add the log if you have a method for the find function.

services/wallet/src/states.rs
```diff
@@ -1,25 +1,141 @@
...
+
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct RecoveryState {
     next_new_voucher_index: VoucherIndex,
     vouchers: Vouchers<VoucherId>,
     /// [`WalletState`] at the last known LIB.
     /// `None` on fresh start; populated after the first LIB update.
     lib_wallet_state: Option<(HeaderId, WalletState)>,
+    #[serde(skip)]
+    pending_claims: PendingClaims,
```
danielSanchezQ, line 121: Why not? 🤔

services/wallet/src/lib.rs
andrussal: > `copied`?

Changed this a bit differently: instead of copying tuple fields, I added a small KnownVoucher type and made the wallet iterator return that.

services/wallet/src/lib.rs
andrussal: Updated, the selection result is now handled with a match so the success/pending/no-voucher cases are grouped in one place.

services/wallet/src/lib.rs
andrussal: I moved this into ServiceState instead. I also dropped the per-voucher trace log since it did not add much.

services/wallet/src/states.rs
andrussal: Added a comment. We should not recover pending claims because after restart we cannot know whether the reserved claim was submitted, accepted, or dropped, and keeping it could block usable vouchers or notes.

services/wallet/src/lib.rs
```diff
@@ -187,10 +200,17 @@ pub struct UtxoWithKeyId {
     pub key_id: KeyId,
 }
 
-#[derive(Debug)]
-pub struct VoucherCommitmentAndNullifier {
-    pub commitment: VoucherCm,
-    pub nullifier: VoucherNullifier,
+struct BuiltLeaderClaimTx {
```
youngjoon-lee:
```suggestion
struct LeaderClaimTx {
```
I think it's better to simplify this name.

services/wallet/src/states.rs
```diff
@@ -1,25 +1,107 @@
...
 
 use crate::{KeyId, WalletServiceError, WalletServiceSettings};
 
 type VoucherIndex = u64;
 type VoucherId = (KeyId, VoucherIndex);
 pub type Wallet = lb_wallet::Wallet<KeyId, VoucherId>;
 
+#[derive(Debug, Clone)]
+struct PendingClaim {
+    lib_blocks_elapsed: u64,
```
youngjoon-lee: What does this field mean? It's hard to infer the purpose of this field without any comment, or any better name.

services/wallet/src/lib.rs
```diff
@@ -1206,21 +1331,29 @@ where
...
+            lib_update.pruned_blocks.immutable_blocks.values(),
+            storage_adapter,
+        )
+        .await;
+
+        for nullifier in &claimed_nullifiers {
+            state.release_claim_reservation(*nullifier);
+        }
+
+        let lib_blocks_count = lib_update.pruned_blocks.immutable_blocks.len() as u64;
```
youngjoon-lee: I don't understand what `lib_blocks_count` means. The similar name `lib_blocks` is also used in the `states.rs`, but it's not easy to understand what it's for, without reading all modules. There's always only one LIB at a time.

services/wallet/src/states.rs
```diff
@@ -1,25 +1,107 @@
...
+                "Released pending claim reservation"
+            );
+        }
+    }
+
+    fn is_reserved(&self, nullifier: &VoucherNullifier) -> bool {
+        self.claims.contains_key(nullifier)
+    }
+
+    fn advance_lib(&mut self, lib_blocks: u64, expiry_blocks: u64) {
```
youngjoon-lee: I don't understand what this function is for. What do `lib_blocks` and `expiry_blocks` mean? Is this function for evicting pending vouchers (TTL-based)? If so, what do you think about renaming this function to `evict`, similar as other functions below?
https://github.com/logos-blockchain/logos-blockchain/blob/fab72c58826a12f581a43e28c97cb7fb214613b8/services/wallet/src/states.rs#L204-L207

Also, two arguments need to be properly renamed. And, please add a doc comment.

services/wallet/src/states.rs
youngjoon-lee: I don't understand clearly why we cannot know those. Is it because there might be a case which a voucher is reserved, but the node was terminated before the tx is submitted? Yeah, that can happen.

But then, what do you think about just holding `PendingClaims` in memory in the service layer? We don't need to put this in the `RecoveryState` if we are not going to load it later.

services/wallet/src/lib.rs
```diff
@@ -1077,26 +1157,71 @@ where
     }
 
     fn find_claimable_vouchers(
-        wallet: &Wallet,
+        state: &ServiceState<'_>,
         tip: HeaderId,
     ) -> Result<Vec<ClaimableVoucherInfo>, WalletServiceError> {
```
youngjoon-lee: It seems that you're also adding a similar function `find_available_claimable_voucher` in the `ServiceState`. Can we probably have a single function that does everything that both of those functions want?

tests/src/tests/mantle/leader.rs
```diff
@@ -35,74 +37,18 @@ const NODE_COUNT: usize = 1;
...
     let mut block_stream = node.client.blocks_stream().await.unwrap();
 
     // Submit a tx with a LeaderClaim operation
     wait_for_claimable_vouchers(&node.client, Duration::from_secs(30)).await;
 
     let tx_hash = claim_leader_rewards(&node.client, Duration::from_secs(30)).await;
-
-    // Wait for the claim tx to be included in the chain
-    // TODO: Check if wallet balance is increased by improving wallet
-    // to track reward UTXOs in the wallet: https://github.com/logos-blockchain/logos-blockchain/issues/2627
```
youngjoon-lee, line 86: Why did you remove this TODO?

tests/src/tests/mantle/leader.rs
```diff
@@ -35,74 +37,18 @@ const NODE_COUNT: usize = 1;
...
-        target_slot,
-        Duration::from_mins(5),
-    )
-    .await;
-
-    // Subscribes to process blocks to check if our tx is included in the end
     let mut block_stream = node.client.blocks_stream().await.unwrap();
 
     // Submit a tx with a LeaderClaim operation
     wait_for_claimable_vouchers(&node.client, Duration::from_secs(30)).await;
```
youngjoon-lee: Is 30s enough for two epoch transitions? Also, please don't remove this comment:
```
// Wait for two epoch transitions.
// 0->1: vouchers (blocks) are collected but not added to MMR
// 1->2: vouchers are added to MMR and become claimable
```

wallet/src/voucher.rs
```diff
@@ -3,6 +3,12 @@ use std::collections::HashMap;
 use lb_core::mantle::ops::leader_claim::{VoucherCm, VoucherNullifier};
 use serde::{Deserialize, Serialize};
 
+#[derive(Clone, Copy, Debug)]
+pub struct KnownVoucher {
+    pub nullifier: VoucherNullifier,
+    pub commitment: VoucherCm,
+}
```
youngjoon-lee:
```suggestion
pub struct Voucher {
    pub nullifier: VoucherNullifier,
    pub commitment: VoucherCm,
}
```
We can use a more general name.

services/wallet/src/states.rs
```diff
@@ -1,25 +1,107 @@
...
 pub struct RecoveryState {
     next_new_voucher_index: VoucherIndex,
     vouchers: Vouchers<VoucherId>,
     /// [`WalletState`] at the last known LIB.
     /// `None` on fresh start; populated after the first LIB update.
     lib_wallet_state: Option<(HeaderId, WalletState)>,
+    // After restart we cannot know whether a reserved claim was submitted,
+    // accepted, or dropped, so recovering it could block usable vouchers/notes.
+    #[serde(skip)]
+    pending_claims: PendingClaims,
```
youngjoon-lee, line 121: It seems that this PR doesn't have the reservation mechanism for funding notes, which doesn't match the PR description saying:
> concurrent claim requests do not reuse the same voucher or funding notes

If so, doesn't it mean that this PR fixes the issue "partially"? It still seems possible that the race condition(?) could happen. I think that it's a fundamental issue that the wallet service has, not only for leader_claim tx, but also any kind of operations that use funding notes. Do you think we need to address this issue right now? or do you think that merging this approach would make our life easier now?

Also, I'd like to hear why you dropped the funding note reservation that you implemented in the 1st commit in this PR.

cc @danielSanchezQ

youngjoon-lee (commented):
Thank you. This was actually a leftover. Previously the chain leader service has some context to build leader_claim txs, but we refactored the architecture and now the wallet service has the full context. So, this flow change is sound.
I think we need to improve some parts before merging this.

services/wallet/src/states.rs
youngjoon-lee: Or, what do you think about putting all responsibilities to the user? We already provide an API to get a list of claimable vouchers. We can add an API for a list of funding notes (or just notes owned by a certain pk).
Then, the user can pick one of the vouchers and one of the funding notes, in order to ask the wallet service to build and submit a tx. 
In this model, we don’t need to have the reservation/pending management with TTL that doesn’t look so convincing to me. 

Of course, the user can make mistakes, and should memorize which vouchers and notes he used. But, IMO, that’s fine because the current wallet service doesn’t completely solve all concurrency issues caused by the late finality.

danielSanchezQ: ideally we should make this easy. Its ok for now going for a completely manual process for advance users. But at some we need something simplified. It should be a button to claim vouchers and thats it. That would build the tx and submit everything.

services/wallet/src/lib.rs
danielSanchezQ: I agree with that, a single funtion that returns 2 sets of claimable and non-claimable ones.

tests/src/tests/mantle/leader.rs
danielSanchezQ: Ideally we do not do anything time based btw, cant we reuse the timeservice and use slots/epochs?

services/wallet/src/states.rs
andrussal: Good point, I should update the PR description after the latest commits. The current PR only reserves claim vouchers, not funding notes.

I initially had funding-note reservation as well, but removed it after David’s feedback because the faucet currently relies on note reuse as a form of rate limiting, and for this testnet leader-claim should not need funding notes anyway because we expect 0 gas fees.

I agree note reuse is still a general wallet issue for other funded operations. I’d keep that separate from this PR though. This PR is mainly about making leader-claim safer by not letting concurrent claim requests pick the same voucher.

andrussal: Regarding user responsibility, since this is for the dashboard journey, I think it makes sense to keep the main flow simple: user clicks claim, and the node/wallet builds and submits the tx.

services/wallet/src/states.rs
andrussal: I originally skipped serializing it because I was worried about stale reservations after restart. But thinking about it again, persisting voucher reservations seems safer here, because after restart we still want to avoid picking the same voucher while an earlier claim may be in flight.

>  Is it because there might be a case which a voucher is reserved, but the node was terminated before the tx is submitted? Yeah, that can happen.

I think it is ok to accept that case as a compromise because the stale reservation is temporary. If the node stops after reserving but before submitting, the voucher may be hidden after restart, but only for a short period of time.

services/wallet/src/states.rs
andrussal: Renamed this to `immutable_blocks_since_reservation`.

services/wallet/src/lib.rs
andrussal: Renamed this to `new_immutable_blocks_count`, meaning the number of blocks that became immutable in this LIB update.

services/wallet/src/states.rs
andrussal: Renamed this to `evict_expired`. The expiry is based on newly immutable blocks, not wall-clock time.

tests/src/tests/mantle/leader.rs
andrussal: Restored the comment.

services/wallet/src/lib.rs
andrussal: Removed `find_available_claimable_voucher`. The reservation path now calls `state.claimable_vouchers(tip)?` directly.

tests/src/tests/mantle/leader.rs
andrussal: You’re right, 30s can be too short here. I restored the comment and changed the timeout back to 5 minutes. 

> Ideally we do not do anything time based btw, cant we reuse the timeservice and use slots/epochs?

I agree that waiting on explicit state is better for e2e tests. In this case I think polling the claimable-vouchers API is the right state to wait for, because the next step only needs a voucher to be claimable. The duration is just the outer timeout so the test can fail if that state never appears.

wallet/src/voucher.rs
andrussal: Renamed `KnownVoucher` to `Voucher`.

tests/src/tests/mantle/leader.rs
youngjoon-lee: Thanks. I agree. Previously, I implemented this test to wait first until two epochs pass. But this PR removed it. But I think it’s okay because waiting for a claimable voucher is a superset of waiting for two epoch transitions.

services/wallet/src/states.rs
youngjoon-lee: Thank you. Now I agree with both of you guys. I understand that we want to make the wallet service as smart as possible, so users don’t need to implement any logic separately. 
And, I understand that the note reservation is not urgent now because gas price in the testnet is 0. 
So, I agree going with this approach.

services/wallet/src/states.rs
youngjoon-lee: > If the node stops after reserving but before submitting, the voucher may be hidden after restart, but only for a short period of time.

But, how long the voucher is hidden depends on how large the `max_immutable_blocks_since_reservation` is. This PR is setting it to `security_param` (`k`) which will be 2160 in the mainnet. It means that the voucher will stay hidden for about 18h if `f=1/30`. 

What was the reasoning of setting it to `k`? It'd be great to leave the reason in the comment.

andrussal: I used `k` because in this approach the reservation is meant to cover the period until the claim either appears in LIB or can be treated as not finalized. That keeps us from reusing the same voucher while the earlier claim may still become final.

But you’re right that `k` is very long on mainnet. Maybe releasing when we first see the claim in a block is a better compromise. It is less conservative if that block does not reach LIB, but it would avoid holding reservations for a long time.

youngjoon-lee: I understand that using `k` is the safest choice. 

I think, it's a good compromise to release a reservation once the tx is included in the chain, but we can do it as a separate PR if you think it will make this PR larger.

I'm approving this PR now.

youngjoon-lee (approved):

services/wallet/src/states.rs
andrussal: Thank you!

Actually, I think keeping the reservation until LIB is the cleaner behavior.

My main concern with releasing on first inclusion is that it is still before LIB, so the same voucher becomes selectable again. The `k`-based fallback is conservative, but I think the long wait is acceptable here because it only affects the rare case where submit succeeded and the claim never finalized.

PR #2924, opened 2026-06-11
fix(cryptarchia): use strict_add for Epoch and Slot

youngjoon-lee:

This PR fixes the finding `5` in the audit https://github.com/logos-blockchain/logos-blockchain/pull/2922. 
> Unchecked Epoch - Epoch / Epoch - NumberOfEpochs subtraction panics/wraps

In this PR, I removed `Sub` impls from `Epoch` and `Slot` because they're not used anywhere.

Also, I replaced `Add` impls with `strict_add` which panics when overflow. In almost all cases, we don't want epochs and slots to silently saturate. 
I chose `strict_add` rather than `checked_add + expect`, because it's shorter and it also prints the caller info when it panics (even in release build):
```
thread 'main' (6709414) panicked at src/main.rs:3:22:
attempt to add with overflow
```

Yes

@youngjoon-lee 

Not directly related to the spec.

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/consensus/cryptarchia-engine/src/lib.rs
+++ b/consensus/cryptarchia-engine/src/lib.rs
@@ -917,7 +917,7 @@ pub mod tests {
 
         let slot = Slot::genesis();
 
-        assert_eq!(slot + 10u64, Slot::from(10));
+        assert_eq!(slot.strict_add(10.into()), Slot::from(10));
 
         let id_100 = hash(&100u64);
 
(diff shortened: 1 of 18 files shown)


danielSanchezQ (approved):

ntn-x2 (approved):
Ideally we can finally enable the lint group for unsafe arithmetic. I will try to enable it and see how much noise it generates, after we merge this.

PR #2926, opened 2026-06-11
chore(mantle): convert ChannelWithdrawOp to NomEncode/Decode trait

ntn-x2:

Following the new pattern introduced with bounded vecs.

Yes.

@ntn-x2 

N/A

No.

### DIFF
--- a/core/src/mantle/encoding.rs
+++ b/core/src/mantle/encoding.rs
@@ -19,7 +19,7 @@ use crate::{
         nom::{NomBoundedVec, NomDecode as _, NomEncode as _},
         ops::{
             Op, OpProof,
-            channel::{ChannelId, Ed25519PublicKey},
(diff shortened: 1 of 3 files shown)


hansieodendaal (approved):
LGTM

**Note:** We could add a hardcoded wire-format test vector for `ChannelWithdrawOp` stronger regression protection, as encode/decode roundtrips can still pass if both sides accidentally change the wire format together.

ntn-x2:
> Note: We could add a hardcoded wire-format test vector for ChannelWithdrawOp stronger regression protection, as encode/decode roundtrips can still pass if both sides accidentally change the wire format together.

Agree! Those tests were not present in the codebase, and I would suggest we add them as a separate PR, maybe putting those tests behind version modules, like `v1`, `v2`, etc. Feel free to take on the task!

PR #2927, opened 2026-06-11
fix(serde): Use fixed size non len prefix for arrays and use nom encoding for non human readable SignedMantleTx

danielSanchezQ:
### Remove bincode length prefixes for fixed-size data
  
  Bincode (fixint) was emitting a redundant 8-byte length prefix for every
  fixed-size field serialized via `serialize_bytes` — signatures, public keys,
  and Groth16 proofs whose sizes are compile-time constants.

  ### Changes
  - **`lb_utils::serde`**: `serialize_bytes_array` / `deserialize_bytes_array`
    now encode `[u8; N]` as a fixed-size tuple (exactly N bytes, no prefix).
    An explicit tuple visitor is required since serde only implements array
    `Deserialize` for N ≤ 32. Fixes `Ed25519PublicKey` (32B) and
    `Ed25519Signature` (64B) everywhere.
  - **`ZkSignature`** (kms/keys): pi_a/pi_b/pi_c encoded as tuples — saves
    24 bytes of prefixes per zk signature.
  - **PoL / PoC proofs** (128B): delegate to the shared helpers. PoC also
    gains the previously missing human-readable (hex) branch.

  ### Tests
  - New regression tests pin the exact encodings: a 64-byte array encodes to
    exactly 64 bytes, and an empty block to exactly 372 bytes.
  - Regenerated 4 hash-dependent test vectors: blend reward hamming distances
    are computed over the bincode bytes of `BlendingToken`, so expected
    distances/seeds changed.

  ### Breaking
  - ⚠️  Wire/storage format change: blocks and proposals are not byte-compatible
    with previous versions (re-sync or migration required).
  - Consensus IDs (`TxHash`, `HeaderId`, `ContentId`) are unaffected — none are
    derived from bincode bytes. Blend reward hamming distances **do** change.

> [!IMPORTANT]  
> explained.  

@thomaslavaur @youngjoon-lee @danielSanchezQ 

Yes, specification is complete. This is fixing an implementation mistake on serialization.

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/core/src/block/deser.rs
+++ b/core/src/block/deser.rs
@@ -69,4 +69,36 @@ mod tests {
         assert_eq!(block.header().id(), restored.header().id());
         assert_eq!(block.signature(), restored.signature());
     }
+
+    #[test]
+    fn test_bincode_fixed_size_fields_have_no_length_prefix() {
+        const VERSION: usize = 4; // u32 enum variant tag
+        const PARENT_BLOCK: usize = 32;
+        const SLOT: usize = 8;
+        const BLOCK_ROOT: usize = 32;
+        const POL_PROOF: usize = 128;
+        const ENTROPY_CONTRIBUTION: usize = 32;
+        const LEADER_KEY: usize = 32;
+        const VOUCHER_CM: usize = 32;
+        const SIGNATURE: usize = 64;
+        const TX_COUNT: usize = 8; // u64 Vec length (genuinely variable)
+        const EXPECTED: usize = VERSION
+            + PARENT_BLOCK
+            + SLOT
+            + BLOCK_ROOT
+            + POL_PROOF
+            + ENTROPY_CONTRIBUTION
+            + LEADER_KEY
+            + VOUCHER_CM
+            + SIGNATURE
+            + TX_COUNT;
+
+        let block = make_empty_block();
+        let bytes = bincode::serialize(&block).expect("bincode serialization should succeed");
+        assert_eq!(
+            bytes.len(),
+            EXPECTED,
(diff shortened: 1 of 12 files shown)


core/src/block/deser.rs
```diff
@@ -69,4 +69,36 @@ mod tests {
...
+            + SLOT
+            + BLOCK_ROOT
+            + POL_PROOF
+            + ENTROPY_CONTRIBUTION
+            + LEADER_KEY
+            + VOUCHER_CM
+            + SIGNATURE
+            + TX_COUNT;
+
+        let block = make_empty_block();
```
thomaslavaur, line 96: would be nice to have a test for a non-empty block. You can reuse the block_root_test_vectors test I added to my branch tl/pos_channel  to build a block with one of each operation type

thomaslavaur (approved):

core/src/block/deser.rs
danielSanchezQ: ill check, thanks!

ntn-x2 (approved):
Looks good! But are we sure there is nothing in our dependency tree that can solve the same issue? Somewhere in the `serde_with` or `generic_array` crates?

danielSanchezQ:
> Looks good! But are we sure there is nothing in our dependency tree that can solve the same issue? Somewhere in the `serde_with` or `generic_array` crates?

Its reusing what we have but just changing the vecs/de/serialize_bytes that adds the prefix for non prefix constructs. Maybe it could be simplified tho.

bacv (approved):

PR #2928, opened 2026-06-11
chore(mantle): migrate SDPDeclareOp to NomEncode/Decode trait

ntn-x2:

Same as usual. Built on top of https://github.com/logos-blockchain/logos-blockchain/pull/2926. Tried to keep the changes as minimal as possible for the time being to make reviewing easier. Enforcing invariants in the type system means we don't need to check the number of locators at runtime, making a whole error case impossible, hence it's been cleaned up.

Yes.

@ntn-x2 

Yes.

No.

### DIFF
--- a/core/src/mantle/nom/sdp/declare.rs
+++ b/core/src/mantle/nom/sdp/declare.rs
@@ -1,0 +1,114 @@
+use lb_key_management_system_keys::keys::{Ed25519PublicKey, ZkPublicKey};
+use lb_utils::bounded_vec::BoundedVec;
+use nom::{
+    IResult, Parser as _,
+    combinator::{map, map_res},
+};
+
+use crate::{
+    mantle::{
+        NoteId,
+        nom::{NomBoundedVec, NomDecode, NomEncode},
+        ops::sdp::SDPDeclareOp,
+    },
+    sdp::{
+        Locator, Locators, MAX_DECLARATION_LOCATOR_COUNT, MAX_LOCATOR_BYTE_SIZE, ProviderId,
+        ServiceType,
+    },
+};
+
+impl NomEncode for ServiceType {
+    fn encode(&self) -> Vec<u8> {
+        <Self as AsRef<u8>>::as_ref(self).encode()
+    }
+}
+
+impl NomDecode for ServiceType {
+    type Output = Self;
+
+    fn decode(bytes: &[u8]) -> IResult<&[u8], Self::Output> {
+        map_res(u8::decode, Self::try_from).parse(bytes)
+    }
+}
+
+type NomLocator<'a> = NomBoundedVec<'a, u8, 0, MAX_LOCATOR_BYTE_SIZE, 2>;
+
+impl NomEncode for Locator {
+    fn encode(&self) -> Vec<u8> {
+        let bounded_bytes =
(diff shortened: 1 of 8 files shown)


core/src/mantle/nom/sdp/declare.rs
```diff
@@ -0,0 +1,114 @@
...
+        map_res(u8::decode, Self::try_from).parse(bytes)
+    }
+}
+
+type NomLocator<'a> = NomBoundedVec<'a, u8, 0, MAX_LOCATOR_BYTE_SIZE, 2>;
+
+impl NomEncode for Locator {
+    fn encode(&self) -> Vec<u8> {
+        let bounded_bytes =
+            BoundedVec::new_unchecked(<Self as AsRef<[u8]>>::as_ref(self).to_owned());
```
hansieodendaal, line 39: I think we still need to do an encode-side invariant check here otherwise the new code is weaker than the old `encode_locator`, and we also need a failing unit test:
```suggestion
        let locator_bytes = <Self as AsRef<[u8]>>::as_ref(self).to_owned();
        assert!(
            locator_bytes.len() <= MAX_LOCATOR_BYTE_SIZE,
            "Fatal error in 'NomEncode for Locator' - {} locator bytes exceeds \
                {MAX_LOCATOR_BYTE_SIZE}",
            locator_bytes.len()
        );
        let bounded_bytes = BoundedVec::new_unchecked(locator_bytes);
```

hansieodendaal (commented):
Looking good, just the one issue mentioned below to consider please.

core/src/mantle/nom/sdp/declare.rs
ntn-x2: The reason why the assert is not there anymore (as it has been removed for all previous encoding methods as well) is that we bake the maximum length in the type system now. In this case not directly as a `BoundedVec` since we wrap a `Multiaddr` instead, but we still check the maximum bound at compile time. If we were to change here, we would need to change also how `NomBoundedVec` does encoding, since it calls `encode_slice` called on the inner bounded vec, which could have also been constructed using a `new_unchecked` function.

ntn-x2: Maybe I can a `debug_assert` at most, I think that would be ok.

hansieodendaal: I don’t think `debug_assert!` is enough here. `Locator` is currently `pub struct Locator(Multiaddr);`, which is not bounded in the type of `Locator` itself, so it is still possible to construct or hold a `Locator` whose encoded bytes exceed `MAX_LOCATOR_BYTE_SIZE`, and then call `NomEncode` on it.
The issue needs to be caught in release mode.

ntn-x2: But we already enforce all bounds when creating a `Locator`, the same way we do with `BoundedVec`s. I don't understand what is the difference?

ntn-x2: Would a `BoundedMultiaddr` make it maybe clearer then?

ntn-x2: https://github.com/logos-blockchain/logos-blockchain/pull/2928/commits/df87f349454dcee541dad89541bc4f8aa6ce297e

hansieodendaal (approved):
LGTM

PR #2934, opened 2026-06-12
feat: add more descriptive zone sdk events for user feedback

### LINKED ISSUE #2900: zone-sdk: Emit more states for Tx (submitted, pending, safe, finalized)
Currently user cannot distinguish if something was just submitted in the zone-sdk or was it also submitted to the mempool. It would be useful to make a split and emit more events so the user can track:
submitted -> pending (in the mempool) -> safe (maybe in the future, when we know somehting is safe) -> finalized (already present with TxsFinalized event).

hansieodendaal:

Adds zone-sdk transaction lifecycle events with source tagging:
- Introduces `TxStatusChanged`, `TxStatus`, and `TxSource`.
- Emits local accept, mempool pending, on-chain, orphaned, and finalized states.
- Tracks local transaction identity separately from pending state to preserve `Local` classification after cleanup.
- Adds cucumber coverage for the full local transaction lifecycle.
- Keeps existing `BlocksProcessed` behavior intact.

Closes #2900.

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/deployment/tui-zone/src/lib.rs
+++ b/deployment/tui-zone/src/lib.rs
@@ -137,7 +137,7 @@ fn handle_event(
             }
             state.save_checkpoint(checkpoint);
         }
-        Event::TxStatusChanged { .. } | Event::TurnNotification { .. } => {}
+        Event::MempoolPending(_) | Event::TurnNotification { .. } => {}
     }
 }
 
--- a/deployment/tui-zone/src/lib.rs
+++ b/deployment/tui-zone/src/lib.rs
@@ -137,7 +137,7 @@ fn handle_event(
             }
             state.save_checkpoint(checkpoint);
         }
-        Event::TurnNotification { .. } => {}
+        Event::TxStatusChanged { .. } | Event::TurnNotification { .. } => {}
     }
 }
 
--- a/tests/cucumber_tests/features/zone.feature
+++ b/tests/cucumber_tests/features/zone.feature
@@ -128,6 +128,11 @@ Feature: Zone SDK
       | MSG_1 | a1   |
       | MSG_2 | a2   |
       | MSG_3 | a3   |
+    And sequencer "SEQ_A" emits the full transaction lifecycle for zone messages in 30 seconds:
+      | alias |
+      | MSG_1 |
+      | MSG_2 |
+      | MSG_3 |
     Then the zone indexer returns messages in any order in 360 seconds:
       | alias |
       | MSG_1 |
@@ -269,6 +274,10 @@ Feature: Zone SDK
       | alias  |
       | MSG_B1 |
       | MSG_B2 |
+    And sequencer "SEQ_B" observed mempool pending events for zone messages:
+      | alias  |
+      | MSG_B1 |
+      | MSG_B2 |
     Then sequencer "SEQ_B" has 3 pending publish txs in 180 seconds
     Then sequencer "SEQ_B" has 1 pending publish txs in 180 seconds
     And the zone indexer returns messages in any order in 360 seconds:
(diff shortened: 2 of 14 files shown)


pradovic:
Hey @hansieodendaal. Overall looks good and matches what we agreed, nice work! A few things stood out that I think might be worth considering/fixing before merge. Sharing my notes:

1. Orphaned(Local) evicts from local_txs — I think this is a bug
```
  zone_sequencer.rs:406-408
  TxStatus::Orphaned(TxSource::Local) | TxStatus::Finalized(TxSource::Local) => {
      state.remove_local_tx(&tx_hash);
  }
  ```

  This is exactly the case we discussed: orphan is per-branch, not permanent. On a deeper reorg the same hash can land in adopted/finalized on the new branch. Because we already evicted, the follow-up events fire as Other:
  - next OnChain(...) → OnChain(Other) :x:
  - next Finalized(...) → Finalized(Other) :x:
  - next Orphaned(...) cycle → Orphaned(Other) :x:

  Suggested fix: drop Orphaned(TxSource::Local) from that match arm, keep only Finalized(Local).

2. Event::MempoolPending re-fires on every resubmit
```
  zone_sequencer.rs:374-385
  if success {
      if let Some(state) = self.state.as_mut() {
          state.mark_pending_inscription_posted(&tx_hash); // ← return value ignored
      }
      self.queue_tx_status(tx_hash, TxStatus::PendingMempool);
  }

```

  mark_pending_inscription_posted already returns a bool that's true only on the first successful post in this runtime (state.rs:454-462).
  Right now we ignore it, so every resubmit-interval tick fires another PendingMempool. From the agreed lifecycle (AcceptedLocally → PendingMempool → OnChain → Finalized), I'd expect a single PendingMempool transition.

  Suggested fix: gate the event on the bool.
  ```
  if state.mark_pending_inscription_posted(&tx_hash) {
      self.queue_tx_status(tx_hash, TxStatus::PendingMempool);
  }
```

  Or, if you intentionally want it per-post, please document that on the variant.

3. TxStatus broadcasts beat BlocksProcessed to the consumer

  actor.rs:109-126 — queue_block_status_events sends immediately on the broadcast, while the matching BlocksProcessed goes through buffered_events and only comes back on the next next_event pop. So a consumer wired to both will see Finalized(...) before BlocksProcessed{ finalized: [..] }.

Not a blocker, but worth a note on subscribe_tx_status that updates can interleave with drive-loop events.

4. Possible doc improvements
  - On TxStatus (enum-level doc): explicitly say it's a tx-hash lifecycle, not a publish-intent lifecycle, and that republishing creates a
  new lifecycle keyed by the new hash. Maybe include the three example flows you sketched (simplest / simpler / complex case).
  - On TxStatus::Orphaned: current doc says "Previously tracked tx was invalidated by a canonical chain update" — reads as permanent. Add:  it's per-branch; the same hash can resurface as OnChain/Finalized after a reorg.
  - On TxSource::Other: mention it includes "evicted from local tracking due to the bound." Without this, a consumer treating Other as a hard "not ours" signal will be silently wrong once the bound is hit.

  5. Bonus thoughts on local_txs bound (not blocking)

  Just walked the lifecycle to confirm the housekeeping piece:

  - Populated by submit_inscription, submit_atomic_withdraw, submit_other via track_local_tx — covers both fresh publishes and checkpoint
  restore (since restore_pending_tx routes through the same submit_*).
  - Pruned by:
    - soft cap prune_local_tx_tracking(max) — called after restore and on every AcceptedLocally
    - exact removal on Finalized(Local) :white_check_mark: (and on Orphaned(Local) :x:, see number 1)
  - Bound: SequencerConfig.max_local_tx_tracking = 1_000.

  On a busy sequencer 1k is easy to exceed during congestion, and we get silent degradation to Other with no event. That's fine if it's documented (see number 4). Longer term we may want a time-based bound too for txs that never resolve, but we can revisit that when we work on getting more info from the node.

pradovic (approved):
Looks great! A nice improvement!

PR #2942, opened 2026-06-15
fix(sdp): filter out inactive declarations when snapshotting

youngjoon-lee:

When taking a SDP snapshot, we were correctly filtering declarations that are becoming active. But, we were not filtering out declarations that became inactive but not-yet-GC-ed. 

This is a bug that we've had since we moved the snapshot into `EpochState`.

This PR implements the filtering-out logic by changing `EpochState` to hold only the active set instead of the entire `SdpLedger`.

Yes

@youngjoon-lee 
spec: @madxor 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/ledger/src/mantle/sdp/mod.rs
+++ b/ledger/src/mantle/sdp/mod.rs
@@ -840,7 +840,7 @@ mod tests {
 
         // The declaration is still present in the live ledger (no GC)
         assert!(ledger.get_declaration(&declaration_id).is_some());
-        // but active_declarations at epoch 10 must filter it out.
+        // but active_declarations at epoch 5 must filter it out.
         assert!(
             ledger
                 .active_declarations(5.into(), &config.service_params)
--- a/ledger/src/mantle/sdp/mod.rs
+++ b/ledger/src/mantle/sdp/mod.rs
@@ -282,11 +282,6 @@ impl<R: Rewards> ServiceState<R> {
 /// Returns true if the declaration is active at `current_epoch`:
 /// an activity message has been accepted within `inactivity_period` epochs,
 /// and its withdrawal (if any) has not yet taken effect.
-///
-/// We don't check `declaration.active >= current_epoch` because genesis
-/// declarations which are initialized with `active = 2` must be treated as
-/// active from the epoch 0. To make it clearer, the spec should update
-/// `active` to `Option` and define a clearer way to handle it.
 fn is_active(declaration: &Declaration, current_epoch: Epoch, config: &ServiceParameters) -> bool {
     declaration.active.strict_add(config.inactivity_period) >= current_epoch
         && declaration
--- a/ledger/src/mantle/sdp/mod.rs
+++ b/ledger/src/mantle/sdp/mod.rs
@@ -279,6 +279,21 @@ impl<R: Rewards> ServiceState<R> {
     }
 }
 
+/// Returns true if the declaration is active at `current_epoch`:
+/// an activity message has been accepted within `inactivity_period` epochs,
+/// and its withdrawal (if any) has not yet taken effect.
+///
+/// We don't check `declaration.active >= current_epoch` because genesis
+/// declarations which are initialized with `active = 2` must be treated as
+/// active from the epoch 0. To make it clearer, the spec should update
+/// `active` to `Option` and define a clearer way to handle it.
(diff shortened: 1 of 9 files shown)


ledger/src/mantle/sdp/rewards/blend/current_epoch.rs
```diff
@@ -105,13 +105,10 @@ impl CurrentEpochTracker {
         );
 
         let declarations = last_epoch_state
```
ntn-x2: What I see from the code, is that we access `declarations.len()` and then we call `.iter()` on it. Can we avoid `collect`ing it into a vector at all?

danielSanchezQ (approved):

services/blend/src/membership/service.rs
```diff
@@ -30,11 +30,11 @@ pub fn membership_info_from_epoch_state<NodeId>(
...
 {
-    let declarations = epoch_state.sdp.declarations();
-    let mut nodes: Vec<ZkNode<NodeId>> = declarations
-        .iter()
-        .filter(|(service_type, _)| matches!(service_type, ServiceType::BlendNetwork))
-        .flat_map(|(_, declarations)| declarations.values())
+    let mut nodes: Vec<ZkNode<NodeId>> = epoch_state
+        .active_declarations
+        .for_service(&ServiceType::BlendNetwork)
+        .into_iter()
```
ntn-x2: What about calling `.iter()` instead, since we are collecting anyway at the end into a vector?

ledger/src/mantle/sdp/mod.rs
```diff
@@ -732,6 +781,124 @@ mod tests {
...
+            let new_epoch_state = next_epoch_state(epoch.into(), last_epoch_state.clone());
+            (ledger, _) = ledger
+                .try_apply_header(&config, &last_epoch_state, &new_epoch_state)
+                .unwrap();
+            last_epoch_state = new_epoch_state;
+        }
+
+        // The declaration is still present in the live ledger (no GC)
+        assert!(ledger.get_declaration(&declaration_id).is_some());
+        // but active_declarations at epoch 10 must filter it out.
```
ntn-x2:
```suggestion
        // but active_declarations at epoch 5 must filter it out.
```

ledger/src/mantle/sdp/mod.rs
```diff
@@ -279,6 +279,21 @@ impl<R: Rewards> ServiceState<R> {
...
 
+/// Returns true if the declaration is active at `current_epoch`:
+/// an activity message has been accepted within `inactivity_period` epochs,
+/// and its withdrawal (if any) has not yet taken effect.
+///
+/// We don't check `declaration.active >= current_epoch` because genesis
+/// declarations which are initialized with `active = 2` must be treated as
+/// active from the epoch 0. To make it clearer, the spec should update
+/// `active` to `Option` and define a clearer way to handle it.
+fn is_active(declaration: &Declaration, current_epoch: Epoch, config: &ServiceParameters) -> bool {
```
ntn-x2, line 285: Can we simply expose a method to initialize a genesis declaration that is valid from epoch `0` instead, and add a lower bound check as well?

ledger/src/mantle/sdp/rewards/blend/current_epoch.rs
youngjoon-lee: Thanks. Done 9f1d1e92d02e1cc5fd54eca9ef07dd18e21f871f

services/blend/src/membership/service.rs
youngjoon-lee: Thanks. 043bfc185c65fce353f597aaf04a85abe27b1054

ledger/src/mantle/sdp/mod.rs
youngjoon-lee: This field name `active` is confusing. This is used only for checking whether the declaration should be marked as **inactive**, not for checking when the declaration becomes idle->active. The idle->active transition must be done by the `EpochState` snapshot logic, instead, (by taking a snapshot in advance for the future before new declarations are added).

I removed this comment and just added a TODO comment f7441ffc2b7e9a887be11165339de04aa76c460f, so we can clarify the purpose of the `active` field. I will open a PR soon and suggest a change in the spec.

ntn-x2 (approved):

PR #2855, opened 2026-06-04
fix(sdp): prevent stake-less service provision

youngjoon-lee:
## 1. What does this PR implement?

We had the "stake-less service provision" issue described in the [RFC: Remove concept of session](https://app.notion.com/p/nomos-tech/RFC-Remove-Concept-of-a-Session-f39261aa09df826db83e81613bb454dc?source=copy_link#357261aa09df80d79dcbeccd5487b5a5). This issue exists also in the previous session-based model (if `lock_period` is very short). 
In short, providers can continue providing services even after withdrawing their declarations, because SDP uses the snapshot from 2 epochs ago, and because the locked note is unlocked immediately as soon as the withdraw tx is accepted. 

The RFC introduces several fixes to prevent this issue. This PR implements all of them:
- [Delay unlocking the note](https://app.notion.com/p/nomos-tech/RFC-Remove-Concept-of-a-Session-f39261aa09df826db83e81613bb454dc?source=copy_link#357261aa09df807bb4c6cfa7d4863a5f) by setting `Declaration.withdraw = current_epoch + 2` when an withdraw message is accepted. The note is unlocked later once `Declaration.withdraw` epoch is reached.
- [Remove the `lock_period` parameter](https://app.notion.com/p/nomos-tech/RFC-Remove-Concept-of-a-Session-f39261aa09df826db83e81613bb454dc?source=copy_link#367261aa09df80c098f8eefe68e19c9d), as delaying note unlocking inherently provides the locking mechanism.

Also, I found that the previous implementation had a bug (even in the previous session-based model):
- Withdrawn declarations were being removed immediately when withdraw messages were accepted. But, they shouldn't be removed until the retention period has passed, as defined in the [previous spec](https://app.notion.com/p/nomos-tech/1-0-0-Service-Declaration-Protocol-1fd261aa09df819ca9f8eb2bdfd4ec1d?source=copy_link#1fd261aa09df813292abe4e33371a9c2). This PR fixes it in the way introduced by the [RFC](https://app.notion.com/p/nomos-tech/1-0-0-Service-Declaration-Protocol-33d261aa09df808f834bc728ae0c5e0f?source=copy_link#33d261aa09df80a88428cc9c1ef1afe1).  
[...]

PR #2943, opened 2026-06-15
feat(sdp): remove GC

youngjoon-lee:

In a [thread](https://discord.com/channels/973324189794697286/1514045411675865221/1514620529280024617) in Discord, we decided to drop GC from the spec and the implementation, for simplicity. 

Now declarations remain in the ledger until they are withdrawn, even if they have become inactive. They are withdrawn (unlocked and removed) after the [2-epoch delay](https://app.notion.com/p/nomos-tech/RFC-Remove-Concept-of-a-Session-f39261aa09df826db83e81613bb454dc?source=copy_link#357261aa09df80d79dcbeccd5487b5a5) since the withdraw message was accepted (this behavior was implemented by PR #2855). So, GC is no longer necessary.

This change is specified in the [RFC](https://app.notion.com/p/nomos-tech/RFC-Remove-Concept-of-a-Session-f39261aa09df826db83e81613bb454dc?source=copy_link#37d261aa09df800c887dd7f28a2fc736).

Yes

@youngjoon-lee 
spec: @madxor 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.


core/src/sdp/mod.rs
ntn-x2: Can't this be computed from `active`? And also, why is it an optional? `withdrawn` indicates, to me, the epoch at which the declaration has been withdrawn, but the comment says `The epoch at which the declaration must be withdrawn`, which I think is not correct.

ntn-x2 (commented):
I left one comment, but then realized this branch needs to be rebased since it contains changes from a different PR, which makes reviewing more challenging.

core/src/sdp/mod.rs
youngjoon-lee: I understand that it's confusing. This is how the spec is defined currently. When the withdraw message is accepted, the declaration and its locked note is not withdrawn immediately. Previously, it was withdrawn immediately, but changed recently by [RFC](https://app.notion.com/p/nomos-tech/RFC-Remove-Concept-of-a-Session-f39261aa09df826db83e81613bb454dc?source=copy_link#357261aa09df80d79dcbeccd5487b5a5) and PR #2855. So, I updated the comment (line 199) accordingly. 

In short, when the withdraw message is accepted, we set `withdrawn = Some(current_epoch + 2)`. When the `withdrawn` epoch reaches, we unlock the note. 
The `withdrawn` is `None` if no withdraw message has been accepted.

I'd like to suggest to rename it to `to_be_withdrawn` and update the [spec](https://app.notion.com/p/nomos-tech/1-0-0-Service-Declaration-Protocol-33d261aa09df808f834bc728ae0c5e0f?source=copy_link#33d261aa09df80658da9ee2232b92692).

youngjoon-lee:
@danielSanchezQ @ntn-x2 @davidrusu This PR for removing GC from SDP is now ready for reviews again!

core/src/sdp/mod.rs
youngjoon-lee: Done by a separate PR https://github.com/logos-blockchain/logos-blockchain/pull/2979

ledger/src/mantle/sdp/mod.rs
```diff
@@ -205,66 +208,37 @@ impl<R: Rewards> ServiceState<R> {
...
+                if locked_notes
+                    .is_locked_for_service(&declaration.locked_note_id, &declaration.service_type)
+                {
+                    locked_notes
+                        .unlock(declaration.service_type, &declaration.locked_note_id)
+                        .expect("unlocking note from withdrawn declaration must be successful if it hasn't been unlocked yet");
+                }
+                Some(*id)
             })
             .collect();
```
ntn-x2, line 236: I don't think we need to collect here.

ntn-x2 (approved):

ledger/src/mantle/sdp/mod.rs
youngjoon-lee: `self.declarations` is a `rpds`, which doesn't support `retain`. So, I took this approach that collects ids into `to_remove` first and remove them from the `self.declaration`, because we cannot remove elements from `rpds` while iterating over it.

ntn-x2: Ok got it! Perhaps a short comment would help explain why the need to collect then 😊

youngjoon-lee:
@danielSanchezQ Could you please have a quick look at this PR? After merging this, we can finally merge [one](https://github.com/logos-blockchain/logos-blockchain/pull/2922) of the David's audit PRs.

ledger/src/mantle/sdp/mod.rs
```diff
@@ -1065,9 +1032,10 @@ mod tests {
             Epoch::new(4) // epoch when the activity message is submitted/accepted
         );
 
-        // Move forward to the epoch 7. The declaration should be still present
-        // because the activity message was accepted at epoch 4.
-        for epoch in 5..=7 {
+        // Move forward to the epoch 8 where declaration will become inactive.
```
hansieodendaal: Maybe we could retain the `.try_apply_header(&` invariant testing below? It may be a small regression in defensive testing.

hansieodendaal (approved):
LGTM - on small test nit.

ledger/src/mantle/sdp/mod.rs
youngjoon-lee: Good point. I improved tests to check the epoch boundary more explicitly 9166d8386

ledger/src/mantle/sdp/mod.rs
```diff
@@ -44,7 +43,7 @@ impl Service {
         last_epoch_state: &EpochState,
         epoch_state: &EpochState,
         locked_notes: &mut LockedNotes,
-        config: &ServiceParameters,
+        config: ServiceParameters,
```
danielSanchezQ, line 46: nit: Why this change? even if the parameters are copy it would still be better to have a single ref than inner copies? Not really important tho.

danielSanchezQ (approved):

ledger/src/mantle/sdp/mod.rs
youngjoon-lee: Yeah. Clippy suggested to use a call by value. Let me do a last check before merging

youngjoon-lee: > warning: this argument (8 byte) is passed by reference, but would be more efficient if passed by value (limit: 8 byte)

Clippy complains about it because the size of `ServiceParameters` is only 8 bytes. I think it's okay to follow clippy's suggestion because it will also suggest to use call by ref if we add more fields into `ServiceParameters` in the future.

danielSanchezQ: Ok, makes sense, I really did not check the inner fields sizes, so fine by me.

PR #2953, opened 2026-06-15
chore(zone-sdk): add a cmd facade for publish/subscribe commands

pradovic:

Adds a cloneable `SequencerClient` (sequencer.client()) that mirrors the SequencerHandle surface async, so publishes can originate from any task — not just the drive loop. Carries the watch/broadcast subscribe methods too, so it's the single "everything I need from outside the drive loop" handle.

Usage:
```
    let client = sequencer.client();   // cheap to clone, move into any task
    client.publish(payload).await?;
    let ready = client.subscribe_ready();
```

 The drive loop must still be polled for client publish awaits to resolve; commands route through an internal channel processed inside next_event().

  Also: next_event() returns Event (not Option<Event>) — looping over transient empty steps is now internal. Drops dead events() Stream method.

Yes

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/tests/cucumber_tests/features/zone.feature
+++ b/tests/cucumber_tests/features/zone.feature
@@ -109,6 +109,45 @@ Feature: Zone SDK
     And I stop all nodes
 
   @zone_ci
+  Scenario: Publishes issued while the node is down are accepted locally and posted on reconnect
+    Given the genesis block has the following wallet resources:
+      | account_index | token_count | token_amount |
+      | 1             | 3           | 100000       |
+    And I have a cluster with capacity of 1 nodes
+    And I start nodes with wallet and sequencer resources:
+      | node_name | account_index | wallet_name | connected_to | sequencers |
+      | NODE_1    | 1             | WALLET_1A   |              | SEQ_A      |
+    When node "NODE_1" is at height 1 in 120 seconds
+    And I start zone sequencer "SEQ_A" with indexer
+    # Take the node down: the sequencer enters its reconnect loop, but its
+    # in-process SequencerClient stays alive.
+    When I stop node "NODE_1"
+    # Publishes issued via SequencerClient while the node is down must be
+    # accepted locally and return promptly — the inscription is appended to
+    # state and posted on reconnect, the same as a republish-while-down. With
+    # the pre-fix behavior these client calls would block until connectivity
+    # returned (the request was never drained during reconnect), hanging here.
+    # This step publishes without waiting for inclusion (no node calls), so it
+    # works while the node is down; inclusion/finalization is asserted after
+    # restart below.
+    And sequencer "SEQ_A" submits the following zone messages without waiting for inclusion:
+      | alias | data           |
+      | MSG_1 | While down (1) |
+      | MSG_2 | While down (2) |
+      | MSG_3 | While down (3) |
+    # Bring the node back; the locally-queued inscriptions are posted, mined
+    # and adopted, preserving publish order.
+    When I restart node "NODE_1"
+    Then all zone messages are safe in 120 seconds
+    And all zone messages are finalized in 180 seconds
+    And the zone indexer returns messages in this order:
+      | alias |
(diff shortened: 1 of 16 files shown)


hansieodendaal (commented):
Looking good, with nice code re-use for the two pathways. I have two comments to consider please.

1. The `handle().publish(...)` path and the new async `SequencerClient::publish(...).await` path have different liveness semantics during reconnect. Semantics for the two paths should be the same.
  `handle().publish(...)` can accept a publish locally after `Ready` even when the block stream is temporarily disconnected, whereas the async client request is only processed from inside `next_event()` after `ensure_connected().await` succeeds. Since `ensure_connected()` can perform network calls and sleep for `reconnect_delay`, a client request that arrives while reconnect is in progress can wait unnecessarily.
  I think the two publish pathways should preserve the same acceptance semantics. Could we structure reconnect so pending `ActorRequest`s are processed before/while reconnecting? For example, the reconnect delay could be selected against `request_rx.recv()`, allowing client requests to interrupt reconnect waiting and be processed promptly. That would better preserve the existing `handle().publish(...)` behavior where post-`Ready` publishes are accepted locally and posted once connectivity/turn state allows.

2. It would be useful to add a focused test that alternates publish sources on the same sequencer combined with a node shutdown-restart.
```
1. Start the node and initialize the sequencer until it has emitted `Ready`.
2. Shut down the node so the sequencer enters reconnect behavior.
3. While the node is down, issue:
   - one publish via `SequencerClient`;
   - one publish from the drive-loop path via `sequencer.handle().publish(...)`;
   - one more publish via `SequencerClient`.
4. Assert all three publish calls return promptly, i.e. they are accepted locally and do not wait for node connectivity.
5. Wait a short period, e.g. 5 seconds, while the node remains down.
6. Restart the node.
7. Verify all three inscriptions are eventually posted/mined/adopted.

This would cover both expected semantics:
- `SequencerClient` should preserve the same post-`Ready` local-acceptance behavior as `SequencerHandle`, even during reconnect.
- Alternating client-originated and drive-loop-originated publishes should share the same state path without corrupting ordering, checkpoints, or parentage.
```

tests/src/cucumber/steps/manual_nodes/utils.rs
```diff
@@ -765,6 +765,39 @@ pub async fn start_node(
     Ok(())
 }
 
+/// Stop a node and leave it down.
+///
+/// Unlike [`restart_node`], which brings it back up and waits for readiness,
+/// this leaves the node down. Used to exercise reconnect behavior: while the
+/// node is down, an in-process sequencer enters its reconnect loop but its
+/// `SequencerClient` stays alive.
```
hansieodendaal: This is general code - I do not think it should be tied to a sequencer in hte comments.
```suggestion
/// Unlike [`restart_node`], which brings it back up and waits for readiness,
/// this leaves the node down, useful to exercise reconnect behavior while the
/// node is down.
```

tests/cucumber_tests/features/zone.feature
```diff
@@ -108,6 +108,45 @@ Feature: Zone SDK
...
+    # in-process SequencerClient stays alive.
+    When I stop node "NODE_1"
+    # Publishes issued via SequencerClient while the node is down must be
+    # accepted locally and return promptly — the inscription is appended to
+    # state and posted on reconnect, the same as a republish-while-down. With
+    # the pre-fix behavior these client calls would block until connectivity
+    # returned (the request was never drained during reconnect), hanging here.
+    # This step publishes without waiting for inclusion (no node calls), so it
+    # works while the node is down; inclusion/finalization is asserted after
+    # restart below.
```
hansieodendaal: No need to describe the previous failure case, and I think the step is clear enough, does not warrant a comment.
```suggestion
```

hansieodendaal (approved):
LGTM

(Just two over-zealous comments to remove please)

PR #2957, opened 2026-06-16
feat(chain-network): proactive tip-poll lag watchdog

davidrusu:

Adds a **proactive tip-poll lag watchdog** to the `chain-network` service.

Today, recovery from a gossip stall (partial eclipse / network partition) is purely *reactive*: a node only catches up if a brand-new proposal happens to arrive referencing an unknown parent, which then drives the orphan downloader. If gossip simply goes quiet, the node can sit behind indefinitely without a restart or fresh IBD.

This watchdog closes that gap. On a slot-tick cadence (~one expected block interval, `ceil(1/f)` slots, where `f` is the active slot coefficient), it compares the local tip slot against the current slot. When the node has fallen behind by more than `lag_threshold_blocks` such intervals (default `3`), it samples a handful of connected peers with `GetTip` and hands the most-advanced reported tip (strictly ahead of local height) to the **existing** `OrphanBlocksDownloader`, which performs the actual catch-up download and full validation.

Key points:
- **No new download path.** It reuses the orphan downloader as-is — `enqueue_orphan(tip, local_tip, lib)` already fetches from a target back to known blocks, including the target block itself.
- **Trust model unchanged.** Polled tips are unauthenticated peer hearsay, but enqueuing one only triggers a *download*: every downloaded block still goes through normal `apply_block` validation, and a bogus tip is simply dropped on download failure. This is the same trust model orphan handling already relies on.
- **Quiet by default when healthy.** A `should_poll()` guard skips polling while a catch-up is already draining; the cadence gate keeps polling to roughly once per block interval; only tips strictly ahead of local height are enqueued (duplicates dedupe for free via `AlreadyInQueue`).
- **Default-on, fully config-gated** via `TipPollConfig { enabled, lag_threshold_blocks, max_peers_to_sample }`.

### Changes
- `sync/config.rs`: new `TipPollConfig`, added to `SyncConfig` (serde-defaulted); threaded through node binary config.
- `network`: `NetworkAdapter::sample_tips(max_peers)` with a default no-op impl (other adapters / test mocks compile unchanged); libp2p impl samples connected peers and requests tips concurrently.
- `relays.rs`: `TimeService` relay wired into chain-network for slot ticks.
- `lib.rs`: slot-tick subscription, `TipPollParams::derive` (integer `ceil(1/f)` from the active slot coeff), new `select!` arm → `poll_peer_tips_if_behind`, plus a pure `select_catchup_tip` helper.
- `metrics.rs`: `tip_poll_triggered_total`, `tip_poll_enqueued_total`.

> [!NOTE]
> No specification document is associated with this change — it is a resilience/liveness improvement to the existing sync subsystem, not a spec implementation. The rationale and trust model are documented inline (doc-comments on `poll_peer_tips_if_behind`, `TipPollParams::derive`, and `TipPollConfig`) and in the commit message. Reviewers familiar with the orphan downloader (`sync/orphan_handler.rs`) and IBD (`bootstrap/ibd.rs`) have the necessary context.

Accountable: @davidrusu. No external spec author; requesting review from the consensus/sync owners (see CODEOWNERS for `services/chain/chain-network`).

N/A — no associated specification document.

N/A — no associated specification document.

## Checklist

* [x] 2. Description added.
* [ ] 3. Context and links to Specification document(s) added. — _N/A: resilience improvement, no spec doc._
* [x] 4. Main contact(s) added.
* [ ] 5. Implementation and Specification are 100% in sync. — _N/A: no spec._
* [ ] 6. Link PR to a specific milestone. — _maintainer to set._
* [x] 7. New or changed logs were reviewed against the logging guidelines (correct levels + per-subsystem `LOG_TARGET`).

### DIFF
--- a/services/chain/chain-network/src/lib.rs
+++ b/services/chain/chain-network/src/lib.rs
@@ -641,15 +641,16 @@ where
     /// Proactive tip-poll lag watchdog.
     ///
     /// Fires on a slot-tick cadence (`params.cadence_slots`, ≈ one expected
-    /// block interval). When the local tip has fallen behind the current slot by
-    /// more than `params.lag_threshold_slots`, it samples a handful of peers
-    /// with `GetTip` and hands the most-advanced tip that is strictly ahead of
-    /// the local height to the orphan downloader, which performs the actual
-    /// catch-up download and full validation.
+    /// block interval). When the local tip has fallen behind the current slot
+    /// by more than `params.lag_threshold_slots`, it samples a handful of
+    /// peers with `GetTip` and hands the most-advanced tip that is strictly
+    /// ahead of the local height to the orphan downloader, which performs
+    /// the actual catch-up download and full validation.
     ///
     /// Polled tips are unauthenticated peer hearsay, but enqueuing one only
     /// triggers a download: every downloaded block still goes through normal
-    /// `apply_block` validation, and a bogus tip is dropped on download failure.
+    /// `apply_block` validation, and a bogus tip is dropped on download
+    /// failure.
     async fn poll_peer_tips_if_behind(
         network_adapter: &NetAdapter,
         cryptarchia: &CryptarchiaServiceApi<Cryptarchia, RuntimeServiceId>,
@@ -781,8 +782,8 @@ struct TipPollParams {
 
 impl TipPollParams {
     /// Derive the polling cadence and lag threshold from the active slot
-    /// coefficient `f`. The expected number of slots between blocks is `1/f`, so
-    /// the cadence is `ceil(1/f)` slots and the lag threshold is
+    /// coefficient `f`. The expected number of slots between blocks is `1/f`,
+    /// so the cadence is `ceil(1/f)` slots and the lag threshold is
     /// `lag_threshold_blocks` such intervals.
     async fn derive<Cryptarchia, RuntimeServiceId>(
         config: &TipPollConfig,
@@ -809,8 +810,7 @@ impl TipPollParams {
         }
 
         let cadence_slots = denominator.div_ceil(numerator).max(1);
-        let lag_threshold_slots =
-            cadence_slots.saturating_mul(config.lag_threshold_blocks.get());
+        let lag_threshold_slots = cadence_slots.saturating_mul(config.lag_threshold_blocks.get());
 
         Ok(Self {
             cadence_slots,
--- a/services/chain/chain-network/src/lib.rs
+++ b/services/chain/chain-network/src/lib.rs
@@ -701,7 +701,11 @@ where
         };
         metrics::tip_poll_triggered_total();
 
-        let tips = network_adapter.sample_tips(params.max_peers).await;
+        let tips: Vec<GetTipResponse> = network_adapter
+            .sample_tips(params.max_peers)
+            .await
+            .collect()
+            .await;
 
         // Pick the most-advanced reported tip that is strictly ahead of us.
         let Some((height, _, tip)) = select_catchup_tip(tips, info.height) else {
--- a/services/chain/chain-network/src/lib.rs
+++ b/services/chain/chain-network/src/lib.rs
@@ -694,25 +694,11 @@ where
             return;
         }
 
-        let info = match cryptarchia.info().await {
-            Ok(info) => info.cryptarchia_info,
-            Err(e) => {
-                warn!(target: LOG_TARGET, %e, "tip poll: failed to read local chain info");
-                return;
-            }
-        };
-
-        let tip_slot = u64::from(info.slot);
-        let lag = current_slot.saturating_sub(tip_slot);
-        if lag <= params.lag_threshold_slots {
+        let Some(info) =
+            Self::lagging_local_info(cryptarchia, current_slot, params.lag_threshold_slots).await
+        else {
             return;
-        }
-
-        debug!(
-            target: LOG_TARGET,
-            lag, tip_slot, current_slot,
-            "Chain tip lagging; polling peers for their tip"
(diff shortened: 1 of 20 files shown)


services/chain/chain-network/src/network/mod.rs
```diff
@@ -33,6 +33,17 @@ pub trait NetworkAdapter<RuntimeServiceId> {
...
+    /// Sample up to `max_peers` currently-connected peers and request their
+    /// chain tip via `GetTip`, concurrently. Per-peer failures are dropped, so
+    /// the returned vector only contains the responses that came back.
+    ///
+    /// Used by the proactive tip-polling lag watchdog. The default
+    /// implementation returns an empty vector, effectively disabling the
+    /// watchdog for adapters that don't support peer sampling.
+    async fn sample_tips(&self, _max_peers: usize) -> Vec<GetTipResponse> {
+        Vec::new()
+    }
```
danielSanchezQ: I would just not do this, forcing the implementors that doesnt support that to do it explicetely.

services/chain/chain-network/src/network/adapters/libp2p.rs
```diff
@@ -274,6 +274,33 @@ where
...
+        }
+
+        let requests = sampled.into_iter().map(|peer| self.request_tip(peer));
+
+        futures::future::join_all(requests)
+            .await
+            .into_iter()
+            .filter_map(Result::ok)
+            .collect()
+    }
```
danielSanchezQ: Maybe we could have returned a stream instead of awaiting and collecting all of them here ?

danielSanchezQ (commented):
Could be cleaner. But lets try it. Will let @youngjoon-lee give the last 👍

services/chain/chain-network/src/network/mod.rs
davidrusu: Done in 5025f7c4 — removed the default impl, `sample_tips` is now a required trait method so adapters opt in explicitly. The two test mocks (`ibd.rs`, `orphan_handler.rs`) implement it returning an empty stream.

services/chain/chain-network/src/network/adapters/libp2p.rs
davidrusu: Done in 5025f7c4 — `sample_tips` now returns `BoxedStream<GetTipResponse>`: it fires the GetTip requests and returns a stream over the response receivers, driven concurrently via `FuturesUnordered`, and the watchdog collects at the call site.

I went with streaming over the owned `oneshot` receivers rather than cloning `self` into per-peer futures, because the latter would require adding `Clone + 'static` bounds on `RuntimeServiceId` that ripple up through the service. This mirrors the existing `request_blocks_from_peers` pattern.

davidrusu: it reads a lot more complicated to me, but sure

services/chain/chain-network/src/lib.rs
```diff
@@ -613,6 +820,84 @@ where
...
 }
 
+/// From a set of polled tip responses, pick the most-advanced tip that is
+/// strictly ahead of `local_height`. Ties on height are broken by the higher
+/// slot. `Failure` responses are ignored. Returns `(height, slot, tip)`.
+fn select_catchup_tip(
+    tips: impl IntoIterator<Item = GetTipResponse>,
+    local_height: u64,
+) -> Option<(u64, Slot, HeaderId)> {
+    tips.into_iter()
```
youngjoon-lee: @davidrusu I'd like to discuss about this design choice. What do you think about enqueuing all `tips` to the `orphan_downloader` instead of selecting only one out of them? The `orphan_downloader` will deduplicate them and will reject them if the queue is full. We may need to exclude tips that we already have in the chain. Then, we don't need to have this `select_catchup_tip` function.

But, let's me guess your intention first:

Because we are choosing the highest tip in this function, we will lose catching up a peer's tip that is currently behind us but will grow faster in the future. But, the probability is low. Even if it's the case, the next polling will discover the tip grown faster.

Also, we don't want to flood the orphan downloader by this proactive polling because we want it to handle orphan blocks that will be received from the network. It is okay to ignore tips that are not the highest but higher than us in this function because they'll be detected in the later pollings if they grow faster.

If this is the reasoning you intended, I think we can go with this approach.

davidrusu: > Also, we don't want to flood the orphan downloader by this proactive polling because we want it to handle orphan blocks that will be received from the network. It is okay to ignore tips that are not the highest but higher than us in this function because they'll be detected in the later pollings if they grow faster.

This was my main concern.

In the rc10 devnet logs I was seeing a lot of orphan queue fully saturated logs and new orphans being dropped.

I didn't want to increase the pressure on this queue further.

youngjoon-lee (approved):

services/chain/chain-network/src/network/adapters/libp2p.rs
danielSanchezQ: yeah, claude usally do not chose the proper constructions (which we actively use in the codebase), should be better from [6e17216](https://github.com/logos-blockchain/logos-blockchain/pull/2957/commits/6e17216a14a4609a8dc56b7c7837ee68687efaab)

ntn-x2:
Yeah I agree it could be cleaned up and maybe even removed based on the Discord convo, but let's merge it and release it for now. I am not sure what's the need for the feature and the config that seem to be doing the same thing, but let's merge and deploy this, once some cucumber tests pass.

ntn-x2:
Enough tests have passed to be confident this works as expected. Merging now.

PR #2959, opened 2026-06-16
feat(blend): use key nullifiers as message IDs and implement winning slot streams

ntn-x2:

Circuts [0.4.2](https://github.com/logos-blockchain/logos-blockchain-circuits/releases/tag/v0.4.2) introduced a change that bound the key nullifier of a PoQ to a specific slot + message index combination. This means that our old approach of using a single winning slot to generate all proofs does not work anymore. 

Until now, we were using public keys as message identifiers, and since we were generating a fresh key per layer, we did not have any issues with duplicates. Nevertheless, the PoQ is bound to a key nullifier, not to the public key used in a message, hence using public keys for duplicates does not enforce the constraints that PoQ imposes. For that, we need to switch to key nullifiers, but doing that without updating the way we generate proofs would result in all proofs having the same key nullifier, which each node's de-duplication logic would filter out.

Hence, this PR introduces the following changes:

* Change the message ID from being its public key (as was defined in an old version of the spec) to the key nullifier: this allows us to enforce the requirements that PoQ states in terms of the maximum total number of messages that a node can emit as the sum of core quota (already previously enforced) and leadership quota (was not enforced before, it's enforced now)
* Because we need to know ALL winning slots now, and because epochs can have a very large number of winning slots, the notification mechanism for winning slots has been updated: we don't iterate over all slots synchronously in the leader service's loop, but we offload the computation to when subscribers (such as Blend) subscribe to the stream: at that point we find all the winning slots and use a bounded channel to handle backpressure
* Since the new circuits compute the key nullifier from the used slot, before this PR each node would generate leadership proofs using the same slot, which would result in the same key nullifier, which would result in the same PoSel, which would result in the same peer being chosen for all block proposals being blended. This is bad especially if the chosen peer is unreachable and if no redundancy is set, because that means that all blocks proposed by the node will never be delivered, increasing the chances of deep forks. Using a fresh slot every time, we ensure a new key nullifier is computed, which makes the selection of the receiver node random again, making the system more resilient to faulty Blend core nodes
* Because we now have a stream of winning slots instead, the signature on the leadership proof generators and types that depend on those has been changed to take the stream instead and react to the stream elements to generate new leadership proofs

I tried to keep it as concise as possible, limiting the diff. I suggest to review the changes with a bottom-up approach:

* `blend-message`
* `blend-scheduling`
* `blend-network`
* `chain-leader-service`
* `blend-service`
* `node-binary`

@ntn-x2 

Yes.

No, it brings the implementation up to date with the spec.


services/chain/chain-leader/src/lib.rs
```diff
@@ -39,26 +39,43 @@ use lb_tx_service::{
...
+    Pin<Box<dyn Stream<Item = WinningPolEpochSlots> + Send + Sync + Unpin>>;
 
+pub struct WinningPolEpochSlots {
+    pub epoch: Epoch,
+    pub slots: WinningPolSlotStream,
+}
+
+/// A lazy stream of one epoch's winning-slot leadership private inputs (one
+/// item per winning slot), as handed to a Blend subscriber.
+pub type WinningPolSlotStream = Pin<Box<dyn Stream<Item = LeaderPrivate> + Send + Sync + Unpin>>;
```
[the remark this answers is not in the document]
ntn-x2: `Sync` bound required by proof generators, ignoring the suggestion.

services/chain/chain-leader/src/lib.rs
```diff
@@ -39,26 +39,43 @@ use lb_tx_service::{
...
     mempool::{MempoolAdapter as _, adapter::MempoolAdapter},
     relays::CryptarchiaConsensusRelays,
     wallet::{LeaderWalletError, fund_and_sign_leader_claim_tx},
 };
 
-pub(crate) type WinningPolInfo = (LeaderPrivate, LeaderPublic, Epoch);
+/// The per-subscriber stream of per-epoch winning slots. Each item
+/// carries a single epoch and that epoch's stream of winning slots.
+pub type WinningPolEpochSlotsStream =
+    Pin<Box<dyn Stream<Item = WinningPolEpochSlots> + Send + Sync + Unpin>>;
```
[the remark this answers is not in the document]
ntn-x2: `Sync` bound required by proof generators, ignoring the suggestion.

blend/scheduling/src/message_blend/provers/mod.rs
```diff
@@ -11,6 +15,15 @@ pub mod leader;
...
 mod test_utils;
 
+/// A stream of winning-slot leadership inputs, one item per winning slot.
+///
+/// The leadership proof generator pulls a fresh winning slot from this stream
+/// for each new data message (advancing the slot every `message_quota` proofs),
+/// so each message gets a distinct key nullifier. Backpressure on the
+/// underlying channel keeps the producer from materializing the whole epoch.
+pub type WinningPolInfoStream =
+    Pin<Box<dyn Stream<Item = ProofOfLeadershipQuotaInputs> + Send + Sync>>;
```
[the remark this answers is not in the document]
ntn-x2: `Sync` bound required by proof generators, ignoring the suggestion.

services/chain/chain-leader/src/leadership.rs
```diff
@@ -177,186 +183,240 @@ pub enum PrivateInputsError {
...
+    kms: &(impl KmsAdapter<RuntimeServiceId, KeyId = KeyId> + Sync),
+    start_slot: Slot,
+    slot_tx: &mpsc::Sender<LeaderPrivate>,
+) {
+    let slots_per_epoch = ledger_config.epoch_length();
+    let epoch_first_slot: u64 = ledger_config
+        .epoch_config
+        .starting_slot(&epoch_state.epoch, ledger_config.base_period_length())
+        .into();
+    let epoch_last_slot = epoch_first_slot.saturating_add(slots_per_epoch);
```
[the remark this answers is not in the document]
ntn-x2: https://github.com/logos-blockchain/logos-blockchain/pull/2959/commits/36901d9fa871cfe6d818fe222c13325396819608

services/chain/chain-leader/src/lib.rs
```diff
@@ -691,25 +694,44 @@ where
...
+                let (epoch_handoff_sender, epoch_handoff_receiver) =
+                    mpsc::channel(WINNING_POL_EPOCH_HANDOFF_BUFFER_SIZE);
+                tokio::spawn(search_for_winning_slots(
+                    cryptarchia.clone(),
+                    wallet.clone(),
+                    kms.clone(),
+                    time_relay.clone(),
+                    ledger_config.clone(),
+                    epoch_handoff_sender,
+                ));
```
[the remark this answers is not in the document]
ntn-x2: https://github.com/logos-blockchain/logos-blockchain/pull/2959/commits/36901d9fa871cfe6d818fe222c13325396819608

services/chain/chain-leader/src/leadership.rs
```diff
@@ -177,186 +183,242 @@ pub enum PrivateInputsError {
...
+            slot,
+            &winning_slots_sender,
+        );
+        let next_epoch = next_epoch_tick(&mut slot_timer, epoch);
+        tokio::pin!(epoch_winning_slot_compute_task, next_epoch);
+        // We poll both next epoch and the scan task, so that we can stop the scan task
+        // if a new epoch starts without stopping the main loop.
+        current_slot_tick = tokio::select! {
+            () = &mut epoch_winning_slot_compute_task => (&mut next_epoch).await,
+            tick = &mut next_epoch => tick,
```
youngjoon-lee, line 299: Should we stop the scan task if a new epoch starts before the task is done, because the blend expects to receive winning slots of the **new** epoch with giving up listening for the previous epoch winning slots?

services/chain/chain-leader/src/leadership.rs
```diff
@@ -177,186 +183,242 @@ pub enum PrivateInputsError {
...
-        kms: &(impl KmsAdapter<RuntimeServiceId, KeyId = KeyId> + Sync),
-    ) {
-        if let Some((last_processed_epoch, _)) =
-            self.last_processed_epoch_and_found_first_winning_slot
+    // Process one epoch at a time, starting from whichever slot is current when
+    // we subscribe. Each iteration handles a single epoch; the `tokio::select!`
+    // at the end yields the first tick of the next epoch to process, or `None`
+    // when the tick stream ends (which ends the loop).
+    let mut current_slot_tick = slot_timer.next().await;
+    while let Some(SlotTick { slot, epoch }) = current_slot_tick {
```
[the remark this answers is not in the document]
ntn-x2: If the current slot is already ongoing, most of the computation to check whether it's winning and proposing a block has already happened, so checking that at this point is already too late. Waiting for the new slot is more correct as we run ahead of it in the spawned task and are able to perform computation based on whether it's a winning one or not.

services/chain/chain-leader/src/leadership.rs
```diff
@@ -177,186 +183,242 @@ pub enum PrivateInputsError {
...
+    slot_tx: &mpsc::Sender<LeaderPrivate>,
+) {
+    let slots_per_epoch = ledger_config.epoch_length();
+    let epoch_first_slot: u64 = ledger_config
+        .epoch_config
+        .starting_slot(&epoch_state.epoch, ledger_config.base_period_length())
+        .into();
+    let epoch_last_slot = epoch_first_slot
+        .checked_add(slots_per_epoch)
+        .expect("Epoch slot calculation overflow.");
```
youngjoon-lee: Don't we need `-1` to compute `epoch_last_slot`? I understand that the current code works because we use the range `scan_starting_slot..epoch_last_slot` below (not `scan_starting_slot..=epoch_last_slot`). But, I think it's more readable if we do `-1` and `..=`.

services/chain/chain-leader/src/leadership.rs
```diff
@@ -177,186 +183,242 @@ pub enum PrivateInputsError {
...
+    // Skip slots earlier than the start slot: a mid-epoch subscriber does not
+    // waste work on slots it has already passed.
+    let scan_starting_slot = u64::from(start_slot).max(epoch_first_slot);
+
+    for slot in scan_starting_slot..epoch_last_slot {
+        for UtxoWithKeyId { utxo, key_id } in eligible_utxos {
```
youngjoon-lee, line 386: If we have two UTXOs winning at this slot, this function will emit two `LeaderPrivate`s with the same slot number. I think it's not a big issue because distinct secrets keep nullifiers distinct. But, it would be clearer if we emit only one output for each slot, as we documented in many places. 
I think we had this issue(?) from the very beginning.

services/chain/chain-leader/src/leadership.rs
```diff
@@ -177,186 +183,242 @@ pub enum PrivateInputsError {
...
+            let public_inputs = public_inputs_for_slot(epoch_state, slot.into(), latest_tree);
```
youngjoon-lee: We can move this line to before `for UtxoWithKeyId ... in eligible_utxos`

blend/scheduling/src/message_blend/provers/leader/mod.rs
youngjoon-lee: Please correct me if I'm wrong. It seems that this `self.proofs_stream` is not finite because it is built from a per-epoch stream which scans only the limited number of slots within the epoch.
If so, is it safe to `expect` here?

services/blend/src/core/mod.rs
```diff
@@ -839,12 +839,17 @@ where
...
             Some(pol_secret_info) = secret_pol_info_stream.next() => {
                 if current_epoch_info.epoch == pol_secret_info.epoch {
-                    crypto_processor.set_epoch_private(pol_secret_info.poq_private_inputs.clone(), pol_secret_info.epoch);
+                    // Apply now: move the winning-slot stream into the current processor.
+                    crypto_processor.set_epoch_private(pol_secret_info.winning_pol_info_stream, pol_secret_info.epoch);
+                    latest_secret_pol_info = None;
+                } else {
+                    // Belongs to an upcoming epoch: keep it to seed that epoch's
+                    // processor when the rotation happens.
+                    latest_secret_pol_info = Some(pol_secret_info);
```
youngjoon-lee, line 848: If `secret_pol_info_stream` yields new `pol_secret_info`s quickly, the `latest_secret_pol_info` will keep only the last one yielded. But, is it fine because the blend service is interested in only the last (current) epoch?

youngjoon-lee (commented):
Thank you. I left some questions and suggestions but I'm not sure if I caught up all changes 100% correctly. Please let me know if I'm wrong.

services/chain/chain-leader/src/leadership.rs
ntn-x2: I am not sure if I fully understand the question. But if a new epoch starts, we want to drop the previous epoch scan and move over to the new epoch, yes. Blend will receive a new stream item with the new epoch and the new slot winning stream, that it can start polling.

services/chain/chain-leader/src/leadership.rs
ntn-x2: Why would that be an issue tho? In this case, all Blend (and other consumers) cares, is to get a list of winning slots. If more than a slot is winning, that's not an issue, is it? It will be able to generate more leadership proofs than it actually needs. But if you think it's clearer to think about, I can stop the loop when one winning UTXO is found.

services/blend/src/core/mod.rs
ntn-x2: Yes, correct. If the stream yields a new epoch, it means the previous one is to be discarded indeed, and the latest will be used to be matched against the incoming epoch public info as well.

services/chain/chain-leader/src/lib.rs
```diff
@@ -691,25 +694,44 @@ where
...
-                    });
-            }
             LeaderMsg::Claim { sender } => {
                 Self::handle_claim_message(cryptarchia, wallet, config, mempool, sender).await;
             }
+            LeaderMsg::PotentialWinningPolEpochSlotStreamSubscribe { sender } => {
+                // Spawn a dedicated, off-main-loop task that scans winning slots
+                // for this subscriber and streams them in, with the bounded
+                // channel providing backpressure.
+                let (epoch_handoff_sender, epoch_handoff_receiver) =
```
danielSanchezQ, line 719: Why a task? Cant we have a stream to poll?

danielSanchezQ (approved):

blend/scheduling/src/message_blend/provers/leader/mod.rs
ntn-x2: Yeah this is a good point. In the previous version we were using a single slot so all proofs would always be generated. This is now different. I changed that in https://github.com/logos-blockchain/logos-blockchain/pull/2959/commits/2504cd63da9811c693d4f9e80d0854e93b557ab2.

services/chain/chain-leader/src/lib.rs
ntn-x2: There's two separate components here: a per-epoch tick and a per-winning-slot tick. I am checking whether we can get rid of the task here in favor of a stream, so that the consumer would be responsible to poll it under its own schedule. Let's see if I get something not too ugly. Good point tho.

services/chain/chain-leader/src/leadership.rs
ntn-x2: https://github.com/logos-blockchain/logos-blockchain/pull/2959/commits/e6551af899986faac6921caaf5a6db19dfb180ac

services/chain/chain-leader/src/leadership.rs
ntn-x2: https://github.com/logos-blockchain/logos-blockchain/pull/2959/commits/4ffdd4755bc8142a7f4fc5e229e19690922f1fc9

services/chain/chain-leader/src/lib.rs
danielSanchezQ: @ntn-x2 We can open an issue and do it later too.

services/chain/chain-leader/src/leadership.rs
youngjoon-lee: > It will be able to generate more leadership proofs than it actually needs.

Yeah, I don't think it's actually an issue. While reading doc comments in many places, I set a mental model that the stream yields items "per slot". But here, two items can have the same slot number. That was confusing to me, but it's okay if that's not an issue.

youngjoon-lee (approved):

PR #2963, opened 2026-06-17
feat(c-bindings): deposit operation

pradovic:
This PR adds channel deposit operations to the FFI bindings. There are two deposit calls:
  - channel_deposit_with_notes — the simpler call; the caller supplies the exact notes to deposit (their whole value enters the channel). A 1:1 port of the existing POST /channel/deposit handler.
  - channel_deposit — the automatic call; the caller passes an amount and the binding
  selects the notes itself, splitting one via a transfer when no exact-value note exists.

  It also adds get_wallet_notes (+ free_wallet_notes), which exposes a wallet address's spendable notes (note ID, value). This is needed so callers can discover the note IDs to pass to channel_deposit_with_notes — the existing get_balance binding only returned a total and discarded the per-note list.

Yes

@pradovic 

yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/c-bindings/src/api/wallet.rs
+++ b/c-bindings/src/api/wallet.rs
@@ -25,7 +25,6 @@ use lb_node::{
     generic_services::{CryptarchiaService, WalletService as NodeWalletService},
 };
 use lb_wallet_service::{ClaimableVoucherInfo, TipResponse, api::WalletApi};
-use num_bigint::BigUint;
 use overwatch::services::status::ServiceStatus;
 
 use crate::{
@@ -745,11 +744,8 @@ pub unsafe extern "C" fn transfer_funds(
     } else {
         lb_core::header::HeaderId::from(unsafe { *arguments.optional_tip })
     };
-    let change_public_key = {
-        let change_public_key_bytes =
-            unsafe { std::slice::from_raw_parts(arguments.change_public_key, 32) };
-        ZkPublicKey::from(BigUint::from_bytes_le(change_public_key_bytes))
-    };
+    let change_public_key =
+        unwrap_or_return_error!(unsafe { parse_public_key(arguments.change_public_key) });
     let funding_public_keys = {
         let funding_public_keys_pointers = unsafe {
             std::slice::from_raw_parts(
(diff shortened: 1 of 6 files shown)


c-bindings/src/api/wallet.rs
```diff
@@ -645,3 +784,670 @@ pub unsafe extern "C" fn transfer_funds(
...
+#[repr(C)]
+pub struct ChannelDepositWithNotesArguments {
+    /// Optional pointer to the header ID to build against. If null, the current
+    /// tip is used.
+    pub optional_tip: *const HeaderId,
+    /// The 32-byte channel ID to deposit into.
+    pub channel_id: *const u8,
+    /// Array of pointers to 32-byte note IDs to deposit. Their combined value
+    /// is the deposited amount.
+    pub input_note_ids: *const *const u8,
```
danielSanchezQ: I think is time to introduce a note_id type maybe?

c-bindings/src/api/wallet.rs
```diff
@@ -645,3 +784,670 @@ pub unsafe extern "C" fn transfer_funds(
...
+    let mut selected = Vec::new();
+    let mut total: Value = 0;
+    for (id, value) in notes {
+        if total >= amount {
+            break;
+        }
+        selected.push(id);
+        total = total.saturating_add(value);
+    }
+    (total >= amount).then_some((selected, total))
```
danielSanchezQ, line 1158: this is a naive implementation, at some point we should use some dynamic programming here (knapsack problem). Also, isnt this in the wallet already? 
Otherwise this should be somewhere else to be reused, c-bindings is probably not the best place.

c-bindings/src/api/wallet.rs
pradovic: Agreed. https://github.com/logos-blockchain/logos-blockchain/pull/2963/changes/4baa53409aa6a5556f3780ab4f98f9681eb755f8

c-bindings/src/api/wallet.rs
pradovic: ﻿﻿ Good points, agreed on both.

 Yes, it's in Wallet::fund_tx (wallet/src/lib.rs:164/:515), same naive largest-first + change heuristic (so knapsack would apply there too).

  `channel_deposit` can't just reuse fund_tx because of the split path: (1) we need the note ID of the freshly-created deposit output to point the ChannelDeposit at it, but MantleTxBuilder exposes no accessor for the pending transfer's outputs; and (2) the deposit consumes that same-tx output, so sign_tx (resolves keys from committed ledger) fails with MissingInputNote -> hence the manual assembly + sign_tx_with_zk.

  Agreed it shouldn't live in c-bindings. 
  
  We could create a reusable selection primitive in the wallet crate that the FFI calls, or  a wallet-level method that builds + signs the whole atomic transfer+deposit so the FFI just calls one function (and knapsack has one home)? 
  
  Should we try to do it in this PR or leave it as is with todo?

danielSanchezQ: Ok, let do it in a later pr. imo this should be some kind of buider, can be an specific builder on top of mantletxbuilder. But will do for now.

danielSanchezQ (approved):

PR #2968, opened 2026-06-18
fix(chain-network): replace queued orphan parent with descendant orphan

### LINKED ISSUE #2960: Deduplicate orphans in the queue
When inserting a new orphan into the queue, check if its parent is in the queue. If so, replace the parent with the new orphan, because we're going to download all ancestors of the orphan in the end.

youngjoon-lee:

Resolves #2960 

When enqueuing a new orphan, we now check if there is its parent in the queue. If so, we replace the parent with the new orphan, in order to decrease the queue pressure.

Also, I found that we were returning the `QueueIsFull` error more aggressively than necessary. We can silently ignore orphan blocks if they're already in the queue, without returning the error. I fixed it by reordering the logic in the `enqueue` function.

Yes

@youngjoon-lee 

This is an implementation details, not defined in the chain sync spec.

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/services/chain/chain-network/src/sync/orphan_handler.rs
+++ b/services/chain/chain-network/src/sync/orphan_handler.rs
@@ -128,9 +128,37 @@ where
     pub fn enqueue_orphan(
         &mut self,
         block_id: HeaderId,
+        parent_id: Option<HeaderId>, // `None` if the caller knows only `block_id`
         current_tip: HeaderId,
         lib: HeaderId,
     ) -> Result<(), OrphanEnqueueError> {
+        if let DownloaderState::Downloading(download) = &self.state
+            && download.orphan_block_id() == block_id
+        {
+            debug!(target: LOG_TARGET, ?block_id, "Orphan block is already being downloaded, skipping enqueue");
+            return Err(OrphanEnqueueError::AlreadyDownloading);
+        }
+
+        if self.pending_orphans_queue.contains_key(&block_id) {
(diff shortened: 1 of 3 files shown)


services/chain/chain-network/src/sync/orphan_handler.rs
```diff
@@ -128,9 +128,37 @@ where
...
+        {
+            debug!(
+                target: LOG_TARGET,
+                ?block_id, ?parent_id,
+                "replacing queued parent orphan with descendant"
+            );
+            self.remove_orphan(&parent_id);
+            metrics::orphan_blocks_replaced_total();
+        }
+
```
danielSanchezQ, line 161: Wait, I don't get why we replace downloading the parent for the child (?)

youngjoon-lee: When we try to enqueue a new orphan block (that was just discovered), we now check if its parent is already in the queue. If it's the case, we remove the parent from the queue because both blocks are on the same branch. When the descendant orphan (that is just being enqueued) will be processed by the orphan downloader, we will download all ancestors of it, including the parent block that we just removed from the queue. 

So, in short, we do this to save the space in the queue. 
Catching up the history from the block-100 is a super-set of catching up from the block-99.

danielSanchezQ: Understood, thanks!

danielSanchezQ (approved):

PR #2970, opened 2026-06-18
fix(mantle): omit empty transfer op and avoid needless funding input

davidrusu:

Two related fixes to transaction construction. Previously every transaction ended up carrying a `Transfer` op even when it moved no value, because of two independent behaviours:

**(a) `MantleTxBuilder::build` always appended the pending transfer.** When a tx moved no value (no inputs, no outputs) this produced an empty transfer that carries no information and would fail validation with `TransferError::NoInputTransfer`. `build()` now only appends the pending transfer when it is non-empty (new `TransferOp::is_empty`), allowing genuinely transfer-less transactions (e.g. an SDP op that pays no fee).

**(b) `WalletState::fund_tx` always pulled in an input before checking if funding was needed.** The loop started at the first UTXO and only evaluated `funding_delta` *after* `extend_ledger_inputs`, so the zero-input state was never tested. As a result a tx that needs no funding — it pays no fee (gas prices are `0` today, see #2490), or the caller already balanced it — still pulled in an input and added a change note to hand that value straight back: an unnecessary **self-transfer**. It also returned `InsufficientFunds` when no funding was needed but the wallet had no spare notes.

`fund_tx` now checks the funding delta of the tx *as supplied* first:
- `== 0` → return it untouched (no input, no change),
- `> 0` → add only a change note (no new input),
- `< 0` → fall back to the existing loop that pulls in UTXOs.

Together these remove the spurious Transfer op that was appearing on fee-less transactions (e.g. SDP Active txs).

Changes are local to `MantleTxBuilder::build` / `TransferOp::is_empty` (`core/`) and `WalletState::fund_tx` (`wallet/`). `funding_delta = (Σinputs − Σoutputs) − gas_cost`, so a non-positive delta means the tx is already funded. The two fixes are coupled: with (b), an already-funded fee-less tx returns a builder with an empty pending transfer, and (a) is what makes `build()` omit it so the tx stays valid.

Three existing wallet tests (`test_fund_tx_zero_funds`, `test_fund_tx_all_locked_notes`, `test_fund_tx_respects_pk_list`) asserted `InsufficientFunds` using zero gas prices on an empty tx — which no longer needs funding. They now use `GasPrices::new(1, 1)` so they still exercise the underfunded path and the locked-note / pk-list filtering they were written to check. New test `test_fund_tx_no_fee_adds_no_input` covers the no-fee path: no input pulled in, no transfer op in the built tx.

@davidrusu (accountable). Reviewers TBD.

N/A — transaction-builder / wallet funding correctness fixes; no specification change.

No.

## Checklist

* [x] 2. Description added.
* [x] 7. New or changed logs were reviewed against the logging guidelines. (No log changes.)

### DIFF
--- a/core/src/mantle/tx_builder.rs
+++ b/core/src/mantle/tx_builder.rs
@@ -253,10 +253,12 @@ impl MantleTxBuilder {
     // TODO: Change this to a `Result` if genesis tx already contains max number of
     // ops.
     pub fn build(mut self) -> Result<MantleTx, TxBuilderError> {
-        self.mantle_tx
-            .0
-            .try_push(Op::Transfer(self.pending_transfer))
-            .map_err(|err| TxBuilderError::from((err, TooManyTag::Ops)))?;
+        if !self.pending_transfer.is_empty() {
+            self.mantle_tx
+                .0
+                .try_push(Op::Transfer(self.pending_transfer))
+                .map_err(|err| TxBuilderError::from((err, TooManyTag::Ops)))?;
+        }
         Ok(self.mantle_tx)
     }
 }
--- a/core/src/mantle/ops/transfer.rs
+++ b/core/src/mantle/ops/transfer.rs
@@ -25,6 +25,11 @@ impl TransferOp {
         Self { inputs, outputs }
     }
 
+    #[must_use]
+    pub const fn is_empty(&self) -> bool {
+        self.inputs.is_empty() && self.outputs.is_empty()
+    }
+
     pub fn balance(&self, utxos: &Utxos) -> Result<i128, TransferError> {
         let mut balance: i128 = 0;
         let input_amount = self.inputs.amount(utxos)?;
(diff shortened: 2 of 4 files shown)


core/src/mantle/tx_builder.rs
```diff
@@ -253,10 +253,12 @@ impl MantleTxBuilder {
     // TODO: Change this to a `Result` if genesis tx already contains max number of
     // ops.
     pub fn build(mut self) -> Result<MantleTx, TxBuilderError> {
-        self.mantle_tx
-            .0
-            .try_push(Op::Transfer(self.pending_transfer))
-            .map_err(|err| TxBuilderError::from((err, TooManyTag::Ops)))?;
+        if !self.pending_transfer.is_empty() {
```
andrussal, line 256: Question: do we consider fully empty transactions valid? Because `build()` now can return an empty `MantleTx` when no ops were added, I wanted to check whether that is expected.

andrussal (approved):
LGTM!

(One comment)

core/src/mantle/tx_builder.rs
davidrusu: yes, nothing in our specs prevent it.. we need to study this to make sure it's not a DOS vector though. I'll open an issue!

davidrusu: https://github.com/logos-blockchain/logos-blockchain/issues/3031

PR #2974, opened 2026-06-18
chore(nom): migrate SDPActiveOp to NomEncode/Decode trait

ntn-x2:

Removed most of the code that was dealing with vectors, since a Blend activity proof has a well-defined, statically-known size, so no need for additional runtime logic.

Yes.

@ntn-x2 

N/A

No.


youngjoon-lee (approved):

core/src/mantle/nom/sdp/mod.rs
```diff
@@ -21,3 +25,160 @@ impl NomDecode for DeclarationId {
...
+            return Err(nom::Err::Error(Error::new(
+                bytes,
+                nom::error::ErrorKind::Fail,
+            )));
+        }
+        let (bytes, epoch) = Epoch::decode(bytes)?;
+        let (bytes, signing_key) = Ed25519PublicKey::decode(bytes)?;
+        let (bytes, proof_of_quota) = ProofOfQuota::decode(bytes)?;
+        let (bytes, proof_of_selection) = ProofOfSelection::decode(bytes)?;
+        Ok((
```
hansieodendaal, line 89: Should we not verify that remaining `bytes` are empty here before returning? It feels to me decode should fail if given too may bytes.

hansieodendaal (commented):
Looking good overall, just one query below.

core/src/mantle/nom/sdp/mod.rs
ntn-x2: That's the thing: nom does zero-copy stuff, so the `bytes` that are remaining belong to the next type that is encoded in the tx. The only place where we want to check that no bytes are left is in the outermost container, so at the edge where decoding happens, i.e., when receiving a block proposal from the network. Otherwise, from the perspective of the `SDPActiveOp`, it does not know what the remaining bytes represent.

hansieodendaal (approved):
LGTM

PR #2975, opened 2026-06-18
chore(nom): do not manually parse

ntn-x2:

Built on top of https://github.com/logos-blockchain/logos-blockchain/pull/2974.

This PR removes all calls to `parse` from the types and delegates that to the primitive types, encapsulating the logic. The overall checks should be performed at the serialization/deserialization boundary, which will be a whole `SignedMantleTx`, when we get to the point of implementing `NomEncode` and `NomDecode` on that type.

Yes.

@ntn-x2 

N/A

No.

### DIFF
--- a/core/src/mantle/nom/mod.rs
+++ b/core/src/mantle/nom/mod.rs
@@ -2,8 +2,7 @@ use lb_cryptarchia_engine::Epoch;
 use lb_groth16::{Fr, fr_from_bytes, fr_to_bytes};
 use lb_key_management_system_keys::keys::ZkPublicKey;
 use nom::{
-    IResult, Parser as _,
-    combinator::{map, map_res},
+    IResult,
     error::{Error, ErrorKind},
     number::complete::{le_u16, le_u32, le_u64, u8},
 };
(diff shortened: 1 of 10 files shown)


danielSanchezQ (approved):

youngjoon-lee (approved):

core/src/mantle/nom/mod.rs
```diff
@@ -149,10 +152,12 @@ impl NomDecode for Ed25519PublicKey {
...
-        map_res(NomArray::<u8, 32>::decode, |key_bytes: [u8; 32]| {
-            Self::from_bytes(&key_bytes).map_err(|_| Error::new(bytes, ErrorKind::Fail))
-        })
-        .parse(bytes)
+        let (bytes, inner) = NomArray::<u8, _>::decode(bytes)?;
+        Ok((
+            bytes,
+            Self::from_bytes(&inner)
+                .map_err(|_| nom::Err::Error(Error::new(bytes, ErrorKind::MapRes)))?,
+        ))
```
hansieodendaal, line 160: For all places where this pattern is used, the input slice attached to the nom error now does not contain the full complement of bytes as was the case previously, but only the remainder bytes. This could be misleading for downstream diagnostics.

Please verify whether switching to `Error::new(bytes, ErrorKind::MapRes)` is intentional here.

hansieodendaal (commented):
Looking good, just one clarification needed please.

core/src/mantle/nom/mod.rs
ntn-x2: What are you suggesting here then?

hansieodendaal: Maybe one of these?
- Keep the original input bytes in scope for error creation.
- Use an error marker/context so the error can tell which conversion failed (and remove the remaining bytes in the error).

ntn-x2: Got it. Please check https://github.com/logos-blockchain/logos-blockchain/pull/2975/commits/c687e670edd07e88e14d086cb5b495146eee8b7b.

hansieodendaal (approved):
LGTM

PR #2976, opened 2026-06-18
fix(wallet): don't add an input when the tx is already funded

davidrusu:

`WalletState::fund_tx` always pulled in at least one of the wallet's UTXOs as an input *before* checking whether funding was needed. The loop starts at the first UTXO and only evaluates `funding_delta` after `extend_ledger_inputs`, so the zero-input state is never tested. Consequences:

- A transaction that needs no funding — e.g. it pays no fee (gas prices are `0` today, see #2490), or the caller already balanced it — still pulled in an input and then added a change note to hand that value straight back: an unnecessary **self-transfer**. (This is the bundled Transfer op we see on SDP Active txs.)
- If no funding was needed but the wallet held no spare notes, `fund_tx` returned `InsufficientFunds` even though nothing needed funding.

This PR checks the funding delta of the transaction *as supplied* first:
- `== 0` → return it untouched (no input, no change),
- `> 0` → add only a change note (no new input),
- `< 0` → fall back to the existing loop that pulls in UTXOs.

The change is local to `WalletState::fund_tx` in `wallet/src/lib.rs`. `funding_delta = (Σinputs − Σoutputs) − gas_cost`, so a non-positive delta means the tx is already funded.

**Stacked on #2970** (`fix/skip-empty-transfer-op`). When the tx is already funded with no inputs/outputs (the no-fee case), `fund_tx` now returns a builder with an empty pending transfer; #2970 is what makes `build()` omit that empty transfer so the resulting tx is valid. This PR targets that branch and should merge after it.

Three existing tests (`test_fund_tx_zero_funds`, `test_fund_tx_all_locked_notes`, `test_fund_tx_respects_pk_list`) asserted `InsufficientFunds` using zero gas prices on an empty tx — which no longer needs funding. They now use `GasPrices::new(1, 1)` so they still exercise the underfunded path (and the locked-note / pk-list filtering they were written to check). New test `test_fund_tx_no_fee_adds_no_input` covers the no-fee path: no input pulled in, no transfer op in the built tx.

@davidrusu (accountable). Reviewers TBD.

N/A — wallet funding-logic correctness fix; no specification change.

No.

## Checklist

* [x] 2. Description added.
* [x] 7. New or changed logs were reviewed against the logging guidelines. (No log changes.)

### DIFF
--- a/wallet/src/lib.rs
+++ b/wallet/src/lib.rs
@@ -183,6 +183,22 @@ impl WalletState {
         // Consume large valued notes first to ensure we converge.
         utxos.sort_by_key(|utxo| -i128::from(utxo.note.value));
 
+        // The transaction may already be funded before we add any of the
+        // wallet's UTXOs, for example when it pays no fee or the caller already
+        // supplied inputs. In that case we must not pull in an extra input (and
+        // the change note it would require).
+        match tx_builder.funding_delta::<G>()?.cmp(&0) {
+            Ordering::Equal => return Ok(tx_builder.clone()),
+            Ordering::Greater => {
+                if let Some(tx_with_change) = tx_builder.clone().return_change::<G>(change_pk)? {
+                    return Ok(tx_with_change);
+                }
+                // The change note costs more than the surplus covers, so fall
+                // through and pull in more UTXO's below.
+            }
+            Ordering::Less => {}
+        }
+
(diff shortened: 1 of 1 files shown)


davidrusu:
Folding this into #2970 per review — both changes ship as one PR.

PR #2978, opened 2026-06-19
feat(wallet): track leader reward outputs

andrussal:

In order to make claimed leader rewards visible in wallet balances, this PR emits a LeaderRewardClaimed event when a leader claim is applied.

The event records the reward public key and amount. The wallet derives the corresponding reward note from that event, so leader rewards are reflected in balance once the claim appears in an applied block.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


core/src/events/mod.rs
```diff
@@ -79,6 +83,12 @@ pub enum EventPayload {
         amount: Value,
         metadata: Metadata,
     },
+    /// A leader claim operation created the reward note for its beneficiary.
+    LeaderRewardClaimed {
+        voucher_nullifier: VoucherNullifier,
+        pk: ZkPublicKey,
+        amount: Value,
+    },
```
davidrusu, line 95: I think this should hold the generated UTXO, not just the pk/amount
```suggestion
    LeaderRewardClaimed {
        voucher_nullifier: VoucherNullifier,
        utxo: Utxo,
    },
```

wallet/src/lib.rs
```diff
@@ -340,6 +340,38 @@ impl WalletState {
...
+fn leader_reward_utxos_from_events(events: &Events) -> impl Iterator<Item = Utxo> + '_ {
+    events.iter().filter_map(|event| match event {
+        Event::Tx {
+            op_id,
+            payload: EventPayload::LeaderRewardClaimed { pk, amount, .. },
+            ..
+        } => Some(Utxo::new(*op_id, 0, Note::new(*amount, *pk))),
+        _ => None,
+    })
+}
```
davidrusu, line 372: yeah, lets put the utxo in the event rather than rederiving the UTXO here. (we want to avoid duplicating this detail about how leader utxo's are created (output index 0, op_id is the txhash, etc.)

davidrusu (approved):
Generally looks good, but lets just change the event to hold the utxo so that we don't spread the utxo derivation logic around the code base.

andrussal:
> Generally looks good, but lets just change the event to hold the utxo so that we don't spread the utxo derivation logic around the code base.

Good point, I changed to utxo.

core/src/mantle/ops/leader_claim.rs
```diff
@@ -212,8 +213,20 @@ impl Operation<LeaderClaimValidationContext<'_>> for LeaderClaimOp {
...
+            ctx,
+            std::iter::once(Event::from_tx(
+                tx_hash,
+                self.op_id(),
+                EventPayload::LeaderRewardClaimed {
+                    voucher_nullifier: self.voucher_nullifier,
+                    utxo,
+                },
+            ))
+            .collect(),
```
danielSanchezQ: Nitpick: why not construct it `Events(vec![Event::from_tx(..)])`?
Or `impl From<Event> for Events` rather than wrap everything in an iterator for a single item.

core/src/mantle/channel.rs
```diff
@@ -409,22 +409,25 @@ mod tests {
...
+                    payload:
+                        EventPayload::Deposit {
+                            channel_id,
+                            amount,
+                            metadata,
+                        },
+                } => Some((*tx_hash, *op_id, *channel_id, *amount, metadata.clone())),
+                _ => None,
+            })
+            .expect("events should include deposit event");
```
danielSanchezQ: you can keep the pattern matching on find:
```suggestion
        let Event::Tx {
                    tx_hash,
                    op_id,
                    payload:
                        EventPayload::Deposit {
                            channel_id,
                            amount,
                            metadata,
                        },
                }) = events
            .iter()
            .find(|e| matches!(e, Event::Tx)
            .expect("events should include deposit event");
```

ledger/src/lib.rs
```diff
@@ -1112,22 +1113,25 @@ mod tests {
...
-            op_id,
-            payload,
-        } = events.iter().next().unwrap().clone()
-        else {
-            panic!("expected a Tx event")
-        };
-        assert_eq!(tx_hash, tx.hash());
+        let (event_tx_hash, op_id, event_channel_id, amount, metadata) = events
+            .iter()
+            .find_map(|event| match event {
```
danielSanchezQ: Same as above

wallet/src/lib.rs
```diff
@@ -340,6 +340,37 @@ impl WalletState {
...
+    known_keys: &HashMap<ZkPublicKey, KeyId>,
+    utxos: &mut rpds::HashTrieMapSync<NoteId, Utxo>,
+    pk_index: &mut rpds::HashTrieMapSync<ZkPublicKey, rpds::HashTrieSetSync<NoteId>>,
+) {
+    if !known_keys.contains_key(&utxo.note.pk) {
+        return;
+    }
+
+    let note_id = utxo.id();
+    *utxos = utxos.insert(note_id, utxo);
```
danielSanchezQ: there is an `insert_mut` that should be used instead.

wallet/src/lib.rs
```diff
@@ -340,6 +340,37 @@ impl WalletState {
...
+
+    let note_set = pk_index
+        .get(&utxo.note.pk)
+        .cloned()
+        .unwrap_or_else(rpds::HashTrieSetSync::new_sync)
+        .insert(note_id);
+    *pk_index = pk_index.insert(utxo.note.pk, note_set);
```
danielSanchezQ: same `insert_mut` iirc

danielSanchezQ (approved):
Just nitpicks, otherwise 🟢

wallet/src/lib.rs
```diff
@@ -340,6 +340,37 @@ impl WalletState {
     }
 }
 
+fn insert_owned_utxo<KeyId>(
```
youngjoon-lee:
```suggestion
fn insert_utxo_if_owned<KeyId>(
```
I think it's better to rename this function because the caller doesn't know whether the utxo is owned or not. It is checked by this function.

youngjoon-lee (approved):

wallet/src/lib.rs
andrussal: > I think it's better to rename this function because the caller doesn't know whether the utxo is owned or not. It is checked by this function.

Done, renamed it to `insert_utxo_if_owned`.

PR #2979, opened 2026-06-19
refactor(sdp): rename `Declaration::withdrawn` to `withdraw_at`

youngjoon-lee:

The `Declaration` struct was having a field `withdrawn` as defined in the [spec](https://app.notion.com/p/nomos-tech/1-0-0-Service-Declaration-Protocol-33d261aa09df808f834bc728ae0c5e0f?source=copy_link#33d261aa09df80658da9ee2232b92692). But, it could be misleading. Declarations are not withdrawn immediately as soon as the withdraw message is accepted. Instead, they are scheduled to be withdrawn at the epoch `current + 2` (since [this RFC](https://app.notion.com/p/nomos-tech/RFC-Remove-Concept-of-a-Session-f39261aa09df826db83e81613bb454dc?source=copy_link#357261aa09df80d79dcbeccd5487b5a5) and PR #2855). This was also pointed out in https://github.com/logos-blockchain/logos-blockchain/pull/2943/changes#r3418331695.

So, this PR renames the `withdrawn` field to `withdraw_at`, which is much clearer, I think.
And I’ll also propose updating the RFC accordingly to Marcin.

The `Declaration` struct also has another field `active` which is also not so clear, I think. I will propose a new name in a separate PR. 

Yes

@youngjoon-lee 
spec: @madxor 

Yes, although it would be great to reflect this naming update in the spec as well.

Yes, as mentioned above.

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/core/src/mantle/ops/sdp/active.rs
+++ b/core/src/mantle/ops/sdp/active.rs
@@ -40,13 +40,13 @@ impl Operation<SDPActiveValidationContext<'_>> for SDPActiveOp {
         };
 
         // Check the declaration hasn't been withdrawn
-        // (Return error if `withdrawn` epoch has passed)
-        if let Some(withdrawn) = declaration.withdrawn
-            && withdrawn <= ctx.epoch
+        // (Return error if `scheduled_withdrawal_epoch` epoch has passed)
+        if let Some(withdraw_at) = declaration.withdraw_at
+            && withdraw_at <= ctx.epoch
         {
             return Err(SdpError::DeclarationWithdrawn {
                 declaration_id: self.declaration_id,
-                withdrawn_epoch: withdrawn,
+                withdraw_at,
             });
         }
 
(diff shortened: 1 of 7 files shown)


ntn-x2 (approved):
Looks good but I think we need to clarify what withdrawal means. Is it the act of withdrawing at epoch `N`, or the fact that the declaration is removed at epoch `N+2`?

danielSanchezQ (approved):

youngjoon-lee:
> Looks good but I think we need to clarify what withdrawal means. Is it the act of withdrawing at epoch `N`, or the fact that the declaration is removed at epoch `N+2`?

After PR https://github.com/logos-blockchain/logos-blockchain/pull/2943 is merged, withdrawal will imply removal. There will be no separate concept of removal anymore. So, it means that the declaration will be withdrawn + removed at the `withdraw_at = N+2` epoch.

Before the PR #2943 is merged, the `withdraw_at` field means "the locked note of the declaration will be unlocked at the `withdraw_at = N+2` epoch". The declaration will be removed at `withdraw_at + retention_period`. 

But I think it’s sufficient to focus only on the post-PR semantics.

PR #2938, opened 2026-06-12
feat: add cli commands to tui-sequencer enabling roundtrip wallet-zone-wallet demo

hansieodendaal:
## 1. What does this PR implement?

**Implementation:**

Adds a manual Zone deposit/withdrawal workflow to `tui-zone` and supporting cucumber manual-control tooling.

- Adds `tui-sequencer` subcommands for interactive inscription, key generation, `state full` / `state balance`, deposit, withdrawal prepare/sign/combine/submit, and channel config apply/prepare/sign/combine/submit.
- Adds file-based JSON contracts for exported wallet funds, withdrawal intents, withdrawal signatures, signed withdrawal transactions, config intents, config signatures, and signed config transactions.
- Extends cucumber manual transaction controls with wallet funds export support for manual Zone demos.
- Adds a non-interactive sequencer driver loop for one-shot commands so deposits, config updates, and withdraw submissions are driven until observed on-chain, with optional finalized waiting.
- Persists CLI sequencer checkpoints from block-processing events so one-shot commands can resume from the same channel state without replaying from genesis.
- Adds channel-id validation across transferred files and supports explicit `--channel-id` for non-admin authorized keys.
- Makes one-shot command summaries print compact channel balance output, while `state full` exposes the full channel state snapshot.
- Expands the manual demo documentation to cover deposits, single-signer withdrawal, 2-of-3 withdrawal, and multi-signer config update flows.
- Adds focused unit coverage for TUI command helpers and file-contract validation.

**Testing (manual demo sequence):**

1. Start the cucumber `Transactions manual control` scenario and export wallet funds for `WALLET_1A` and `WALLET_2A`.
2. Start `tui-sequencer`, create the Zone channel with `seq-a.key`, publish one inscription to validate the channel, then stop the interactive sequencer.
3. Generate `seq-b.key`, `seq-c.key`, and later `seq-d.key` for multi-signer authorization flows.
4. Submit two deposits into the Zone using the exported wallet funds, then confirm cucumber wallet balances reflect the deposits.
5. Run the single-signer withdrawal flow:
       prepare intent -> sign with `seq-a.key` -> combine -> submit -> confirm `WALLET_1A` balance increases.
6. Reconfigure the Zone channel to authorize `seq-a`, `seq-b`, and `seq-c`, with `withdraw-threshold = 2` and `configuration-threshold = 2`.
7. Run the multi-signer withdrawal flow:
[...]

PR #2982, opened 2026-06-21
fix: zone sdk stale last_msg_id bug

hansieodendaal:

Fixed a Zone SDK stale-parent bug in prepare_tx.

- `ZoneSequencer::do_prepare_tx` previously used `self.last_msg_id` directly when building a parented channel inscription bundle. That diverged from the normal publish path, which uses `compute_publish_parent()` to derive
  the current safe parent from sequencer state.

- This could produce stale-parent transactions after checkpoint restore or after another command advanced the channel tip. In the manual TUI flow, a multi-signature withdraw prepared after a config update could be
  accepted locally but never land on-chain because the prepared bundle was still parented to an older message.

- The fix makes `prepare_tx` use the same parent-selection path as publish, so manually prepared transactions follow the current channel lineage instead of trusting checkpoint-derived `last_msg_id`.

- Normal interactive publish did not hit this because it already used `compute_publish_parent()`; the bug was isolated to SDK callers using the “prepare but don’t fully publish now” API path.

**Note:** This bug surfaced in #2938.

Yes

@hansieodendaal @pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/zone-sdk/src/sequencer/zone_sequencer.rs
+++ b/zone-sdk/src/sequencer/zone_sequencer.rs
@@ -830,12 +830,13 @@ where
         data: Inscription,
     ) -> Result<(MantleTx, MsgId, Ed25519Signature), Error> {
         self.ensure_ready()?;
+        let parent = self.compute_publish_parent();
         Ok(build_prepare_tx(
             ops,
             self.channel_id,
             &self.signing_key,
             data,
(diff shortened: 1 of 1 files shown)


pradovic (approved):
Good catch!

PR #2997, opened 2026-06-22
fix(wallet): track channel withdraw outputs

andrussal:

This fixes wallet balance tracking after channel withdraws.

Withdraw outputs are now passed to the wallet through block events, using the same approach as leader reward claims instead of reconstructing them in wallet code. This fixes the stale balance shown in the UI after a withdraw.

Includes focused wallet coverage and an e2e check for the balance update.

Related [Discord](https://discord.com/channels/973324189794697286/1518016374897643710)

---

One possible follow-up is to keep moving wallet sync toward typed block events.

At the moment wallet balance tracking is a bit mixed: some data comes from parsing operations directly, while leader rewards and channel withdraws now come from events. It might be cleaner over time if each relevant operation event carried the wallet-visible data it already knows about, and then the matching wallet-side operation parsing could be removed.

For example:
- `Deposit` could include spent note IDs
- `Withdraw` already includes created UTXOs
- `Transfer` could include spent note IDs and created UTXOs
- `LeaderRewardClaimed` already includes the reward UTXO
- SDP events could include locked/unlocked note IDs

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/core/src/mantle/ops/channel/withdraw.rs
+++ b/core/src/mantle/ops/channel/withdraw.rs
@@ -2,7 +2,7 @@ use nom::IResult;
 use serde::{Deserialize, Serialize};
 
 use crate::{
-    events::Events,
+    events::{Event, EventPayload, Events},
     mantle::{
         TxHash,
         channel::{Channels, Error},
@@ -64,6 +64,7 @@ pub struct WithdrawValidationContext<'a> {
 pub struct WithdrawExecutionContext {
     pub channels: Channels,
     pub utxos: Utxos,
+    pub tx_hash: TxHash,
 }
 
 impl Operation<WithdrawValidationContext<'_>> for ChannelWithdrawOp {
@@ -151,9 +152,21 @@ impl Operation<WithdrawValidationContext<'_>> for ChannelWithdrawOp {
             })
         }?;
 
-        // Add the ouputs to the ledger
+        // Add the outputs to the ledger
+        let output_utxos = self.outputs.utxos(self).collect::<Vec<_>>();
         ctx.utxos = self.outputs.execute(ctx.utxos, self);
 
-        Ok((ctx, Events::new()))
+        let events = Event::from_tx(
+            ctx.tx_hash,
+            self.op_id(),
+            EventPayload::Withdraw {
+                channel_id: self.channel_id,
+                amount: amount_withdraw,
+                utxos: output_utxos,
+            },
+        )
+        .into();
+
+        Ok((ctx, events))
     }
 }
--- a/core/src/mantle/ops/channel/withdraw.rs
+++ b/core/src/mantle/ops/channel/withdraw.rs
@@ -153,9 +153,9 @@ impl Operation<WithdrawValidationContext<'_>> for ChannelWithdrawOp {
         }?;
 
         // Add the outputs to the ledger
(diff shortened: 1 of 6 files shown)


wallet/src/lib.rs
```diff
@@ -257,6 +261,10 @@ impl WalletState {
             insert_utxo_if_owned(*utxo, known_keys, &mut utxos, &mut pk_index);
         }
 
+        for utxo in &block.channel_withdraw_utxos {
+            insert_utxo_if_owned(*utxo, known_keys, &mut utxos, &mut pk_index);
+        }
+
```
danielSanchezQ, line 267: Can we collapse this in a single loop with chained iterators instead?

danielSanchezQ (approved):

wallet/src/lib.rs
andrussal: > Can we collapse this in a single loop with chained iterators instead?

Done, collapsed the reward and withdraw UTXO loops with `chain`.

youngjoon-lee:
@andrussal Yesterday, I opened a PR https://github.com/logos-blockchain/logos-blockchain/pull/3000 which fixes a bug in the wallet. That PR may conflict with this PR, but I will resolve the conflicts if you merge this PR first. 
I think, my PR improves some of the things that you mentioned as the "follow-up" in your PR description, though there's something more to improve (as I mentioned in my PR description).

core/src/mantle/ops/channel/withdraw.rs
```diff
@@ -151,9 +152,21 @@ impl Operation<WithdrawValidationContext<'_>> for ChannelWithdrawOp {
             })
         }?;
 
-        // Add the ouputs to the ledger
+        // Add the outputs to the ledger
+        let output_utxos = self.outputs.utxos(self).collect::<Vec<_>>();
         ctx.utxos = self.outputs.execute(ctx.utxos, self);
```
youngjoon-lee, line 156: I think you can move the line 156 to the line 165, because the comment "Add the outputs to the ledger" doesn't explain what the line 156 does.

youngjoon-lee (approved):

andrussal:
> @andrussal Yesterday, I opened a PR #3000 which fixes a bug in the wallet. That PR may conflict with this PR, but I will resolve the conflicts if you merge this PR first.
> I think, my PR improves some of the things that you mentioned as the "follow-up" in your PR description, though there's something more to improve (as I mentioned in my PR description).

Thanks! I think this PR can go first.

For the follow-up, I was wondering if ordered ledger events could become the single source of truth for wallet-visible changes. That would give the built-in wallet and external wallets the same interface for what changed inside each applied block, instead of each wallet having to define its own mapping from operation data to wallet updates.

youngjoon-lee:
> Thanks! I think this PR can go first.
> 
> For the follow-up, I was wondering if ordered ledger events could become the single source of truth for wallet-visible changes. That would give the built-in wallet and external wallets the same interface for what changed inside each applied block, instead of each wallet having to define its own mapping from operation data to wallet updates.

I'm not sure if I understand your point correctly, but I think my idea is somehow similar as what you said. I think we can define the interface which can be used by all block consumer modules. I will try it soon.

PR #2999, opened 2026-06-22
chore(API): Attach id (TxHash) on human readable formats for signedmantletx

danielSanchezQ:
It just adds the tx id as part of human readable responses used in the API and the bindings. This should help with interoperability without need of reimplementing the hashing somewhere else. It leaves untouched the binary representation use for wire formats.

Its actually really easy

@danielSanchezQ @youngjoon-lee @hansieodendaal 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.


hansieodendaal (commented):
I think the new test should validate that the new id field is actually present in the JSON output, otherwise it could be removed and the test will still pass. We could add this:

```rust
let json_value: serde_json::Value = serde_json::from_str(&json)?;
assert!(
    json_value.get("id").is_some(),
    "Human-readable JSON must include the transaction id"
);
```

danielSanchezQ:
> I think the new test should validate that the new id field is actually present in the JSON output, otherwise it could be removed and the test will still pass. We could add this:
> 
> ```rust
> let json_value: serde_json::Value = serde_json::from_str(&json)?;
> assert!(
>     json_value.get("id").is_some(),
>     "Human-readable JSON must include the transaction id"
> );
> ```

Added in [5b5ac1b](https://github.com/logos-blockchain/logos-blockchain/pull/2999/commits/5b5ac1b2a7e18382079585cfe63b893b9948e5fd) thanks @hansieodendaal

hansieodendaal (approved):
LGTM

bacv (approved):

youngjoon-lee (approved):

hansieodendaal:
All e2e and cucumber tests fail with 
```
    Location:
        nodes/node/binary/src/cli/mod.rs:449:13
    Error: Unrecognized fields in value: ["cryptarchia.genesis_block.transactions.0.id"]
```

Asking about it I get this answer (untested):

---

Fix: make id truly optional in human-readable deserialization.

Your error proves strict config loading is seeing id as unrecognized at:

```
  nodes/node/binary/src/cli/mod.rs:449
    path cryptarchia.genesis_block.transactions.0.id
```

In this PR, _id is declared with `#[serde(default, alias = "id")]`.
That can fail in practice with strict unknown-field checks depending on derive behavior/context.

**What to change**
In core/src/mantle/tx.rs, update the helper struct field from alias-based underscore field to a direct optional id field:

```rust
#[derive(Deserialize)]
struct SignedMantleTxHelper {
    // Consume optional human-readable `id` emitted by Serialize.
    // It's informational only; actual id is recomputed in Self::new.
    #[serde(default)]
    id: Option<TxHash>,
    mantle_tx: MantleTx,
    ops_proofs: Vec<OpProof>,
}
```
Then destructure ignoring id:

```rust
let SignedMantleTxHelper {
    id: _,
    mantle_tx,
    ops_proofs,
} = SignedMantleTxHelper::deserialize(deserializer)?;
```

**Why this fixes it**
- `serde_ignored` / strict unknown-key loaders only care that incoming keys are consumed.
- A real `id` field guarantees `"id"` is consumed.
- `Option<TxHash> + #[serde(default)]` keeps backward compatibility when older payloads omit id.

**Also update tests**
Your tests are close already; keep/add assertions for:
- Serialize includes id
- Deserialize works without id
- serde_ignored sees no ignored fields (the strict path)

danielSanchezQ:
> All e2e and cucumber tests fail with

What is weird is that I have a test using serde_ignored but it didnt triggered that.

danielSanchezQ:
@hansieodendaal, Im closing this PR. The issue is that we use the txs for config and we should not touch that format. Instead im gonna create a wrapper type for what we return in the API and the bindings where the field would be used.

PR #3000, opened 2026-06-22
fix(wallet): apply txs/ops in source order to avoid phantom UTXOs

youngjoon-lee:
### Issues

The way how the wallet service processes blocks was different from how the chain-leader and the chain service build/process blocks.

A tx in a block may depend on other txs in the same block. For example, 
```
block: [tx1, tx2]
tx1: transfer note1 -> note2
tx2: transfer note2 -> note3
```
In this reason, the chain service processes txs one by one sequentially.

However, the wallet service was flattening the input notes of all txs. It was processing all input notes first, and then was processing output notes. This is wrong, and cause phantom notes.
```
all_inputs: note1, note2
all_outputs: note2, note3

remove note1 from wallet
remove note2 from wallet (but no-op because note2 doesn't exist in the wallet)
add note2 to the wallet (a "phantom UTXO" that doesn't exist in the ledger)
add note3 to the wallet
```

This issue was found from the following error from the faucet. If the phantom note, `note2`, is the largest note owned by `funding_pk`, it will be always picked by the wallet service and cause the error.
image

### Fix

Respect the tx order when processing blocks in the wallet service. For that, I refactored the `WalletBlock` structure.

In addition, after https://github.com/logos-blockchain/logos-blockchain/pull/2978 and https://github.com/logos-blockchain/logos-blockchain/pull/2997 were merged, the wallet should track tx events as well as tx operations. The way how those PRs apply the events to the wallet is also wrong, in the same reason I described above. So, in this PR, we pair the events with the tx operations to bulid the `WalletBlock` correctly.

### Future work

I have an idea to modularize the block processing nicer and use it for both ledger and wallet. I will open an issue and implement it soon

Yes

@youngjoon-lee 

We don't have spec for wallet.
The Mantle spec is correct and doesn't need any change.

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.


bacv (approved):
Thank you very much! I couldn't reproduce error after this change.

danielSanchezQ (approved):
Much cleaner too!!

PR #3001, opened 2026-06-22
chore: change logic to specify custom deployments

ntn-x2:

As discussed internally, we initially envisioned the system of deployments without knowing exactly our release strategy. Now, it is clear that the current solution has some limitations. The most important one is that it is hard to keep all potential well-known deployments shipped in a single binary compatible with each other. E.g. releasing a new binary candidate that works for a new devnet most likely won't work with the old testnet, so it would be wrong to ship a `testnet` well-known config in such binary. In general, even if we version each and every deployment we have (i.e., all devnet deployments) as well-known deployment, it becomes a mess to handle and also new versions of the binary might ignore some fields or require new ones, and of course we don't keep track of any versions.

Hence the solution devised is simple: each binary has a single well-known deployment that is embedded at compile time from the `nodes/node/binary/src/config/deployment/settings.yaml` file. For our `master` node, this is equivalent to the old `standalone-deployment-config.yaml`. When releasing a new candidate, it will be the upcoming devnet spec, and when releasing a new version, it will be the upcoming testnet spec. This means that a testnet binary won't be usable "by default" against a devnet deployment, and that is by design, as we do not guarantee compatibility. It will be up to the users to get a hold of the devnet deployment settings and use the old way of overriding it with `--deployment devnet-deployment.yaml`.

Another major change is that we won't be merging releases into `master` at all anymore, while releases will keep branching off of `master` as is already the case: this means that on `master` we keep a `0.0.0` version for all workspace crates. Each release process will change that to the right version before building the binaries and publishing the release.

See context above.

@ntn-x2 

N/A

No.


scripts/standalone-genesis-ceremony.sh
```diff
@@ -4,4 +4,4 @@
   --providers deployment/ceremony/genesis/standalone/providers.yaml \
   --faucet deployment/ceremony/genesis/standalone/faucet.yaml \
   --deployment deployment/ceremony/genesis/standalone/deployment-template.yaml \
-  --output nodes/node/standalone-deployment-config.yaml
+  --output nodes/node/binary/src/config/deployment/settings.yaml
```
bacv: Everything looks good, but I would still consider the standalone config a seperate case, and wouldn't put it to the "well known" section even on master. The standalone config is special because it has `nodes/node/standalone-node-config.yaml` keys inside the genesis block, which allows this pair of configs to produce blocks without connecting to any remote peers. 

I would keep master well known config as a blank deployments config with all values set to zeroes, this way it would be quite easy for devs to update it when the format changes.

ntn-x2: Reverted relevant changes: https://github.com/logos-blockchain/logos-blockchain/pull/3001/commits/f5332df6a7fe55463aa5895f92967cc9f9ceeb32.

youngjoon-lee (approved):
I agree that this approach is clean.

davidrusu (approved):

hansieodendaal (commented):
I appreciate the intent of this PR but have some doubts.

- Pubsub topic version isolation regression risk
Changing from explicit versioned topics (e.g. `...devnet-0.1.3-rc.7...`) to templated `.../X.Y.Z` can weaken safety unless version values are always concretely provisioned per release/deployment.
We should keep enforced version provisioning to avoid cross-release interference.

- Missing version-specific test strategy
The PR strategy doesn’t clearly define version-targeted tests, and relying on generic CI is insufficient.
We should add explicit version-specific unit/e2e coverage as part of release preparation.

- **Release-process gate proposal**
For each new release, add a version-tagged deployment config into master, validate it through unit/e2e tests, then branch for release.
We should make this a required gate, and test effective loaded config/invariants (not just config path selection).

ntn-x2:
@hansieodendaal 

> Pubsub topic version isolation regression risk
Changing from explicit versioned topics (e.g. ...devnet-0.1.3-rc.7...) to templated .../X.Y.Z can weaken safety unless version values are always concretely provisioned per release/deployment.
We should keep enforced version provisioning to avoid cross-release interference.

That is still the case. We just mark it as a template on `master`, but our CI workflows still have steps to set the right values. This PR is not aiming at changing that just yet, but rather remove multi-deployment binaries.

> Missing version-specific test strategy
The PR strategy doesn’t clearly define version-targeted tests, and relying on generic CI is insufficient.
We should add explicit version-specific unit/e2e coverage as part of release preparation.

Agree, but once again, it is not within the scope of this PR. Please feel free to take the current release process and make it more robust, as this is a team effort. We will tackle all these issues one at the time, but this PR is not concerned with that.

> Release-process gate proposal
For each new release, add a version-tagged deployment config into master, validate it through unit/e2e tests, then branch for release.
We should make this a required gate, and test effective loaded config/invariants (not just config path selection).

Not sure I fully understand your proposal here. We can still add CI jobs and release checklist items that are to be enforced on a release branch. We don't need stuff to be merged on `master` for that to happen. We should already check that CI is green by the time we fork off `master`, and if more checks are needed on the release branch, we can do that. Again, the release process is going to be incrementally improved over the next weeks.

bacv (approved):

PR #3003, opened 2026-06-22
chore(blend): set data replication factor to 1 for devnet deployment template

ntn-x2:


ntn-x2:
Clippy failure is unrelated and already fixed in a different PR

bacv (approved):

PR #3006, opened 2026-06-23
fix(mantle): bind LeaderClaim voucher nullifier via the proof's public input

### LINKED ISSUE #2990: [CRITICAL] LeaderClaim dedup uses op.voucher_nullifier (not the proof-bound nullifier) → unlimited re-claim of leader rewards
# [CRITICAL] LeaderClaim dedup uses the op's `voucher_nullifier`, not the proof-bound one → unlimited re-claim of leader rewards (value theft)

## Summary

The Proof-of-Claim (PoC) is verified against the proof's **own** `voucher_nf` (which the circuit
binds to the voucher secret), but the double-spend / replay set is keyed on the
`LeaderClaimOp.voucher_nullifier` field — an **independent, attacker-controlled wire value** that is
**never checked to equal the proof-bound nullifier**. A holder of a single valid voucher can
therefore claim it an unlimited number of times: each claim reuses a (freshly generated, tx-bound)
valid proof for the same voucher while supplying a different, unused `voucher_nullifier` so the dedup
check never fires. This drains the entire `claimable_rewards` pool and mints reward UTXOs to the
attacker — value inflation / theft and a consensus-integrity break.

## Severity

**Critical** (value conservation / token inflation). Reachable by any party that legitimately owns
one voucher (i.e. any leader who has earned a single reward), on the authoritative block-execution
path. No fork required — all nodes deterministically mint the bogus rewards.

## Root cause

There are two distinct nullifier values that are never bound together:
1. `LeaderClaimOp.voucher_nullifier` — decoded from the transaction bytes (attacker-controlled).
   Used for the dedup set check **and** insertion.
2. `Groth16LeaderClaimProof.voucher_nf` — the nullifier the ZK proof actually binds to the voucher
   secret. Used as a public input during PoC verification.

`LeaderClaimOp::validate` checks (a) `op.voucher_nullifier` ∉ set, (b) `rewards_root` matches, and
(c) the PoC verifies against `proof.voucher_nf` — but never asserts
`op.voucher_nullifier == proof.voucher_nf`. The `leader.rs::claim` comment even states the binding
"is expected to happen outside of this function" — it happens nowhere (a repo-wide search for any
equality check between `voucher_nullifier` and `voucher_nf` finds none).

The `mantle_tx_hash` public input does **not** mitigate this: the voucher owner knows the voucher
secret and can generate a fresh, valid proof bound to each new transaction's hash, all sharing the
same proof-bound `voucher_nf`.

## Code reference

- `core/src/mantle/ops/leader_claim.rs:180-200` — `validate`: dedup on `self.voucher_nullifier`
  (line 182); PoC verified via `proof_of_claim.verify(&LeaderClaimPublic{ voucher_root,
  mantle_tx_hash })` (line 192) — note `voucher_nullifier` is **not** among the verified public
  inputs.
- `core/src/mantle/ops/leader_claim.rs:202-217` — `execute`: inserts `self.voucher_nullifier`
  (line 207) and mints the reward UTXO.
- `core/src/proofs/leader_claim_proof.rs:64-78` — `verify` uses `self.voucher_nf()` (the proof's
  own stored nullifier) as the PoC public input.
- `ledger/src/mantle/leader.rs:185-209` — `claim`: dedup/insert on `op.voucher_nullifier`; comment
  defers the binding "outside".
- `ledger/src/lib.rs:664-694` — block-execution arm: calls `validate` then `execute` (so the gap is
  on the consensus path).

## Worked failure scenario (reproduced by a failing test)

A test was added at `core/src/mantle/ops/leader_claim.rs`,
`tests::validate_rejects_reclaim_of_already_spent_voucher`, and it **fails** against current code:

```
assertion `left == right` failed: a voucher whose proof-bound nullifier is already
spent must not be reclaimable by supplying a different op.voucher_nullifier
  left: Ok(())
 right: Err(DuplicatedVoucherNullifier)
```

The test: generate a valid proof for voucher secret `s` (proof-bound `voucher_nf = N`); record `N`
as already spent; then call `validate` for an op with `voucher_nullifier = M ≠ N` and the same valid
proof. It returns `Ok(())` — i.e. the already-claimed voucher is re-claimable.

Attack loop: for `i = 1, 2, …`, build a `LeaderClaim` tx with `voucher_nullifier = M_i` (fresh unused
value) and a fresh proof for the same voucher bound to that tx's hash. Each passes, inserts `M_i`,
and mints `reward_amount` to the attacker. As bogus nullifiers accumulate, `n_unclaimed` shrinks and
per-claim reward grows, draining `claimable_rewards`.

## Suggested fix

In `LeaderClaimOp::validate`, bind the two nullifiers:

```rust
if self.voucher_nullifier != *ctx.proof_of_claim.voucher_nf() {
    return Err(LeaderClaimError::InvalidPoC); // or a dedicated NullifierMismatch
}
```

Better still, **remove** `LeaderClaimOp.voucher_nullifier` entirely and use
`proof.voucher_nf()` for both the dedup check and the insertion, so there is only one nullifier and
it is always the proof-bound one. Keep the added regression test.

davidrusu:

Fixes #2990 (**Critical** — reward-pool drain).

The bug: the Proof-of-Claim was verified against a `voucher_nf` **stored in `Groth16LeaderClaimProof`**, while the double-spend set was keyed on the `LeaderClaimOp`'s **separate** `voucher_nullifier` field — two copies of the same value, with nothing forcing them equal. The JSON path's derived `Deserialize` let them diverge: an attacker proves voucher `N` but dedups under an unused `M`, re-claiming the same voucher under fresh nullifiers and draining `claimable_rewards`. (See #2990 for the quantified exploit.)

**We fix the bug by making the bad state unrepresentable.** Instead of carrying the nullifier twice and checking the copies agree, we remove the duplicate: `Groth16LeaderClaimProof` becomes just `{ proof }`, and `op.voucher_nullifier` is fed in as the PoC verifier's public input (via `LeaderClaimPublic`). There is now exactly **one** nullifier — the one on the op — so "a proof whose nullifier differs from the op's" is no longer a representable state. The proof verifies only when `op.voucher_nullifier` is the value the circuit derived from the proven voucher, which is what binds the dedup key to the voucher. A forged/mismatched nullifier simply fails verification (`InvalidPoC`); the separate equality check (and its error variant) is gone because there is nothing left to keep in sync.

- The binary wire already never carried the nullifier twice (`decode` reconstructed it from the op); this removes the redundant in-memory/JSON copy and unifies the two paths.
- Regression test `validate_rejects_op_nullifier_not_matching_proof`: a claim supplying a nullifier other than the proven voucher's fails verification. `validate_accepts_valid_proof_of_claim` still passes.
- Spec: [Proof of Leadership](https://lip.logos.co/blockchain/raw/cryptarchia-proof-of-leadership.html) / [Anonymous Leaders Reward](https://lip.logos.co/blockchain/raw/bedrock-anonymous-leaders-reward.html).

> PoL / anonymous-leaders-reward spec authors: please add as reviewers. Author accountable for resolution.

Yes — one-nullifier-per-voucher dedup is implied by the spec; the implementation had decoupled the dedup key from the proof.

No (it removes a redundant on-wire/in-memory field; the proof statement is unchanged).

## Checklist

* [x] 2. Description added.
* [x] 5. Implementation and Specification are in sync.
* [x] 7. New/changed logs reviewed (none added).

---

### DIFF
--- a/core/src/mantle/encoding.rs
+++ b/core/src/mantle/encoding.rs
@@ -157,11 +157,8 @@ fn decode_op_proof<'a>(input: &'a [u8], op: &Op) -> IResult<&'a [u8], OpProof> {
         }
 
         // ProofOfClaimProof = Groth16
-        Op::LeaderClaim(leader_claim_op) => map(decode_groth16, |proof| {
-            OpProof::PoC(Groth16LeaderClaimProof::new(
-                proof,
-                leader_claim_op.voucher_nullifier,
-            ))
+        Op::LeaderClaim(_) => map(decode_groth16, |proof| {
+            OpProof::PoC(Groth16LeaderClaimProof::new(proof))
         })
         .parse(input),
 
@@ -1253,18 +1250,14 @@ mod tests {
             pk: ZkPublicKey::from(BigUint::from(0u64)),
         };
 
-        let mantle_tx = MantleTx(Ops::new_unchecked(vec![Op::LeaderClaim(
-            leader_claim_op.clone(),
-        )]));
+        let mantle_tx = MantleTx(Ops::new_unchecked(vec![Op::LeaderClaim(leader_claim_op)]));
 
         let empty_gas_context =
             MantleTxGasContext::new(HashMap::new(), HashMap::new(), GasPrices::new(0, 0));
         let predicted_size = predict_signed_mantle_tx_size(&mantle_tx, &empty_gas_context);
 
-        let poc_proof = Groth16LeaderClaimProof::new(
-            CompressedGroth16Proof::from_bytes(&[0u8; 128]),
-            leader_claim_op.voucher_nullifier,
-        );
+        let poc_proof =
+            Groth16LeaderClaimProof::new(CompressedGroth16Proof::from_bytes(&[0u8; 128]));
 
         // Construct directly to skip proof verification (dummy proof won't verify)
         let signed_tx = SignedMantleTx {
@@ -1281,15 +1274,12 @@ mod tests {
         use crate::proofs::leader_claim_proof::Groth16LeaderClaimProof;
 
         let proof_bytes: [u8; 128] = core::array::from_fn(|i| i as u8);
-        let voucher_nf = VoucherNullifier::default();
-        let poc_proof = Groth16LeaderClaimProof::new(
-            CompressedGroth16Proof::from_bytes(&proof_bytes),
-            voucher_nf,
-        );
+        let poc_proof =
+            Groth16LeaderClaimProof::new(CompressedGroth16Proof::from_bytes(&proof_bytes));
 
         let leader_claim_op = LeaderClaimOp {
             rewards_root: RewardsRoot::default(),
-            voucher_nullifier: voucher_nf,
+            voucher_nullifier: VoucherNullifier::default(),
             pk: ZkPublicKey::from(BigUint::from(0u64)),
         };
         let op = Op::LeaderClaim(leader_claim_op);
@@ -1303,7 +1293,6 @@ mod tests {
             decoded,
             OpProof::PoC(Groth16LeaderClaimProof::new(
                 CompressedGroth16Proof::from_bytes(&proof_bytes),
-                voucher_nf,
             ))
         );
     }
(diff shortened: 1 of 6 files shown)


youngjoon-lee (approved):
much safer. thank you 👍

PR #3007, opened 2026-06-23
fix(mantle): enforce strictly-increasing signature indices in ChannelMultiSigProof

### LINKED ISSUE #2985: CHANNEL_CONFIG validation skips the duplicate-index check → multisig authorization bypass
> ⚠️ **Security-sensitive.** This is an authorization bypass. Consider private
> disclosure (security advisory) before any public discussion.

## Summary

`ChannelConfigOp::validate` enforces the signature *count* against
`configuration_threshold` but never checks that the signature **indexes are unique**. The
spec requires the indexes to be strictly increasing (hence unique). A `ChannelMultiSigProof`
carrying duplicate indexes therefore passes validation, letting the holder of a **single**
accredited key satisfy an N-of-M configuration multisig.

## Severity

High confidence (clear spec divergence + authorization gap). Exploitability: the binary
p2p/block decode path is a backstop (see below), so the clean attack surface is the JSON RPC
`add_tx` endpoint corrupting the receiving node's channel state; a node that builds a block on
that corrupted state forks itself off.

## Spec reference

`logos-lips/docs/blockchain/raw/bedrock-v1.1-mantle-specification.md`, `CHANNEL_CONFIG`
Validation:

```python
# Check that indexes are ordered to avoid duplication
for i in range(len(proof.indexes)-1):
    assert proof.indexes[i] < proof.indexes[i+1]   # strictly increasing ⇒ unique
```

## Code reference

`core/src/mantle/ops/channel/config.rs:100-140` — `validate()` checks only
`signatures.len() == channel.configuration_threshold` and then verifies each signature
against `accredited_keys[index]`. **No uniqueness check.** The comment at `:101-102` claims
it is "enforced by the proof structure," but that dedup/well-formedness lives only in
`ChannelMultiSigProof::new()` (`core/src/proofs/channel_multi_sig_proof.rs:62-66`).

`ChannelMultiSigProof` derives `Deserialize` on its private `signatures` field
(`channel_multi_sig_proof.rs:55-59`), so serde-based deserialization bypasses `new()`:

- JSON mempool path: `nodes/node/binary/src/api/handlers.rs:683` → `Json<SignedMantleTx>` →
  derived `Deserialize` → duplicates preserved → reaches `validate()`.
- Binary p2p/block decode path **is safe**: `core/src/mantle/encoding.rs:270`
  (`decode_channel_multi_sig_proof`) routes through `ChannelMultiSigProof::new()`, which
  sorts/dedups and rejects duplicate indexes.

The sibling **`CHANNEL_WITHDRAW`** path *does* enforce uniqueness
(`core/src/mantle/tx.rs:525-532`, `verify_channel_withdraw` builds a `HashSet` and rejects
duplicate indexes). This asymmetry indicates the `CHANNEL_CONFIG` omission is unintentional.

## Reproduction (worked, currently failing)

`core/src/mantle/ops/channel/config.rs`, test
`spec_divergence_tests::channel_config_rejects_duplicate_indexes_per_spec`:

```
test ...::channel_config_rejects_duplicate_indexes_per_spec ... FAILED
panicked: spec requires strictly-increasing signature indexes, but validate()
accepted two signatures at the same index — a single key satisfied a 2-of-2
configuration threshold (auth bypass)
```

The test builds a 2-of-2 channel, then submits two copies of one key's signature at index 0
via a serde-deserialized proof (the JSON path), and shows `validate()` returns `Ok`.

## Impact

A single accredited key can pass an N-of-M `configuration_threshold`, then rotate out the
other sequencers, lower `withdraw_threshold`, and drain channel funds. Bounded by the binary
decode backstop to the JSON-submission surface / self-forking, but still a real authorization
gap and spec violation.

## Suggested fix

Add the strictly-increasing / uniqueness check inside `ChannelConfigOp::validate` (mirror
`verify_channel_withdraw`), **and/or** give `ChannelMultiSigProof` a custom `Deserialize`
that routes through `new()` so no deserialization path can produce an unnormalized proof.

davidrusu:

Fixes #2985 (**High** — multisig threshold bypass).

`CHANNEL_CONFIG` validation relied on the `ChannelMultiSigProof` invariant (ordered, unique signature indices) but that invariant was only enforced in `ChannelMultiSigProof::new`. The type `#[derive(Deserialize)]`d its private field, so the JSON mempool path (`add_tx`) bypassed `new` and could carry a proof with duplicate indices — letting a holder of a **single** accredited key satisfy a threshold-of-N config multisig by repeating its signature under one index.

**We fix it by making a non-monotonic proof unrepresentable**, at the type boundary rather than per-consumer:

- The spec requires the signature indices to be **strictly increasing**:
  ```python
  for i in range(len(proof.indexes)-1):
      assert proof.indexes[i] < proof.indexes[i+1]
  ```
- `ChannelMultiSigProof::new` now enforces exactly that — it rejects any input whose indices aren't strictly increasing (ordered **and** unique), with no silent sort/dedup (previously it normalized by sorting, which diverged from the spec's *assert*).
- `Deserialize` is implemented manually to route through `new`, so the invariant holds on **every** construction path (serde/JSON included), not just the binary one.

Consequently a duplicate-index or out-of-order proof can't be constructed at all, and both `CHANNEL_CONFIG` and `CHANNEL_WITHDRAW` can rely on the invariant without re-checking.

- Builders already produce strictly-increasing order: the wallet/sequencer construct signatures via `enumerate()` (indices 0,1,2,…), and the binary decoder feeds wire order into `new` (which now rejects a non-monotonic peer-supplied proof — spec-conformant).
- The `verify_channel_withdraw` HashSet uniqueness check is now redundant (the type guarantees it) but is left as harmless defense-in-depth; can be removed in a follow-up.
- Tests: `rejects_repeated_index`, `rejects_unsorted_indices`, `accepts_strictly_increasing_indices`, and `deserialize_rejects_non_monotonic_indices` (the serde path).
- Spec: [Mantle v1.1 — CHANNEL_CONFIG / CHANNEL_WITHDRAW Validation](https://lip.logos.co/blockchain/raw/bedrock-v1.1-mantle-specification.html#channel_config).

> Mantle/channels spec authors: please add as reviewers. Author accountable for resolution.

Yes — this aligns the implementation with the spec's strictly-increasing index assertion (the prior code normalized by sorting instead of asserting).

No.

## Checklist

* [x] 2. Description added.
* [x] 5. Implementation and Specification are in sync.
* [x] 7. New/changed logs reviewed (none added).

---

### DIFF
--- a/core/src/proofs/channel_multi_sig_proof.rs
+++ b/core/src/proofs/channel_multi_sig_proof.rs
@@ -52,31 +52,44 @@ pub enum Error {
     TooManySignatures { actual: usize, maximum: usize },
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Serialize)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+// Serde goes through `ChannelMultiSigProofRepr` via `try_from`/`into`: `Deserialize`
+// routes through `new`, so the well-formedness invariant (strictly-increasing
+// indices) is upheld on every serde path too — a non-monotonic proof is
+// unrepresentable no matter how it is constructed, and no consumer needs to
+// re-check. The repr keeps the `{ "signatures": [..] }` wire form.
+#[serde(
+    try_from = "ChannelMultiSigProofRepr",
+    into = "ChannelMultiSigProofRepr"
+)]
 pub struct ChannelMultiSigProof {
     // Invariant: signature indices are strictly increasing (hence ordered and
-    // unique), as required by the spec. Upheld by `new` AND by the custom
-    // `Deserialize` impl below, so a non-monotonic proof is unrepresentable no
-    // matter how it is constructed — every consumer can rely on the invariant
-    // without re-checking.
+    // unique), as required by the spec.
     signatures: Vec<IndexedSignature>,
 }
 
-impl<'de> Deserialize<'de> for ChannelMultiSigProof {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: serde::Deserializer<'de>,
-    {
-        // Deserialize the raw fields, then route through `new` so the well-formedness
-        // invariant holds for serde paths too (e.g. the JSON mempool path). Without
-        // this, the derived `Deserialize` would let a caller construct a proof with
-        // non-monotonic / duplicate signature indices, defeating threshold checks.
-        #[derive(Deserialize)]
-        struct Raw {
-            signatures: Vec<IndexedSignature>,
+/// Serde wire representation of [`ChannelMultiSigProof`] — a struct with a
+/// `signatures` field. Kept separate so the public type's (de)serialization
+/// is forced through `new` (via the `TryFrom`/`From` impls below) while
+/// preserving the `{ "signatures": [..] }` JSON shape.
+#[derive(Serialize, Deserialize)]
+struct ChannelMultiSigProofRepr {
+    signatures: Vec<IndexedSignature>,
+}
+
+impl TryFrom<ChannelMultiSigProofRepr> for ChannelMultiSigProof {
+    type Error = Error;
+
+    fn try_from(repr: ChannelMultiSigProofRepr) -> Result<Self, Self::Error> {
+        Self::new(repr.signatures)
+    }
+}
+
+impl From<ChannelMultiSigProof> for ChannelMultiSigProofRepr {
+    fn from(proof: ChannelMultiSigProof) -> Self {
+        Self {
+            signatures: proof.signatures,
         }
-        let Raw { signatures } = Raw::deserialize(deserializer)?;
-        Self::new(signatures).map_err(serde::de::Error::custom)
     }
 }
 
@@ -174,9 +187,10 @@ mod tests {
     }
 
     /// Regression test for #2985: a non-monotonic proof must be unrepresentable
-    /// via serde too, not just via `new`. The derived `Deserialize` would have
-    /// let the JSON mempool path bypass the well-formedness check; the custom
-    /// `Deserialize` routes through `new`, so deserialization fails.
+    /// via serde too, not just via `new`. A derived `Deserialize` would have
+    /// let the JSON mempool path bypass the well-formedness check; routing
+    /// serde through `new` (via `#[serde(try_from)]`) makes deserialization
+    /// fail.
     #[test]
     fn deserialize_rejects_non_monotonic_indices() {
         // Two distinct signatures sharing index 0 — not strictly increasing, so
@@ -194,15 +208,20 @@ mod tests {
             "a non-monotonic proof must not be deserializable"
         );
 
-        // A well-formed proof still round-trips.
+        // A well-formed proof still round-trips, and keeps the `{ "signatures": [..] }`
+        // JSON shape.
         let ok = ChannelMultiSigProof::new(vec![
             IndexedSignature::new(0, sig(1)),
             IndexedSignature::new(1, sig(2)),
         ])
         .expect("distinct indices are well-formed");
+        let serialized = serde_json::to_string(&ok).expect("serialize proof");
+        assert!(
+            serialized.starts_with("{\"signatures\":"),
+            "expected the `{{ signatures: [..] }}` shape, got {serialized}"
+        );
         let round_tripped: ChannelMultiSigProof =
-            serde_json::from_str(&serde_json::to_string(&ok).expect("serialize proof"))
-                .expect("well-formed proof round-trips");
+            serde_json::from_str(&serialized).expect("well-formed proof round-trips");
         assert_eq!(round_tripped, ok);
     }
 }
--- a/core/src/proofs/channel_multi_sig_proof.rs
+++ b/core/src/proofs/channel_multi_sig_proof.rs
@@ -1,4 +1,4 @@
-use std::{cmp::Ordering, collections::HashSet};
+use std::cmp::Ordering;
 
 use lb_key_management_system_keys::keys::Ed25519Signature;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


youngjoon-lee (approved):
Oh 🙈, thank you!

core/src/proofs/channel_multi_sig_proof.rs
```diff
@@ -46,57 +46,62 @@ impl Ord for IndexedSignature {
...
+        struct Raw {
+            signatures: Vec<IndexedSignature>,
+        }
+        let Raw { signatures } = Raw::deserialize(deserializer)?;
+        Self::new(signatures).map_err(serde::de::Error::custom)
+    }
+}
+
 impl ChannelMultiSigProof {
     pub fn new(signatures: Vec<IndexedSignature>) -> Result<Self, Error> {
```
ntn-x2, line 97: Just casually stumbling upon this PR. I think a better design, in my opinion, would be to implement `TryFrom<Vec<IndexedSignature>>`, and then for the type itself derive `Deserialize` with `#[serde(try_from = "Vec<IndexedSignature")]`. We re-use code, and we use existing Rust traits to implement what seems to be a fallible constructor.

davidrusu: Good idea! yes, lets try this

davidrusu: @ntn-x2 let me know your thoughts on the new impl

ntn-x2: I think there is no need for that machinery at all. It can be as simple as

```diff
diff --git a/core/src/mantle/encoding.rs b/core/src/mantle/encoding.rs
index 1b1020578..b3a853777 100644
--- a/core/src/mantle/encoding.rs
+++ b/core/src/mantle/encoding.rs
@@ -227,7 +227,7 @@ fn decode_channel_multi_sig_proof(input: &[u8]) -> IResult<&[u8], ChannelMultiSi
         .map(|(signature, index)| IndexedSignature::from((index, signature)))
         .collect();
 
-    ChannelMultiSigProof::new(signatures)
+    ChannelMultiSigProof::try_new(signatures)
         .map(|proof| (input, proof))
         .map_err(|_| nom::Err::Failure(Error::new(input, ErrorKind::Verify)))
 }
@@ -708,7 +708,7 @@ mod tests {
 
         // ChannelConfig creates the channel just-in-time, so no signatures are
         // required for validation — empty proof is well-formed.
-        let config_proof = ChannelMultiSigProof::new(vec![]).unwrap();
+        let config_proof = ChannelMultiSigProof::try_new(vec![]).unwrap();
 
         // Encode and decode roundtrip test (no hardcoded test vector since signatures
         // are deterministic)
@@ -922,7 +922,7 @@ mod tests {
 
         // Create a signed tx and encode it to get actual size. New channel
         // → empty proof (no signatures required for just-in-time create).
-        let config_proof = ChannelMultiSigProof::new(vec![]).unwrap();
+        let config_proof = ChannelMultiSigProof::try_new(vec![]).unwrap();
         let signed_tx =
             SignedMantleTx::new(mantle_tx, vec![OpProof::ChannelMultiSigProof(config_proof)])
                 .unwrap();
@@ -1110,7 +1110,7 @@ mod tests {
         let op_sig = signing_key.sign_payload(&txhash.as_signing_bytes());
         // Create a signed tx and encode it to get actual size. ChannelConfig
         // creates the channel here, so its proof has no signatures.
-        let config_proof = ChannelMultiSigProof::new(vec![]).unwrap();
+        let config_proof = ChannelMultiSigProof::try_new(vec![]).unwrap();
         let signed_tx = SignedMantleTx::new(
             mantle_tx,
             vec![
@@ -1223,7 +1223,7 @@ mod tests {
         // creates the channel here, so its proof has no signatures.
         let txhash = mantle_tx.hash();
         let op_ed25519_sig = signing_key1.sign_payload(&txhash.as_signing_bytes());
-        let config_proof = ChannelMultiSigProof::new(vec![]).unwrap();
+        let config_proof = ChannelMultiSigProof::try_new(vec![]).unwrap();
         let signed_tx = SignedMantleTx::new(
             mantle_tx,
             vec![
@@ -1340,7 +1340,7 @@ mod tests {
             },
         )]));
         let tx_hash = mantle_tx.hash();
-        let proof = ChannelMultiSigProof::new(vec![IndexedSignature::new(
+        let proof = ChannelMultiSigProof::try_new(vec![IndexedSignature::new(
             0,
             signing_key.sign_payload(tx_hash.as_signing_bytes().as_ref()),
         )])
diff --git a/core/src/mantle/tx.rs b/core/src/mantle/tx.rs
index 534939f6c..0627c8786 100644
--- a/core/src/mantle/tx.rs
+++ b/core/src/mantle/tx.rs
@@ -780,7 +780,7 @@ mod tests {
                 )
             })
             .collect();
-        let proof = ChannelMultiSigProof::new(signatures).unwrap();
+        let proof = ChannelMultiSigProof::try_new(signatures).unwrap();
         SignedMantleTx::new(mantle_tx, vec![OpProof::ChannelMultiSigProof(proof)]).unwrap()
     }
 
diff --git a/core/src/proofs/channel_multi_sig_proof.rs b/core/src/proofs/channel_multi_sig_proof.rs
index e9c7f898d..4830fb35e 100644
--- a/core/src/proofs/channel_multi_sig_proof.rs
+++ b/core/src/proofs/channel_multi_sig_proof.rs
@@ -53,48 +53,15 @@ pub enum Error {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
-// Serde goes through `ChannelMultiSigProofRepr` via `try_from`/`into`: `Deserialize`
-// routes through `new`, so the well-formedness invariant (strictly-increasing
-// indices) is upheld on every serde path too — a non-monotonic proof is
-// unrepresentable no matter how it is constructed, and no consumer needs to
-// re-check. The repr keeps the `{ "signatures": [..] }` wire form.
-#[serde(
-    try_from = "ChannelMultiSigProofRepr",
-    into = "ChannelMultiSigProofRepr"
-)]
+#[serde(try_from = "Vec<IndexedSignature>")]
 pub struct ChannelMultiSigProof {
     // Invariant: signature indices are strictly increasing (hence ordered and
     // unique), as required by the spec.
     signatures: Vec<IndexedSignature>,
 }
 
-/// Serde wire representation of [`ChannelMultiSigProof`] — a struct with a
-/// `signatures` field. Kept separate so the public type's (de)serialization
-/// is forced through `new` (via the `TryFrom`/`From` impls below) while
-/// preserving the `{ "signatures": [..] }` JSON shape.
-#[derive(Serialize, Deserialize)]
-struct ChannelMultiSigProofRepr {
-    signatures: Vec<IndexedSignature>,
-}
-
-impl TryFrom<ChannelMultiSigProofRepr> for ChannelMultiSigProof {
-    type Error = Error;
-
-    fn try_from(repr: ChannelMultiSigProofRepr) -> Result<Self, Self::Error> {
-        Self::new(repr.signatures)
-    }
-}
-
-impl From<ChannelMultiSigProof> for ChannelMultiSigProofRepr {
-    fn from(proof: ChannelMultiSigProof) -> Self {
-        Self {
-            signatures: proof.signatures,
-        }
-    }
-}
-
 impl ChannelMultiSigProof {
-    pub fn new(signatures: Vec<IndexedSignature>) -> Result<Self, Error> {
+    pub fn try_new(signatures: Vec<IndexedSignature>) -> Result<Self, Error> {
         Self::validate_well_formedness(&signatures)?;
         Ok(Self { signatures })
     }
@@ -136,7 +103,7 @@ impl TryFrom<Vec<IndexedSignature>> for ChannelMultiSigProof {
     type Error = Error;
 
     fn try_from(value: Vec<IndexedSignature>) -> Result<Self, Self::Error> {
-        Self::new(value)
+        Self::try_new(value)
     }
 }
 
@@ -156,7 +123,7 @@ mod tests {
             IndexedSignature::new(0, sig(2)),
         ];
         assert!(matches!(
-            ChannelMultiSigProof::new(signatures),
+            ChannelMultiSigProof::try_new(signatures),
             Err(Error::IndicesNotStrictlyIncreasing(_))
         ));
     }
@@ -170,7 +137,7 @@ mod tests {
             IndexedSignature::new(0, sig(2)),
         ];
         assert!(matches!(
-            ChannelMultiSigProof::new(signatures),
+            ChannelMultiSigProof::try_new(signatures),
             Err(Error::IndicesNotStrictlyIncreasing(_))
         ));
     }
@@ -181,7 +148,7 @@ mod tests {
             IndexedSignature::new(0, sig(1)),
             IndexedSignature::new(1, sig(2)),
         ];
-        let proof = ChannelMultiSigProof::new(signatures)
+        let proof = ChannelMultiSigProof::try_new(signatures)
             .expect("strictly-increasing indices are well-formed");
         assert_eq!(proof.signatures().len(), 2);
     }
@@ -210,7 +177,7 @@ mod tests {
 
         // A well-formed proof still round-trips, and keeps the `{ "signatures": [..] }`
         // JSON shape.
-        let ok = ChannelMultiSigProof::new(vec![
+        let ok = ChannelMultiSigProof::try_new(vec![
             IndexedSignature::new(0, sig(1)),
             IndexedSignature::new(1, sig(2)),
         ])
diff --git a/ledger/src/lib.rs b/ledger/src/lib.rs
index f8cde900f..529aac5c5 100644
--- a/ledger/src/lib.rs
+++ b/ledger/src/lib.rs
@@ -1039,7 +1039,7 @@ mod tests {
 
         let config_tx = MantleTx([Op::ChannelConfig(config_op.clone())].into());
         let config_tx_hash = config_tx.hash();
-        let config_proof = ChannelMultiSigProof::new(vec![IndexedSignature::new(
+        let config_proof = ChannelMultiSigProof::try_new(vec![IndexedSignature::new(
             0,
             signing_key.sign_payload(config_tx_hash.as_signing_bytes().as_ref()),
         )])
@@ -1205,7 +1205,7 @@ mod tests {
         };
         let withdraw_tx = MantleTx([Op::ChannelWithdraw(withdraw.clone())].into());
         let withdraw_tx_hash = withdraw_tx.hash();
-        let withdraw_proof = ChannelMultiSigProof::new(vec![IndexedSignature::new(
+        let withdraw_proof = ChannelMultiSigProof::try_new(vec![IndexedSignature::new(
             0,
             signing_key.sign_payload(withdraw_tx_hash.as_signing_bytes().as_ref()),
         )])
@@ -1293,7 +1293,7 @@ mod tests {
         let wrong_key = Ed25519Key::from_bytes(&[42; 32]);
         let withdraw_tx = MantleTx([Op::ChannelWithdraw(withdraw.clone())].into());
         let withdraw_tx_hash = withdraw_tx.hash();
-        let invalid_proof = ChannelMultiSigProof::new(vec![IndexedSignature::new(
+        let invalid_proof = ChannelMultiSigProof::try_new(vec![IndexedSignature::new(
             0,
             wrong_key.sign_payload(withdraw_tx_hash.as_signing_bytes().as_ref()),
         )])
@@ -1510,7 +1510,7 @@ mod tests {
         ];
         let config_tx = MantleTx(Ops::new_unchecked(ops.clone()));
         let config_tx_hash = config_tx.hash();
-        let config_proof = ChannelMultiSigProof::new(vec![IndexedSignature::new(
+        let config_proof = ChannelMultiSigProof::try_new(vec![IndexedSignature::new(
             0,
             sk1.sign_payload(config_tx_hash.as_signing_bytes().as_ref()),
         )])
diff --git a/tests/src/cucumber/steps/manual_zone/support.rs b/tests/src/cucumber/steps/manual_zone/support.rs
index d14801899..6c8b9432f 100644
--- a/tests/src/cucumber/steps/manual_zone/support.rs
+++ b/tests/src/cucumber/steps/manual_zone/support.rs
@@ -1330,7 +1330,7 @@ pub async fn submit_zone_withdraw(
             })?;
 
     let withdraw_proof =
-        match ChannelMultiSigProof::new(vec![IndexedSignature::new(0, withdraw_sig)]) {
+        match ChannelMultiSigProof::try_new(vec![IndexedSignature::new(0, withdraw_sig)]) {
             Ok(proof) => proof,
             Err(error) => {
                 return Err(ZoneTestError::SubmitWithdraw {
diff --git a/zone-sdk/src/sequencer/tx_builder.rs b/zone-sdk/src/sequencer/tx_builder.rs
index 33d7f7d10..ff2a91a8a 100644
--- a/zone-sdk/src/sequencer/tx_builder.rs
+++ b/zone-sdk/src/sequencer/tx_builder.rs
@@ -29,7 +29,7 @@ pub(super) fn build_atomic_withdraw_ops_proofs(
     own_sig: Ed25519Signature,
 ) -> Result<Vec<OpProof>, Error> {
     let withdraw_proof =
-        ChannelMultiSigProof::new(vec![IndexedSignature::new(own_key_index, own_sig)])
+        ChannelMultiSigProof::try_new(vec![IndexedSignature::new(own_key_index, own_sig)])
             .map_err(|e| Error::Network(format!("multi-sig proof assembly failed: {e:?}")))?;
     let mut ops_proofs = Vec::with_capacity(tx.ops().len());
     for op in tx.ops() {
@@ -126,7 +126,7 @@ pub(super) fn create_channel_config_tx(
             )
         })
         .collect();
-    let proof = ChannelMultiSigProof::new(signatures).unwrap();
+    let proof = ChannelMultiSigProof::try_new(signatures).unwrap();
 
     SignedMantleTx {
         ops_proofs: vec![OpProof::ChannelMultiSigProof(proof)],

```

ntn-x2: So removing the `ChannelMultiSigProofRepr` type altogether and renaming `new` to `try_new`.

danielSanchezQ: @ntn-x2 would you mind opening a PR with those changes on top of this branch?

ntn-x2: Sure!

ntn-x2: PR here: https://github.com/logos-blockchain/logos-blockchain/pull/3013.

thomaslavaur:
I didn't review but why are we enforcing an order (it's not how it's done in the specs). We just want to avoid duplicates. If we want to stick to the specs we shouldn't enforce any order.

davidrusu:
@thomaslavaur check the validation of the channel config / withdraw.

We enforce strictly increasing indices https://lip.logos.co/blockchain/raw/bedrock-v1.1-mantle-specification.html#channel_config

core/src/proofs/channel_multi_sig_proof.rs
davidrusu: Lets just go with this Repr machinery, its clunky but it works and keeps the json wire shape backwards compatible

thomaslavaur:
That's true for the Chanel Config I completly forgot we introduced that. But that's still not the case of the channel withdraw op. I think the spec needs an update to match the two

hansieodendaal (approved):
LGTM

PR #3010, opened 2026-06-23
fix(mempool): preserve requested transaction order

### LINKED ISSUE #2989: Block reconstruction reorders txs by hash (BTreeSet), diverging from builder order → block_root mismatch / production stall
# Block reconstruction reorders transactions by hash, diverging from the builder's order → block_root mismatch / block-production stall

## Summary

A block proposal carries `references` — an **ordered** list of tx hashes — and `header.block_root`
is the Merkle root over the transactions **in that order**. The validator, however, reconstructs
the transactions in **sorted-hash order** (a `BTreeSet`), discarding the `references` order, and
recomputes/checks `block_root` over the sorted order. Whenever the builder's transaction order is
not already hash-sorted, the recomputed root mismatches and **every validator rejects the block**.

The builder's order is *not* generally hash-sorted: the block-assembly loop reorders transactions to
respect intra-block dependencies (a tx `B` that spends an output of tx `A` is placed after `A`, even
when `hash(B) < hash(A)`). Such a block is then rejected by all validators. Because the builder keeps
re-selecting the same transactions from its mempool every slot, this can stall block production.

## Severity

High (liveness / consensus). Masked in the common case (independent transactions, where the builder's
hash-sorted mempool view already matches the validator's sorted reconstruction), but a block bundling a
dependent transaction pair in reverse-hash order is permanently unacceptable, and a poison pair sitting
in mempools can cause repeated block rejection. Also a consensus-split / interop landmine: the spec
defines reconstruction over `references` order, so any builder that doesn't emit hash-sorted order
(fee-priority, FIFO, a second client implementation) produces blocks this validator rejects.

## Spec reference

`logos-lips/docs/blockchain/raw/bedrock-v1.1-block-construction.md`:
- §Proposal Construction, step 3: `references = [mantle_txhash(tx) for tx in mempool_transactions]`
- step 4: "Compute the `header.block_root` as the root of the Merkle tree constructed from the
  `mempool_transactions` transactions used to build `references`."
- §Block Proposal Reconstruction / Validation: the proposal is reconstructed from the referenced
  transactions (in `references` order) and validated.

The block_root and reconstruction are defined over the builder's transaction order as listed in
`references`; the validator must preserve that order.

## Code reference

Builder (preserves dependency/selection order, root over that order):
- `services/chain/chain-leader/src/lib.rs:607-676` — mempool view → dependency-retry loop
  (lines 630-658, which intentionally orders a dependency before its dependent) → `txs` →
  `Block::create(parent, slot, proof, txs, key)`.
- `core/src/utils/merkle.rs:14-30` — `calculate_block_root` maps `transactions.iter()` to leaves
  in input order (no sort); the root is order-dependent.

Validator (reorders by hash, discarding `references` order):
- `services/tx-service/src/tx/service.rs:452-483` — `partition_transactions_by_availability` does
  `let keys_set: BTreeSet<_> = hashes.into_iter().collect();` then fetches in `keys_set` (sorted)
  order; `found_transactions` is therefore sorted by hash.
- `services/tx-service/src/lib.rs:48-50` — `into_found()` returns that sorted `Vec` as-is.
- `services/chain/chain-network/src/lib.rs:944-949` — `reconstructed_transactions =
  mempool_response.into_found();` then `Block::reconstruct(header, reconstructed_transactions, sig)`.
- `core/src/block/mod.rs:151-154` — `reconstruct` recomputes `calculate_content_id` over the given
  order and rejects on `header.block_root() != calculated` (`BlockRootMismatch`).

(The mempool `view` itself is hash-sorted — `services/tx-service/src/backend/pool.rs:121-127` — which
is exactly why the independent-tx common case works and masks the bug.)

## Worked failure scenario

1. User submits tx `A`, then tx `B` that spends an output (note) created by `A`. Suppose
   `hash(B) < hash(A)`.
2. A leader builds a block. Mempool view is hash-sorted: `[B, A]`. The dependency-retry loop tries
   `B` first (invalid — `A` not yet applied), applies `A`, then applies `B` in the next round.
   Builder order = `[A, B]`; `block_root = merkle([A, B])`; `references = [hash(A), hash(B)]`.
3. A validator reconstructs: `get_transactions_by_hashes([hash(A), hash(B)])` →
   `BTreeSet{hash(B), hash(A)}` → `into_found() = [B, A]` (sorted). `Block::reconstruct` computes
   `merkle([B, A]) != header.block_root` → `BlockRootMismatch` → **block rejected**.
4. Every validator runs the same path → the block is rejected network-wide. The leader re-selects
   `A,B` next slot and produces the same rejected block; the pair can stall block production while it
   remains in mempools.

## Suggested fix

Reconstruct transactions in `references` order, not sorted order. Make
`partition_transactions_by_availability` (and/or `reconstruct_block_from_proposal`) return the found
transactions in the exact order of the requested `hashes` (e.g. build a `hash -> tx` map, then map
over the original `hashes` vec), instead of collecting request hashes into a `BTreeSet`. This makes
the validator's reconstruction and execution order equal the builder's `references` order, so the
Merkle root matches and intra-block dependencies execute correctly. Add a test with a dependent tx
pair in reverse-hash order asserting the reconstructed block validates.

andrussal:

This PR fixes mempool transaction lookup by preserving the order requested by the caller.

This matters for block proposal reconstruction, where validators must rebuild the block using the same transaction order as the leader-provided transaction references. A focused tx-service test was added to cover this contract.

Closes #2989

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/services/tx-service/src/lib.rs
+++ b/services/tx-service/src/lib.rs
@@ -19,7 +19,8 @@ pub use tx::{service::TxMempoolService, settings::TxMempoolSettings};
 /// Response for `GetTransactionsByHashes` request
 #[derive(Debug, Clone)]
 pub struct TransactionsByHashesResponse<Item, Key> {
-    /// Transactions that were found in the mempool
+    /// Transactions that were found in the mempool, ordered like the requested
+    /// hashes.
     found: Vec<Item>,
     /// Hashes of transactions that were not found in the mempool
     not_found: BTreeSet<Key>,
--- a/services/tx-service/src/tx/service.rs
+++ b/services/tx-service/src/tx/service.rs
@@ -5,7 +5,7 @@ pub mod openapi {
 }
 
 use std::{
-    collections::BTreeSet,
+    collections::{BTreeMap, BTreeSet},
     fmt::{Debug, Display},
     marker::PhantomData,
     pin::Pin,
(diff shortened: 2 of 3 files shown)


hansieodendaal (approved):
LGTM

PR #3012, opened 2026-06-23
refactor(core): drop `EventPayload::Withdraw` identical to `ChannelWithdraw` op

youngjoon-lee:

(Stacked on PR #3000)

Since PR https://github.com/logos-blockchain/logos-blockchain/pull/2997 was merged, we've been emitting/persisting a `Withdraw` event per `ChannelWithdraw` operation. But, we don't actually need the event because it is 100% identical to `ChannelWithdraw` operation. The `ChannelWithdraw` operation already contains all useful information for users and the wallet. 

This PR removes the redundant `Withdraw` event and simplify the wallet tracking the `ChannelWithdraw` operation.

Yes

@youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/core/src/events/mod.rs
+++ b/core/src/events/mod.rs
@@ -93,11 +93,6 @@ pub enum EventPayload {
         voucher_nullifier: VoucherNullifier,
         utxo: Utxo,
     },
-    Withdraw {
-        channel_id: ChannelId,
-        amount: Value,
-        utxos: Vec<Utxo>,
-    },
 }
 
 impl TryFrom<Bytes> for Events {
--- a/core/src/mantle/channel.rs
+++ b/core/src/mantle/channel.rs
@@ -469,31 +469,7 @@ mod tests {
             updated.channels.channel_state(&channel_id).unwrap().balance,
             4
         );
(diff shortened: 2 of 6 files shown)


danielSanchezQ (approved):
More red than green, always welcome!

andrussal (approved):
My original thought was that we could have events for all ledger effects and then consumers can just listen events to build state they need. And this way we also don't need to mix events and block ops checking. 

But I agree that if we don't follow the idea of using only event stream for wallet state then we can just use op directly.

youngjoon-lee:
> My original thought was that we could have events for all ledger effects and then consumers can just listen events to build state they need. And this way we also don't need to mix events and block ops checking.
> 
> But I agree that if we don't follow the idea of using only event stream for wallet state then we can just use op directly.

@andrussal I also thought about the approach you mentioned. One downside is that we will store events for all operations to the storage, then. Let me first try to think about the interface which combines operations and events nicely. We anyway need to see operations in a block because we have to respect the tx/op order, as implemented in PR #3000.
If it doesn't work well, we can define/persist events for all operations.

PR #3013, opened 2026-06-23
chore: remove unnecessary serde machinery

ntn-x2:

Remove the unnecessary serde machinery when deserializing `ChannelMultiSigProof`, and remove the fallible `new` method in favor of a single `TryFrom<Vec>` implementation. This code will be updated once we migrate to bounded vectors for proofs as well.

Yes.

@ntn-x2 

N/A

No.

### DIFF
--- a/core/src/proofs/channel_multi_sig_proof.rs
+++ b/core/src/proofs/channel_multi_sig_proof.rs
@@ -53,7 +53,7 @@ pub enum Error {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
-#[serde(try_from = "Vec<IndexedSignature>")]
+#[serde(try_from = "Vec<IndexedSignature>", into = "Vec<IndexedSignature>")]
 pub struct ChannelMultiSigProof {
     // Invariant: signature indices are strictly increasing (hence ordered and
     // unique), as required by the spec.
@@ -100,6 +100,12 @@ impl TryFrom<Vec<IndexedSignature>> for ChannelMultiSigProof {
     fn try_from(value: Vec<IndexedSignature>) -> Result<Self, Self::Error> {
         Self::validate_well_formedness(&value)?;
         Ok(Self { signatures: value })
+    }
+}
+
+impl From<ChannelMultiSigProof> for Vec<IndexedSignature> {
+    fn from(value: ChannelMultiSigProof) -> Self {
+        value.signatures
     }
 }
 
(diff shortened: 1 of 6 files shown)


danielSanchezQ (approved):
Imo better. Lets wait for @davidrusu anyway.

core/src/proofs/channel_multi_sig_proof.rs
```diff
@@ -210,16 +179,12 @@ mod tests {
...
             IndexedSignature::new(0, sig(1)),
             IndexedSignature::new(1, sig(2)),
         ])
         .expect("distinct indices are well-formed");
         let serialized = serde_json::to_string(&ok).expect("serialize proof");
-        assert!(
-            serialized.starts_with("{\"signatures\":"),
-            "expected the `{{ signatures: [..] }}` shape, got {serialized}"
-        );
         let round_tripped: ChannelMultiSigProof =
```
davidrusu, line 188: This was the main reason for the machinery, I was trying to maintain the json wire structure to avoid breaking any clients.

Do we just break the json format arbitrarily?

ntn-x2: Makes sense. In that case, I guess the machinery is to remain in place indeed.

danielSanchezQ: wouldnt just be a `#[serde(rename = "signatures")]` suffice?

ntn-x2: No because one deserializes a collection, the other a map. But @davidrusu I think a simple `deserialize_with` with a small function would be simpler, instead of a struct that simply calls `TryFrom`. The function itself can call `TryFrom` instead. But anyway, because we don't want to introduce unnecessary breaking change, I am fine with either way. The least code, the better tho.

ntn-x2: Or actually even more correct would be to use an IndexSet if uniqueness and order are required, or a HasSet if order is not important (as Thomas was bringing it up). No need to use a Vector when there are better types available. But I could also do that as part of the non migration when I will get to it.

davidrusu: the design in the spec is specifically an ordered index list. The point was to push sorting client side, and just have a linear scan through the index list on the validation side.

So a typed IndexSet should not be doing the hashset dedup thing, it should just enfoce that what ever the client provided was in fact ordered by the client.

davidrusu: Lets just go with using the Repr mechanism for now? it is inelegant but I think it gets the job done while keeping our wire types backwards compatible

ntn-x2: Yup. I will think about a better way when I get to refactor proofs. I will close this PR, feel free to go ahead with the other one.

PR #3017, opened 2026-06-24
fix(core): emit SdpNoteUnlocked event when the note is actually unlocked

youngjoon-lee:
## 1. What does this PR implement?

Since PR #2855, we have been not unlocking notes as soon as SDP withdraw operations are accepted. Instead, we are unlocking the notes after 2-epoch delays, as mentioned in the [RFC](https://app.notion.com/p/nomos-tech/RFC-Remove-Concept-of-a-Session-f39261aa09df826db83e81613bb454dc?source=copy_link#357261aa09df80d79dcbeccd5487b5a5).

But, when I implemented PR #2855, I forgot to change the wallet. The wallet is currently unlocking the notes that it is tracking as soon as the withdraw operation is received from the chain service, but that is wrong now. 

In this PR,
- Defined a new event `HeaderEvent::SdpNoteUnlocked` which is emitted when a note is actually unlocked after the 2-epoch delay. I named it as `HeaderEvent` because that's for the events produced by processing a block header (not block txs). 
- Renamed the existing `EventPayload` to `TxEventPayload`. Now we can clearly distinguish tx events from header events.
- Modified the wallet to react to `HeaderEvent::SdpNoteUnlocked`.

## 2. Does the code have enough context to be clearly understood?

Yes

## 3. Who are the specification authors and who is accountable for this PR?

@youngjoon-lee 
spec: @madxor 

## 4. Is the specification accurate and complete?

Yes

## 5. Does the implementation introduce changes in the specification?

No. The RFC should be merged soon.

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [x] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [x] 2. Description added.
* [x] 3. Context and links to Specification document(s) added.
* [x] 4. Main contact(s) (developers and specification authors) added
* [x] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 6. Link PR to a specific milestone.
* [x] 7. New or changed logs were reviewed against the [logging guidelines](https://github.com/logos-blockchain/logos-blockchain/blob/master/CONTRIBUTING.md#logging-guidelines).

PR #3018, opened 2026-06-24
feat(wallet): add `locked` field in the balance API response

youngjoon-lee:

This is a follow-up PR of https://github.com/logos-blockchain/logos-blockchain/pull/3017.

This PR improves the `sdp_ops_e2e` test to check if the note has been actually unlocked in the end.

For that, we need to extend the `GET wallet/<pk>/balance` API. Previously it was returning the value of each note. This PR extend it to return the `locked: bool` for each note, so that the e2e test can check if the note has been unlocked.

No client in this repo (e.g. faucet, c-binding) is not broken. The faucet is interested in only the total balance. The c-binding simply ignores the `locked: bool` field, as before (i.e., it doesn't distinguish locked notes from other notes).

Any JSON consumers outside of this repo may be broken. Please let me know if there's any.

Yes

@youngjoon-lee 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/tests/src/tests/mantle/sdp/ops.rs
+++ b/tests/src/tests/mantle/sdp/ops.rs
@@ -230,6 +230,17 @@ async fn sdp_ops_e2e() {
             .iter()
             .any(|declaration| declaration.provider_id == provider_id)
     );
+
+    // Check that the note has been unlocked in the wallet.
+    let resp = node0
+        .get_wallet_balance(spare_note_secret_key.to_public_key(), None)
+        .await
+        .expect("balance request should succeed");
+    let note = resp
+        .notes
+        .get(&locked_note_id)
+        .expect("the previously-locked note should be tracked by the wallet");
+    assert!(!note.locked);
 }
 
 /// Test that SDP declaration is correctly restored after validator restart.
--- a/tests/src/tests/mantle/sdp/ops.rs
+++ b/tests/src/tests/mantle/sdp/ops.rs
@@ -24,7 +24,7 @@ use lb_core::{
         WithdrawMessage,
     },
 };
-use lb_key_management_system_service::keys::{Ed25519Key, Ed25519Signature, ZkKey};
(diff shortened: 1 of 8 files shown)


tests/src/tests/mantle/sdp/ops.rs
```diff
@@ -230,6 +230,17 @@ async fn sdp_ops_e2e() {
...
+    // Check that the note has been unlocked in the wallet.
+    let resp = node0
+        .get_wallet_balance(spare_note_secret_key.to_public_key(), None)
+        .await
+        .expect("balance request should succeed");
+    let note = resp
+        .notes
+        .get(&locked_note_id)
+        .expect("the previously-locked note should be tracked by the wallet");
+    assert!(!note.locked);
```
andrussal: Should we also check earlier that this note was actually locked after the declaration?

andrussal (approved):

hansieodendaal (approved):
LGTM

tests/src/tests/mantle/sdp/ops.rs
youngjoon-lee: good point! 75b28bb97

danielSanchezQ:
There is one issue with this which is that it will change the c-bindings api (even if the code itself doesnt change the content does) which means the UI may stop working (it should not tho). Lets push this for after the release if possible as we just intend to use it for tests.
cc @youngjoon-lee @hansieodendaal @andrussal

youngjoon-lee:
> There is one issue with this which is that it will change the c-bindings api (even if the code itself doesnt change the content does) which means the UI may stop working (it should not tho). Lets push this for after the release if possible as we just intend to use it for tests. cc @youngjoon-lee @hansieodendaal @andrussal

Yeah. Thank you for pointing it out. The API change in this PR is necessary ONLY for testing. So, we don’t need to merge this now. 
But we still need to merge the base PR #3017 to fix a bug. @danielSanchezQ Could you pls have a quick review for the base PR, so we can include it to the new release? It doesn’t change any API. (Acutally that’s why I split PRs)

PR #3019, opened 2026-06-24
fix(ibd): fan out IBD requests across discovered peers

### LINKED ISSUE #2217: Consider re-checking peer connections in IBD
When IBD tries to initiate downloads from peers, it tries to establish a libp2p stream expecting that a libp2p connection has been already established by the network service.

But, if no connection is not yet established by the network service (e.g., due to network issues), IBD will fail with the following error:
> 2026-02-17T13:25:32.357337Z ERROR Swarm::poll: logos_blockchain_cryptarchia_sync::libp2p::behaviour: Error while processing tip request: Peer 12D3KooWKnBcZqCWebf1LP6RtQFPHXpfe4yzmDz9U1antsXfCj9C: Stream error: failed to open stream: io error: Dial error: no addresses for peer.

We need to consider introducing a mechanism in IBD to re-check peers' connections before giving up the peers.

### LINKED ISSUE #2758: IBD: download blocks from discovered peers as well as IBD peers
https://discord.com/channels/973324189794697286/1500920951007875182/1501386546136486072

> Regarding IBD, the reason why I proposed keeping the IBD logic as it is and just expanding peers is under an assumption that most of discovered peers also provide the honest chain. Also, we have PBP (24h in the spec). But, I agree that this is a very early design and is maybe not so efficient in real world.
I understand that point of fetching tips from trusted nodes. It is like following the Polar Star. Where the node download blocks from is not very important, as long as he can catch up the Polar Star in the end. After finishing other urgent tasks, I will implement it. We don't need to change the spec because the spec already defines all request-responses needed. 

We can do this with https://github.com/logos-blockchain/logos-blockchain/issues/2759

### LINKED ISSUE #2759: Consider using orphan sync logic directly to IBD
https://discord.com/channels/973324189794697286/1500920951007875182/1501386546136486072
> Now I think we can even remove IBD. We only do orphan backfillings in the Bootstrapping mode. In that mode, the node periodically checks if it has reached the tips of trusted nodes. Once it reaches them, it stops the periodical tip check and start PBP. The mode is switched to Online once PBP passed. 
In this way, we can spread requests to any nodes who forward new blocks, while trying to catch up the (moving) tips of trusted nodes.
A benefit is that we can make IBD implementation much thinner. Since we have chainsync network protocol used for both IBD and orphans, the IBD logic can do only the periodic tip check.
A downside is that the catch-up may be slow if the network doesn't add blocks to the honest chain for a long time. But, I think it's unlikely to happen.

youngjoon-lee:

Closes #2758 and #2759

### Issues

In devnets, we've found that IBD often fails because all new nodes in the network send IBD requests to the same 4 peers (configured as `ibd.peers` in YAML). It's because those 4 nodes are limiting the max concurrent chainsync requests to avoid DoS. 

In this reason, we've disabled IBD in the devnet. But, it's better to enable IBD in the following reason:
- We want to catch up trusted peers before starting PBP. IBD and PBP are for protecting us from long-range attacks. We may be able to rely on only a long PBP (without IBD), but I think it's not a good idea because PBP may end before the node downloads the meaningful number of blocks (e.g., network errors, etc.). I think it's better to explicitly catch up the trusted peers before switching to online mode.
  - I honestly think that 24h PBP in the spec is too long. If IBD has completed against the trusted peers, we don’t really need to wait for the long PBP. But, this is out of scope for this PR. I’m just saying this to explain why I think IBD is necessary. 

### Solution

A basic idea to make IBD scalable is: Use `ibd.peers` only for getting the trusted tips, and download ancestors from "any" peers (connected or discovered). In other words, we will use `ibd.peers` as the North Stars, and fan out block downloads across a larger set of peers.

To achieve that, I decided to reuse the `OrphanDownloader` which already has the following features:
- Queuing multiple target tips (i.e., orphan blocks)
- Fanning out block downloads across all connected/discovered peers (with a limit)

Thanks to `OrphanDownloader`, I eliminated most of IBD code. Now, the `ibd.rs` is a simple wrapper of `OrphanDownloader` that does the following:
- Fetch tips from the trusted peers and enqueue them to the downloader.
- Drain the downloader until there's no block left in the queue.
- Repeat these until the local chain catches up every trusted peers' tip.

This is actually the architecture that we should have had earlier. But we postponed it because IBD was implemented before the `OrphanDownloader` took its current shape.

### Performance Considerations

This PR implements the following considerations:

#### Avoid `QueueIsFull` error

IBD creates `OrphanDownloader` by setting the queue capacity to the number of `ibd.peers` because we always enqueue at most `ibd.peers.len()` number of tips into the downloader. Also, we don't fetch/enqueue new tips until we drain/process the queue completely. Of course, when IBD is done, we drop the `OrphanDownloader` and create the new one with a larger capacity for the main event loop of chain-network service.

#### Don't overload `ibd.peers`

People tend to configure `ibd.peers` with the most well-known nodes in the network. It means that already every node in the network will send tip requests to those well-known nodes. 

To not overload well-known nodes, we introduce exponential backoff for tip requests. This is also good to not give up IBD too early. If it is failed to fetch tips from all `ibd.peers` in the end after retries+backoff, IBD is stopped and the chain-network service is terminated (as before).

#### Peer discovery delay

IBD is started almost as soon as a node starts. But, peer discovery is probably not so fast. It means that the 1st block download request may be send to only the `ibd.peers` because there is no discovered peer yet. If they have room to process requests, it's fine. If not, they will reject requests (`max_inbound_requests`). Then, the node will sleep for a while before starting a next round (for fetching new tips and downloading blocks). 

While testing this PR with devnet, I saw that 1 or 2 peers are discovered after 2~3 seconds since the node starts. So, the block download requests are successfully fanned out across discovered peers in the end.

I also [considered/implemented](https://github.com/logos-blockchain/logos-blockchain/pull/3019/commits/04bcef4ea5fc41c24bfbda0727db03428f310672) having a warm-up phase which waits at most 10s until at least 1 peer is discovered before starting IBD. But, I ended up [removing](https://github.com/logos-blockchain/logos-blockchain/pull/3019/commits/a6e724d1d6d6b9f4da62b544fb2447bc60879ebf) that behavior because I think the delay between rounds that I mentioned above is enough and even better. Also, I don't want early-joining nodes to wait 10s for the warm-up even though there might be not many nodes to discover, except `ibd.peers`.

#### Keeping v.s. Removing the IBD implementation

As I mentioned above, I think we need IBD. But, I considered if we can completely remove `ibd.rs` and rely on the orphan downloader and tip-poll watchdog that was recently added. 

However, I ended up keeping `ibd.rs`. The tip-fetch logic in the `ibd.rs` might look similar as the tip-poll watchdog. But, that is actually much simpler and has different requirements. 

One example is that, during IBD, we want to explicitly follow a set of trusted peers. In contrast, the watchdog samples random peers. Of course, we could change the specification and adjust the goals of IBD accordingly, but I think the current IBD spec is already reasonable and intuitive, so I don’t see a strong need to change it.

We might be able to bake all of its requirements into the tip-poll watchdog, but I didn't want to make the watchdog complicated because of IBD. I think it is better to clearly separate tools for each purpose.

#### No concurrent download 
The previous IBD downloaded blocks from all `ibd.peers` concurrently. But, the new implementation doesn't do it because the `OrphanDownloader` doesn't support concurrent downloads.

I think that's fine because most of `ibd.peers` may share the long common chain (deeper than LIB) if they were not forked permanently. Also, it doesn't make sense to choose `ibd.peers` that were permanently forked.

#### No eager target re-adjustment
In the previous IBD, we re-fetched the peer's tip every time we download 1,000 blocks, because we wanted to react to the peer's chain reorgs as soon as possible. But, we don't have that behavior now because the `OrphanDownloader` doesn't support it.

Of course, we can implement the behavior on top of `OrphanDownloader`, but I don't think it's worth it. Most of `ibd.peers` may share the long common chain. Chain reorgs can happen within the short `k`-block range (e.g., 2K in mainnet). So, considering the 1K batch size of download, we don't really need a complex logic to re-adjust the target tips eagerly. We're already re-fetching tips after draining all downloads. That's enough.

###  Next PRs

- Split the existing `chainsync.max_inbound_requests` into `max_inbound_tip_requests` and `max_inbound_download_requests`, so that we can limit those two kinds of requests separately. Even, we can configure the initial 4 nodes to deny download requests.

Yes

@youngjoon-lee 
spec: @youngjoon-lee 

Yes. 

No because we use the same requests/responses defined in the spec.

## Checklist

> [!WARNING]  

* [x] 2. Description added.


services/chain/chain-network/src/bootstrap/ibd.rs
```diff
@@ -103,318 +109,149 @@ where
...
-                Ok(Some(download)) => {
-                    self.start_download(download, &mut downloads);
+        downloader: &mut OrphanBlocksDownloader<NetAdapter, RuntimeServiceId>,
+    ) {
+        // Use `timeout` because `downloader.next()` can stall forever if a
+        // download fails and leaves the queue empty: the downloader returns
+        // `Pending` with no internal wake source. The timeout lets us re-check
+        // `should_poll` and exit cleanly in that case.
+        // TODO: improve `OrphanBlocksDownloader` to simplify this.
+        while downloader.should_poll() {
```
danielSanchezQ, line 201: This is really weird tbh. We should either have a stream and consume this or something else, a sync call for polling makes no sense?

danielSanchezQ (approved):
Image
🚀

services/chain/chain-network/src/bootstrap/ibd.rs
```diff
@@ -131,308 +135,208 @@ where
...
-        debug!(
-            ?target, local_tip = ?initial_cryptarchia_info.tip, local_tip_height = initial_cryptarchia_info.height,
-            local_lib = ?initial_cryptarchia_info.lib, ?peer,
-            "requesting blocks from peer",
+        config: IbdConfig<NetAdapter::PeerId>,
+        orphan_config: &OrphanConfig,
+    ) -> Result<(), Error> {
+        let mut downloader = OrphanBlocksDownloader::new(
+            self.network.clone(),
+            config.peers.len().try_into().unwrap(),
```
danielSanchezQ: Lets use `expect` instead of `unwrap`

services/chain/chain-network/src/bootstrap/ibd.rs
danielSanchezQ: Coming back again, why isn't downloader a stream?

services/chain/chain-network/src/bootstrap/ibd.rs
```diff
@@ -131,308 +135,208 @@ where
...
-
-        self.try_initiate_download(*download.peer(), download.last(), downloads)
-            .await
+        let mut unsynced = HashSet::new();
+        for tip in tips {
+            if !self.block_processor.has_processed_block(tip).await? {
+                unsynced.insert(tip);
+            }
+        }
+        Ok(unsynced)
```
danielSanchezQ: This can be constructed with a stream from tips and collect all of them in an async way, as a hashset is not ordered we do not need to do it sequentially. It can be a futures unordered depending the query to the processor.

services/chain/chain-network/src/bootstrap/ibd.rs
```diff
@@ -131,308 +135,208 @@ where
...
+async fn fetch_tips_with_retry<NetAdapter, RuntimeServiceId>(
+    network: &NetAdapter,
+    config: &IbdConfig<NetAdapter::PeerId>,
+) -> Result<HashSet<HeaderId>, AllPeersFailed>
+where
+    NetAdapter: NetworkAdapter<RuntimeServiceId> + Sync,
+    NetAdapter::PeerId: Copy + Clone + Eq + Hash + Debug + Send + Sync,
+    RuntimeServiceId: Sync,
+{
+    (|| fetch_tips(network, config.peers.iter().copied()))
```
danielSanchezQ, line 264: Why this closure?

danielSanchezQ (commented):
Its much simpler. I left some questions but otherwise I think is good to go.

services/chain/chain-network/src/bootstrap/ibd.rs
youngjoon-lee: Actually the downloader is a stream, so that we can use it in the main event loop of chain-network service. The stream yields blocks as soon as they are downloaded, regardless of which orphan request they are corresponded to. If there's no block to download, the stream yields nothing, instead of being closed, because it expected that more orphan requests would be added/requested in the future.

I think this design is fine for the orphan block handling in the main event loop, but it might not be optimal for IBD in this PR. I want to improve the downloader in a next PR. I think, we can probably improve or replace the `downloader.should_poll` function somehow, but I didn't want to touch it in this PR. I need to cross-check it with Andrus.

services/chain/chain-network/src/bootstrap/ibd.rs
youngjoon-lee: c1202230c4149132e0867a50120a5082de3bc05e done.

services/chain/chain-network/src/bootstrap/ibd.rs
youngjoon-lee: That is actually required by [`backon::Retryable`](https://docs.rs/backon/latest/backon/trait.Retryable.html), and we have a similar code in our [libp2p](https://github.com/logos-blockchain/logos-blockchain/blob/aae224cc21b2105f91554e59354024a750eac58a/libp2p/src/behaviour/nat/address_mapper/protocols/pcp_core/connection.rs#L57) crate. I left a comment to explain why we need the closure: 53f379fe2b92eecdc5b395a0ca547bf0591e3262

services/chain/chain-network/src/bootstrap/ibd.rs
youngjoon-lee: I just added aae224cc21b2105f91554e59354024a750eac58a to parallelize `has_processed_block` calls.

I also tried to introduce the full pipelining that covers both fetching tips and checking processed blocks, but it was not trivial and not easy to read because of error cases and retry. Let me think how to improve it.

youngjoon-lee: I gave another try, but now I think the current shape is good enough and easy to read. 

I tried to pipeline the entire flow from fetching tips to draining downloads, to prevent one slow peer from blocking IBD to proceed to next tips of other fast peers. But it was not easy to read, and the benefit is not big. 

In the current shape, we fetch tips from all peers first and start draining the downloader queue. But, I think it's fine because fetching tips from peers is lightweight. We have the retry mechanism for fetching tips, but the retry is triggered only when it's failed to fetch tips from "all" peers. One slow peer can't block IBD too long because we have the request timeout.

services/chain/chain-network/src/bootstrap/ibd.rs
danielSanchezQ: Ah i see, I was not familiar with Retryable. It makes sense the closure now. Thanks!

danielSanchezQ (approved):

PR #3021, opened 2026-06-24
feat(zone-sdk): fire channel updates on conflict and adjust policies

pradovic:

This PR expands the orphaned field in the BlocksProcessed update and emits channel updates only on conflict (when channel view changes, not on every L1 branch fork).

The update is channel-driven, not L1-driven: it is emitted only when the channel lineage changes, either because the channel view extends or because it switches to a different channel branch due to a conflict. An L1 branch change that leaves the channel lineage unchanged emits no update.

On conflict, orphaned now includes the full delta back to the LCA, plus the invalidated pending set. Previously, it only included the invalidated pending set. This gives consumers enough information to apply a more accurate policy when reconciling their local state.

This PR also extends the current conflict handling policies to support deeper forks.

A possible future optimization is to track the pending set per channel branch instead of only for the canonical branch. That would reduce the conflict candidate set and avoid some mempool churn when channel views alternate frequently. This is not required for correctness, and would increase the amount of state we need to maintain.

Yes

@pradovic 

No

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


pradovic:
@davidrusu Note here that `BlocksProcessed` event will fire on any view change (branch change) and would cover 3 cases:
1. Branch change without conflict would emit orphaned and adopted delta diff so the user can adjust state with current view -> this is needed so when the conflict arises the lca diff is actually correct in respect to the users state.
2. Extension with conflict -> orphaned would then basically just have invalidated pending set if any.
3. Branch change with conflict -> orphaned = {block delta + invalidated set} and adopted same as is

zone-sdk/src/sequencer/types.rs
```diff
@@ -261,26 +261,32 @@ pub enum TxSource {
...
-///    publishes) to local state, revert `orphaned` (yours that can no longer
-///    land). Both being empty is a no-op.
-/// 3. For each entry in `orphaned`, decide whether to republish (with a fresh
-///    parent — SDK handles parent selection).
+/// 2. On [`Event::BlocksProcessed`]: rebuild state by reverting every
+///    `orphaned` entry you had applied and applying every `adopted` entry
+///    (dedup `adopted` by `this_msg` against your outbox so you don't
+///    double-apply your own publishes). Entries in `orphaned` that you never
+///    applied (our never-landed pending) are a no-op for state — key your
+///    rollback by `this_msg` so they simply don't match.
```
davidrusu: > (dedup `adopted` by `this_msg` against your outbox so you don't double-apply your own publishes)

This should not be a possible case. The users state was at the tip of the orphaned vec. The user could not have applied any of the adopted messages onto that state.

```suggestion
/// 2. On [`Event::BlocksProcessed`]: rebuild state by reverting every
///    `orphaned` entry you had applied and applying every `adopted` entry.
```

zone-sdk/src/sequencer/types.rs
```diff
@@ -261,26 +261,32 @@ pub enum TxSource {
...
+/// 3. Republish the orphans you still want, **skipping any whose payload is
+///    already in `adopted`** (back on the channel under a different
+///    `this_msg`). Republishing an orphan you didn't submit is harmless — the
+///    SDK re-chains it onto the current tip — so no ownership filter is needed.
```
davidrusu: This is too prescriptive and still stuck in the 1 zone tx == 1 inscription model, can we reduce this to:
```suggestion
/// 3. Process the orphans so that no useful information is lost. e.g. If inscriptions are your Zone blocks, extract all the Zone transactions from the block and re-insert it into the Zone mempool.
```

zone-sdk/src/sequencer/types.rs
```diff
@@ -261,26 +261,32 @@ pub enum TxSource {
...
 #[derive(Debug, Clone)]
 pub struct ChannelUpdate {
-    /// Our pending whose original signed tx is permanently invalid because
-    /// a competing inscription claimed the parent slot (or because the
-    /// parent is now off the canonical chain transitively). These need a
-    /// user decision — re-creation requires your signing key.
+    /// Inscriptions removed from the channel — the block delta plus our pending
+    /// that can no longer land (including pending that never landed, so it
+    /// appears nowhere else). Revert from state and treat as republish
+    /// candidates.
```
davidrusu:
```suggestion
    /// Inscriptions removed from the channel — the safe and pending inscriptions
    /// that can no longer finalize due to a conflicting channel inscription (parent-double-spend).
    /// Returned to user for processing.
```

deployment/tui-zone/src/state.rs
davidrusu: Why do we have a `published` and `adopted` msgs? Shouldn't it just be `pending` and `finalized`?

We are just trying to model the linear channel message chain here.

```suggestion
pending: Vec<Msg>,
finalized: Vec<Msg>,
```

deployment/tui-zone/src/state.rs
```diff
@@ -22,8 +22,11 @@ use crate::message::Msg;
 /// resuming from a persisted state and re-receiving backfill is harmless.
 pub trait ZoneState: Send {
     fn on_adopted(&mut self, adopted: &[InscriptionInfo]);
-    /// Remove our orphaned entry from `published`. Caller is expected to
-    /// auto-republish via `sequencer.handle().publish`.
+    /// Drop an orphaned inscription: remove it from our outbox (`published`,
+    /// so the caller can auto-republish via `sequencer.handle().publish`) and
+    /// from the channel view (`adopted`, since it fell off canonical). An
+    /// orphan we never applied to `adopted` (our never-landed pending) is
+    /// simply absent there, so that removal is a no-op.
```
davidrusu: We shouldn't be tracking the outbox in the TUI zone state, the ZoneSDK manages the outbox (pending) for us.

TUI Zone State can keep the channel chain as two vecs to render those states: `[finalized msgs] [pending msgs]`, but it shouldn't do anything else besides that.

zone-sdk/DECENTRALIZED_SEQUENCING.md
```diff
@@ -265,7 +265,7 @@ if let Event::BlocksProcessed { channel_update, .. } = event {
 }
 ```
 
-The integration tests provide a reference policy that runs this re-publish loop automatically — see `OrphanRepublishPolicy` in `tests/src/cucumber/steps/manual_zone/support.rs`. Wiring such a policy into the drive loop is the recommended pattern for any sequencer running in a competing-write environment.
+The integration tests provide reference policies that run this re-publish loop automatically, in `tests/src/cucumber/steps/manual_zone/support.rs`: `OrphanRepublishPolicy` (simple republish), `SortedConflictPolicy` (republish only when it preserves the channel's sorted order, otherwise discard), and `BalanceAwarePolicy` (republish only when the account balance still allows it). Wiring one into the drive loop is the recommended pattern for any sequencer running in a competing-write environment.
```
hansieodendaal: We should add `RepublishLineagePolicy` to the list

hansieodendaal (approved):
I think this is better behaviour altogether from the Zone SDK side, and the example policies in the cucumber test code illustrate this nicely. I welcome the addition of tracing intent in those policies, as I always thought it was missing.

One small correctness/documentation issue remains, but otherwise this looks fine.

From @davidrusu:
- Zone SDK docs should be sanitized as suggested.
- I propose we open a follow-up PR after this one is merged to add different policy models to the TUI Sequencer code, selectable at startup. That would demonstrate lifecycle policies not only in the cucumber code, but also in the stand-alone TUI Sequencer application.

LGTM!

hansieodendaal (approved):
LGTM (again!)

davidrusu (approved):

PR #3016, opened 2026-06-23
test: improve devnet manual cucumber scenarios

hansieodendaal:
## 1. What does this PR implement?

This PR refactors the cucumber transaction harness to prepare, sign, and submit wallet transactions concurrently, for continuous round-robin and next-wallet workloads, in order to better stress the mempool. For these scenarios, the wallet transaction pipeline was split into reserve/finalize phases so UTXO selection stays serialized while the expensive proof/signing work runs in parallel, then submissions are pushed concurrently as well. Chain-source backfill was also improved by allowing a fallback peer for wallet feed recovery, which makes feed catch-up more resilient when the chosen node lags.

## 2. Does the code have enough context to be clearly understood?

Yes

## 3. Who are the specification authors and who is accountable for this PR?

@hansieodendaal @andrussal 

## 4. Is the specification accurate and complete?

Yes

## 5. Does the implementation introduce changes in the specification?

No

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [X] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [X] 2. Description added.
* [X] 3. Context and links to Specification document(s) added.
* [X] 4. Main contact(s) (developers and specification authors) added
* [X] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [X] 6. Link PR to a specific milestone.
* [X] 7. New or changed logs were reviewed against the [logging guidelines](https://github.com/logos-blockchain/logos-blockchain/blob/master/CONTRIBUTING.md#logging-guidelines).

Issue #2569, opened 2026-04-17
Unable to restart the node from 'Online' mode


grigor-aramyan:
After successful bootstrap and run for couple of days the node (and laptop) shut down because power supply went off and all tries to restart it after that were resulting to the same error, "couldn't retrieve block parent for xxx from storage during recovery".

Removing 'state' folder and starting the node as usual resolved the issue. Attaching archived 'state' that was causing the issue in case there's a need to investigate.

I'm on Ubuntu 24.04 LTS, running logos node v0.1.2

[state.tar.gz](https://github.com/user-attachments/files/26829392/state.tar.gz)

xAlisher:
Greetings from a node running at an orange grove in Valencia 🍊

Hit the exact same issue — \`Could not retrieve block parent for [tip] from storage during recovery\` — after a kernel panic (dirty shutdown mid-write). Node crash-looped every 10 seconds indefinitely.

**Root cause (what we think is happening):** The tip pointer in the recovery snapshot advances to a new block, but the parent block data isn't fully written to RocksDB before the crash. On restart, the chain service tries to walk \`(lib, tip]\` from storage, finds the tip block but not its parent, and panics instead of recovering gracefully.

**Workaround — no full resync needed:**

The tip can be rolled back to the LIB directly in the recovery snapshot:

\`\`\`bash
# Stop the node first
systemctl --user stop logos-node

# Backup
cp state/live-v0.1.2/recovery/consensus/chain_service.json \
   state/live-v0.1.2/recovery/consensus/chain_service.json.bak

# Set tip = lib
python3 -c "
import json
path = 'state/live-v0.1.2/recovery/consensus/chain_service.json'
with open(path) as f:
    d = json.load(f)
print('tip before:', d['tip'])
print('lib:       ', d['lib'])
d['tip'] = d['lib']
with open(path, 'w') as f:
    json.dump(d, f)
print('tip after: ', d['tip'])
"

# Restart
systemctl --user start logos-node
\`\`\`

The node will log \`0 blocks recovered. finishing initialization tip_height=N lib_height=N\` and start IBD from the LIB forward — losing only the few unfinalized blocks between LIB and the broken tip, rather than resyncing from genesis.

**Suggested fix upstream:** in the recovery path at \`services/chain/chain-service/src/lib.rs:1089\`, catch the missing-parent case and fall back to LIB automatically rather than panicking. This is effectively what the workaround does manually.

---

[...]

PR #3024, opened 2026-06-24
fix(chain): fall back from missing recovered tip

andrussal:

This PR is related to the recovery inconsistency in [#2569](https://github.com/logos-blockchain/logos-blockchain/issues/2569).

In that case, recovery state can point further ahead than what RocksDB can reconstruct.

This PR changes recovery to first prove that the whole `(LIB, tip]` branch can be loaded from storage. If the recovered tip path or block body is missing, chain-service falls back to the recovered LIB and persists a rebuilt recovery state from that safe point.

This makes startup robust against the broken state from [#2569](https://github.com/logos-blockchain/logos-blockchain/issues/2569), together with the storage batch fix in [#3016](https://github.com/logos-blockchain/logos-blockchain/pull/3016), but it does not remove the underlying split between file-based Overwatch recovery state and RocksDB chain data. Longer term, it may make sense to store Overwatch recovery state in RocksDB too, but that likely needs changes on the Overwatch side.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/services/chain/chain-service/src/lib.rs
+++ b/services/chain/chain-service/src/lib.rs
@@ -105,6 +105,17 @@ pub enum Error {
     ParentIdNotFound(HeaderId),
 }
 
+struct InitializedCryptarchia {
+    cryptarchia: Cryptarchia,
+    pruned_blocks: PrunedBlocks<HeaderId>,
+    fell_back_to_lib: bool,
+}
+
+struct RecoveryBlocks<Tx> {
+    blocks: Vec<Block<Tx>>,
+    fell_back_to_lib: bool,
+}
+
 #[derive(Derivative)]
 #[derivative(Debug)]
 pub enum ConsensusMsg<Tx> {
@@ -595,7 +606,11 @@ where
 
         let (mut current_slot, mut slot_timer) = Self::get_slot_timer(&relays).await?;
 
-        let (mut cryptarchia, pruned_blocks) = self
+        let InitializedCryptarchia {
+            mut cryptarchia,
+            pruned_blocks,
+            fell_back_to_lib,
+        } = self
             .initialize_cryptarchia(
                 &bootstrap_config,
                 ledger_config.clone(),
@@ -612,6 +627,14 @@ where
             relays.storage_adapter(),
         )
         .await;
+
+        if fell_back_to_lib {
+            Self::update_state(
+                &cryptarchia,
+                storage_blocks_to_remove.clone(),
+                &self.service_resources_handle.state_updater,
+            );
+        }
 
         let sync_blocks_provider: BlockProvider<_, _> =
             BlockProvider::new(relays.storage_adapter().storage_relay.clone());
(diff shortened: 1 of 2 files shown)


bacv (approved):

danielSanchezQ (approved):

services/chain/chain-service/src/lib.rs
```diff
@@ -1227,6 +1250,63 @@ where
...
+        tip: HeaderId,
+        lib: HeaderId,
+        storage: StorageAdapter<Storage, Tx, RuntimeServiceId>,
+    ) -> Result<Vec<Block<Tx>>, Error> {
+        let ids = Self::load_block_ids_from_storage(tip, lib, storage.clone())
+            .try_collect::<Vec<_>>()
+            .await?;
+
+        let mut blocks = Vec::new();
+        for id in ids.into_iter().rev().skip(1) {
```
youngjoon-lee, line 1266: Please add a comment about why we do `rev()` and `skip(1)`. Maybe, you can copy the removed comments below.

services/chain/chain-service/src/lib.rs
```diff
@@ -1227,6 +1250,63 @@ where
         ))
     }
 
+    async fn load_recovery_blocks_from_storage(
+        tip: HeaderId,
+        lib: HeaderId,
+        storage: StorageAdapter<Storage, Tx, RuntimeServiceId>,
+    ) -> Result<Vec<Block<Tx>>, Error> {
```
youngjoon-lee, line 1257: Previously we avoided collecting `Block`s into `Vec`. We loaded each `Block` and applied it immediately to the cryptarchia before loading next `Block`s. 

But, I understand that it's not easy to use streaming in this new flow. Let's improve this later if it's urgent to fix this issue.

services/chain/chain-service/src/lib.rs
```diff
@@ -1227,6 +1250,63 @@ where
...
+        tip: HeaderId,
+        lib: HeaderId,
+        storage: StorageAdapter<Storage, Tx, RuntimeServiceId>,
+    ) -> RecoveryBlocks<Tx> {
+        if tip == lib {
+            return RecoveryBlocks {
+                blocks: Vec::new(),
+                fell_back_to_lib: false,
+            };
+        }
```
youngjoon-lee, line 1288: It'd be great if we add a comment, saying "Since we already have LIB in Cryptarchia, we don't need to load any block". Someone may wonder why we return early here.

youngjoon-lee (approved):

youngjoon-lee (changes requested):
Hm, wait. It seems that we were persisting the new recovery state only after the `process_block` succeeds (i.e., only after updating Cryptarchia and persisting blocks successfully). If so, I don't clearly understand why some nodes fail to load blocks in the `initialize_cryptarchia`. Could you please explain what this PR is trying to fix, in more detail?

If I understand this PR correctly, this PR is trying to handle an inconsistent situation which the recovery file contains the "latest" state, but the storage doesn't have those latest blocks, right? I don't clearly understand how the situation could happen. Is it because RockDB doesn't persist values immediately (fs-sync, etc.)?

youngjoon-lee:
> Hm, wait. It seems that we were persisting the new recovery state only after the `process_block` succeeds (i.e., only after updating Cryptarchia and persisting blocks successfully). If so, I don't clearly understand why some nodes fail to load blocks in the `initialize_cryptarchia`. Could you please explain what this PR is trying to fix, in more detail?
> 
> If I understand this PR correctly, this PR is trying to handle an inconsistent situation which the recovery file contains the "latest" state, but the storage doesn't have those latest blocks, right? I don't clearly understand how the situation could happen. Is it because RockDB doesn't persist values immediately (fs-sync, etc.)?

Hm, okay. I read through the issue https://github.com/logos-blockchain/logos-blockchain/issues/2569, and a community member mentioned his guess about fsync. I agree. The `Put` operation is returns once the value is appended to the memtable and is written to the WAL via `write()` syscall. But the value lands in the kernel page cache, not on disk, if the sync mode is off.

It'd be great if we introduce a way to always guarantee the durability order (rocksdb -> recovery file). But, I agree with having this defensive recovery.

youngjoon-lee (approved):

andrussal:
> Hm, wait. It seems that we were persisting the new recovery state only after the `process_block` succeeds (i.e., only after updating Cryptarchia and persisting blocks successfully). If so, I don't clearly understand why some nodes fail to load blocks in the `initialize_cryptarchia`. Could you please explain what this PR is trying to fix, in more detail?
> 
> If I understand this PR correctly, this PR is trying to handle an inconsistent situation which the recovery file contains the "latest" state, but the storage doesn't have those latest blocks, right? I don't clearly understand how the situation could happen. Is it because RockDB doesn't persist values immediately (fs-sync, etc.)?

Here I think yes the issue is that we need to keep database and file system in sync but we don’t have an atomic way of saving both together. I believe that in this relatively short time window file system write may succeed but database writes may still be in memory when node crash .  

We could also use sync writes so every time memory is flushed to disk when we write but generally this is not recommended as it is much slower. Although it might be fine given we produce blocks quite slowly. But even with sync we still have two independent storage systems and some inconsistencies can still happen.

I think best solution would be to use database backend also for recovery and then that should significantly reduce chance for corruption. But currently I think it is not quite straightforward to implement it.

youngjoon-lee:
> I think best solution would be to use database backend also for recovery and then that should significantly reduce chance for corruption. But currently I think it is not quite straightforward to implement it.

I agree with this

danielSanchezQ:
> I think best solution would be to use database backend also for recovery and then that should significantly reduce chance for corruption. But currently I think it is not quite straightforward to implement it.

Yes, we should implement this in db better than in file (plain json) and its something we should do next iteration if you all agree. Its a breaking change but we can write a migration tool to move from one recovery to other.

services/chain/chain-service/src/lib.rs
andrussal: > Previously we avoided collecting `Block`s into `Vec`. We loaded each `Block` and applied it immediately to the cryptarchia before loading next `Block`s.
> 
> But, I understand that it's not easy to use streaming in this new flow. Let's improve this later if it's urgent to fix this issue.

Yeah, agreed, this uses more memory than the old replay path. I think it is acceptable for this defensive fix, but if we move recovery state into DB later, we should revisit this and restore a cleaner replay path.

PR #3028, opened 2026-06-25
feat(ibd): DRAFT: retry tip requests

youngjoon-lee:

Stacked on PR #3019 

This PR adds a retry mechanism for requesting tips to the trusted peers.

IBD basically fails if none of the trusted peers fail to return their tip (e.g., due to `max_inbound_requests`). This failure can happen if all nodes in the network send tip requests to the same trusted nodes. 

To avoid giving up on IBD too early, we are adding a retry mechanism.

Tip requests themselves are very lightweight. However, the current chainsync implementation uses the common `max_inbound_requests` limit for both tip requests and download requests. As a result, if trusted nodes are busy serving many download requests, they may reject incoming tip requests as well, causing IBD failures for other nodes. In a follow-up PR, I will introduce separate limits for tip requests and download requests.

Yes

@youngjoon-lee 

Yes

No. This is an implementation detail.

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/nodes/node/binary/src/config/cryptarchia/serde/network.rs
+++ b/nodes/node/binary/src/config/cryptarchia/serde/network.rs
@@ -1,4 +1,7 @@
-use core::num::{NonZeroU64, NonZeroUsize};
+use core::{
+    num::{NonZeroU64, NonZeroUsize},
+    time::Duration,
+};
 use std::collections::HashSet;
 
 use libp2p::PeerId;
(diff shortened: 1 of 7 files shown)


nodes/node/binary/src/config/cryptarchia/serde/network.rs
```diff
@@ -25,6 +28,26 @@ pub struct BootstrapConfig {
...
+    pub attempts: NonZeroUsize,
+    /// Fixed delay between attempts.
+    pub delay_between_attempts: Duration,
+}
+
+impl Default for TipsFetchConfig {
+    fn default() -> Self {
+        Self {
+            attempts: NonZeroUsize::new(3).unwrap(),
+            delay_between_attempts: Duration::from_secs(1),
```
ntn-x2, line 48: Isn't this a small interval? Would it make sense to have a backoff mechanism, like after 1, 2, and 4 seconds? Or maybe 2, 4, and 8?

youngjoon-lee: yeah I will add the exponential backoff.

PR #3032, opened 2026-06-26
fix(chain-network): negative-cache dead-fork blocks to stop orphan storms

davidrusu:
## Why

On **2026-06-24** a stale fork *below the immutable boundary (LIB)* triggered a ~20 GB/h log storm that took down multiple nodes: the fork's blocks were re-gossiped endlessly, each one reconstructed, re-applied (failing `ParentMissing`), and re-enqueued for orphan download (~51×/hour per orphan at peak).

This is a small, self-contained fix entirely in **chain-network** — an in-memory **negative cache** of block ids known to be on dead forks. No storage, pruning, or consensus changes. Targets the **`0.1.2`** commit (base `0.1.2-base`).

## What

New bounded (FIFO, 100k-entry) `NegativeCache` in `services/chain/chain-network/src/negative_cache.rs`. It is consulted as blocks arrive:

**On each gossiped proposal** (`handle_incoming_proposal`), before any reconstruct/apply/orphan work:
1. **id in cache** → drop it.
2. **parent in cache** → the block extends a dead fork, so it is dead too: cache its id (so its own descendants are dropped by rule 1) and drop it.

**When processing the orphan download queue:**
3. if the **first block of a download diverges below LIB** (`slot < lib_slot`), the orphan is on an unwinnable fork: cache the orphan tip and cancel the stream.

Rule 3 seeds the cache; rules 1 & 2 then absorb the dead fork's re-gossip and its descendants, eliminating the reprocessing amplification.

### Supporting change

`CryptarchiaInfo` gains a `lib_slot` field (the running node already exposes this) so the `ParentMissing` error carries the LIB slot into the orphan downloader (threaded through `enqueue_orphan` → `OrphanInfo`). That nudged `Error` past clippy's `result_large_err` threshold, so `CryptarchiaInfo` is now **boxed** inside the two `ParentMissing` variants — the lint's recommended fix, which also shrinks the error.

## Tests

- 3 new unit tests for `NegativeCache` (insert/contains, FIFO eviction, dedup).
- All 34 `chain-network` lib tests pass.
- `cargo clippy --workspace --all-targets --all-features` clean under `-D warnings` (incl. pedantic/nursery).

## Note for review

Rule 3 only seeds the cache when a dead-fork download actually **yields** a first block below LIB. If such downloads instead error out (e.g. `BlockNotFound` when ancestry is entirely unavailable network-wide), rule 3 won't fire and rules 1 & 2 won't be seeded — worth validating against real node behavior which path dominates during a storm.

### DIFF
--- a/.github/workflows/code-check.yml
+++ b/.github/workflows/code-check.yml
@@ -78,7 +78,7 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Install cargo-hack
-        uses: taiki-e/install-action@5651179950649c44da31d346537e20c0534f0f25  # Version 2.49.35
+        uses: taiki-e/install-action@13608cbb45b01feb47ef444ab1a42dc41ad56f1a  # Version 2.79.11
         with:
           tool: cargo-hack
 
@@ -191,7 +191,7 @@ jobs:
             -- nodes/node/standalone-node-config.yaml --check-config --deployment nodes/node/standalone-deployment-config.yaml
 
       - name: Install cargo-nextest
-        uses: taiki-e/install-action@5651179950649c44da31d346537e20c0534f0f25 # Version 2.49.35
+        uses: taiki-e/install-action@13608cbb45b01feb47ef444ab1a42dc41ad56f1a # Version 2.79.11
         with:
           tool: cargo-nextest
 
--- a/.github/workflows/end-to-end-with-cucumber.yml
+++ b/.github/workflows/end-to-end-with-cucumber.yml
@@ -66,7 +66,7 @@ jobs:
         run: echo "LOGOS_BLOCKCHAIN_CIRCUITS=$HOME/.logos-blockchain-circuits" >> $GITHUB_ENV
 
       - name: Install cargo-nextest
-        uses: taiki-e/install-action@5651179950649c44da31d346537e20c0534f0f25 # Version 2.49.35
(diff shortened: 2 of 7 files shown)


youngjoon-lee:
@davidrusu Oh, sorry David. I missed this PR, and opened/merged a separate PR https://github.com/logos-blockchain/logos-blockchain/pull/3033 (after we had a call on 26 June). 

In my opinion, we can close this PR in the following reasons.

Differences:
- Base
  - Your PR: `0.1.2-base`
  - My PR was based on master and merged.  
- When we write to cache
  - Your PR: Only when the 1st block of a download is older than LIB
  - My PR: When a block is rejected by chain-service with non-recoverable error (except ParentNotFound)
- When we read cache
  - Your PR: To reject incoming blocks from gossipsub in the chain-network service
  - My PR:
    - To not enqueue blocks to the orphan downloader
    - To not start download when dequeuing a block from the orphan downloader

My PR covers wider range of errors, but uses the cache only for orphan downloader. But, I think it's fine because blocks coming from gossipsub will be immediately tossed to the chain service, which will reject them if they are not acceptable, as I mentioned in the PR description #3033.

If the chain-network <> chain communication makes a log storm, we can minimize the communication by utilizing the cache. But I think, the block rejection is basically the chain service's role. So, I think it's good to keep the cache scoped to the orphan downloader.

PR #3034, opened 2026-06-26
chore: re-organize deployment files for new release process

ntn-x2:

We recently decided to change our release process by not merging back into `master`. So, I wanted to make the current structure a bit easier to follow and more structured with the goal of making clear what files are to be used as templated for ALL releases, what files are templates for a specific release _type_, and what files/values are release-specific. The first two will be changed on `master` and serve as base for all future releases, while the latter will only be changed on the release branch and will never be merged back on `master` by design.

### Changes
- **Per-environment layout.** All environment-specific genesis inputs now live under `deployment/environments/<env>/` (`devnet`, `testnet`, `standalone`):

```
deployment/
  compose.yml  compose.run.yml  compose.setup.yml      # moved from repo root
  compose.static.yml -> compose.run.yml                # infra entrypoint symlink
  compose.tracing.yml
  .env.example  .env.devnet  .env.testnet              # moved from repo root, next to compose
  ceremony/genesis/
    shared/                       # (optional) inputs common to ALL envs — room reserved
    <env>/                        # devnet | testnet | standalone
      inscribe.yaml               # TEMPLATE entropy + PER-RELEASE chain_id/genesis_time
      template/                   # PER-TYPE genesis blueprint
        deployment-template.yaml  stakeholders.yaml  providers.yaml  faucet.yaml
```

- **Genesis inputs grouped by environment** under `deployment/ceremony/genesis/<env>/`, each with a per-type `template/` and a per-release `inscribe.yaml`. The env dirs sit beside a reserved `shared/`, so cross-env files have a natural home.
- **Compose + env files moved into `deployment/`** (`compose.{yml,run.yml,setup.yml}`, the `compose.static.yml` symlink, and `.env.{example,devnet,testnet}`), so all Compose config lives in one place next to `compose.tracing.yml`. Internal relative paths were rewritten to resolve from `deployment/` (`./deployment`→`.`, `./nodes`→`../nodes`, build `context: .`→`..`, `include` → sibling).
- **Responsibility split by comment sections** instead of new files: `inscribe.yaml` uses `# TEMPLATE` (entropy) / `# PER-RELEASE` (chain_id, genesis_time); `.env.<env>` uses `# DEPLOYMENT TYPE` / `# PER-RELEASE` (`NODE_IMAGE_LABEL`).
- **Consumers updated** to the new paths: `.github/workflows/genesis-ceremony.yml`, `scripts/standalone-genesis-ceremony.sh`, a doc comment in `tools/blockchain-tools/src/bin/genesis.rs`, and the release / release-candidate issue templates (inscribe path, `.env` path, and the `compose.static.yml` symlink commands).

Yes.

@ntn-x2 

N/A

No.


.github/ISSUE_TEMPLATE/release-candidate.md
```diff
@@ -30,7 +30,7 @@ Most of the template content is the same or very similar to what is in `release.
 
 ## Devnet genesis (optional, only whenever a new devnet deployment - with a new genesis - is required)
 
-- [ ] Update the inscription file at `deployment/ceremony/genesis/devnet/inscribe.yaml` by setting the `chain_id` field to `X.Y.Z-rc.N` and the `genesis_time` field to be approximately 10 mins in the future, following the existing [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) datetime format. No need to change the `entropy_sources`
+- [ ] Update the inscription file at `deployment/environments/devnet/ceremony/genesis/inscribe.yaml` by setting the `chain_id` field to `X.Y.Z-rc.N` and the `genesis_time` field to be approximately 10 mins in the future, following the existing [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) datetime format. No need to change the `entropy_sources`
```
bacv: `environments` in the path seems unnecessary, as they are implied by the `deployment` dir. 
Also I would prefer `devnet` and `testnet` dirs after the `ceremony/genesis` because that would allow a natural placement of shared files between envs like
```
ceremony
  - genesis
    - shared.yaml
    - devnet
      - devnet_specific.yaml
    - testnet
      - testnet_specific.yaml

File: .github/ISSUE_TEMPLATE/release-candidate.md [hunk 2]
```diff
@@ -68,7 +68,7 @@ Most of the template content is the same or very similar to what is in `release.
 - [ ] Verify the Logos Blockchain tools Docker image was properly built and pushed to the [GitHub container registry][logos-tools-image-container-registry]
 - [ ] Wait for the new Docker image to be built after the release is published. It must have the `X.Y.Z-rc.N` tag.
 - [ ] Checkout `devnet` branch again and change the `compose.static.yml` symlink to now point to `compose.run.yml` with `ln -s -f compose.run.yml compose.static.yml`
-- [ ] Update `.env.devnet` file to contain `NODE_IMAGE_LABEL=X.Y.Z-rc.N` set to version being released
+- [ ] Update `deployment/environments/devnet/.env` file to contain `NODE_IMAGE_LABEL=X.Y.Z-rc.N` set to version being released
```
bacv: This `.env` file is used by the docker compose, I would prefer it to be located near the compose file, seems more straightforward to me. If we go with this approach, we need to update how docker compose is run on the infra side, to read not `.env.<deployment>` but `.env` and also add a step where we symlink this `.env` to the deployment specific `.env`.

.github/ISSUE_TEMPLATE/release-candidate.md
ntn-x2: The point is that the `deployment` folder already has other stuff in it, and having environments on the same level seemed confusing to me. `devnet`, `testnet`, and `standalone` on the same level as `faucet`, `tui-zone` and `l2-sequencer-archival-demo` does not feel right. Either we maintain the `environments` folder, or we group the rest of the stuff into one or more sub-directories, in my opinion. But if we go with `ceremony/genesis/<env>`, then this is not an issue anymore. I can change that.

.github/ISSUE_TEMPLATE/release-candidate.md
ntn-x2: I was actually thinking to move the compose file inside the `deployment` directory as well. What do you think?

.github/ISSUE_TEMPLATE/release-candidate.md
bacv: > But if we go with ceremony/genesis/<env>, then this is not an issue anymore

But this is what we already have?

> on the same level as faucet, tui-zone and l2-sequencer-archival-demo does not feel right.

These are shared between the deployments, so at the moment the `deployment` directory contains:
```
ceremony
faucet
nginx
scripts
...
```
and the environment specifics inside those dirs. If we go the other way around, then we need to duplicate everything per environment? Maybe I misunderstood your suggestion.

ntn-x2: I did not see any environment-specific files in those folders, that's why I assumed they would work with any environment via flags.

> But this is what we already have?

Yeah what I am saying is that we can revert to that structure indeed, I am fine with that. Will actually do that.

.github/ISSUE_TEMPLATE/release-candidate.md
bacv: > I was actually thinking to move the compose file inside the deployment directory as well

Yeah, I think that would be a proper place for `compose.setup.yml` and `compose.run.yml`, then we would symlink from there during deployment.

.github/ISSUE_TEMPLATE/release-candidate.md
bacv: > I did not see any environment-specific files in those folders

Yet :D

bacv:
Lets postpone this PR until we release the testnet 0.2.0.

.github/ISSUE_TEMPLATE/release-candidate.md
ntn-x2: What do you think? https://github.com/logos-blockchain/logos-blockchain/pull/3034/commits/2cc96e835d6374c8b9958ef80f75731f83fe5fd6

.github/ISSUE_TEMPLATE/release-candidate.md
ntn-x2: https://github.com/logos-blockchain/logos-blockchain/pull/3034/commits/2cc96e835d6374c8b9958ef80f75731f83fe5fd6

bacv (approved):
Looks good now, sorry for the trouble. Also are you planning to make a split of optional "genesis ceremony" deployment, as we discussed we could decouple genesis block creation from the "cleanup" during the deployment.

ntn-x2:
@bacv that's done in this PR: https://github.com/logos-blockchain/logos-blockchain/pull/3051. I will rebase that on top of `master` after merging this.

PR #3036, opened 2026-06-26
feat(c-bindings): Surface FFI errors

strinnityk:
Surface FFI errors to callers instead of dumping everything to `stderr`. This way, we let callers decide what to do with those.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/c-bindings/src/errors.rs
+++ b/c-bindings/src/errors.rs
@@ -1,6 +1,8 @@
+use std::ffi::{CStr, CString, c_char};
+
 #[derive(Default, PartialEq, Eq, Debug)]
 #[repr(C)]
-pub enum OperationStatus {
+pub enum OperationStatusCode {
     #[default]
     Ok = 0x0,
     NotFound = 0x1,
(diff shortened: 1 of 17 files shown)


c-bindings/src/errors.rs
```diff
@@ -17,11 +19,37 @@ pub enum OperationStatus {
...
+#[derive(Default)]
+#[repr(C)]
+pub struct OperationStatus {
+    pub code: OperationStatusCode,
+    /// A NUL-terminated description of the error, or null when `code` is
+    /// `Ok` or no message is available.
+    ///
+    /// The caller must free this with
+    /// [`free_cstring`](crate::api::memory::free_cstring).
+    pub message: *mut c_char,
```
danielSanchezQ, line 31: why mut?

c-bindings/src/errors.rs
```diff
@@ -17,11 +19,37 @@ pub enum OperationStatus {
     ValidationError = 0xC,
 }
 
+/// Result of an FFI operation: a status code plus an optional human-readable
+/// message describing what went wrong.
+#[derive(Default)]
+#[repr(C)]
+pub struct OperationStatus {
+    pub code: OperationStatusCode,
```
danielSanchezQ, line 25: we do have to transform this later into the logos-core new result we just integrated 😩

danielSanchezQ (approved):

c-bindings/src/errors.rs
strinnityk: Freeing requires `*mut`, thus `*mut` on owned strings signals ownership, not intent to mutate.

c-bindings/src/errors.rs
strinnityk: You say that like it's a bad thing?

danielSanchezQ: > You say that like it's a bad thing?

yes, we just finished migrating the old code system 😭

strinnityk: You're right, but there's a (good) catch: We don't need to throw away the changes we made to the module.

At most we need an adapter that transforms the FFI's `OperationStatus` to the module's. And error handling as desired, which we really couldn't do before since all was dumped into `stderr`.

TL;DR: Just a tiny bit of glue missing :)

Issue #3054, opened 2026-06-30
0.2.0: lgpd blockchain_module is a devnet build (devnet-0.2.0-rc.1) — testnet operators following the Quick Start can't chainsync


xAlisher:
## Summary

Following the official **0.2.0 Quick Start** (Logos Core module path) to join the **public testnet** produces a node that connects to the documented bootstrap peers but **never chainsyncs** — because the `blockchain_module 0.2.0` pulled by `lgpd` is built for **devnet** (`devnet-0.2.0-rc.1`), while the documented testnet bootstrap peers run **`testnet-0.2.0`**. The two networks' chainsync protocol strings don't match, so the peers reject the sync.

## Reproduction

1. Per the 0.2.0 Quick Start:
   ```
   lgpd download blockchain_module --version 0.2.0 --output ./
   lgpm --modules-dir ./modules install --file blockchain_module-0.2.0.lgx
   logoscore -m ./modules -D && logoscore load-module blockchain_module
   logoscore call blockchain_module generate_user_config '{"initial_peers":[
     "/ip4/65.109.51.37/udp/3000/quic-v1/p2p/12D3KooWFrouXfmrR4nsLMtE7wu15DoMJ6VtoUtHinREZCvbWHar",
     "/ip4/65.109.51.37/udp/3001/quic-v1/p2p/12D3KooWJRGau8M1rjT7R5e4YYsgdFhsMX35nRDtMwCDjxQkXAHz",
     "/ip4/65.109.51.37/udp/3002/quic-v1/p2p/12D3KooWQXJavMDTRscjauFSgVAB1VLB6Rzpy2uY5SU9Tk7927tb",
     "/ip4/65.109.51.37/udp/50001/quic-v1/p2p/12D3KooWSQc7CcGtvWDPF1yCbBthFnQjprfCVHmfmNDUrSmqQsU1"]}'
   logoscore call blockchain_module start user_config.yaml ""
   ```
2. Node reaches `Bootstrapping`, connects (`n_peers=4`), but stays at genesis (`slot=0 height=0`).
3. Logs, repeating for every peer:
   ```
   ERROR logos_blockchain_cryptarchia_sync::libp2p::behaviour: Error while processing tip request:
     Peer <id>: Stream error: failed to open stream:
     remote peer does not support /logos-blockchain-devnet-0.2.0-rc.1/chainsync/1.0.0
   ```

## Root cause (the network id is baked into the build)

```
# lgpd module (Logos Core path):
[...]

PR #3055, opened 2026-06-30
docs(release): add ZK circuits prerequisite to node quick-start

xAlisher:
## What

Adds a **ZK circuits** prerequisite to the node quick-start (`.github/release/release-content.md`).

## Why

Following the quick-start as written, the node **fails to start** — the Prerequisites only download the *node binary*, but the standalone `logos-blockchain-node` loads proving/verifying keys at launch and aborts without `LOGOS_BLOCKCHAIN_CIRCUITS` pointing at a circuits directory:

```
Error: No such file or directory (os error 2)
  nodes/node/binary/src/main.rs:83
```

Hit while bringing up a 0.2.0 testnet node (Linux x86_64). I had to reuse a `v0.4.2` circuits dir from a prior release to get past it.

## Note / follow-up

There is **no circuits asset attached to the 0.2.0 node release**, so even with this doc fix an operator has nowhere to download the matching circuits — the artifact should be published alongside the node binary (and the exact version pinned). Filed as #3054, which also covers the related issue that the `lgpd` `blockchain_module 0.2.0` is a **devnet** build (`devnet-0.2.0-rc.1`) while the testnet bootstrap peers run `testnet-0.2.0`.

Draft — happy to adjust the wording / circuits version once the artifact + version are confirmed.

*Filed with AI assistance (Claude Code).*

### DIFF
--- a/.github/release/release-content.md
+++ b/.github/release/release-content.md
@@ -8,6 +8,17 @@
     tar -xzf logos-blockchain-node-<arch>-<version>.tar.gz
    ```
 
+2. Download and unzip the **ZK circuits**, then point the node at them:
+
+   ```bash
+    tar -xzf logos-blockchain-circuits-<arch>-<version>.tar.gz
+    export LOGOS_BLOCKCHAIN_CIRCUITS="$PWD/logos-blockchain-circuits-<arch>-<version>"
+   ```
+
+   > ⚠️ The node loads proving/verifying keys at launch and **will not start** without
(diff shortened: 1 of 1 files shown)


danielSanchezQ:
@xAlisher this should not happen and its a bug. Does it break in basecamp module `0.2.0`?

davidrusu:
@xAlisher we've baked the circuits into the releases, there is no separate circuits to be downloaded anymore

danielSanchezQ:
Closing this, is not relevant anymore.

PR #3068, opened 2026-07-01
chore: add bounded size and transaction count for blocks

hansieodendaal:

This PR introduces bounded block transactions as a first-class invariant across block creation, serialization/deserialization (_including serde deserialization_), and network ingestion. 

**_Replaces #2921._:** The new type `StorageBounedVec` was removed and replaced with a size validation helper.

At the end state, block transactions are enforced by a bounded container that validates _maximum transaction count_ whereas a size validation helper validates _maximum total serialized size_.

This enforcement is applied consistently through core and service paths, so oversized/invalid transaction sets are rejected at boundaries rather than being accepted and failing later.

Why this matters
- Prevents blocks from exceeding protocol transaction/count/size constraints.
- Makes block validity checks deterministic and centralized.
- Aligns mempool/block assembly and network block intake with the same bounded-transaction rules.
- Preserves real transaction-size semantics (serialized byte size), so limits are meaningful, not placeholder checks.

Network block intake warrants further discussion: previously, the `libp2p` adapter at `async fn request_blocks_from_peer` accepted a raw wire block if it passed deserialized without any validation - the same checks are now implemented as when reconstructing a block from a proposal. 
In practice, this turns the `libp2p` adapter into a hard validation gate:
- peers cannot inject blocks with transaction vectors that violate count/size bounds;
- the same canonical block constraints used in core logic are enforced at network ingress;
- downstream services can rely on already-bounded block transaction data.

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


core/src/block/mod.rs
```diff
@@ -52,13 +55,41 @@ pub struct References {
...
+            signature: raw.signature,
+            transactions: raw.transactions,
+        };
+
+        block
+            .try_into_bounded_block()
+            .map_err(serde::de::Error::custom)
+    }
+}
+
```
danielSanchezQ, line 87: I think this really makes no sense. We have 2 types which are the same just to bridge one from the other.

core/src/block/mod.rs
```diff
@@ -98,74 +137,114 @@ impl<Tx> Block<Tx> {
...
+        } = self;
+
+        if header.slot() == Slot::from(0) {
+            // Genesis block: skip non-genesis bounded + signature validation.
+            return Ok(Self {
+                header,
+                signature,
+                transactions,
+            });
+        }
```
danielSanchezQ: We do have a genesis definition if im not mistaken. We should use genesis to bypass this so its effectively explicit in code if possible (?)

core/src/mantle/genesis_tx.rs
```diff
@@ -45,6 +47,18 @@ pub struct GenesisTx {
...
+        let mut bytes = Vec::new();
+        bytes.extend(encode_mantle_tx(self.mantle_tx()));
+        let ops_proofs_size = self.tx.ops_proofs.bytes_size().unwrap_or_default() as usize;
+        bytes.resize(bytes.len() + ops_proofs_size, 0);
+        bytes.extend(self.cryptarchia_parameter().encode());
+        // Clamp at MAX_BLOCK_SIZE to keep sizing bounded for genesis deserialization.
+        bytes.len().min(MAX_BLOCK_TRANSACTIONS_SIZE)
+    }
+}
+
```
danielSanchezQ: This should not be necessary with genesis block?

danielSanchezQ (commented):
Ok, so GensisBlock is an alias atm, maybe we should make it a wrapper type instead:

```
pub type GenesisBlock = Block<GenesisTx>;
```
to
```
pub struct GenesisBlock(Block<GenesisTx>);
```
Then we implement deserialize for it with the custom skip rules and `AsRef<Block<GenesisTx>>` so we can use it in places we were using it. This may be a bigger refactor than expected maybe. But its more explicit usage of the type system imo.

We can remove `RawBlock` and enforce checks in the block itself (?) and/or at least we change the types in RawBlock to reuse the BoundedVec types instead?

core/src/block/mod.rs
hansieodendaal: It is just getting by the Rust compile limitations, but I will take a closer look, this could maybe be simplified now as the other `StorageBoundedType` was removed.

hansieodendaal: Ok, the way I understand it, this pattern is the correct one to use when custom validation has to happen after serde deserialization as the deserializer will get infinite recursion because it calls itself.

I will move `struct RawBlock<Tx> {` into the ` fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>` funtion to make it less intrusive,

utils/src/bounded_vec.rs
```diff
@@ -11,8 +8,12 @@ use thiserror::Error;
...
     #[error("Input cannot be empty.")]
     EmptyInput,
-    #[error("Length {actual} exceeds static maximum of {max}")]
-    TooLong { actual: usize, max: usize },
+    #[error("Item count {count} is below minimum of {min}")]
+    TooFewItems { count: usize, min: usize },
+    #[error("Item count {count} exceeds static maximum of {max}")]
+    TooManyItems { count: usize, max: usize },
+    #[error("Index {index} is out of bounds for length {len}")]
+    IndexOutOfBounds { index: usize, len: usize },
```
ntn-x2: It seems this error is mapped elsewhere, but it's never actually returned here? So it can probably be removed?

core/src/block/mod.rs
```diff
@@ -52,13 +55,41 @@ pub struct References {
     pub mempool_transactions: Vec<TxHash>,
 }
 
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize)]
 pub struct Block<Tx> {
     header: Header,
     signature: Ed25519Signature,
     transactions: Vec<Tx>,
```
ntn-x2: This is what I was talking about when I mentioned that refactoring bottom-up was probably a better idea than refactoring top-down. Is there any reason why this field here cannot be a `BoundedVec` directly? In that case, we can directly derive serde and (shortly) `NomCodec` because the bounds are already enforced by the bounded vec type.

core/src/block/mod.rs
```diff
@@ -81,12 +112,20 @@ impl Proposal {
     }
 }
 
+/// Validated transaction payload for non-genesis blocks.
+///
+/// `Block<Tx>` stores transactions as a plain `Vec<Tx>` because `GenesisBlock =
```
ntn-x2: Does this mean the genesis block can have an unlimited number of transactions?

core/src/block/mod.rs
```diff
@@ -98,74 +137,114 @@ impl<Tx> Block<Tx> {
...
-                count: tx_size,
-                max: MAX_BLOCK_SIZE,
-            });
-        }
+        Ok(block)
+    }
 
-        let calculated_content_id = Self::calculate_content_id(&transactions);
-        if header.block_root() != &calculated_content_id {
+    fn verify_internal(&self) -> Result<(), Error>
```
ntn-x2: If we're going the type-safe approach, I would make this function consume `self` and return a verified block type of some sort, with all the invariants enforced, so that it's not possible to delete this validation line by mistake in a future refactor.

core/src/block/genesis.rs
```diff
@@ -31,8 +31,12 @@ pub enum Error {
     InvalidGenesisTx(#[from] genesis_tx::Error),
     #[error("add_notes called with empty iterator")]
     EmptyNotes,
+    #[error("too few notes for genesis transfer outputs: attempted {actual}, min {min}")]
+    TooFewNotes { actual: usize, min: usize },
     #[error("too many notes for genesis transfer outputs: attempted {actual}, max {max}")]
     TooManyNotes { actual: usize, max: usize },
+    #[error("Index {index} is out of bounds for length {len}")]
+    IndexOutOfBounds { index: usize, len: usize },
```
ntn-x2, line 39: Unused, if we remove the unused underlying `IndexOutOfBounds` error from the bounded vec type as well.

core/src/mantle/ops/channel/inscribe.rs
```diff
@@ -19,7 +19,7 @@ use crate::{
     },
 };
 
-pub const MAX_BYTES: usize = MAX_BLOCK_SIZE * 7 / 8;
+pub const MAX_BYTES: usize = MAX_BLOCK_TRANSACTIONS_SIZE * 7 / 8;
```
ntn-x2, line 25: Maybe it's time to at least comment on how we take 7/8ths?

utils/src/bounded_vec.rs
hansieodendaal: It is returned by `BoundedVec::try_remove` when `index >= self.len()`, and there is a unit test covering that case. The mapping in genesis is for callers that propagate bounded-vector mutation errors through genesis-builder errors.

core/src/block/mod.rs
hansieodendaal: Good catch, with the refactor @danielSanchezQ suggested this is now possible

core/src/block/mod.rs
hansieodendaal: Fixed with above suggestion

utils/src/bounded_vec.rs
ntn-x2: Yes I though `try_remove` now returns `None` for those cases. Sorry.

core/src/block/mod.rs
hansieodendaal: I changed this to consume `self` for now:

```rust
fn into_verified(self) -> Result<Self, Error>
```

so validation is tied to the value being returned rather than just mutating/checking by reference. Introducing a separate `VerifiedBlock<Tx>` type could be a follow-up PR if we want to make the distinction explicit in the type system.

core/src/block/genesis.rs
hansieodendaal: See above comment

core/src/mantle/ops/channel/inscribe.rs
hansieodendaal: Added a comment

core/src/block/mod.rs
```diff
@@ -81,91 +107,120 @@ impl Proposal {
...
+        transactions: BlockTransactions<Tx>,
         signing_key: &Ed25519Key,
     ) -> Result<Self, Error>
     where
         Tx: Transaction<Hash = TxHash> + StorageSize,
     {
+        if slot == Slot::genesis() {
+            return Err(Error::Validation("expected non-genesis slot".to_owned()));
+        }
+
```
youngjoon-lee, line 132: Above, `GenesisBlock` wraps `Block`. `GenesisBlock` requires the slot 0, but `Block` requires non-zero slot here. Doesn't it cause any conflict/confusion? What do you think about removing this constraint?

core/src/block/mod.rs
```diff
@@ -81,91 +107,120 @@ impl Proposal {
...
         let signature = header.sign(signing_key)?;
 
-        Ok(Self {
+        let block = Self {
             header,
             signature,
             transactions,
-        })
+        };
+        block.validate_total_transactions_size()?;
```
youngjoon-lee, line 154: Do you think we can replace this with `into_verified`?

core/src/block/mod.rs
hansieodendaal: There are no conflicts at the moment, and insofar as confusion, I would think the constraints make it clear what the expectation is.

Would this be a blocking issue for you? Do you propose a more definitive contract with creation, for example `ZeroSlot` and `NonZeroSlot` derivatives?

core/src/block/mod.rs
hansieodendaal: I considered it at the time, and looked at it again, but that will result in duplicate verification steps and wasted cycles:
- `calculate_content_id` will be calculated twice.

core/src/mantle/transactions/tx.rs
```diff
@@ -176,6 +176,12 @@ impl MantleTxGasContext {
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct MantleTx(pub Ops);
 
+impl StorageSize for MantleTx {
+    fn storage_size(&self) -> usize {
+        encode_mantle_tx(self).len()
+    }
+}
+
```
danielSanchezQ, line 184: Why is this needed now? I think we only did implement this for `AuthenticatedMantleTx` for a reason.

core/src/sdp/mod.rs
```diff
@@ -99,63 +99,104 @@ pub struct InactivityPeriodTooSmall {
     pub period: NumberOfEpochs,
 }
 
-pub const MAX_LOCATOR_BYTE_SIZE: usize = 329;
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
+#[serde(try_from = "Multiaddr")]
+struct BoundedMultiaddr<const MAX_SIZE: usize>(Multiaddr);
```
danielSanchezQ: Uhm, why is this in this pr?

core/src/sdp/mod.rs
danielSanchezQ, line 1: I think this whole module comes from a rebase or something, do not understand why it has to deal with the block.

danielSanchezQ (changes requested):

core/src/block/mod.rs
youngjoon-lee: I don't see any correctness issue. I was just confused because, while reading the code, I set a mental model that `Block::header::slot` is always `> 0`. But, I found that `GenesisBlock` wraps `Block`. It breaks the mental model that I set. 

For example, in this architecture, we cannot build `GenesisBlock` from an existing `Block`. But, I think we don't have that use case.

I think, the current architecture is fine. We may be able to clearly separate two block types by wrapping a common type that doesn't have the slot invariant. But, I'm not sure if it is necessary/meaningful. We can think about it later if necessary.

youngjoon-lee (approved):
I did another round of review. I think this PR achieves its goal. There might be some improvement we can try, as I mentioned in the comment above. But, I don't think it's critical. Thank you!

youngjoon-lee (approved):
I did another round of review. I think this PR achieves its goal. There might be some improvement we can try, as I mentioned in the comment above. But, I don't think it's critical. Thank you!

core/src/sdp/mod.rs
hansieodendaal: Yeah, bad merge

core/src/sdp/mod.rs
hansieodendaal: ditto bad merge

core/src/mantle/transactions/tx.rs
hansieodendaal: `TryFrom<Bytes> for Block<Tx>` needs `StorageSize`. Previously `StorageSize` was only used in tests and `impl StorageSize for MantleTx` was declared inside the test module of `core/src/block/mod.rs`.

Call path:
```rust
fn try_from(bytes: Bytes) -> Result<Self, Self::Error> 
    ...
    fn into_verified(self) -> Result<Self, Error>
        ...
        fn validate_total_transactions_size(&self) -> Result<usize, Error>
```

danielSanchezQ (approved):

PR #3071, opened 2026-07-02
chore: strengthen bounded-vec invariant protection

hansieodendaal:

This PR is a defensive refinement that removes mutable escape hatches from `BoundedVec` and wrapper types (`Outputs`, `Inputs`) to enforce stricter bounds checking and encapsulation. All length-changing operations now go through validated
checked methods, preventing clients from accidentally or intentionally violating the `MIN..=MAX` length invariant.

Invariant Safety:

- No client can bypass bounds checks by obtaining mutable slices.
- All length mutations protected by `try_push()`, `try_pop()`, `try_remove()` with `MIN/MAX` validation.
- `new_unchecked()` remains as documented escape hatch for trusted internal code only.

Encapsulation:

- Ledger `Outputs` and `Inputs` now own all mutation policy.
- Single source of truth for length-changing operations per wrapper type.

Clarity:

- Intent is explicit: safe element mutation via iter_mut(), safe insertion via `try_push(), etc.
- Errors are specific: `OutputsError`, `InputsError` instead of generic `BoundedError`.
- API is audit-friendly: all mutable paths are validated in one place.

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/utils/src/bounded_vec.rs
+++ b/utils/src/bounded_vec.rs
@@ -1,7 +1,4 @@
-use core::{
-    ops::{Deref, DerefMut},
-    slice::Iter,
-};
+use core::{ops::Deref, slice::Iter};
 use std::{str::FromStr, vec::IntoIter};
 
 use serde::{Deserialize, Serialize};
@@ -11,8 +8,12 @@ use thiserror::Error;
 pub enum BoundedError {
     #[error("Input cannot be empty.")]
     EmptyInput,
-    #[error("Length {actual} exceeds static maximum of {max}")]
-    TooLong { actual: usize, max: usize },
+    #[error("Item count {count} is below minimum of {min}")]
+    TooFewItems { count: usize, min: usize },
+    #[error("Item count {count} exceeds static maximum of {max}")]
+    TooManyItems { count: usize, max: usize },
+    #[error("Index {index} is out of bounds for length {len}")]
+    IndexOutOfBounds { index: usize, len: usize },
 }
 
 /// `Vec<T>` whose length is statically enforced to be in the range `[MIN,
@@ -81,13 +82,52 @@ impl<T, const MIN: usize, const MAX: usize> BoundedVec<T, MIN, MAX> {
 
     pub fn try_push(&mut self, item: T) -> Result<(), BoundedError> {
         if self.len() >= MAX {
-            return Err(BoundedError::TooLong {
-                actual: self.len() + 1,
+            return Err(BoundedError::TooManyItems {
+                count: self.len() + 1,
                 max: MAX,
             });
         }
         self.0.push(item);
         Ok(())
+    }
+
+    pub fn try_pop(&mut self) -> Result<Option<T>, BoundedError> {
+        if self.is_empty() {
+            return Ok(None);
+        }
+
+        let new_len = self.len() - 1;
+        if new_len < MIN {
+            return Err(BoundedError::TooFewItems {
+                count: new_len,
+                min: MIN,
+            });
+        }
+
+        Ok(self.0.pop())
+    }
+
+    pub fn try_remove(&mut self, index: usize) -> Result<T, BoundedError> {
+        if index >= self.len() {
+            return Err(BoundedError::IndexOutOfBounds {
+                index,
+                len: self.len(),
+            });
+        }
+
+        let new_len = self.len() - 1;
+        if new_len < MIN {
+            return Err(BoundedError::TooFewItems {
+                count: new_len,
+                min: MIN,
+            });
+        }
+
+        Ok(self.0.remove(index))
+    }
+
+    pub fn iter_mut(&mut self) -> core::slice::IterMut<'_, T> {
+        self.0.iter_mut()
     }
 }
 
@@ -107,12 +147,17 @@ impl<T, const MIN: usize, const MAX: usize> TryFrom<Vec<T>> for BoundedVec<T, MI
     type Error = BoundedError;
 
     fn try_from(value: Vec<T>) -> Result<Self, Self::Error> {
-        if value.len() < MIN {
+        if value.len() < MIN && value.is_empty() {
             return Err(BoundedError::EmptyInput);
+        } else if value.len() < MIN {
+            return Err(BoundedError::TooFewItems {
+                count: value.len(),
+                min: MIN,
+            });
         }
         if value.len() > MAX {
-            return Err(BoundedError::TooLong {
-                actual: value.len(),
+            return Err(BoundedError::TooManyItems {
+                count: value.len(),
                 max: MAX,
             });
         }
@@ -133,7 +178,8 @@ where
 
 impl<T, const MIN: usize, const MAX: usize> From<T> for BoundedVec<T, MIN, MAX> {
     fn from(value: T) -> Self {
-        const { assert!(MAX >= 1, "Max size cannot be zero.") }
+        const { assert!(MIN <= 1, "Single element is below BoundedVec MIN") }
+        const { assert!(MAX >= 1, "Single element exceeds BoundedVec MAX") }
         Self([value].into())
     }
 }
@@ -186,9 +232,12 @@ impl<T, const MIN: usize, const MAX: usize> Deref for BoundedVec<T, MIN, MAX> {
     }
 }
 
-impl<T, const MIN: usize, const MAX: usize> DerefMut for BoundedVec<T, MIN, MAX> {
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        &mut self.0
+impl<'a, T, const MIN: usize, const MAX: usize> IntoIterator for &'a mut BoundedVec<T, MIN, MAX> {
+    type Item = &'a mut T;
+    type IntoIter = std::slice::IterMut<'a, T>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.iter_mut()
     }
 }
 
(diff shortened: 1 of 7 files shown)


utils/src/bounded_vec.rs
```diff
@@ -186,9 +232,12 @@ impl<T, const MIN: usize, const MAX: usize> Deref for BoundedVec<T, MIN, MAX> {
     }
 }
 
-impl<T, const MIN: usize, const MAX: usize> DerefMut for BoundedVec<T, MIN, MAX> {
```
ntn-x2, line 189: A mutable reference to a slice allows to change the slice elements, not its length. So it's not unsafe, and that's by design. You can't push new items or remove existing items from a slice, but you can (if needed mutably) iterate ove them, or access with indices, that's it.

utils/src/bounded_vec.rs
```diff
@@ -133,7 +178,8 @@ where
 
 impl<T, const MIN: usize, const MAX: usize> From<T> for BoundedVec<T, MIN, MAX> {
     fn from(value: T) -> Self {
-        const { assert!(MAX >= 1, "Max size cannot be zero.") }
+        const { assert!(MIN <= 1, "Single element is below BoundedVec MIN") }
```
ntn-x2: Why? A bounded vec with minimum length of 0 is fine.

utils/src/bounded_vec.rs
```diff
@@ -81,14 +82,53 @@ impl<T, const MIN: usize, const MAX: usize> BoundedVec<T, MIN, MAX> {
...
+                count: self.len() + 1,
                 max: MAX,
             });
         }
         self.0.push(item);
         Ok(())
     }
+
+    pub fn try_pop(&mut self) -> Result<Option<T>, BoundedError> {
+        if self.is_empty() {
```
ntn-x2: Pop can just call the `remove` implementation below with the right index, right?

ntn-x2: Also because the remove does not have the empty check, but it should.

utils/src/bounded_vec.rs
```diff
@@ -107,12 +147,17 @@ impl<T, const MIN: usize, const MAX: usize> TryFrom<Vec<T>> for BoundedVec<T, MI
     type Error = BoundedError;
 
     fn try_from(value: Vec<T>) -> Result<Self, Self::Error> {
-        if value.len() < MIN {
+        if value.len() < MIN && value.is_empty() {
```
ntn-x2, line 144: Why? If MIN is 2 and value has one element, we should still fail.

ntn-x2: But also this is not correct, because we a MIN of zero, an empty input is fine. No?

ntn-x2 (changes requested):
Will review again once my first round of comments is addressed. I would like to know the rationale behind many of the decisions you took here, if possible.

utils/src/bounded_vec.rs
hansieodendaal: Yeah, `try_pop` can be routed to `try_remove`
An empty vec is checked by this invariant already: `if index >= self.len() `

utils/src/bounded_vec.rs
hansieodendaal: I expanded the tests to show they all pass as intended
- An empty vec with `value.len() < MIN` gives `BoundedError::EmptyInput)`
- A non-empty vec with `value.len() < MIN` gives `BoundedError::TooFewItems `
- A non-empty vec with `value.len() > MAX ` gives `BoundedError::TooManyItems`

hansieodendaal: Expanded tests
```rust
    type TestBoundedVector = BoundedVec<u8, 2, 4>;
    type TestBoundedVectorMin0 = BoundedVec<u8, 0, 4>;

    #[test]
    fn try_from_accepts_lengths_within_bounds() {
        TestBoundedVectorMin0::try_from(vec![]).unwrap();
        TestBoundedVectorMin0::try_from(vec![1]).unwrap();
        TestBoundedVectorMin0::try_from(vec![1, 2]).unwrap();
        TestBoundedVectorMin0::try_from(vec![1, 2, 3]).unwrap();
        TestBoundedVectorMin0::try_from(vec![1, 2, 3, 4]).unwrap();

        TestBoundedVector::try_from(vec![1, 2]).unwrap();
        TestBoundedVector::try_from(vec![1, 2, 3]).unwrap();
        TestBoundedVector::try_from(vec![1, 2, 3, 4]).unwrap();
    }

    #[test]
    fn try_from_rejects_input_below_min() {
        assert_eq!(
            TestBoundedVector::try_from(vec![]),
            Err(BoundedError::EmptyInput)
        );
        assert_eq!(
            TestBoundedVector::try_from(vec![1]),
            Err(BoundedError::TooFewItems { count: 1, min: 2 })
        );
    }

    #[test]
    fn try_from_rejects_input_above_max() {
        assert_eq!(
            TestBoundedVector::try_from(vec![1, 2, 3, 4, 5]),
            Err(BoundedError::TooManyItems { count: 5, max: 4 })
        );
        assert_eq!(
            TestBoundedVectorMin0::try_from(vec![1, 2, 3, 4, 5]),
            Err(BoundedError::TooManyItems { count: 5, max: 4 })
        );
    }
```

utils/src/bounded_vec.rs
ntn-x2: Yeah what I am suggesting is that `try_remove` should include the empty check, and `try_pop` just call `try_remove` with the right index. But I think the emptiness check is actually irrelevant since we check for the index and final length anyway.

utils/src/bounded_vec.rs
ntn-x2: Ok I see that we use both checks to return a more specific error variant, instead of `TooFewItems` in all cases. I might be fine with that. I think it would be easier to read `if value.is_empty() && MIN > 0`, although I don't necessarily see a strong need for checking this case specifically, but I won't oppose it.

utils/src/bounded_vec.rs
hansieodendaal: `From<T>` creates a `BoundedVec` with length exactly `1`, so length `1` must satisfy the bounds: `MIN <= 1 && MAX >= 1`.

A bounded vec with `MIN = 0` is still valid here, since `0 <= 1`. The assertion was not intended to reject `MIN = 0`; it only rejects bounds where a single-element construction would violate the minimum, e.g. `MIN > 1`.

I agree the current assertion messages are misleading, though. I’ll clarify them as:

```rust
const { assert!(MIN <= 1, "Single-element construction is invalid for minimum bound > 1") }
const { assert!(MAX >= 1, "Single-element construction is invalid for maximum bound < 1") }
```

I also added a test to demonstrate.

utils/src/bounded_vec.rs
```diff
@@ -186,9 +232,12 @@ impl<T, const MIN: usize, const MAX: usize> Deref for BoundedVec<T, MIN, MAX> {
     }
 }
 
-impl<T, const MIN: usize, const MAX: usize> DerefMut for BoundedVec<T, MIN, MAX> {
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        &mut self.0
+impl<'a, T, const MIN: usize, const MAX: usize> IntoIterator for &'a mut BoundedVec<T, MIN, MAX> {
```
ntn-x2, line 245: I'm fine having the `IntoIterator` impl, but I would like to leave the `DerefMut` into a slice, since as I explained above it's not really a safety issue since access to a slice won't allow for changing the number of elements.

utils/src/bounded_vec.rs
hansieodendaal: Agreed. I conflated “mutable access” with “length-changing access” here. The invariant only depends on length, and &mut [T] cannot alter length. I’ll revert this.

utils/src/bounded_vec.rs
ntn-x2: Ok, fair enough, I missed that case.  Good catch!

utils/src/bounded_vec.rs
```diff
@@ -96,18 +96,12 @@ impl<T, const MIN: usize, const MAX: usize> BoundedVec<T, MIN, MAX> {
...
-                count: new_len,
-                min: MIN,
-            });
-        }
-
-        Ok(self.0.pop())
+        self.try_remove(self.len() - 1).map(Some)
     }
 
     pub fn try_remove(&mut self, index: usize) -> Result<T, BoundedError> {
```
ntn-x2, line 102: `remove` panics, and `try_remove` (still experimental) returns an `Option`. Maybe it would be worth returning `None` here as well? In that case we would not even need to check for emptiness in the `try_pop` method since that case of the empty vector would be caught in this if check here.

ntn-x2 (approved):
Left one final comment for further simplifying the new API, but it's not blocking. Feel free to do what you want with the suggestion.

utils/src/bounded_vec.rs
hansieodendaal: Thanks for the suggestion, I saw the inconsistency and decided to fix it the other way round.

Previous:
`MIN=0, len=0: try_pop -> Ok(None)`
`MIN=2, len=2: try_pop -> Err(BoundedError::IndexOutOfBounds { ..})`

Now consistent:
`MIN=0, len=0: try_pop -> Ok(None)`
`MIN=2, len=2: try_pop -> Ok(None)`

PR #3073, opened 2026-07-02
refactor(kms): propagate execute errors

### LINKED ISSUE #2079: KMS service should bubble up errors to callers
If the key given by the caller service doesn't exist in KMS, the KMS service silent print a log and doesn't nothing.
Then, the response channel that the caller is waiting for is dropped, and the caller has nothing it can do.

andrussal:

This PR makes KMS operator execution errors observable to callers.

Previously `KmsServiceApi::execute` only reported that the execute message was queued. If the backend failed afterwards, the service logged the error but callers never received a response — most panicked at that point, others could continue with incorrect assumptions. This now replies with the execution result, preserves the backend error in the API, and updates wallet/leader/Blend call sites to handle that result according to their local policy.

It also avoids advancing wallet voucher state when voucher derivation fails, keeps Blend startup misconfiguration fail-fast, and adds proposal-failure metrics for leadership proof failures so persistent KMS/operator issues are visible.

Closes [#2079](https://github.com/logos-blockchain/logos-blockchain/issues/2079)

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


strinnityk (approved):
I don't think I'd call this `fix`, though.

services/key-management-system/src/lib.rs
```diff
@@ -188,17 +188,121 @@ where
...
+                })),
+                reply_channel,
+            },
+            &mut backend,
+        )
+        .await;
+
+        assert!(rx.await.expect("execute response should be sent").is_ok());
+    }
+}
```
strinnityk: Didn't we already have similar tests to these?

andrussal:
> I don't think I'd call this `fix`, though.

Fair, refactor is probably better.

services/key-management-system/src/lib.rs
andrussal: > Didn't we already have similar tests to these?

Yeah, fair point. I think this doesn’t add much value besides testing that errors are sent through the channel, which probably doesn’t need a separate test.

services/chain/chain-leader/src/lib.rs
```diff
@@ -440,8 +440,24 @@ where
...
+                        {
+                            Ok(proof) => proof,
+                            Err(e) => {
+                                error!(target: LOG_TARGET, "Failed to build leadership proof for slot {slot:?}: {e}");
+                                continue;
+                            }
+                        };
 
-                       if let Some((proof, signing_key)) = build_proof_for(&eligible_aged, latest_tree, &epoch_state, slot, &wallet_api, &kms_api).await {
+                        if let Some((proof, signing_key)) = proof {
```
ntn-x2, line 463: Small fmt issue :)

ntn-x2 (approved):
LGTM!

youngjoon-lee (approved):

PR #3076, opened 2026-07-02
refactor(mantle): Transactions and Op encoding

strinnityk:

Mostly just moving code around, without behavioural changes, to improve how crowded mantle's encoding has become and to prepare for the `Transaction` refactor:

- Pulled all the transaction stuff into its own module.
- Renamed `encoding.rs` into `codec.rs` everywhere.
- Move each codec to its specific module. The ones without a specific module remain in `mantle/codec.rs`, split into inner mods (`primitives`/`crypto`) so it also eases out later into a per-file split.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.


core/src/mantle/transactions/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+// ==============================================================================
+// These limits are not designed to mimic system limits, but rather to prevent
+// unbounded memory usage from malicious inputs. They prevent memory
+// over-allocation attacks where untrusted input specifies allocation sizes.
+// Values are chosen to not limit normal operations while preventing excessive
+// memory usage (e.g., 68GB allocation). As an example, if the network currently
+// limits maximum transaction size to 1MiB, for memory safety limits we can
+// allow 4MiB.
+pub const MAX_OPS_PER_TX: usize = u8::MAX as usize;
+pub type Ops = UpperBoundedVec<Op, MAX_OPS_PER_TX>;
```
danielSanchezQ, line 29: Just to double check, this does come from earlier and you just moved right?

core/src/mantle/codec.rs
```diff
@@ -0,0 +1,250 @@
+pub mod primitives {
```
danielSanchezQ: I would really prefer to have files instead of nested modules to be honest.

core/src/mantle/codec.rs
```diff
@@ -0,0 +1,250 @@
...
+            let ts = OffsetDateTime::now_utc();
+            let data = encode_unix_timestamp(&ts);
+            let (remaining, value) = decode_unix_timestamp(&data).unwrap();
+            assert_eq!(value, ts.truncate_to_second());
+            assert!(remaining.is_empty());
+        }
+    }
+}
+
+pub mod crypto {
```
danielSanchezQ: definetely better to separate this into smaller files
```
/codec
    mod.rs
    crypto.rs
    /primitives
        encoding.rs
        decoding.rs
...
```

core/src/mantle/ledger.rs
```diff
@@ -14,14 +14,26 @@ use thiserror::Error;
...
+// Values are chosen to not limit normal operations while preventing excessive
+// memory usage (e.g., 68GB allocation). As an example, if the network currently
+// limits maximum transaction size to 1MiB, for memory safety limits we can
+// allow 4MiB.
+const MAX_TRANSACTION_INPUTS: usize = u8::MAX as usize;
+const MAX_TRANSACTION_OUTPUTS: usize = u8::MAX as usize;
+pub type BoundedUtxos = UpperBoundedVec<Utxo, MAX_TRANSACTION_INPUTS>;
+pub type BoundedInputs = UpperBoundedVec<NoteId, MAX_TRANSACTION_INPUTS>;
+pub type BoundedOutputs = UpperBoundedVec<Note, MAX_TRANSACTION_OUTPUTS>;
+
```
danielSanchezQ, line 36: Same here, double check, we moved this were this introduced now?

core/src/proofs/channel_multi_sig_proof.rs
```diff
@@ -140,6 +140,64 @@ impl TryFrom<Vec<IndexedSignature>> for ChannelMultiSigProof {
     }
 }
 
+pub mod encoding {
```
danielSanchezQ: this one may be fine

danielSanchezQ: Will this be replaced by a NomEncode/Decode trait impl later?

core/src/proofs/channel_multi_sig_proof.rs
```diff
@@ -140,6 +140,64 @@ impl TryFrom<Vec<IndexedSignature>> for ChannelMultiSigProof {
...
+    pub fn encode_channel_multi_sig_proof(proof: &ChannelMultiSigProof) -> Vec<u8> {
+        let mut bytes = Vec::new();
+        bytes.extend(encode_uint16(proof.signatures().len() as ChannelKeyIndex));
+        bytes.extend(proof.signatures().iter().flat_map(|signature| {
+            encode_ed25519_signature(&signature.signature)
+                .into_iter()
+                .chain(encode_uint16(signature.channel_key_index))
+        }));
+        bytes
+    }
```
danielSanchezQ: If we cant allocate the vec before hand would be better to chain the bytes and collect. But this comes from legacy so its fine. Just leaving as a not in case we could change it later if we implement NomEncode/Decode

danielSanchezQ (changes requested):
I would use files instead of nested modules, its easier to locate imo. Otherwise seems good to go. Left some comments too.

core/src/mantle/transactions/mod.rs
strinnityk: Yes, moved from `mantle/encoding.rs`.

core/src/mantle/ledger.rs
strinnityk: Same thing, moved from `mantle/encoding.rs`.

core/src/mantle/codec.rs
strinnityk: Perfectly fine with that, just decided to do incremental steps at first.

core/src/proofs/channel_multi_sig_proof.rs
strinnityk: It was... until Antonio stole the first place from me 😂

danielSanchezQ (approved):

PR #3078, opened 2026-07-02
chore: migrate proofs to NomEncode and NomDecode

ntn-x2:

First PR for migrating proofs to the new `NomEncode`/`NomDecode` trait system, and adding well-known fixtures to capture future regressions.

One thing that is not yet tackled, is how to map from an op to a proof. I will open a Discord thread about changing the wire format so that each op is immediately followed by its required proof type (specified via an associated type on a `Op` trait). That will make the code much easier to follow, and also would allow us to fail immediately if an op is not followed by a properly encoded proof as required. But anyway, this is a separate discussion. This PR is on purpose limited in scope and only migrate the encode and decode functions for those proofs, for now.

Yes.

@ntn-x2 

Yes.

No.


core/src/proofs/channel_multi_sig_proof.rs
```diff
@@ -94,20 +99,19 @@ impl From<ChannelMultiSigProof> for ChannelMultiSigProofRepr {
...
     /// must be strictly increasing (so they are ordered and unique, per the
-    /// `CHANNEL_CONFIG` / `CHANNEL_WITHDRAW` spec), and the count must not
-    /// exceed `ChannelKeyIndex::MAX`.
+    /// `CHANNEL_CONFIG` / `CHANNEL_WITHDRAW` spec).
     ///
     /// This validates structural correctness only. Cryptographic validity
     /// (signature verification, threshold requirements, index-to-key
     /// correspondence) must be checked separately.
-    fn validate_well_formedness(signatures: &[IndexedSignature]) -> Result<(), Error> {
+    fn validate_well_formedness(signatures: IndexedSignatures) -> Result<IndexedSignatures, Error> {
```
hansieodendaal, line 136: Why not this? Why take ownership and pass it back out again?
```rust
fn validate_well_formedness(signatures: &IndexedSignatures) -> Result<(), Error> {
```

hansieodendaal (changes requested):
Nice job overall - I only picked up one main issue, if I am not mistaken.

The old decode path for `ChannelMultiSigProof` called `validate_well_formedness`, whereas the new derived `NomCodec` path bypasses it. `ChannelMultiSigProof` should not derive `NomCodec` while it has a fallible well-formedness invariant: the derived Nom decoder constructs the struct directly from decoded fields, bypassing `try_new` / `validate_well_formedness`.

That means a Nom wire payload with duplicate or non-increasing `channel_key_index` values can still become a `ChannelMultiSigProof`, even though those cases are intended to be unrepresentable and downstream validation relies on that uniqueness/order invariant.

Could we remove `NomCodec` from `ChannelMultiSigProof` and add hand-written Nom impls in this module instead? For example:

```rust
impl NomEncode for ChannelMultiSigProof {
    fn encode(&self) -> Vec<u8> {
        self.signatures.encode()
    }
}

impl NomDecode for ChannelMultiSigProof {
    fn decode(bytes: &[u8]) -> nom::IResult<&[u8], Self> {
        let (remaining, signatures) = IndexedSignatures::decode(bytes)?;
        let proof = Self::try_new(signatures).map_err(|_| {
            nom::Err::Failure(nom::error::Error::new(
                remaining,
                nom::error::ErrorKind::Verify,
            ))
        })?;
        Ok((remaining, proof))
    }
}
```

We should also add Nom regression tests to prove the wire decoder rejects invalid proofs, e.g.:
- `nom_decode_rejects_repeated_index`
- `nom_decode_rejects_descending_indices`

core/src/proofs/channel_multi_sig_proof.rs
ntn-x2: Yes the old function was taking a slice, but this is a bounded vector, so I thought about iterating over a type with a well-known size instead. It's a minor detail tho.

ntn-x2:
@hansieodendaal good catch. I fixed the issue in https://github.com/logos-blockchain/logos-blockchain/pull/3078/commits/d0e4d5d7631f98ac102786a8724c8ae6214d8579. I implemented `NomEncode` and `NomDecode` by hand for the `IndexedSignatures` type instead, since that's where the condition must be met. I left the `NomCodec` automatic implementation for the op itself, since it will call `decode` on the inner type.

hansieodendaal (approved):
LGTM!

danielSanchezQ (approved):

PR #3079, opened 2026-07-02
test(zone-sdk): failing test — gap-backfilled inscriptions never surface as adopted

davidrusu:
## What

Adds one deliberately **failing** unit test, `sequencer::block_fetch::tests::canonical_backfill_gap_inscriptions_surface_as_adopted`, demonstrating a bug in the channel-view model introduced on `pr/zone-sdk-test-policy`. This branch is stacked on top of `pr/zone-sdk-test-policy` (the test depends on the new `channel_lineage` / `detect_channel_update` API), so the diff against master includes that branch's commits; the last commit is the test.

## The bug

`handle_block_event` captures `old_lineage` (block_fetch.rs, after the backfill block) **after** `backfill_canonical` has already inserted the stream-gap blocks into state. `channel_lineage`'s forward walk bridges through the freshly backfilled inscriptions, so they land on the "before" side of the diff, `adopted = new − old` comes out empty, and `detect_channel_update` returns `None` — no consumer ever learns those inscriptions landed until finalization. This violates the invariant already documented at the capture site ("the 'before' side of the diff isn't contaminated") for the backfill path.

## Test scenario

Chain `G ← B1 ← B2 ← B3`, one canonical branch, LIB pinned at genesis:

1. **B1** (live) carries inscription **A** — correctly reported as adopted.
2. **B2** carries inscription **Y** (child of A) — *skipped by the block stream*.
3. **B3** (live) arrives with its parent missing → `backfill_canonical` fetches B2 and inserts it **before** `old_lineage` is captured.

Expected: this event's `ChannelUpdate` reports Y as adopted.
Actual: `channel_update` is `None`:

```
panicked at zone-sdk/src/sequencer/block_fetch.rs:
the backfilled gap advanced the channel tip; expected a channel update
```

## Verified fix (not included)

Moving the `old_lineage` capture to before the canonical backfill (one line) makes this test pass, and all 34 existing zone-sdk unit tests still pass with the reorder. Left out so the PR demonstrates the failure; happy to push the fix here too if preferred.

Note: the capture would still sit after the **LIB** backfill (step 1). That's probably correct — LIB-backfilled inscriptions surface via the finalized path, and counting them as already-known avoids double-reporting — but worth a deliberate look since it's the same mutate-before-snapshot pattern.


davidrusu:
Added a second, more realistic test: `sequencer::actor::tests::stream_gap_surfaces_backfilled_inscriptions_as_adopted`.

Instead of calling `handle_block_event` directly, it drives a real `ZoneSequencer` end-to-end through `next_event()` with the realistic trigger for the canonical backfill:

1. The live block stream delivers **B1** (inscription A) and then **ends** — a stream drop.
2. **B2** (inscription Y, child of A) is mined while the sequencer is disconnected.
3. The sequencer reconnects inside `next_event()`; the new stream resumes at **B3**, whose parent is unknown locally, so `backfill_canonical` fetches B2.

Per the `Event::Ready` contract ("any tx invalidated by the catch-up surfaces via `ChannelUpdate::orphaned` on the next `BlocksProcessed` once the stream resumes") and the `ChannelUpdate` consumer pattern ("mirror the channel: revert every orphaned entry and apply every adopted entry"), the post-reconnect `BlocksProcessed` must carry Y as `adopted`. It doesn't:

```
inscription mined during the stream gap must surface as adopted on the next
BlocksProcessed after reconnect; got adopted=[], orphaned=[]
```

Same root cause as the unit test, now observable purely from the public API — a consumer following the documented mirroring pattern silently loses Y from its channel view until finalization. Both tests pass with the one-line reorder (capture `old_lineage` before `backfill_canonical`), under which all 33 pre-existing zone-sdk unit tests remain green.

pradovic (approved):
Looks good! I have added the suggested fix. 

I have also discovered another similar bug with orphaned messages. Finalized messages were also reported as orphans in some cases. This is fixed as well now, plus I added couple of more unit tests to cover all of the mentioned cases.

zone-sdk/src/sequencer/state.rs
```diff
@@ -779,16 +782,23 @@ impl TxState {
         let old_ids: HashSet<MsgId> = old_lineage.iter().map(|i| i.this_msg).collect();
         let new_ids: HashSet<MsgId> = new_lineage.iter().map(|i| i.this_msg).collect();
 
+        // Each lineage stops at the LIB of its capture time, so a LIB
+        // advance between the captures shifts the diff's lower boundary.
+        // Mask the finalized prefix on both sides so the shifted floor
+        // doesn't read as adopted/orphaned content.
+        let mut finalized = self.finalized_prefix_ids(old_lineage);
+        finalized.extend(self.finalized_prefix_ids(&new_lineage));
```
hansieodendaal, line 824: Could this hide orphans when competing branches end with the same replayed `ChannelConfig`?

For example, starting from the same finalized tip `P`:

- old branch: `[Inscribe U(parent=P), Config X]`
- new finalized branch: `[Inscribe V(parent=P), Config X]`

pradovic: Good catch! Fixed.
We will have config lineage soon so we might need to revisit this but the fix was simple. Thanks!

I have also added a unit test as well.

hansieodendaal (approved):
LGTM!

PR #3070, opened 2026-07-02
feat(core): set genesis gas prices to 1

youngjoon-lee:
## 1. What does this PR implement?

- Change the genesis gas prices from 0 to 1 
  - Storage gas price $P_{STR}(0) = 1$ ([spec](https://lip.logos.co/blockchain/raw/storage-markets.html#protocol-constants))
  - Execution gas price $b_{exec}[0] = 1$ ([spec](https://lip.logos.co/blockchain/raw/execution-market.html#notation))
- Enable/modify some unit tests and e2e tests.

This PR is the last piece to enable TKE. The following prerequisite PRs have been already merged by Petar and Andrus:
- https://github.com/logos-blockchain/logos-blockchain/pull/3080
- https://github.com/logos-blockchain/logos-blockchain/pull/3104
- https://github.com/logos-blockchain/logos-blockchain/pull/3083
- https://github.com/logos-blockchain/logos-blockchain/pull/3109

## 2. Does the code have enough context to be clearly understood?

Yes

## 3. Who are the specification authors and who is accountable for this PR?

@youngjoon-lee 

## 4. Is the specification accurate and complete?

yes

## 5. Does the implementation introduce changes in the specification?

no

## Checklist

> [!WARNING]  
> Do not merge the PR if any of the following is missing:

* [x] 1. The PR title follows the Conventional Commits [specification](https://www.notion.so/How-to-open-a-Pull-Request-215261aa09df80deb538ed59f5e4e0b5).
* [x] 2. Description added.
* [x] 3. Context and links to Specification document(s) added.
* [x] 4. Main contact(s) (developers and specification authors) added
* [x] 5. Implementation and Specification are 100% in sync including changes. This is critical.
* [x] 6. Link PR to a specific milestone.
* [x] 7. New or changed logs were reviewed against the [logging guidelines](https://github.com/logos-blockchain/logos-blockchain/blob/master/CONTRIBUTING.md#logging-guidelines).

PR #3080, opened 2026-07-03
feat(wallet): add pending-note reservation

andrussal:

This PR makes the wallet service track which notes it has already handed out as funding, so concurrent transactions no longer compete for the same UTXO.

Until now the wallet selected funding notes from confirmed UTXOs with no tracking of in-flight transactions. Concurrent requests funded from the same largest note and produced conflicting transactions, of which only one could be included. Zero genesis gas prices masked the problem; concurrent_leader_claims exposed it once fees became non-zero ([#3070](https://github.com/logos-blockchain/logos-blockchain/issues/3070)).

Wallet now keeps an ephemeral set of reserved note ids and excludes them from subsequent funding A reservation is released when the note is observed spent in a processed block, when build or signing fails, or after pending_note_expiry_blocks of LIB progress (default 10, configurable; serde-defaulted so existing configs are unaffected). 

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/services/wallet/src/lib.rs
+++ b/services/wallet/src/lib.rs
@@ -202,6 +202,7 @@ pub struct UtxoWithKeyId {
 struct LeaderClaimTx {
     signed_tx: SignedMantleTx,
     voucher_nullifier: VoucherNullifier,
+    funded_notes: Vec<NoteId>,
 }
 
 struct LeaderClaimTxRequest {
@@ -244,6 +245,17 @@ pub struct WalletServiceSettings {
     pub known_keys: HashMap<KeyId, ZkPublicKey>,
     pub voucher_master_key_id: KeyId,
     pub recovery_path: PathBuf,
+    /// How much LIB progress a pending note reservation survives before being
+    /// evicted. Notes funded into in-flight transactions are excluded from
+    /// funding until they are observed spent in a block or this many immutable
+    /// blocks have passed since the reservation.
+    #[serde(default = "default_pending_note_expiry_blocks")]
+    pub pending_note_expiry_blocks: u64,
+}
+
+#[must_use]
+pub const fn default_pending_note_expiry_blocks() -> u64 {
+    10
 }
 
 impl FileBackendSettings for WalletServiceSettings {
@@ -412,7 +424,7 @@ where
         loop {
             tokio::select! {
                 Some(msg) = service_resources_handle.inbound_relay.recv() => {
-                    Self::handle_wallet_message(msg, &mut state, &voucher_master_key_id, &storage_adapter, &cryptarchia_api, &kms, &epoch_config).await;
+                    Box::pin(Self::handle_wallet_message(msg, &mut state, &voucher_master_key_id, &storage_adapter, &cryptarchia_api, &kms, &epoch_config)).await;
                 }
                 Ok(event) = new_block_receiver.recv() => {
                     Self::handle_new_block(event.block_id, &mut state, &storage_adapter, &cryptarchia_api, &epoch_config).await;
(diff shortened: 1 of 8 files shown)


services/wallet/src/lib.rs
```diff
@@ -244,6 +245,17 @@ pub struct WalletServiceSettings {
     pub known_keys: HashMap<KeyId, ZkPublicKey>,
     pub voucher_master_key_id: KeyId,
     pub recovery_path: PathBuf,
+    /// How much LIB progress a pending note reservation survives before being
+    /// evicted. Notes funded into in-flight transactions are excluded from
+    /// funding until they are observed spent in a block or this many immutable
+    /// blocks have passed since the reservation.
+    #[serde(default = "default_pending_note_expiry_blocks")]
+    pub pending_note_expiry_blocks: u64,
```
youngjoon-lee, line 253: I have a question. It seems that this PR is taking a different approach from the pending voucher tracking: https://github.com/logos-blockchain/logos-blockchain/pull/2923#discussion_r3440032626. 

For pending vouchers, we wait `k` blocks before expiring a pending voucher. But, in this PR, we wait this configured number of blocks (or until the tx is included in the chain). 

Could you please explain how you end up with this different approach?

youngjoon-lee: I guess it's because of the difference how vouchers and notes are managed inside `Wallet`? 

Notes are managed inside `WalletState`, and the Wallet holds`wallet_states: BTreeMap<HeaderId, WalletState>` to handle reorgs. New `WalletState` is created once a new block is observed, and this new state doesn't contain spent notes in the block.
On the other hand, vouchers are not managed in per-block way. They're stored inside `Wallet` as a flat structure. Vouchers are removed from the wallet only when they're observed in immutable part of the chain.

I think this is the reason why we manage pending vouchers and notes in different ways. Could you confirm my understanding?

andrussal: Yes, I agree with this. Regular spends update the wallet state at first inclusion, while vouchers update it only at LIB. But mostly this is about user experience: if we released vouchers at first inclusion, the wallet state would still be the same, so when the user claims vouchers again they'd just fail to claim.

There's also the expiry in addition to inclusion, and now I think the expiry can actually be the same for spends and vouchers — waiting the full k blocks to retry a lost claim isn't really necessary. That part might be worth changing for vouchers, maybe in a separate PR.

youngjoon-lee: I understand the rationale. Thank you.

youngjoon-lee (approved):
the approach looks good, imo.

davidrusu:
We will need some follow up work to introduce rate-limiting in the faucet here, otherwise you can spam the the faucet with this change and receive quite a lot of stake very rapidly and threaten consensus.

andrussal:
> We will need some follow up work to introduce rate-limiting in the faucet here, otherwise you can spam the the faucet with this change and receive quite a lot of stake very rapidly and threaten consensus.

@davidrusu please take a look at https://github.com/logos-blockchain/logos-blockchain/pull/3092

davidrusu:
> > We will need some follow up work to introduce rate-limiting in the faucet here, otherwise you can spam the the faucet with this change and receive quite a lot of stake very rapidly and threaten consensus.
> 
> @davidrusu please take a look at #3092

Ah perfect, thanks!

services/wallet/src/lib.rs
youngjoon-lee: FYI: we can simplify managing pending vouchers, though it's not directly related to this PR.
https://discord.com/channels/973324189794697286/1523906458075402401/1523909144241897492

danielSanchezQ (approved):
Nice stuff @andrussal , thanks!

PR #3081, opened 2026-07-03
feat(mantle): extract MantleTxContext from MantleTxBuilder

pradovic:

This PR extracts MantleTxContext from MantleTxBuilder to enable node wallet to fetch the current context when funding the tx and give accurate gas prices. 

This is also the first PR towards implementing TKE in zone-sdk. It enables exposing the needed HTTP API fund call that will be used in integrations scenarios.

Yes

spec. @davidrusu 

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


nodes/node/binary/src/api/handlers.rs
```diff
@@ -991,7 +991,7 @@ where
             )
             .await?;
 
-        let tx_fee = funded_tx_builder.gas_cost::<MainnetGasConstants>()?;
+        let tx_fee = funded_tx_builder.gas_cost::<MainnetGasConstants>(&tx_context)?;
```
davidrusu: The tx fee is the net balance of the mantle tx, we don't need the context anymore to compute the fee

```suggestion
        let tx_fee = funded_tx_builder.net_balance();
```

You need to also to a (checked) convertion of the the i126 balance to GastCost

nodes/node/binary/src/generic_services/sdp/wallet.rs
```diff
@@ -45,7 +49,7 @@ where
             .map_err(|e| SdpWalletError::WalletApi(e.into()))?
             .response;
 
-        let tx_fee = funded.gas_cost::<MainnetGasConstants>()?;
+        let tx_fee = funded.gas_cost::<MainnetGasConstants>(&context)?;
```
davidrusu: same here, we don't need context to compute tx fee, it's `funded.net_balance()`

nodes/node/binary/src/generic_services/sdp/wallet.rs
```diff
@@ -78,7 +83,7 @@ where
...
```
davidrusu: ditto

nodes/node/binary/src/generic_services/sdp/wallet.rs
```diff
@@ -111,7 +117,7 @@ where
...
```
davidrusu: here too

services/wallet/src/lib.rs
```diff
@@ -1176,21 +1188,22 @@ where
...
         let funded_tx_builder = state.wallet().fund_tx::<MainnetGasConstants>(
             request.tip,
             &tx_builder,
             request.funding_pk,
             [request.funding_pk],
+            &context,
         )?;
 
-        let tx_fee = funded_tx_builder.gas_cost::<MainnetGasConstants>()?;
+        let tx_fee = funded_tx_builder.gas_cost::<MainnetGasConstants>(&context)?;
```
davidrusu: this should be `funded_tx_builder.net_balance()`

The paid fee may be higher than the gas cost (in the case where we add a priority tip, this is not yet implemented)

Still worth logging the gas_cost seperately in the debug log below, but the `tx_fee > request.max_tx_fee` should be on the `net_balance()`, not `gas_cost()`

davidrusu (approved):
Just need to switch the `tx_fee = gas_cost()` to `tx_fee = net_balance()` semantics. 

This change is also more correct, `gas_cost() == fee` semantics is not true when priority fee is introduced.

nodes/node/binary/src/generic_services/sdp/wallet.rs
```diff
@@ -15,6 +15,15 @@ where
...
 }
 
+/// The paid fee is the tx's net balance (inputs minus outputs); it can exceed
+/// the gas cost once priority tips are introduced.
+fn tx_fee(funded: &MantleTxBuilder) -> Result<GasCost, SdpWalletError> {
+    let net_balance = funded.net_balance();
+    u64::try_from(net_balance)
+        .map(GasCost::new)
+        .map_err(|_| SdpWalletError::NegativeNetBalance(net_balance))
+}
```
danielSanchezQ: Shouldn't this be a method of the builder which returns an error that we can embedded later in the SdpWalletError instead?
I see we are using this pattern in different places.

danielSanchezQ (approved):
Nice stuff, thanks! I left a comment which I think it would be better to read, otherwise is good to go.

youngjoon-lee (commented):
looks good!

youngjoon-lee (approved):

nodes/node/binary/src/generic_services/sdp/wallet.rs
pradovic: Very nice catch, I missed this! Will quickly refactor it before merging.

nodes/node/binary/src/generic_services/sdp/wallet.rs
```diff
@@ -45,7 +45,7 @@ where
             .map_err(|e| SdpWalletError::WalletApi(e.into()))?
             .response;
 
-        let tx_fee = funded.gas_cost::<MainnetGasConstants>()?;
+        let tx_fee = funded.tx_fee()?;
```
hansieodendaal, line 53: The call to `.fund_tx(...)` returns a `TipInfoResponse` that includes `TipResponse::tip`.
The following `.sign_tx(...)` is called without using this tip.

 If the chain tip advances between funding and signing, the transaction may be funded against one ledger/gas context but signed against another. 

We can call `sign_tx(Some(tip), funded)` instead of `.sign_tx(None, funded)`.

Also applicable to `async fn withdraw_tx(` and `async fn active_tx(` below.

hansieodendaal (commented):
Looking good - I just picked up one thing to consider for and edge case correctness.

nodes/node/binary/src/generic_services/sdp/wallet.rs
pradovic: Good catch, it was existing before but fixing in this PR as it is small change.

core/src/mantle/transactions/builder.rs
```diff
@@ -209,6 +211,16 @@ impl MantleTxBuilder {
...
 
+    /// The fee this transaction actually pays: its net balance (inputs minus
+    /// outputs). This can exceed the raw gas cost once priority tips are
+    /// introduced. Only meaningful once the builder is funded/balanced.
+    pub fn tx_fee(&self) -> Result<GasCost, TxBuilderError> {
+        let net_balance = self.net_balance();
+        u64::try_from(net_balance)
+            .map(GasCost::new)
+            .map_err(|_| TxBuilderError::NegativeNetBalance { net_balance })
+    }
```
danielSanchezQ, line 226: Also, we should probably add this check if its missing from the build. We should not allow to build a negative balanced tx. Just double checking as it may be missing from the older implementation.

core/src/mantle/transactions/builder.rs
danielSanchezQ, line 278: Yes, its missing! lets add this just before returning Ok:

```rust
let _ = self.tx_fee()?;
Ok(self.mantle_tx)
```
so we short circuit any other check

danielSanchezQ (changes requested):

core/src/mantle/transactions/builder.rs
pradovic: @danielSanchezQ  gas_cost() calls build() internally (builder.rs:228), and fund_tx works by repeatedly computing funding_delta on deliberately underfunded builders. It needs to see a negative delta to know it must pull more UTXOs. If we put self.tx_fee()? in build() and every one of those probes returns Err(NegativeNetBalance) instead of Ok(-N): fund_tx errors on anything that needs funding, transfer_funds breaks (starts as outputs-only), and the core all_ops test (asserts funding_delta == -40) fails. Two test-framework helpers also  build unbalanced builders for inspection.

Maybe I could extract build_unchecked for this purpose?

danielSanchezQ: >  gas_cost() calls build() internally (builder.rs:228)

for this case we could have an unchecked version indeed

> fund_tx works by repeatedly computing funding_delta on deliberately underfunded builders. It needs to see a negative delta to know it must pull more UTXOs

this check could be decoupled completely, we could add a method to be used instead of build.

But calls to build should avoid building an **invalid transaction**

pradovic: Does this makes sense?

https://github.com/logos-blockchain/logos-blockchain/pull/3081/changes/b1a369aad948f1ba794d70f4fcbdff6e4ffefb0c

nodes/node/binary/src/generic_services/sdp/wallet.rs
pradovic: https://github.com/logos-blockchain/logos-blockchain/pull/3081/changes/b1a369aad948f1ba794d70f4fcbdff6e4ffefb0c

hansieodendaal: These are looking good form my side, thanks!

core/src/mantle/transactions/builder.rs
pradovic: Hmm, actually there is a bug like this, will commit the fix asap

pradovic: Please check this commit as well: https://github.com/logos-blockchain/logos-blockchain/pull/3081/changes/3d075c2ead0f7e63b168ae9ac427bd0b913a2407.

The build() check false-negatived on the >32-input coin join integration test: chunked funding pushes Transfer ops whose NoteIds the builder can't price, so net_balance() only sees part of the tx. 

Fix: check runs only when no raw Transfer ops are pushed -> whenever the builder can prove the balance; the rest defers to ledger validation. All production paths (I think) still get the full check (node wallet never chunks). Becomes unconditional once the builder supports multiple managed transfers (existing TODO, if I understand it correctly).

Do you think the fix is OK or too fuzzy? The build will validate now if it can.

davidrusu: this feels wrong. I think this is pointing something is wrong with this builder.

I think the `push_op` and `extend_ops` API is too loose. 

Instead we probably need individual builder functions for each op:
- `fn push_transfer(inputs: Vec<Utxo>, outputs: Note)`
- `fn push_channel_config(...)`

But this feels like it should be done in a follow up PR. @danielSanchezQ wdyt of dropping this build time validation until we improve the builder API to prevent invalid transactions from being built?

pradovic: I agree this indeed sounds like a better approach. We can remove the build time validation, or leave it like this as a transitional step before the follow up PR. @danielSanchezQ Let me know what you think

danielSanchezQ: > this feels wrong. I think this is pointing something is wrong with this builder.
> 
> I think the `push_op` and `extend_ops` API is too loose.
> 
> Instead we probably need individual builder functions for each op:
> 
> * `fn push_transfer(inputs: Vec<Utxo>, outputs: Note)`
> * `fn push_channel_config(...)`
> 
> But this feels like it should be done in a follow up PR. @danielSanchezQ wdyt of dropping this build time validation until we improve the builder API to prevent invalid transactions from being built?

I think the issue is that we do not perform any kind of validation when pushing the ops. Which needs to be fixed and can be done in a different pr indeed. Adding a method for each case seems a bit overload, but can be considered too.

pradovic: Ok, reverting the validation, and let's fix it in the follow up PR, thanks!

danielSanchezQ (approved):

PR #3082, opened 2026-07-03
chore: migrate CryptarchiaParameter to NomEncode and NomDecode trait

ntn-x2:

Built on top of the [proofs refactoring](https://github.com/logos-blockchain/logos-blockchain/pull/3078) to avoid unnecessary conflicts, but the two PRs are logically independent of each other.

Also I questioned whether it makes sense to force cast a timestamp into a `u64`, but I left the logic unchanged for the time being. I only constrained the maximum size the chain ID string can have, because it was previously unbounded. So I wrote a small `BoundedString` wrapper to enforce min and max bounds on a valid UTF-8 string. For now it lives next to the `ChainId` type, but it could be moved elsewhere if we want to include strings in some ops, for some reason.

Yes.

@ntn-x2 

Yes.

No.


core/src/mantle/nom/time.rs
```diff
@@ -0,0 +1,28 @@
...
+impl NomEncode for OffsetDateTime {
+    fn encode(&self) -> Vec<u8> {
+        u64::try_from(self.unix_timestamp())
+            .expect("timestamp fits in u64")
+            .encode()
+    }
+}
+
+impl NomDecode for OffsetDateTime {
+    fn decode(bytes: &[u8]) -> IResult<&[u8], Self> {
```
hansieodendaal, line 18: This function could return consistent error types for `u64::decode` and `i64::try_from`.

core/src/mantle/transactions/genesis_tx.rs
```diff
@@ -266,57 +272,170 @@ impl<'de> Deserialize<'de> for GenesisTx {
...
 
-/// Cryptarchia parameters encoded as an inscription in the genesis block.
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct CryptarchiaParameter {
-    pub chain_id: String,
-    pub genesis_time: OffsetDateTime,
-    pub epoch_nonce: Fr,
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
+#[serde(try_from = "String")]
+struct BoundedString<const MIN_SIZE: usize, const MAX_SIZE: usize>(String);
```
hansieodendaal, line 280: I think `BoundedString` should live alongside "utils/src/bounded_vec.rs"

core/src/mantle/transactions/genesis_tx.rs
```diff
@@ -266,57 +272,170 @@ impl<'de> Deserialize<'de> for GenesisTx {
...
+                "Max size cannot be more than the maximum allowed byte size for a chain ID."
+            );
+        }
+        Self::try_from(value.into_inner())
+    }
+}
+
+pub const MAX_CHAIN_ID_SIZE: usize = u64::MAX as usize;
+type ChainIdBoundedVec = BoundedVec<u8, 1, MAX_CHAIN_ID_SIZE>;
+type ChainIdBoundedString = BoundedString<1, MAX_CHAIN_ID_SIZE>;
```
hansieodendaal, line 346: I think the upper bound on these are excessive and should be bound by some practical limitation like block size.

core/src/mantle/transactions/genesis_tx.rs
```diff
@@ -266,57 +272,170 @@ impl<'de> Deserialize<'de> for GenesisTx {
...
+{
+    fn as_ref(&self) -> &str {
+        &self.0
+    }
+}
+
+impl<const MIN_SIZE: usize, const MAX_SIZE: usize> TryFrom<String>
+    for BoundedString<MIN_SIZE, MAX_SIZE>
+{
+    type Error = String;
```
hansieodendaal, line 299: Why `type Error = String;` and not a defined type? Same comment elsewhere in `BoundedString`.

core/src/mantle/transactions/genesis_tx.rs
```diff
@@ -266,57 +272,170 @@ impl<'de> Deserialize<'de> for GenesisTx {
...
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        if value.len() < MIN_SIZE {
+            return Err(format!("String must be at least {MIN_SIZE} bytes: {value}"));
+        }
+        if value.len() > MAX_SIZE {
+            return Err(format!("String must not exceed {MAX_SIZE} bytes: {value}"));
```
hansieodendaal, line 306: The error messages use {value} which could expose very long strings in logs. Consider truncating long values in error messages for readability.

hansieodendaal (commented):
Nice further application of consistent NomEncode and NomDecode. I have some concerns w.r.t. the practical bounds definition of `ChainIdBoundedString` as noted below, where `BoundedString` should live and then some other lesser comments.

core/src/mantle/nom/time.rs
ntn-x2: This is handled elsewhere in https://github.com/logos-blockchain/logos-blockchain/pull/3086.

core/src/mantle/transactions/genesis_tx.rs
ntn-x2: You are right. I follwed the same approach we introduced a while ago for `BoundedMultiaddr`, but I agree it makes sense to have a `bounded` module in the utils crate. I will do that in a separate PR tho, as this one is already conflicting with Alex's refactoring work, and I'd like to merge this ASAP, if you're fine with that. I'll open a follow-up right away to move this type in the utils crate.

core/src/mantle/transactions/genesis_tx.rs
ntn-x2: Yes I agree, I initially went for calculating the maximum size as the maximum size of an inscription, minus any other info that goes into the genesis inscription. Nevertheless, for the same reason as above, let's go ahead with this and change it in a different PR, if you agree.

core/src/mantle/transactions/genesis_tx.rs
ntn-x2: I will fix this in a follow-up PR where I migrate this type in the utils crate, if that's ok.

core/src/mantle/transactions/genesis_tx.rs
ntn-x2: Same as above. I will fix this in a follow-up PR where I migrate this type in the utils crate, if that's ok.

hansieodendaal (approved):
LGTM with promises of follow-up PRs as noted.

youngjoon-lee (approved):

PR #3083, opened 2026-07-03
fix(e2e): non-zero gas price

andrussal:

This PR adds fixes for e2e tests the non-zero gas price use case.

`concurrent_leader_claims` still ignored until https://github.com/logos-blockchain/logos-blockchain/pull/3080 gets merged, `@zone_ci` red until zone-sdk funding lands

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


tests/cucumber_tests/features/transactions.feature
```diff
@@ -27,7 +27,7 @@ Feature: Transactions
       | account_index | token_count | token_amount |
       | 1             | 2           | 1000         |
       | 2             | 0           | 0            |
-    And we have a sponsored genesis fee account with 2 tokens of 997 value each
+    And we have a sponsored genesis fee account with 2 tokens of 20000 value each
```
hansieodendaal, line 30: With exact fee accounting becoming important now, I think we should add a fee assertion step to this scenario.

tests/cucumber_tests/features/transactions.feature
```diff
@@ -50,7 +50,7 @@ Feature: Transactions
...
```
hansieodendaal, line 53: ditto fee assertion step

tests/cucumber_tests/features/transactions.feature
```diff
@@ -70,7 +70,7 @@ Feature: Transactions
   Scenario: Many nodes with wallets startup
     Given the genesis block has the following wallet resources:
       | account_index | token_count | token_amount |
-      | 1             | 2           | 1000         |
+      | 1             | 2           | 2000         |
```
hansieodendaal, line 73: Will it be possible to add an assertion for the exact value of the two change outputs? Would that be consistent going forward? If the fee structure change, we can then update the expected values.

tests/src/tests/mantle/sdp/ops.rs
```diff
@@ -106,32 +101,32 @@ async fn sdp_ops_e2e() {
     };
     let declaration_id = declaration.id();
 
-    let declare_mantle_tx = fund_sdp_transaction(
+    let (declare_tx, _declare_fee) = funded_signed_tx(
```
hansieodendaal, line 104: Wdyt about collecting and using the `_declare_fee`, `_withdraw_fee`, etc. with the option to do some assertions on it? Just an idea.

hansieodendaal (commented):
Looking good.

We should probably rebase on top of #3083.

Some questions/suggestions below.

youngjoon-lee (approved):
great

tests/cucumber_tests/features/transactions.feature
andrussal: > With exact fee accounting becoming important now, I think we should add a fee assertion step to this scenario.

I think in general this is a good idea, but what do you think of keeping fee assertions in dedicated fee tests? That way there's a single place to reason about and maintain them — if the fee structure changes, a few clearly-named fee scenarios break as a signal, rather than fee literals in every transfer scenario needing a retune.

tests/src/tests/mantle/sdp/ops.rs
andrussal: > Wdyt about collecting and using the `_declare_fee`, `_withdraw_fee`, etc. with the option to do some assertions on it? Just an idea.

mh, I think asserting the fee here would just test the builder's gas calculation, and this is an arbitrary location where we happen to use the builder — same code path as every other call site. I think this is better covered by a focused builder test, wdyt?

andrussal:
> We should probably rebase on top of #3083.

Good idea — assuming you mean [#3070](https://github.com/logos-blockchain/logos-blockchain/pull/3070) :) I'll rebase on top of it at some later point, once that branch is up to date with master. My branch is already on latest master.

tests/cucumber_tests/features/transactions.feature
hansieodendaal: Good suggestion - it could be a consecutive PR

tests/src/tests/mantle/sdp/ops.rs
hansieodendaal: Yes, that is better

hansieodendaal (approved):
LGTM!

PR #3086, opened 2026-07-05
feat(mantle): change genesis_time from OffsetDateTime to u32

youngjoon-lee:

Now https://github.com/logos-co/logos-lips/pull/370 clearly defines that `genesis_time` should be `u32`. This PR implements it by defining a concrete type `GenesisTime(u32)` which implements `NomCodec`.

yes

@youngjoon-lee 
spec: @davidrusu 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.


core/src/mantle/codec/primitives.rs
```diff
@@ -49,12 +46,11 @@ mod encoding {
         s.as_bytes().to_vec()
     }
 
-    pub fn encode_unix_timestamp(ts: &OffsetDateTime) -> Vec<u8> {
-        encode_uint64(
-            ts.unix_timestamp()
-                .try_into()
-                .expect("timestamp fits in u64"),
-        )
+    pub fn encode_u32_unix_timestamp(ts: &OffsetDateTime) -> Vec<u8> {
```
ntn-x2: We're implementing `NomEncode` and `NomDecode` for all types involved in wire encoding/decoding, including the `CryptarchiaParameter` type in https://github.com/logos-blockchain/logos-blockchain/pull/3082. I suggest one of two things: if we are specifying logic specific for the genesis timestamp, I would suggest we create a newtype around the `OffsetDateTime` and implement `NomEncode` and `NomDecode` on that, so that this particular trimming, potentially panicking logic, is limited to this type. If we are to provide a generic implementation for encoding `OffsetDateTime` that could appear elsewhere, e.g., in ops for some reason, then we need a non-panicking implementation. But we can go with option 1 for now.

youngjoon-lee: I will rebase this branch on top of #3082 and implement a new type only for the genesis timestamp.

ntn-x2: Thanks! In that case, please remove the current implementation on the `OffsetDateTime` itself as it's flawed and not needed if we have a newtype for the genesis start time which bakes in genesis assumptions, thank you!

youngjoon-lee: @ntn-x2 Thank you. I rebased/improved this PR. I hope this aligns well with our convention.

danielSanchezQ (approved):
I just have one little concern. Does this change config file representation too?

core/src/mantle/transactions/genesis_tx.rs
```diff
@@ -443,11 +443,39 @@ impl NomDecode for ChainId {
     }
 }
 
+/// Time at which the chain should start. Bounded by u32 since a far-future
```
ntn-x2: Technically we use a `u32` because timestamps are represented as `i64`, and we only care about the positive part, right? So we would still be able to represent the maximum possible value for a timestamp.

ntn-x2 (approved):
Very good, thank you very much!

youngjoon-lee:
> I just have one little concern. Does this change config file representation too?

No. Config files are unchanged. The ceremony template `inscribe.yaml` still expects an ISO8601 string. The deployment config is unchanged because genesis_tx is encoded as bytes. 

EDIT: Ah, now I think I should update the `inscription` bytes in `nodes/node/standalone-deployment-config.yaml`. I will update it now. 

Updated 1ddea31c2

core/src/mantle/transactions/genesis_tx.rs
youngjoon-lee: ah, ah, right! I fixed the comment. [3f29f29](https://github.com/logos-blockchain/logos-blockchain/pull/3086/commits/3f29f29d0c7d0c60ef555888ea1e714f6f29b853)

core/src/mantle/codec/primitives.rs
```diff
@@ -103,10 +99,11 @@ mod tests {
         assert_eq!(value, str);
         assert!(remaining.is_empty());
 
-        // Test Unix Timestamp
+        // Test Unix Timestamp (spec: u32 LE, 4 bytes)
```
hansieodendaal: We could possibly add some targeted tests here. The current roundtrip test confirms the happy path and the 4-byte length, but it would be useful to also assert the exact byte sequence for a known timestamp and cover the `u32` conversion limits.

For example:
- a fixed timestamp like `1000` encodes to `[0xE8, 0x03, 0x00, 0x00]`
- `u32::MAX` roundtrips successfully
- timestamps outside the `u32` range panic or fail as expected

That would make the intended wire format and boundary behavior explicit.

hansieodendaal (commented):
I have just one suggestion to add, apart from the existing comments.

core/src/mantle/codec/primitives.rs
youngjoon-lee: Thank you. Added tests 321798440.

PR #3090, opened 2026-07-06
chore(deployment): Persistent grafana and datasources volumes

bacv:

Persist data between deployments. Each service has it's own volume incase we'll want to cleanup prometheus and loki but not grafana user defined contents.

Yes

@bacv

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/deployment/compose.tracing.yml
+++ b/deployment/compose.tracing.yml
@@ -15,6 +15,7 @@ services:
       - ./tracing/grafana/datasources.yaml:/etc/grafana/provisioning/datasources/datasources.yaml:z
       - ./tracing/grafana/dashboards.yml:/etc/grafana/provisioning/dashboards/dashboards.yml:z
       - ./tracing/grafana/dashboards:/var/lib/grafana/dashboards/:z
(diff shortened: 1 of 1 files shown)


ntn-x2 (approved):
Nice. What if we actually want to delete state between runs? Should it be included in the setup compose somehow?

bacv:
> What if we actually want to delete state between runs?

We could, but ideally such maintenance should be decoupled from the deployment. If we see that we don't need data from previous runs, we could just remove volumes for prometheus and loki, this way we'd always get data from the latest run only.

hansieodendaal (approved):
LGTM

ntn-x2:
Well, it would be one step before a new deployment, right? The same way we clean blockchain state, we would want to clean metrics. Unless we explicitly keep them for analysis, I guess.

bacv:
> Well, it would be one step before a new deployment, right?

You know what, I'll remove prometheus and loki volumes, we'll only persist grafana customization. For logs we have a file access still that we could download before deploying a new version.

PR #3092, opened 2026-07-06
feat(faucet): serialize drips through a queue and add per-key cooldown

andrussal:

Fixes the faucet silently dropping concurrent requests while returning success.

**Problem**

- All concurrent requests build transactions spending the faucet's single note, so only one can land per block.
- Every caller still gets a 200 with a tx hash — but the losing transactions are dropped at block inclusion, so those hashes never appear on chain and the users never receive funds or an error.

**Change**

- Requests go into a queue; a worker pays them out one per block, in order. Callers get an immediate `202 {"status":"queued"}` and funds arrive a few blocks later.
- Repeat requests for the same key within the cooldown (`--cooldown-secs`, default 5 min) get `429` with the remaining wait.
- The response no longer contains a tx hash — the transaction is built asynchronously.
- Also fixes a debug-build startup panic (duplicate `-d` short flag).

**Relation to #3080**

That PR makes the wallet reserve notes for in-flight transactions, which turns the silent drops into "insufficient funds" errors from a fully funded faucet. With the queue, requests wait their turn instead.

Yes

@andrussal

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/deployment/faucet/src/server.rs
+++ b/deployment/faucet/src/server.rs
@@ -1,4 +1,8 @@
-use std::sync::Arc;
+use std::{
+    collections::HashMap,
+    sync::{Arc, Mutex},
+    time::{Duration, Instant},
+};
 
 use axum::{
     Router,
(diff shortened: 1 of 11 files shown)


deployment/faucet/src/server.rs
```diff
@@ -9,11 +13,53 @@ use axum::{
...
+            .expect("cooldown mutex should not be poisoned");
+        let now = Instant::now();
+        cooldowns.retain(|_, expires_at| *expires_at > now);
+
+        if let Some(expires_at) = cooldowns.get(key) {
+            return Err(expires_at.duration_since(now));
+        }
+
+        cooldowns.insert(key.to_owned(), now + self.cooldown);
+        drop(cooldowns);
```
bacv, line 49: Was this clippy?

bacv (approved):

deployment/faucet/src/server.rs
andrussal: > Was this clippy?

Yep, looked this up: `has_significant_drop` - https://rust-lang.github.io/rust-clippy/master/#significant_drop_tightening

PR #3093, opened 2026-07-06
chore(utils): introduce a Bounded wrapper and type aliases for String and Multiaddr

ntn-x2:

We had very similar logic for `BoundedMultiaddr` and `BoundedString`. So there was room for improvement. We introduce a `Bounded<T, MIN, MAX>` type which wraps around a type `T` and enforces lower and upper bounds. Then we expose aliases for `BoundedMultiaddr` and `BoundedString`, with also some alias-specific implementations.

Unrelated, but still part of the unaddressed comments on the review of https://github.com/logos-blockchain/logos-blockchain/pull/3082, we change the maximum length of chain ID to be `u8::MAX` bytes long, at most, by using the newly-introduced `BoundedString`. Using an alias to the `Bounded` type also made the error case returned an enum and not a `String` anymore.

Yes.

@ntn-x2 

N/A

No.

### DIFF
--- a/core/src/sdp/mod.rs
+++ b/core/src/sdp/mod.rs
@@ -11,7 +11,7 @@ use blake2::{Blake2b, Digest as _};
 use bytes::Bytes;
 use lb_cryptarchia_engine::Epoch;
 use lb_key_management_system_keys::keys::ZkPublicKey;
-use lb_utils::bounded::{Bounded, BoundedVec, NonEmptyBoundedVec, UpperBoundedVec};
+use lb_utils::bounded::{BoundedVec, NonEmptyBoundedVec, UpperBoundedVec};
 use multiaddr::{Multiaddr, Protocol};
 use nom::{
     IResult,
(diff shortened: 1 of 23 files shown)


core/src/sdp/mod.rs
```diff
@@ -99,104 +99,62 @@ pub struct InactivityPeriodTooSmall {
...
-}
-
 pub const MAX_LOCATOR_BYTE_SIZE: usize = 329;
 
+/// A [`Multiaddr`] whose byte length is bounded to `[0,
+/// MAX_LOCATOR_BYTE_SIZE]`.
+///
+/// `Multiaddr` is foreign to `lb_utils`, so it cannot implement `BoundedLen`;
+/// the byte length is measured here and validated through the shared
+/// [`Bounded::check_len`] gate.
```
hansieodendaal:
```suggestion
/// A [`Multiaddr`] whose byte length is bounded to `[0,
/// MAX_LOCATOR_BYTE_SIZE]`.
///
/// The shared `lb_utils::bounded` wrapper enforces the byte-length invariant
/// using `Multiaddr::len()`. `Locator::try_from` performs the additional
/// locator-specific validation below, such as rejecting unspecified, loopback,
/// multicast, documentation, and link-local addresses.
```

hansieodendaal (approved):
LGTM! Nicely done.

Just one minor doc comment suggestion.

core/src/sdp/mod.rs
ntn-x2: Done in https://github.com/logos-blockchain/logos-blockchain/pull/3093/commits/902a7a042db4167ee4f220be2c394889f4d00f89.

PR #3096, opened 2026-07-07
feat(tracing): Otlp http exporter in tracing crate

bacv:

Otlp supports grpc and http to publish the tracing related data, by default we've supported grpc, now we also support http.

Yes

@bacv

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/tools/config/src/tracing.rs
+++ b/tools/config/src/tracing.rs
@@ -25,6 +25,7 @@ impl GeneralTracingConfig {
                         endpoint: LOKI_ENDPOINT.try_into().unwrap(),
                         service_name: host_identifier.clone(),
                         authorization_header: None,
+                        protocol: lb_tracing::OtlpProtocol::Grpc,
                     }),
                     stdout: true,
                     file: None,
@@ -37,6 +38,7 @@ impl GeneralTracingConfig {
                     sample_ratio: DEBUG_TRACING_SAMPLE_RATIO,
                     service_name: host_identifier.clone(),
                     authorization_header: None,
+                    protocol: lb_tracing::OtlpProtocol::Grpc,
                 }),
                 filter: tracing::filter::Layer::Env(tracing::filter::EnvConfig {
                     filters: HashMap::from([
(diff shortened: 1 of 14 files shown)


tools/config/src/tracing.rs
```diff
@@ -37,6 +38,7 @@ impl GeneralTracingConfig {
                     sample_ratio: DEBUG_TRACING_SAMPLE_RATIO,
                     service_name: host_identifier.clone(),
                     authorization_header: None,
+                    protocol: lb_tracing::OtlpProtocol::Grpc,
```
andrussal, line 41: Do we need to be backwards compatible here as well and use `#[serde(default)]`?

andrussal (commented):
I'm not sure if I missed something in the setup, but when I enable `Http` and point metrics at a dummy server, the node panics on the first export and nothing ever arrives:

```
ERROR logos_blockchain::node: A panic occurred
panic_payload="there is no reactor running, must be called from the context of a Tokio 1.x runtime"
```

Dummy collector: `python3 -m http.server 4318`

```yaml
logger:
  otlp:
    endpoint: "http://127.0.0.1:4318/v1/logs"
    service_name: "http-repro"
    protocol: Http
metrics: !Otlp
  endpoint: "http://127.0.0.1:4318/v1/metrics"
  service_name: "http-repro"
  protocol: Http
```

bacv:
> ERROR logos_blockchain::node: A panic occurred

Got it panicking too, running the standalone config panics only after a while, didn't test it properly. Fixed it with https://github.com/logos-blockchain/logos-blockchain/pull/3096/commits/3ef265517b2df807100bdfa4f06a6da750c92511, another way would be to initialize otlp layer inside the tracing service `run` method, but as I understand we'd get similar effect, since blocking client is spawned inside the thread.

tools/config/src/tracing.rs
bacv: > Do we need to be backwards compatible

By default, the `init-config` command generates otlp for logging, tracing and metrics as `None`. Let's leave it explicit?

andrussal (approved):
LGTM!

PR #3098, opened 2026-07-07
feat(deployment): Expose otlp endpoint for external nodes

bacv:

Expose otlp authenticated endpoint for external nodes to send metrics to. The allow list can be modified via web filebrowser by editing the `/node-data/tracing/otel_tokens.map` file.

Add volunteer data to the grafana.

@bacv

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/deployment/compose.run.yml
+++ b/deployment/compose.run.yml
@@ -139,6 +139,14 @@ services:
       - "${NODE3_API_PORT}:18083/tcp"
     entrypoint: /etc/logos-blockchain/scripts/run_node.sh
 
+  filebrowser:
+    container_name: filebrowser
+    image: filebrowser/filebrowser:v2-s6
+    volumes:
+      - node-data:/srv
+    command: ["--noauth", "-b", "/filebrowser"]
+    restart: on-failure
+
   nginx:
     container_name: nginx
     image: docker.io/library/nginx:stable-trixie-perl
--- a/deployment/nginx/run.conf
+++ b/deployment/nginx/run.conf
@@ -9,6 +9,11 @@ http {
     sendfile on;
     keepalive_timeout 65;
 
+    map $http_otlp_authorization $is_otlp_authorized {
+        default 0;
+        include /node-data/tracing/otel_tokens.map;
+    }
+
     # Public
     server {
         listen 13000;
@@ -43,6 +48,21 @@ http {
             proxy_pass http://faucet:18090/faucet/;
         }
 
+        # Otlp http collector for external nodes.
+        location /otlp {
+            if ($is_otlp_authorized = 0) {
+                return 401 '{"error": "Unauthorized"}\n';
+            }
+
+            proxy_pass http://otel:4318/v1/metrics;
+            proxy_set_header Host $host;
+            proxy_set_header X-Real-IP $remote_addr;
+            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+
+            # OTLP payloads can be large; prevent Nginx from blocking them.
+            client_max_body_size 20M;
+        }
+
         # Static content on https://$EVN_TITLE_STRING.blockchain.logos.co/web/.
         location / {
             root /usr/share/nginx/html;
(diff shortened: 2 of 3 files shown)


deployment/nginx/run.conf
```diff
@@ -9,6 +9,11 @@ http {
     sendfile on;
     keepalive_timeout 65;
 
+    map $http_otlp_authorization $is_otlp_authorized {
```
andrussal: This may be different from node if I understand this correctly:

```rust
if let Some(auth) = config.service.authorization_header {
        builder = builder.with_headers(HashMap::from([("authorization".to_owned(), auth)]));
    }
```

deployment/nginx/run.conf
```diff
@@ -9,6 +9,11 @@ http {
...
+        default 0;
+        include /node-data/tracing/otel_tokens.map;
```
hansieodendaal, line 14: This file is not ensured to exists before nginx starts? Maybe we could create the directory and an empty/default map file during setup or in `run_nginx.sh` before `exec nginx`, otherwise a fresh deployment can fail at nginx config load time.

deployment/scripts/run_nginx.sh
```diff
@@ -1,9 +1,19 @@
...
+
 # Container can't write to the mounted host dir.
 mkdir /usr/share/nginx/html/
 cp -r /usr/share/nginx/html_template/* /usr/share/nginx/html/
 
 envsubst '$ENV_TITLE_STRING' < /usr/share/nginx/html/index.html.template > /usr/share/nginx/html/index.html
 
+# If we update otlp auth tokens at runtime we need to reload the nginx service.
+(
+  while inotifywait -e modify /node-data/tracing/otel_tokens.map; do
```
hansieodendaal, line 16: If the file `otel_tokens.map` does not exist at startup, `inotifywait` can exit and the reload loop will stop permanently.

deployment/compose.run.yml
```diff
@@ -139,6 +139,14 @@ services:
       - "${NODE3_API_PORT}:18083/tcp"
     entrypoint: /etc/logos-blockchain/scripts/run_node.sh
 
+  filebrowser:
+    container_name: filebrowser
+    image: filebrowser/filebrowser:v2-s6
+    volumes:
+      - node-data:/srv
+    command: ["--noauth", "-b", "/filebrowser"]
```
hansieodendaal, line 147: This seems to mount all of node-data into `Filebrowser` and disables authentication with `--noauth`, and that write access is exposed to much more than the OTLP token allowlist. Maybe we need explicit access controls around this endpoint?

hansieodendaal (commented):
This is not my forte at all, so these comments may miss the mark completely and be out of context. In that case, please "beer" with me -:).

deployment/nginx/run.conf
bacv: You are correct, fixed in https://github.com/logos-blockchain/logos-blockchain/commit/a74c7eeb52ab288f628e96d334cdc56c88765a4d

deployment/compose.run.yml
bacv: > be we need explicit access controls around this endpoint?

We actually rely on the same mechanism that that protects the `/node-data` endpoint, provided by infra. I think that's enough for this use case.

The main idea here is that all http services are only accessible via ngnix, and nginx runs on two ports: 13000 and 13001, one is public, another is protected by the basic auth:
https://github.com/logos-blockchain/logos-blockchain/pull/3098/changes#diff-395533430e68efce5bb98c47ee7329df9b482bcbbb0cbc2ab71ca8d89b8f0ff3L54-R89

Basic auth is managed by infra team.

deployment/nginx/run.conf
bacv: > This file is not ensured to exists before nginx starts?

Good question, I think this falls under the similar category as our `user_config.yaml` for init nodes, updated related deployment docs in notion.

deployment/scripts/run_nginx.sh
bacv: > If the file otel_tokens.map does not exist at startup

Just for you https://github.com/logos-blockchain/logos-blockchain/pull/3098/commits/ad179cf97a9872cb62e4e12cb4520a392302fd25

andrussal (approved):

hansieodendaal (approved):
LGTM!

PR #3101, opened 2026-07-07
test: add new atomic wallet scanner

hansieodendaal:

Added a new wallet scanner that replaces the old wallet UTXO feed and transaction hash tracking paths in cucumber tests. It moves wallet tracking from an on-demand/feed collector model to a background scanner per fork group, which selects a majority-backed best node, streams blocks in order, and updates wallet UTXO state and observed transaction hashes atomically as blocks are processed.

**Key changes:**

- Added `tests/src/common/wallet/scanner` with accounting, block streaming, group config, runtime state, and scanner task orchestration.
- Removed the old wallet feed/backfill implementation.
- Reworked wallet sync, balance checks, transaction submission, and snapshot handling to consume scanner-backed tracked wallet state.
- Added scanner catch-up/runtime wiring to cucumber world setup and node startup flows.
- Improved best-node selection for scanner use across fork groups.
- Added a funding-wallet transaction scenario to `transactions.feature`.

This gives cucumber scenarios a single chain-aligned wallet state view, reducing stale per-node wallet state issues during stress, fork, and snapshot flows.

**Reason for the change:**

Devnet-linked stress test scenarios were too complex to manage and made performant, as increased chain height deteriorated wallet and transaction feed scanning.

**Validation**

Performed a 6 concurrent cucumber test run for all CI cucumber tests:
```
cargo test -p logos-blockchain-tests --features cucumber --test cucumber -- --tags "@cli_ci or @blend_ci or @cryptarchia_ci or @mempool_ci or @smoke_ci or @transactions_ci or @zone_ci"
```
```
[Summary]
9 features
55 scenarios (55 passed)
717 steps (717 passed)
```

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


hansieodendaal (commented):
CoPilot review:

**Blocking or high-value review points**

- `tests/src/cucumber/world.rs#L965-L974`: `CucumberWorld::drop` does not stop `wallet_scanner_runtime`. If a scenario fails before reaching an explicit “stop all nodes” step, the scanner tasks can keep running after the world is dropped because they hold cloned Arcs and node clients. Please abort the scanner runtime in `Drop`, e.g. `if let Some(runtime) = self.wallet_scanner_runtime.take() { runtime.cancel(); }`, similar to the faucet task cleanup.
- `tests/src/cucumber/world.rs#L1364-L1378`: `ensure_wallet_scanner_started` returns immediately once a scanner runtime exists, but scanner configs are built only from the nodes/wallets present at first startup. The old feed path supported later source registration; with this code, scenarios that add nodes or wallets after the scanner has started will not be tracked by the existing scanner. Please either restart/reconfigure the scanner when `nodes_info`/`wallet_info` changes, or document/assert that scanner startup only happens after all wallet-bearing nodes are registered.
- `tests/src/common/wallet/scanner/runner.rs#L146-L164`: `wait_for_scanner_catch_up` only checks `target_height` vs `applied_height`, so a group with `status = Error` can still be considered caught up if its last target was already reached. That can let wallet reads proceed on stale/erroring scanner state. Consider treating `ScannerStatus::Error`/`last_error` as a wait condition or immediate diagnostic failure unless a subsequent iteration clears it.

**Possible Improvements**

- `tests/src/common/wallet/scanner/runner.rs#L603-L609`: each scanner iteration extends the shared observed hash set with the scanner’s entire historical `observed_transaction_hashes` set. This is correct but can become increasingly expensive on long stress/devnet-linked runs. Consider tracking the newly observed hashes per iteration and extending the shared set only with the delta.

hansieodendaal:
> CoPilot review:
> 
> Please either restart/reconfigure the scanner when nodes_info/wallet_info changes, or document/assert that scanner startup only happens after all wallet-bearing nodes are registered.

Kept this behaviour as is; it will only influence long running devnet-linked scenarios, and will be tested in due course and PR'd when required.

tests/src/common/wallet/scanner/runner.rs
```diff
@@ -0,0 +1,870 @@
...
+        rescan.tip,
+        streamed_block_count,
+        slot_from,
+        rescan.slot,
+    );
+
+    Ok(())
+}
+
+fn push_checkpoint(checkpoints: &mut VecDeque<ScannerCheckpoint>, checkpoint: ScannerCheckpoint) {
```
andrussal, line 815: I am wondering if this should be deeper, otherwise a reorg deeper than the window has nothing to roll back to and the scanner can't recover.

tests/src/common/wallet/tracked_wallet.rs
```diff
@@ -127,6 +127,8 @@ impl TrackedWallet {
         self.utxos_by_note.clear();
         self.utxos_by_note
             .extend(utxos.into_iter().map(|utxo| (utxo.id(), utxo)));
+        self.pending_state
```
andrussal, line 130: When this is called because of a rollback with older state, it forgets pending notes that came into existence later. Probably worth considering if this can break things.

tests/src/common/wallet/scanner/group.rs
```diff
@@ -0,0 +1,256 @@
...
+            height: merged_height,
+            slot: merged_slot,
+            source_node_names: merged_source_node_names,
+            rescan_blocks: merged_rescan_blocks,
+        } = seed
+        else {
+            unreachable!("merged seed is always a snapshot");
+        };
+
+        if *merged_tip != tip || *merged_height != height || *merged_slot != slot {
```
andrussal, line 257: I am wondering if this is robust given that each node snapshot is saved while nodes are still running?

tests/src/common/wallet/scanner/runner.rs
```diff
@@ -0,0 +1,870 @@
...
+    pub fn cancel(self) {
+        for handle in self.handles {
+            handle.abort();
+        }
+    }
+
+    /// Graceful cancellation. Lets the current scanner iteration complete,
+    /// then stops before starting the next iteration.
+    pub async fn shutdown_after_current_iteration(self) {
+        self.shutdown_requested.store(true, Ordering::Relaxed);
```
andrussal, line 108: Do all active code paths actually check this flag?

tests/src/common/wallet/scanner/group.rs
```diff
@@ -0,0 +1,256 @@
...
+
+    for wallet in world.wallet_info.values() {
+        let group_id = world
+            .node_to_group
+            .get(&wallet.node_name)
+            .cloned()
+            .unwrap_or_default();
+        let entry = groups
+            .entry(group_id)
+            .or_insert_with(|| (Vec::new(), Vec::new(), BTreeMap::new()));
```
andrussal, line 90: Should we just error if no group? Probably invalid scenario.

tests/src/common/wallet/scanner/accounting.rs
```diff
@@ -0,0 +1,440 @@
...
+        &self.observed_transaction_hashes
+    }
+
+    #[must_use]
+    /// Return the number of wallets tracked by this accounting instance.
+    pub const fn tracked_wallet_count(&self) -> usize {
+        self.tracked_wallets.len()
+    }
+
+    fn apply_transaction(&mut self, tx: &SignedMantleTx) {
```
andrussal, line 138: It might be better to reuse logic from `chain/state.rs`.

andrussal (commented):
Looks very nice!

I like the new scanner approach. Managing this state is quite complex, and fully dedicated logic seems to work better here than the general purpose feed.

(Though I still think the general purpose feed collector we have can be useful for other purposes, it is a unified way of collecting chain data, so it might help avoid duplicate work in some cases.)

tests/src/common/wallet/scanner/accounting.rs
hansieodendaal: This is conceptually right, but I do not think it is worth doing in this PR. Scanner accounting and chain/state.rs are now near-duplicate logic. Reuse would reduce drift, but it touches the accounting core after the tests are passing.

However, I will give it another round to see if we can reduce the technical debt, or defer to a follow-up PR.

tests/src/common/wallet/scanner/group.rs
hansieodendaal: Good observation, I will make this case give better feedback.

tests/src/common/wallet/scanner/group.rs
hansieodendaal: Good catch, although the complex cucumber wallet restore test passed, `pub async fn prepare_wallet_snapshot_from_active_node_tips(` has the wrong shape and is working against the new wallet scanner, as it was waiting for wallert scanner catch up from all groups that use best node selection as filter, but directly afterwards calling cryptarchia info for each node again and pinning wallet state to a potential forked or lagging node.

Fixing this.

tests/src/common/wallet/scanner/runner.rs
hansieodendaal: Yes, both "I stop all k8s manual nodes" and "I stop all nodes" steps call this path

hansieodendaal: It is not strictly required as the tokio tasks will be shutdown eventually, but makes for better warn-free logs when scenarios ends.

tests/src/common/wallet/scanner/runner.rs
hansieodendaal: Good point, increasing that depth dynamically from 8 to 128 with increasing chain height and adding safe fallback to scan from genesis again if a suitable snapshot is not available.

tests/src/common/wallet/tracked_wallet.rs
hansieodendaal: Good catch. I tuned this behavior against the long-running local continuous round-robin and next-wallet scenarios, and current CI coverage is passing with that setup.
That said, rollback-to-older-state can prune pending notes that only existed on the newer branch, so I will keep monitoring this in longer-running local runs and devnet-linked scenarios and revisit if I see regressions.

andrussal: > Good catch. I tuned this behavior against the long-running local continuous round-robin and next-wallet scenarios, and current CI coverage is passing with that setup. That said, rollback-to-older-state can prune pending notes that only existed on the newer branch, so I will keep monitoring this in longer-running local runs and devnet-linked scenarios and revisit if I see regressions.

Yes, I think this is fine, thank you

andrussal (approved):
Thank you for addressing comments, I think this is good to go

PR #3103, opened 2026-07-08
chore: implement Nom traits for MantleTx

ntn-x2:
`SignedMantleTx` is the last one and I will leave it for a separate PR to @strinnityk, as we might want to change few things before (e.g., the `Vec<OpProof>` does not implement `NomEncode` and `NomDecode` because it's not bounded, so we need to perhaps represent signed txs as a bounded `Vec<(Op, OpProof)>` instead, so that we can enforce at compile time, as part of the type system, that ops and proofs count match.


hansieodendaal (approved):
LGTM

danielSanchezQ (approved):

PR #3105, opened 2026-07-08
fix(blend): proper wire encoding without any dynamically sized types

ntn-x2:

By design, all Blend messages are supposed to look the same. That means that, given the number of encapsulations, all messages have the same length. We can enforce that when decoding messages from the wire. Also, we can use arrays and boxed slices in Rust because we are never growing those layers once a message is created, and that is also by definition.

To make things easier to follow, I introduced `WireEncode` and `WireDecode` traits, similar to what we do for nom encoding, but with an additional context when decoding, which types can request to make judgment about the expected size of the input being decoded. In practice, most types use `()` as the context, with some types requesting the number of encapsulation layers. This allows for a modular, bottom-up approach to implementing encoding/decoding logic which also stays separate from the rest of the types implementations.

At the top level, the `EncapsulatedMessage` type, which is only ever received by a node, exposes a `decode` function only, with a test-only `encode` one for tests. On the other hand, the verified counterparts `EncapsulatedMessageWithVerifiedPublicHeader` and `EncapsulatedMessageWithVerifiedSignature`, which are both supposed to only be sent over the wire and never received, expose only an `encode` function.

Yes.

@ntn-x2 

N/A

No. But we might want to use this implementation to generate a test vector to include in the spec.

### DIFF
--- a/blend/message/src/codec.rs
+++ b/blend/message/src/codec.rs
@@ -1,0 +1,142 @@
+use lb_blend_proofs::{
+    quota::{PROOF_OF_QUOTA_SIZE, ProofOfQuota},
+    selection::{PROOF_OF_SELECTION_SIZE, ProofOfSelection},
+};
+use lb_key_management_system_keys::keys::{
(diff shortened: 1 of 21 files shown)


youngjoon-lee (approved):
great

PR #3107, opened 2026-07-08
refactor(cryptarchia-engine): block-tree invariants by construction

diogofriggo:

No new protocol behavior. This PR restructures the cryptarchia-engine's internal block tree so that the invariants the engine used to maintain by convention are now maintained by construction, while keeping the public API and wire format compatible with master. The observable behavior of the engine is unchanged except for the deltas listed below: all pre-existing tests pass unmodified, and the dependent chain-service compiles and passes its test suite without a single change.

## Motivation

This started as a fix for a small TODO on master https://github.com/logos-blockchain/logos-blockchain/blob/81a6a37694b9889c98caca8a30c1ce10b5133461/consensus/cryptarchia-engine/src/lib.rs#L409-L410

Eliminating `lib_depth` meant expressing "which forks are prunable" directly against the LIB rather than through a depth recomputed from the tip. Pulling that thread exposed how much of the block tree's correctness rested on conventions — hand-maintained parent walks, a tips set kept in sync by habit, pruning phases whose order was load-bearing but unenforced — and the fix grew into this refactor: making those conventions structural. The original TODO is resolved: the tip-relative `lib_depth` (and its round-trip — `tip.length - lib.length`, converted straight back to `lib.length` inside the pruning filter) is gone; stale-fork pruning is now driven by the LIB's own length, symmetric with `prune_immutable_blocks`.

### Design

**`Branch` / `Block` / `Role` split** (`block.rs`, renamed from `branch.rs`): a block's immutable facts and its mutable position in the tree are now different types.

- `Branch<Id>` (public, unchanged shape): plain data — id, parent, slot, length. Safe to copy and hold across engine mutations; nothing in it can go stale. Serialization format is identical to master, including the genesis-is-its-own-parent convention (pinned by a test).
- `Role` (crate-internal): `Lib` (the in-memory tree root), `Internal { children_count }`, or `Tip`. Roles change as the tree evolves, so they live only in the tree.
- `Block<Id>` (crate-internal): the map entry — `branch` plus `role`. Block values never escape the tree; public APIs hand out `Branch` copies.

The `children_count` on interior blocks lets fork pruning cascade: removing a fork demotes its fork point back to a `Tip` exactly when its last child is removed, so shared stale segments are cleaned up across multiple pruning passes without recomputing LCAs.

**Lineage walks** are centralized in `LineageIterator` (start block inclusive down to the LIB inclusive), used by LCA computation, density lookups, reorg collection, `nth_ancestor` and both pruning passes. `prune_fork` is a single iterator pipeline using a small `WithIsLast` adapter to apply the boundary decrement on the final removed block.

**LIB update is a typestate** (`typestate.rs`). Updating the LIB is a three-phase mutation in a precise order (stale forks must be pruned while the immutable segment still exists; the immutable walk terminates at the old LIB's role marker; the new LIB is re-tagged last). Each phase consumes a token only the previous phase produces, so reordering the phases is a compile error rather than a runtime corruption:

```rust
typestate::LibUpdate::new(new_lib)
    .prune_stale_forks(self)
    .prune_immutable_blocks(self)
    .commit(self)
```

### Public API compatibility

`config.rs` and `time.rs` are untouched. All public types, methods, bounds, const-ness, serde formats and error strings are identical to master, with these exceptions:

1. **Removed: `ForkDivergenceInfo`.** It was dead code (no consumers in this repository); external consumers would get a loud compile error.
2. **Duplicate block re-application no longer perturbs the tree.** On master, re-applying a known block overwrote its entry and re-inserted its id into the tips set (adding a non-leaf to the fork-choice candidate set). Now a duplicate passes the same validations as a new block (`ParentMissing` / `InvalidSlot` behave exactly as on master) and a fully valid duplicate is a logged no-op. `Ok`/`Err` results are unchanged.
3. **`Branches::lca` asserts its precondition.** Passing branches that are not in the tree now panics with a message; on master such inputs produced an index panic or a non-terminating walk. Valid inputs are unaffected. The signature's lifetimes were also relaxed (callers compile unchanged).

### Robustness

- Every panic-capable expression (unwraps, indexing, arithmetic) either carries a `SOUNDNESS:` comment stating the invariant that makes it unreachable, or fails with a message naming the violated assumption. Arithmetic on protocol values uses `checked_*`/`strict_*` operations; the fork-choice expressions are kept byte-identical to master.
- `assert_tree_consistent` (test-only) recomputes the tree's ground truth from parent links — child counts vs roles, tips ⇔ leaves, single LIB root, lineage integrity — and is exercised throughout the scenario tests.

### Testing

25 unit tests (up from 12), all of master's tests preserved. New coverage:

- tree invariants through forks, a reorg and two LIB advances;
- LCA in all orientations (ancestor/descendant, unequal/equal lengths, self-LCA) plus the not-in-tree panic contract;
- fork-choice: equal-length ties, the `maxvalid_mc` rejection arm from both sides of `k`;
- duplicates: valid no-op against a full engine snapshot (including a block with children), and both validation errors ordered before the no-op;
- pruning: same-slot immutable blocks (report collapses by slot, the tree removes all), reorg + LIB advance in a single `receive_block` (reorged blocks are collected against the pre-prune tree);
- LIB parent conventions (genesis self-parent; an advanced LIB keeps its pruned parent's id) — the contract chain-service's storage fallback depends on;
- serde wire-format pinning and round-trip;
- the `WithIsLast` adapter and the genesis-only LIB clamp.

### Future work

- Property-based testing: random insert/reorg/advance sequences with `assert_tree_consistent` as the oracle (proptest is already a workspace dependency).
- `Branch::parent` is kept for compatibility; migrating consumers to an `Option`-returning accessor and removing the self-parent convention is a separate breaking change (TODO in `block.rs`).

Specification (executable): https://github.com/logos-blockchain/logos-blockchain-specs/blob/master/cryptarchia/cryptarchia.py — `fork_choice` and the Genesis/Praos rules (`maxvalid_bg`, around lines 416 and 678), and the LIB computation ("the k-th block in the chain", around line 396). Behavioral reference tests: https://github.com/logos-blockchain/logos-blockchain-specs/blob/master/cryptarchia/test_fork_choice.py

The consensus rules themselves (`maxvalid_bg`, `maxvalid_mc`, the LIB trailing the tip by the security parameter) are unchanged and kept expression-identical to master where practical.

Implementation details that are idiosyncratic (not derived from the specification) are documented in code where they live:

- the `Role` taxonomy and `children_count` bookkeeping (`block.rs`), with a worked multi-pass pruning example on `prune_fork`;
- the typestate tokens (`typestate.rs` module docs state the three ordering facts they enforce and why each is load-bearing);
- the tree-root ("genesis is its own parent") convention and the advanced LIB keeping its pruned parent's id — documented on `Branch::parent` and pinned by tests; chain-service's storage fallback depends on it;
- duplicate re-application as a validated no-op — rationale documented at the guard in `apply_header` (the insertion path is not idempotent under count bookkeeping).

- Accountable for this PR: @diogofriggo
- Specification author(s) (main authors of `cryptarchia.py` in the specs repository): @davidrusu, @zeegomo, @youngjoon-lee
- Suggested code reviewers (main historical authors of this crate): @ntn-x2, @youngjoon-lee, @zeegomo

The refactor surfaced three behaviors that appear implementation-defined rather than specified. This PR preserves the existing semantics for all three; confirming them with the specification authors is advisable:

1. **Duplicate block re-application**: previous behavior mutated internal state (see API delta 2); this PR makes it a validated no-op. Neither behavior appears specified.
2. **Equal-length fork ties**: the first-seen chain wins (now pinned by a test). Tie-breaking does not appear in the specification.
3. **Forks diverging exactly at the LIB are retained** (only strictly deeper divergences are pruned). The boundary condition does not appear specified.

No. Protocol-level behavior is unchanged; the deltas listed in section 1 are Rust API surface only. If the section-4 clarifications result in specification updates, they would document existing behavior rather than change it.

## Checklist

* [x] 1. PR title follows Conventional Commits (`refactor(cryptarchia-engine): …`).
* [x] 2. Description added.
* [x] 3. Context and links to specification document(s) added (section 2).
* [x] 4. Main contacts (developers and specification authors) added (section 3).
* [x] 5. Implementation and specification are in sync: no protocol behavior changed; three implementation-defined behaviors are flagged in section 4 for clarification (all preserved as-is).
* [ ] 6. PR linked to a milestone
* [x] 7. New and changed logs reviewed against the logging guidelines: the two touched logs (`prune_fork`'s pruning summary, `apply_header`'s duplicate no-op) are `debug` on the existing `cryptarchia::engine` target; both explain "why a block was treated a certain way", matching the guideline's `debug` definition. No new `error`/`warn`/`info` logs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4488,6 +4488,7 @@ dependencies = [
  "logos-blockchain-utils",
  "rpds",
  "serde",
+ "serde_json",
  "serde_with",
  "thiserror 2.0.18",
  "time",
--- a/consensus/cryptarchia-engine/Cargo.toml
+++ b/consensus/cryptarchia-engine/Cargo.toml
@@ -27,5 +27,8 @@ time       = { workspace = true }
 tokio      = { features = ["time"], optional = true, workspace = true }
 tracing    = { workspace = true }
 
+[dev-dependencies]
+serde_json = { workspace = true }
+
 [features]
 tokio = ["dep:tokio"]
--- a/consensus/cryptarchia-engine/src/block.rs
+++ b/consensus/cryptarchia-engine/src/block.rs
@@ -1,0 +1,215 @@
+use core::hash::Hash;
+use std::num::NonZero;
+
+use rpds::HashTrieMapSync;
+
+use crate::time::Slot;
+
+/// Holds the immutable facts about a block: its identity, parentage, slot
+/// and chain length.
+///
+/// Values of this type are plain data, safe to copy and hold across engine
+/// mutations. A block's *role* in the tree (LIB, fork point, tip) changes as
+/// the tree evolves, so it is deliberately not part of this type: role
+/// questions are only answerable by the live block tree.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct Branch<Id> {
+    id: Id,
+    parent: Id,
+    slot: Slot,
+    length: u64,
+}
+
+impl<Id: Copy> Branch<Id> {
+    pub(crate) const fn new(id: Id, parent: Id, slot: Slot, length: u64) -> Self {
+        Self {
+            id,
+            parent,
+            slot,
+            length,
+        }
+    }
+
+    /// The block id.
+    pub const fn id(&self) -> Id {
+        self.id
+    }
+
+    /// The parent block id.
+    ///
+    /// Tree-root convention: the block the tree was created from is its own
+    /// parent. Code that walks parent links must therefore either stop when
+    /// `parent() == id()` or treat a failed lookup of the parent as the end
+    /// of the in-memory lineage (the parent of an advanced LIB is real but
+    /// pruned from memory).
+    ///
+    /// Instead of manually walking parent links, use the `LineageIterator`
+    /// which walks back the chain starting from a provided initial block
+    /// and naturally stops at the LIB block.
+    pub const fn parent(&self) -> Id {
+        self.parent
+    }
+
+    /// The slot the block belongs to.
+    pub const fn slot(&self) -> Slot {
+        self.slot
+    }
+
+    /// The chain length up to and including this block.
+    pub const fn length(&self) -> u64 {
+        self.length
+    }
+}
+
+/// A block's current role in the block tree.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Role {
+    /// The root of the in-memory tree: the latest immutable block.
+    /// Its ancestors have been pruned from memory.
+    Lib,
+    /// An interior block with at least one child.
+    ///
+    /// The count lets fork pruning turn a block back into a `Role::Tip`
+    /// when its last remaining child is removed.
+    Internal { children_count: NonZero<usize> },
+    /// A leaf: a chain tip with no children.
+    Tip,
+}
+
+/// A block-tree map entry: the immutable block data plus its current role.
+///
+/// `Block` values must not escape the tree, the role is only true for as
+/// long as the tree is not mutated. Public APIs hand out `Branch` copies
+/// instead, which cannot go stale.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct Block<Id> {
+    pub(crate) branch: Branch<Id>,
+    pub(crate) role: Role,
+}
+
+impl<Id: Copy> Block<Id> {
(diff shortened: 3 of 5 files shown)


davidrusu:
Before undertaking any refactors or large changes, join our discord and coordinate with the team to ensure this is something we want.

We can't dedicate time to reviewing arbitrary refactors if you do not coordinate ahead of time with us.

PR #3109, opened 2026-07-08
feat(zone-sdk): enable non zero gas price (tke)

pradovic:

This PR enables tke in zone-sdk. 

Know limitations: 
1. Current resubmitting tries with the same tx (same price)
2. Publishing form zone-sdk when node is offline now returns an error and it is up to the user to continue publishing when it gets ready signal again, or retry until it passes (we can solve this better in the follow up PR if needed)

TODO:
1. Provision fee in fund wallet call to mitigate know limitation 1 mid-term (needs node update, so left for follow up PR)
2. switch state to be per inscription ID instead of tx_uuid -> this would solve both limitation 1 and 2 (this is larger scope, needs to be tested and deliberately left for future) 

Yes

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


youngjoon-lee (approved):
I focused on `tx_builder.rs`. Looks good!

Since I haven't followed all details about how sequencer manages the queue of inscriptions, I couldn't fully understand the changes in `actor.rs` and `handle.rs`. I will catch up on that separately.

Also, thank you very much for updating cucumber tests.

youngjoon-lee:
Now I understand what fast-fail exactly means. 

If the sequencer knows that the node is offline (i.e., `self.connected == false`), it doesn't try to send `/wallet/fund` requests. Instead, it returns error to the user immediately. No queuing and no internal retry. 

The user will retry the tx manually once the sequencer emits Ready signal.

Could you confirm my understanding?

pradovic:
> Now I understand what fast-fail exactly means.
> 
> If the sequencer knows that the node is offline (i.e., `self.connected == false`), it doesn't try to send `/wallet/fund` requests. Instead, it returns error to the user immediately. No queuing and no internal retry.
> 
> The user will retry the tx manually once the sequencer emits Ready signal.
> 
> Could you confirm my understanding?

Yes, exactly. The connected flag is true while the block stream is active. We used to queue publish requests in this case as we could create txs without calling the node, but now we need the fund call working. 

I think this can be solved better  by switching zone-sdk state to be per inscription id instead of tx_uuid. But this is larger scope, and returning error is more honest, so I kept it like this for this PR not to block the TKE. I can work on optimizing this separately.

hansieodendaal (approved):
LGTM!

**Note:** I suspect calling `pub(super) async fn submit_atomic_zone_deposit_transaction(` in a tight loop when gas costs are going to be included will break related test(s) as `current_available_utxos_for_wallet` will not know about pending transactions, causing double spends, but we can fix that if/when it happens in the tests,

PR #3111, opened 2026-07-09
test(fee): fee market tests

andrussal:

This PR adds fee e2e tests.

It also includes 4 unit tests which currently fail, because it seems that the current logic fails to adapt to market changes properly (prices don't rise under demand and don't recover after quiet periods — failure messages have the details). 

There is also an issue where it seems the storage price is never updated at all — usage doesn't reach the price update. The PR includes a proposed fix for it.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.


tests/cucumber_tests/features/fees.feature
```diff
@@ -40,3 +40,89 @@ Feature: Fees
     And I fund an inscription transaction on node "NODE_1" as "FUNDED_INSCRIPTION"
     Then transaction "FUNDED_INSCRIPTION" is included on node "NODE_1" in 120 seconds
     Then I stop all nodes
+
+  # The execution gas price after every block must equal the spec's calculation,
+  # recomputed here independently. Storage usage must also reach the pricing
+  # update and move its price at an epoch boundary. Together these guard the
+  # consensus math and that the storage market remains connected to network
+  # usage; price viability is covered by ledger market-liveness tests.
+  @transactions_ci
```
hansieodendaal: I think we can add a ` @fees_ci` tag now and run them in their own cucumber ci task, although it has to do with transactions, it is a separate topic.

tests/cucumber_tests/features/fees.feature
```diff
@@ -40,3 +40,89 @@ Feature: Fees
...
+      | node_name | account_index | wallet_name | connected_to |
+      | NODE_1    | 1             | WALLET_1A   |              |
+      | NODE_1    | 2             | WALLET_2A   |              |
+    When node "NODE_1" is at height 1 in 180 seconds
+    And I submit an exactly funded self-transfer from wallet "WALLET_1A" via node "NODE_1" as "TRAFFIC_A"
+    And I submit an exactly funded self-transfer from wallet "WALLET_2A" via node "NODE_1" as "TRAFFIC_B"
+    And I record per-block gas prices on node "NODE_1" for 125 slots in 300 seconds
+    Then transaction "TRAFFIC_A" is included on node "NODE_1" in 60 seconds
+    And transaction "TRAFFIC_B" is included on node "NODE_1" in 60 seconds
+    And recorded execution gas prices follow the execution market spec reference
```
hansieodendaal, line 73: As a test designer, what ensures that the detailed comment above this test is asserted? I would have expected to see some assertion about "epoch boundary" before.

tests/cucumber_tests/features/fees.feature
```diff
@@ -40,3 +40,89 @@ Feature: Fees
...
+      | 4             | 1           | 2            |
+    And I have a cluster with capacity of 1 nodes
+    And I start nodes with wallet resources:
+      | node_name | account_index | wallet_name | connected_to |
+      | NODE_1    | 1             | FUNDER_A    |              |
+      | NODE_1    | 2             | FUNDER_B    |              |
+      | NODE_1    | 3             | RECIPIENT_A |              |
+      | NODE_1    | 4             | RECIPIENT_B |              |
+    When node "NODE_1" is at height 2 in 240 seconds
+    And I concurrently fund transfers of 500 LGO from wallets "FUNDER_A" and "FUNDER_B" to wallets "RECIPIENT_A" and "RECIPIENT_B" via node "NODE_1" as "TX_A" and "TX_B"
```
hansieodendaal: Question about this step
```
    And I concurrently fund transfers of 500 LGO from wallets "FUNDER_A" and "FUNDER_B" to wallets "RECIPIENT_A" and "RECIPIENT_B" via node "NODE_1" as "TX_A" and "TX_B"
```
See related comment at `fn concurrently_fund_transfers`
Maybe we should generalize this step to take arguments as a table, then we are not limited by size two.
An interesting take on this _for a different PR and scenario and test case_ would be to choose different nodes for each transfer, or to submit different transactions from the same wallet to recipient(s) using different nodes.

tests/cucumber_tests/features/fees.feature
```diff
@@ -40,3 +40,89 @@ Feature: Fees
     And I fund an inscription transaction on node "NODE_1" as "FUNDED_INSCRIPTION"
     Then transaction "FUNDED_INSCRIPTION" is included on node "NODE_1" in 120 seconds
     Then I stop all nodes
+
+  # The execution gas price after every block must equal the spec's calculation,
```
hansieodendaal, line 44: These are nice _fees_ test additions. I think we could enhance them by also using the `NODE_1` built-in wallet in the mix, to make sure we get what we expect.

tests/src/common/fee_spec.rs
```diff
@@ -0,0 +1,274 @@
+//! Shared helpers for the fee e2e scenarios.
+//!
+//! `ExecutionMarketReference` and the `SPEC_*` constants are the spec's price
+//! update written out again by hand from the spec documents (not imported
+//! from node code), so the conformance scenario notices if the node's
+//! pricing math changes. The rest are plain helpers for building and
+//! checking fee-paying transactions.
```
hansieodendaal, line 7: Spec reference perhaps?

tests/cucumber_tests/features/fees.feature
hansieodendaal, line 44: I am missing a step in a test in this feature that purposefully add a tip that would be required for fee increase at epoch boundary? Only cases where we have zero tip and a shortfall tip, not a positive tip.

tests/src/cucumber/steps/fees/actions.rs
```diff
@@ -0,0 +1,328 @@
...
+        }
+    }
+
+    Err("blocks stream ended before the requested slots were observed".to_owned())
+}
+
+/// The streamed block plus any ancestors the stream subscription missed,
+/// oldest first: walks parent links back until it reconnects with the last
+/// recorded block.
+async fn block_and_missed_ancestors(
```
hansieodendaal, line 214: We could also use `/cryptarchia/blocks_range"` endpoint from genesis to the current tip/slot.

tests/src/cucumber/steps/fees/actions.rs
```diff
@@ -0,0 +1,328 @@
...
+        slot: u64::from(block.header.slot),
+        execution_price: prices.execution_base_gas_price.into_inner(),
+        storage_price: prices.storage_gas_price.into_inner(),
+        block_execution_gas: fee_spec::spec_block_execution_gas(block)?,
+    })
+}
+
+/// Fires two wallet transfer requests at the same time, funded from the same
+/// key set, and records both transactions under their aliases.
+pub async fn concurrently_fund_transfers(
```
hansieodendaal: Maybe this could be a general vec, not limited by size two.
`concurrently_fund_n_transfers`

tests/src/cucumber/steps/fees/assertions.rs
```diff
@@ -0,0 +1,307 @@
...
+    info!(
+        target: TARGET,
+        "Storage gas price responded to usage over {} blocks", records.len(),
+    );
+
+    Ok(())
+}
+
+/// Checks two included transactions spent no common ledger input.
+pub async fn transactions_spent_disjoint_inputs(
```
hansieodendaal: Can this ever fail? Do consensus, mempool and block proposal rules not guarantee this?

tests/src/cucumber/steps/fees/assertions.rs
```diff
@@ -0,0 +1,307 @@
...
+    for _ in 0..1_000 {
+        if Some(cursor) == genesis_id {
+            break;
+        }
+
+        let Some(block) = client
+            .block(&cursor)
+            .await
+            .map_err(|source| format!("block fetch failed: {source}"))?
+        else {
```
hansieodendaal: We could use the `"/cryptarchia/blocks_range"` endpoint for this

tests/src/cucumber/steps/fees/assertions.rs
```diff
@@ -0,0 +1,307 @@
...
+/// Per the [1.0.0] Storage Market spec, usage recorded during an epoch
+/// drives the price update at its boundary. Guards that the live storage
+/// market remains connected to network usage.
+pub fn storage_prices_respond_to_usage(world: &CucumberWorld, step: &Step) -> StepResult {
+    let records = recorded_gas_prices(world, step)?;
+
+    let first_price = records[0].storage_price;
+    if records
+        .iter()
+        .all(|record| record.storage_price == first_price)
```
hansieodendaal: Maybe we could use a stronger assertion here? Otherwise, it could pass for either 1 → 2 or 1 → 0.
```suggestion
        .any(|record| record.storage_price > first_price)
```

tests/src/common/fee_spec.rs
```diff
@@ -0,0 +1,274 @@
...
+    // which depends on the fee - so recompute until the fee stops changing.
+    let mut fee = fee_for_output(utxo.note.value);
+    loop {
+        let recomputed = fee_for_output(output_for_fee(fee));
+
+        if recomputed == fee {
+            break;
+        }
+        fee = recomputed;
+    }
```
hansieodendaal, line 189: Maybe we can add a fail-safe exit condition here, guarding against an encoding-size boundary that could theoretically oscillate between two fee values and hang the scenario.

tests/src/common/fee_spec.rs
```diff
@@ -0,0 +1,274 @@
...
+    },
+};
+use lb_testing_framework::configs::wallet::WalletAccount;
+
+use crate::common::wallet::transfer_proofs_for_funded_wallet_tx;
+
+/// [1.0.0] Execution Market: base fee starts at 1.
+pub const SPEC_GENESIS_EXECUTION_PRICE: u128 = 1;
+/// [1.0.0] Storage Market: storage price starts at 1.
+pub const SPEC_GENESIS_STORAGE_PRICE: u64 = 1;
```
hansieodendaal: Is this still used?

tests/src/common/fee_spec.rs
```diff
@@ -0,0 +1,274 @@
...
+}
+
+/// The mandatory fee of a transaction at the given prices: execution gas *
+/// base price + encoded signed size * storage price ([1.5.0] Mantle).
+///
+/// # Panics
+///
+/// Panics on gas overflow, which small test transactions cannot hit.
+#[must_use]
+pub fn fee_of<State: VerificationState>(tx: &SignedMantleTx<State>, prices: &GasPrices) -> u64 {
```
hansieodendaal: Is this still used?

hansieodendaal (commented):
Very nice addition, good job! 

I left some comments to consider/answer, and I think the PR description could also be out of date now that issues identified by this PR have been addressed and merged to master.

tests/cucumber_tests/features/fees.feature
andrussal: > I think we can add a ` @fees_ci` tag now and run them in their own cucumber ci task, although it has to do with transactions, it is a separate topic.

Good idea. I’ll move these scenarios to a dedicated @fees_ci suite so failures and artifacts are reported separately.

tests/cucumber_tests/features/fees.feature
andrussal: > As a test designer, what ensures that the detailed comment above this test is asserted? I would have expected to see some assertion about "epoch boundary" before.

Good point. The price records indirectly showed the boundary, but the scenario did not assert it explicitly. I’ll add a step that verifies the recording crossed an epoch boundary before asserting the storage-price change.

tests/cucumber_tests/features/fees.feature
andrussal: > Question about this step
> 
> ```
>     And I concurrently fund transfers of 500 LGO from wallets "FUNDER_A" and "FUNDER_B" to wallets "RECIPIENT_A" and "RECIPIENT_B" via node "NODE_1" as "TX_A" and "TX_B"
> ```
> 
> See related comment at `fn concurrently_fund_transfers` Maybe we should generalize this step to take arguments as a table, then we are not limited by size two. An interesting take on this _for a different PR and scenario and test case_ would be to choose different nodes for each transfer, or to submit different transactions from the same wallet to recipient(s) using different nodes.

Agreed. I changed this to a table-based step, including the node for each transfer, so it is no longer limited to two requests.

tests/cucumber_tests/features/fees.feature
andrussal: > These are nice _fees_ test additions. I think we could enhance them by also using the `NODE_1` built-in wallet in the mix, to make sure we get what we expect.

Good point. The payment case now uses the built-in node wallet, while the inscription case uses the user wallet, so both paths are covered.

tests/src/common/fee_spec.rs
andrussal: > Spec reference perhaps?

Added links to the corresponding spec versions.

tests/src/cucumber/steps/fees/actions.rs
andrussal: > We could also use `/cryptarchia/blocks_range"` endpoint from genesis to the current tip/slot.

Good suggestion. The recorder now uses blocks_range for existing history and the live feed for new blocks.

tests/src/cucumber/steps/fees/assertions.rs
andrussal: > Can this ever fail? Do consensus, mempool and block proposal rules not guarantee this?

Yes, once both transactions are included, consensus already guarantees their inputs are disjoint. The intended check was the wallet’s pending-note reservation: if it reused an input, one transaction would not be included. So asserting that both transactions land is sufficient, and I removed the redundant input check.

tests/src/common/fee_spec.rs
andrussal: > Maybe we can add a fail-safe exit condition here, guarding against an encoding-size boundary that could theoretically oscillate between two fee values and hang the scenario.

Good catch. I added cycle detection and a bounded number of attempts.

hansieodendaal (approved):
LGTM

Potential follow-up:
- https://github.com/logos-blockchain/logos-blockchain/pull/3111#discussion_r3673226510
- Update PR description
- Prove that the storage-price rise occurs at the epoch boundary - a regression can incorrectly change the storage price halfway through an epoch.

PR #3113, opened 2026-07-09
chore(zone-sdk): unify pending set with channel view

pradovic:

This PR fixes inconsistency with pending set and channel view. Pending set now includes other sequencers tx as well, and emits adopted in the extension case only if it expands the pending set. 

The consumer pattern is to treat published messages as part of the state, and react to channel update as before, reducing the need to dedup in the adopted case. It also updated integration tests to follow this pattern.

This makes the information more stable in following case:
```
LIB <- A <- B
LIB <- A      # Block containing B is reorged out
LIB <- A <- B # we retry B and it goes in again
```

Yes

spec @davidrusu 

@pradovic 
Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


zone-sdk/src/sequencer/state.rs
```diff
@@ -145,6 +145,59 @@ impl TxState {
...
+            .push(tx_hash);
+        self.pending.insert(
+            tx_hash,
+            PendingInscription {
+                tx_hash,
+                signed_tx,
+                parent_msg,
+                this_msg,
+                payload,
+                withdraws: None,
```
hansieodendaal: Observed transactions are always stored as plain `PendingInscription` with `withdraws: None`. Could external or restored observed transactions be atomic `inscription`+`withdraw` bundles?

hansieodendaal (commented):
Looking good, just some questions/observations to clear up please:

- In `sequencer/block_fetch.rs‎`, `pub(super) async fn fetch_and_process_blocks<Node>(` does not call `observe_channel_inscriptions- ` before `state.process_block`. Is that intentional? 
If I understand correctly, it means other-sequencer channel transactions first discovered during LIB backfill are not inserted into the pending set, which appears to violate the new “pending mirrors channel view” invariant. 
- In `sequencer/types.rs‎` at `pub struct PublishResult {` the doc comment still says consumers use publish results “to dedup the same entry when it later shows up in ChannelUpdate::adopted.” Needs update?

zone-sdk/src/sequencer/state.rs
pradovic: Good catch, this seems like a bug, I will update the code. That applies for the suggestions in the global comment as well. I will address all of those, some with code some with additional comments. Thanks!

The reason for it was that before we only cared about inscription tip update, but now that we retry other sequencer txs we should have withdraw populated as well, we should retry all od them.

I dont like the idea that this can catch also prepare_tx made transactions in the ame format, but maybe this is actually fine as we treat this format as atomic withdraw

pradovic: Done

pradovic:
> Looking good, just some questions/observations to clear up please:
> 
>     * In `sequencer/block_fetch.rs‎`, `pub(super) async fn fetch_and_process_blocks<Node>(` does not call `observe_channel_inscriptions- ` before `state.process_block`. Is that intentional?
>       If I understand correctly, it means other-sequencer channel transactions first discovered during LIB backfill are not inserted into the pending set, which appears to violate the new “pending mirrors channel view” invariant.
> 
>     * In `sequencer/types.rs‎` at `pub struct PublishResult {` the doc comment still says consumers use publish results “to dedup the same entry when it later shows up in ChannelUpdate::adopted.” Needs update?

I have addressed both topics + the one from comment. Let me know how it looks. Thanks!

zone-sdk/src/sequencer/block_fetch.rs
```diff
@@ -192,28 +195,52 @@ where
...
+                    if inscribe.is_some() {
+                        multi_inscribe = true;
+                    } else {
+                        inscribe = Some(i);
+                    }
+                }
+                Op::ChannelWithdraw(w) if w.channel_id == channel_id => {
+                    withdraws.push(w.clone());
+                }
+                _ => {}
```
hansieodendaal, line 224: Having difficulty wrapping my head around this `multi_inscribe`. Do we not need to track those at all, even if the doc comment says _... skipped naturally, as are the exotic multi-inscribe shapes `restore_pending_tx` tracks as opaque_? It seems wrong.
- `==1 Op::ChannelInscribe` - track
- `>=1 Op::ChannelWithdraw` - track
- `==1 Op::ChannelInscribe` + `>=1 Op::ChannelWithdraw` - track
- `>1 Op::ChannelInscribe` - _do not track_
- `>1 Op::ChannelInscribe` + `>=1 Op::ChannelWithdraw` - _do not track_

hansieodendaal (commented):
Hi, thanks for the updates, having a further question on the new additions.

zone-sdk/src/sequencer/block_fetch.rs
pradovic: Hmm, adopted should not adopt anything other then:
1. Single inscription - publish message
2. Inscirption + Vec<Withdraws) - atomic withdraw

I will remove it from observing in adopted. The asymmetry comes from the fact that we also support prepare_tx and config update tx which end up in pending_other. I will fix this now, but we should think in the future how to support this better (arbitrary txs). Current contract is that we only track inscription and atomic withdraw for the concurrent sequencers scenario

pradovic: Deffered for the follow up PR. Will merge this PR and open a follow up asap

hansieodendaal (approved):
LGTM (with follow-up PR to address the final comments)

PR #3116, opened 2026-07-09
feat(wallet): add priority tip to support zone-sdk

pradovic:

This PR adds an optional priority tip to wallet fund call and propagates it to API and c-bindings. It is a hardening PR for tke aware zone-sdk.

Zone-sdk has periodic resubmission of transactions. Since the price can change over time this could cause tx to be dropped and make periodic resubmission ineffective. It was decided not to recreate tx at the moment as this needs more testing and has a larger scope.

 @davidrusu proposed to add priority fee to decrease the chance of price becoming larger than pending transactions. This seems like an effective short/mid term solution until we have more information and decide whether to recreate transactions on each resubmission.

Yes

@davidrusu 

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


wallet/src/lib.rs
```diff
@@ -223,15 +227,21 @@ impl WalletState {
         // Consume large valued notes first to ensure we converge.
         utxos.sort_by_key(|utxo| -i128::from(utxo.note.value));
 
+        // The funding target: the tx's excess balance over the mandatory fee
+        // must end up exactly at `priority_fee` (the execution tip).
+        let target = i128::from(priority_fee);
```
youngjoon-lee: nit: I think, it's clearer to call this `delta_target`?

zone-sdk/src/sequencer/tx_builder.rs
```diff
@@ -20,6 +20,16 @@ use lb_key_management_system_service::keys::{Ed25519Key, Ed25519Signature};
...
 
+/// Execution tip paid on top of the mandatory fee when funding a transaction,
+/// buffering gas-price movement between funding and inclusion (the base fee
+/// moves at most 12.5% per block — the ledger updates it as
+/// `fee × (11_176_760 + avg_gas) / 12_773_440` with `avg_gas` bounded by the
+/// block gas limit of `3_193_360`, giving factors of exactly 0.875..1.125 —
+/// so this covers a few blocks of drift at current fee levels).
+///
+/// TODO: promote to [`FundingConfig`] if clients need to tune it.
+const PRIORITY_FEE: Value = 200;
```
youngjoon-lee, line 29: nit: I think, it's better to not put the equation in this comment, because the meaning of `fee` and magic numbers are not clear for zone-sdk readers. And, the ledger logic could change in the future. IMO, it's sufficient to just mention that the base fee moves at most 12.5% per block, and the `PRIORITY_FEE` covers a few blocks of drift in the current spec. Not important, though.

---------

Question: I have a question to clearly understand what we expect for priority fee, here.

If I understand correctly, the flow is:
- We fund a tx before submission (not per-resubmission).
- When funding the tx, we target to make the funding delta equal to the `PRIORITY_FEE`.
- If the tx is accepted by the node before the gas price changes, the entire `PRIORITY_FEE` goes to the block proposer. 
- If the tx is rejected by the node for some reason, and if it is resubmitted after the gas price increases, the `PRIORITY_FEE - fee_increase` goes to the block proposer. (`fee_increase` is a result of gas price increase).

And, the name 'priority' comes from the fact that block proposers will try to maximize their revenue when picking txs, right? So, this 200 fee is basically used as a buffer from perspective of zone-sdk, but it also motivates block proposers to pick the txs as early as possible?

youngjoon-lee (approved):

zone-sdk/src/sequencer/tx_builder.rs
pradovic: Thanks! Agreed about the comment, good point, I did not take users in the account when writing the comment.

From zone-sdk perspective it is really a buffer to decrease the change of price going up, as we don't have dynamic tx recreation right now (we can add this in the future if needed).

Regarding the name, it was suggested by @davidrusu  so maybe better for him to comment, but I agree with your breakdown, this is how I see it as well. The name suggests priority.

youngjoon-lee: Thank you for your confirmation. I think the name "priority" makes also sense. And, IIRC, the same name is used in Ethereum. So, it's fine by me.

pradovic: Thank you!

Makes sense to me as well, I think that is why David proposed it as a more general name for API.

pradovic: I just reread your understanding, and leaving just 2 clarifications for completeness, but I think we are on the page:
1. Current implementation does not priorities txs when applying txs, if I understand correctly, but it is a possibility I guess.

2. it's not specific to periodic esubmission — the tx doesn't change at all. Whether it sat in the mempool or was re-sent byte-identically, the tip is evaluated at inclusion time: 200 − fee_increase. And the boundary: if the price rises so much that the increase exceeds 200, the tx's balance no longer covers the mandatory fee and it gets dropped at block assembly — that's where the buffer runs out. So this is kind of like a best effort fix, to avoid recreating txs. Only if the user chooses to submit the message again via publish_message the new tx will be made with a new funding.

So currently it acts as a buffer for zone-sdk, as we don't handle rejected txs by the ledger at all and don't follow the mempool, at the moment. This will be improved in the future.

hansieodendaal (commented):
I have two comments to consider please:

**Comment 1.** 
We should not require another UTXO when no positive change output can be created.

`delta_with_change` includes the additional gas cost of a hypothetical change output. Therefore, when `delta_with_change == target`, adding change would produce a zero-valued note and is rightly impossible—but the original builder, without that output, has a lower gas cost and is already fundable. It would retain the requested priority fee plus the avoided change-output cost as additional tip.

Could we distinguish this “no positive change to return” outcome from genuinely insufficient funds, so `fund_tx` can retain the already-fundable transaction rather than consume another UTXO?

Example:

Item | Value
-- | --
Transaction net balance | 1,000
Mandatory gas without change | 754
Mandatory gas with change | 794
Requested priority fee | 206
delta_with_change | 1,000 − 794 = 206
Change amount | 1,000 − 794 − 206 = 0
Actual excess without change | 1,000 − 754 = 246
Excess above requested priority fee | 246 − 206 = 40


**Comment 2.** 
We have a semantic mismatch between the method name/result and its caller:

`return_change` returning `None` naturally means “there is no valid positive change output.”
But `fund_tx` interprets that result as “continue adding inputs.”

pradovic:
> I have two comments to consider please:
> 
> **Comment 1.** We should not require another UTXO when no positive change output can be created.
> 
> `delta_with_change` includes the additional gas cost of a hypothetical change output. Therefore, when `delta_with_change == target`, adding change would produce a zero-valued note and is rightly impossible—but the original builder, without that output, has a lower gas cost and is already fundable. It would retain the requested priority fee plus the avoided change-output cost as additional tip.
> 
> Could we distinguish this “no positive change to return” outcome from genuinely insufficient funds, so `fund_tx` can retain the already-fundable transaction rather than consume another UTXO?
> 
> Example:
> Item 	Value
> Transaction net balance 	1,000
> Mandatory gas without change 	754
> Mandatory gas with change 	794
> Requested priority fee 	206
> delta_with_change 	1,000 − 794 = 206
> Change amount 	1,000 − 794 − 206 = 0
> Actual excess without change 	1,000 − 754 = 246
> Excess above requested priority fee 	246 − 206 = 40
> 
> **Comment 2.** We have a semantic mismatch between the method name/result and its caller:
> 
> `return_change` returning `None` naturally means “there is no valid positive change output.” But `fund_tx` interprets that result as “continue adding inputs.”

Thank you!

I think I agree, but it existed before this PR imo -> with target 0 the same window exists: inputs that cover the fee but where the leftover is smaller than the change note's own cost (e.g. fee 754 without change / 794 with: inputs of 755–794 can pay the fee but can't create positive change), so fund_tx pulls another UTXO or fails. 
 
`test_fund_tx_unfundable_region` on master documents exactly this. This PR only moved the target from 0 to priority_fee, so the window shifted but the behavior is unchanged. 

If this understanding is correct we should fix it in a separate PR as it changes behavior for all fund callers (small excess kept as tip instead of pulling a note) and needs its own test updates, while this PR keeps priority_fee = 0 bit-identical to today.

Sorry, maybe I am getting it all wrong, please correct if I am not seeing it the right way.

@youngjoon-lee Could you advise here as you know the wallet code better. Was this behavior intentional?

hansieodendaal:
> If this understanding is correct we should fix it in a separate PR...

Agreed

hansieodendaal (approved):
LGTM (with new PR(s) to address the additional comments)

zone-sdk/src/sequencer/tx_builder.rs
youngjoon-lee: Yeah, right. We're on the same page. Thank you for your confirmation!

youngjoon-lee:
> @youngjoon-lee Could you advise here as you know the wallet code better. Was this behavior intentional?

Good point! Yes, this issue existed before this PR. I just opened an issue https://github.com/logos-blockchain/logos-blockchain/issues/3128 to address that point.

PR #3119, opened 2026-07-10
chore: Wire* traits

ntn-x2:

This PR lays the foundation to migrate all types that are ever encoded in a binary format to a common set of traits, specifically `WireEncode`, `WireDecode`, and a set of macros. The workings are very similar to the previous `NomEncode`/`NomDecode` and related tools, with the difference that this set of traits is nom-agnostic, and it supports a `type: Context` when decoding: this was deemed necessary when implementing wire encoding/decoding for Blend messages, which in [this PR](https://github.com/logos-blockchain/logos-blockchain/pull/3105) were using their own trait system, but they have also now been migrated under the same trait system that is implemented for the core components as well.

The ultimate goal is simple: we encode binary encoding/decoding logic once per type, and we enforce at least one fixture that a well-known value of each type is tested against. Additional auto-generated tests include round-trip tests, hence I deleted a few roundtrip tests that are now deprecated.

The new encoding trait introduces something that the old trait system did not have:

* a method to return the encoded length of a type: `fn encoded_length(&self) -> usize`
* a method to append the encoding to a pre-allocated vector: `fn encode_into(&self, out: &mut Vec<u8>)`

These two traits allow a top-level component to create a single vector with the required capacity given by summing up the `encoded_length` of its components, and then call `encode_into` for each of them, instead of collecting several, sparsely-allocated `Vec<u8>`s.

The trait also provides two default implementations:

* `fn encode(&self) -> Box<[u8]>` and `fn encode_to_vec(&self) -> Vec<u8>`, which both pre-allocated a vector of the required capacity and let each component extend the vector. These functions are meant to be called on top-level components only, but the default implementations are of course provided to all types implementing `WireEncode`, and can be overridden for any reason, if needed.

There are still a lot of rough edges that I want to polish, and although this PR seems quite big, the core of the changes consist of the two new crates in the `wire` folder. The rest is mostly migration from the old two traits system into the new one. Future follow-up PRs will improve especially around the well-known fixtures, which are still a bit cryptic, but at least they are present and used to catch any wire-breaking changes going forward. The plan is to move more and more components to this trait system and provide fixtures to be used as test vectors for specs and other implementers.

It's a large PR, but the core of the changes are inside the new `wire` folder. You can skip over the macro machinery, and simply see how the macros are used in the rest of the codebase. The new `WireEncode` and `WireDecode` traits deserve special attention. The rest of the changes are mostly migrating from the old to the new trait system.

@ntn-x2 

N/A

No.


youngjoon-lee (approved):

blend/message/src/.DS_Store
youngjoon-lee, line 1: I think this file has been included by mistake.

ntn-x2:
Closed in favor of multiple smaller PRs. First of the batch: https://github.com/logos-blockchain/logos-blockchain/pull/3198.

PR #3122, opened 2026-07-10
chore(zone-sdk): unify channel-update-txs

pradovic:

This PR unifies channel update orphaned/adopted txs, refactors supported/unsupported tx and fixes a few leftover bugs along the way.

Yes

@pradovic 
@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


zone-sdk/src/sequencer/types.rs
```diff
@@ -70,31 +70,40 @@ pub struct WithdrawArg {
...
+/// A tx reported in a [`ChannelUpdate`], in both `adopted` and `orphaned`.
 ///
-/// The consumer republishes by calling the same SDK method they used
-/// originally with the data carried inside the variant:
-/// - [`OrphanedTx::Inscription`] → [`super::SequencerHandle::publish`] with
-///   `info.payload`
-/// - [`OrphanedTx::AtomicWithdraw`] →
+/// For orphaned entries the consumer republishes by calling the same SDK
+/// method they used originally with the data carried inside the variant:
+/// - [`ChannelUpdateTx::Inscription`] → [`super::SequencerHandle::publish`]
```
hansieodendaal: I think we could drop `super::` in all doc comments
```suggestion
/// - [`ChannelUpdateTx::Inscription`] → [`SequencerHandle::publish`]
```

hansieodendaal (commented):
I like the direction of preserving the ledger-true channel tip and separating transaction classification from retry handling. However, I believe the heuristics for classifying Mantle transactions in `fn classify_channel_txs(txs: &[SignedMantleTx], channel_id: ChannelId) -> Vec<BlockChannelTx>` with multiple `Op::ChannelInscribe` operations are wrong and can create a liveness stalemate until those transactions reach LIB.

A valid transaction can contain multiple inscriptions for the same channel when each subsequent inscription uses the preceding inscription’s message ID as its parent. The ledger applies operations sequentially, so each inscription advances the channel tip before the next is validated.

With this implementation, `classify_channel_txs` retains every inscription, so internal channel-tip tracking remains ledger-correct. However, because `inscribes != 1`, it classifies the enclosing transaction as `BlockChannelTx::Custom`. `Custom` entries are neither mirrored into pending state nor surfaced by `classify_update_tx`, which means they are omitted from both `ChannelUpdate.adopted` and `ChannelUpdate.orphaned`.

Consequently, a consumer that applies non-finalized zone state from `Event::BlocksProcessed.channel_update` can observe the canonical L1 channel tip advancing without receiving the corresponding inscription payloads to apply. Those payloads are only surfaced on finalization, leaving application state behind the canonical channel state in the interim. If a subsequent zone action depends on that missing state, an authorized sequencer can prevent other participants from progressing until LIB advances.

The same problem applies to deposit-plus-inscription transactions: `ChannelDeposit` marks the transaction as `foreign_ops`, making it `Custom` even though deposit + inscription is an intended atomic bridge pattern.

Please add regression coverage for valid multi-inscribe and deposit-plus-inscription transactions, proving that they cannot advance the consumer-visible channel tip without also delivering the corresponding adopted state. More fundamentally, unsupported retry semantics should not cause otherwise valid adopted inscriptions to disappear from the channel-update stream.

We also need unit tests to cover all new behavioral changes introduced by this refactor.

pradovic:
@hansieodendaal  Thank you for the review!

The goal before was to only support the main path: 
1. Publish
2. Publish atomic withdraw

And the expectation was that users would use only this path to use the zone-sdk and channels. However, I agree with your point that there is discrepancy. One possibility would be to completely remove prepare_tx + publish_signed_tx path, and the other is to support it somehow. 

I unified the approach by introducing the third Custom type, which as emitted for the users as is. This way we can cover custom transactions submitted by the API or zone-sdk publish_signed_tx path. Our docs still mention the first 2 above as the main path, but we still emit everything else as Custom for users to handle if there is a specific advanced user that needs to submit custom transactions (this could include various combinations of different ops).

This way the contract is uniform at least, imo, but the preferred way is still publish and publish_atomic_withdraw path.

All finalized tx are emitted as before with a full list of ops.

In the future we can expand this to support smarter builders for custom tx to handle many various use-cases.

Let me know what you think.

hansieodendaal (approved):
LGTM!

Thanks for addressing the concerns, this feels much better than leaving a gate open that can land us in an undefined state.

**Remaining task:** Add dedicated end-to-end/integration coverage proving consumers apply and recover Custom transactions across a real block-processing/reorg flow.

pradovic:
@hansieodendaal I have added the e2e scenario you asked and fixed bugs along the way. It grew the PR a lot as scenario takes a lot of lines + there was also a completeness bug in shedding of custom txs. Let me know if I can merge.

hansieodendaal (approved):
LGTM

PR #3123, opened 2026-07-10
feat(mantle): Verified Transactions revamp

### LINKED ISSUE #2638: refactor `Operation` trait to have all logic
`Operation` trait should contain signature creation/verification and so on. Currently those are spread across services (wallett, ledger, ...).

strinnityk:
Revamp `SignedMantleTx` into a typestate: a transaction carries its verification level in the type, and each level can only be obtained by passing the corresponding checks.
- `preverify()` (Unverified → Preverified) runs the stateless checks (proof count, inscription/leader-claim signatures).
- Stateful checks move out of the ledger into the transaction. `VerifiedOps` yields each op only after verifying it against current state, interleaved with execution so ops can depend on earlier ops in the same tx.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.


core/src/mantle/transactions/tx.rs
```diff
@@ -299,26 +311,25 @@ impl Transaction for MantleTx {
...
+//   The current tests behave just like the old code.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct SignedMantleTx {
-    pub mantle_tx: MantleTx,
+pub struct SignedMantleTx<State: VerificationState> {
+    mantle_tx: MantleTx,
     // TODO: make this more efficient
-    pub ops_proofs: Vec<OpProof>,
+    ops_proofs: Vec<OpProof>,
+    state: State,
```
danielSanchezQ: If state has no context then a phantom data is more correct (it will be removed from the actual memmory footprint)

core/src/mantle/transactions/tx.rs
```diff
@@ -383,133 +429,377 @@ pub trait OperationVerificationHelper {
...
-    /// Create a new `SignedMantleTx` and verify that all required proofs are
-    /// present and valid.
-    ///
-    /// This enforces at construction time that:
-    /// - `ChannelInscribe` operations have a valid Ed25519 signature from the
-    ///   declared signer
-    pub fn new(mantle_tx: MantleTx, ops_proofs: Vec<OpProof>) -> Result<Self, VerificationError> {
-        let tx = Self {
+impl<State: VerificationState> SignedMantleTx<State> {
+    fn with_state<T: VerificationState>(self, state: T) -> SignedMantleTx<T> {
```
danielSanchezQ: Not sure this is the way to go. The state is just a typehint as it doesnt contain anything (and probably will not). Instead we should  implement `new` for `unverified` and some other method maybe for the other state. Because this method means we can bridge from any state to any state and it is probably make someone to shot in their feet.

core/src/mantle/transactions/tx.rs
```diff
@@ -383,133 +429,377 @@ pub trait OperationVerificationHelper {
...
+                clippy::unnested_or_patterns,
+                reason = "Clarity on valid op/proof pairs."
+            )]
+            (Op::ChannelConfig(_), OpProof::ChannelMultiSigProof(_))
+            | (Op::ChannelDeposit(_), OpProof::ZkSig(_))
+            | (Op::ChannelWithdraw(_), OpProof::ChannelMultiSigProof(_))
+            | (Op::SDPDeclare(_), OpProof::ZkAndEd25519Sigs { .. })
+            | (Op::SDPWithdraw(_), OpProof::ZkSig(_))
+            | (Op::SDPActive(_), OpProof::ZkSig(_))
+            | (Op::Transfer(_), OpProof::ZkSig(_)) => Ok(()),
```
danielSanchezQ, line 560: @thomaslavaur as per our convo: We can probably check all this signatures except the channel related ones right?

core/src/mantle/transactions/tx.rs
```diff
@@ -383,133 +429,377 @@ pub trait OperationVerificationHelper {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns [`VerificationError`] if the operation at the current index
+    /// fails verification. On error, the cursor is not advanced. In the
+    /// current implementation, the callers are expected to abort since only
+    /// linear verification is supported.
+    pub fn next(
+        &mut self,
+        helper: &impl OperationVerificationHelper,
```
danielSanchezQ, line 848: Why dont we move the `&impl OperationVerificationHelper` to the `VerifiedOps` and we implement `Iterator` on it instead?

core/src/mantle/transactions/tx.rs
```diff
@@ -651,65 +942,100 @@ impl GasCalculator for SignedMantleTx {
...
+impl<State: VerificationState> StorageSize for SignedMantleTx<State> {
     fn storage_size(&self) -> usize {
         self.gas_storage_size() as usize
     }
 }
 
-impl Serialize for SignedMantleTx {
+#[derive(Serialize)]
+#[serde(rename = "SignedMantleTx")]
+struct SignedMantleTxHelper<'a> {
```
danielSanchezQ: Lets not call it Helper, lets call it `SignedMantleTxSerde`

core/src/mantle/transactions/tx.rs
```diff
@@ -651,65 +942,100 @@ impl GasCalculator for SignedMantleTx {
...
         } else {
             encode_signed_mantle_tx(self).serialize(serializer)
         }
     }
 }
 
-impl<'de> Deserialize<'de> for SignedMantleTx {
+#[derive(Deserialize)]
+#[serde(rename = "SignedMantleTx")]
+struct OwnedSignedMantleTxHelper {
```
danielSanchezQ: Likewise

core/src/mantle/transactions/tx.rs
```diff
@@ -651,65 +942,100 @@ impl GasCalculator for SignedMantleTx {
...
+// `Preverify` at that Service's boundary.
+impl<'de> Deserialize<'de> for SignedMantleTx<Preverified> {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let unverified_tx = SignedMantleTx::<Unverified>::deserialize(deserializer)?;
+        unverified_tx.preverify().map_err(serde::de::Error::custom)
+    }
+}
```
danielSanchezQ: This file is getting big, maybe we could move the serde parts to a different one?

core/src/mantle/mod.rs
```diff
@@ -68,11 +72,11 @@ pub trait AuthenticatedMantleTx: Transaction<Hash = TxHash> + GasCalculator + St
...
     ) -> Result<Gas, GasOverflow>;
+}
 
-    fn verify_ops_proofs_with_helper(
-        &self,
-        helper: &impl OperationVerificationHelper,
-    ) -> Result<(), VerificationError>;
+pub trait PreverifiedMantleTx: AuthenticatedMantleTx {
+    /// Returns the cursor to the verified operations in this transaction.
+    fn verified_ops(&self) -> VerifiedOps<'_>;
```
danielSanchezQ, line 80: 👍

ledger/src/mantle/sdp/mod.rs
```diff
@@ -691,53 +665,6 @@ mod tests {
         utxo_tree
     }
 
-    fn apply_declare_with_dummies(
```
danielSanchezQ, line 714: where did all this apply methods went?

core/src/mantle/transactions/tx.rs
thomaslavaur: Even the channel ones can be pre-verified under the condition that the public keys are stored among the transaction in the state

danielSanchezQ: Yes, what it means is that we can check the signature in the pre-verification and in the verification we can just check the membership alone (which is really cheap operation).

core/src/mantle/transactions/tx.rs
strinnityk: This is just a private helper. The initial idea planned for this to be used in a couple more places, but right now it's just used in `into_preverified`. Still, I think it makes `into_preverified` a tiny bit more readable.

core/src/mantle/transactions/tx.rs
strinnityk: Wdym?

core/src/mantle/transactions/tx.rs
strinnityk: To a different file, you mean?

danielSanchezQ: yes a `serde` one with all serde related stuff for example

core/src/mantle/transactions/tx.rs
danielSanchezQ: I mean to implement Iterator for VerifiedOps instead of this next method, but for that we need to embbeded the helper into the `VerifiedOps` or a third-party component that is the iterator.

ledger/src/mantle/sdp/mod.rs
strinnityk: Since the `try_apply_*` methods don't do verification anymore, all the signature scaffolding those dummy helpers were doing is not needed. Thus, they were removed and inlined.

core/src/mantle/transactions/tx.rs
strinnityk: That's done in a branch that follows this one. I'd like to get this one reviewed and merged before publishing it, though.

core/src/mantle/transactions/tx.rs
strinnityk: Interesting, I had discarded that idea since I assumed we needed to regenerate the helper at every step, but it might be possible not to. I'll give it a try.

strinnityk: We can't do it that way.
If we were to do that, the `OperationVerificationHelper` would be holding a reference to the `Ledger` as it was when `VerifiedOps` was instantiated. That is, when no operations were applied.
After the first operation was applied, the `Ledger`'s state would be different, but `VerifiedOps` would still be holding an out of date `Ledger`.

core/src/mantle/transactions/tx.rs
strinnityk: We'll be addressing that in a different PR, right?

core/src/mantle/transactions/tx.rs
danielSanchezQ: True, makes sense, thanks.

core/src/mantle/transactions/tx.rs
danielSanchezQ: Yes, I think it can be expanded in a different PR.

core/src/block/mod.rs
```diff
@@ -472,6 +472,8 @@ mod tests {
     struct TestMantleTx<const SIZE: usize>;
 
     impl<const SIZE: usize> Transaction for TestMantleTx<SIZE> {
+        //noinspection RsTypeCheck: The type is correct, but the linter is confused by
+        // the closure.
```
danielSanchezQ, line 476: wtf

core/src/mantle/transactions/genesis_tx.rs
```diff
@@ -163,10 +160,12 @@ fn valid_cryptarchia_inscription(
 }
 
 impl Transaction for GenesisTx {
+    //noinspection RsTypeCheck: The type is correct, but the linter is confused by
+    // the closure.
```
danielSanchezQ, line 164: what is this?

core/src/mantle/transactions/tx.rs
```diff
@@ -324,21 +344,20 @@ impl Transaction for MantleTx {
...
         signed_tx.mantle_tx
     }
 }
 
-// Deserializing here is dangerous, as it bypasses the verification without
-// confirmation.
-// TODO: Split entity into a system that allows for verification in different
-// stages.
+// TODO: Increase test coverage after type state refactor.
+//   The current tests behave just like the old code.
```
danielSanchezQ, line 358: lets create an issue unless is gonna be tackle shortly after.

core/src/mantle/transactions/tx.rs
```diff
@@ -403,133 +457,399 @@ pub trait OperationVerificationHelper {
...
-                    if !ok {
-                        return Err(VerificationError::InvalidProofOfClaim { op_index: idx });
-                    }
-                }
-                // Other operations are checked by the ledger or don't require verification here
-                _ => {
-                    // TODO: If the op and proof don't match, we are silently
-                    // delaying the error
-                    //  until tx execution.
+    // TODO: Might drop proofs after verification. This TODO is carried over from
```
danielSanchezQ, line 523: yeah, pre-verified proofs could be dropped. Lets open an issue for this too.

core/src/mantle/transactions/tx.rs
```diff
@@ -403,133 +457,399 @@ pub trait OperationVerificationHelper {
...
+                let is_verified = poc.verify(&LeaderClaimPublic {
+                    voucher_nullifier: leader_claim_op.voucher_nullifier.into(),
+                    voucher_root: leader_claim_op.rewards_root.into(),
+                    mantle_tx_hash: tx_hash.to_fr(),
+                });
+
+                if is_verified {
+                    Ok(())
+                } else {
+                    Err(VerificationError::InvalidProofOfClaim { op_index })
```
danielSanchezQ, line 547: Lets make this a method, one for each proof with same return signature

core/src/mantle/transactions/tx.rs
```diff
@@ -403,133 +457,399 @@ pub trait OperationVerificationHelper {
...
+    /// operation has a corresponding proof and that the proofs are of the
+    /// correct type.
+    ///
+    /// # Invariants
+    ///
+    /// - `ops` and `proofs` have the same length
+    /// - Each operation has a corresponding proof of the correct type
+    /// - [`InscriptionOp`](crate::mantle::ops::channel::inscribe::InscriptionOp)
+    ///   and [`LeaderClaimOp`](crate::mantle::ops::leader_claim::LeaderClaimOp) have valid signatures/proofs.
+    pub fn preverify(self) -> Result<SignedMantleTx<Preverified>, VerificationError> {
```
danielSanchezQ, line 591: wherever we call this it should be in a blocking tokio task, as this may be quite heavy. We could also at some point create a `verify_batched` that batches the groth16 proof verification process for the whole Tx. wdyt @thomaslavaur

core/src/mantle/transactions/tx.rs
```diff
@@ -403,133 +457,399 @@ pub trait OperationVerificationHelper {
...
+                    tx_hash,
+                    deposit_sig: channel_deposit_proof,
+                };
+                channel_deposit_op
+                    .validate(&channel_deposit_context)
+                    .map_err(VerificationError::ChannelVerificationError)
+            }
+            // TODO: Duplicate check. `verify_channel_withdraw` and `ChannelWithdrawOp::validate`,
+            //   both called in this arm, overlap in functionality. We probably need to purge the
+            //   `verify_channel_withdraw` function.
```
danielSanchezQ, line 681: If its not withing the next planned PRs lets add an issue.

core/src/mantle/mock.rs
```diff
@@ -39,6 +39,8 @@ impl<M: Serialize + DeserializeOwned + Clone> MockTransaction<M> {
 }
 
 impl<M: Serialize + DeserializeOwned + Clone> Transaction for MockTransaction<M> {
+    //noinspection RsTypeCheck: The type is correct, but the linter is confused by
+    // the closure.
```
danielSanchezQ, line 43: One more, im really curious.

core/src/mantle/mod.rs
```diff
@@ -86,6 +90,8 @@ pub trait GenesisTx: Transaction<Hash = TxHash> {
 }
 
 impl<T: Transaction> Transaction for &T {
+    //noinspection RsTypeCheck: The type is correct, but the linter is confused by
+    // the closure.
```
danielSanchezQ, line 95: lmao

ledger/src/mantle/sdp/mod.rs
```diff
@@ -404,6 +400,8 @@ impl SdpLedger {
         };
 
         // Validate SDP Declare
+        // TODO: Genesis has a different verification flow than `SignedMantleTx`.
+        // Refactor into a   type state.
```
danielSanchezQ, line 408: You mean to add a Genesis state for the Txs? Maybe nice. Lets open an issue.

ledger/src/mantle/mod.rs
```diff
@@ -187,18 +170,8 @@ impl LedgerState {
...
-        config_sigs: &ChannelMultiSigProof,
-        tx_hash: &TxHash,
         block_slot: Slot,
     ) -> Result<(Self, Vec<TxEvent>), Error> {
-        // Validate the SetKeys
-        config_op.validate(&ChannelConfigValidationContext {
-            channels: &self.channels,
-            tx_hash,
-            config_sigs,
-        })?;
```
danielSanchezQ, line 200: I think by removing this there is no other failure point, should we rename to apply_channel_set_keys instead?

ledger/src/lib.rs
```diff
@@ -133,20 +129,21 @@ where
...
         &self,
         id: Id,
         parent_id: Id,
         slot: Slot,
         proof: &LeaderProof,
-        txs: impl Iterator<Item = impl AuthenticatedMantleTx<Context = GasPrices>>,
+        txs: impl Iterator<Item = &'tx Tx>,
     ) -> Result<(Id, LedgerState, Events), LedgerError<Id>>
     where
+        Tx: PreverifiedMantleTx + 'tx + AuthenticatedMantleTx<Context = GasPrices>,
```
danielSanchezQ, line 149: Nice

ledger/src/lib.rs
```diff
@@ -522,210 +525,164 @@ impl LedgerState {
...
-                    return Err(LedgerError::UnsupportedOp);
-                }
-            }
+        loop {
+            let helper = MantleOperationVerificationHelper::new(
+                &self.mantle_ledger,
+                &self.cryptarchia_ledger,
+                config,
+            );
+            let Some(op) = verified_ops.next(&helper).transpose()? else {
```
danielSanchezQ, line 690: The stateful verification happen in this point right? I think its a bit confusing actually. We should maybe do it different:
* Rename next to `next_stateful_verified` (just a suggestion, open to names)
* Move the verification to ops and call from the helper
```rust
let Some(op) = verified_ops.next() else { break };
helper.verify_op(&op)?;
...
```

ledger/src/lib.rs
```diff
@@ -522,210 +525,164 @@ impl LedgerState {
...
+                break;
+            };
+            (self, balance, tx_events) = self.try_apply_op::<_, Constants>(
```
danielSanchezQ, line 693: We should check if after verification passes there is any other possible failure point, otherwise we should refactor the `try_*` methods into infallible ones. Maybe some yes, some not, im not sure right now.

nodes/node/binary/src/api/serializers/blocks.rs
```diff
@@ -3,20 +3,32 @@ use lb_chain_service::Slot;
...
 
-#[derive(Serialize)]
-#[serde(remote = "Block<SignedMantleTx>")]
-pub struct ApiBlockSerializer {
-    #[serde(getter = "Block::header")]
-    #[serde(with = "ApiHeaderSerializer")]
-    header: Header,
-    #[serde(getter = "Block::transactions_vec")]
-    #[serde(with = "crate::api::serializers::transactions::signed_api_transaction_vec")]
-    transactions: Vec<SignedMantleTx>,
```
danielSanchezQ, line 19: Why all this changes?

nodes/node/binary/src/api/serializers/blocks.rs
```diff
@@ -3,20 +3,32 @@ use lb_chain_service::Slot;
...
+    use crate::api::serializers::transactions::adapters::ApiTransactionsRefAdapter;
+
+    pub fn serialize<S, State: VerificationState>(
+        block: &Block<SignedMantleTx<State>>,
+        serializer: S,
+    ) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let mut state = serializer.serialize_struct("Block", 2)?;
```
danielSanchezQ: waht?

nodes/node/binary/src/api/serializers/transactions.rs
```diff
@@ -50,3 +61,13 @@ pub mod signed_api_transaction_vec {
...
+
+pub mod adapters {
+    use lb_core::mantle::{SignedMantleTx, transactions::states::VerificationState};
+    use serde::Serialize;
+
+    #[derive(Serialize)]
+    pub struct ApiTransactionsRefAdapter<'a, State: VerificationState>(
+        #[serde(with = "super::signed_api_transaction_vec")] pub &'a Vec<SignedMantleTx<State>>,
+    );
+}
```
danielSanchezQ: Do we have some checks to test there is no regressions? (it should not)

nodes/node/binary/src/api/handlers.rs
```diff
@@ -764,14 +768,14 @@ where
...
+                    SignedMantleTx<Preverified>,
+                    <SignedMantleTx<Preverified> as Transaction>::Hash,
                     RuntimeServiceId,
                 >,
                 Mempool<
                     HeaderId,
-                    SignedMantleTx,
-                    <SignedMantleTx as Transaction>::Hash,
+                    SignedMantleTx<Preverified>,
+                    <SignedMantleTx<Preverified> as Transaction>::Hash,
```
danielSanchezQ, line 778: Uhm, is this the correct play? Ill probably check later on it directly

nodes/node/binary/src/config/mempool/mod.rs
```diff
@@ -31,7 +34,7 @@ impl ServiceConfig {
 
         TxMempoolSettings {
             network_adapter: Libp2pNetworkAdapterSettings {
-                id: SignedMantleTx::hash,
+                id: SignedMantleTx::<Preverified>::hash,
```
danielSanchezQ, line 37: Ok, does pre-verification happen in deserialization here?

services/chain/chain-service/src/lib.rs
```diff
@@ -363,13 +363,13 @@ impl Cryptarchia {
     }
 
     /// Try to apply a block to the chain.
-    fn try_apply_block<Tx>(
+    fn try_apply_block<'tx, Tx>(
```
danielSanchezQ, line 366: why the lifetime tho?

danielSanchezQ (changes requested):

core/src/block/mod.rs
strinnityk: It's a hint to disable a Rust Rover check that mistakenly shows up as an error.

core/src/mantle/transactions/tx.rs
strinnityk: I left this to be addressed when the `Op` type state is done.

core/src/mantle/transactions/tx.rs
strinnityk: Added as part of the `Op` type state issue.

ledger/src/mantle/sdp/mod.rs
strinnityk: Wasn't thinking in a solution yet, but your proposal does make sense
An alternative that came to mind would be to have the `GenesisTx` be its own type state so we decouple the two flows.

ledger/src/mantle/mod.rs
strinnityk: Thought about that. Didn't go for it since the `fn`s still return a `Result`. Do you think it's worth changing?

ledger/src/lib.rs
strinnityk: It gets even cleaner in following PRs.

ledger/src/lib.rs
strinnityk: Personally it makes sense to me, although I'm biased, ofc.
I think your suggestion makes a lot more sense after we go through the `Op` type state. After that, the `VerifiedOps` is no longer needed. We can just `impl Iterator for SignedMantleTx<Preverified>` and have `Ledger` require `Op<Verified>`, which you'd retrieve like:
```rust
let preverified_tx = _;
let preverified_op = preverified_tx.next();
let verified_op = preverified_op.verify(&helper);
```
Also, I'm thinking the `VerifiedOps` is a bit unnecessary. We could've used a third state. Might update it later.

nodes/node/binary/src/api/serializers/blocks.rs
strinnityk: Due to `serde`'s compatibility with types with nested generics. The remote feature didn't work anymore.

nodes/node/binary/src/api/serializers/blocks.rs
strinnityk: Same as above. `serde`'s remote can't handle nested generics (`Block<Tx<VerificatonState>>`).

nodes/node/binary/src/api/serializers/transactions.rs
strinnityk: The tests we have in place should cover this, but they might not be enough. I'll add a point to the coverage issue.

nodes/node/binary/src/api/handlers.rs
strinnityk: It's the simpler play.
I'm in favour of explicit transitions at the boundaries, but that would've made a huge PR into humongous.
I'll create an issue to address this as well.

services/chain/chain-service/src/lib.rs
strinnityk: The compiler was complaining it needed it here: `Tx: PreverifiedMantleTx + 'tx + ...`

nodes/node/binary/src/config/mempool/mod.rs
strinnityk: Correct. We have two `Deserialize`: into `Unverified` and into `Preverified`.
The `Preverified` one is the same as the `Unverified` but with a `preverify` call.

core/src/block/mod.rs
danielSanchezQ: uhm, ok

nodes/node/binary/src/api/handlers.rs
danielSanchezQ: No, i was just wondering if it was the proper state, but it is because pre-verification happens on wire deresialization so its fine.

services/chain/chain-service/src/lib.rs
danielSanchezQ: This feels unnecesary to be honest. But its fine if we are sure the compiler complains.

nodes/node/binary/src/api/serializers/blocks.rs
danielSanchezQ: Are we sure about that? Cant we bound the old ApiblockSerializer with a txstate generic (`ApiBlockSerializer<TxState>`) and use that generic in the remote? I think there must be a better option that raw/manual serialization here. Even duplicating (macro) for each type could be ok if we reuse the current helpers. Its a safer move imo.

core/src/mantle/transactions/tx.rs
```diff
@@ -403,133 +457,399 @@ pub trait OperationVerificationHelper {
...
+            ops_proofs,
+            state: PhantomData,
+        }
+    }
+
+    #[expect(
+        clippy::too_many_lines,
+        reason = "The match arms are long due to the validation context construction. Split later."
+    )]
+    pub(crate) fn verify_stateful_op(
```
thomaslavaur, line 631: That sounds very repetitive. I may break all your changes but can't we break this state machinery at the operation level as well? So a PreVerified mantle transaction is a mantle transaction with all its ops Preverified?

For example if you take the inscription. the .validate() verify again the signature while you already checked it in `verify_stateless_op()`... I feel like then you need to update the `validate()` method to only take as input a preverified Operation and then you can assume the proof was checked against the PublicKey so you just have to check that the channel PublicKey is the one authorized to sign at the last moment.

This can be generalized by Op and I think it would be eaiser to manage if it's implemented at the op level

thomaslavaur: Because leaving all the validate() like this brings no improvements compare to the stateless version

ledger/src/cryptarchia/mod.rs
```diff
@@ -460,21 +459,8 @@ impl LedgerState {
 
     pub fn try_apply_transfer<Id, Constants: GasConstants>(
```
thomaslavaur, line 460: Then  you don't need this. You can just call the `execute()` method of the op if you don't validate it

ledger/src/mantle/sdp/mod.rs
thomaslavaur, line 435: Same if you remove the validation part you can directly call the execution

thomaslavaur: Or at least they would need a renaming

ledger/src/mantle/mod.rs
thomaslavaur: the name is also wrong it should be `config` instead of set_keys

nodes/node/binary/src/api/serializers/blocks.rs
strinnityk: Well, I did try but failed. The AI was also suggesting a similar approach too. If you have any idea I'd love to try, I don't really like this solution.

core/src/mantle/transactions/tx.rs
strinnityk: Yes, that's the plan. We kept it as-is here to avoid growing this PR.
In a follow-up task we'll transform `Op` into a typestate so it tracks its own verification. That way not only is it self-contained, but it also allows for a clear split between the stateless and stateful checks.

thomaslavaur: So why not leaving the state_less_verification empty for now just as a placeholder for next PRs?

danielSanchezQ: > So why not leaving the state_less_verification empty for now just as a placeholder for next PRs?

the idea is to leave everything as it is current behavior, more refactors will come after this iirc. Ops will have the same machinery in the (near) future.

nodes/node/binary/src/api/serializers/blocks.rs
danielSanchezQ: ok, if have no regressions should be ok, I may take a look at it, because its ugly as 💩 .

ledger/src/mantle/sdp/mod.rs
danielSanchezQ: Yeah, this is what i suggested as well. @strinnityk ideally we move full verification of tx on the outer and then there is no try, but just (safe) execute. iirc its what we have, just the naming is not updated.

nodes/node/binary/src/api/serializers/blocks.rs
strinnityk: I spent a bit of time on this and arrived at something better, I think, but a bit more manual than desired.
TL;DR: Plain structs built via `From<&Foreign>`, each with one small handwritten `serialize` function doing what `remote` would've generated if it could reach through the nesting.
Not as magic as the derive macro, but readable enough.

core/src/mantle/transactions/states.rs
```diff
@@ -0,0 +1,12 @@
+/// Marker trait for valid verification states of a transaction
+pub trait VerificationState {}
+
+/// Unverified state of a Transaction
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Unverified;
+impl VerificationState for Unverified {}
+
+/// Partially verified state of a Transaction
```
youngjoon-lee, line 9: It'd be also good to clearly mention that this type is for transactions that has passed "stateless" checks.

core/src/mantle/transactions/tx.rs
```diff
@@ -403,133 +457,399 @@ pub trait OperationVerificationHelper {
...
+    /// # Errors
+    ///
+    /// Returns [`VerificationError`] if the operation at the current index
+    /// fails verification. On error, the cursor is not advanced. In the
+    /// current implementation, the callers are expected to abort since only
+    /// linear verification is supported.
+    pub fn next(
+        &mut self,
+        helper: &impl OperationVerificationHelper,
+    ) -> Option<Result<&'tx Op, VerificationError>> {
```
youngjoon-lee, line 849: nit: I think, we can rename `VerifiedOps` to `PreverifiedOps` because `new` accepts a preverified tx (not verified tx). Then, we can rename `next` to `verfiy_next`. But not so important. Just a nit.

core/src/mantle/mod.rs
```diff
@@ -44,11 +44,15 @@ pub trait Transaction {
     fn as_signing(&self) -> Vec<u8>;
 }
 
+// TODO: Purge out gas fns
```
youngjoon-lee, line 48: Sorry, I couldn't understand why gas fns should be purged out. We now get these values from other contexts (such as helper)?

ledger/src/lib.rs
```diff
@@ -300,7 +300,7 @@ impl LedgerState {
         self.cryptarchia_ledger
             .update_fee_window(window_index, total_fee_burned);
 
-        // Then compute the amount of the block rewards
+        // Then compute the number of block rewards
```
youngjoon-lee: I think, "amount" is clearer.

zone-sdk/src/sequencer/zone_sequencer.rs
```diff
@@ -139,6 +139,8 @@ pub struct ZoneSequencer<Node> {
     request_rx: mpsc::UnboundedReceiver<ActorRequest>,
 }
 
+pub type PublishReceipt = (PublishResult, SequencerCheckpoint);
```
youngjoon-lee, line 137: I don't think this change is really necessary in this PR, but I agree with having a type alias for this tuple. I guess @pradovic is also happy with it, since it's a small change.

core/src/mantle/transactions/tx.rs
youngjoon-lee: In the following PR, we need to make `ChannelWithdrawOp::validate` safer. Currently, it access to a key by index using `[i]`, which can panic. But, it doesn't actually panic in the current codebase, thanks to `verify_channel_withdraw` which uses `helper.get_key_from_channel_at_index` (that uses `get(i).ok_or(...)` internally). 
If we're going to drop `verify_channel_withdraw` in the following PR, we also need to change `[i]` to something safer in the `ChannelWithdrawOp::validate`.

core/src/mantle/transactions/tx.rs
```diff
@@ -403,133 +457,399 @@ pub trait OperationVerificationHelper {
...
+impl SignedMantleTx<Preverified> {
+    /// Creates a new `SignedMantleTx<Preverified>` without performing any
+    /// verification.
+    ///
+    /// This function is intended for
+    /// [`GenesisTx`](crate::mantle::transactions::genesis_tx::GenesisTx) and
+    /// testing purposes only.
+    #[must_use]
+    #[doc(hidden)]
+    pub const fn new_trusted(mantle_tx: MantleTx, ops_proofs: Vec<OpProof>) -> Self {
```
youngjoon-lee:
```suggestion
    pub(crate) const fn new_trusted(mantle_tx: MantleTx, ops_proofs: Vec<OpProof>) -> Self {
```
This could be `pub(crate)`.

youngjoon-lee (approved):
I focused on the logic correctness, because it seems that you and Daniel have discussed the typestate architecture sufficiently. Also, it looks good to me. 
I've read all the discussions in this PR and understand that we're going to have the following improvements in the upcoming PRs. I hope I've understood the direction correctly :)
- Op-level typestate that removes double-verification
- Renaming `try_apply_*` -> `apply_*` in ledger
- GenesisTx typestate, or any better way.
I'm approving this PR because I agree with those improvements.

I left a few minor comments.

ledger/src/mantle/sdp/mod.rs
strinnityk: Yes, naming is out of date. I just tried to avoid adding (yet) more things to the PR.

danielSanchezQ (approved):
Looking forward to next related PRs.

core/src/mantle/transactions/tx.rs
strinnityk: I would argue `VerifiedOps` makes more sense, since that's what you're consuming with `next`. The verification is hidden behind the iterator, and it's almost invisible except for the need of the `OperationVerificationHelper` (which we might be able to make go away, or at least simplify, in the near future).

core/src/mantle/mod.rs
strinnityk: You nailed it. We're currently providing those methods via other traits. Having those here just adds another level of (unnecessary) indirection.
This is already tackled in a following (not yet published) branch.

zone-sdk/src/sequencer/zone_sequencer.rs
strinnityk: It was not strictly necessary, but I needed to change the signature of a symbol. Adding this alias and updating the remaining usages made that change clearer.

core/src/mantle/transactions/tx.rs
strinnityk: Unfortunately, there's an external call to this: `tools/.../consensus.rs`.

PR #3129, opened 2026-07-14
test: add tokio console profiling with cucumber integration

hansieodendaal:

- Added tokio console profiling to the node, enabled when compiling with feature flag `--features tokio-console`.
- Updated main repo README.md with tokio console profiling additions.
- Added tokio profiling to cucumber tests select-able via `And I will have tokio console profile nodes:`

Note: A follow-up task will be to name tokio tasks after some analysis has been done for easier identification in tokio console, for example we could hve these:
```
network-libp2p-swarm
blend-core-swarm
blend-edge-swarm
chain-network-tip-poll
chain-leader-winning-slot-search
mempool-broadcast-item
wallet-generate-poc
storage-rocksdb-bulk-write
```
A sample implementation will look like:
```rust
let task = async move {
    swarm_handler.run(initial_peers).await;
};

#[cfg(feature = "tokio-console")]
tokio::task::Builder::new()
    .name("network-libp2p-swarm")
    .spawn(task)
    .expect("failed to spawn network-libp2p-swarm");

#[cfg(not(feature = "tokio-console"))]
tokio::spawn(task);
```

**Note:** [Overwatch/pull/141](https://github.com/logos-co/Overwatch/pull/141) implements opt-in Tokio task naming for Overwatch-managed service and state-handler tasks. This would reduce naming service-related tasks in the follow-up task.

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/tests/testing_framework/src/framework/local/provisioning.rs
+++ b/tests/testing_framework/src/framework/local/provisioning.rs
@@ -37,8 +37,10 @@ use tracing::debug;
 
 use crate::{
     LOG_LEVEL,
+    configs::deployment::NodeBinaryProfile,
     diagnostics::{record_system_monitor_event, register_system_monitor_output_file},
     env as tf_env,
+    env::{remove_default_env, replace_default_env},
     framework::LbcEnv,
     node::{
         DeploymentPlan, NodeHttpClient, NodePlan,
@@ -257,9 +259,11 @@ fn build_node_launch_spec(
     let deployment_path = dir.join(DEPLOYMENT_CONFIG_FILE);
     let time_backend =
         env::var("LOGOS_BLOCKCHAIN_TIME_BACKEND").unwrap_or_else(|_| "monotonic".to_owned());
+    let node_binary_profile =
+        NodeBinaryProfile::from_string(&env::var("NODE_BINARY_PROFILE").unwrap_or_default());
 
     Ok(LaunchSpec {
-        binary: node_binary_provider().resolve()?,
+        binary: node_binary_provider(&node_binary_profile).resolve()?,
         files: vec![
             launch_file(USER_CONFIG_FILE, user_yaml.into_bytes()),
             launch_file(DEPLOYMENT_CONFIG_FILE, deployment_yaml.into_bytes()),
@@ -283,11 +287,16 @@ fn launch_file(relative_path: &str, contents: Vec<u8>) -> LaunchFile {
     }
 }
 
-fn node_binary_provider() -> BinaryProviderRef {
-    Arc::new(FallbackBinaryProvider::new([
+fn node_binary_provider(node_binary_profile: &NodeBinaryProfile) -> BinaryProviderRef {
+    let providers: [BinaryProviderRef; 2] = [
         Arc::new(EnvBinaryProvider::new("LOGOS_BLOCKCHAIN_NODE_BIN")),
-        default_node_binary_provider(),
-    ]))
+        match node_binary_profile {
+            NodeBinaryProfile::Normal => default_node_binary_provider(),
+            NodeBinaryProfile::TokioConsole => tokio_console_node_binary_provider(),
+        },
+    ];
+
+    Arc::new(FallbackBinaryProvider::new(providers))
 }
 
 fn default_node_binary_provider() -> BinaryProviderRef {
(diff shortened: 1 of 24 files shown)


tests/testing_framework/src/node/configs/deployment.rs
```diff
@@ -235,6 +280,15 @@ impl DeploymentBuilder {
         let nodes = build_node_plans(node_count, &ids, &node_configs)?;
         self.config.genesis_block = Some(genesis_block);
 
+        if self.config.node_binary_profile == NodeBinaryProfile::Normal {
```
andrussal, line 283: After the first deployment, this becomes a no-op because set_default_env only writes when NODE_BINARY_PROFILE is unset. If a normal scenario runs first, a later Tokio-console scenario will still use the normal profile, and vice versa. Is that intended?

tests/testing_framework/src/framework/local/provisioning.rs
```diff
@@ -318,6 +327,42 @@ fn release_node_binary_path() -> PathBuf {
         .join("logos-blockchain-node")
 }
 
+fn tokio_console_node_binary_provider() -> BinaryProviderRef {
+    if running_in_ci() {
+        Arc::new(PathBinaryProvider::new(release_node_binary_path()))
+    } else {
+        let current = replace_default_env("RUSTFLAGS", "--cfg tokio_unstable");
```
andrussal: Should we pass RUSTFLAGS through the build provider instead? The Cargo command is only executed when the provider is resolved, but RUSTFLAGS is restored before that happens, so this build will not receive --cfg tokio_unstable.

tests/testing_framework/src/framework/local/provisioning.rs
```diff
@@ -283,11 +287,16 @@ fn launch_file(relative_path: &str, contents: Vec<u8>) -> LaunchFile {
     }
 }
 
-fn node_binary_provider() -> BinaryProviderRef {
-    Arc::new(FallbackBinaryProvider::new([
+fn node_binary_provider(node_binary_profile: &NodeBinaryProfile) -> BinaryProviderRef {
```
andrussal, line 316: Would it be safer to use separate binary environment variables for the two profiles? Currently, LOGOS_BLOCKCHAIN_NODE_BIN is always tried first, so a regular binary can accidentally satisfy a Tokio-console request even when it was not built with the required feature. Something like:

```rust
fn node_binary_provider(profile: &NodeBinaryProfile) -> BinaryProviderRef {
    let providers: [BinaryProviderRef; 2] = match profile {
        NodeBinaryProfile::Normal => [
            Arc::new(EnvBinaryProvider::new("LOGOS_BLOCKCHAIN_NODE_BIN")),
            default_node_binary_provider(),
        ],
        NodeBinaryProfile::TokioConsole => [
            Arc::new(EnvBinaryProvider::new(
                "LOGOS_BLOCKCHAIN_TOKIO_CONSOLE_NODE_BIN",
            )),
            tokio_console_node_binary_provider(),
        ],
    };

    Arc::new(FallbackBinaryProvider::new(providers))
}
```

andrussal (approved):
LGTM!

(Left few comments to consider)

tests/testing_framework/src/node/configs/deployment.rs
hansieodendaal: Yes, the presumption was one would either run tokio-console profiling or not, but not a bad idea to scope this for the specific scenario,

hansieodendaal: The only meaningful way tokio console profiling with cucumber will be used is when the targeted scenario(s) is run and kept alive for a fairly long time so the tokio console UI can be observed. It is blocked from running in CI. That being said, any amount of scenarios can be run with a tokio console compiled binary, and only the nodes that are selected will be able to link to the tokio console UI.

tests/testing_framework/src/framework/local/provisioning.rs
hansieodendaal: The idea is not to be able to swap binaries for each scenario, but to use the same binary if multiple scenarios are selected. As explained above, doing a tokio console profiling run is best done with a focused scenario,

The user can start a tokio console profiling scenario, and if a binary is not specified, the appropriate `release-profile` binary will be built. If the wrong binary is specified, the tokio console profiling server will not run inside the node and the scenario will error.

tests/testing_framework/src/framework/local/provisioning.rs
hansieodendaal: Good find! Fixed.

PR #3135, opened 2026-07-16
feat(api): add `discovered_peers` to `/network/info`

youngjoon-lee:

Monitoring how many peers have been discovered is useful. The existing `connected_peers` field doesn't tell us that, because libp2p doesn't actively establish connections to the discovered peers until it becomes necessary. 

Yes

@youngjoon-lee 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/services/network/src/backends/libp2p/swarm/mod.rs
+++ b/services/network/src/backends/libp2p/swarm/mod.rs
@@ -368,7 +368,7 @@ const fn exp_backoff(retry: usize) -> Duration {
 
 #[cfg(test)]
 mod tests {
-    use std::{net::Ipv4Addr, sync::Once, time::Instant};
+    use std::{collections::HashSet, net::Ipv4Addr, sync::Once, time::Instant};
 
     use lb_libp2p::{
         libp2p::core::{ConnectedPoint, Endpoint, transport::PortUse},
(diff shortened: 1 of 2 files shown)


services/network/src/backends/libp2p/swarm/mod.rs
```diff
@@ -266,6 +266,13 @@ impl<R: Clone + Send + RngCore + 'static> SwarmHandler<R> {
                 self.connect(dial);
             }
             NetworkCommand::Info { reply } => {
+                let discovered_peers: Vec<PeerId> = self
+                    .swarm
+                    .kademlia_discovered_peers()
+                    .into_iter()
+                    .map(|peer_info| peer_info.peer_id)
+                    .collect();
+                let n_discovered_peers = discovered_peers.len();
```
hansieodendaal, line 275: Are we sure that the Kademlia helper used here will not the local node as well? I think `kademlia_discovered_peers()` returns every Kademlia entry that has an address.

We could possibly filter out `*swarm.local_peer_id()` from `kademlia_discovered_peers()` before building `discovered_peers` and `n_discovered_peers`.

hansieodendaal (commented):
This will be a great addition.

Maybe we should add a targeted unit/integration test, for example a focused test covering an empty routing table and a routing table containing the local peer plus remote peers.

danielSanchezQ (approved):

services/network/src/backends/libp2p/swarm/mod.rs
youngjoon-lee: The local_peer_id is excluded by libp2p kademlia: https://github.com/libp2p/rust-libp2p/blob/9698607c72ea7b8f5e661963ab0e24384aef1e4e/protocols/kad/src/kbucket.rs#L142-L146.

Since this PR just implements API, I don't want to change the implementation of `lb_libp2p::behaviour::kademlia::Swarm::kademlia_discovered_peers`. I want the API to show exactly what the other crates get and use from the `kademlia_discovered_peers` function.

Also, I didn't add a test for an empty set because it's so trivial.

hansieodendaal (approved):
LGTM

services/network/src/backends/libp2p/swarm/mod.rs
youngjoon-lee: @hansieodendaal 
Just added tests c7b376e8d. Thanks!

PR #3145, opened 2026-07-17
feat(mantle): Enable PoS for channels

thomaslavaur:

This PR implement the [Channel PoS participation RFC](<https://github.com/logos-co/logos-lips/pull/364>)

Yes

@thomaslavaur 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/core/src/header/mod.rs
+++ b/core/src/header/mod.rs
@@ -274,3 +274,302 @@ fn test_serde() {
         HeaderId([0; 32])
     );
 }
+
+/// Block-root / `HeaderId` test-vector generator.
+///
+/// This module does not assert library behaviour: it *emits* reference test
+/// vectors for the block `block_root` computation and the resulting
+/// [`HeaderId`], so that alternative implementations (e.g. the nim
+/// implementation) can be checked for conformance against the canonical Rust
+/// encoding.
+///
+/// It emits three vectors:
+/// - **empty block**: `block_root` of a block with no transactions (`Merkle([])
+///   == [0u8; 32]`);
+/// - **one tx per op kind**: a block whose transactions each carry a single
+///   operation, one per distinct mantle [`Op`] variant; for it we print every
+///   leaf (`tx_hash`) and the resulting `block_root`;
+/// - **`HeaderId`**: a header reusing the previous `block_root`, with a fixed
+///   parent, slot and a deterministic genesis proof of leadership, for which we
+///   print the inputs and the resulting `HeaderId`.
+///
+/// `block_root = Merkle(blake2b256(b"MANTLE_TXHASH_V1" || tx_bytes) leaves)`,
+/// where the Merkle tree pads the leaf set to the next power of two with
+/// all-zero leaves and hashes inner nodes as `blake2b256(left || right)`.
+///
+/// `HeaderId (block_id) = blake2b256( b"BLOCK_ID_V1" || bedrock_version (1B)
+/// ||` `parent_block (32B) || slot_le (8B) || block_root (32B) ||
+/// leader_voucher` `(32B) || entropy_contribution (32B) || proof (128B) ||
+/// leader_key (32B) )`.
+///
+/// The test is `#[ignore]`d so it is skipped by `cargo test --all-features`.
+/// Run it on demand with:
+/// `cargo test -p logos-blockchain-core block_root_test_vectors -- --ignored
+/// --nocapture`
+#[cfg(test)]
+mod block_root_test_vectors {
+    use lb_blend_proofs::{
+        quota::{PROOF_OF_QUOTA_SIZE, VerifiedProofOfQuota},
+        selection::{PROOF_OF_SELECTION_SIZE, VerifiedProofOfSelection},
+    };
+    use lb_cryptarchia_engine::Epoch;
+    use lb_key_management_system_keys::keys::ZkPublicKey;
+    use lb_poseidon2::Fr;
+
+    use super::*;
+    use crate::{
+        mantle::{
+            MantleTx, Note, Op, Transaction as _,
+            channel::{SlotTimeframe, SlotTimeout},
+            ledger::{Inputs, NoteId, Outputs},
+            ops::{
+                channel::{
+                    ChannelId, Ed25519PublicKey, MsgId,
+                    channel_transfer::ChannelTransferOp,
+                    config::{ChannelConfigOp, Keys},
+                    deposit::{DepositOp, Metadata},
+                    inscribe::InscriptionOp,
+                    withdraw::ChannelWithdrawOp,
+                },
+                leader_claim::{LeaderClaimOp, VoucherCm},
+                transfer::TransferOp,
+            },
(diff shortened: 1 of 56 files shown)


deployment/tui-zone/src/run_commands/run_balance.rs
```diff
@@ -1,18 +1,19 @@
...
 };
 
-pub(crate) async fn run_state_balance(args: StateArgs) -> RunResult<()> {
-    let channel_id = resolve_channel_id(&args.node_key)?;
-    let node = node_client(&args.node_key.node_url)?;
-    let channel_state = query_channel_state(&node, channel_id).await;
-    print_channel_balance("balance", &channel_id, channel_state.as_ref());
-    Ok(())
-}
+// TODO: support channel note tracking in the TUI. A channel's balance is now
```
danielSanchezQ, line 8: Lets open an issue for this. Not sure we should maintain this anymore tbh.

wallet/src/lib.rs
```diff
@@ -90,10 +90,10 @@ pub enum WalletOp {
     Lock(NoteId),
     /// Create the reward note.
     LeaderClaim(Utxo),
-    /// Spend the deposited notes.
+    /// The deposited notes are now owned by the channel and can no longer be
+    /// spent by the wallet.
+    /// TODO: support channel note tracking and channel note `PoS` participation
```
danielSanchezQ, line 95: Lets open an issue for this as well

danielSanchezQ (approved):
Awesome job @thomaslavaur !!, this is much better! Lets open the proper issues so we can improve/implement in further PRs!

wallet/src/lib.rs
thomaslavaur: Opened: https://github.com/logos-blockchain/logos-blockchain/issues/3146

core/src/mantle/ops/channel/channel_transfer.rs
```diff
@@ -0,0 +1,124 @@
...
+            return Err(Error::ThresholdUnmet {
+                channel_id: self.channel_id,
+                threshold: channel.transfer_threshold,
+                actual: signatures.len(),
+            });
+        }
+
+        // Check the signatures
+        for sig in signatures {
+            if channel.accredited_keys[sig.channel_key_index as usize]
```
youngjoon-lee: Instead of `[...]`, I'd like to suggest to use `get(..).ok_or(...)` to avoid panics.

wallet/src/lib.rs
```diff
@@ -90,10 +90,10 @@ pub enum WalletOp {
...
     ChannelDeposit(Vec<NoteId>),
-    /// Create the notes withdrawn from the channel.
-    ChannelWithdraw(Vec<Utxo>),
```
youngjoon-lee, line 96: I think, we shouldn't remove this operation because wallet should be able to start using these withdrawn notes.

youngjoon-lee: Hm, okay. Now I understand why you removed this `ChannelWithdraw` from `WalletOp`. Now, the withdraw operation doesn't have `outputs`. 

Nevertheless, I think that wallet should track ChannelTransfer and ChannelWithdraw as well as ChannelDeposit. When ChannelDeposit is detected, wallet should just "mark" the note as unspendable. Then, the wallet can revert it to "spendable" once the note is withdrawn. Similarly, the wallet should track ChannelTransfer — removing input notes and adding output notes if they're owned by the wallet. 

I guess, that's what you mentioned in the TODO comment and #3146, correct?
If so, I can take it over if you want.

core/src/mantle/transactions/tx.rs
```diff
@@ -533,23 +544,22 @@ impl SignedMantleTx {
...
     op_index: usize,
 ) -> Result<(), VerificationError> {
-    let channel_id = &operation.channel_id;
-    let withdraw_threshold = helper.get_channel_withdraw_threshold(channel_id)?;
+    let transfer_threshold = helper.get_channel_transfer_threshold(channel_id)?;
 
     let signatures = proof.signatures();
     let signatures_len = signatures.len();
-    if signatures_len < withdraw_threshold as usize {
+    if signatures_len < transfer_threshold as usize {
```
youngjoon-lee: I know this is not from this PR, but I think it's better to use `!=` instead of `<`, because other functions (such as [`ChannelWithdrawOp::validate`](https://github.com/logos-blockchain/logos-blockchain/blob/b1a7d3e756a9c418ec7e121a2932272fac5b2f43/core/src/mantle/ops/channel/withdraw.rs#L69)) use `!=`.

wallet/src/lib.rs
youngjoon-lee: Also, I think it's better to postpone merging this PR until we update wallet and zone-sdk. Until then, this PR is incomplete, IMO.

core/src/header/mod.rs
```diff
@@ -274,3 +274,302 @@ fn test_serde() {
...
+/// ||` `parent_block (32B) || slot_le (8B) || block_root (32B) ||
+/// leader_voucher` `(32B) || entropy_contribution (32B) || proof (128B) ||
+/// leader_key (32B) )`.
+///
+/// The test is `#[ignore]`d so it is skipped by `cargo test --all-features`.
+/// Run it on demand with:
+/// `cargo test -p logos-blockchain-core block_root_test_vectors -- --ignored
+/// --nocapture`
+#[cfg(test)]
+mod block_root_test_vectors {
```
youngjoon-lee, line 310: Just to confirm my reading of two test vectors added in this PR:

These test vectors are just reference-vector generators, to give alternative implementations (e.g. Nim) a stable reference to check against, not to assert correctness of anything added/changed in this PR. Is that right?

wallet/src/lib.rs
thomaslavaur: This works for user wallets but we also need a « sequencer wallet » that track channel notes under their eddsa key to fund channel transfer and channel withdraw.

the issue is exactly this + the sequencer wallet

core/src/header/mod.rs
thomaslavaur: Yes. They are used to generate the values in the specs

youngjoon-lee (approved):
Thank you! I think we can merge this PR, even though there are some missing pieces to be implemented in the wallet, as I mentioned in https://github.com/logos-blockchain/logos-blockchain/pull/3145#discussion_r3612142604. By merging this PR, we'll lose a functionality of wallet that tracks notes withdrawn from channels. But, we can re-implement it according to the new spec after merging this PR ;)

core/src/mantle/ops/channel/channel_transfer.rs
thomaslavaur: true. I also changed it on the channel withdraw then

core/src/mantle/transactions/tx.rs
thomaslavaur: true Updated!

PR #3150, opened 2026-07-20
chore: convert vec of op proof into bounded vec

hansieodendaal:

Converted `pub ops_proofs: Vec<OpProof>,` to `pub ops_proofs: OpsProofs,` with 
```rust
pub type OpsProofs = UpperBoundedVec<OpProof, MAX_OPS_PER_TX>;
```
 and changed all relevant call sites.

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/core/src/block/genesis.rs
+++ b/core/src/block/genesis.rs
@@ -66,11 +66,11 @@ where
     I: IntoIterator<Item = N>,
     N: Into<Note>,
 {
-    let notes: Vec<Note> = notes.into_iter().map(Into::into).collect();
-    if notes.is_empty() {
+    let mut notes_iter = notes.into_iter().map(Into::into).peekable();
+    if notes_iter.peek().is_none() {
         return Err(Error::EmptyNotes);
     }
-    BoundedOutputs::try_from(notes).map_err(|error| map_notes_bounded_error(&error))
+    BoundedOutputs::try_from_iter(notes_iter).map_err(|error| map_notes_bounded_error(&error))
 }
 
 fn push_note(mut notes: BoundedOutputs, note: Note) -> Result<BoundedOutputs> {
@@ -80,21 +80,17 @@ fn push_note(mut notes: BoundedOutputs, note: Note) -> Result<BoundedOutputs> {
     Ok(notes)
 }
 
-fn extend_non_empty_notes<I, N>(mut existing: BoundedOutputs, notes: I) -> Result<BoundedOutputs>
+fn extend_non_empty_notes<I, N>(existing: BoundedOutputs, notes: I) -> Result<BoundedOutputs>
 where
     I: IntoIterator<Item = N>,
     N: Into<Note>,
 {
-    let mut iter = notes.into_iter().peekable();
-    if iter.peek().is_none() {
+    let mut notes_iter = notes.into_iter().map(Into::into).peekable();
+    if notes_iter.peek().is_none() {
         return Err(Error::EmptyNotes);
     }
-    for note in iter.map(Into::into) {
-        existing
-            .try_push(note)
-            .map_err(|error| map_notes_bounded_error(&error))?;
-    }
-    Ok(existing)
+    BoundedOutputs::try_from_iter(existing.into_iter().chain(notes_iter))
+        .map_err(|error| map_notes_bounded_error(&error))
 }
 
 /// A [`Block`] whose transactions are all [`GenesisTx`] values.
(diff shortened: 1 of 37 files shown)


core/src/block/genesis.rs
```diff
@@ -1317,20 +1314,22 @@ mod tests {
...
+                    )),
+                    Op::SDPDeclare(_) => OpProof::ZkAndEd25519Sigs {
+                        zk_sig: ZkSignature::new(CompressedGroth16Proof::from_bytes(&[0u8; 128])),
+                        ed25519_sig: Ed25519Signature::zero(),
+                    },
+                    other => unreachable!("unexpected genesis op in tests: {}", other.as_str()),
+                })
+                .collect::<Vec<_>>(),
+        )
+        .expect("genesis transaction proofs are bounded");
```
danielSanchezQ: I would expect a `try_collect` here instead of collecting and then `try_from`, can we implement `TryFrom` for an iterator instead? 🤔 so we avoid actually collecting. We could bound to a sizehinted one

hansieodendaal: Yeah, that was a roundabout way, fixed in 322104e

danielSanchezQ: Yeah, ti was fixed where it was used in the test but we kept the *API usage*, would be nice that we actually implement something like:

```rust
impl<T> TryFrom<T> for OpsProofs where T: IntoIterator<Op> { ... }
```

We can use `size_hint` to evaluate the limits if its present, if not just try to consume it in an intermediary `Vec` as you did before. But at least it would be encapsulated in the `TryFrom` implementation.

hansieodendaal: That exact proposal would have conflicted with the standard library’s blanket implementation of  `impl<T, U> TryFrom<U> for T where U: Into<T>`

I added `BoundedOutputs::try_from_iter(` instead
```rust
    pub fn try_from_iter<I>(iterable: I) -> Result<Self, BoundedError>
    where
        I: IntoIterator<Item = T>,
    {
```
and used it at the appropriate call sites.

danielSanchezQ: I dont see the changes @hansieodendaal, maybe you forgot to commit and push :question:

hansieodendaal: > I dont see the changes @hansieodendaal, maybe you forgot to commit and push ❓

Ah sorry, I made a bad merge, now pushed

utils/src/bounded/vec.rs
```diff
@@ -100,6 +100,24 @@ impl<T, const MIN: usize, const MAX: usize> Bounded<Vec<T>, MIN, MAX> {
...
+        for value in iterable {
+            if values.len() == MAX {
+                return Err(BoundedError::TooManyItems {
+                    count: MAX + 1,
+                    max: MAX,
+                });
+            }
+            values.push(value);
+        }
+        Self::try_from(values)
```
danielSanchezQ, line 119: We can use the `size_hint` as suggested before:
```rust
let iter = iterable.into_iter();
if let (_, Some(hint)) = iter.size_hint() && hint > MAX {
    return Err(BoundedError::TooManyItems { count: hint, max: MAX });
}
Self::try_from(iter.collect::<Vec<_>>())
```

Or 
```rust
let iter = iterable.into_iter();
if let (_, Some(hint)) = iter.size_hint() && hint > MAX {
    return Err(BoundedError::TooManyItems { count: hint, max: MAX });
}
for value in iterable {
            if values.len() < MAX {
                return Err(BoundedError::TooManyItems {
                    count: MAX + 1,
                    max: MAX,
                });
            }
            values.push(value);
        }
        Self::try_from(values)
```

is `MAX` an inclusive bound? `[0, MAX]` or not `[0, MAX)`?

hansieodendaal: We could not use `size_hint` due to the following:

`Iterator::size_hint()` returns `(lower_bound, upper_bound)` which means 
```
lower_bound <= actual remaining items <= upper_bound
```

Therefore, an upper bound larger than MAX does not prove there are too many items. It only says there could be that many. A concrete standard-library example is:
```rust
let iter = (0..100).filter(|value| *value < 3);

assert_eq!(iter.size_hint(), (0, Some(100)));
assert_eq!(iter.count(), 3);
```

danielSanchezQ: So, if there could be that many we are safe unless its not known right?
The only non-safe is if the iterator is buggy. 

But fair enough, your simple implementation works in all cases anyway, so lets go with it.

danielSanchezQ (approved):

PR #3162, opened 2026-07-22
refactor(chain-service): reorganize the main event loop into phases

### LINKED ISSUE #2956: refactor(chain-service): simplify waiting for genesis time
## Issues

Currently chain service has the `chain_start_timer` which ticks once the genesis time reaches. Until then, the chain service suppresses the block processing. This behavior was added by PR #2685.

That's good, but it's confusing that the chain service also delays starting PBP if genesis time isn't reached yet. There are too many `if` statements that branch based on `chain_start_time`, and it's not easy to understand what each of them is for.

On the other hand, the chain network service starts IBD regardless of the genesis time and send blocks to the chain service, hoping that the blocks will be processed. But, the chain service ignore those blocks. This behavior even cause a [panic](https://github.com/logos-blockchain/logos-blockchain/blob/c390c219a76e10a86042ccf887356cacccaa6d18/services/chain/chain-service/src/lib.rs#L936) because there are two event loops in the chain service, and the [1st event loop](https://github.com/logos-blockchain/logos-blockchain/blob/c390c219a76e10a86042ccf887356cacccaa6d18/services/chain/chain-service/src/lib.rs#L698) ignores all `ApplyBlock` requests. The ignored requests are tossed to the 2nd event loop which panics on any `ApplyBlock` requests.

## Impact

Not very high because it's a rare case to run the network with the future genesis time.

Also, if there are nodes running at time `t < genesis_time`, it means that they are nodes bootstrapping the network from scratch. Even if their chain network service starts IBD (even though blocks will be ignored by the chain service), IBD will be completed immediately because all nodes in the network have only the genesis block. So, the panic mentioned above won't be triggered in production. But, the code should be still cleaned up.

## Suggestions

From what I understood from PR #2685, the main purpose of adding `chain_start_timer` is introducing the [`AwaitingStart`](https://github.com/logos-blockchain/logos-blockchain/blob/c390c219a76e10a86042ccf887356cacccaa6d18/services/chain/chain-service/src/lib.rs#L734) mode returned for `cryptarchia/info` requests. I agree that it's necessary. But, instead of having the `chain_start_timer` variable with complex `if` statements, I'd like to suggest the following structure:
- Change the time service to yield the `Slot(0)` only when the genesis time reaches.
  - Currently it yields the `Slot(0)` immediately even if `now < genesis_time`, and yields `Slot(1)` at the correct time past the genesis time. It's clearer to yield `Slot(0)` only when the genesis time reaches because that's the definition of `Slot(0)`. 
- Until then, the time service shouldn't mark it as ready because it doesn't have any slot to yield yet.
- The chain service should wait until the time service is ready, and the chain network service should wait until the chain service becomes ready, as they are already doing.
  - In other words, they must become operational only after the genesis time reaches. We don't need to have `if` statements to enable/disable specific operations in those services.
- But then, chain service won't provide responses for `cryptarchia/info` requests until it becomes ready. To workaround it, we can introduce the pre-ready mode for the chain service, so it can respond only to `cryptarchia/info` requests, with `AwaitingStart` responses.

After this changes, all downstreams (such as blend) that subscribe to slots will need a "first slot tick may take a while" contract.

youngjoon-lee:

Refactored the chain service's complex main event loop into explicit phases.

Why? The main event loop was handling multiple channels and timers. It was very hard to read (without AI :)). Some of events becomes unnecessary after IBD or PBP completes, but we were putting everything into the main event loop. Also, we have introduce the `AwaitingStart` concept https://github.com/logos-blockchain/logos-blockchain/pull/2685 to handle the future genesis time, but putting it in the main event loop causes some potential bugs/panics (that I'm not listing here). 

How? I introduced 4 phases. They advance in one direction. Each phase handles only events relevant to it.

None of chain service's behavior changed. Only potential bugs, which could theoretically happen but didn't because of other services' behavior, have been eliminated. 

Breaking changes? Only the response of `GET /cryptarchia/info` has changed:
- Before (`mode` mixed service mode and engine state):
  ```json
  {
    "cryptarchia_info": { "lib": "0x…", "lib_slot": 10, "tip": "0x…",
  "slot": 42, "height": 30 },
    "mode": { "Started": "Online" }        // or "AwaitingStart"
  }
  ```
- After (phase = service phase; state = Cryptarchia engine state, now separate):
  ```json
  {
    "cryptarchia_info": { "lib": "0x…", "lib_slot": 10, "tip": "0x…",
  "slot": 42, "height": 30, "state": "Online" },  // `state` added
    "phase": "Following"                   // AwaitingGenesisTime |
  InitialBlockDownload | ProlongedBootstrapPeriod | Following
  }
  ```

The c-binding is not changed. It still returns the old-style response which delivers limited information (`NotStarted | Bootstrapping | Online`), but I think it's fine for now.

For reviewers, I believe this PR is not hard to read. I minimized diffs.
I didn't introduce any verbose typestate concept because the phase transition is one-direction and each phase requires exactly the same ingredients. Instead, I took the simplest approach which is short and easy to read. 
Please focus on the `run` function of chain service, first. You can also review two commits separately. I refactored the service message type first and then refactored the service. 

EDIT: After getting an approval from Daniel, I introduced typestate: 7f511bddd5a9f2ce01f306c72d8cb1dd9c0a3ff0.

yes

boostrapping spec: https://lip.logos.co/blockchain/raw/cryptarchia-v1-bootstr-sync.html#prolonged-bootstrap-period

diagram: https://app.notion.com/p/Cryptarchia-v1-Bootstrapping-Synchronization-MIGRATED-1e58f96fb65c804aae85f84a2e94392d?source=copy_link#1e58f96fb65c8159834dfe8b434624d1

@youngjoon-lee 
spec: @youngjoon-lee 

yes

no

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/services/chain/chain-service/src/lib.rs
+++ b/services/chain/chain-service/src/lib.rs
@@ -51,17 +51,14 @@ use lb_storage_service::{
 use lb_time_service::TimeService;
 use overwatch::{
     DynError, OpaqueServiceResourcesHandle,
-    services::{AsServiceId, ServiceCore, ServiceData, state::StateUpdater},
+    services::{AsServiceId, ServiceCore, ServiceData, relay::InboundRelay, state::StateUpdater},
 };
 use relays::BroadcastRelay;
 use serde::{Deserialize, Serialize, de::DeserializeOwned};
 use serde_with::serde_as;
 use thiserror::Error;
 use time::OffsetDateTime;
-use tokio::{
-    sync::{broadcast, mpsc, oneshot, watch},
-    time::Instant,
-};
+use tokio::sync::{broadcast, mpsc, oneshot, watch};
 use tracing::{Level, debug, error, info, instrument, span, trace, warn};
 use tracing_futures::Instrument as _;
 
@@ -113,6 +110,8 @@ pub enum Error {
     HeaderIdNotFound(HeaderId),
     #[error("Parent header ID not found for child={0}")]
     ParentIdNotFound(HeaderId),
+    #[error("Awaiting genesis time")]
+    AwaitingGenesisTime,
 }
 
 struct InitializedCryptarchia {
@@ -210,10 +209,17 @@ pub enum Query {
 pub(crate) type HeaderIdStream =
     Pin<Box<dyn Stream<Item = Result<HeaderId, Error>> + Send + 'static>>;
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
-pub enum ChainServiceMode {
-    AwaitingStart,
-    Started(State),
+/// The phase of the chain service, which advances one-directionally
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
+pub enum ChainServicePhase {
+    /// The genesis time is in the future. Only read-only queries are served.
+    AwaitingGenesisTime,
+    /// Applying blocks while waiting for Initial Block Download to complete.
+    InitialBlockDownload,
+    /// The Prolonged Bootstrap Period is running.
+    ProlongedBootstrapPeriod,
+    /// Following the chain in real time
+    Following,
 }
 
 #[serde_as]
@@ -221,7 +227,7 @@ pub enum ChainServiceMode {
 #[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
 pub struct ChainServiceInfo {
     pub cryptarchia_info: CryptarchiaInfo,
-    pub mode: ChainServiceMode,
+    pub phase: ChainServicePhase,
 }
 
 #[serde_as]
@@ -233,6 +239,7 @@ pub struct CryptarchiaInfo {
     pub tip: HeaderId,
     pub slot: Slot,
     pub height: u64,
+    pub state: State,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
@@ -358,6 +365,7 @@ impl Cryptarchia {
             tip: tip_branch.id(),
             slot: tip_branch.slot(),
             height: tip_branch.length(),
+            state: *self.state(),
         }
     }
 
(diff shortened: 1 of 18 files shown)


services/chain/chain-service/src/lib.rs
```diff
@@ -191,18 +204,25 @@ pub enum ConsensusMsg<Tx> {
...
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
+pub enum ChainServicePhase {
+    /// The genesis time is in the future. Only read-only queries are served.
+    AwaitingGenesisTime,
+    /// Applying blocks while waiting for Initial Block Download to complete.
+    InitialBlockDownload,
+    /// The Prolonged Bootstrap Period is running.
+    ProlongedBootstrapPeriod,
+    /// Following the chain in real time
+    Following,
```
danielSanchezQ: This is simple enough, but imo we could make it better If we want to enforce advancing one-directionally.

We could keep this enum as a tag `ChainServicePhaseTag` type state, then have something link:

```rust
trait ChainServicePhase {};
...
struct AwaitingGenesisTime;
struct InitialBlockDownload;
...
impl ChainServicePhase for AwaitingGenesisTime {}
...
pub struct ChainServicePhase<Phase: ChainServicePhase> {
    
}
```

That way we can compose something like this in the main:

```rust
        AwaitingGenesisTime::new(&mut self, Self::create_genesis_timer(starting_state))
        .process_awaiting().await
        .process_download(ibd_skipped).await
        .process_prolonged_bootstrap_period().await
        .process_following.await;
```

danielSanchezQ (approved):
Its nice. I suggested something that may be better. I'll let you consider the effort for it. Thanks!

services/chain/chain-service/src/lib.rs
youngjoon-lee: Yes. We can introduce the real type state. Now that you reviewed on the service flow in this PR, I will try this system. Some functions will be moved across modules, but I think it will be fine since you already reviewed this

youngjoon-lee: @danielSanchezQ I defined typestate for phases: [7f511bd](https://github.com/logos-blockchain/logos-blockchain/pull/3162/commits/7f511bddd5a9f2ce01f306c72d8cb1dd9c0a3ff0). Many functions have been moved to the new modules, but no function bodies have changed. Please see the summary below:
- `src/service/mod.rs`: Defines a new struct `Service` which defines all common operations for all phases. Some operations are even used by `src/lib.rs` to initialize the chain before starting the service, but I will refactor it later.
- `src/service/phases/`: Defines a new trait `Phase` and structs that implement the trait. The overall structure follows your suggestion above.
- `src/lib.rs`: Only functions for initializing (recoverying) the chain remain in this module. As I mentioned, I will clean up this later.

PR #3164, opened 2026-07-22
feat(merkle): Dynamic merkle tree

danielSanchezQ:

* Extracts the dynamic merkle tree implementation from the utxo tree
* Made the implementation generic over a `MerkleHasher` trait
* Migrate curren utxo tree to use new generic implementation

Its mostly moving things around otherwise yes.

@danielSanchezQ @thomaslavaur 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.


dynamic-merkle/src/lib.rs
```diff
@@ -1,41 +1,97 @@
...
+    /// Root of a fully-empty subtree of the given `height`.
+    ///
+    /// Implement with [`empty_subtree_root`] to get a cached implementation.
+    fn empty_subtree_root(height: usize) -> Self::Hash;
+}
 
-const EMPTY_VALUE: Fr = <Fr as AdditiveGroup>::ZERO;
+/// Height of the tree excluding the root, i.e. the length of every Merkle path
+/// and the base-2 logarithm of the tree's leaf capacity (`2^32` items).
+pub const TREE_HEIGHT_EXCEPT_ROOT: usize = 32;
```
thomaslavaur, line 47: should it be EXPECT instead of EXCEPT ?

dynamic-merkle/src/lib.rs
```diff
@@ -1,41 +1,97 @@
...
+            })[height]
         }
-        hashes
-    })[height]
+    };
 }
 
 #[derive(::serde::Serialize, ::serde::Deserialize, Clone, Debug, PartialEq, Eq)]
-enum Node<Item> {
+enum Node<Item, Hash> {
```
thomaslavaur, line 90: would benefit a small comment to not mix Hash and Hasher imo

dynamic-merkle/src/lib.rs
```diff
@@ -248,38 +296,75 @@ impl<Item: AsRef<Fr>> Node<Item> {
 /// insertions. Compared to a MPT, the height of this tree is predictable and
 /// bounded by the number of items, allowing for efficient and simple proof of
 /// memberships for `PoL`.
```
thomaslavaur: nit: this is very specific we can state: "for example for PoL" instead maybe?

thomaslavaur (commented):
I think it's pretty good looks better imo and so we can do the same for any object

dynamic-merkle/src/lib.rs
danielSanchezQ: nono, this is rigth (tho confusing)

dynamic-merkle/src/lib.rs
thomaslavaur: just to recall Hash is the type of the leaves and nodes not the hash function

danielSanchezQ: This is pretty private, I think its fine, as it is bound where it is used

dynamic-merkle/src/lib.rs
danielSanchezQ: copy pasted, will fix

thomaslavaur (approved):

youngjoon-lee (approved):

PR #3167, opened 2026-07-22
chore: add exact-size zk verifier json input

hansieodendaal:

Added exact-size ZK verifier JSON input to `ZkSignVerifierInputsJson` using a lower + upper bounded vec with bounds of exactly 33 for both.

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


thomaslavaur (approved):
lgtm

zk/proofs/zksign/src/public.rs
```diff
@@ -18,32 +19,34 @@ impl ZkSignVerifierInputs {
 
 #[derive(Deserialize)]
 #[serde(transparent)]
-pub struct ZkSignVerifierInputsJson(Vec<Groth16InputDeser>);
+pub struct ZkSignVerifierInputsJson(BoundedVec<Groth16InputDeser, 33, 33>);
```
ntn-x2, line 22: Is it a requirement to have it as a length-prefixed vector? Why not using an array directly?

hansieodendaal: A 33-element fixed-size array will allow less than 33 elements - this has to be exactly 33.

ntn-x2: No. A fixed-sized array always have the same number of elements. What you are talking about is a vector. Am I missing something?

hansieodendaal: We want a container that can only have 33 elements, not variable 0 to 33 elements.

danielSanchezQ: > We want a container that can only have 33 elements, not variable 0 to 33 elements.

33? 🤔

danielSanchezQ: image
we want 32 not 33?

thomaslavaur: no we want 33 because it's the 32 public keys of the notes + the mantle tx hash to bind the proof (the msg here)

ntn-x2: @hansieodendaal that's what a fixed-size array gives you. Rhe only difference is that a bounded vec would prefix its length, an array would not. Other than that, if we always need 33 elements, we should use an array, which is NOT a variable-size container.

danielSanchezQ: > no we want 33 because it's the 32 public keys of the notes + the mantle tx hash to bind the proof (the msg here)

Ah ok, thanks for the explanation, I was thinking on the pks only.

hansieodendaal: > @hansieodendaal that's what a fixed-size array gives you. Rhe only difference is that a bounded vec would prefix its length, an array would not. Other than that, if we always need 33 elements, we should use an array, which is NOT a variable-size container.

Ah yes @ntn-x2 , I conflated a fixed-size array with a vector having a maximum capacity. Thanks for pointing out using a fixed-size array directly, I will make a small follow-up PR for his.

**Edit:** Fixed in #3191

PR #3173, opened 2026-07-23
feat(c-bindings): implement missing ffi Node impl

pradovic:

This PR adds missing ffi for full Node implementation.

It also updates the existing stream call with stream closed termination (NULL pointer).

It also updates existing stream calls plus cryptarchia info with missing fields.

After this PR we will need to update the consumers of stream handling (new JSON and null termination).

Yes

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


c-bindings/src/api/subscriptions.rs
```diff
@@ -1,154 +1,170 @@
...
     logging, return_error_if_null_pointer,
 };
 
-#[derive(Serialize)]
-#[serde(rename = "SignedMantleTx")]
-#[derive(Clone)]
-pub struct TxWithId {
-    id: TxHash,
-    #[serde(flatten)]
-    tx: SignedMantleTx<Unverified>,
```
danielSanchezQ, line 32: This have value, we embedded the TxId which is necessary for uis and other stuff, but we completely removed it here. Was it replaced somehow somewhere else? I have not seen it.

danielSanchezQ (commented):
Overall ok, but we are breaking API instead of expanding it. This currently breaks our UI module. Can we just expand but keep the blocks call format for the UI to keep functioning as it was? Or just move the format for the new API calls so its a matter of changing the call? TxIds are really needed there.

c-bindings/src/api/subscriptions.rs
pradovic: The new payload does carry every tx id, just at a different path. The HTTP wire format we now emit serializes each transaction as {"mantle_tx": {"hash": …, "ops": …}, "ops_proofs": …}. The id is right there as mantle_tx.hash (see nodes/node/binary/src/api/serializers/transactions.rs:10-11, the serializer injects the computed hash). The old TxWithId existed because c-bindings predated that serializer.
Master's HTTP API embeds the hash natively now. Does that makes sense?

pradovic:
@danielSanchezQ Thanks for the fast review!  I get you point, I was thinking about the similar think. Worth noting the module pins `liblogos_blockchain.dylib`, so nothing breaks in production. The break lands when binding version is bumper and this PR breaks ABI anyway (CryptarchiaInfo gained lib_slot).

However we could maybe:
- Reinstate subscribe_to_new_blocks exactly as it is on master (old shape, TxWithId and all) — UI module keeps working with zero changes.
- Our new-schema call gets its own name (subscribe_to_processed_blocks), alongside subscribe_to_lib_blocks.
- Mark the old call deprecated in its doc comment, pointing at the new one and at the mantle_tx.hash path for tx ids — with removal once the UI migrates

In this case CryptarchiaInfo gaining lib_slot mid-struct is currently an ABI break too, but can move it to the end of the struct, so field offsets stay stable and it becomes a pure expansion as well. With that, the PR is fully additive: old binaries keep working, and the new calls/fields are opt-in at recompile.

danielSanchezQ (approved):
Be aware that we will have to asses the changes at module and UI level. Go for it, thanks!

bacv (approved):
Looks good, module built and starts a node, we'll still need to expose the new calls through the module itself.

PR #3176, opened 2026-07-23
chore(ci): Use a merge group for heavy CI

danielSanchezQ:
## Enable merge queue for `master`

Moves the heavy, self-hosted Cucumber and end-to-end integration test suites off the per-PR path and into GitHub's merge queue. These suites run after approval and enqueueing, against the temporary merge-group commit, rather than on every PR push or stale-branch update.

The nextest unit-test suite remains part of the normal pull-request CI and is not rerun in the merge queue.

### Why

With **Require branches to be up to date before merging** enabled, pull requests may need to be updated before merging, retriggering expensive CI. Slow runs against commits that have already become stale continue consuming self-hosted runner capacity while newer work waits.

The merge queue validates the proposed merge against the latest `master`, including any pull requests ahead of it in the queue, before allowing it to merge.

### Changes

- **`code-check.yml`**
  - Continues to run on `pull_request` and pushes to `master`.
  - The self-hosted nextest unit-test jobs remain part of pull-request CI.
  - It does not run again for `merge_group` events.

- **`cucumber-integration-tests.yml`**
  - Triggers on both `pull_request` and `merge_group`.
  - On pull requests, the heavy self-hosted matrix is skipped and the lightweight **`Cucumber result`** check succeeds without occupying a self-hosted runner.
  - In the merge queue, both Linux and macOS Cucumber jobs run against `github.sha`, which is the temporary merge-group commit.
  - **`Cucumber result`** succeeds when at least one OS lane passes and fails when both lanes fail.

- **`end-to-end-integration-tests.yml`**
  - Triggers on both `pull_request` and `merge_group`.
  - On pull requests, the heavy self-hosted matrix is skipped and the lightweight **`E2E result`** check succeeds without occupying a self-hosted runner.
  - In the merge queue, both Linux and macOS end-to-end jobs run against `github.sha`.
  - **`E2E result`** succeeds when at least one OS lane passes and fails when both lanes fail.

### Required repository settings

These settings must be configured separately for `master`:

- Enable **Require merge queue**.
- Disable **Require branches to be up to date before merging**.
- Keep the required approval rules enabled.
- Require these stable merge-queue checks:
  - **`Cucumber result`**
  - **`E2E result`**
- Remove the old OS-specific Cucumber and end-to-end checks from the required-check list.
- Do not require checks from `code-check.yml` for the merge queue unless that workflow is also updated to report those checks for `merge_group` events. This includes the two **`Run test suite`** nextest checks.

> The Cucumber and end-to-end suites still run on both supported operating systems in the merge queue. A failure on one OS is tolerated because these suites can be flaky; the queued merge is blocked and removed from the queue only when both OS lanes fail.
>
> Nextest continues to provide feedback on the pull request but is not rerun as part of merge-group validation.

A bit of ci magic is around tbh

@danielSanchezQ @hansieodendaal @andrussal 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.


andrussal:
Only thing I would consider changing: instead of requiring both macOS and Linux cucumber runs to pass, could we somehow require that at least one of them passes? Otherwise occasional cucumber flakiness can easily undermine the whole solution.

.github/workflows/code-check.yml
```diff
@@ -118,7 +123,9 @@ jobs:
 
   tests:
     name: Run test suite
-    if: ${{ !startsWith(github.event.pull_request.title, '[WIP]') && !contains(github.event.label.name,  'DO NOT MERGE') }}
+    # Heavy, self-hosted suite: run ONLY inside the merge queue so it executes once,
+    # after the PR is approved and queued - never on unreviewed pushes or stale rebases.
+    if: ${{ github.event_name == 'merge_group' }}
```
hansieodendaal: These nextest unit tests are generally fast enough, and adds immediate feedback. In my opinion, these should not go on the merge queue.

danielSanchezQ:
> Only thing I would consider changing: instead of requiring both macOS and Linux cucumber runs to pass, could we somehow require that at least one of them passes? Otherwise occasional cucumber flakiness can easily undermine the whole solution.

The problem with that is that if its not required, then its run in each PR in parallel, which is somewhat what we are trying to avoid here.
We would have to try this, and if it doesn't work find an alternate solution.

hansieodendaal (commented):
Looks good in general.

I agree with @andrussal that we should only require one successful e2e and one successful cucumber before merge, not successful for both on Linux and MacOS runners.

andrussal:
> > Only thing I would consider changing: instead of requiring both macOS and Linux cucumber runs to pass, could we somehow require that at least one of them passes? Otherwise occasional cucumber flakiness can easily undermine the whole solution.
> 
> The problem with that is that if its not required, then its run in each PR in parallel, which is somewhat what we are trying to avoid here. We would have to try this, and if it doesn't work find an alternate solution.

I meant something different actually: still required and still queue-only, just not both. maybe something like an extra job that passes if either OS run passed, and that one is the required check?

hansieodendaal:
> I meant something different actually: still required and still queue-only, just not both. maybe something like an extra job that passes if either OS run passed, and that one is the required check?

So this?

normal CI pass + approval -> merge queue -> 1 of e2e pass + 1 of cucumber pass -> merged

andrussal:
> > I meant something different actually: still required and still queue-only, just not both. maybe something like an extra job that passes if either OS run passed, and that one is the required check?
> 
> So this?
> 
> normal CI pass + approval -> merge queue -> 1 of e2e pass + 1 of cucumber pass -> merged

played around with the either-OS idea, something like this might work (not 100% sure, would need a real run to confirm):

each OS run marks itself as passed with a last step, and the job exposes that to other jobs — a failed run never reaches the step, so its value stays empty:

```yaml
  tests:
    # ...existing job unchanged...
    outputs:
      self-hosted-linux: ${{ steps.report.outputs.self-hosted-linux }}
      self-hosted-macos: ${{ steps.report.outputs.self-hosted-macos }}
    steps:
      # ...existing steps unchanged...
      - name: Report success
        id: report
        run: echo "${{ matrix.platform.label }}=success" >> "$GITHUB_OUTPUT"
```

then a new job fails only if both stayed empty, i.e. both OSes failed:

```yaml
  cucumber-result:
    name: Cucumber result
    needs: tests
    if: always()
    runs-on: ubuntu-latest
    steps:
      - name: Check that at least one OS passed
        run: |
          if [ "${{ needs.tests.outputs.self-hosted-linux }}" = "success" ] || \
             [ "${{ needs.tests.outputs.self-hosted-macos }}" = "success" ]; then
            exit 0
          fi
          exit 1
```

and in branch protection "Cucumber result" becomes the required check instead of the two OS ones. both OS runs still happen in the queue and stay visible, only the merge gate loosens

hansieodendaal:
> played around with the either-OS idea, something like this might work (not 100% sure, would need a real run to confirm):

Nice - we could also extend this to the e2e matrix.

andrussal (approved):

.github/workflows/end-to-end-integration-tests.yml
```diff
@@ -1,7 +1,8 @@
+# Heavy, self-hosted suite: runs ONLY inside the merge queue. A PR reaches the
+# queue only after it satisfies branch protection (required approvals), so this
+# expensive CI executes once, on the queued merge commit, after review.
 on:
-  pull_request:
-  push:
-    branches: [ master ]
+  merge_group:
```
hansieodendaal, line 6: I think we need the same `Check that at least one OS passed` step/logic here for `End-to-end integration tests`, but it could be fine-tuned in a follow-up PR.

hansieodendaal (approved):
LGTM

**Note:** The nextest unit tests are generally fast enough and adds immediate feedback. In my opinion, these should not go on the merge queue, but it could be fine-tuned in a follow-up PR.

.github/workflows/cucumber-integration-tests.yml
```diff
@@ -202,3 +203,34 @@ jobs:
...
+        run: echo "${{ matrix.platform.label }}=success" >> "$GITHUB_OUTPUT"
+
+  # This is the stable required check for the Cucumber suite.
+  # On pull requests it succeeds without starting either self-hosted runner.
+  # In the merge queue it passes if at least one OS passed, so a single flaky
+  # OS run does not eject the PR. Both OS runs still execute and remain visible.
+  cucumber-result:
+    name: Cucumber result
+    needs: tests
+    if: always()
```
ntn-x2, line 220: I think there were some issues with manually cancelling workflows that have a `if: always()` condition. The right way of doing it would be `if: !interrupted()`: https://stackoverflow.com/questions/77799456/workflow-stuck-and-not-cancellable

.github/workflows/end-to-end-integration-tests.yml
```diff
@@ -166,3 +167,34 @@ jobs:
...
+        run: echo "${{ matrix.platform.label }}=success" >> "$GITHUB_OUTPUT"
+
+  # This is the stable required check for the end-to-end suite.
+  # On pull requests it succeeds without starting either self-hosted runner.
+  # In the merge queue it passes if at least one OS passed, so a single flaky
+  # OS run does not eject the PR. Both OS runs still execute and remain visible.
+  e2e-result:
+    name: E2E result
+    needs: tests
+    if: always()
```
ntn-x2, line 184: Same as above: https://stackoverflow.com/questions/77799456/workflow-stuck-and-not-cancellable.

.github/workflows/cucumber-integration-tests.yml
hansieodendaal: Good catch! The cancellation concern is valid.

Fixed in #3193

.github/workflows/end-to-end-integration-tests.yml
hansieodendaal: ditto #3193

PR #3179, opened 2026-07-23
fix(ffi): don't install a process::exit panic hook in the embedded cdylib

dlipicar:
## Problem

A host process embedding the node via the `logos_blockchain` cdylib dies with **SIGABRT** on a routine, read-only query. Observed in Logos Basecamp: `logos_host` aborts inside `get_cryptarchia_info`, which the blockchain UI polls **every 2 seconds** while a node is running.

Symptom in the log is a double panic:

```
panicked at library/std/src/thread/local.rs:428:25:
thread panicked while processing panic. aborting.
FATAL: module 'blockchain_module' crashed (signal 6)
```

## Root cause

`run_node_from_config` unconditionally installed the `log_and_exit_hook` panic hook, which calls `std::process::exit(1)`. That is correct for the standalone daemon — but the same function also backs the cdylib's `start_lb_node`, and `std::panic::set_hook` is **process-global**. So once a host started an embedded node, a panic on *any* thread terminated the entire host.

It gets worse than a plain exit. `exit()` runs the host's C++ static destructors, which call back into the node's own `stop_node`/`shutdown_node` FFI — on a thread whose thread-locals are already being destroyed. That second panic during panic is what turns it into an `abort`:

```
get_cryptarchia_info (FFI)
  -> panic
  -> log_and_exit_hook -> std::process::exit(1)
     -> __cxa_finalize_ranges          [C runtime]
        -> ~LogosBlockchainModule()    [host C++]
           -> stop_node (FFI)          [re-enters tokio mid-teardown]
              -> panic during panic -> abort()
```

### What actually panics

The panic originates upstream, in Overwatch (`ae887f41`, `overwatch/src/overwatch/handle.rs:78-86`):

```rust
let Ok(()) = self.send(OverwatchCommand::Relay(..)).await else {
    unreachable!("Service relay should always be available");
};
```

`send` is a tokio mpsc send that returns `Err` as soon as the receiver is dropped — i.e. once Overwatch begins shutting down. So this is a race: any `cryptarchia_info` poll landing during/after shutdown hits that `unreachable!`. (`handle.rs:118` has the same issue in `status_watcher`.) Worth fixing there too by returning a `RelayError`, but the embedded node should not be fatally sensitive to it either way.

## Changes

**1. Don't install a `process::exit` panic hook from library code.** The hook moves out of `run_node_from_config` into a documented `install_panic_hook()`, called only from the standalone binary. Behaviour there is unchanged — still installed only on the node-run path, not for CLI subcommands.

**2. Stop panics unwinding across the FFI boundary.** New `guard_ffi` helper wraps a body in `catch_unwind` and converts a panic into an `OperationStatus` (`RuntimeError`) instead of unwinding out of an `extern "C"` fn (which aborts). Applied to all **27 FFI operation entry points**.

The pure `free_*` deallocators and the `is_ok`/`is_error` accessors are intentionally left unwrapped: they run no node logic.

Both parts are needed. (1) alone would still abort at the `extern "C"` boundary; (2) alone wouldn't help while the exit hook runs first. Together, the query returns a recoverable error and the host stays alive.

## Verification

Toolchain 1.96.0, matching `rust-toolchain.toml`:

- `cargo check -p logos-blockchain-c -p logos-blockchain-node` — exit 0
- `cargo clippy -p logos-blockchain-c -p logos-blockchain-node --all-targets` — exit 0, **zero warnings in any touched file** (`--all-targets` also type-checks test code)

Applies identically to the `0.2.0` tag if a `0.2.x` backport is wanted.

### DIFF
--- a/c-bindings/src/api/blend.rs
+++ b/c-bindings/src/api/blend.rs
@@ -82,29 +82,31 @@ pub unsafe extern "C" fn blend_join_as_core_node(
     locator: *const c_char,
     locked_note_id: *const u8,
 ) -> FfiStatusResult<DeclarationId> {
-    return_error_if_null_pointer!(node);
-    return_error_if_null_pointer!(locator);
-    return_error_if_null_pointer!(locked_note_id);
+    crate::macros::guard_ffi(|| {
+        return_error_if_null_pointer!(node);
+        return_error_if_null_pointer!(locator);
+        return_error_if_null_pointer!(locked_note_id);
 
-    let locator = unwrap_or_return_error!(unsafe { parse_locator(locator) });
-    let locked_note_id = unwrap_or_return_error!(unsafe { parse_locked_note_id(locked_note_id) });
+        let locator = unwrap_or_return_error!(unsafe { parse_locator(locator) });
+        let locked_note_id = unwrap_or_return_error!(unsafe { parse_locked_note_id(locked_note_id) });
 
-    let node = unsafe { &*node };
+        let node = unsafe { &*node };
 
-    let result: StatusResult<sdp::DeclarationId> = node.get_runtime_handle().block_on(async {
-        lb_api_service::http::blend::blend_join_network::<
-            BlendService<RuntimeServiceId>,
-            BlendBroadcastSettings<RuntimeServiceId>,
-            RuntimeServiceId,
-        >(node.get_overwatch_handle(), locator, locked_note_id)
-        .await
-        .map_err(|error| {
-            OperationStatus::error(
-                OperationStatusCode::RelayError,
-                format!("Failed to join blend network: {error}"),
-            )
-        })
-    });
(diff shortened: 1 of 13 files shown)


strinnityk:
We're addressing two things here which are worth separating:
1. Moving the panic hook out of `run_node_from_config` into `install_panic_hook()`, and keep it main-only seems to make sense.
2. `guard_ffi` is handling issues without taking into account what our `panic`s mean: They are invariants that tell us the node's internal state is not safe/trustworthy anymore.
	Instead, I think we should add a poisoned flag into the FFI. On the first panic `guard_ffi` catches, set it. Every call after that short-circuits to a dedicated error (e.g. `Unrecoverable`). Without it, we've just moved the abort from the process to a silent zombie node.

Do you want us to take this over?

dlipicar:
@strinnityk

> Do you want us to take this over?

 by all means!

PR #3180, opened 2026-07-24
test(cucumber): recover wallet scanner from reorged snapshot seed tips

andrussal:
## Problem

The wallet snapshot only stores the scanner's latest applied tip. The tip is captured while nodes are still producing blocks, so a reorg between capture and shutdown can leave the snapshot pointing at a block that got discarded. After restart the rescan looks for that tip in the canonical block stream and never finds it. There was no recovery path for this case, so the scanner just retried the same lookup forever and wallet steps timed out.

Seen in CI on both linux and macos runners — same error repeating for ~90 min across all retries of the snapshot scenarios.

## Fix

The scanner already keeps a few recent checkpoints in memory for reorg rollback, so the snapshot now stores the last 8 of those (tip, height, slot, wallet utxos) instead of a single tip. On restore, if the newest tip isn't found on the chain, seeding falls back to the next older checkpoint. If none verify, it falls through to the existing reset-to-genesis replay, so the scanner can't get stuck anymore.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/tests/src/common/wallet/scanner/config.rs
+++ b/tests/src/common/wallet/scanner/config.rs
@@ -3,7 +3,7 @@ use std::{collections::BTreeMap, sync::Arc, time::Duration};
 use lb_core::{header::HeaderId, mantle::Utxo};
 use lb_testing_framework::NodeHttpClient;
 
-use super::state::SharedWalletScannerState;
+use super::state::{ScannerStateCheckpoint, SharedWalletScannerState};
 use crate::{
     common::wallet::{TrackedWalletKeys, WalletUtxos},
     cucumber::{
@@ -57,6 +57,9 @@ pub enum ScannerSeed {
         source_node_names: Vec<String>,
         /// Number of trailing slots to verify before tailing from the seed tip.
         rescan_blocks: u64,
+        /// Older checkpoints to fall back to when the seed tip is not found on
+        /// the restored chain, newest first.
+        fallback_checkpoints: Vec<ScannerStateCheckpoint>,
     },
 }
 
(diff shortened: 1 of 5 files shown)


hansieodendaal (approved):
LGTM

**Non-blocking**

Maybe these behavioral regression tests in `runner.rs` around snapshot recovery?

1. A reorged primary seed tip with one or more newer fallback checkpoints also absent from the restored chain, followed by an older valid checkpoint; assert the scanner retries checkpoints in newest-first order, skips the invalid ones, reseeds from the valid checkpoint, and reaches Tailing on the restored canonical tip.

2. A reorged primary seed tip where every stored fallback is also absent, asserting the scanner exhausts fallbacks, resets to genesis replay, and still reaches `Tailing` rather than repeatedly reporting the same missing-tip error.

andrussal:
> LGTM
> 
> **Non-blocking**
> 
> Maybe these behavioral regression tests in `runner.rs` around snapshot recovery?
> 
> 1. A reorged primary seed tip with one or more newer fallback checkpoints also absent from the restored chain, followed by an older valid checkpoint; assert the scanner retries checkpoints in newest-first order, skips the invalid ones, reseeds from the valid checkpoint, and reaches Tailing on the restored canonical tip.
> 2. A reorged primary seed tip where every stored fallback is also absent, asserting the scanner exhausts fallbacks, resets to genesis replay, and still reaches `Tailing` rather than repeatedly reporting the same missing-tip error.

I looked into adding these tests, but an E2E setup would require constructing some fairly specific fork and snapshot states and would likely be fragile. The alternative is adding a block-source abstraction just for testing, which feels like too much extra machinery for this PR. Since this is non-blocking, I’d leave it out for now.

PR #3183, opened 2026-07-24
chore: add bounded and validated merkle paths

hansieodendaal:

This PR bounds and validates MMR `MerklePath` values. It:

* replaces the unbounded sibling `Vec<Fr>` with a collection capped at 32 elements;
* keeps `MerklePath` generic for smaller MMR heights;
* validates that tracked paths contain exactly `MAX_HEIGHT - 1` siblings before mutation;
* validates tracked leaf indices before updating paths;
* preserves the existing `Fr` JSON and binary serialization format;
* rejects oversized paths during deserialization;
* converts voucher paths into the exact 32-element form required by the PoC circuit;
* propagates invalid-path errors through wallet append and signing flows instead of panicking;
* rejects unsupported MMR heights, including height zero.

**Why is this required?**

- Previously, `MerklePath` exposed an unbounded `Vec<Fr>` with no relationship to the configured MMR height.
- Malformed or incompatible recovered paths could therefore deserialize successfully and later panic when the MMR updated them using direct indexing.
- A similar panic was possible during Proof of Claim construction, where an arbitrary-length path was indexed while constructing the circuit’s fixed 32-element witness.
- A simple upper bound is not sufficient because valid generic MMRs can use smaller heights. The implementation therefore enforces:
  * globally: at most 32 siblings;
  * for a specific MMR: exactly `MAX_HEIGHT - 1` siblings;
  * for PoC: exactly 32 siblings.

**Compatibility**

- Valid path serialization remains unchanged.
- Existing smaller-height MMRs remain supported.
- Malformed recovered paths now return typed errors when first used for MMR updates or PoC construction, rather than causing out-of-bounds indexing panics.
- This PR does not attempt to validate all corrupted internal MMR state. Its scope is limited to `MerklePath` shape, height, leaf-index, and PoC compatibility.

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


youngjoon-lee:
Sorry, I misclicked the `Update branch` button :)

core/src/proofs/merkle.rs
```diff
@@ -18,15 +19,47 @@ pub fn merkle_path_to_witness<T: Copy>(path: &MerklePath<T>) -> (Vec<T>, Vec<boo
...
 /// Converts an [`lb_mmr::MerklePath`] to the witness format expected by
 /// the circuit: siblings in bottom-top order, selectors reversed.
-pub fn mmr_path_to_witness(path: &lb_mmr::MerklePath) -> (Vec<Fr>, Vec<bool>) {
-    let items = path.siblings.clone();
+pub fn mmr_path_to_witness(
+    path: &lb_mmr::MerklePath,
+) -> Result<
+    (
+        [Fr; lb_poc::VOUCHER_MERKLE_TREE_HEIGHT],
+        [bool; lb_poc::VOUCHER_MERKLE_TREE_HEIGHT],
```
youngjoon-lee:
```suggestion
        [Fr; lb_poc::VOUCHER_MERKLE_PATH_LEN],
        [bool; lb_poc::VOUCHER_MERKLE_PATH_LEN],
```
I know this is not part of this PR, but I think we should rename "tree height" (defined as 32, not 33) to "path len" (32). I can do it as a next PR.

mmr/src/lib.rs
youngjoon-lee: We can reuse `assert_supported_max_height` for this.
Maybe we can rename the `assert_...` function to `assert_acceptable_height`.

mmr/src/lib.rs
```diff
@@ -119,17 +119,29 @@ where
         &self,
         elem: T,
         paths: &mut [MerklePath],
-    ) -> Result<(Self, MerklePath), MmrFull> {
+    ) -> Result<(Self, MerklePath), PushWithPathsError> {
+        // This assertion establishes the invariant required by the internal
+        // path construction below, independently of caller-supplied paths.
+        assert_supported_max_height(MAX_HEIGHT);
```
youngjoon-lee: We are already doing this assertion in the `MerkleMountainRange::new`. Do we also need to have the same assertion in this function?

mmr/src/lib.rs
```diff
@@ -119,17 +119,29 @@ where
...
         if self.roots.peek().is_some_and(|r| r.height == MAX_HEIGHT) {
-            return Err(MmrFull);
+            return Err(PushWithPathsError::MmrFull(MmrFull));
+        }
+
+        for path in paths.iter() {
+            path.validate_for_height(MAX_HEIGHT)?;
+            path.validate_leaf_index(self.len())?;
```
youngjoon-lee, line 172: Instead of these manual checks, it'd have been great if we can make `MerkleMountainRange<33>` to accept only `MerklePath<33-1>`. But, as far as I know, it's possible only in Rust nightly ([`generic_const_exprs`](https://doc.rust-lang.org/beta/unstable-book/language-features/generic-const-exprs.html)) :)

youngjoon-lee (approved):
Great. Thank you!

mmr/src/lib.rs
danielSanchezQ: > Instead of these manual checks, it'd have been great if we can make `MerkleMountainRange<33>` to accept only `MerklePath<33-1>`. But, as far as I know, it's possible only in Rust nightly ([`generic_const_exprs`](https://doc.rust-lang.org/beta/unstable-book/language-features/generic-const-exprs.html)) :)

I've been praying for this shit for ages, the day the will include it we will be retired 😭

mmr/src/lib.rs
hansieodendaal: Nice pick-up!

mmr/src/lib.rs
hansieodendaal: You’re right that the assertion is redundant for normally constructed MMRs. Since derived `Deserialize` is the path that bypasses `new()`, I’ll validate `MAX_HEIGHT` in a custom `Deserialize` implementation and return a serde error there. Then the repeated assertion in `push_with_paths()` can be removed while retaining the per-path validation.

hansieodendaal: @youngjoon-lee please see changes in https://github.com/logos-blockchain/logos-blockchain/pull/3183/commits/1c15c6f323a4320e8482a2bfea21ba0dc4a98bb2

youngjoon-lee (approved):

PR #3186, opened 2026-07-26
chore(core): remove `impl GasCalculator` for `MantleTx` and define `minimum_*` functions

youngjoon-lee:

We are estimating the gas cost of a `MantleTx` when funding the tx.
But, the gas cost of `MantleTx` cannot be computed accurately because we don't know how many proofs will be attached for each operation after the tx is funded. For most operations, proof sizes and the number of proofs are fixed. But for a `ChannelConfig` op creating a new channel, we don't know how many proofs will be attached when `SignedMantleTx` will be built, because the ledger skips the proof verification for that operation if the channel doesn't exist. So, the user can specify any number of proofs for that operation, in theory.

Nevertheless, we are already motivating users to set 0 proofs to pay less fess. Also, I think it's not a good idea to update the Mantle spec just because of the funding logic in the wallet.

So, this PR doesn't change any logic. Instead, it clarifies that a gas cost computed from `MantleTx` (not `SignedMantleTx`) is just a prediction, not an accurate value, by removing `impl GasCalculator for MantleTx` and defining functions with the `minimum_*` prefix explicitly.

These functions are currently being used by only wallet, when funding txs. This PR doesn't refactor `MantleTx` to limit the use of those function to the wallet. Instead, we use the `minimum_*` prefix to inform callers that we are not calculating an exact value. I think this is good enough, instead of taking a big surgery.

Yes

Spec: [Mantle: ChannelConfig validation](https://lip.logos.co/blockchain/raw/bedrock-v1.1-mantle-specification.html#validation-2)

@youngjoon-lee 
spec: @thomaslavaur 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/core/src/mantle/transactions/builder.rs
+++ b/core/src/mantle/transactions/builder.rs
@@ -7,7 +7,7 @@ use thiserror::Error;
 
 use crate::{
     mantle::{
-        GasCalculator as _, GasConstants, Note, NoteId, Op, Utxo, Value,
+        GasConstants, Note, NoteId, Op, Utxo, Value,
         gas::{GasCost, GasOverflow},
         ledger::{BoundedUtxos, Inputs, Outputs},
         ops::{
@@ -56,7 +56,7 @@ impl From<(BoundedError, BoundedTag)> for TxBuilderError {
 ///
 /// The builder is intentionally free of any [`MantleTxContext`]: gas prices are
 /// tip-dependent, so the context is supplied as a parameter to the fee-aware
-/// methods ([`Self::gas_cost`], [`Self::funding_delta`],
+/// methods ([`Self::minimum_gas_cost`], [`Self::funding_delta`],
 /// [`Self::return_change`]) at the moment they run. This keeps the builder
 /// serializable, so a partially built tx can be handed to a wallet (e.g. over
 /// HTTP) to be funded against a freshly fetched context.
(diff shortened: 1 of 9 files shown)


thomaslavaur (approved):
Can you double check this is only used by the wallet?

Because if it's used to decode the op then it's not good as the ledger doesn't enforce the signature to be empty... In this case we can enforce it

youngjoon-lee:
> Can you double check this is only used by the wallet?
> 
> Because if it's used to decode the op then it's not good as the ledger doesn't enforce the signature to be empty... In this case we can enforce it

@thomaslavaur 
Currently the `predict_signed_mantle_tx_size` function is used only by the wallet's funding logic, but it can be used by other modules in the future because it is defined in the `lb-core` crate.
```
predict_signed_mantle_tx_size
    └─ MantleTx::signed_serialized_size
        └─ MantleTx::storage_gas_consumption
            └─ MantleTx::{total_gas_cost, storage_gas_cost}
                └─ TxBuilder::gas_cost / funding_delta
                    └─ wallet/src/lib.rs, services/wallet/src/lib.rs
```
I considered moving the function into `lb-wallet` to limit it to be used only for the wallet. But then, we have to take a big surgery to move `MantleTx: GasCalculator` and `TxBuilder` out from `lb-core`. 

Hm, let me think how to make this better, without touching the spec.

youngjoon-lee:
> Hm, let me think how to make this better, without touching the spec.

@thomaslavaur 
I tried to refactor `MantleTx` and `MantleTxBuilder` to limit the use of `predict_...` function to only the wallet. But, I realized that the refactoring is too big, and it may be good enough to just rename functions of `MantleTx` to clarify that a gas cost computed from `MantleTx` is just a prediction. I wrote details in the updated PR description, but this is a summary.
In short, we expose a `minimum_gas_cost` function instead of `gas_cost`. This informs callers that this is not always an exact value.

thomaslavaur (approved):
I think it's good and clear

danielSanchezQ (approved):

PR #3195, opened 2026-07-28
feat(zone-sdk): make funding priority fee configurable

andrussal:

The zone SDK funds transactions with a hardcoded tip of 200 on top of the fee. This PR turns it into a config option (`FundingConfig::priority_fee`, plus a `--priority-fee` flag in tui-zone). 

Background: the balance of a funded tx is fixed when it's signed, but the ledger prices it at inclusion time. If gas prices rise by more than the tip in between, the tx becomes invalid and can never be included. The fundamental problem is that the zone cannot detect this — nothing in the network reports a failed transaction, so the SDK keeps re-submitting the same a transaction that can't currently be included forever and the publish never completes. One of the zone cucumber tests fails on the gas price fix branch because of this. A configurable tip lets tests and deployments set a buffer sized for the price movement they expect.

A larger tip only makes this failure less likely. A transaction that still becomes invalid, for this or any other reason, will be re-submitted forever. Possible follow-ups — tip as a percentage of the fee, re-funding stuck transactions when prices change.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/deployment/tui-zone/src/cli.rs
+++ b/deployment/tui-zone/src/cli.rs
@@ -104,6 +104,15 @@ pub struct NodeKeyArgs {
     /// `--funding-pk`.
     #[arg(long, default_value_t = 1_000_000, env = "MAX_TX_FEE")]
     pub max_tx_fee: u64,
+
+    /// Execution tip paid on top of the mandatory fee when funding via
+    /// `--funding-pk`; capped by `--max-tx-fee`.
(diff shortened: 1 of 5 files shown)


hansieodendaal (approved):
LGTM

pradovic (approved):

PR #3198, opened 2026-07-28
chore: Wire* traits

ntn-x2:

See https://github.com/logos-blockchain/logos-blockchain/pull/3119 for a description of the trait. This PR is simply a part 1/N where the original PR is split up into smaller ones. In this PR, the new trait and macros are introduced, and the trait is implemented on Blend types and KMS keys and signatures (used in Blend).

**The follow-up PR has been merged into this branch before merging this branch on `master` to speed things up.**

Yes.

@ntn-x2 

N/A

No.


danielSanchezQ (approved):
`Wire` is a bit confusing, as the only thing that goes through here are (iirc) mantletx and blend related stuff, any other upper level message is wrapped by bincode (?)
But otherwise fine 🟢

ntn-x2:
@danielSanchezQ actually no. Or rather: yes that is the case today, but ideally we might want to switch to this format also for other things like stuff we store in DB. I don't yet know what will be the components that could benefit from this, but especially for the fixture part, we might want to implement the trait for anything we would want to catch regressions for. Do you have a better name in mind than`Wire`? Something with the `bin` prefix would probably be too confusing.

danielSanchezQ:
> @danielSanchezQ actually no. Or rather: yes that is the case today, but ideally we might want to switch to this format also for other things like stuff we store in DB. I don't yet know what will be the components that could benefit from this, but especially for the fixture part, we might want to implement the trait for anything we would want to catch regressions for. Do you have a better name in mind than`Wire`? Something with the `bin` prefix would probably be too confusing.

If we move all formats to this (manually packing/unpacking) we would have to specify the format in the specs. It this bundled with the idea of versioning our messages?
I want to understand this well, I don't want to go from 2 methods encode/decode with bincode to a whole re-inventing the wheel system. Lets sync on this discussion.

wire/src/bounded_vec.rs
```diff
@@ -0,0 +1,312 @@
...
+impl<T, const MIN: usize, const MAX: usize> WireExamples for BoundedVec<T, MIN, MAX>
+where
+    T: WireExamples,
+{
+    fn fixtures() -> WireFixtures<Self> {
+        let count = MIN.max(1);
+
+        let mut values = Vec::with_capacity(count);
+        let mut bytes = Vec::new();
+        encode_length_prefix_into::<MAX>(count, &mut bytes);
```
hansieodendaal: Since `[0, 0]` is a valid bound, could this use zero elements when `MAX == 0` and could we add a regression test for `BoundedVec<u8, 0, 0>`?

wire/src/bounded_vec.rs
```diff
@@ -0,0 +1,312 @@
...
+impl<T, const MIN: usize, const MAX: usize> WireDecode for BoundedVec<T, MIN, MAX>
+where
+    T: WireDecode,
+{
+    type Context = T::Context;
+
+    fn decode<'input>(
+        input: &'input [u8],
+        context: &Self::Context,
+    ) -> Result<(&'input [u8], Self), DecodeError> {
```
hansieodendaal: Could we add a regression test that defends against memory allocation attacks? Something like 
```ruat
#[test]
fn huge_declared_length_does_not_preallocate_from_the_wire() {
    type Wide = BoundedVec<u8, 1, { u64::MAX as usize }>;

    let err = Wide::decode(&u64::MAX.to_le_bytes()).unwrap_err();

    assert!(matches!(err, DecodeError::UnexpectedEnd { .. }));
}
```
I do not think `let mut items = Vec::with_capacity(len);` is the correct option here. 
`let mut items = Vec::new();` or something like `let mut items = Vec::with_capacity(len.min(1024));` would be much safer.

codec/src/bounded_vec.rs
```diff
@@ -0,0 +1,312 @@
...
+    }
+}
+
+impl<T, const MIN: usize, const MAX: usize> WireDecode for BoundedVec<T, MIN, MAX>
+where
+    T: WireDecode,
+{
+    type Context = T::Context;
+
+    fn decode<'input>(
```
hansieodendaal, line 101: Should `BoundedVec` support zero-length wire types such as `[u8; 0]`? If zero-length elements are not required, perhaps reject successful element decodes that make no input progress. If they are supported, we need a separate safe element-count limit independent of `MAX`, as a large length prefix could cause an effectively unbounded CPU loop.

core/src/mantle/ops/mod.rs
```diff
@@ -118,91 +112,88 @@ impl<'de> Deserialize<'de> for Op {
...
-    fn decode(bytes: &[u8]) -> IResult<&[u8], Self> {
-        let (bytes, opcode) = u8::decode(bytes)?;
+impl WireDecode for Op {
+    type Context = ();
+
+    fn decode<'input>(
+        input: &'input [u8],
+        (): &Self::Context,
+    ) -> Result<(&'input [u8], Self), DecodeError> {
+        let (input, opcode) = u8::decode(input, &())?;
```
hansieodendaal, line 164: Why has the `CHANNEL_TRANSFER ` leg been dropped from this decode function? Gremlin or intent? Encoding still emits opcode `0x14`, so a `ChannelTransfer` currently encodes successfully but fails to decode with `UnknownDiscriminant`.
```
CHANNEL_TRANSFER => ChannelTransferOp::decode(bytes)
                .map(|(bytes, op)| (bytes, Self::ChannelTransfer(op))),
```

hansieodendaal (commented):
Looking good! Some observations below.

core/src/mantle/ops/mod.rs
ntn-x2: Good catch. Some rebase messed this up. Fixed in https://github.com/logos-blockchain/logos-blockchain/pull/3198/commits/9ee3a385575681359c17cf2f43a91ca7d77d39e6.

Since this PR changes a lot of files, rebases are a mess. I will go over and address your other comments in a follow-up PR once this is merged.

PR #3200, opened 2026-07-28
refactor(transactions): unify Op preverification

### LINKED ISSUE #2638: refactor `Operation` trait to have all logic

strinnityk:
Adds a `preverify` stage to the `Operation` trait, splitting op verification into two typed steps:
- `preverify`: stateless, context-free checks
- `verify`: stateful checks against ledger state

`SignedMantleTx` now delegates verification to each `Op`.
Old `verify_stateless` (and similar) references have been removed as well in favour of a single naming style.

Yes

@strinnityk 

N/A

N/A

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/core/src/mantle/ops/channel/channel_transfer.rs
+++ b/core/src/mantle/ops/channel/channel_transfer.rs
@@ -61,13 +61,28 @@ pub struct ChannelTransferExecutionContext {
 }
 
 impl Operation<ChannelTransferValidationContext<'_>> for ChannelTransferOp {
+    type PreverificationContext<'a>
+        = ()
+    where
+        Self: 'a;
     type ExecutionContext<'a>
         = ChannelTransferExecutionContext
     where
         Self: 'a;
-    type Error = Error;
+    type VerificationError = Error;
+    type ExecutionError = Error;
 
-    fn validate(&self, ctx: &ChannelTransferValidationContext<'_>) -> Result<(), Self::Error> {
+    fn preverify(
+        &self,
+        _context: &Self::PreverificationContext<'_>,
+    ) -> Result<(), Self::VerificationError> {
+        self.verify_stateless()
+    }
+
+    fn verify(
+        &self,
+        ctx: &ChannelTransferValidationContext<'_>,
+    ) -> Result<(), Self::ExecutionError> {
         verify_channel_multi_sig(
             &self.channel_id,
             ctx.proof,
@@ -129,7 +144,7 @@ impl Operation<ChannelTransferValidationContext<'_>> for ChannelTransferOp {
     fn execute(
         &self,
         mut ctx: Self::ExecutionContext<'_>,
-    ) -> Result<(Self::ExecutionContext<'_>, Vec<TxEvent>), Self::Error> {
+    ) -> Result<(Self::ExecutionContext<'_>, Vec<TxEvent>), Self::ExecutionError> {
         // Remove the inputs from the ledger and from the channel.
         ctx.utxos = self.inputs.execute(ctx.utxos)?;
         for note_id in self.inputs.iter() {
--- a/core/src/mantle/ops/channel/channel_transfer.rs
+++ b/core/src/mantle/ops/channel/channel_transfer.rs
@@ -29,13 +29,6 @@ impl ChannelTransferOp {
     pub fn utxos(&self) -> impl Iterator<Item = Utxo> {
         self.outputs.utxos(self)
     }
-
-    pub fn verify_stateless(&self) -> Result<(), Error> {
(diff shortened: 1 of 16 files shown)


youngjoon-lee (commented):
thank you!

youngjoon-lee (approved):

core/src/mantle/ops/channel/channel_transfer.rs
```diff
@@ -61,13 +54,31 @@ pub struct ChannelTransferExecutionContext {
...
     type ExecutionContext<'a>
         = ChannelTransferExecutionContext
     where
         Self: 'a;
-    type Error = Error;
+    type VerificationError = Error;
+    type ExecutionError = Error;
 
-    fn validate(&self, ctx: &ChannelTransferValidationContext<'_>) -> Result<(), Self::Error> {
+    fn preverify(
```
ntn-x2, line 68: Do you think these operations should return some associated type that the implementor itself decides? Simply returning `Result<(), Error>` means that the compiler won't complain that the operation has not been pre-verified. I think we should return an associated type that implements a specific trait, e.g., we can define `type PreVerified: VerifyTrait`, and then define the `verify` function on the `VerifyTrait` trait, so it's impossible to forget to run the pre-verify step. 

Sorry I don't have full context and I was away the last couple of weeks, but I would like to see a bit more compilation-time checks even if we define these traits, which are definitely nice to have. What do you think? Cc @danielSanchezQ

danielSanchezQ: > Do you think these operations should return some associated type that the implementor itself decides? Simply returning `Result<(), Error>` means that the compiler won't complain that the operation has not been pre-verified. I think we should return an associated type that implements a specific trait, e.g., we can define `type PreVerified: VerifyTrait`, and then define the `verify` function on the `VerifyTrait` trait, so it's impossible to forget to run the pre-verify step.
> 
> Sorry I don't have full context and I was away the last couple of weeks, but I would like to see a bit more compilation-time checks even if we define these traits, which are definitely nice to have. What do you think? Cc @danielSanchezQ

There is probably an upcoming PR that adds typestate to the Ops which will do exactly as you say. Or at least is in the roadmap.

strinnityk: My idea was to actually make the it be `-> Result<Self, ...>` but that would mean adding `Sized` or having an extra (or two) associated types which would be pointing to `Self`, I guess.
I wasn't fully convinced about any of those, and since these methods are called from a wrapping typestate struct, that situation can be handled there.
That being said, if you guys feel it's a good idea I'm absolutely not opposed to changing the signature, quite the contrary.

ntn-x2: Yes but `Result<Self>` still has the same issue, right? At least from the "consumer" perspective, calling and handling result or not calling it at all would be the same. Hence it would be easy to "forget" to call the method and check for the result. I don't have the whole context regarding what you have in mind, so I cannot make any judgement at the moment. I would just like it to be impossible, aka a hard, red error at compile time, if someone forgets to run the required checks. So forcing the type to be consumed in a specific way by doing a fixed sequence of checks, probably from less to most expensive.

danielSanchezQ: > My idea was to actually make the it be `-> Result<Self, ...>` but that would mean adding `Sized` or having an extra (or two) associated types which would be pointing to `Self`, I guess. I wasn't fully convinced about any of those, and since these methods are called from a wrapping typestate struct, that situation can be handled there. That being said, if you guys feel it's a good idea I'm absolutely not opposed to changing the signature, quite the contrary.

But there is no states in this PR yet? What I mean, if what Antonio means is enforced later after the typestate inclusion?
If not yeah, it would be better to include the typestate here.

strinnityk: Correct, this PR doesn't include states yet. Just provides all necessary methods as part of the trait.

strinnityk: Agreed, and thinking it through further: a witness/token-return approach doesn't fix this either as long as `preverify`/`verify`/`execute` stay methods on a public trait.

The way to enforce this would be not to impl the trait at all, but to provide those functions are internal, reachable only through the typestate.

This PR is intentionally just laying down the trait/methods as the capability layer. There's a lot more layers than I expected to this refactor, so I'm trying to keep it split so PRs are lighter.

PR #3202, opened 2026-07-28
feat(pow): Implement Empowering core Op

danielSanchezQ:

Includes `ClaimPoWRewards` Op and partial verification implementation. There is a missing point which happens to need a block height which is not really easily available. This later need to be integrated as part of the missing Mantle execution and Blend new behaviour branches. For now this should be a simple first approach on needed items.

Its not really much code

@madxor @youngjoon-lee @danielSanchezQ 

Spec is in sketch atm, but from calendar preassures we already started implementation.

🤷 Not to current.

## Checklist

> [!WARNING]  

* [x] 2. Description added.


core/src/mantle/ops/codec.rs
```diff
@@ -79,6 +80,7 @@ fn encode_op_proof(proof: &OpProof, op: &Op) -> Vec<u8> {
             }
             OpProof::ZkSig(sig) => sig.encode(),
             OpProof::PoC(poc) => poc.encode(),
+            OpProof::None => Vec::with_capacity(0),
```
ntn-x2: Would it make more sense to make the proof itself an `Option` instead of having a variant for the `None` case?

danielSanchezQ: Then we would have to spread an optional through all implementations and/or check if they come with one or not. I think this is more straight forward. Also the encoding is splicit now. You can check the new version.

core/src/mantle/fixtures/ops/pow.rs
```diff
@@ -0,0 +1,33 @@
...
+
+codec_fixtures!(
+    ClaimPowRewardOp,
+    ClaimPowRewardOp {
+        epoch_nonce: Fr::ZERO,
+        block_hash: [0u8; 32],
+        public_key: ZkPublicKey::new(Fr::ZERO),
+    } => "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+    ClaimPowRewardOp {
+        epoch_nonce: Fr::ONE,
```
ntn-x2: Just a nit, perhaps epoch nonce fixture should be `0`, so we use `0`, `1`, and `2`s for the different components.

core/src/mantle/ops/mod.rs
```diff
@@ -254,13 +265,36 @@ pub struct ZkAndEd25519Proof {
...
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum OpProof {
     Ed25519Sig(Ed25519Signature),
     ZkSig(ZkSignature),
     ZkAndEd25519Sigs(ZkAndEd25519Proof),
     PoC(Groth16LeaderClaimProof),
     ChannelMultiSigProof(ChannelMultiSigProof),
+    None(NoOpProof),
```
ntn-x2, line 297: Do we need the `NoOpProof` here?

core/src/mantle/ops/mod.rs
```diff
@@ -254,13 +265,36 @@ pub struct ZkAndEd25519Proof {
     pub ed25519_sig: Ed25519Signature,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct NoOpProof;
```
ntn-x2, line 269: I think we can get rid of this since we have the `OpProof` enum. If we were to delete that enum (I am in favor since we parse ops to know what proofs to expect), then this would make more sense, in my opinion.

core/src/mantle/ops/pow.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub epoch_reward_pool: PowReward,
+    pub current_epoch_nonce: Epoch,
+    pub previous_epoch_nonce: Epoch,
+    pub blocks_height: HashMap<Hash, u64>,
+}
+
+impl ClaimPoWRewardVerificationContext<'_> {
+    /// Claiming must be enabled for this block context (pool can cover the
+    /// reward)
+    fn pow_reward_enabled(&self) -> Result<(), ClaimPowRewardError> {
```
ntn-x2:
```suggestion
    fn are_pow_rewards_enabled(&self) -> Result<(), ClaimPowRewardError> {
```

core/src/mantle/ops/pow.rs
```diff
@@ -0,0 +1,224 @@
...
+            return Err(ClaimPowRewardError::InsufficientPoolBalance {
+                pool: self.epoch_reward_pool,
+                reward: self.epoch_pow_reward,
+            });
+        }
+        Ok(())
+    }
+
+    /// On-chain `block_hash` window check
+    pub fn accept_claim<const WINDOW: u64>(
```
ntn-x2, line 120: Do we want this to be a non-const value instead, in case we make this configurable in the deployment? Would that make sense?

core/src/mantle/ops/pow.rs
```diff
@@ -0,0 +1,224 @@
...
+        &self,
+        block_id: Hash,
+    ) -> Result<(), ClaimPowRewardError> {
+        let Some(&block_height) = self.blocks_height.get(&block_id) else {
+            return Err(ClaimPowRewardError::MissingBlock { block_id });
+        };
+
+        let check_height: i128 = i128::from(self.current_block_height) - i128::from(block_height);
```
ntn-x2: Why not use a `checked_sub` instead, considering `None` as the negative case below? Much cleaner I think.

core/src/mantle/ops/pow.rs
```diff
@@ -0,0 +1,224 @@
...
+        if check_height.is_negative() {
+            return Err(ClaimPowRewardError::OutOfWindowHeight {
+                height: block_height,
+            });
+        }
+        if !(0 <= check_height && check_height <= i128::from(WINDOW)) {
```
ntn-x2: If we use a `checked_sub` above, this would be a `u64`, and it would be >= `0` by definition, so we only need to check if it's less than `WINDOW`.

core/src/mantle/ops/pow.rs
```diff
@@ -0,0 +1,224 @@
...
+        &self,
+        claim_epoch_nonce: ZkHash,
+    ) -> Result<(), ClaimPowRewardError> {
+        let previous_epoch_nonce = ZkHasher::digest(&[fr_from_mod_bytes(
+            &self.previous_epoch_nonce.into_inner().to_le_bytes(),
+        )]);
+        let current_epoch_nonce = ZkHasher::digest(&[fr_from_mod_bytes(
+            &self.current_epoch_nonce.into_inner().to_le_bytes(),
+        )]);
+        if claim_epoch_nonce != previous_epoch_nonce && claim_epoch_nonce != current_epoch_nonce {
```
ntn-x2: What about checking equality with `previous_epoch_nonce` before deciding whether we need to hash for `current_epoch_nonce`? Or viceversa.

core/src/mantle/ops/pow.rs
```diff
@@ -0,0 +1,224 @@
...
+            });
+        }
+        Ok(())
+    }
+
+    fn validate_difficulty_reward(
+        &self,
+        puzzle_ticket: PuzzleTicket,
+    ) -> Result<(), ClaimPowRewardError> {
+        let ticket_as_fr = puzzle_ticket.0;
```
ntn-x2: Since `PoWNullifier` is a new type introduced here, we could also let it implement `PartialOrd` with an `Fr` directly, so we run comparison directly without accessing `.0`. Just a nit.

core/src/mantle/ops/pow.rs
```diff
@@ -0,0 +1,224 @@
...
+    ) -> Result<(), ClaimPowRewardError> {
+        if self.pow_nullifiers.contains(&puzzle_ticket) {
+            return Err(ClaimPowRewardError::DoubleClaimed);
+        }
+        Ok(())
+    }
+}
+
+pub struct ClaimPoWRewardExecutionContext {
+    _phantom: std::marker::PhantomData<()>, // fake content to be removed
```
ntn-x2, line 189: Why?

core/src/mantle/ops/pow.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+impl ProvableOperation for ClaimPowRewardOp {
+    type Proof = NoOpProof;
```
ntn-x2, line 193: Ok I guess that's the reason why we have a `NoOpProof` definition.

core/src/mantle/ops/mod.rs
danielSanchezQ: Its because of the encoding, and because with the new verification traits we need to actually pass a proof. Its kinda explicit, I like it. But im open to options.

core/src/mantle/ops/mod.rs
danielSanchezQ: If you check the verification we need to pass a proof. I can probably set the proof to `()`, but then the encoding of it would be in the general OpProof wrapper instead. But for any other proof is in the proof itself.

core/src/mantle/ops/pow.rs
danielSanchezQ: Supposedly this is static. But it may make sense to be deployment configurable. It can be done in a further PR when this values and stuff is actually in the spec. wdyt?

core/src/mantle/ops/pow.rs
danielSanchezQ: Much better indeed, thanks.

core/src/mantle/ops/pow.rs
danielSanchezQ: Ill revert the conditions to short circuit the check, thanks.

core/src/mantle/ops/pow.rs
danielSanchezQ: Its just used here and its contained, il add an `as_fr` method instead.

core/src/mantle/ops/pow.rs
danielSanchezQ: This is gone in upper PR, I had some lifetime and then I moved it. Its more a leftover than anything. Disregard it.

core/src/mantle/ops/pow.rs
danielSanchezQ: Yes it is 😅

core/src/mantle/ops/pow.rs
ntn-x2: Sure, I think it would be best to have them dynamically configurable instead. Fine if it's done in a separate PR.

danielSanchezQ:
@ntn-x2 most of stuff disscused changed here: https://github.com/logos-blockchain/logos-blockchain/pull/3202/commits/e5ce102b846792ef38b1e78e7ba1d7b2a6323e35

youngjoon-lee (approved):

ntn-x2 (approved):
All good! Minus the change from consts to dynamic configs to be done in a separate PR 🚀🚀🚀

strinnityk (approved):

PR #3206, opened 2026-07-29
test: improve ci merge queue integration tests

hansieodendaal:

Run Linux and macOS Cucumber/E2E lanes concurrently, but report success (stop merge-queue tests) as soon as either OS passes the complete suite.

The existing result jobs use `needs: tests`, which waits for both matrix lanes. This allowed a slow macOS lane to delay `Cucumber result` until after the merge-queue timeout, despite Linux already passing.

#### Changes

- Added shared `check-peer-success` action:
  - Detects when the other OS has fully passed.
  - Stops the remaining lane at its next test-phase boundary.
  - Does not cancel the workflow or interrupt a running test command.
- Added shared `check-any-lane-result` action:
  - Polls the current workflow every 30 seconds.
  - Passes when either OS completes successfully.
  - Fails only after both OS lanes finish without success.
- Updated Cucumber and E2E workflows to use the shared actions.
- Kept matrix `fail-fast: false` so one failed OS does not cancel the other.
- Kept E2E `--no-fail-fast` so each running nextest phase can report multiple failures.
- Kept heavy tests restricted to `merge_group`.

#### Result

- Either Linux or macOS may satisfy the suite.
- The required result no longer waits unnecessarily for the other OS.
- The remaining lane skips later phases, conserving self-hosted runner resources.
- If both OS lanes fail, the merge is blocked.

Required check names remain unchanged:

- `Cucumber result`
- `E2E result`

No branch-protection or ruleset changes are required.

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.

### DIFF
--- a/.github/workflows/cucumber-integration-tests.yml
+++ b/.github/workflows/cucumber-integration-tests.yml
@@ -246,7 +246,7 @@ jobs:
     name: Cucumber result
     if: ${{ !cancelled() }}
     runs-on: ubuntu-latest
-    timeout-minutes: 175
+    timeout-minutes: 230
     permissions:
       actions: read
       contents: read
--- a/.github/workflows/cucumber-integration-tests.yml
+++ b/.github/workflows/cucumber-integration-tests.yml
@@ -242,36 +242,36 @@ jobs:
   # It fails once both lanes have completed without reporting success.
   #
   # On pull requests it succeeds immediately without starting self-hosted jobs.
-cucumber-result:
-  name: Cucumber result
-  if: ${{ !cancelled() }}
-  runs-on: ubuntu-latest
-  timeout-minutes: 175
-  permissions:
-    actions: read
-    contents: read
-  steps:
-    # On pull requests, report the required check without running or checking
-    # either self-hosted lane.
-    - name: Report pull-request result
-      if: ${{ github.event_name != 'merge_group' }}
-      run: echo "Cucumber integration tests run only in the merge queue"
+  cucumber-result:
+    name: Cucumber result
+    if: ${{ !cancelled() }}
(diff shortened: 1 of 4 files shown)


.github/workflows/cucumber-integration-tests.yml
```diff
@@ -204,33 +228,50 @@ jobs:
...
+  # OS has completed the entire lane and its `Report success` step passed.
+  # It fails once both lanes have completed without reporting success.
+  #
+  # On pull requests it succeeds immediately without starting self-hosted jobs.
   cucumber-result:
     name: Cucumber result
-    needs: tests
     if: ${{ !cancelled() }}
     runs-on: ubuntu-latest
+    timeout-minutes: 175
```
andrussal: Maybe make this longer so we don't fail on a very slow macos runner — Monday one lane took 3h17m (run 30353203117), just over this cap. ofc ideally we don't have slow runners at all.

andrussal (approved):
Nice, I think this makes sense.

.github/workflows/cucumber-integration-tests.yml
hansieodendaal: Yeah, that is something I considered. The current attempt coincides with a request to reduce `Build concurrency` in the merge queue from 5 to 3, which should alleviate runner pressure already. We also have `Status-check timeout` in the merge queue as 180 at the moment; this directly influences that setting of ` timeout-minutes: 175`.

My thoughts were that we should closely monitor overall merge queue performance after this PR and the requested changes to `Build concurrency`; we should have ample headroom.

I will increase the time-out and also request changes to `Status-check timeout`.

PR #3207, opened 2026-07-29
chore: update toolchain to rust 1.97(.1)

ntn-x2:

Update toolchain to version 1.97.1 and fmt nightly version accordingly.

The new Clippy introduces two pedantic warnings that I decided to ignore for the time being, but that should be tackled at some point for consistent styling: [`inline_modules`](https://rust-lang.github.io/rust-clippy/master/index.html?search=inline_#inline_modules) and [`inline_trait_bounds`](https://rust-lang.github.io/rust-clippy/master/index.html?search=inline_#inline_trait_bounds).

Also, a new feature allows us to specify a cargo env variable to explicitly deny warnings, removing the need to specify the `RUSTFLAGS` env variable which would result in crate recompilation. Those flags can be re-introduced for something else than this now, if needed.

Yes.

@ntn-x2 

N/A

No.

### DIFF
--- a/.github/actions/cargo-hack-check-cached/action.yml
+++ b/.github/actions/cargo-hack-check-cached/action.yml
@@ -41,7 +41,7 @@ runs:
     - name: Run Cache-Aware Cargo Hack Check
       shell: bash
       env:
-        RUSTFLAGS: "-D warnings"
+        CARGO_BUILD_WARNINGS: deny
       run: python3 ./scripts/cargo-hack-check/script.py --continue-on-failure
 
     - name: Save Cargo Hack Keys (main)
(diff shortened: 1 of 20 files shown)


hansieodendaal (approved):
LGTM

**Note:** I think we should hold off merging this PR until we can point to the similarly updated Overwatch crate (https://github.com/logos-co/Overwatch/pull/147).

ntn-x2:
@hansieodendaal why? They are not related. Lints are run on workspace crates only.

hansieodendaal:
> @hansieodendaal why? They are not related. Lints are run on workspace crates only.

Never mind, it was just so we use the same toolchain for each when compiling a node, but it should not matter.

danielSanchezQ (approved):

PR #3209, opened 2026-07-29
fix(cryptarchia-engine): avoid potential infinite loops in walk-back functions

### LINKED ISSUE #2702: avoid infinite loops in `Branches` logically
`Branches` has several functions which potentially cause infinite loops:
- `lca`
- `walk_back_before`
- `walk_back_to_block`

Check how they are used by callers, and find a way to eliminate infinite loops completely.

youngjoon-lee:

Closes #2702

Several functions in the `cryptarchia-engine` had parent-block accesses with `self.branches[&X.parent]` in loops. They cause infinite loops potentially if they hit the genesis block or the LIB just loaded from the recovery state, because its parent is itself. 

This PR changes the following functions to avoid the potential infinite loops:
- `Branches::lca`
  - Now, it returns `Option` if LCA can't be found in the tree, instead of panicking or infinite loop.
  - This change doesn't affect callers because they always use right values to call the function. So, I added `expect(..)` to the callers.
- `Branches::walk_back_before`
  - Now, it stops at the oldest block in the tree, if it is hit before the target slot.
  - I confirmed that this change doesn't change callers' behavior.
- `Branches::walk_back_to_block`
  - Now, it stops at the oldest block in the tree, if it is hit before the target block.
  - I confirmed that this change doesn't change callers' behavior.

Yes

@youngjoon-lee 

N/A

N/A

## Checklist

> [!WARNING]  

* [x] 2. Description added.

### DIFF
--- a/consensus/cryptarchia-engine/src/lib.rs
+++ b/consensus/cryptarchia-engine/src/lib.rs
@@ -83,7 +83,9 @@ where
 
     let forks = branches.branches();
     for chain in forks {
-        let lowest_common_ancestor = branches.lca(&cmax, &chain);
+        let lowest_common_ancestor = branches
+            .lca(&cmax, &chain)
+            .expect("local chain and fork must have a common ancestor");
         let m = cmax.length - lowest_common_ancestor.length;
         if m <= k {
             // Classic longest chain rule with parameter k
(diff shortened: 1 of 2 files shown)


danielSanchezQ (approved):
Nice catch @youngjoon-lee , thanks!

PR #3210, opened 2026-07-29
feat(mempool): evict expired transactions after configurable ttl

andrussal:

This adds a TTL for pending mempool transactions (default 1h): anything that stays pending longer than the TTL is evicted from the pool.

Transactions are removed from the mempool in two ways: when they appear in an applied block, or when a node acting as leader finds them invalid while building a block. That second path only runs on the proposer. A node that doesn't propose never validates its pending transactions, so invalid ones are never cleared and the mempool grows indefinitely.

Configurable via mempool.tx_ttl in deployment settings; null disables eviction.

Yes

@andrussal 

N/A

N/A

## Checklist

* [x] 2. Description added.

### DIFF
--- a/services/tx-service/src/backend/pool.rs
+++ b/services/tx-service/src/backend/pool.rs
@@ -8,7 +8,7 @@ use std::{
 
 use async_trait::async_trait;
 use futures::Stream;
-use indexmap::IndexSet;
+use indexmap::IndexMap;
 use lb_core::mantle::transactions::hash::PrefixedKey;
 use lb_log_targets::mempool;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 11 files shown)


danielSanchezQ (commented):
We wanted to have this from some time ago. Now, this PR is pretty straight forward. I would like to see a configurable policy system which is not completely add-hoc to a single behaviour. Even if for now the ttl is the first policy.
Something in the line of having an analytical struct/entity that flags Txs depending the enalbled policies. We can discuss about this in the proper channels if its not that clear.
wdyt?

andrussal:
> We wanted to have this from some time ago. Now, this PR is pretty straight forward. I would like to see a configurable policy system which is not completely add-hoc to a single behaviour. Even if for now the ttl is the first policy. Something in the line of having an analytical struct/entity that flags Txs depending the enalbled policies. We can discuss about this in the proper channels if its not that clear. wdyt?

agree, was thinking about this a bit now more. one way to frame it — maybe two separate things:

```
tx ──► gate ──► pool ◄── evictor
        │                  │
   disqualify            evict
   (invalid)         (old · no room)
```

|            | gate | evictor |
|------------|------|---------|
| role       | rejects invalid txs — as strictly as the node can verify (structure, sigs, later ledger) | frees the pool of valid txs it won't keep carrying — too old, or no room |
| why split | invalid is invalid on every node, so hardcoded | ttl/size are just local limits, so config |

these two depend on each other a bit: right now gate does nothing so ttl has to clean up everything, invalid or just stale, we can't tell. once gate can actually verify stuff, invalid txs leave through it and eviction is left with capacity + txs that can't be verified. ttl probably never disappears completely

on the eviction side something like — each policy a self-contained type: sees txs at add/remove, keeps its own state, answers one question, pool just executes:

```rust
struct TtlPolicy {
    tx_ttl: Duration,
    inserted: IndexMap<TxHash, u64>,
}

impl TtlPolicy {
    fn on_add(&mut self, key: TxHash, now: u64);
    fn on_remove(&mut self, key: &TxHash);
    fn expired(&self, now: u64) -> Vec<TxHash>;
}

struct Evictor {
    ttl: Option<TtlPolicy>,
    budget: Option<BudgetPolicy>,   // later: when pool is over size, evict lowest-fee first
}
```

ttl from this PR would be the first evictor policy, size/budget next. not sure the gate side should be configurable at all — just as strict as we can verify. wdyt

andrussal:
> > We wanted to have this from some time ago. Now, this PR is pretty straight forward. I would like to see a configurable policy system which is not completely add-hoc to a single behaviour. Even if for now the ttl is the first policy. Something in the line of having an analytical struct/entity that flags Txs depending the enalbled policies. We can discuss about this in the proper channels if its not that clear. wdyt?
> 
> agree, was thinking about this a bit now more. one way to frame it — maybe two separate things:

@danielSanchezQ pushed a version of this — ttl extracted into evictor + policy module. same behavior. roughly what you had in mind?

danielSanchezQ:
> @danielSanchezQ pushed a version of this — ttl extracted into evictor + policy module. same behavior. roughly what you had in mind?

taking a look, thanks!

services/tx-service/src/backend/pool.rs
```diff
@@ -8,26 +8,55 @@ use std::{
...
+    backend::{MemPool, MempoolError, RecoverableMempool, evictor::Evictor},
     metrics,
     storage::MempoolStorageAdapter,
 };
 
 const REMOVED_ITEM_GRACE_PERIOD: Duration = Duration::from_mins(10);
 const LOG_TARGET: &str = mempool::POOL;
 
+/// Default time a pending transaction may stay in the pool before eviction.
+pub const DEFAULT_TX_TTL: Duration = Duration::from_hours(1);
```
danielSanchezQ: I would make it a bit more safe, 1 hour seems sort (?)

danielSanchezQ (approved):
Abstraction looks much better now, and adding more policies in the future should be easier too. Thanks!

services/tx-service/src/backend/pool.rs
andrussal: > I would make it a bit more safe, 1 hour seems sort (?)

Good point, I increased it to one day now.

PR #3217, opened 2026-07-30
fix(blend): encapsulation padding logic

ntn-x2:

This PR fixes two minor issues in the code logic that are not really visible in network runs:

1. If a node tries to encapsulate less than the expected number of layers, we were not doing any additional padding. Since we have a TODO to allow for less than the expected number of layers in case the underlying proof system fails, this fix is needed so that a message that is padded less than the expected number of times still has the same length as a fully-encapsulated message
2. The Blend payload spec defines padding on the payload body as being random, but we were just filling the remaining part with zeroes. This is not really visible from casual external observers because everything is encapsulated, but it might leave the door open to chosen plaintext attacks, so we fix that as well.

The fixtures have been updated because we are not using a padding of zeroes anymore.

Yes.

@madxor for the spec. @youngjoon-lee for the initial implementation of the encapsulation/decapsulation logic. @ntn-x2 for the follow-up PR.

Yes. This PR brings the implementation up to date with the spec.

No.


youngjoon-lee (approved):
Thank you. I've missed this mismatch between impl and spec, since I wrote this code before the spec was written, iirc.

PR #3220, opened 2026-07-30
feat(zone-sdk): remove indexer

pradovic:

Per discussion with LEZ, this PR removes the indexer as a separate component of zone-sdk. Users can use existing zone-sdk, together with ChannelUpdates to get both finalized and non-finalized state. Providing a key that is non-existing in rotation makes it read-only. It then also supports a use-case where use might supply key that becomes valid in the future.

All tests that use indexer kept the same structure but are initializing a separate read-only zone-sequencer for this.

Yes

@pradovic  @davidrusu 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


tests/src/cucumber/steps/manual_zone/assertions.rs
```diff
@@ -3,36 +3,33 @@ use std::{
...
-            if expected.contains(&payload) {
-                seen.insert(payload);
-            }
-        })
-        .await?;
+        let seen: HashSet<Inscription> =
+            replayed_inscription_payloads(&replay_finalized_history(reader).await?)
+                .into_iter()
+                .filter(|payload| expected.contains(payload))
+                .collect();
```
hansieodendaal, line 32: Could we avoid cold-starting and replaying from genesis on every polling iteration?

The previous indexer flow retained a slot cursor and fetched only newly finalized data. This now creates a fresh sequencer without a checkpoint for every poll, making the work proportional to the complete channel history multiplied by the number of retries.

tests/src/cucumber/steps/manual_zone/support.rs
```diff
@@ -964,77 +964,91 @@ pub async fn wait_for_turn_to_write(
...
+        let mut finalized = Vec::new();
+        loop {
+            match sequencer.next_event().await {
+                Event::BlocksProcessed {
+                    finalized: batch, ..
+                } => finalized.extend(batch),
+                Event::Ready => return finalized,
+                Event::MempoolPending(_) | Event::TurnNotification { .. } => {}
+            }
+        }
```
hansieodendaal, line 999: Can `Ready` safely be used as the end-of-replay marker here? Could the replay expose an explicit successful backfill-complete boundary, perhaps tied to a captured LIB/checkpoint, rather than terminating on Ready?

Would this be plausible currently for the helper:
- wait until its three-minute timeout after completing backfill if no new block arrives, or
- return on `Ready` before collecting finalizations carried by the first live block's buffered `BlocksProcessed` event.

hansieodendaal (commented):
Looking good!

I have two questions/comments to consider please.

tests/src/cucumber/steps/manual_zone/assertions.rs
pradovic: We could, but I think it does not influence the tests much. We are not testing long networks with indexer restart. What do you think?

The way to avoid it now is not cursor but checkpoint, same as the regular sequencer flow.

hansieodendaal: We should actually add long-running Zone SDK scenarios similar to long-running transaction scenarios, run locally, not run in CI. For those this would be a massive problem.

pradovic: Makes sense, but I would live this for a special PR.

pradovic: [https://github.com/logos-blockchain/logos-blockchain/issues/3242](https://github.com/logos-blockchain/logos-blockchain/issues/3242). I have created an issue. We can do it whenever any of us has an empty plate.

tests/src/cucumber/steps/manual_zone/support.rs
pradovic: Oups, sorry I missed this comment.

The helper is correct for its documented contract imo and the tests are not flaky. 

Ready is a complete boundary for what the helper promises (a snapshot up to the LIB at connect time): backfill emits one BlocksProcessed per batch and the buffer is drained every drive turn, so all backfill-range finalizations are collected before Ready can fire.

On your two scenarios: (1) Ready does additionally need one live block after backfill, so a halted chain would run into the timeout — but the cluster mints blocks continuously, so it lands within a slot, and a halted chain fails the scenario regardless; (2) the buffered BlocksProcessed behind Ready only carries items beyond the connect-time snapshot, and the wait steps poll this helper repeatedly, so the next call picks those up.
 
An explicit captured-LIB boundary would work too, but it adds a consensus-info round-trip and boundary bookkeeping to change no observable behavior in these tests.

I went for the simple solution as it might be enough for the tests. What do you think?

hansieodendaal: Ok, let's go with it. We can always revisit this when required.

hansieodendaal (approved):
LGTM

PR #3221, opened 2026-07-30
chore(deployment): Docker tools image tags

bacv:

When building tools image, add additional label if git commit is tagged. This would produce `devnet` + `0.2.1-rc.3` labeled images.

Previously we labeled tools images only by `devnet` and `testnet`, now we add version tag too.

N/A

N/A

N/A

## Checklist

* [ ] 2. Description added.


pradovic (approved):

.github/workflows/build-logos-tools.yml
```diff
@@ -35,6 +40,11 @@ jobs:
       - name: Print run info
         run: |
           echo "Building Logos tools image for ${{ matrix.platform }} with tag: ${{ inputs.tag }}"
+          if [ -n "${{ inputs.custom_version }}" ]; then
```
ntn-x2, line 43: Should we let the workflow fail if a custom tag is being specified when triggering the workflow from a tagged commit? I think it would be a safer approach here.

ntn-x2 (approved):
I'm not sure if we need it for backwards compatibility, but to me it seems that `inputs.tag` might not be required anymore, if we use the commit tag or a manually-specified one now. We could remove the `devnet`/`testnet` additional input.

PR #3224, opened 2026-07-31
fix(encoding): implement mitigations for malicious inputs on bounded vec decoding

ntn-x2:

Addresses leftover comments from https://github.com/logos-blockchain/logos-blockchain/pull/3198#pullrequestreview-4817050570.

Yes.

@ntn-x2, @hansieodendaal reviewer on the original PR.

N/A

No.

### DIFF
--- a/codec/src/bounded_vec.rs
+++ b/codec/src/bounded_vec.rs
@@ -124,8 +124,8 @@ where
         for _ in 0..len {
             let (next, item) = T::decode(rest, context)?;
 
-            // If we decode more than `ZERO_LENGTH_ELEMENT_MAX` correctly without consuming
-            // the input, bail out to avoid hanging for too long.
+            // If we decode once without consuming any input bail out to avoid hanging for
+            // too long.
             // TODO: This logic can be made a compile-time check once generics become more
             // powerful and we get const generics expressions. For now, we just check it at
             // runtime.
--- a/codec/src/bounded_vec.rs
+++ b/codec/src/bounded_vec.rs
@@ -74,12 +74,6 @@ fn decode_length_prefix<const MAX_LENGTH: usize>(
     }
 }
 
-/// Upper bound, in elements, on how much a decoder may pre-allocate from a
-/// declared length before decoding anything. Beyond this the `Vec` grows
-/// amortised, so the memory a peer can make us commit stays proportional to the
-/// bytes it actually sent.
-const PREALLOC_LIMIT: usize = 1024;
-
 /// Largest `MAX` for which an element type that decodes without consuming input
 /// is still supported.
 ///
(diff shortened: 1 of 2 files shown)


codec/src/bounded_vec.rs
```diff
@@ -110,9 +125,24 @@ where
             return Err(DecodeError::length_out_of_bounds::<Self>(len, MIN, MAX));
         }
 
-        let mut items = Vec::with_capacity(len);
+        // Pre-allocate up to a max limit to avoid memory amplification attacks.
+        let mut items = Vec::with_capacity(len.min(PREALLOC_LIMIT));
+
```
hansieodendaal: Although I originally proposed something like this, looking at it now, I think we could be more vigilant here. For a large element type, a small length prefix can still trigger a very large allocation before any element bytes have been received. For example, a capacity of 1024 for `[u8; 1 MiB]` would request roughly 1 GiB.

For example, we could consider either use `Vec::new()`, or derive the initial capacity from a fixed byte budget using `size_of::<T>()?`. It would also be useful to test this with a large element type rather than only u8/u64.

codec/src/bounded_vec.rs
```diff
@@ -110,9 +125,24 @@ where
...
         for _ in 0..len {
             let (next, item) = T::decode(rest, context)?;
+
+            // If we decode more than `ZERO_LENGTH_ELEMENT_MAX` correctly without consuming
+            // the input, bail out to avoid hanging for too long.
```
hansieodendaal: I think we could improve this comment a bit.

When the type’s declared MAX exceeds 1,024, it rejects immediately on the first element that decodes without consuming input, not if we _decode more than_.

hansieodendaal (commented):
Nice defense-in-depth with associated tests!

Two comments to consider please.

codec/src/bounded_vec.rs
ntn-x2: I agree. Let's go with `Vec::new()` instead. A bit less efficient but safer.

codec/src/bounded_vec.rs
ntn-x2: No the idea is to allow for up to `ZERO_LENGTH_ELEMENT_MAX` zero-size decoding before failing. If the number of decodings is within the limit, we accept it. This is a safeguard against long-lasting loops, so failing after the first decoding is pointless and restricts the set of types that can be decoded. We set a somewhat higher limit (which I think could be made much higher) only for values that would result in a lot of "wasted" CPU, although this is something we developers use, it's not really something that can easily be exploited unless we set the conditions for this to happen. But anyway, I am fine with enforcing this at runtime, but we will migrate to compile-time once generics become more capable.

codec/src/bounded_vec.rs
ntn-x2: https://github.com/logos-blockchain/logos-blockchain/pull/3224/commits/9a832d541b3e01125e97c9a096acaf3bad288d61

codec/src/bounded_vec.rs
ntn-x2: Sorry I confused one thing for another, as I had two things mixed un in my head. Yes you are right, I just updated the comment as this is just a runtime replacement of what will in the future be a compile-time error: https://github.com/logos-blockchain/logos-blockchain/pull/3224/commits/1383049bdb0311608366c8ace7605893c6ba468b. Sorry about the confusion.

hansieodendaal (approved):
LGTM!

PR #3227, opened 2026-07-31
chore: unused code cleanup

ntn-x2:

In an attempt to eliminate code that has been forgotten in between refactor and new implementations, I have asked AI to analyze the codebase and eliminate any component that is not used at the moment. Of course, we might want to keep around functions that are not yet used, but I think we could re-introduce them once they are needed, as they would only potentially increase complexity for everyone else that might need to deal with those unused functions at the moment.

Yes.

@ntn-x2 

N/A

No.


ntn-x2:
@andrussal @hansieodendaal can you guys please check the deletions around cucumber and testing framework do actually make sense?

andrussal:
> @andrussal @hansieodendaal can you guys please check the deletions around cucumber and testing framework do actually make sense?

Looks fine to remove. One thing we could also check across the whole PR is whether these deletions leave more code that has become unused, and clean that up recursively as well(unless it already did that).

ntn-x2:
> Looks fine to remove. One thing we could also check across the whole PR is whether these deletions leave more code that has become unused, and clean that up recursively as well(unless it already did that).

Anything non-public that becomes unused is caught by Clippy and rustc. The AI went one step further and checked public code, and recursively deleted stuff that became unused. So I think it's fine for this one pass.

hansieodendaal (approved):
LGTM!

Nice to see code go away in this way.

ledger/src/mantle/sdp/mod.rs
```diff
@@ -526,15 +526,6 @@ impl SdpLedger {
...
     }
 
-    /// Commitment to the declarations of `service_type`, or `None` if the
-    /// service is not declared.
-    #[must_use]
-    pub fn declarations_root(&self, service_type: ServiceType) -> Option<lb_core::crypto::Hash> {
-        self.services
-            .get(&service_type)
-            .map(|service| service.declarations().root())
-    }
```
youngjoon-lee: This function and other functions that end with `_root` shouldn't be removed because they're going to be used to implement the fast bootstrapping RFC.

ntn-x2: Right. Re-introduced them in https://github.com/logos-blockchain/logos-blockchain/pull/3227/commits/984eb6a87c44a7d8e04a223b3d50002804d7a563.

core/src/mantle/channel.rs
```diff
@@ -219,23 +219,12 @@ impl Channels {
         self.channels.root()
     }
 
-    #[must_use]
-    pub fn channel_notes_root(&self) -> Hash {
-        self.channel_notes.root()
-    }
-
```
youngjoon-lee: This function will be necessary for implementing fast bootstrapping, but it doesn't really matter. We can re-introduce this function when it's necessary.

youngjoon-lee (approved):
thank you!

merkle/blake2btree/src/lib.rs
```diff
@@ -65,9 +65,6 @@ where
 /// Updating an item replaces its leaf with another one, keeping its position.
 pub type Blake2bTree<Key, Item> = MerkleTree<Key, Item, Blake2bLeaf>;
 
-/// Compressed form of a [`Blake2bTree`].
-pub type CompressedBlake2bTree<Key, Item> = CompressedMerkleTree<Key, Item>;
```
danielSanchezQ, line 69: This stuff will conflict with the revert of the fast bootstrapping code. I suggest we do this cleaning after that is revert.

danielSanchezQ (approved):
We should  block this until the fastbootstrapping changes are merged. Otherwise nice to trim this out.

merkle/blake2btree/src/lib.rs
ntn-x2: Nice test for the merge queue logic! 😃 but yeah I'll wait and rebase afterwards.

PR #3228, opened 2026-07-31
chore: implement remaining serde json boundedness

hansieodendaal:

This PR implements the remaining high-value boundedness fixes for:

- inscription Serde conversions;
- MMR sibling-path Serde conversions.

#### Inscription Serde conversions

Inscription Serde decoding now uses a bounded adapter that:

- Checks the maximum hex character length before decoding human-readable input;
- Limits binary sequence decoding to the configured inscription size;
- Rejects oversized input before constructing an unrestricted byte vector;
- Preserves the existing serialization format.

#### MMR sibling-path Serde conversions

MMR sibling paths now deserialize through a bounded visitor that:

- Limits the number of decoded field elements to the supported maximum;
- Rejects oversized JSON and binary paths before building an unrestricted vector;
- Preserves the existing serialized representation.

**Compressed Merkle-tree recovery** => See #3247

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


core/src/codec/bincode/mod.rs
```diff
@@ -51,7 +51,31 @@ pub fn serialized_size<T: Serialize>(item: &T) -> Result<u64> {
 
 /// Deserialize an object directly from bytes
 pub fn deserialize<T: DeserializeOwned>(data: &[u8]) -> Result<T> {
+    deserialize_with_limit(data, MAX_DESERIALIZATION_BYTES)
+}
+
+fn deserialize_with_limit<T: DeserializeOwned>(data: &[u8], max: u64) -> Result<T> {
```
ntn-x2:
```suggestion
fn deserialize_with_limit<T: DeserializeOwned>(data: &[u8], max: usize) -> Result<T> {
```

core/src/codec/mod.rs
```diff
@@ -4,6 +4,14 @@
 pub(crate) mod bincode;
 pub mod errors;
 
+/// Maximum serialized input accepted by the binary deserialization boundary.
+///
+/// This is an operational recovery ceiling rather than a protocol limit. It
+/// is deliberately high enough for resource-rich nodes to restore large
+/// persisted states while ensuring that binary deserialization is not
+/// unbounded.
+pub const MAX_DESERIALIZATION_BYTES: u64 = 8 * 1024 * 1024 * 1024;
```
ntn-x2: Mhh I mean, can we actually comfortably define a codebase-wise limit like this? Would it not be rather be better to have the limit specified on a case-by-case depending on what the node is trying to deserialize? In most cases stuff will have known sizes, and for dynamically-sized types like vectors, sets, and maps, we could define the limit itself. I am just not sure it's a good idea to have a (anyway very large) limit for all deserializations regardless of their context.

utils/src/lib.rs
```diff
@@ -189,5 +193,89 @@ pub mod serde {
                 res.map_err(Error::custom)
             }
         }
+
+        pub fn deserialize_bounded<'de, T, const MAX: usize, D>(
```
ntn-x2, line 193: I think we can update the `deserialize` function to simply call `deserialize_bounded` with `usize::MAX` instead. So we remove duplication.

utils/src/lib.rs
```diff
@@ -189,5 +193,89 @@ pub mod serde {
...
+                deserializer.deserialize_byte_buf(BoundedBytesVisitor::<MAX>)?
+            };
+
+            T::try_from(bytes).map_err(Error::custom)
+        }
+
+        struct BoundedBytesVisitor<const MAX: usize>;
+
+        impl<'de, const MAX: usize> Visitor<'de> for BoundedBytesVisitor<MAX> {
+            type Value = Vec<u8>;
```
ntn-x2: Why not a `UpperBoundedVec<u8, MAX>`?

utils/src/lib.rs
```diff
@@ -189,5 +193,89 @@ pub mod serde {
...
+                    )));
+                }
+                Ok(bytes.to_vec())
+            }
+
+            fn visit_borrowed_bytes<E>(self, bytes: &'de [u8]) -> Result<Self::Value, E>
+            where
+                E: Error,
+            {
+                self.visit_bytes(bytes)
```
ntn-x2: This is already the default implementation, so we can remove the manual implementation of `visit_borrowed_bytes`.

utils/src/lib.rs
```diff
@@ -189,5 +193,89 @@ pub mod serde {
...
+            }
+
+            fn visit_byte_buf<E>(self, bytes: Vec<u8>) -> Result<Self::Value, E>
```
ntn-x2, line 250: This also has a default implementation of `self.visit_bytes(&v)`, so we can avoid manually implementing it.

utils/src/lib.rs
```diff
@@ -189,5 +193,89 @@ pub mod serde {
...
+                    )));
+                }
+                Ok(bytes)
+            }
+
+            fn visit_seq<A>(self, mut sequence: A) -> Result<Self::Value, A::Error>
+            where
+                A: SeqAccess<'de>,
+            {
+                let mut bytes = Vec::with_capacity(sequence.size_hint().unwrap_or(0).min(MAX));
```
ntn-x2: I think we can check for `bytes.len()` before we call `next_element()`, so we avoid consuming the collection if we are already at the maximum number of expected elements. But I think we could go even further and simply collect the sequence into a vector and call one of the other visitor methods instead, with the goal of, once again, reducing duplication.

mmr/src/path.rs
```diff
@@ -129,7 +132,36 @@ mod serde_siblings {
...
+            fn expecting(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                formatter.write_str("a sequence of MMR sibling field elements")
+            }
+
+            fn visit_seq<A>(self, mut sequence: A) -> Result<Self::Value, A::Error>
+            where
+                A: SeqAccess<'de>,
+            {
+                let mut siblings = Vec::with_capacity(MAX_MERKLE_PATH_SIBLINGS);
+                while let Some(FrWrap(sibling)) = sequence.next_element()? {
```
ntn-x2, line 152: Can't we simply collect the sequence into an array of the expected size? So this will automatically fail if we have more or less elements than the expected number.

ntn-x2: Also I feel maybe we can abstract this implementation into a generic visitor instead, and expose a utility function for it, instead of implementing a visitor for each type we need it for? Since the logic seems to be the same, minus the type of the collection we iterate over.

merkle/dynamic-merkle/src/lib.rs
```diff
@@ -478,43 +491,60 @@ impl<H: MerkleHasher> DynamicMerkleTree<H> {
...
-
         Self {
             root,
-            holes,
+            holes: RedBlackTreeSetSync::new_sync(),
             _hasher: PhantomData,
         }
     }
+
+    fn build_sparse_subtree<I>(
```
ntn-x2: Isn't this out of scope? I would rather have this logic in a different PR since this PR is about serialization bounds.

utils/src/lib.rs
ntn-x2: We could actually just call this function `deserialize` and use a default of `usize::MAX` for the `MAX` generic. Then we would have a single function instead.

ntn-x2 (commented):
Good idea to tackle this issue. Nevertheless, I think we can make it in a simpler way by writing less lines of code, and also I would rather have the merkle compression logic in a separate PR, if possible.

danielSanchezQ (changes requested):
I agree with @ntn-x2 , could we have the sparse tree logic in a separated PR? Also different people need to check on that too.

mmr/src/path.rs
hansieodendaal: > Can't we simply collect the sequence into an array of the expected size? So this will automatically fail if we have more or less elements than the expected number.

The expected sibling count is contextual rather than globally fixed. Deserialization only knows the global maximum; the exact `max_height - 1` count is checked later by `validate_for_height`. An array of `MAX_MERKLE_PATH_SIBLINGS` would reject valid paths for smaller MMR heights.

hansieodendaal: > Also I feel maybe we can abstract this implementation into a generic visitor instead, and expose a utility function for it, instead of implementing a visitor for each type we need it for? Since the logic seems to be the same, minus the type of the collection we iterate over.

I agree that a reusable generic bounded-sequence visitor would be useful, but I would prefer to handle that as part of #3233 rather than introduce the wider abstraction in this focused PR.

For now, this PR keeps the implementation local while ensuring that oversized sibling sequences are rejected before an unrestricted vector is constructed.

utils/src/lib.rs
hansieodendaal: - Check before `next_element`: not applied because that rejects valid input containing exactly `MAX` elements. `try_push` now centralizes the correct bound check.
- Collect the sequence first: intentionally not applied because an unrestricted intermediate Vec would defeat early boundedness.

utils/src/lib.rs
hansieodendaal: Changes applied in bb0dfc4643ea1bf9abaa56acfa1cc8c3e5db5af9

utils/src/lib.rs
hansieodendaal: Changes applied in bb0dfc4643ea1bf9abaa56acfa1cc8c3e5db5af9

utils/src/lib.rs
hansieodendaal: Changes applied in bb0dfc4643ea1bf9abaa56acfa1cc8c3e5db5af9

utils/src/lib.rs
hansieodendaal: Changes applied in bb0dfc4643ea1bf9abaa56acfa1cc8c3e5db5af9

utils/src/lib.rs
```diff
@@ -175,18 +181,96 @@ pub mod serde {
...
+                Ok(UpperBoundedVec::new_unchecked(bytes))
+            }
+
+            fn visit_seq<A>(self, mut sequence: A) -> Result<Self::Value, A::Error>
+            where
+                A: SeqAccess<'de>,
+            {
+                // An empty vector always satisfies an upper-only bound.
+                let capacity = sequence.size_hint().unwrap_or(0).min(MAX);
+                let mut bytes = UpperBoundedVec::new_unchecked(Vec::with_capacity(capacity));
```
ntn-x2, line 267: Maybe would be good to have a `with_capacity` method for a bounded vec as well, that checks whether the capacity is within the `MIN..MAX` bounds?

ntn-x2 (approved):
Looks good. Just a comment that is probably not going to be addressed in this PR.

utils/src/lib.rs
hansieodendaal: Yeah, good idea, created an issue for this: #3261.

danielSanchezQ (approved):

PR #3230, opened 2026-07-31
chore: fix remaining alternate wire deserialisation paths boundedness

hansieodendaal:

This PR implements the remaining alternate wire/deserialisation paths boundedness fixes:

- `MantleTx` binary Serde validation;
- `PaddedPayloadBody` length validation;
- chain-sync sender frame-size enforcement.

#### `MantleTx` binary Serde

The binary `MantleTx` Serde path now:

- Bounds the opaque encoded transaction envelope using the existing `MAX_BLOCK_TRANSACTIONS_SIZE` limit;
- Rejects valid transactions followed by trailing bytes;
- Matches the stricter behavior already present for `SignedMantleTx`.

Added regression tests:
- `binary_serde_rejects_trailing_bytes_inside_transaction_envelope`
- `binary_serde_rejects_oversized_transaction_envelope`

#### `PaddedPayloadBody` validation

`PaddedPayloadBody` now validates `actual_len` immediately during:

- Serde deserialization;
- Binary decoding.

Invalid lengths are rejected before the object can expose an invalid body state.

Added regression tests:
- `binary_decode_rejects_invalid_actual_length`
- `serde_deserialize_rejects_invalid_actual_length`

#### Chain-sync sender frame limit

The sender now rejects messages larger than `MAX_MSG_LEN` before writing the length prefix. This keeps sender behavior consistent with receiver-side enforcement and avoids producing frames that conforming peers must reject.

Added regression test:
- `sender_rejects_messages_above_frame_limit`

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


core/src/mantle/transactions/mantle_tx.rs
```diff
@@ -183,11 +184,91 @@ impl<'de> Deserialize<'de> for MantleTx {
...
+            } else {
+                Err(serde::de::Error::custom(
+                    "MantleTx binary encoding contains trailing bytes",
+                ))
+            }
+        }
+    }
+}
+
+fn deserialize_bounded_bytes<'de, const MAX: usize, D>(deserializer: D) -> Result<Vec<u8>, D::Error>
```
ntn-x2:
```suggestion
fn deserialize_bounded_bytes<'de, const MAX: usize, D>(deserializer: D) -> Result<UpperBoundedVec<u8, MAX>, D::Error>
```

core/src/mantle/transactions/mantle_tx.rs
```diff
@@ -183,11 +184,91 @@ impl<'de> Deserialize<'de> for MantleTx {
...
+where
+    D: Deserializer<'de>,
+{
+    struct Visitor<const MAX: usize>;
+
+    impl<'de, const MAX: usize> serde::de::Visitor<'de> for Visitor<MAX> {
+        type Value = Vec<u8>;
```
ntn-x2:
```suggestion
        type Value = UpperBoundedVec<u8, MAX>;
```

core/src/mantle/transactions/mantle_tx.rs
```diff
@@ -183,11 +184,91 @@ impl<'de> Deserialize<'de> for MantleTx {
...
+                )));
+            }
+            Ok(bytes.to_vec())
+        }
+
+        fn visit_borrowed_bytes<E>(self, bytes: &'de [u8]) -> Result<Self::Value, E>
+        where
+            E: serde::de::Error,
+        {
+            self.visit_bytes(bytes)
```
ntn-x2: This is a default impl, as I mentioned in the other PR. We can remove this manual impl.

core/src/mantle/transactions/mantle_tx.rs
ntn-x2: Can't this be made the default `Deserialize` implementation for a bounded vector directly?

consensus/cryptarchia-sync/src/libp2p/packing.rs
```diff
@@ -29,15 +29,7 @@ where
     Writer: AsyncWriteExt + Send + Unpin,
 {
     let packed_message = message.to_bytes()?;
```
ntn-x2, line 31: I mean, ideally we would change the signature of `to_bytes` or add a `to_bounded_bytes` that returns a bounded vector instead? Then we don't need to check the length prefix anymore since if the conversion succeeds, we know we have a collection of bytes below the expected maximum.

blend/message/src/message/payload.rs
```diff
@@ -69,6 +69,33 @@ pub struct PaddedPayloadBody {
...
 
+#[serde_as]
+#[derive(Deserialize)]
+struct PaddedPayloadBodySerde {
+    actual_len: u16,
+    #[serde_as(as = "serde_with::Bytes")]
+    padded: Box<[u8; MAX_PAYLOAD_BODY_SIZE]>,
+}
+
+impl<'de> Deserialize<'de> for PaddedPayloadBody {
```
ntn-x2: Couldn't this just be done with a simple deserialize function that we could use to decorate the `actual_len` field? The `padded` field logic is already good, and since we just want to check that the `actual_len` is less than a constant, we can simply use a very small custom deserialize function for it and use `#[serde(deserialize_with = "path")]`. But good catch nonetheless.

blend/message/src/message/payload.rs
```diff
@@ -124,6 +151,13 @@ impl BinaryDecode for PaddedPayloadBody {
         (): &Self::Context,
     ) -> Result<(&'input [u8], Self), DecodeError> {
         let (input, actual_len) = u16::decode(input, &())?;
+        if usize::from(actual_len) > MAX_PAYLOAD_BODY_SIZE {
+            return Err(DecodeError::length_out_of_bounds::<Self>(
```
ntn-x2, line 142: This is more an `InvalidValue` type of error, not really a length out of bounds.

hansieodendaal: I would disagree:
- `InvalidValue`: bytes were well-sized but semantically invalid.
- `LengthOutOfBounds`: a decoded length fell outside its permitted bounds.

consensus/cryptarchia-sync/src/libp2p/packing.rs
hansieodendaal: Good suggestion; I agree that a bounded serialization API would be the cleaner abstraction, bit within scope of this PR, so I will make a separate PR for it.

core/src/mantle/transactions/mantle_tx.rs
hansieodendaal: Changes in https://github.com/logos-blockchain/logos-blockchain/pull/3230/commits/a8bb4af1bccf6790f60eac44588dbf9a71b4ca6e

core/src/mantle/transactions/mantle_tx.rs
hansieodendaal: Changes in https://github.com/logos-blockchain/logos-blockchain/commit/a8bb4af1bccf6790f60eac44588dbf9a71b4ca6e

blend/message/src/message/payload.rs
hansieodendaal: Changes in https://github.com/logos-blockchain/logos-blockchain/commit/a8bb4af1bccf6790f60eac44588dbf9a71b4ca6e

consensus/cryptarchia-sync/src/libp2p/packing.rs
hansieodendaal: See new issue created: #3241

core/src/mantle/transactions/mantle_tx.rs
ntn-x2: I guess we could have a newtype for it, if we want to be full Rusty, that wraps a `u16` and has also a `Tryfrom` implementation.

ntn-x2 (approved):
Looks good. I think we could have a newtype for the deserializable actual length of a padded payload body (my bad, I should have introduced it), but otherwise looks good, thanks!

hansieodendaal:
> Looks good. I think we could have a newtype for the deserializable actual length of a padded payload body (my bad, I should have introduced it), but otherwise looks good, thanks!

See #3244

core/src/mantle/transactions/mantle_tx.rs
hansieodendaal: See https://github.com/logos-blockchain/logos-blockchain/issues/3244

youngjoon-lee (approved):

PR #3237, opened 2026-08-02
fix(zone-sdk): shed orphans on config update

### LINKED ISSUE #3236: Zone-sdk: shed invalidated inscriptions on config update
Config update is changing the tip which can invalidate pending inscriptions. Zone-sdk should emit them as orphans. 

This is a current solution and might need to be updated when the config lineage is introduced.

pradovic:

There was a bug in zone-sdk as it was not emitting orphaned inscriptions on config update. This PR fixes this by considering config update as well on `detect_channel_update`.

Closes https://github.com/logos-blockchain/logos-blockchain/issues/3236.

Yes

@pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.

### DIFF
--- a/zone-sdk/src/sequencer/actor.rs
+++ b/zone-sdk/src/sequencer/actor.rs
@@ -474,16 +474,19 @@ where
             Some(update) => {
                 Self::log_channel_update(&update);
 
+                let new_channel_tip = update.new_channel_tip;
+                let built = self.build_channel_update(update);
+
                 let has_pending = self
                     .state
                     .as_ref()
                     .is_some_and(TxState::has_pending_inscriptions);
 
-                if !update.orphaned.is_empty() || !has_pending {
-                    self.last_msg_id = update.new_channel_tip;
+                if !built.orphaned.is_empty() || !has_pending {
+                    self.last_msg_id = new_channel_tip;
                 }
 
-                self.build_channel_update(update)
+                built
             }
             None => ChannelUpdate {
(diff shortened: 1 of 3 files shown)


hansieodendaal (commented):
Good direction.

Could you add a unit test through the block-event/actor path that proves both of these?

* A config-only update is processed, with the invalidated pending inscription emitted as orphaned and removed from pending state.
* The resulting checkpoint uses the config tip as `last_msg_id`.

I’m not convinced the current test proves this because it calls `detect_channel_update()` and `shed_off_branch_pending()` separately. In the production path, the empty update may still be filtered out before shedding, and `last_msg_id` appears to be updated before the stale pending inscription is removed.

pradovic:
> Good direction.
> 
> Could you add a unit test through the block-event/actor path that proves both of these?
> 
>     * A config-only update is processed, with the invalidated pending inscription emitted as orphaned and removed from pending state.
> 
>     * The resulting checkpoint uses the config tip as `last_msg_id`.
> 
> 
> I’m not convinced the current test proves this because it calls `detect_channel_update()` and `shed_off_branch_pending()` separately. In the production path, the empty update may still be filtered out before shedding, and `last_msg_id` appears to be updated before the stale pending inscription is removed.

You are right! Good catch! Fixing, thanks!

hansieodendaal (approved):
LGTM

PR #3243, opened 2026-08-03
fix: serialization logic for reference txs in a block

ntn-x2:

On the way to implement the [block compression proposal](https://github.com/logos-co/logos-lips/pull/389), I noticed that the current codebase does not (de)serialize tx references in a block proposal properly. We serialize and deserialize using bincode, but that is different than what the spec defines. Also, in the upcoming block compression changes, we will use a bounded vector with a 2-byte length prefix as we do for mantle components, to make it easier to understand. The use of a 2-byte length prefix renders useless also adding the number of elements to a block's header, reducing the blast radius of the new spec.

Hence, this PR:

1. Fixes those divergences by implementing `BinaryEncode` and `BinaryDecode` on all components of a block proposal, including the block proposal itself
2. Remove the `NetworkMessage` enum wrapping when sending and parsing a block proposal via the gossipsub network: a gossipsub for a specific topic already expects a specific message type, so we don't need to wrap that in an enum.
3. Use `BinaryEncode` and `BinaryDecode` instead of encoding using bincode when sending to and receiving from the wire a block proposal.

Yes.

@ntn-x2 implementor, @madxor spec writer

We are in the process of updating it.

Kinda. We need to specify the wire format for a block proposal. Since we are implementing binary encoding/decoding instead of relying on bincode, we can fully specify the format of a block proposal. I will work closely with @madxor to add that to the block construction spec that is being updated as part of the block compression LIP.


services/chain/chain-common/src/messages.rs
```diff
@@ -1,7 +0,0 @@
-use lb_core::block::Proposal;
-use serde::{Deserialize, Serialize};
-
-#[derive(Clone, Debug, Serialize, Deserialize)]
-pub enum NetworkMessage {
-    Proposal(Proposal),
-}
```
danielSanchezQ, line 7: Vestiges of the past

danielSanchezQ (approved):

services/chain/chain-common/src/messages.rs
ntn-x2: The whole crate is now gone 👋👋👋

PR #3246, opened 2026-08-03
feat(zone-sdk): channel wallet tracker

pradovic:

This PR is adding channel wallet view and interface to enable atomic_withdraw and later atomic_deposits with new channel notes model.

**How PR 2 will use this:** 
- `publish_atomic_withdraw` will build a single tx of [ChannelInscribe, ChannelTransfer, ChannelWithdraw]. The transfer consumes channel notes selected from this PR's wallet (explicit note ids chosen by the zone, or a mechanical default preferring finalized, oldest notes) and cuts exact recipient notes plus change; the withdraw then releases the recipient notes, whose ids are derived locally from the transfer op. 
- The SDK builds [CHANNEL_INSCRIBE, CHANNEL_TRANSFER, CHANNEL_WITHDRAW] client-side. The transfer's inputs are channel-note ids selected from the tracked set, outputs are the recipient + change notes, and the withdraw's inputs are the recipient note ids derived locally (derive_note_id over the transfer's op id). Channel ops need no wallet help: their inputs are ids the ledger validates against channel_notes, and their proofs are Ed25519 threshold signatures from the accredited key — no ZkSignature, no note ownership lookup.
- The SDK sends those ops to wallet/fund. The node computes the gas over them (opaque payload from its perspective), appends a TRANSFER spending its own funding_pk notes, and returns the funded tx plus the transfer's ZkSignature proof over the final tx hash.
- The SDK signs that funded hash with the sequencer key, builds the ChannelMultiSigProofs for transfer/withdraw and the Ed25519Sig for the inscribe, and attaches the wallet's transfer proof last, matching op order.

Selection and reservation happen inside the drive loop against the wallet's tip view, so a bundle can never double-spend a note already claimed by an in-flight withdraw. Consumers use the same view read-only via channel_wallet() (channel balance, deposit crediting), and the finalized `ChannelTransfer` ops surfaced in `BlocksProcessed` let a zone replicate its note state event-sourced. The wallet base rides in`SequencerCheckpoint`, so restarts restore it and backfill replays only the gap since the checkpoint's LIB.

Ref spec: https://github.com/logos-co/logos-lips/blob/master/docs/blockchain/raw/bedrock-v1.1-mantle-specification.md.

For easier review, as tests have a lot of lines of code:
- types: ChannelNote/DepositGroup/ChannelWalletView + FinalizedOp::ChannelTransfer + match arms
- channel_wallet.rs: the tracker + its unit tests
- wiring: block_fetch paths + process_block signature
- exposure: getter/handle/client + checkpoint field
- cucumber

Yes

@thomaslavaur @pradovic 

Yes

No

## Checklist

> [!WARNING]  

* [ ] 2. Description added.


thomaslavaur (commented):
I think it's good from what I understand

hansieodendaal (commented):
Nice addition.

A dust attack will break this model: deposits are permissionless, so an adversary can saturate a channel with thousands of 1 LGO deposits. The channel wallet should retain the age, public key, and value of every deposited note. `DepositGroup` is the wrong public model here.

The withdrawal client should also be able to determine its dust-spend policy—for example, ignoring values below 100 LGO, or opportunistically consuming dust alongside a withdrawal when it fits within the input limit.

I think we also need a cucumber test covering dust selection, either in this PR or the next: create 300 low-value outputs (1 LGO each) in one block followed by several larger deposits (10,000 LGO each) in a later blocks, deposit them into the channel, then withdraw from the channel.

The test should verify that the withdrawal remains valid under the 255-input limit and that the configured dust policy determines whether the 1 LGO notes are ignored or opportunistically spent.

pradovic:
> Nice addition.
> 
> A dust attack will break this model: deposits are permissionless, so an adversary can saturate a channel with thousands of 1 LGO deposits. The channel wallet should retain the age, public key, and value of every deposited note. `DepositGroup` is the wrong public model here.
> 
> The withdrawal client should also be able to determine its dust-spend policy—for example, ignoring values below 100 LGO, or opportunistically consuming dust alongside a withdrawal when it fits within the input limit.
> 
> I think we also need a cucumber test covering dust selection, either in this PR or the next: create 300 low-value outputs (1 LGO each) in one block followed by several larger deposits (10,000 LGO each) in a later blocks, deposit them into the channel, then withdraw from the channel.
> 
> The test should verify that the withdrawal remains valid under the 255-input limit and that the configured dust policy determines whether the 1 LGO notes are ignored or opportunistically spent.

I agree, I think we can solve this by extending the deposit event -> per-note (note_id, value, pk) in TxEventPayload::Deposit. I will open a separate PR for this and then update this PR after it is merged. This should simplify the existing model and make it more correct. Thanks!

Withdrawal client will be in a PR after this, and then we will tackle the policy.

@thomaslavaur Do you agree? Could you also take a look if we missed something in logic?

thomaslavaur:
No agreed. The sequencers could even chose to ignore notes under a certain threshold in value if it's clear for the entire zone

tests/src/cucumber/steps/manual_zone/steps.rs
```diff
@@ -667,6 +668,36 @@ async fn step_submit_zone_deposit_transaction(
     .await
 }
 
+#[when(
+    expr = "I submit zone deposit transaction {string} into channel of {string} consuming notes valued {string} with metadata {string}"
+)]
```
hansieodendaal, line 673: Is the expectation that this function return all matching notes, or just one of each specified if found?

hansieodendaal (commented):
Looking good, thanks. Some comments here and below, some possible for follow-up PR(s):

- I think we should version checkpoints, if versions do not match, restore fully.
- We still need to demonstrate in cucumber that a dust attack will not succeed. The scenario should create >1 coinsplits so that we have >255 similar small-valued UTXOs in the same block that is below the minimum fee when summed.

tests/src/cucumber/steps/manual_zone/steps.rs
pradovic: It consumes exactly one note per listed value, in order — valued "1,2" picks one 1-LGO UTXO and one 2-LGO UTXO

pradovic:
> Looking good, thanks. Some comments here and below, some possible for follow-up PR(s):
> 
>     * I think we should version checkpoints, if versions do not match, restore fully.
> 
>     * We still need to demonstrate in cucumber that a dust attack will not succeed. The scenario should create >1 coinsplits so that we have >255 similar small-valued UTXOs in the same block that is below the minimum fee when summed.

Thanks for the reivew. 

The dust test is agreed for the second PR when we have withdraw implemented and enabled as well.  I still think it is better to add it there as this is a fairly large PR already, and the dust test will be better with withdraw implemented.

Regarding checkpoint versioning, it's a nice to have but still not mandatory I think. We can do it when we have time, after all the high-priority features are implemented.

hansieodendaal (approved):
LGTM!

PR #3247, opened 2026-08-03
fix: make compressed merkle-tree recovery bounded

hansieodendaal:

This PR implements bounded compressed Merkle-tree recovery, split from #3228. It eliminates deserialization work amplification where a small compressed input containing a high leaf position could trigger billions of explicit hole insertions.

It now:
- Uses checked `TryFrom` conversion instead of infallible From;
- Rejects positions outside the fixed `2^32` tree capacity;
- Rejects duplicate logical keys and duplicate positions;
- Uses sparse reconstruction rather than materializing every missing position;

The previous recovery implementation treated every gap as an explicit hole:
```
while current_pos < pos {
    tree = tree.insert_hole(current_pos);
    current_pos += 1;
}
```

This made recovery complexity depend on the highest encoded position rather than the number of stored items. A valid item at position `2^31 + 1` would require more than two billion hole insertions before recovery could complete.

Sparse recovery introduces implicit empty subtrees, so the dynamic tree also needed to:

- Hash empty left subtrees correctly;
- Find the lowest empty position from the tree structure;
- Avoid relying exclusively on the separately stored hole set during insertion.

The regression test `test_sparse_recovery_does_not_materialize_gaps` reconstructs a tree with an item at `(1 << 31) + 1`, then verifies that the next insertion uses position `0`. This demonstrates that the previous “everything is an explicit hole” approach was not merely inefficient—it was infeasible in practice and would likely time out or exhaust memory.

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


merkle/dynamic-merkle/src/lib.rs
```diff
@@ -478,43 +497,60 @@ impl<H: MerkleHasher> DynamicMerkleTree<H> {
...
-
         Self {
             root,
-            holes,
+            holes: RedBlackTreeSetSync::new_sync(),
             _hasher: PhantomData,
         }
     }
+
+    fn build_sparse_subtree<I>(
```
youngjoon-lee, line 562: It'd be great if we add a doc comment for this function, because the algorithm is not very simple to read.

merkle/dynamic-merkle/src/lib.rs
```diff
@@ -478,43 +497,60 @@ impl<H: MerkleHasher> DynamicMerkleTree<H> {
...
+            return Arc::new(Node::Empty { height });
+        };
+        let capacity = 1usize << height;
+        assert!(
+            position >= start && position - start < capacity,
+            "indices must be strictly increasing and within bounds"
+        );
+        if height == 0 {
+            let (position, value) = items.next().expect("peeked item must be available");
+            assert_eq!(position, start, "indices must be strictly increasing");
```
youngjoon-lee: Could you please explain what `start` means here and why `position` should be `== start`?

merkle/dynamic-merkle/src/lib.rs
```diff
@@ -478,43 +497,60 @@ impl<H: MerkleHasher> DynamicMerkleTree<H> {
...
+        };
+        let right = if items
+            .peek()
+            .is_some_and(|(position, _)| *position < start + capacity)
+        {
+            Self::build_sparse_subtree(items, midpoint, height - 1)
+        } else {
+            Arc::new(Node::Empty { height: height - 1 })
+        };
+        Arc::new(Node::new_inner::<H>(left, right))
```
youngjoon-lee: What if `midpoint` is out of range of `capacity` because `start = 9` and `capacity = 10`?

merkle/dynamic-merkle/src/lib.rs
youngjoon-lee: Can we now remove this `holes` field?

youngjoon-lee: If I understand the changes correctly, now the comment `// Can only be found in the right...` is outdated.

youngjoon-lee, line 108: It seems that we can simplify this function as `H::compress(&left.value::<H>(), &right.value::<H>())`.

merkle/tree/src/lib.rs
```diff
@@ -268,38 +272,104 @@ where
     }
 }
 
-impl<Key, Item, Leaf> From<CompressedMerkleTree<Key, Item>> for MerkleTree<Key, Item, Leaf>
+#[derive(Error, Debug, Clone, PartialEq, Eq)]
+pub enum RecoveryError {
+    #[error("compressed Merkle tree position {position} exceeds capacity {capacity}")]
+    PositionOutOfBounds { position: usize, capacity: usize },
+    #[error("compressed Merkle tree contains duplicate keys")]
+    DuplicateKey,
```
youngjoon-lee: Maybe we can add a `position` to this error type for easy debugging.

merkle/dynamic-merkle/src/lib.rs
thomaslavaur: How would you recover removed leaves then?

merkle/dynamic-merkle/src/lib.rs
```diff
@@ -465,8 +484,8 @@ impl<H: MerkleHasher> DynamicMerkleTree<H> {
         })
     }
 
-    /// Rebuilds a tree placing each leaf hash at its given index, filling the
-    /// gaps between indices with holes.
+    /// Rebuilds a tree placing each leaf hash at its given index, representing
+    /// gaps as implicit empty subtrees.
```
thomaslavaur: subtrees?? You mean empty tree of depth 0?

thomaslavaur (commented):
Could you explain how holes are managed now? I didn't get it

merkle/dynamic-merkle/src/lib.rs
hansieodendaal: Yes. Sparse recovery can create empty subtrees on either side now. I removed the right-subtree-only assumption.

hansieodendaal: Agreed - `value()` now handles `Node::Empty` correctly as well, so the duplicated matching was no longer necessary.

hansieodendaal: Doc comment updated.

`holes` is no longer used to find the next insertion position; `first_empty_index()` derives that from the tree.

I kept `holes` because removals are still recorded there and it is part of the existing `DynamicMerkleTree` serialized format. Removing it would be a separate serialization change.

For compressed recovery, missing positions are simply treated as free positions.

merkle/dynamic-merkle/src/lib.rs
hansieodendaal: Doc csomment updated.

`Node::Empty { height: 0 }` represents one empty leaf, while larger heights compactly represent the entire `2^height` empty range.

merkle/dynamic-merkle/src/lib.rs
hansieodendaal: Doc comment added.

merkle/dynamic-merkle/src/lib.rs
hansieodendaal: Doc comment added.

`start` was the absolute first leaf index covered by that recursive subtree. I renamed it to `subtree_start` and documented the interval. At height `0` the subtree covers exactly one `position`, so the item must be at `subtree_start`.

merkle/dynamic-merkle/src/lib.rs
hansieodendaal: Doc comment added.

`capacity` (renamed to `subtree_capacity`) here is the subtree width rather than an absolute position limit. The represented range is `[subtree_start, subtree_start + subtree_capacity]`, so the midpoint is inside that range.

merkle/tree/src/lib.rs
hansieodendaal: `DuplicateKey` now carries the position of the duplicate occurrence.

hansieodendaal:
> Could you explain how holes are managed now? I didn't get it

Sparse gaps are now represented directly in the tree as `Node::Empty { height }`, which can cover many empty positions without materializing them individually.

Positions explicitly freed by `remove()` remain `Leaf(None)` and are still tracked in `holes`.

`insert()` now uses `first_empty_index()` on the tree, so it finds the lowest free position in either representation. If that position is also in `holes`, it is removed from the set.

For compressed recovery, only occupied positions are serialized, so any missing position is treated as free.

youngjoon-lee:
@hansieodendaal @thomaslavaur I think, most of confusion we're having comes from the fact that we still have the `holes` field, but it is not read anywhere. Please correct me if I'm wrong. And, the internal shape of a tree is not preserved during a compressed serde roundtrip, because the holes are restored as empty subtrees.

I'm not saying that this asymmetric compressed serde is not good. It does its job and makes the deserialization time reasonable. But, I'm thinking we may be able to make it symmetric.

We can change the removal operation. Instead of filling an empty position with a hole, we can fill it with an empty subtree, as this PR is doing when deserializing a tree. If two siblings are empty, we can merge them as a bigger empty subtree. I believe, it's not expensive.

In this way, the serde becomes completely symmetric. The tree has a single/consistent behavior.

If this idea works well, we can simplify `Node::Leaf { value: Option<Hash> }` to `Node::Leaf { value: Hash }`, because we're gonna use `Node::Empty { height: 0 }` to represent a hole.

I believe, this is an ultimate way that is easier to understand/maintain. Wdyt? I can also try it if you’re busy with the others.

hansieodendaal:
> @hansieodendaal @thomaslavaur I think, most of confusion we're having comes from the fact that we still have the `holes` field ...

Yes, this is a much better direction @youngjoon-lee! Proposal implemented, we can evaluate it.

merkle/dynamic-merkle/src/lib.rs
```diff
@@ -462,8 +509,10 @@ impl<H: MerkleHasher> DynamicMerkleTree<H> {
         self.root.path::<H>(index)?.try_into().ok()
     }
 
-    /// Rebuilds a tree placing each leaf hash at its given index, filling the
-    /// gaps between indices with holes.
+    /// Rebuilds a tree placing each leaf hash at its given index, representing
+    /// gaps as implicit empty subtrees. Empty sibling subtrees are represented
```
youngjoon-lee:
```suggestion
    /// gaps as empty subtrees. Empty sibling subtrees are represented
```
I think, now we don’t need to call it “implicit” empty subtrees? Pls let me know if I misunderstood the meaning of “implicit” here.

merkle/dynamic-merkle/src/lib.rs
```diff
@@ -475,43 +524,77 @@ impl<H: MerkleHasher> DynamicMerkleTree<H> {
...
-    // This is only for maintaining holes information when recovering
-    // the tree from a compressed format, should not be used otherwise.
-    fn insert_hole(&self, index: usize) -> Self {
+        let mut items = items.into_iter().peekable();
+        let root = Self::build_sparse_subtree(&mut items, 0, TREE_HEIGHT_EXCEPT_ROOT);
         assert!(
-            index < self.root.capacity(),
-            "Index out of bounds for inserting an empty node"
+            items.next().is_none(),
+            "indices must be strictly increasing and within bounds"
```
youngjoon-lee: I guess, this panic message is outdated? Isn’t this for checking if `build_sparse_subtree` consumed all items?

merkle/dynamic-merkle/src/lib.rs
youngjoon-lee: Thank you. 

What I asked was at least having a doc comment about what this function does and what the sparse subtree means. 

Now, the first sentence is good. Plus, I think it’d be great if we move the last sentence “represents unoccupied ranges with…” to the end of the first sentence, because it explains what “sparse” means, if I understood it correctly. 

Other paragraphs are good to have, especially how the `items` iterator should behave.

merkle/dynamic-merkle/src/lib.rs
```diff
@@ -475,43 +524,77 @@ impl<H: MerkleHasher> DynamicMerkleTree<H> {
...
+        let subtree_capacity = 1usize << height;
+        assert!(
+            position >= subtree_start && position - subtree_start < subtree_capacity,
+            "indices must be strictly increasing and within bounds"
+        );
+        if height == 0 {
+            let (position, value) = items.next().expect("peeked item must be available");
+            assert_eq!(
+                position, subtree_start,
+                "indices must be strictly increasing"
```
youngjoon-lee: If I understood the doc comment correctly, I think this panic message should be “the consumed items must be at the subtree_start position since this subtree should represent an empty leaf”.

merkle/dynamic-merkle/src/lib.rs
```diff
@@ -150,14 +156,64 @@ impl<Hash> Node<Hash> {
 }
 
 impl<Hash: Copy> Node<Hash> {
+    fn canonicalize<H>(node: &Arc<Self>) -> Result<Arc<Self>, &'static str>
```
youngjoon-lee: What does `canonicalize` mean here? A doc comment should explain it. It seems that this function just checks if all `height` values are correct, and if left/right siblings have the same height, correct? 
Also, there is a similar assertion in the `new_inner` which is called by this function. 

Is this function for checking invariants when deserializing a tree?

merkle/dynamic-merkle/src/lib.rs
```diff
@@ -524,22 +607,21 @@ impl<H: MerkleHasher> Eq for DynamicMerkleTree<H> {}
 
 /// [`serde`](::serde) support for [`DynamicMerkleTree`].
 ///
-/// The tree serializes as its root node and the set of holes; on
-/// deserialization the two are reassembled into a tree. Requires the hasher's
+/// The tree serializes its canonical root node; deserialization canonicalizes
```
youngjoon-lee: What does “canonical root” mean here? Do we have a concept of non-canonical root? I’m asking to know if I’m missing anything. 

If not, I think we can simply call it “root” to avoid confusion.

merkle/tree/src/lib.rs
```diff
@@ -268,38 +272,106 @@ where
...
 }
 
-#[derive(::serde::Serialize, ::serde::Deserialize)]
+#[derive(::serde::Serialize)]
 #[serde(transparent)]
 pub struct CompressedMerkleTree<Key, Item> {
     items: BTreeMap<usize, (Key, Item)>,
 }
 
+impl<'de, Key, Item> ::serde::Deserialize<'de> for CompressedMerkleTree<Key, Item>
```
youngjoon-lee, line 336: Why do we need this impl? Deriving `Deserialize` is not sufficient? I’d like to understand why deserializing a compress tree should be changed in this PR, because I thought that this PR touches only the way how a dynamic merkle tree is recovered from a compress merkle tree.

youngjoon-lee (commented):
Thank you. The logic looks good.

I left some comments for readability, and some questions.

merkle/dynamic-merkle/src/lib.rs
youngjoon-lee: If I understood correctly, the word “canonical” implies an old context that the previous serde was asymmetric. 

Now that we have the symmetric serde scheme, a tree, recovered by this code, always has the same shape as a tree before the serialization. So, I think, we don’t really need to call this logic “canonicalize”. We don’t need to (shouldn’t) worry about asymmetricity if our code is correct.

But I think I am missing something. Please take this comment as a minor opinion. If the current name is good in your opinion, please go ahead and

merkle/dynamic-merkle/src/lib.rs
hansieodendaal: See updates in https://github.com/logos-blockchain/logos-blockchain/pull/3247/commits/a4addfbb3cb46a548766777518c3d7f11b5ff4b8.

merkle/dynamic-merkle/src/lib.rs
hansieodendaal: See updates in https://github.com/logos-blockchain/logos-blockchain/pull/3247/commits/a4addfbb3cb46a548766777518c3d7f11b5ff4b8.

merkle/dynamic-merkle/src/lib.rs
hansieodendaal: Actually, the assertion was redundant.

See updates in https://github.com/logos-blockchain/logos-blockchain/pull/3247/commits/a4addfbb3cb46a548766777518c3d7f11b5ff4b8.

merkle/dynamic-merkle/src/lib.rs
hansieodendaal: Function renamed and doc comment added.

See updates in https://github.com/logos-blockchain/logos-blockchain/pull/3247/commits/a4addfbb3cb46a548766777518c3d7f11b5ff4b8.

merkle/dynamic-merkle/src/lib.rs
hansieodendaal: Doc comment updated.

See updates in https://github.com/logos-blockchain/logos-blockchain/commit/a4addfbb3cb46a548766777518c3d7f11b5ff4b8.

merkle/tree/src/lib.rs
hansieodendaal: Explanation added in doc comment.

See updates in https://github.com/logos-blockchain/logos-blockchain/commit/a4addfbb3cb46a548766777518c3d7f11b5ff4b8.

youngjoon-lee (approved):
Thank you so much for taking my comments into account, Hansie, even though my review was slow. I think, now the structure is clean and efficient.

hansieodendaal:
> Thank you so much for taking my comments into account, Hansie, even though my review was slow. I think, now the structure is clean and efficient.

Good comments make for a good result!

PR #3248, opened 2026-08-03
feat(pow): Empowering Mantle

danielSanchezQ:

Implements an integrates PoW rewards execution in the mantle ledger. There is a few granted values that are not yet defined in the spec which whould have to be updated later. For not rewards are disabled through the constans.

Spec is not finalized yet and I expect a few changes. But overall logic is there. Rewards pool, refilling, blocks slot tracking, op execution etc.

@madxor @youngjoon-lee @danielSanchezQ 

Spec is in sketch atm, but from calendar preassures we already started implementation.

Yes, spec uses block height but here we already use block slots. This was talked privately with @madxor and it will be adapted in the spec later.

## Checklist

> [!WARNING]  

* [x] 2. Description added.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


core/src/events/mod.rs
```diff
@@ -117,6 +118,11 @@ pub enum TxEventPayload {
         voucher_nullifier: VoucherNullifier,
         utxo: Utxo,
     },
+    /// A `PoW` claim operation created a reward not for its beneficiary
```
ntn-x2:
```suggestion
    /// A `PoW` claim operation created a reward note for its beneficiary
```

core/src/mantle/transactions/verification_helper.rs
```diff
@@ -51,6 +55,34 @@ pub trait OperationVerificationHelper {
         channel_id: &ChannelId,
         key_index: &ChannelKeyIndex,
     ) -> Result<Ed25519PublicKey, VerificationError>;
+
```
ntn-x2, line 57: Interesting trait..... 😃

ledger/src/mantle/pow/mod.rs
```diff
@@ -0,0 +1,658 @@
...
+
+impl PowState {
+    /// Create the genesis `PoW` state: pool and per-claim reward seeded from
+    /// the genesis endowment, initial difficulty derived from them, and no
+    /// claims or seen blocks yet.
+    #[must_use]
+    pub fn new() -> Self {
+        // TODO: Setup values when decided
+        Self {
+            reward_pool: POW_REWARD_POOL_GENESIS,
```
ntn-x2, line 68: Also, how are the rewards disabled for genesis is we are using values `> 0` here?

ledger/src/lib.rs
```diff
@@ -443,6 +472,15 @@ impl LedgerState {
         // Accumulate storage gas consumed so the storage market can update the
         // price at the next epoch rotation.
         self = self.add_storage_gas_consumed(total_block_storage_gas)?;
+        self.update_pow_difficulty(
```
ntn-x2: I had to ask AI for confirmation here. The `try_apply_contents` is called from a different context for block proposer (once per proposed tx) and validators (once with all txs in a block). I think the two results can diverge from each other, leading the proposer to produce blocks that no other node would be able to verify. Can you double check if this is actually true?

core/src/mantle/ops/pow.rs
```diff
@@ -74,23 +96,33 @@ pub enum ClaimPowRewardError {
...
     pub reward_difficulty: PowTarget,
+    /// Nullifiers of already-claimed `PoW` solutions.
     pub pow_nullifiers: &'a rpds::HashTrieSetSync<PowNullifier>,
     // needed not in spec yet
+    /// `sigma_e`: reward amount per claim for the current epoch.
     pub epoch_pow_reward: PowReward,
+    /// `R_PoW`: current balance of the `PoW` reward pool.
     pub epoch_reward_pool: PowReward,
+    /// Nonce of the current epoch.
     pub current_epoch_nonce: Epoch,
```
ntn-x2: A nonce is an `Epoch`? Is it not a verifiable random `Fr` instead?

ledger/src/mantle/pow/mod.rs
```diff
@@ -0,0 +1,658 @@
...
+    /// `sigma_e`: reward per claim, fixed for the epoch.
+    epoch_reward: PowReward,
+    /// `d_reward`: the REWARD threshold, retargeted every block
+    #[serde(with = "serde_fr")]
+    reward_difficulty: PowTarget,
+    /// Rewards collected during the current epoch, added to the
+    /// `reward_pool` at the next epoch boundary.
+    refill_rewards: PowReward,
+    /// Spent `PoW` solutions, retained only for the acceptance
+    nullifiers: HashTrieSetSync<PowNullifier>,
```
ntn-x2: When is this set cleared?

ntn-x2: Does it make sense to clear items as the verification window elapses?

ledger/src/lib.rs
```diff
@@ -333,12 +357,17 @@ impl LedgerState {
         )
         .checked_add(total_fee_tip)?;
 
+        let pow_reward = (reward_numerator * POW_REWARD_SHARE_NUMERATOR
```
ntn-x2: This is an unsafe arithmetic operation. We are truncating a `u128` into a (wrapped) `u64`.

core/src/mantle/ops/pow.rs
```diff
@@ -174,8 +206,32 @@ impl ClaimPoWRewardVerificationContext<'_> {
...
+    /// Hash of the transaction carrying this claim.
+    pub tx_hash: TxHash,
+    /// Unspent transaction outputs, extended with the reward note.
+    pub utxos: Utxos,
+}
+
+impl ClaimPoWRewardExecutionContext {
+    /// Deduct the paid-out `epoch_reward` from the `reward_pool`.
+    const fn decrement_reward_pool(&mut self) {
+        self.reward_pool = self.reward_pool.saturating_sub(self.epoch_reward);
```
ntn-x2: If we saturate it means we are paying out more than we could. We need to deal with that.

ledger/src/mantle/pow/mod.rs
```diff
@@ -0,0 +1,658 @@
...
+    const RATE_DEN: u64 = 100;
+    /// Expected number of reward claims per block.
+    const TARGET_CLAIM_PER_BLOCK: u64 = 0;
+    /// Expected number of blocks per epoch.
+    const EXPECTED_BLOCKS_PER_EPOCH: u64 = 0;
+
+    /// Full denominator of the per-epoch payout rate.
+    #[must_use]
+    fn denominator() -> u64 {
+        Self::RATE_DEN * Self::TARGET_CLAIM_PER_BLOCK * Self::EXPECTED_BLOCKS_PER_EPOCH
```
ntn-x2: This produces a `0` denominator, which is not possible. Also the return type should be at the very least a `NonZeroU64` instead.

core/src/mantle/ops/pow.rs
```diff
@@ -74,23 +96,33 @@ pub enum ClaimPowRewardError {
     InvalidPoWRewardTicket,
     #[error("Ticket was already claimed")]
     DoubleClaimed,
-    #[error("Out of window height ({height})")]
-    OutOfWindowHeight { height: u64 },
+    #[error("Out of window slot ({slot:?})")]
+    OutOfWindowSlot { slot: Slot },
```
ntn-x2: Should we also add the current slot, for easier debugging?

ledger/src/mantle/pow/mod.rs
```diff
@@ -0,0 +1,658 @@
...
+    /// window-of-acceptance check and pruned as they age out of
+    /// [`SLOT_WINDOW`]. Keyed by the wire-format block hash — the same
+    /// value a `ClaimPowRewardOp` anchors to — so consensus state stays
+    /// independent of the node's header-id type.
+    block_slots: HashTrieMapSync<Hash, Slot>,
+}
+
+/// Errors that can occur while applying `PoW` state transitions.
+#[derive(thiserror::Error, Debug, Clone, PartialEq, Eq)]
+pub enum Error {}
```
ntn-x2: Do we need this?

ledger/src/mantle/pow/mod.rs
```diff
@@ -0,0 +1,658 @@
...
+        for id in prunable {
+            self.block_slots.remove_mut(&id);
+        }
+    }
+
+    /// Apply an epoch transition: on epoch change, refill the reward pool
+    /// from the rewards collected during `previous_epoch`.
+    pub(crate) fn try_apply_header(
+        &self,
+        _config: (),
```
ntn-x2: Is this needed?

ntn-x2 (commented):
This PR feels still a bit raw, in my opinion. I highlighted some points that could be improved, but in particular maybe we should replace the constants with deployment parameters.

ledger/src/mantle/pow/mod.rs
danielSanchezQ: The check from the constants trait is disabled if `RATE_NUM` is zero. Independently of the genesis pools.

core/src/mantle/transactions/verification_helper.rs
danielSanchezQ: This is gonna be refactored to provide the contexts directly, cc @strinnityk

ledger/src/mantle/pow/mod.rs
danielSanchezQ: Yeah, changed it, thanks!

core/src/mantle/ops/pow.rs
danielSanchezQ: Yes, but this is computed in the check directly. The context gets the current and past epoch. We could pre-compute it, but then why bother changing the check function to only zkhash them if needed ? 😝

ledger/src/lib.rs
danielSanchezQ: Ill double check, this is a good catch. It depends if the states are being kept or discarded.

core/src/events/mod.rs
danielSanchezQ: Applied later, thanks!

core/src/mantle/ops/pow.rs
danielSanchezQ: Yeah, I dont remember seeing anything in the spec about it. cc @madxor

ledger/src/mantle/pow/mod.rs
danielSanchezQ: nah, leftover, removing

ledger/src/mantle/pow/mod.rs
danielSanchezQ: Leftover, removing

ledger/src/lib.rs
danielSanchezQ: @ntn-x2 really nice catch!!! I moved it from the header for the same reason, i didnt remember the contents where being used like that too. Its moved not to try_update which is called when updating from a block. And it also holds just for the new returned state. Thanks!!

danielSanchezQ:
@ntn-x2 fixes here: https://github.com/logos-blockchain/logos-blockchain/pull/3248/commits/af10be97816e936fa8d0106e531957ec25ed42ea
Thanks for the review!

youngjoon-lee (approved):
Thank you. I mostly focused on checking if the implementation matches with the RFC.

ledger/src/lib.rs
```diff
@@ -219,9 +237,24 @@ impl LedgerState {
...
+        Id: Into<BlockHash>,
     {
-        let (state, header_events) = self.try_apply_header(slot, proof, config)?;
-        let (state, tx_events) = state.try_apply_contents::<_, _, Constants>(config, txs)?;
+        let (mut state, header_events) = self.try_apply_header(slot, proof, config)?;
+        // Record the applied block among the recently seen blocks `PoW`
+        // claims may anchor to. This is the canonical apply path, where the
+        // block's id is known — unlike a proposer's direct
+        // `try_apply_header` call for a block still being built.
+        state.mantle_ledger.add_seen_block(block_id.into(), slot);
```
ntn-x2, line 247: I guess we would need to add the seen block after trying to apply its content, right?

core/src/mantle/ops/pow.rs
```diff
@@ -81,23 +103,33 @@ pub enum ClaimPowRewardError {
     InvalidPoWRewardTicket,
     #[error("Ticket was already claimed")]
     DoubleClaimed,
-    #[error("Out of window height ({height})")]
-    OutOfWindowHeight { height: u64 },
+    #[error("Out of window slot ({slot:?}) vs block slot ({current_slot:?})")]
```
ntn-x2, line 105: Should these be the other way around? Block slot is the anchor slot and window slot is the current slot? Or did I get this wrong?

danielSanchezQ: Slot is the pow referenced block slot, the current block slot is the current ledger state block slot. Its confusing tho.

core/src/mantle/ops/pow.rs
ntn-x2: Ok, then I guess we could rename these to be simply `current_epoch` and `past_epoch`? If we have nonce in there, but it's not a nonce, it's confusing :D

ledger/src/lib.rs
danielSanchezQ: Well we would have seen the block even if not accepted. The header was accepted already, should be guarantee that a leadership was correctly won, even if the contents where wrong. I think the behavior here doesnt change much whenever we place this call in this method. It just records the block slot so its available later. And trims older slot blocks.

core/src/mantle/ops/pow.rs
madxor: By the RULES OF (TOKEN)ECONOMICS we can only pay until we have the funds, then we simply don't pay. https://app.notion.com/p/nomos-tech/EmPoWering-the-Logos-Blockchain-Proposal-37a261aa09df8071b28efabcbfc9dca6?source=copy_link#37a261aa09df818482b3e8774bbf3aa3

core/src/mantle/transactions/verification_helper.rs
```diff
@@ -51,6 +55,34 @@ pub trait OperationVerificationHelper {
...
+    // `PoW` claim validation inputs, one per
+    // [`ClaimPoWRewardValidationContext`] field. The current epoch comes from
+    // [`Self::get_epoch`] and the current block slot from
+    // [`Self::get_block_slot`].
+    //
+    // [`ClaimPoWRewardValidationContext`]: crate::mantle::ops::pow::ClaimPoWRewardValidationContext
+
```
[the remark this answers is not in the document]
danielSanchezQ: Fixed

core/src/mantle/transactions/verification_helper.rs
```diff
@@ -51,6 +55,34 @@ pub trait OperationVerificationHelper {
...
+    /// `R_PoW`: current balance of the `PoW` reward pool.
+    fn get_pow_reward_pool(&self) -> PowReward;
+
+    /// The epoch preceding [`Self::get_epoch`], whose nonce is also accepted
+    /// for claims mined just before an epoch boundary.
+    fn get_previous_epoch(&self) -> Epoch;
+
+    /// Slots of the blocks a claim may anchor to, keyed by block hash;
+    /// used for the window-of-acceptance check.
+    fn get_blocks_slot(&self) -> HashMap<Hash, Slot>;
```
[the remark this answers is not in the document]
danielSanchezQ: Fixed

core/src/mantle/ops/pow.rs
ntn-x2: OK so if there is not enough funds we should fail to pay out, but the current logic pays out let's say `10`, and then removes at most 10 from the pool, but this seems to be wrong. If there is less than 10 in the pool, we should not pay out 10.

ntn-x2 (approved):
Looks good to me now! The only open point is about whether we should perform a saturating subtraction after paying out the reward, or whether we should perform a checked subtraction before, and then pay out the reward if enough funds are in the pool.

core/src/mantle/ops/pow.rs
danielSanchezQ: If its like it means that this could be failable hence building, or verification fails to include the Tx (?)

madxor: We should not pay out so we reject the transaction. It should be known a the moment of block building if there are any tokens left to pay out, so the block builder should avoid including such transaction.

strinnityk (approved):

PR #3251, opened 2026-08-04
feat: replace full tx hashes with prefixes in block proposals

ntn-x2:

Partial implementation of the [block compression spec](https://github.com/logos-co/logos-lips/pull/389). Block proposals now carry transaction hash prefixes instead of full prefixes, and each node tries to reconstruct the block by considering all candidate combinations, up to the maximum number of attempts as specified by the spec.

A follow-up PR will update the Blend side of things, including removing the gossipsub settings from the wire payload and make that a settings of the service itself, then we will have only encoded (and encapsulated) block proposals travelling over the wire.

Yes.

@madxor spec @ntn-x2 PR

It is being updated.

Yes, they are being taken care of.


core/src/mantle/transactions/hash.rs
```diff
@@ -3,13 +3,50 @@ use bytes::Bytes;
...
 serde_bytes_newtype!(TxHash, 32);
 
+/// Number of leading hash bytes a block proposal uses to refer to a
+/// transaction.
+const REFERENCE_PREFIX_BYTES: usize = 8;
+
+/// The leading [`REFERENCE_PREFIX_BYTES`] bytes of a [`TxHash`] — how a block
+/// proposal refers to a transaction that is already in a validator's mempool.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Hash, PartialOrd, Ord, BinaryCodec)]
+pub struct TxHashPrefix(pub [u8; REFERENCE_PREFIX_BYTES]);
```
danielSanchezQ, line 23: Nitpick, we could have use `Cow` so when calling `prefix` on `TxHash` we return a reference to the actual bytes instead?
Maybe its not worth it, but it makes sense that the `TxHashPrefix` is view into the actual thing. Let me know what you think.

services/chain/chain-network/src/lib.rs
```diff
@@ -987,37 +1005,110 @@ where
...
+    mempool: &MempoolAdapter<Item>,
+) -> Result<Vec<Vec<Item>>, Error>
+where
+    Item: MantleTxWithProofs<Hash = TxHash> + Clone + Send + Sync + 'static,
+{
+    let mut candidates = Vec::with_capacity(proposal.mempool_transactions().len());
+    let mut combinations = 1usize;
+
+    for (index, prefix) in proposal.mempool_transactions().iter().copied().enumerate() {
+        let found = candidates_for_reference(index, prefix, mempool).await?;
```
danielSanchezQ: Maybe we can unwrap this loop into futures, or we can actually create a batch method to the mempool?

services/tx-service/src/backend/pool.rs
```diff
@@ -150,6 +183,17 @@ where
...
+            if let Entry::Occupied(mut bucket) = self.by_prefix.entry(key.key_prefix()) {
+                // Bucket order is irrelevant — only reference order matters, and
+                // that comes from a block proposal — so a swap-remove is fine to avoid shifting
+                // remaining elements around.)
+                if let Some(position) = bucket.get().iter().position(|held| held == key) {
+                    bucket.get_mut().swap_remove(position);
+                }
+                if bucket.get().is_empty() {
+                    bucket.remove();
+                }
```
danielSanchezQ: This should probably be its own function.

danielSanchezQ (approved):
A few nitpicks and improvements I think we could do. Overall logic seems ok

core/src/mantle/transactions/hash.rs
ntn-x2: Yes I thought about it, did not spend too much time investigating it. We need the owned value to be stored in the mempool's index of prefixes, but I can give it a try to see if a Cow makes things any better without sacrificing simplicity.

services/chain/chain-network/src/lib.rs
ntn-x2: I preferred not to make this method a batch one since we would then need the mempool to be aware of the limits, which I think belong to this chain network layer. Letting the mempool return everything and enforce the limits later defeats the limits themselves. But, I like your suggestion to parallelize this work, so I will try to integrate that indeed.

core/src/mantle/transactions/hash.rs
danielSanchezQ: > Yes I thought about it, did not spend too much time investigating it. We need the owned value to be stored in the mempool's index of prefixes, but I can give it a try to see if a Cow makes things any better without sacrificing simplicity.

I think it should be simple. I can give it a shot, the prefix struct is pretty contained and short so it would just live there.

youngjoon-lee (approved):

core/src/mantle/transactions/hash.rs
ntn-x2: Actually the prefix is only 8 bytes, and it's `Copy`, so it's very cheap to copy. A `Cow` itself would already take 16 bytes instead because it stores the type + additional references for its semantic. Also, as I said before, we need an owned value basically all the time when creating a block proposal and in the mempool storage, so in this case a `CoW` does not really give us anything, but quite the opposite.  So I'll maintain the owned, 8-byte, `Copy` type instead, in this case.

services/tx-service/src/backend/pool.rs
ntn-x2: Fixed in https://github.com/logos-blockchain/logos-blockchain/pull/3251/commits/d67b98374e20a09584adaf3ecb6e0df125ad36f5.

services/chain/chain-network/src/lib.rs
ntn-x2: Actually parallelizing here also does not make sense, since we want to enforce an upper bound on the amount of combinations we try, so parallelizing that just defeats the purpose. We need to keep it sequential and check after every combination if we are within the limit.

PR #3254, opened 2026-08-04
fix(zone-sdk): fix flaky deposit zone test

pradovic:

There was a flaky tests that were using deposit. The fix is simple, deposit margin increased.


hansieodendaal (approved):
LGTM

PR #3256, opened 2026-08-04
fix(zone-sdk): update correct tip on bundle custom txs

pradovic:

There is a bug in updating the tip on custom tx bundle with multiple ops. The tip was correct after mined but not in local pending state.

There are 2 cases:
1. If custom tx does not carry inscriptions it does not matter, config updates would reset the tip and orphan non valid inscriptions
2. if custom tx does have an inscription it needs  to chain properly to preserve the inscription order correctly.

### DIFF
--- a/zone-sdk/src/sequencer/state.rs
+++ b/zone-sdk/src/sequencer/state.rs
@@ -43,15 +43,24 @@ struct PendingOtherTx {
     signed_tx: SignedMantleTx<Unverified>,
     first_parent: Option<MsgId>,
     last_msg: Option<MsgId>,
+    /// True when the tx bundles an inscription with its other ops (an atomic
+    /// bundle): it is part of the message chain, so publishes must chain
+    /// after it. A pure config cut carries no inscription and is not chained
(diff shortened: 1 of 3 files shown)


zone-sdk/src/sequencer/state.rs
hansieodendaal: This ordering guarantee no longer holds for the full return value.

`pending_other` entries are appended using `HashMap` iteration order below, but this PR now allows opaque custom bundles to form parent/child dependencies through their derived `last_msg`. A dependent bundle may therefore be returned before the bundle that establishes its parent tip.

hansieodendaal (commented):
Looking good, just one comment to consider please.

zone-sdk/src/sequencer/state.rs
pradovic: Good catch. It was self-healing, but much better to have it correctly. Fixed, thanks!

pradovic: I have also added a regression test for this

hansieodendaal (approved):
LGTM!

PR #3258, opened 2026-08-05
feat(blend): support compressed block proposals in blend

ntn-x2:

This is the last PR needed to support the draft block compression RFC.

In this PR, which builds on https://github.com/logos-blockchain/logos-blockchain/pull/3251, we do the following:

* Change the padding in Blend to the new maximum size, which matches the new size defined in the spec
* Change the encoding of what represents a "data message": we used to blend the block proposal with the gossipsub settings together, but that is wrong. So we take the settings out and pass them to the network adapter on creation, so now what is encapsulated is _just_ the (padded) block proposal, as it should be

Yes.

@madxor spec, @ntn-x2 impl

This PR provides the reference implementation to bring the spec up to date.

See previous section.


youngjoon-lee (approved):
Great! 

I agree that now we should take the gossipsub setting out of a blend message. At the very beginning, we considered blending various kinds of messages through the blend network, each of which would be gossiped on a different topic. But, now it's not the case anymore.

ntn-x2:
Will merge into the base feature branch, then merge everything as one PR so it's faster.

PR #3264, opened 2026-08-05
fix(zone-sdk): config update key index & round robin turn reset the tip

pradovic:

The config update was always using the [0] index. When this was fixed another bug was found, the channel tip was not reset on turn change. The fixes are trivial.

This PR fixes both, and adds an integration test:
1. seq A publishes configu
2. seq A and B publish inscriptions
3. seq b publishes new config
4. seq A and B publish inscriptions
5. assert everything is correct in the end


zone-sdk/src/sequencer/zone_sequencer.rs
```diff
@@ -795,18 +795,34 @@ where
...
+        // out of scope, so reject it early instead of submitting a
+        // transaction that can only die at block assembly.
+        let signer = match &self.channel_state {
+            None => None,
+            Some(channel) => {
+                if channel.configuration_threshold != 1 {
+                    return Err(Error::Network(format!(
+                        "channel config update needs {} signatures; only single-signer \
+                         (configuration_threshold 1) updates are supported",
+                        channel.configuration_threshold
```
hansieodendaal, line 814: I think the error msg can be improved here. Instead of simply saying “not supported,” it should mention that it is not supported by the one-shot `channel_config` helper, as multi-party configs are supported when using the prepare/sign/combine/submit workflow already provided by TUI-Zone. ("deployment/tui-zone/README.manual-demo.md:265")

hansieodendaal (commented):
Looking good, just one minor wording issue.

zone-sdk/src/sequencer/zone_sequencer.rs
pradovic: Fixed!

hansieodendaal (approved):
LGTM!

PR #3272, opened 2026-08-05
feat(mantle): enforce channel config lineage via parent message id

thomaslavaur:

This PR implements the Channel Config lineage <https://github.com/logos-co/logos-lips/pull/396>

Yes

@thomaslavaur 

Yes

No

## Checklist

> [!WARNING]  

* [x] 2. Description added.


youngjoon-lee (approved):
I focused on the `core` part. The `zone-sdk` needs to be reviewed by Petar.

zone-sdk/src/sequencer/block_fetch.rs
hansieodendaal, line 352: `ChannelConfigOp` now carries the authoritative lineage parent, but this helper derives the parent from the previous entry in the same transaction and defaults to `MsgId::root()`. I think this is incorrect when the config is the first channel operation in a transaction but follows an earlier block: it will report the wrong lineage parent and may corrupt downstream adoption or reorg handling.

Proposed to the operation's parent directly:

```rust
parent_msg: config.parent,
```

We could also add a regression test where a config with a non-root parent is the first channel entry in its transaction.

hansieodendaal (commented):
Looking good, one possible correctness issue.

zone-sdk/src/sequencer/block_fetch.rs
thomaslavaur: Thanks Hansie nice catch!

hansieodendaal (approved):
LGTM!

pradovic (changes requested):
@thomaslavaur, Nice work!

I think we might have 2 issues in zone-sdk, if I am not mistaken:

1. Config resubmission became turn-gated. Moving configs into the pending-inscription set makes them lineage links (good), but that set's resubmit path applies inscription rules: resubmit_pending skips entries while can_publish_inscription_now() is false and counts them against max_pending_publish_depth. The ledger imposes neither on configs — ChannelConfigOp::verify checks parent and threshold signatures, no round_robin. On a fixed-writer channel (posting_timeframe = 0), a non-leading accredited key that submits a config and loses the first post never resubmits it: its turn never comes. Suggestion: keep configs in pending_other, whose retries are ungated — the pending/pending_other split then coincides exactly with the ledger's turn boundary (turn-gated publishes vs turn-free tip-advancers). opaque_lineage already anchors them via config.parent in this PR, so shed_off_branch_pending_other sheds stale ones correctly. The missing piece is publish_parent chaining through pending_other entries (first_parent → last_msg) so in-flight configs still anchor the next publish — that walk already exists in the open pr/fix-custom-tx-valid-tip branch and the two changes compose (the config parent even removes that PR's special case for parentless config-led bundles). Unless this needs to be turn-gated, then we need a spec/mantle update and not zone-sdk update.

2. Orphaned configs are misreported. A stale pending config is now correctly shed — but it surfaces as ChannelUpdateTx::Inscription with an empty payload. Consumers can't distinguish "my config was cut" from "an empty inscription was cut", and a republish policy would republish an empty inscription as recovery — the wrong operation. Reporting it as Custom would be equally misleading (it isn't a foreign shape), and dropping it silently loses the recovery signal. Suggestion: add a Config variant to PendingTx/ChannelUpdateTx carrying the ChannelConfigOp (and tx hash). That makes recovery possible: the op holds everything needed to re-issue via channel_config against the new tip. The same variant on FinalizedOp would fix the parallel asymmetry on the finalized stream (configs currently appear there as synthetic empty-payload inscriptions) — fine as a follow-up if preferred.

Doc note, not a gap: this PR flips config retry semantics — previously a config was valid at any tip, so byte-identical resubmission was always correct; now a stale config is permanently invalid, and recovery means re-issuing (new signatures, new id). Automatic for configuration_threshold == 1 (the sequencer re-signs alone); above that it restarts signature collection, which is the spec's acknowledged cost. Worth stating in the sequencer docs so integrators don't rely on the old reland-anywhere behavior.

hansieodendaal:
> I think we might have 2 issues in zone-sdk, if I am not mistaken:
> 

Good catch! Agreed. I think we could:
- Make config retry behavior independent of inscription round-robin/depth controls while preserving lineage-aware ordering.
- Introduce a first-class config representation in pending/orphan update events, so a consumer can distinguish config recovery from a blank inscription.
- Add tests for:
  - retrying a config while the submitting key is not allowed to inscribe;
  - orphaning a config and verifying the emitted update is typed as configuration, not an empty inscription;
  - config recovery expectations for threshold-1 versus multi-signature configurations.

pradovic:
I think we should also reconsider (sorry if it has already be done, then just dismiss the comment) having a inscription as parent of the config. Multi sig configs might be slower to compile then regular inscriptions, so you would essentially be racing with last known inscriptions and orphaning would give you something slow to republish when the tip is updated, as it would update the inscription tip thus losing the atomicity with inscription anyway. Separate config lineage might help here.  As zones might be producing inscriptions fast (although turn based, so it depends which sequnecer is posting the config).

It depends do we need to point to exact time (exact inscription) for the particular config, or is it ok for the config update to just cut in the middle at any time (but with valid config parent) and orphan invalidate/orphan rest of the inscriptions.

hansieodendaal:
> I think we should also reconsider ...

I think the real choice is:

- _Need config to be ordered at an exact message boundary?_
  Use a shared lineage (as is proposed in this PR) plus a locking/halt/reservation protocol to allow time to change the config.
- _Need configurations to remain live while the channel is actively posting?_
  Use a separate config lineage, active in the next block after it is mined.
- _Need every state change to be jointly authorized?_
  Threshold-sign inscriptions too, >= config threshold.

thomaslavaur:
hummm yes we just want to avoid configuration replayability... we don't even need ordering (that's why I initially proposed just a nonce incrementatal).

Maybe we just need the configuration to have it's own lineage

zone-sdk/src/sequencer/actor.rs
```diff
@@ -493,6 +493,31 @@ where
...
+        let seen: HashSet<_> = channel_update
+            .orphaned
+            .iter()
+            .map(ChannelUpdateTx::tx_hash)
+            .collect();
+        for tx in stale_configs {
+            if !seen.contains(&tx.mantle_tx().hash()) {
+                channel_update.orphaned.push(ChannelUpdateTx::Custom(tx));
+            }
+        }
```
hansieodendaal, line 524: I think this needs a typed config recovery path rather than surfacing the shed transaction as `ChannelUpdateTx::Custom`.

A pending `ChannelConfigOp` becomes permanently unmineable when another config advances `config_tip_hash`: it cannot be retried byte-for-byte, and for a multisig config recovery means rebuilding the operation against the new config parent and collecting signatures again. A consumer receiving `Custom(tx)` cannot reliably distinguish that case from an arbitrary transaction shape it owns itself.

Could we add a config-specific representation? For example:

- `PendingTx::Config(ChannelConfigOp / ConfigInfo)`
- `ChannelUpdateTx::Config(...)`

Consumers need the ability to restart the config workflow instead of treating this as an opaque orphan.

The same representation could also replace the synthetic empty `PendingTx::Inscription` returned by `do_channel_config`, so configs are consistently represented as configs throughout the SDK.

hansieodendaal (commented):
Looking good, just one possible improvement.

zone-sdk/src/sequencer/state.rs
```diff
@@ -676,6 +674,67 @@ impl TxState {
             .collect()
     }
 
+    /// Shed pending config-carrying txs whose config parent can no longer
+    /// reach the mined config tip: removed from retry and returned whole for
+    /// orphan reporting.
+    pub fn shed_stale_pending_configs(
```
youngjoon-lee: It seems that this function can falsely orphan a config tx that already landed on the chain.

Scenario:
Our pending config `C` (parent = `B`) is included in the chain, and a successor config `D` (parent = `C`) is also included right after.
But, the zone-sdk didn't process those blocks yet. 
In this case, this function may be called with `config_tip = D` which is returned from the node, and the `landable` is seeded with `D`.
So, this function may think that the config `C` is not landable because its parent `B` is not in the `landable`. 
Also, `C` is not in the `safe` set because zone-sdk didn't process its block yet.
As a result, `C` is shed and reported as `Orphaned`. After for a while, it will be reported as `OnChain` once its block arrives, but it could be confusing.

Let me think how to fix this...

youngjoon-lee (commented):
`lb-core` looks correct. I'm taking a look at `zone-sdk` but it's taking me some time since I'm not very familiar with the codebase.

zone-sdk/src/sequencer/state.rs
youngjoon-lee: Suggestion:

We might be able to use a similar strategy that zone-sdk is already doing for tracking the message tip per block. 
https://github.com/logos-blockchain/logos-blockchain/blob/fc0b5a59466c384f33abc43483acb84716742e90/zone-sdk/src/sequencer/state.rs#L888-L894
Instead of fetching `config_tip` from a node, zone-sdk may be able to track the config tip per block. Then, this `shed_...` function can make a judgement based on only information that is obtained when zone-sdk processed blocks. 

I might be wrong since I'm not used to the codebase. It'd be helpful if @pradovic checks this.

pradovic:
I don't think this is a complete implementation for zone-sdk. I will work on implementing the zone-sdk part in the same PR, or open a separate PR for review that we can merge into this one.

zone-sdk/src/sequencer/state.rs
pradovic: Yes, I think you are correct and we should be able to use the similar strategy as well. I synced with @thomaslavaur and I will take over the zone-sdk part, it will be easier.

hansieodendaal: _"But, the zone-sdk didn't process those blocks yet."_

I believe this a general liveness issue; blocks need to be processed as soon as possible otherwise wasteful work will be done. Applicable to any zone operation.

pradovic:
@thomaslavaur, @hansieodendaal , @youngjoon-lee I have opened a separate PR for zone-sdk changes: https://github.com/logos-blockchain/logos-blockchain/pull/3408. Let's review it separately, and if/when it is good we can merge it to this one before merging it all to master. Please see the PR description as it explains the reasoning and let me know does it makes sense.

zone-sdk/src/sequencer/state.rs
```diff
@@ -630,6 +675,34 @@ impl TxState {
...
     }
 
+    /// On a config-tip change, shed the pending entries **not on this branch's
+    /// tip** (not in the safe set) — the not-yet-mined tail a config may have
+    /// invalidated. Mined/on-branch entries are excluded: a config never
+    /// invalidates a landed inscription, so orphaning one would re-post an
+    /// on-chain original as a duplicate. The caller resets the chaining pointer
+    /// to the message tip so re-posts land as a competing branch. Unchanged
+    /// config tip → empty.
+    pub fn shed_pending_inscriptions_on_config(&mut self, tip: HeaderId) -> Vec<PendingTx> {
```
hansieodendaal, line 700: `shed_pending_inscriptions_on_config` removes all pending inscriptions whenever `config_tip_hash` changes, regardless of whether their parent is still the current `tip_message`, whether the signer remains accredited, or whether the configuration actually affects their validity. Since configs now explicitly do not advance the message lineage, a config that only changes e.g. `transfer_threshold` or keeps the same posting key should not invalidate queued inscriptions. 

We should only shed pending inscriptions when the new channel view makes them invalid, or leave valid pending message-lineage entries intact and let normal turn/retry rules handle them.

zone-sdk/src/sequencer/actor.rs
hansieodendaal: The prior request for a typed config representation remains applicable in this version of the PR.

hansieodendaal (commented):
Nice updates. Please see comments for consideration.

zone-sdk/src/sequencer/state.rs
pradovic: This is explained in my PR description. It was agreed with David to first open a PR with non optimized version treating the config as possible invalidation. The tip is correct for republish after orphaning so at worst this creates a same lineage alternate chain and would collaps eventually. 
The improved criterion, for the different separate PR is to find a first invalid inscription based on th accredited keys and orphan it eith its children. However some of the children might still be valid based on keys so it collapses to the same situation. So I would argue this current approach should be correct. I have also added a test scenario to test it. Does that makes sense?

hansieodendaal: I do not think they should be removed outright; the node will reject what became invalid. In my mind optimization would be to proactively identify what would be rejected without waiting for the node to act.

pradovic: The node would reject but we don't have this info from the node, imo. I mean it would not apply them but there is no rejecting event. I think we have two options here:
1. Identify first invalid based on accredited keys and its children
2. Orphan everything right away but treat adoption as source od truth as we do
as the first one also has the problem same as second one, but it is self healing. The user will just sometimes republish some inscriptions again, but at the correct tip, creating alternate same chain in mempool, and only one (or merged combination) will land, and adoption would self correct it.

We cannot know for sure what is invalidated, and it could be in the chain a-b-c in the mempool, that only b is invalid, so orphaning b and c will still create a competing c inscription that will compete with the old one. So it collapses to the same problem, imho

So I am leaving optimization for the different PR but it will have the same handling, as optimization is not perfect either as we always orphan/adopt linear chain we cannot skip some.

pradovic: To be precise, we only orphan the not-yet-mined entries — anything already mined on the branch is left alone. So a re-post can at most produce a same-lineage alternate that the ledger dedups; it can never create a second copy of something already on chain. That's why "orphan and let adoption correct it" is safe, not lossy.

hansieodendaal: I suppose it is not too bad, whatever is in the mempool will either be mined or rejected, irrespective of what the sequencer does. In the end the sequencers will be aligned with what is on the blockchain.

zone-sdk/src/sequencer/state.rs
```diff
@@ -1728,256 +1843,329 @@ mod tests {
...
+        assert!(!state.pending_other_contains(&stale_hash));
+    }
 
-        let old_lineage = state.channel_lineage(b1);
+    /// The finalized config tip must survive a warm restart. It lives only in
+    /// `finalized_config` once its block prunes below LIB, so a checkpoint has
+    /// to carry it — otherwise `config_tip_at` falls back to `root` after
+    /// restart and a later config would chain on a stale parent.
+    #[test]
+    fn finalized_config_survives_checkpoint_restore() {
```
hansieodendaal, line 1931: Nice, this covers the underlying `TxState` fallback nicely.

Could we also add one checkpoint-level regression that goes through `build_checkpoint` → serialize/deserialize → `ZoneSequencer::init_with_config` (and ideally a resumed backfill batch)? That would verify the new field is wired through the real persistence/resume path, rather than only proving that `set_finalized_config` itself restores the expected value.

zone-sdk/src/sequencer/state.rs
```diff
@@ -1728,256 +1843,329 @@ mod tests {
...
+        let tx = SignedMantleTx::new(
+            RawMantleTx([Op::ChannelConfig(config)].into()),
+            OpsProofs::empty(),
+        );
+        (tx, config_msg)
+    }
+
+    /// Wrap a config tx as it is classified when mined: a `Custom` block tx
+    /// carrying the config on the config lineage (no message-tip entry).
+    fn config_block_tx(
```
hansieodendaal: Could we use the clean `BlockChannelTx::Config` path here? A single-config transaction is classified that way in production, whereas `BlockChannelTx::Custom { config_entries: ... }` is the mixed/unknown-shape path.

The mixed path is already covered separately in `mixed_config_tx_is_custom_but_advances_the_config_tip`, so making this helper return `BlockChannelTx::Config` would let the stale-config tests exercise both production representations in their respective tests.

hansieodendaal (approved):
LGTM!

**Non-blocking:** Two test improvement suggestions below.

pradovic (approved):
LGTM!

youngjoon-lee (approved):

PR #3275, opened 2026-08-06
feat: new type for PoQ circuits bit-sized numbers

ntn-x2:

Previously, we were using `u64` for values that would then be part of PoQ generation. But the circuit expects those values to be at most 20 bit long. We would find out way too late that those values would be unusable. So we introduce a new type that makes that error a compile time error, that can be checked on and acted early on rather than when it actually happens. This means most of the zk crate operations are now error-free since the input is already assumed to be of the expected maximum bit length.

I updated the Blend stack to use the new value for core quota, leadership quota, and key index, which all refer to the same underlying type with different aliases.

The new type is defined in `zk/groth16/src/circuit_integer.rs`. All the other changes are cascading effects after changing the expected type for core/leadership quota and key index from `u64` to the new type.

Yes.

@ntn-x2 

N/A

No.


zk/groth16/src/circuit_integer.rs
```diff
@@ -0,0 +1,237 @@
...
+
+use crate::{Fr, Groth16Input};
+
+/// An unsigned integer constrained to `0..=2^BITS - 1`.
+///
+/// The bound is an invariant of the type: every value of
+/// `CircuitInteger<BITS>` fits in `BITS` bits, so conversions into circuit
+/// inputs are infallible. Constants are checked when they are compiled
+/// ([`Self::new_const`]); only values that are genuinely unknown until runtime
+/// need [`Self::new`].
```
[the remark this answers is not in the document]
ntn-x2: Fixed in https://github.com/logos-blockchain/logos-blockchain/pull/3275/commits/6dc53144c85447e3bb3bf211703bce09f0689e58.

zk/groth16/src/circuit_integer.rs
```diff
@@ -0,0 +1,237 @@
...
+///
+/// Circuits are expected to alias this type at the width of the signal they
+/// feed, so that widening or narrowing a signal is a one-line change that
+/// surfaces every invalidated constant as a build failure.
+#[derive(Copy, Clone, Default, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+#[serde(try_from = "u64", into = "u64")]
+pub struct CircuitInteger<const BITS: u8>(u64);
```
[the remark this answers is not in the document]
ntn-x2: Fixed in https://github.com/logos-blockchain/logos-blockchain/pull/3275/commits/6dc53144c85447e3bb3bf211703bce09f0689e58.

blend/scheduling/src/message_blend/provers/core_and_leader/mod.rs
```diff
@@ -125,7 +125,7 @@ where
         tracing::trace!(
             target: LOG_TARGET,
             epoch = ?self.core_proofs_generator.settings.epoch,
-            quota = self.core_proofs_generator.settings.public_inputs.core.quota,
+            quota = %self.core_proofs_generator.settings.public_inputs.core.quota,
```
thomaslavaur, line 130: just a curiosity but what does % do? First time I see it in rust in front of a variable

services/blend/src/core/mod.rs
```diff
@@ -1707,7 +1708,7 @@ where
                 tracing::warn!(target: LOG_TARGET, "Recovered data message should be present in the recovery state but was not found.");
             }
             // Each data message that is sent is one less cover message that should be generated, hence we consume one core quota per data message here.
-            state_updater.consume_core_quota(1);
+            state_updater.consume_core_quota(Quota::new::<1>());
```
thomaslavaur: nit but would be valuable for the tests readability to only use Quota::ZERO and Quota::ONE

thomaslavaur (approved):

blend/scheduling/src/message_blend/provers/core_and_leader/mod.rs
ntn-x2: It uses the `Display` implementation instead of the `Debug` one. In this case it does not matter, so maybe I can just revert that change.

services/blend/src/core/mod.rs
ntn-x2: Good point.

blend/scheduling/src/message_blend/provers/core_and_leader/mod.rs
ntn-x2: I ok I can't remove it because the type itself does not implement `tracing::Value`. I think it's fine to then delegate to its `Display` implementation in this case.

services/blend/src/core/mod.rs
ntn-x2: Addressed in https://github.com/logos-blockchain/logos-blockchain/pull/3275/commits/fc3e1d549688d7f2582b93083c003de944e54fd7.

blend/message/src/crypto/proofs.rs
```diff
@@ -29,18 +29,19 @@ pub struct PoQVerificationInputsMinusSigningKey {
...
             core: CoreInputs {
                 zk_root: ZkHash::default(),
-                quota: 1,
+                quota: Quota::ONE,
             },
             leader: LeaderInputs {
                 pol_ledger_aged: ZkHash::default(),
                 pol_epoch_nonce: ZkHash::default(),
-                message_quota: 1,
+                message_quota: Quota::new::<1>(),
```
thomaslavaur: Quota::ONE?

blend/message/src/reward/epoch.rs
```diff
@@ -191,19 +193,19 @@ mod tests {
...
     #[test]
     fn test_token_count_bit_len() {
-        let core_quota = 5;
+        let core_quota: Quota = Quota::new::<5>();
         let num_core_nodes = 2;
         // ceil(log2(10 + 1))
         assert_eq!(token_count_bit_len(core_quota, num_core_nodes).unwrap(), 4);
 
-        let core_quota = 0;
+        let core_quota: Quota = Quota::new::<0>();
```
thomaslavaur: Quota::ZERO

thomaslavaur (approved):

blend/message/src/crypto/proofs.rs
ntn-x2: Ops!

blend/message/src/reward/epoch.rs
ntn-x2: Ops!

youngjoon-lee (approved):

PR #3276, opened 2026-08-06
feat(utils): add with_capacity constructor to BoundedVec

### LINKED ISSUE #3261: Add `with_capacity` method for a bounded vec
Maybe would be good to have a `with_capacity` method for a bounded vec as well, that checks whether the capacity is within the `MIN..MAX` bounds?

_Originally posted by @ntn-x2 in https://github.com/logos-blockchain/logos-blockchain/pull/3228#discussion_r3718503744_

ygd58:
Fixes #3261.

Adds two new constructors to `BoundedVec`:

### `with_capacity(capacity: usize) -> Result<Self, BoundedError>`
Pre-allocates storage at runtime while enforcing `[MIN, MAX]` bounds.
Returns `BoundedError::CapacityOutOfBounds { min, max, capacity }` when the requested capacity is outside the allowed range.

### `with_const_capacity<const CAPACITY: usize>() -> Self`
Pre-allocates storage with bounds enforced at **compile time** via `const { assert!(...) }`.
Panics at compile time when `CAPACITY > MAX` or `MIN > 0`.

### Also adds
- `BoundedError::CapacityOutOfBounds { min, max, capacity }` variant
- 5 unit tests covering: within bounds, at max, above max, below min, const capacity

### DIFF
--- a/utils/src/bounded/vec.rs
+++ b/utils/src/bounded/vec.rs
@@ -37,14 +37,19 @@ impl<T, const MIN: usize, const MAX: usize> Bounded<Vec<T>, MIN, MAX> {
     /// Constructs an empty vector with at least the specified capacity.
     ///
     /// The `capacity` must be within the `[MIN, MAX]` bounds; returns
-    /// [`BoundedError::CapacityOutOfBounds`] when `capacity` is outside this range.
+    /// [`BoundedError::CapacityOutOfBounds`] when `capacity` is outside this
+    /// range.
     ///
     /// This does **not** change the length of the vector (it is still zero
     /// after construction), but pre-allocates space so that at least
     /// `capacity` elements can be pushed without reallocation.
     pub fn with_capacity(capacity: usize) -> Result<Self, BoundedError> {
         if MIN > 0 || capacity > MAX {
-            return Err(BoundedError::CapacityOutOfBounds { min: MIN, max: MAX, capacity });
+            return Err(BoundedError::CapacityOutOfBounds {
+                min: MIN,
+                max: MAX,
+                capacity,
+            });
         }
         Ok(Self::new_unchecked(Vec::with_capacity(capacity)))
     }
@@ -57,14 +62,8 @@ impl<T, const MIN: usize, const MAX: usize> Bounded<Vec<T>, MIN, MAX> {
     #[must_use]
     pub fn with_const_capacity<const CAPACITY: usize>() -> Self {
         const {
-            assert!(
-                MIN == 0,
-                "Cannot construct empty BoundedVec when MIN > 0"
-            );
-            assert!(
-                CAPACITY <= MAX,
-                "Requested capacity exceeds BoundedVec MAX"
-            );
+            assert!(MIN == 0, "Cannot construct empty BoundedVec when MIN > 0");
+            assert!(CAPACITY <= MAX, "Requested capacity exceeds BoundedVec MAX");
         }
         Self::new_unchecked(Vec::with_capacity(CAPACITY))
     }
@@ -756,7 +755,6 @@ mod tests {
         assert!(v.as_inner().capacity() >= 5);
     }
 
-
     #[test]
     fn with_capacity_within_bounds_succeeds() {
         type V = BoundedVec<u32, 0, 10>;
@@ -778,7 +776,11 @@ mod tests {
         type V = BoundedVec<u32, 0, 4>;
         assert!(matches!(
             V::with_capacity(5),
-            Err(BoundedError::CapacityOutOfBounds { min: 0, max: 4, capacity: 5 })
+            Err(BoundedError::CapacityOutOfBounds {
+                min: 0,
+                max: 4,
+                capacity: 5
+            })
         ));
     }
 
@@ -787,8 +789,11 @@ mod tests {
         type V = BoundedVec<u32, 2, 10>;
         assert!(matches!(
             V::with_capacity(5),
-            Err(BoundedError::CapacityOutOfBounds { min: 2, max: 10, capacity: 5 })
+            Err(BoundedError::CapacityOutOfBounds {
+                min: 2,
+                max: 10,
+                capacity: 5
+            })
         ));
     }
-
 }
--- a/utils/src/bounded/vec.rs
+++ b/utils/src/bounded/vec.rs
@@ -44,11 +44,8 @@ impl<T, const MIN: usize, const MAX: usize> Bounded<Vec<T>, MIN, MAX> {
     /// after construction), but pre-allocates space so that at least
     /// `capacity` elements can be pushed without reallocation.
     pub fn with_capacity(capacity: usize) -> Result<Self, BoundedError> {
-        if capacity < MIN {
-            return Err(BoundedError::TooFewItems { count: capacity, min: MIN });
-        }
-        if capacity > MAX {
-            return Err(BoundedError::TooManyItems { count: capacity, max: MAX });
+        if capacity < MIN || capacity > MAX {
+            return Err(BoundedError::CapacityOutOfBounds { min: MIN, max: MAX, capacity });
         }
         Ok(Self::new_unchecked(Vec::with_capacity(capacity)))
     }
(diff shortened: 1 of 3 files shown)


utils/src/bounded/vec.rs
```diff
@@ -34,6 +34,25 @@ impl<T, const MIN: usize, const MAX: usize> Bounded<Vec<T>, MIN, MAX> {
...
+    /// The `capacity` must be within the `[MIN, MAX]` bounds; returns
+    /// [`BoundedError::TooFewItems`] when `capacity < MIN` and
+    /// [`BoundedError::TooManyItems`] when `capacity > MAX`.
+    ///
+    /// This does **not** change the length of the vector (it is still zero
+    /// after construction), but pre-allocates space so that at least
+    /// `capacity` elements can be pushed without reallocation.
+    pub fn with_capacity(capacity: usize) -> Result<Self, BoundedError> {
+        if capacity < MIN {
+            return Err(BoundedError::TooFewItems { count: capacity, min: MIN });
```
ntn-x2: We can use a different error variant for this, there's nothing wrong about it.

ntn-x2 (commented):
Thanks for this contribution! I think this PR is almost ready to go. I would ask you to also add a `with_const_capacity<const usize: CAPACITY>` that we can then run a compile-time assertion against with

```rust
const {
            assert!(
                MIN == 0,
                "Cannot construct empty BoundedVec when MIN > 0"
            );
            assert!(
                CAPACITY <= MAX,
                "Requested capacity exceeds BoundedVec MAX"
            );
        }
```

ygd58:
Added `with_const_capacity<const CAPACITY: usize>()` with compile-time assertions for both `CAPACITY <= MAX` and `MIN == 0`. Also added a unit test for it. Note: `Vec::with_capacity` is not yet stable as a const fn, so the method itself is not `const` — but the bounds checks run at compile time via `const { assert!(...) }`. @ntn-x2

ntn-x2 (approved):
✅

ntn-x2 (changes requested):
Actually can you add a new error variant for the new method?

ygd58:
Thanks for the approval and feedback! Could you clarify which method you'd like a new error variant for — `with_capacity` (the runtime version) or `with_const_capacity` (which currently uses compile-time panics)? And what would the new variant represent — a specific 'capacity out of bounds' error separate from the existing TooFewItems/TooManyItems? @ntn-x2

ntn-x2:
@ygd58 the compile-time version does not need a new variant since it panics at compile time. The runtime version needs one. A simply `CapacityOutOfBounds` variant with min, max, and provided value would do it.

ygd58:
Added `BoundedError::CapacityOutOfBounds { min, max, capacity }` and updated `with_capacity` to use it. Tests updated accordingly. @ntn-x2

ntn-x2 (approved):

utils/src/bounded/vec.rs
```diff
@@ -34,6 +34,42 @@ impl<T, const MIN: usize, const MAX: usize> Bounded<Vec<T>, MIN, MAX> {
         Self::new_unchecked(Vec::new())
     }
 
+    /// Constructs an empty vector with at least the specified capacity.
+    ///
+    /// The `capacity` must be within the `[MIN, MAX]` bounds; returns
+    /// [`BoundedError::TooFewItems`] when `capacity < MIN` and
+    /// [`BoundedError::TooManyItems`] when `capacity > MAX`.
```
strinnityk: This comment is out of date.

ntn-x2:
@ygd58 the PR is in the queue. Could you please just update the PR description with the latest changes? Thank you.

strinnityk (approved):

ygd58:
Updated PR description and fixed the doc comment — now references `BoundedError::CapacityOutOfBounds` instead of the old variants. @strinnityk @ntn-x2

utils/src/bounded/vec.rs
```diff
@@ -397,7 +432,7 @@ mod tests {
         let mut bv = TestBoundedVectorMin2::try_from(vec![1, 2, 3, 4]).unwrap();
         assert_eq!(
             bv.try_push(5),
-            Err(BoundedError::TooManyItems { count: 5, max: 4 })
+            Err(BoundedError::CapacityOutOfBounds { min: 0, max: 4, capacity: 5 })
```
hansieodendaal: `try_push` must return `BoundedError::TooManyItems`

utils/src/bounded/vec.rs
```diff
@@ -432,11 +467,11 @@ mod tests {
     fn try_from_rejects_input_above_max() {
         assert_eq!(
             TestBoundedVectorMin2::try_from(vec![1, 2, 3, 4, 5]),
-            Err(BoundedError::TooManyItems { count: 5, max: 4 })
+            Err(BoundedError::CapacityOutOfBounds { min: 0, max: 4, capacity: 5 })
```
hansieodendaal: `try_from` must return `BoundedError::TooManyItems`

utils/src/bounded/vec.rs
```diff
@@ -432,11 +467,11 @@ mod tests {
...
         );
         assert_eq!(
             TestBoundedVectorMin0::try_from(vec![1, 2, 3, 4, 5]),
-            Err(BoundedError::TooManyItems { count: 5, max: 4 })
+            Err(BoundedError::CapacityOutOfBounds { min: 0, max: 4, capacity: 5 })
```
hansieodendaal: ditto `try_from`

utils/src/bounded/vec.rs
```diff
@@ -623,7 +658,7 @@ mod tests {
     fn try_from_iter_rejects_items_above_maximum() {
         let result = TestBoundedVectorMin0::try_from_iter([1, 2, 3, 4, 5]);
 
-        assert_eq!(result, Err(BoundedError::TooManyItems { count: 5, max: 4 }));
+        assert_eq!(result, Err(BoundedError::CapacityOutOfBounds { min: 0, max: 4, capacity: 5 }));
```
hansieodendaal: `try_from_iter` must return `BoundedError::TooManyItems`

utils/src/bounded/vec.rs
```diff
@@ -34,6 +34,41 @@ impl<T, const MIN: usize, const MAX: usize> Bounded<Vec<T>, MIN, MAX> {
...
+    ///
+    /// This does **not** change the length of the vector (it is still zero
+    /// after construction), but pre-allocates space so that at least
+    /// `capacity` elements can be pushed without reallocation.
+    pub fn with_capacity(capacity: usize) -> Result<Self, BoundedError> {
+        if capacity < MIN || capacity > MAX {
+            return Err(BoundedError::CapacityOutOfBounds { min: MIN, max: MAX, capacity });
+        }
+        Ok(Self::new_unchecked(Vec::with_capacity(capacity)))
+    }
```
hansieodendaal, line 55: `capacity` is allocation metadata, not the vector’s length. This constructor always returns an empty `Vec`, so `BoundedVec<T, 2, 10>::with_capacity(5)` currently succeeds with `len() == 0`, despite the type requiring at least two elements.

Could we make this consistent with `empty()` and `with_const_capacity()` by rejecting construction when `MIN > 0`? A capacity reservation alone cannot satisfy the lower length bound.

After this change, I would expect both these tests to pass:
```rust
type Valid = BoundedVec<u32, 0, 10>;
let valid = Valid::with_capacity(5);
assert!(valid.is_ok());
assert_eq!(valid.unwrap().len(), 0);

type Invalid = BoundedVec<u32, 2, 10>;
let invalid = Invalid::with_capacity(5);
assert!(invalid.is_err());
```

hansieodendaal (changes requested):
Hi there, nice addition, but I think we are breaking the boundedness contract as is.

ygd58:
Fixed all three issues:
- `with_capacity` now rejects `MIN > 0` with `CapacityOutOfBounds` (capacity alone cannot satisfy the lower length bound, consistent with `empty()` and `with_const_capacity()`)
- Restored `TooManyItems`/`TooFewItems` for `try_push`, `try_from`, `try_from_iter` tests
- Updated `with_capacity` tests to use `BoundedVec<u32, 0, N>` for valid cases, added `with_capacity_rejects_nonzero_min` test
- All 74 utils tests pass @hansieodendaal

hansieodendaal (commented):
LGTM!

ntn-x2:
@hansieodendaal can you approve, so we can add it to the merge queue? Thanks!

hansieodendaal (approved):
LGTM

ntn-x2:
@ygd58 would you mind fixing the CI errors once again?

ygd58:
Applied `cargo +nightly fmt` to fix formatting. CI should pass now. @ntn-x2

ntn-x2:
@ygd58 there are other issues beside formatting.

ygd58:
Fixed the non-exhaustive pattern errors — added `CapacityOutOfBounds` arms to the two match expressions in `genesis.rs`. Both are unreachable in their call context but required for exhaustiveness. @ntn-x2

PR #3277, opened 2026-08-06
fix(zone-sdk): increase funding in flaky integration tests

pradovic:

Zone tests with concurrent republish policies can get flaky on slow CI runs: when many publishes orphan at once, the burst of re-funding requests can momentarily exhaust the funding wallet's free notes (the rest are reserved by in-flight transactions until they land or expire), and a republish that fails to fund is dropped for good. This PR bumps the funding top-up from 100 to 150 notes in the three concurrent scenarios as a quick mitigation, so the burst never touches the whole note set


hansieodendaal (approved):
LGTM

PR #3281, opened 2026-08-07
feat(core): uncle-ref: carry uncle headers, committed by `body_root`

youngjoon-lee:

This is the 1st PR for the [Uncle Reference RFC](https://github.com/logos-co/logos-lips/pull/385). More PRs will be stacked on this. 

In summary, the RFC introduces uncle references to improve the accuracy of TSI (total stake inference). The previous TSI counts only blocks in the honest chain, but doesn’t count blocks in forks. As a result, the stake has been 15% underestimated. So, the RFC proposes to count the uncles (in the forks) as well. 

This PR updates the `Proposal` and `Block` structures to carry `uncle_headers`, which will be filled/verified in the next PRs. 
- `Proposal` and `Block` carry `uncle_headers`. 
- `block_root` is replaced with `body_root`, which commits to both the carried `uncle_headers` and the transaction Merkle root. This name change is also specified in the RFC.
- `MAX_PAYLOAD_BODY_SIZE` becomes 10000 bytes; blend codec fixtures regenerated.

Yes. Please read the RFC I linked above. 

`Max_Body_Length` in the RFC (34574) is outdated. It should be updated according to the recent changes: block compression and tx hash prefix. 

No


core/src/block/uncle.rs
```diff
@@ -0,0 +1,63 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::header::Header;
+
+/// Maximum number of uncles a block may reference.
+pub const MAX_UNCLES: usize = 4;
+
+/// Signed headers of the uncles referenced by a block.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize, BinaryCodec)]
+pub struct UncleHeaders(UpperBoundedVec<SignedHeader, MAX_UNCLES>);
```
ntn-x2, line 13: Do we need a newtype for this? Can't this just be a type alias instead?

core/src/block/fixtures/mod.rs
```diff
@@ -20,16 +20,24 @@ fn three_references() -> BlockTransactionReferences {
...
+fn uncle(byte: u8) -> SignedHeader {
+    SignedHeader::new(header(byte), Ed25519Signature::from_bytes(&[byte; _]))
+}
+
+fn header(byte: u8) -> Header {
     Header::new(
-        HeaderId::from([0x11u8; 32]),
-        ContentId::from([0x22u8; 32]),
-        Slot::from(42u64),
+        HeaderId::from([byte; 32]),
```
ntn-x2: Can we use different values for different fields, so we actually test their encodings?

ntn-x2 (approved):
I think `RawGenesisBlock` should also deserialize the new uncle references field. Other than that, the PR looks ok. There's still a bunch of `block_root` references to update.

core/src/block/uncle.rs
youngjoon-lee: In this PR, this type doesn't have any methods. But, in next PRs, we add several functions to the `UncleHeaders`. So, I think it's good to have this type. I forgot to add a PR comment on this file.

core/src/block/fixtures/mod.rs
youngjoon-lee: Thank you for your review. I applied your suggestions:
- fixture: e22631bfe48337490dfd4ec8b0ed1b703ae91e23
- deserialize `uncle_headers` for `RawGenesisBlock`: 5f0f470e7eb3802ca4d7ae5875a01293132fd9a0
- change comments: `block_root` -> `body_root`: 9d5b4d6094561ad535f0a26fb1b519510ec818d5

blend/message/src/message/payload.rs
```diff
@@ -5,7 +5,7 @@ use serde_with::serde_as;
 
 use crate::Error;
 
-pub const MAX_PAYLOAD_BODY_SIZE: usize = 8555;
+pub const MAX_PAYLOAD_BODY_SIZE: usize = 10000;
```
hansieodendaal, line 18: Maybe we should add some reasoning in a doc comment about this `MAX_PAYLOAD_BODY_SIZE = 10000` value.

hansieodendaal (commented):
Looking good overall! 

I think we may be missing a few follow-ups before this is complete (_if not done in the consecutive PRs, then please ignore_):

1. **Preserve uncle headers through public block/API serialization.**  
   `core::Block` and `Proposal` now carry `uncle_headers`, but the public API path still appears to serialize only the header and transactions.

2. **Make `RawGenesisBlock` faithfully deserialize the new field.**  
   In `core/src/block/genesis.rs`, `RawGenesisBlock` does not currently deserialize `uncle_headers`; the reconstructed block instead always inserts `UncleHeaders::empty()`.

3. **Clarify or enforce uncle signature and eligibility validation.**  
   `core/src/block/uncle.rs` introduces `SignedHeader`, while `Block::into_verified()` validates the block signature and the `body_root`. However, it does not appear to verify the signatures or protocol eligibility of the carried uncle headers themselves. 

4. We have good coverage for encoding, body-root binding, tampering, ordering, and maximum proposal size. I think we still need tests for the public API representation, _maybe_ rejecting non-empty uncle headers in genesis, and - once implemented - verification of the uncle headers’ own signatures/eligibility.

youngjoon-lee:
> Looking good overall!
> 
> I think we may be missing a few follow-ups before this is complete (_if not done in the consecutive PRs, then please ignore_):

Thank you @hansieodendaal. I just addressed your comments.

1. Yes. I added the `uncle_headers` field to the `ApiBlock`: 2e9ad930505022ec2163cf6a3152863fe9703e16. I didn't add a test for the round-trip serde of `ApiBlock`, because two different/similar structs are defined in both `lb-node` and `lb-common-http-client`. It's not easy to write tests in a future-safe way. I will open a separate PR to refactor those structs for safe testing.

2. Already resolved by 5f0f470e7eb3802ca4d7ae5875a01293132fd9a0 - https://github.com/logos-blockchain/logos-blockchain/pull/3281#pullrequestreview-4883521508

3. Implemented in the PR4 https://github.com/logos-blockchain/logos-blockchain/pull/3284

4. Added a test: 8d00d6cdb6029192c83300d8c72b704a85ea5b81  Genesis block rejects non-empty uncle refs.

blend/message/src/message/payload.rs
youngjoon-lee: Thanks! 92e7407e96fcf892dc086f5bdab1f1797f7f8516

davidrusu: fyi, tx references are changing to 16 bytes (up from 8 bytes) https://github.com/logos-co/logos-lips/pull/389

hansieodendaal (approved):
LGTM!

davidrusu (approved):

PR #3282, opened 2026-08-07
feat(consensus): uncle-ref: add the uncle reference window to the config

youngjoon-lee:

This is the 2nd PR for the [Uncle Reference RFC](https://github.com/logos-co/logos-lips/pull/385). 

- Add the uncle reference window `w_u` to the Cryptarchia config. 
- Add the window absorption parameter `W` to the deployment setting, and derive `w_u` from it. 

Yes. Please see the RFC I linked above. 

Yes for this part.

No.


consensus/cryptarchia-engine/src/config.rs
```diff
@@ -13,6 +16,9 @@ pub struct Config {
     /// `f`, the rate of occupied slots
     slot_activation_coeff: NonNegativeRatio,
     stake_inference_learning_rate: NonNegativeF64,
+    /// `w_u`, the maximum number of slots by which a referenced uncle may
+    /// precede the block referencing it.
+    uncle_reference_window: NonZero<u64>,
```
davidrusu: Lets change this to be `W` the average number of blocks by which a referenced uncle's parent may ...

and then we can derive the slot window from uncle_reference_window + lottery_constants when we need it.

youngjoon-lee: We already have `W` in the deployment config: https://github.com/logos-blockchain/logos-blockchain/pull/3282/changes#diff-c1202c3ad20a3e266adf585e7a74245739473560b772c7326c49c9809d03e25aR23. 

And, when we build a Cryptarchia config here, we derive `w_u` from `W/f`. This Cryptarchia config holds only `w_u` because we don't need `W` after the derivation.

That is, we let users configure only `W`.

davidrusu: right, I saw that, but I think we should keep W here as well, and derive w_u when you need it in verification and block building.

Basically, don't pre-compute w_u, just use W everywhere in all configs. (ideally we would read this directly from deployment config if possible to avoid repeating it in these service configs)

We are making a similar change in the spec to drop `w_u`, we will only have `W` defined.

youngjoon-lee: Ah, I just read https://github.com/logos-co/logos-lips/pull/385#discussion_r3773914528. Okay, it makes sense. I'll change the code.

youngjoon-lee: 7e155ad028c6068a03c2611d04fd0b3c43f63796 Now we don't have `w_u` in the config. We have only `W` and derive `w_u` only when we need it.

Also, I suggested to rename `window_absorption_parameter` (`W`) to `uncle_reference_window_in_block` in the RFC: https://github.com/logos-co/logos-lips/pull/385#discussion_r3779572992. If it's accepted, I will change this code even after this PR is merged.

nodes/node/binary/src/config/cryptarchia/deployment.rs
```diff
@@ -46,12 +48,22 @@ impl Settings {
         .get()
     }
 
+    /// # Panics
+    /// If `W` exceeds `0.6k`, which would put the uncle reference window
+    /// outside the finalization window.
     #[must_use]
     pub fn consensus_config(&self) -> ConsensusConfig {
+        let max = (0.6 * f64::from(self.security_param.get())).floor() as u32;
```
hansieodendaal: We could change this to integer arithmetic (`3 * k / 5`), and add explicit boundary tests (accepted vs. rejected) for the `W` constraint.

hansieodendaal (approved):
LGTM!

One non-blocking comment below.

ledger/src/cryptarchia/mod.rs
```diff
@@ -926,6 +926,7 @@ pub mod tests {
             NonZero::new(1).unwrap(),
             NonNegativeRatio::new(1, 10.try_into().unwrap()),
             1f64.try_into().expect("1 > 0"),
+            NonZero::new(100).unwrap(), // W = 10 at f = 1/10
```
davidrusu: 12?

nodes/node/binary/src/config/cryptarchia/deployment.rs
davidrusu: Actually, even easier, this constraint has just been dropped from the spec in https://github.com/logos-co/logos-lips/pull/385/changes/d2c3009e9714908c18b008eccec4fb30c624bb0d

@youngjoon-lee lets remove this assert entirely, since this constraint does not exist anymore

services/chain/chain-leader/src/leadership.rs
```diff
@@ -728,6 +728,7 @@ mod pol_tests {
                 NonZero::new(5).unwrap(),
                 NonNegativeRatio::new(1, 10.try_into().unwrap()),
                 1f64.try_into().expect("1 > 0"),
+                NonZero::new(100).unwrap(), // W = 10 at f = 1/10
```
davidrusu: 12?

wallet/src/lib.rs
```diff
@@ -1837,6 +1837,7 @@ mod tests {
             NonZero::new(1).unwrap(),
             NonNegativeRatio::new(1, 10.try_into().unwrap()),
             1f64.try_into().expect("1 > 0"),
+            NonZero::new(100).unwrap(), // W = 10 at f = 1/10
```
davidrusu: 12

nodes/node/binary/src/config/cryptarchia/deployment.rs
youngjoon-lee: Ah, I was confused by the “virtual upper bound” written in the RFC change. Sorry. I just removed this assertion: 01a4a2294f15c1b2399ad033f926b7d67fed115b

ledger/src/cryptarchia/mod.rs
youngjoon-lee: Ahhhh, sorry. I made leftovers.. I didn’t check these files because these are tests that are not directly related to the uncle-ref. Nevertheless it is better to use 12. I just modified all files: 0568521e6477d8bc2c7ae4315674e18056c9f5aa

davidrusu (approved):
⛵ nice

PR #3283, opened 2026-08-07
feat(chain): uncle-ref: select uncles when proposing a block

youngjoon-lee:

This is the 3rd PR for the [Uncle Reference RFC](https://github.com/logos-co/logos-lips/pull/385). 

This PR is the block-proposer-side implementation. A block proposer can select/include uncles to a new block that he proposes. 

This PR implements the way how uncles are selected. The rule is defined in the RFC. Also, in a next PR, block validators will verify uncles by following some part of the same rule. 

- Chain leader service selects uncles when proposing a block. 
  - I left detailed doc comments to explain how uncles are selected. I believe, those comments explain the logic clearly. 
- The cryptarchia-engine `Branch` tracks `uncle_slots`, so occupied slots are known for selection.

A next PR would be the block-validator-side implementation: verifying uncle references included in a block when validating the block.

## 2. Does the code have enough context to be clearly?

Yes. See the `Uncle Selection` part in the RFC. 

The selection follows the revised validity rule agreed with @madxor (`0 < A.slot - U.parent.slot <= w_u`), which bounds all walks by the window. The spec update is pending simulation results.

Yes, the parent-window rule above; agreed with @madxor, spec update pending.

### DIFF
--- a/consensus/cryptarchia-engine/src/lib.rs
+++ b/consensus/cryptarchia-engine/src/lib.rs
@@ -10,11 +10,15 @@ use std::{
 };
 
 pub use config::*;
+use lb_utils::bounded::UpperBoundedVec;
 use rpds::{HashTrieMapSync, HashTrieSetSync};
 use thiserror::Error;
 pub use time::{Epoch, EpochConfig, Slot};
 
 pub(crate) const LOG_TARGET: &str = "cryptarchia::engine";
+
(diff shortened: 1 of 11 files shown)


davidrusu (approved):

hansieodendaal (approved):
LGTM

I liked the illustrative unit tests!

PR #3284, opened 2026-08-07
feat(chain): uncle-ref: verify uncles when applying a block

youngjoon-lee:

This is the 4th PR for the [Uncle Reference RFC](https://github.com/logos-co/logos-lips/pull/385). 

In a previous PR https://github.com/logos-blockchain/logos-blockchain/pull/3283, we implemented selecting/including uncles to a new block proposal. This PR implements verifying those uncles. If a block contains uncles that don’t meet criteria, it must be rejected, so that we can converge to the consistent total stake, resulting in the consistent epoch nonce in the end. 

- Verify uncles when applying a block. Reject the block if any of uncles are invalid. 
  - I left comments to explain how the validation works, as simple as possible. I believe that is sufficient to understand the logic. 
- This PR doesn’t touch TSI. 

A next PR is the final one of this series. It will update TSI to account for uncles. 

Please see the `Block Header Validation` part in the RFC. 

This PR implements the revised rule agreed with @madxor (`0 < A.slot - U.parent.slot <= w_u`); the spec update is pending simulation results.

Yes, the parent-window rule above; agreed with @madxor, spec update pending.

### DIFF
--- a/services/chain/chain-service/src/uncle.rs
+++ b/services/chain/chain-service/src/uncle.rs
@@ -1,0 +1,463 @@
+//! The uncle validity rules
+
+use std::collections::HashSet;
+
+use lb_core::{
+    block::{Block, SignedHeader, UncleHeaders},
+    codec::SerializeOp as _,
+    header::HeaderId,
+    proofs::leader_proof::LeaderProof as _,
+};
+use lb_cryptarchia_engine::Branch;
+
+use crate::{Cryptarchia, Error};
+
+#[expect(
+    clippy::multiple_inherent_impl,
+    reason = "grouping the uncle validity rules separately from the main impl"
+)]
+impl Cryptarchia {
+    /// Verifies every uncle carried by the block against the chain the block
+    /// extends.
+    pub(crate) fn verify_uncles<Tx>(&self, block: &Block<Tx>) -> Result<(), Error> {
+        if block.uncle_headers().is_empty() {
+            return Ok(());
+        }
+
+        let header = block.header();
+        let slot = header.slot();
+        let parent = self
+            .consensus
+            .branches()
+            .get(&header.parent())
+            .ok_or_else(|| Error::ParentMissing {
+                parent: header.parent(),
+                info: Box::new(self.info()),
+            })?;
+
+        // Each uncle must be strictly older than the block. The window rule
+        // `0 < slot - uncle_parent_slot <= w_u` is enforced by the ancestry
+        // verification below, and it implies that the uncle itself is also
+        // within the window since an uncle is newer than its parent.
+        for uncle in block.uncle_headers().iter() {
+            if uncle.header().slot() >= slot {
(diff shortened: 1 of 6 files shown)


services/chain/chain-service/src/uncle.rs
```diff
@@ -0,0 +1,470 @@
...
+                    uncle: uncle.header().id(),
+                    reason: UncleError::ParentNotOnChain,
+                });
+            }
+        }
+        Ok(())
+    }
+
+    /// Verifies the signature and the leadership proof carried by an uncle.
+    fn verify_uncle_proofs(&self, uncle: &SignedHeader) -> Result<(), UncleError> {
```
davidrusu, line 126: feels odd that this is not just a `verify_header()` call. It should be identical verification as a normal header no?

davidrusu (approved):
Looks clean! One question about consolidating uncle header verfication with normal header verification if possible.

Also, not part of this PR, but do we intend to trigger orphan backfilling for missing uncles?

youngjoon-lee:
> Also, not part of this PR, but do we intend to trigger orphan backfilling for missing uncles?

Thank you for your review! 

We don't need to trigger backfilling. Actually we changed RFC last week to avoid backfilling: https://github.com/logos-co/logos-lips/pull/385#discussion_r3702568253.
The RFC says that only the 1st block of a branch (after LCA) can be selected/verified as an uncle. 
So, we just check that an uncle's parent is on our local chain, and check if the uncle's PoL is valid. By doing so, we can make sure that the uncle was generated by a real lottery win based on its parent (which really exists).

Because of this RFC change, less uncles will be chosen, but Marcin ran simulation to check that the number of uncles is still sufficient after this change.

services/chain/chain-service/src/uncle.rs
youngjoon-lee: I initially thought about it. But, we don't have any function that exactly matches with `verify_header()` you mentioned. Verifications are mixed and also separated in the current codebase:
- we have [`Block::into_verified`](https://github.com/logos-blockchain/logos-blockchain/blob/7d696f5243dcd7618acb4e9371a68ce2855c131f/core/src/block/mod.rs#L240) which verifies header and body all together because we need the body to validate `Header::body_root`. This is the stateless check.
- we validate PoL when applying the block to the ledger because we need the `parent_state`. This is the stateful check.

So, let's do it like this. I will open another PR that refactors the validations, to reuse functions as much as I can, to avoid future bugs. I think it's better to do it in a separate PR because it may be little verbose.

youngjoon-lee: Opened a separate PR on top of this stack: https://github.com/logos-blockchain/logos-blockchain/pull/3314

davidrusu:
> We don't need to trigger backfilling. Actually we changed RFC last week to avoid backfilling

nice change, yes that's simpler

PR #3285, opened 2026-08-07
feat(ledger): uncle-ref: include uncles to TSI

youngjoon-lee:

This is the 5th (last) PR for the [Uncle Reference RFC](https://github.com/logos-co/logos-lips/pull/385). 

Previously, the TSI (Total Stake Inference) counted the number of blocks in the honest chain within the first `6*k/f` slots. 
Now, the TSI counts the number of distinct occupied slots:
- When applying a block to the ledger, now we accumulate the slot of the block and its uncles to the `BlockDensity`, so that the TSI can account for uncles.
- TSI doesn’t need to verify each uncle because they are verified when the block is validated.

Please see the `Epoch State Pseudocode` and the TSI part of the RFC. 

Yes

No

### DIFF
--- a/services/chain/chain-leader/src/lib.rs
+++ b/services/chain/chain-leader/src/lib.rs
@@ -609,9 +609,23 @@ where
 
         let tx_stream: Pin<Box<_>> = Box::pin(txs_stream);
 
+        let uncle_headers = cryptarchia_api
+            .select_uncles(parent, slot)
+            .await
+            .unwrap_or_else(|err| {
+                error!(target: LOG_TARGET, ?slot, %err, "failed to select uncles");
+                // A proposal without uncles is still valid
+                UncleHeaders::empty()
+            });
+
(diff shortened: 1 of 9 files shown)


services/chain/chain-leader/src/lib.rs
```diff
@@ -609,9 +609,23 @@ where
...
 
+        let uncle_headers = cryptarchia_api
+            .select_uncles(parent, slot)
+            .await
+            .unwrap_or_else(|err| {
+                error!(target: LOG_TARGET, ?slot, %err, "failed to select uncles");
+                // A proposal without uncles is still valid
+                UncleHeaders::empty()
+            });
+
```
youngjoon-lee, line 620: This function call was already added by a previous PR #3283 at line 673 below. But, this PR moved it to here at line 612, because we need the returned `uncle_headers` at line 623.

davidrusu (approved):
nice!

PR #3286, opened 2026-08-07
test: add future dated fee projections

hansieodendaal:

Add future-dated transaction fee projections for Cucumber wallet transactions.

Changes:
  - Add transaction fee horizon and policy types.
  - Project storage gas prices across future epochs using the protocol’s `9/8` maximum increase.
  - Track the epoch and chain tip at which fees were prepared.
  - Add configurable epoch headroom for continuous transaction scenarios.
  - Apply projected fee ceilings when calculating transaction fees.
  - Validate prepared transactions against the live epoch and gas prices before submission.
  - Add clearer fee-horizon and fee-expiry errors.
  - Improve manual transaction flows, wallet funding, and transaction bookkeeping.
  - Update transaction scenarios to exercise fee headroom across price changes.
  - Add unit coverage for projection rounding, zero headroom, overflow, and horizon construction.
  - Isolate scenario wallet keys from externally configured node wallets.
  - Remove only scenario-specific keys from node `known_keys` and KMS configuration.
  - Preserve unrelated node, voucher-master, and funding wallets.
  - Load effective epoch configuration from the deployed node settings.
  - Extend Cucumber diagnostics with node wallet, epoch, and fee-policy state.

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


andrussal:
I’m wondering if we could achieve the desired fee safety more locally by fetching the current prices at the start of each transaction batch, applying the storage-price headroom directly when preparing it, and rebuilding if the epoch changes before submission. Would that let us avoid threading the fee policy through the wallet stack and storing it globally on CucumberWorld?

hansieodendaal:
> I’m wondering if we could achieve the desired fee safety ...

The issue we have is that with these large, long-running scenarios, where we need to burst-submit thousands of transactions, transaction preparation time is measured in hours; to start again if the projected epoch headroom changed is not feasible. When transactions are burst-submitted, all of them must succeed together.

andrussal:
> transaction preparation time is measured in hours; to start again if the projected epoch headroom changed is not feasible

Ah, I see. Given that constraint, I think this approach is justified. Thanks for clarifying.

tests/src/common/wallet/transaction/signed.rs
```diff
@@ -1,32 +1,49 @@
...
 pub struct SignedWalletTransaction {
     signed_tx: SignedMantleTx<Preverified>,
     tx_hash: TxHash,
     reserved_inputs: WalletReservedInputs,
-    spent_fee: u64,
+    paid_fee: u64,
+    mandatory_fee_at_preparation: u64,
+    prepared_at_tip: Option<HeaderId>,
+    prepared_at_epoch: Option<u32>,
+    valid_through_epoch: Option<u32>,
```
andrussal: `TransactionFeeHorizon` already encapsulates these values. Would it make sense for `SignedWalletTransaction` to keep an `Option<TransactionFeeHorizon>` instead of copying the individual fields? I also couldn’t find consumers of the current getters, so perhaps the duplicated fields can be removed entirely if per-transaction horizon metadata is not needed.

tests/src/cucumber/world.rs
```diff
@@ -888,6 +897,12 @@ pub struct CucumberWorld {
     pub node_provisioned_wallet_pks: HashSet<ZkPublicKey>,
     /// Manual: Scenario-level fee sponsor configuration and accounting.
     pub fee_state: ScenarioFeeState,
+    /// Fee policy shared by all transactions prepared in the current
+    /// long-running cycle.
+    pub transaction_fee_policy: Option<TransactionFeePolicy>,
+    /// Explicitly configured epoch headroom for long-running transactions.
+    /// Zero means use the live fee prices without epoch projection.
+    pub transaction_epochs_headroom: u32,
```
andrussal: Could the fee policy remain cycle-local and be passed through preparation and validation explicitly? Both `transaction_epochs_headroom` and `transaction_fee_policy` appear to belong to one cycle, while storing them in CucumberWorld requires manually setting and clearing shared state.

andrussal (approved):
LGTM

(Left 2 comments, nothing blocking tho)

tests/src/common/wallet/transaction/signed.rs
hansieodendaal: Changes implemented in https://github.com/logos-blockchain/logos-blockchain/pull/3286/commits/89dcdbd4a76fbcc6de137e23491b7e4d1125f0a6

tests/src/cucumber/world.rs
hansieodendaal: Changes implemented in https://github.com/logos-blockchain/logos-blockchain/pull/3286/commits/89dcdbd4a76fbcc6de137e23491b7e4d1125f0a6

andrussal (approved):
LGTM

PR #3287, opened 2026-08-08
fix(wallet): wallet/fund during non-finalized window hands double note

pradovic:
This might be related to testnet inscription production stop, as a possible cause (needs to be confirmed from testnet logs).

2 integration tests added to illustrate the the call handing double notes:
1. Soaking wallet with fund calls over non-finalized window over forks produces 2 txs with the same note -> main failure path 
2. Node restart loses reservation state -> restart scenario could also produce multiple tx with the same note

Reason: 

It is by current design of the /wallet/fund call:
The reservation map is one flat, branch-agnostic set — services/wallet/src/states.rs:130-132:
```

struct PendingNotes {
    notes: HashMap<NoteId, u64>,
}
```

The wallet keeps ledger state per branch (prune_states), but reservations have no branch dimension.

Every processed block — any branch — releases destructively. The service subscribes to all new blocks (lib.rs:408, 456-457 → handle_new_block → state.apply_block), and apply_block does (states.rs:302-307):
```
pub fn apply_block(&mut self, block: &WalletBlock) -> Result<(), WalletError> {
    self.wallet.apply_block(block)?;
    self.pending_notes.remove_spent(&block.spent_note_ids()); // -> this removes global reservation, but allowed utxos differ between branches                       
    ...
}
```

The doc comment: states.rs:120-128: "funded notes are released as soon as they are observed spent in a tip block... Releasing too early merely reintroduces a conflicting tx, the same failure mode as having no reservation at all." says the same thing.

Zone-sdk is using `/wallet/fund` call in sequential manner over non-finalized window, so it could get the same note for the different transaction which would stop the lineage chain.  

In the test 1, this is manifested by 2 transactions producing the same tx_id (as the amount is the same), so they would end up as one x in the end. 

The troubleshoot the logs in this case:
`grep "Reserved pending note" log.log | grep note_number` + pair it with `InexistingNote` eviction from logs as well.

Suggested fixes:
1. Don't remove reservation when observed on chain and rely only on existing `﻿﻿﻿pending_note_expiry_blocks ` timeout - Done in this PR.
2. Add timeout based refunding of the stale txs to zone-sdk or add storage state to wallet, so it survives restart handing double notes and extreme cases of too early timeout from the node - left for future, currently unguarded.
﻿

### DIFF
--- a/services/wallet/src/states.rs
+++ b/services/wallet/src/states.rs
@@ -159,10 +159,6 @@ impl PendingNotes {
         self.notes.keys().copied().collect()
     }
 
-    fn remove_spent(&mut self, spent: &HashSet<NoteId>) {
-        self.notes.retain(|note_id, _| !spent.contains(note_id));
-    }
-
     /// Evict reservations whose LIB-progress age reached the configured limit.
     ///
     /// Each LIB update adds `new_immutable_blocks_count` to every reservation's
@@ -301,7 +297,8 @@ impl<'u> ServiceState<'u> {
 
     pub fn apply_block(&mut self, block: &WalletBlock) -> Result<(), WalletError> {
         self.wallet.apply_block(block)?;
-        self.pending_notes.remove_spent(&block.spent_note_ids());
+        // Note reservations are released after `pending_note_expiry_blocks`
+        // (see `evict_expired`), not on spent.
         self.update_state();
         Ok(())
     }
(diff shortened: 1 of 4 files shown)


andrussal (approved):
LGTM

I think this is the right quick fix. A more robust long-term solution would be for Zone SDK to rebuild stale transactions with fresh funding and current fees, which would also handle transactions stuck because their fees are too low.

youngjoon-lee (approved):
🙏 thank you.

I agree, this is a quick solution to mitigate the issue, but doesn't eliminate the issue completely. We probably shouldn't rely on `pending_note_expiry_blocks=10`, even if it's large.
If a node submitted a `tx1` with a note `N` to the mempool, but if `tx1` is not selected by block proposers while 10+ blocks are proposed, then the note `N` will be released from the `PendingNotes` in the node, and the node can select `N` again for a new `tx2`. Then, `tx1` and `tx2` are collapsed. 

Let me think about any solution that is more ultimate.

PR #3288, opened 2026-08-10
feat: add epoch headroom to fee projections built into the node

hansieodendaal:

See [this discussion](https://discordapp.com/channels/973324189794697286/1536655245403361363) for background.

`epoch_headroom` is the amount of future chain time from transaction preparation for which the wallet provisions projected mandatory-fee increases. For example, 0.2 means one fifth of an epoch of lookahead; it does not create transaction expiry.

* Add fixed-point `epoch_headroom` fee projection with integer-tenths semantics.
* Project storage prices across actual epoch boundaries using per-boundary `ceil(price × 9 / 8)`.
* Project execution prices using the current EMA, target-load block assumptions, and shared ledger arithmetic.
* Expose fee policies through public funding and transfer APIs and return fee-horizon quote metadata.
* Preserve legacy `priority_fee` behavior and enforce `max_tx_fee`.
* Integrate fee-horizon funding into Zone SDK transactions.
* Improve Zone Cucumber reliability with a persistent indexer and epoch-transition diagnostics.
* Add documentation and focused coverage.
* Added cucumber: `Scenario: Fee horizon keeps a funded transaction valid across an epoch price increase`

**API**

```json
{
  "fee_policy": {
    "epoch_headroom": 1.3,
    "priority_fee": 500
  }
}
```

`epoch_headroom` is stored internally as integer tenths. Extra decimal precision is truncated, not rounded (`1.39 -> 1.3`). Negative, non-finite, overflowing, or effective values above `100.0` are rejected.

Legacy funding requests using top-level `priority_fee` remain supported. Ambiguous simultaneous priority-fee configuration is rejected.

**Implementation notes**

* Fee projection is resolved in the wallet service from one preparation tip.
* `MantleTxBuilder` remains chain-context independent.
* Storage projection is deterministic across covered epoch transitions.
* Execution projection is explicitly an estimate, not a guaranteed ceiling.
* Zone SDK delegates fee projection to the wallet via `fee_policy`.
* Cucumber Zone assertions reuse a persistent read-only indexer, and existing consensus queries can emit epoch/gas-price diagnostics.

**Documentation**

Added `docs/fee-horizon.md` covering semantics, arithmetic, API usage, compatibility, assumptions, and guarantee boundaries.

Yes

@hansieodendaal 

Yes

No

## Checklist

> [!WARNING]  

* [X] 2. Description added.


docs/fee-horizon.md
```diff
@@ -0,0 +1,133 @@
+# Fee horizon projection
+
+`epoch_headroom` is the amount of future elapsed chain time, measured in epochs
```
youngjoon-lee, line 3: It'd be helpful if you add a short definition of epoch headroom in the PR description. I couldn't understand most of the contents in the PR description before reaching to this file (which is mentioned only in the last sentence of the PR description).

core/src/mantle/gas.rs
```diff
@@ -65,6 +69,378 @@ impl From<Value> for GasPrice {
...
+    Negative,
+    /// The supplied numeric value was NaN or infinite.
+    #[error("epoch_headroom must be finite")]
+    NonFinite,
+    /// The supplied value exceeded the explicit input bound.
+    #[error("epoch_headroom exceeds the maximum of 100.0 epochs")]
+    TooLarge,
+}
+
+fn parse_decimal_tenths(value: &str) -> Result<u16, EpochHeadroomError> {
```
youngjoon-lee, line 207: It'd be great if we add a doc comment for this function.

youngjoon-lee: Instead of implementing this long function, is there any crate we can use? ~~For example, `serde_json` might have a function to parse a number which is either an integer or a float.~~

services/wallet/src/fee.rs
```diff
@@ -0,0 +1,293 @@
...
+    let slots_per_epoch = epoch_config.epoch_length(consensus_config.base_period_length());
+    let prepared_at_slot = ledger.slot();
+    let prepared_at_epoch = ledger.epoch_state().epoch;
+    let horizon_slots = horizon_slots(headroom, slots_per_epoch)?;
+    let valid_until_slot = Slot::from(
+        u64::from(prepared_at_slot)
+            .checked_add(horizon_slots)
+            .ok_or(FeeProjectionError::SlotOverflow)?,
+    );
+    let valid_until_epoch: Epoch = (u64::from(valid_until_slot) / slots_per_epoch)
```
youngjoon-lee: This is correct, but we can use `epoch_config::epoch(slot, consensus_config.base_period_length())`.

zone-sdk/src/sequencer/types.rs
```diff
@@ -139,11 +139,19 @@ pub struct FundingConfig {
...
 }
 
 impl FundingConfig {
     /// Default execution tip.
     pub const DEFAULT_PRIORITY_FEE: Value = 200;
+    /// Default fee horizon used by Zone SDK transactions.
+    pub const DEFAULT_EPOCH_HEADROOM: EpochHeadroom = match EpochHeadroom::from_tenths(10) {
+        Ok(headroom) => headroom,
+        Err(_) => panic!("1.0 epoch is a valid fee horizon"),
+    };
```
youngjoon-lee, line 154: It seems that we don't have a CLI arg which uses this default value. I think we need something similar as: https://github.com/logos-blockchain/logos-blockchain/blob/f31dc50b179c8d2648e3ae6b61b79edf4f860f04/deployment/tui-zone/src/cli.rs#L106-L113

core/src/mantle/gas.rs
```diff
@@ -65,6 +69,378 @@ impl From<Value> for GasPrice {
...
+}
+
+/// Applies the same execution-market recurrence as [`update_execution_market`]
+/// using the ledger's historical `u128 as u64` conversion semantics.
+///
+/// Consensus state has historically relied on this infallible conversion. The
+/// checked variant is used by fee projection so an input that exceeds the
+/// representable gas-price range is reported instead of silently truncated.
+#[must_use]
+pub fn update_execution_market_for_ledger(
```
youngjoon-lee, line 119: We are duplicating the same algorithm that the `update_execution_market` already has. Can we probably have a common private function that return `(u128, u128)`, so that each public function can convert the return values properly?

youngjoon-lee: Or, what do you think about having only the `update_execution_market` function? I don't think it's good to rename this function with the `_for_ledger` suffix, which is a downstream of this crate. 

Then, I think, the ledger can panic if `GasOverflow` happens. But I'm not 100% sure. The previous & current implementation silently wraps the value when casting `u128` into `u64`. If this wrapping is correct, I think we should rename this function to `update_execution_market_wrapping` for something better. 

@thomaslavaur Could you please confirm that if the ledger should proceed even if the updated execution market fee exceeds `u64::MAX`? Or, is it okay to panic in this case? I think, panicking is a right choice. The wrapping converts `u64::MAX + 1` into `0`, which I guess it's not a right behavior we intend.

Actually, this topic is out of scope for this PR.

services/wallet/src/fee.rs
```diff
@@ -0,0 +1,293 @@
...
+        execution_base_gas_price: projected_execution,
+        storage_gas_price: projected_storage,
+    };
+    let live_context = ledger.tx_context();
+    let projected_context = MantleTxContext {
+        gas_context: live_context
+            .gas_context
+            .with_gas_prices(projected_prices.clone()),
+        leader_reward_amount: live_context.leader_reward_amount,
+    };
```
youngjoon-lee: I think, this can be simplified?
```suggestion
    let projected_context = live_context.clone_with_live_prices(projected_prices);
```

docs/fee-horizon.md
youngjoon-lee: Also, is this a decision already discussed in Discord or anywhere else? If so, it would be great to put a link in the PR description.

core/src/mantle/gas.rs
thomaslavaur: Yes panicking is ok. I think I updated mantle last week to explicitely explain the type conversion and how it should react!

thomaslavaur: yes exactly. You can check the last version of mantle it's
```python
tx_mandatory_fee = mandatory_gas_fees(signed_tx)  # int128
tx_balance = get_transaction_balance(signed_tx)   # int128
assert tx_mandatory_fee <= tx_balance
tx_priority_tip = checked_uint64(tx_balance - tx_mandatory_fee)
```

and 

```python
def checked_uint64(value: int) -> TokenValue:
        assert 0 <= value <= UINT64_MAX
        return value
```

youngjoon-lee: Thank you for your confirmation. I opened an issue https://github.com/logos-blockchain/logos-blockchain/issues/3295. I can pick it up after finishing some other urgent tasks, if no one else has picked it up by then.

docs/fee-horizon.md
```diff
@@ -0,0 +1,133 @@
...
+`100.09` is accepted as `100.0`, while `100.10` truncates to `100.1` and is
+rejected because the maximum is `100.0`. Negative and non-finite values remain
+invalid.
+
+For a preparation tip with slot `s`, and `E` slots per epoch, the wallet
+calculates the slot horizon as:
+
+```text
+horizon_slots = ceil(epoch_headroom_tenths * E / 10)
+valid_until_slot = s + horizon_slots
```
youngjoon-lee, line 18: When reading this paragraph, it's not easy to guess the purpose of `valid_until_slot`. I think, we need a short explanation about why this variable is introduced in this paragraph, at least.

From what the next chapter says, it seems that this value is only for counting the number of epoch boundaries that the tx can endure. Then, I think we can just directly use `s + horizon_slots` in the next chapter. For me, `valid_until_slot` sounds like: the tx is going to become invalid strictly after the `slot`.

docs/fee-horizon.md
hansieodendaal: Applied review comments in 974b68ac01864b4ac81ba0cc8cc9d1182a8bf1e4

core/src/mantle/gas.rs
hansieodendaal: Applied review comments in 974b68ac01864b4ac81ba0cc8cc9d1182a8bf1e4

services/wallet/src/fee.rs
hansieodendaal: Applied review comments in 974b68ac01864b4ac81ba0cc8cc9d1182a8bf1e4

zone-sdk/src/sequencer/types.rs
hansieodendaal: Applied review comments in 974b68ac01864b4ac81ba0cc8cc9d1182a8bf1e4

services/wallet/src/fee.rs
hansieodendaal: Applied review comments in 974b68ac01864b4ac81ba0cc8cc9d1182a8bf1e4

docs/fee-horizon.md
hansieodendaal: Applied review comments in 974b68ac01864b4ac81ba0cc8cc9d1182a8bf1e4

